PR #16, opened 2014-12-20
Add frequently misspelled UK shops

math1985:
See also this thread on the talk-gb mailing list:
https://lists.openstreetmap.org/pipermail/talk-gb/2014-December/016922.html


jfirebaugh:
Committed (with conservative omission of Thompson) as 375f18029677cae4fd60432bcea573026034e1fb, thanks!

PR #20, opened 2015-05-16
Add some German nouns which are not names

Nakaner:
Eisdiele is like Eiscafe a type of POI not a name. Imbiss is a German noun for a type of fast-food POI. Bäcker is another noun for Bäckerei (bakery) and no name.


jfirebaugh:
Committed as c933f26b8bf24116fcf7851594e0bcea407e31c1, thanks!

PR #22, opened 2015-05-16
Deichmann is a shoe shop, no clothes shop

Nakaner:
Deichmann is a shoe shop, no clothes shop.


RicoElectrico:
I'm afraid you would have to fix it in OSM, as this particular file is generated automatically from planet pbf and will get overwritten at build.

jfirebaugh:
Yep, this needs to be corrected in the source data (OSM).

PR #23, opened 2015-05-16
Tiefgarage is an underground garage, not a name

simonpoole:
See subject


jfirebaugh:
Committed as f06ee3b0ef1d0cde8827d05265acfdbf9fe29602, thanks!

PR #26, opened 2015-10-23
Modified existing entries add added new ones

ghost:
Dunkin' Donuts is the correct spelling. Removed nix value fast food.

Tim Hortons: Removed nix value cafe

AFAIK there is no consenus between cafe and fast_food for many chain coffee shops. Especially since many have drive-through service and serve sandwiches. Until there is a consensus I don't think it makes sense to set them arbitrarily.

Whole Foods Market: swapped values to correct

7-Eleven: added match 7/11, a default that shows up before 7-Eleven in iD. Or is there another convenience store chain called 7/11?

Lowes: The correct name for the DIY store is Lowe's, but there are other stores such as Lowes Foods and Lowe's Market so canonicalization is troublesome.

Panera Bread: remove nix value fast_food. Paying for the food first is the major distinction between fast_food and restaurant.

Pizza Hut: remove nix value fast_food. See above.

Baskin-Robbins: correct spelling includes a "-"

### DIFF
--- a/canonical.json
+++ b/canonical.json
@@ -150,12 +150,9 @@
             "Costa Coffee"
         ]
     },
-    "Dunkin Donuts": {
+    "Dunkin' Donuts": {
         "matches": [
-            "Dunkin' Donuts"
-        ],
-        "nix_value": [
-            "fast_food"
+            "Dunkin Donuts"
         ],
         "tags": {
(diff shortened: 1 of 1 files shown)


zeromap:
@jfirebaugh Would you take a look at this when you get a chance? Thanks.

PR #27, opened 2015-11-25
Add some nouns that are being misused as names to the filter

simonpoole:
The additions to the list have been checked against their geographic use to avoid mistakenly removing foriegn language nouns that are being used as chain names.


simonpoole:
For whatever reson this was based on the wrong branch.

PR #33, opened 2016-06-25
Improve handling of "Netto"

simonpoole:
Netto are two different chains in Germany that have nothing to do with eachother, wit:

https://netto.de/

and

https://www.netto-online.de/Filial-Angebote.chtm


math1985:
Thanks!

math1985:
cosmetics and perfumery are not yet included in the output, they will be included next time the script is run.

@bhousel Question, does iD take the generated .json file from this repository, or does it run the script itself?

simonpoole:
@math1985 the topNames.json file should be regenreated too, while that doesn't have to happen -so- often, tagging changes and new stuff gets mapped (not to mention chains changing their name and going bust and and and). 

I've typically updated it one per major vespucci release, which is roughly every half a year.

math1985:
> the topNames.json file should be regenreated too, 

I know, currently in the process of downloading the latest planet.

bhousel:
> @bhousel Question, does iD take the generated .json file from this repository, or does it run the script itself?

Looks like [iD copies it](https://github.com/openstreetmap/iD/blob/6fdb86be0409ac388f72e18e42e018dd4fd2fcb6/Makefile#L133).

After that, iD does [some processing](https://github.com/openstreetmap/iD/blob/6fdb86be0409ac388f72e18e42e018dd4fd2fcb6/build.js#L88-L108) to take the contents of `name-suggestions.json` and turn them into preset suggestions.  It looks like the only interesting thing that that code does is take only the suggestion with the highest count.

For example, that means that if the list contains "Walmart" as `shop=department_store` with `count: 517` and "Walmart" as a `shop=supermarket` with `count: 644`, iD will only suggest the `shop=supermarket` tagging.

PR #39, opened 2017-03-05
First pass at generating JOSM presets.

maxerickson:
Builds a JOSM preset file as briefly discussed at https://www.reddit.com/r/openstreetmap/comments/5x7ohz/consistent_naming_tagging_for_reoccuring_items/

The JOSM preset search considers the tags in the preset so the extra tags don't need to be in the name.

To make the presets easily available in JOSM all that needs to be done is to add a link to the raw github file here: https://josm.openstreetmap.de/wiki/PresetsSource .

### DIFF
--- a/buildJosm.js
+++ b/buildJosm.js
@@ -5,7 +5,7 @@ var fs = require("fs"),
 var presets=xmlbuilder.create('presets', {version: '1.0', encoding: 'UTF-8'})
   .att('xmlns', 'http://josm.openstreetmap.de/tagging-preset-1.0')
   .att('author', 'Name Suggestion Index');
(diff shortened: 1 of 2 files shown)


math1985:
Great idea to have this, it's a first step towards having name-suggestion-index in JOSM.

math1985:
@bhousel I'm fine with having this in this repository, but let me know if you think there's a better repository for this.

math1985:
@maxerickson Could you add a couple of pines documentation to the script?

bhousel:
@math1985, @maxerickson This looks good to me!  If someone else can verify that the output looks good, let's merge it 👍

bhousel:
Thanks @maxerickson!  Merging per IRC chat that this seems fine.

PR #47, opened 2017-08-16
Address some obvious non-chain names and typos

simonpoole:
Just a handful of obvious fixes when generating the index with a current planet, however there are a lot more.

More general issue: there are a lot of school names (likely from the US) of the form _something_ (historical) these should probably all be removed directly.


canonical.json
```diff
@@ -220,6 +225,16 @@
             "Caisse d'épargne"
         ]
     },
+    "Caixa Geral de Depósitos":{
+        "matches":[
+            "Caixa Geral de Depósitos"
```
math1985: If I'm seeing it well, you have aa match equal to the canonical name here. Did you mean something without accents or something like that?

simonpoole: ah the accent shouldn't be there ... fixing

math1985:
Thanks!

I'm accepting your PR, however the compiled version in the repo still needs to be updated.

simonpoole:
We need to decide on how best to remove the (historical) junk IMHO before re-publishing.

math1985:
Note that although we have a lot of junk, most applications are not harmed by this. For example, in case of missing canonicalization, iD will only give the most used of the two alternatives, and the lesser used alternative will not be displayed to end users.

simonpoole:
Sorry but that is not correct: somebody starting to type "Antioch School" will now get "Antioch School (historical)" I really don't think we want that to happen.

math1985:
Good point.

PR #49, opened 2017-08-17
Make comparision for discarded names case insensitive and add per "element" discarded names

simonpoole:
This simplifies the filter.json file a bit in that it changes the test for discarding a value to be case insensitive (likely doesn't really make sense for non-latin scripts).  If further adds a facility for per "element" discarded tags. 

Example why this is good: previously "Casino" was discarded regardless of what it was attached to as a name, this led to the "Casino" supermarket chain not turning up in the list.

Further discard single digits and "Windmill", normalize McColl's. and I've tried to make the handling of Aldi Süd slightly less messed up.

In reality tagging name as Aldi Süd or Nord is wrong as the brand/name is always ALDI, and the respective operators are Aldi Süd, Nord, Suisse and others (note that these should likely be uppercase but the mixed case tagging is far more popular).


simonpoole:
I've further tried to make the handling of Aldi Süd slightly less messed up.

In reality tagging name as Aldi Süd or Nord is wrong as the brand/name is always ALDI, and the respective operators are Aldi Süd, Nord, Suisse and others (note that these should likely be uppercase but the mixed case tagging is far more popular).

bhousel:
Thanks @simonpoole!

PR #50, opened 2017-08-21
 ignore "stadtmobil CarSharing-Station"

sb12:
This is not even a name but a description. Some iD users used that for car rental companies even outside Europe because it was suggested to them.


bhousel:
Thanks @sb12!

PR #54, opened 2017-09-01
handle media expert

matkoniecz:
based on https://forum.openstreetmap.org/viewtopic.php?pid=661824#p661824


matkoniecz:
note to self: https://forum.openstreetmap.org/viewtopic.php?pid=661825#p661825 has next idea for PR.

PR #58, opened 2017-09-28
T-Mobile and Empik normalization

matkoniecz:
based on https://forum.openstreetmap.org/viewtopic.php?pid=661824#p661824 https://forum.openstreetmap.org/viewtopic.php?pid=661832#p661832 https://wiki.openstreetmap.org/w/index.php?title=WikiProject_Poland/Shops&oldid=1503653

### DIFF
--- a/canonical.json
+++ b/canonical.json
@@ -427,6 +427,11 @@
             "Station Elf"
         ]
     },
+    "Empik":{
+        "matches":[
+            "empik"
+        ]
+    },
     "Engel-Apotheke":{
         "matches":[
             "Engel Apotheke"
@@ -1211,6 +1216,11 @@
     "Superdrug":{
         "nix_value":[
             "chemist"
+        ]
+    },
+    "T-Mobile":{
+        "matches":[
+            "T-mobile"
         ]
     },
     "TGI Friday's":{


math1985:
Thanks!

The files still need to be regenerated.

matkoniecz:
> The files still need to be regenerated.

By running `make`? I did it before making PR (and it failed to modify any files tracked in repository...).

math1985:
Did you get any errors when running `make`? It should update name-suggestions.json and name-suggestions.min.json.

matkoniecz:
No, and it updated name-suggestions.json and name-suggestions.min.json.

Neither is hidden by .gitignore.

`git check-ignore -v name-suggestions.min.json` and `git check-ignore -v name-suggestions.json` have an empty output unlike for example `git check-ignore -v test.pbf`.

I created folder `before` and put files generated before this PR and folder `after` and put files generated after changes introduced by this PR.

`diff before after` is empty - it probably means that as of today this PR introduced no changes in produced rules (maybe because some people changed empik to Empik in OSM - see https://wiki.openstreetmap.org/w/index.php?title=WikiProject_Poland%2FShops&type=revision&diff=1509171&oldid=1503653 ).

PR #63, opened 2017-10-03
Some fixes

### LINKED ISSUE #55: Some changes to default name suggestions to revolve inconsistencies
Some changes to default name suggestions to revolve inconsistencies

- “ALDI”: default should be “Aldi” because also “Aldi Süd” and “Aldi Nord” do not use all-capitals.

- “Adler Apotheke” is wrong orthography. Default should be “Adler-Apotheke” which is correct orthography.

- “Citroen” does probably only exist because many keyboard layouts don’t provide “ë”. Default should be “Citroën”, which also the enterprise itself uses in general.

- “NETTO”: Default should be “Netto” because all the other German supermarkets in the list are also not all-capitals.

- “REWE Getränkemarkt”: Default should be “Rewe Getränkemarkt” because also the main store “Rewe” does not use all-capitals.

- “LAWSON”: I suppose that also here the default should be “Lawson”, but I’m not sure here because I’ve never been in one of these stores.

sommerluk:
Some fixes.

Closes #55 

This PR is somewhat incomplete, because running “make” didn’t work for me. Can anybody do that?

### DIFF
--- a/canonical.json
+++ b/canonical.json
@@ -17,9 +17,9 @@
             "99p Store"
         ]
     },
-    "ALDI":{
+    "Aldi":{
         "matches":[
-            "Aldi"
+            "ALDI"
         ]
     },
     "ARAL":{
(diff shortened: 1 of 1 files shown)


bhousel:
Looks fine, thanks @sommerluk 👍

sommerluk:
Really fast merge! Thanks.

PR #66, opened 2017-10-05
New set of descriptive names

matkoniecz:


bhousel:
Thanks @matkoniecz, this seems fine.. It looks like the removal of the generic "Boutique" is the only effect.

PR #67, opened 2017-10-05
subway is a fast food, not a restaurant

matkoniecz:


bhousel:
I think it's not necessary to add `nix_value` unless the counts are close - in this case Subway is overwhelmingly tagged as "fast_food".   But this change doesn't hurt anything either, so I'll merge it.

matkoniecz:
Main benefit is that it hides this entry from stdout output during make.

Issue #82, opened 2018-01-13
shop=wholesale


HubMiner:
This is continuation to: https://github.com/openstreetmap/iD/pull/4657

Per documentation: https://wiki.openstreetmap.org/wiki/Tag:shop%3Dwholesale
...at least first 3 stores listed in the example section (Costco, Sam's Club, BJ's Wholesale Club) should be updated, possibly same for "Makro" and "Real Canadian Superstore". (Pricesmart is not mentioned in topNames.json)

```
OLD: shop/supermarket OR shop/department_store
NEW: shop=wholesale AND wholesale=supermarket
```
Note, some items are tagged as "amenity/fuel", those should not be changed.

Additionally, there are many name variations especially for Costco fuel station. These probably should be normalized to "Costco Gasoline".

I could attempt this change with some hints. :)

1ec5:
> Additionally, there are many name variations especially for Costco fuel station. These probably should be normalized to "Costco Gasoline".

Good point. I always confuse it with Kroger Fuel and put in “Costco Fuel”.

HubMiner:
bump?

1ec5:
Hi @HubMiner, the readme has [instructions](https://github.com/osmlab/name-suggestion-index#what-to-edit) for contributing to this project. To help push this issue along, please consider editing the files listed there an opening a pull request with your edits.

bhousel:
Looks like this was done in #86 - can this be closed?

HubMiner:
Few days passed after merging this change, and I don't see the intended changes.

**Use case:**
Click Point
Search - Costco

**Expected results:**
a) Costco Wholesale - Warehouse
b) Costco Gasoline - Gas Station

**Actual results:**
a) Costco - Supermarket [bad]
b) Costco Gas - Gas Station [bad]
c) Costco Gasoline - Gas Station [good]

Please re-review https://github.com/osmlab/name-suggestion-index/pull/86
and share some guidance.

Also, I am still not clear if this would affect existing values in OSM, or only suggest when creating new ones.

bhousel:
Yes, this won't show up in iD until we release a v0.1.5 of this project, and pull it into the next iD release which will be v2.7 and happen hopefully in the next week or two.

> Also, I am still not clear if this would affect existing values in OSM, or only suggest when creating new ones.

It only affects users adding new things, not existing data.

HubMiner:
Thanks!

HubMiner:
[...]

PR #86, opened 2018-02-04
82: Costco Warehouse, Costco Gasoline

HubMiner:
https://github.com/osmlab/name-suggestion-index/issues/82


matthijsmelissen:
Thanks!

PR #88, opened 2018-05-06
żabka is not a supermarket

matkoniecz:
Żabka is even given as an example of shop=convenience at https://wiki.openstreetmap.org/wiki/Pl:Tag:shop%3Dconvenience

I looked at some shop=supermarket name=Żabka and nearly all of them are suspected (some from maps.me newbies).


matthijsmelissen:
Agree!

PR #90, opened 2018-05-08
Contribution

sommerluk:
Matches various spelling errors of Oilibya fuel stations. While “Oilibya” is clearly the most popular in OSM, there are various miss-spellings. The combination i+l+i is difficult to read, the the font type they use for their fuel stations doesn’t make it easier either.

Adds also Coqivoire and Foani, two butcher franchaises that are quite popular in Ivory Coast, with automatic refinement tags (`"butcher":"poultry"`).

### DIFF
--- a/canonical.json
+++ b/canonical.json
@@ -328,6 +328,14 @@
             "coop"
         ]
     },
+    "Coqivoire":{
+        "matches":[
+            "COQIVOIRE"
+        ],
+        "tags":{
+            "butcher":"poultry"
+        }
+    },
     "Costa":{
         "matches":[
             "Costa Coffee"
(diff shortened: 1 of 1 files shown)


matthijsmelissen:
Thanks!

matkoniecz:
@sommerluk 

> Foani

Is there chance that you can link their website/twitter account/FB account/official reports/some other confirmation? I need it for https://www.wikidata.org/wiki/Wikidata:Administrators%27_noticeboard#Please_undelete_this_entries_about_brands (since some time for some reason wikidata/wikipedia entry is considered as mandatory and https://www.wikidata.org/w/index.php?title=Q60183335&action=history got deleted and for now temporarily undeleted

matkoniecz:
I added some other - see https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#memo--edit-wikidata

See

PR #107, opened 2018-09-14
remove generic term

matkoniecz:
sklep is simply "shop" in Polish
sklep spożywczy is also completely generic and should rather not appear as a name (and 120% not as a brand)


bhousel:
I caught these in 2072f05

PR #142, opened 2018-09-16
 paczkomat is also machine to leave or pick-up parcels 

matkoniecz:
it is a place where one may pick up or leave parcels in an automated machine
so "vending": "parcel_pickup;parcel_mail_in" fits
it is the same machine as packstation,
sole difference is that it is operating in a different country under a different brand

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -16060,7 +16060,8 @@
     "tags": {
       "amenity": "vending_machine",
       "brand": "InPost",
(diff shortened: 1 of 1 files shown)


matthijsmelissen:
Could you clarify the PR title/commit description? I'm not sure how it describes the changes you made.

matkoniecz:
I updated commit message and copied it into PR title, PR description.

PR #159, opened 2018-09-28
nothing indicates that they all operate under shared brand

matkoniecz:


bhousel:
Yeah there are quite a lot of these in the restaurant category...
I've been removing them in `filters.json` this way:

https://github.com/osmlab/name-suggestion-index/blob/3a3523b75e22401f70b2a55616788f8fd68d77e9/config/filters.json#L27-L34

matkoniecz:
I will redo it in a new PR.

PR #164, opened 2018-09-28
More generic

matkoniecz:


bhousel:
Thanks!  keep them coming, this is super helpful 👍

matkoniecz:
BTW, it turns out that listing entries that

 - exists in my region
 - have no brand tags in my region (or generally)

is quite good way to list entries worth checking. I made simple proof of concept ruby script (that makes a hilarious amount of overpass queries, so I can generate list for about one region in a day)

```
require 'json'

def check_preset_for_country(country)
  data = JSON.parse(File.read('name-suggestion-index/config/canonical.json'))
  data.each do |entry|
    if entry[1]["tags"]["brand"] == nil
      next
    end
    if(count_in("name", entry[1]["tags"]["name"], country) == 0)
      next
    end
    if(count("brand", entry[1]["tags"]["brand"]) <= 1)
      puts
      puts "unusual brand"
      puts entry[0]
    end
  end
end

def count(key, value)
  query = "[out:json][timeout:3600];
  (
  node[\"" + escape_tag(key) + "\"=\"" + escape_tag(value) + "\"];
  way[\"" + escape_tag(key) + "\"=\"" + escape_tag(value) + "\"];
  relation[\"" + escape_tag(key) + "\"=\"" + escape_tag(value) + "\"];
  );
  out meta qt;"
  explanation = "counting #{key}=#{value}"
  data = execute_overpass_query(query, explanation)
  return data.length
end

def count_in(key, value, location_name)
  query = "[out:json][timeout:3600];
  area[name='" + location_name + "']->.searchArea;
  (
  node[\"" + escape_tag(key) + "\"=\"" + escape_tag(value) + "\"](area.searchArea);
  way[\"" + escape_tag(key) + "\"=\"" + escape_tag(value) + "\"](area.searchArea);
  relation[\"" + escape_tag(key) + "\"=\"" + escape_tag(value) + "\"](area.searchArea);
  );
  out meta qt;"
  explanation = "counting #{key}=#{value}"
  data = execute_overpass_query(query, explanation)
  return data.length
end

def escape_tag(string)
  return CartoCSSHelper::OverpassQueryGenerator.overpass_escape(string)
end

def execute_overpass_query(query, explanation)
  json_string = CartoCSSHelper::OverpassQueryGenerator.get_overpass_query_results(query, explanation)
  return JSON.parse(json_string)["elements"]
end

def initialize_overpass_interface
  require 'cartocss_helper'
  CartoCSSHelper::Configuration.set_path_to_folder_for_cache('/media/mateusz/5bfa9dfc-ed86-4d19-ac36-78df1060707c/OSM-cache')
end

initialize_overpass_interface
check_preset_for_country("Polska")
check_preset_for_country("USA")
check_preset_for_country("England")
```

PR #582, opened 2018-10-16
added wiki tags

dieterwarson:


matkoniecz:
Thank you very much for help!

Confirmed by https://overpass-turbo.eu/s/COy

I am merging it, adding your contribution to a project and closing #551 (BTW, it is always a good idea to mention code of fixed issue)

PR #739, opened 2018-10-16
Updated tags for amenity/bank|ABSA

sammansch:
Added  wikidata and wikipedia tags for amenity/bank|ABSA.


bhousel:
Looks good, thanks @sammansch  👍

PR #882, opened 2018-10-16
Added info and removed duplicates for Super 8 Motels

eigenbrot:
There were more counts for tourism/hotel but "Motels" is right in the name so I
kept tourism/motel


bhousel:
Looks good @eigenbrot - yeah looks like this is a case where Super 8 is often wrongly tagged in OSM.  Your contribution will help us fix it 👍

Issue #30, opened 2015-12-24
Chemist/pharmacy confusion


pmailkeey:
I note iD is now showing chemist chains (e.g. Boots, Rowlands, Lloyds, Numark, Well*) as 'amenity pharmacy' when in fact they are building=retail' 'shop=chemist' and should have pharmacy=yes by default. Asda (wal*Mart) in uk often has a pharmacy and pharmacy=yes is more appropriate than a separate 'amenity=pharmacy' - I think. Some have opticians in store too.

That raises the question as to how a mailbox should be shown in Asda - as a node can be used to show the location where as mailbox=yes as a subtag on the building doesn't.

* note chain 'Well' is the name but the branding is displayed as '+Well' - which is how I first came across it - but is in fact incorrect.

math1985:
Note that in the Netherlands, and possibly other countries as well, Boots actually run pharmacies rather than chemists.

math1985:
Boots, Rowlands, Lloyds, Numark and Well are currently (much) more often tagged as amenity=pharmacy than shop=chemist. If you'd like to change this, I'd recommend discussing this with the UK community first. For Boots I can follow your reasoning, but for for example Lloyds, pharmacy seems right to me.

I'm closing this issue for now.

bhousel:
Yeah I agree that we get this wrong in iD.  We do have a `shop=chemist` preset, but we display the word "Chemist" and we show an icon of a little chemistry beaker.   US speakers are not familiar with the UK usage of this word.

We should instead encourage `amenity=pharmacy` to be tagged as a point node within drug stores or whatever other business. This makes more sense because:
1. the pharmacy department is usually a separate entity with different phone number and operating hours
2. pharmacies frequently exist in other non-drug stores, supermarkets (Walmart, Costco, etc)

In my mind the ideal steps going forward would be to:

1. clarify position with tagging lists

2. update wiki

3. update iD to:
  - rename `shop=chemist` preset to say "Drug Store" and change icon.
  - make sure `amenity=pharmacy` says "Pharmacy" and appears only on nodes

4. update name-suggestion-index

5. maproulette challenge to clean up existing chemist/pharmacy mixups 


reopen if you agree?

math1985:
> We should instead encourage amenity=pharmacy to be tagged as a point node within drug stores or whatever other business.

[...]

PR #1276, opened 2018-10-16
Add Walgreens information

iandees:
Fixes https://github.com/osmlab/name-suggestion-index/issues/701

Note that I'm merging "Walgreens Pharmacy" into "Walgreens". I can't think of a Walgreens that doesn't have a pharmacy.


bhousel:
Looks good 👍 
Note that we are keeping pharmacies separate from chemists per https://github.com/osmlab/name-suggestion-index/issues/30#issuecomment-228798011

PR #1630, opened 2018-10-17
#1544 adds  at&t :phone:

jonahadkins:


bhousel:
Nice job thanks @jonahadkins!  I remember this was one I skipped on the first pass because there are several AT&T's - You picked the correct one for AT&T wireless.

PR #1777, opened 2018-10-17
shop/supermarket/IGA add wikidata and wikipedia tags

mraguso2:
Added tags for IGA Supermarket


config/canonical.json
```diff
@@ -30100,6 +30100,8 @@
     "count": 619,
     "tags": {
       "brand": "IGA",
+      "brand:wikidata": "Q5970945",
+      "brand:wikipedia": "en:IGA_(supermarkets)",
```
bhousel: Remove the `_`, then it's good to merge.. 

```suggestion
      "brand:wikipedia": "en:IGA (supermarkets)",
```

mraguso2:
removed "_" from IGA entry
I'm working on another entry that has a "_" in the wikipedia tag - should I just not include underscores?

andrewharvey:
The wikipedia article and the wikidata are pointing to different things...? The wikidata is for the Australian IGA, the wikipedia is for the global IGA.

bhousel:
> The wikipedia article and the wikidata are pointing to different things...? The wikidata is for the Australian IGA, the wikipedia is for the global IGA.

Thanks for noticing @andrewharvey - I just updated the wikidata tag in e2ee17b

PR #1823, opened 2018-10-17
#1699 Jumbo

tstewart15:
This is my first PR here; hopefully done correctly! There were many warnings, but I assume I can ignore. 

Also, when exactly is the `countryCodes` property needed? Jumbo has an [english Wikipedia entry](https://en.wikipedia.org/wiki/Jumbo_(supermarket)) but is a supermarket in the Netherlands/Belgium.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -26192,8 +26192,8 @@
     "count": 94,
     "tags": {
       "brand": "Geers",
-      "brand:wikidata":"Q1497707",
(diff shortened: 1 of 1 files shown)


bhousel:
> Also, when exactly is the `countryCodes` property needed?

It might be useful to restrict where people can add them to OpenStreetMap, but not absolutely necessary.  We're still trying to decide what's best.

PR #1871, opened 2018-10-17
Add tags to Albert

mraguso2:
issue: https://github.com/osmlab/name-suggestion-index/issues/1647
add tags to albert


bhousel:
Thanks @mraguso2 👍

PR #1872, opened 2018-10-17
Add Wiki Tags for Hotel Victoria

kyostra:
added wikidata and wikipedia for hotel victoria

fixes: #1857

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -33397,6 +33397,8 @@
     "count": 81,
     "tags": {
       "brand": "Hotel Victoria",
+      "brand:wikidata": "Q5912087",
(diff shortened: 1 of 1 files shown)


kyostra:
changed tag to #1857 from #1869

matkoniecz:
Thank you very much for a help!

I would say that Hotel Victoria appears to describe hotels that are unrelated to each other, so in this case discarding name would be probably a better idea (I will do it soon).

bhousel:
> I would say that Hotel Victoria appears to describe hotels that are unrelated to each other, so in this case discarding name would be probably a better idea (I will do it soon).

I agree - thanks @matkoniecz

PR #1877, opened 2018-10-17
Added Ross Wiki Tags

mds08011:
Added Wikipedia and Wikidata tags for Ross.  Issue: #1202


bhousel:
Thank you @mds08011  👍

PR #1878, opened 2018-10-17
added wiki tags

dieterwarson:


matkoniecz:
Great! Can you add also change that would fix from

> "brand:wikipedia": "en:Hilton Hotels %26 Resorts",

to

> "brand:wikipedia": "en:Hilton Hotels & Resorts",

?

matkoniecz:
Thank you again for help!

matkoniecz:
I opened #1879 to discuss what should be done with "Ibis" and "Ibis Hotel" duplicate, there is no need to do it here.

dieterwarson:
No problem, thanks for merging!

Issue #1640, opened 2018-10-17
Research needed for: shop/stationery|Office Max


bhousel:
#### :thinking: We need help deciding what to do with "shop/stationery|Office Max"

:point_right: Read the [CONTRIBUTING.md](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) guide for more info, and feel free to ask questions on this issue!

### What's needed

First, do some research on `shop/stationery|Office Max` by searching Google and Wikipedia.

#### If it is an actual brand:

* Edit `config/canonical.json`:
  * Find the entry for `shop/stationery|Office Max`
  * Add tags for `brand:wikidata` and `brand:wikipedia`
  * Optionally add `"countryCodes"` or other tags
* `npm run build`
* Commit your changes and submit a pull request

#### If it is not an actual brand:

* Edit `config/filters.json` to add an expression that discards it
* Edit `config/canonical.json` to remove the entry for `shop/stationery|Office Max`
* `npm run build`
* Commit your changes and submit a pull request

#### If the situation is complicated:

* Leave a comment on this issue with what you find

:sparkles: Thanks!

_This is an autogenerated ticket_

ENT8R:
> My understanding is that this issue has been generated for the single item in the `match` property

The match property does only contain other names a user might use for this brand. In this case it should probably be not removed...
But this issue has already been fixed with https://github.com/osmlab/name-suggestion-index/commit/bef2100ccc09218d7eb75f7950763c385bbdc06a and https://github.com/osmlab/name-suggestion-index/pull/1883

bhousel:
Yes, @ENT8R is right.. "match" is for alternative less-preferred names.

That property really does have a confusing name - I think I should make a `FAQ.md` document for things like this.   Or consider renaming it.. 🤔

Issue #1609, opened 2018-10-17
Research needed for: shop/shoes|Journeys


bhousel:
#### :thinking: We need help deciding what to do with "shop/shoes|Journeys"

:point_right: Read the [CONTRIBUTING.md](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) guide for more info, and feel free to ask questions on this issue!

### What's needed

First, do some research on `shop/shoes|Journeys` by searching Google and Wikipedia.

#### If it is an actual brand:

* Edit `config/canonical.json`:
  * Find the entry for `shop/shoes|Journeys`
  * Add tags for `brand:wikidata` and `brand:wikipedia`
  * Optionally add `"countryCodes"` or other tags
* `npm run build`
* Commit your changes and submit a pull request

#### If it is not an actual brand:

* Edit `config/filters.json` to add an expression that discards it
* Edit `config/canonical.json` to remove the entry for `shop/shoes|Journeys`
* `npm run build`
* Commit your changes and submit a pull request

#### If the situation is complicated:

* Leave a comment on this issue with what you find

:sparkles: Thanks!

_This is an autogenerated ticket_

bonbon12:
Hello, im looking at this issue and just wanted to confirm a few things.
1. What is the "count value" indicating in clothes.json
2. is the "needs wikidata" tag implying that we need to create wikidata for the store?
3. i also noticed that in clothes.json that adidas is missing. can i add this?

bhousel:
Hello!

> 1. What is the "count value" indicating in clothes.json

This is generated automatically based on the last time we processed data from OpenStreetMap. You don't need to change it.

> 2. is the "needs wikidata" tag implying that we need to create wikidata for the store?

It means that someone needs to add the brand to wikidata. The [operating company](https://en.wikipedia.org/wiki/Genesco) has a page and wikidata id, but the brand does not.  So, somebody with experience in editing Wikipedia/Wikidata should do this part.

> 3. i also noticed that in clothes.json that adidas is missing. can i add this?

It looks like Adidas already exists under `shop/sports.json`, so it's not necessary to add it to `shop/shoes.json`

PR #1883, opened 2018-10-17
adds Journeys, OfficeMax #1609 #1640

jonahadkins:


config/canonical.json
```diff
@@ -28715,13 +28717,22 @@
...
-    "match": ["shop/stationery|OfficeMax"],
     "tags": {
-      "brand": "Office Max",
       "name": "Office Max",
       "shop": "stationery"
     }
   },
+  "shop/stationery|OfficeMax": {
+    "count": 115,
+    "match": ["shop/stationery|OfficeMax"],
```
bhousel: I think what you're trying to do here is match "Office Max" with a space..
(the `match` property contains the alternative names that we don't prefer)

```suggestion
    "match": ["shop/stationery|Office Max"],
```

config/canonical.json
```diff
@@ -28715,13 +28717,22 @@
   },
   "shop/stationery|Office Max": {
```
bhousel: and this "Office Max" gets removed since it is matched on another entry

matkoniecz:
Something went wrong, Github diff shows that it reverts many recent changes. I suspect that it may be a Github problem.

bhousel:
yeah the merge in 5aa4055 removed some things it wasn't supposed to..

I ended up manually making the change in bef2100

PR #1891, opened 2018-10-17
Update canonical.json

javirg:
Added Dia supermarket


bhousel:
looks like #1892 supersedes this

PR #1899, opened 2018-10-17
Palí

javirg:


config/canonical.json
```diff
@@ -30760,7 +30760,9 @@
     "tags": {
       "brand": "Palí",
       "name": "Palí",
-      "shop": "supermarket"
+      "shop": "supermarket",
+      "brand:wikidata": "Q1064887",
+      "brand:wikipedia": "en:Walmart_de_México_y_Centroamérica"
```
bhousel: remove the `_` - thanks!
```suggestion
      "brand:wikipedia": "es:Walmart de México y Centroamérica"
```

config/canonical.json
```diff
@@ -31356,7 +31358,7 @@
     "tags": {
       "brand": "Tesco",
       "name": "Tesco",
-      "shop": "supermarket"
+      "shop": "supermarket",
```
bhousel: no trailing `,` here
```suggestion
      "shop": "supermarket"
```

config/canonical.json
bhousel: Also, I don't see any mention of the "Palí" brand on the `en:` page, so I changed it to the `es:` page which mentions it.

config/canonical.json
good-praxis: :thinking: That's odd, this is the trailing comma from master fixed in #1914, how was it committed here?

bhousel: Yeah I was wondering that too...
The merges are coming fast and furious today, so I'm not surprised that there is a little bit of breakage and that every time we `npm run build` things get a little shuffled around.

PR #1905, opened 2018-10-17
Updated info for Comfort [Inn|Suites]

eigenbrot:
Addresses #1843 and #1841

This is a little tricky because Comfort Inn & Suites does not officially exist
anymore. According to Choice Hotels (the parent company) Comfort Inn and Comfort
Suites are two distinct brands. For this reason I added each one to the other's
"nomatch" tag. This might not be the right call.

I also removed the entry for "Comfort Inn & Suites" but set it to match with
"Comfort Inn" because if you go to the Choice Hotels page for "Comfort Inn &
Suites" you are redirected to the "Comfort Inn" reservation page.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -33153,24 +33153,25 @@
   },
   "tourism/hotel|Comfort Inn": {
     "count": 321,
+    "match": [
+      "tourism/hotel|Comfort Inn & Suites"
+    ],
+    "nomatch": ["tourism/hotel|Comfort Suites"],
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks for researching @eigenbrot  - this looks good to me 👍 

In OpenStreetMap, people are encouraged to map whatever the business looks like "on the ground", based on the signage.  So if they are in the middle of a brand update, it's ok for several options to exist.

PR #1914, opened 2018-10-17
Fixing trailing comma

good-praxis:
Hey!
Build was crashing, so I ran canonical through a validator, turns out there was a trailing comma in one of those thousand lines. Removed it quickly, and now the build works :+1:


bhousel:
Thanks! 👍

PR #1916, opened 2018-10-17
adding rewe getraenkemarkt

good-praxis:


config/canonical.json
```diff
@@ -17105,6 +17105,8 @@
     ],
     "tags": {
       "brand": "Rewe Getränkemarkt",
+      "brand:wikidata": "Q573266",
+      "brand:wikipedia": "de: Rewe_Group",
```
bhousel: The `brand:wikipedia` tag needs to be formatted like this.. thanks!
```suggestion
      "brand:wikipedia": "de:Rewe Group",
```

PR #1917, opened 2018-10-17
Add .travis.yml

### LINKED ISSUE #1913: Setup Travis
Because CI is nice

_Originally posted by @bhousel in https://github.com/osmlab/name-suggestion-index/pull/1910/comment#issuecomment-430729658_

ENT8R:
Fixes #1913

I assume that you don't want to receive any emails, so I just disabled notifications... You just need to go to https://travis-ci.org/osmlab/name-suggestion-index and enable Travis for this repo. But I think you know what you have to do 😄

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,14 +19,15 @@
   "scripts": {
     "old": "node build.js && rollup -c",
     "build": "node build_filterNames",
-    "wikimatch": "node match_wikiTags"
+    "wikimatch": "node match_wikiTags",
+    "test": "node validate"
   },
   "devDependencies": {
     "clear": "^0.1.0",
     "colors": "^1.1.2",
     "diacritics": "^1.3.0",
     "json-stringify-pretty-compact": "^1.1.0",
-    "jsonschema": "^1.1.0",
+    "jsonschema": "^1.2.4",
     "linebyline": "^1.3.0",
     "node-fetch": "^2.2.0",
     "osmium": "^0.5.7",
--- a/validate.js
+++ b/validate.js
@@ -1,0 +1,31 @@
+const colors = require('colors/safe');
+const Validator = require('jsonschema').Validator;
+
+const filters = require('./config/filters.json');
+const canonical = require('./config/canonical.json');
+
+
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -19,14 +19,15 @@
...
+    "wikimatch": "node match_wikiTags",
+    "test": "node validate"
   },
   "devDependencies": {
     "clear": "^0.1.0",
     "colors": "^1.1.2",
     "diacritics": "^1.3.0",
     "json-stringify-pretty-compact": "^1.1.0",
-    "jsonschema": "^1.1.0",
+    "jsonschema": "^1.2.4",
```
good-praxis: Probably/Shouldn't break anything, but why the update?

validate.js
```diff
@@ -0,0 +1,31 @@
...
+            if (e.property) {
+                console.error('  ' + colors.yellow(e.property + ' ' + e.message));
+            } else {
+                console.error('  ' + colors.yellow(e));
+            }
+        });
+        console.error();
+        process.exit(1);
+    }
+}
```
good-praxis: Good idea, you should add a build test, and validate that the newly-build files are the same that were committed!

package.json
ENT8R: Why not? It is always good to use the latest version to avoid security issues...

validate.js
ENT8R: I don't really get what you mean... When should this script run?

package.json
good-praxis: Fair point!

validate.js
good-praxis: Well, the build only functions if the json is valid anyway (i.e. see what `npm run build` does) - and if the contributor actually ran the build locally and committed the results, the build files (`/dist/`) shouldn't change after the build.

This travis config will already prevent an array of problems, though, so good job!

matkoniecz:
@bhousel Can you look at this? I see no problems, but in this PR I would not notice if something would be wrong.

bhousel:
Oh yeah - it seems fine!  Looks like it just does the schema validation step only, which is good 👍

PR #1922, opened 2018-10-17
Shop optician synsam & synoptik

femtioelva:


bhousel:
Thanks @femtioelva!

ENT8R:
Just found this right now, but is `sv` or `se` the correct country code for Sweden? This PR uses only `sv` but I'm pretty sure that only `se` is right...

PR #1925, opened 2018-10-17
Bojangles

maxerickson:
#426. The ' at the end of the name matches the official branding but is less common in existing data.


bhousel:
Looks good @maxerickson
I agree we can use this index to nudge mappers towards using the "official branding"

PR #1930, opened 2018-10-17
Fixed build breaking bug in Bojangles' entry

eigenbrot:
Caused by PR #1925. Forgot the commas, d'oh!

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -6647,10 +6647,9 @@
     }
   },
   "amenity/fast_food|Bojangles'": {
-    "count": 153,
     "match": [
-      "amenity/fast_food|Bojangles"
-      "amenity/restaurant|Bojangles"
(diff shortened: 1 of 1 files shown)


eigenbrot:
Also, the addition of the ' to the end of amenity/fast_food|Bojangles' causes an Uncommon entry warning because the entry in keepNames.json does NOT have the ' at the end. How do we resolve this issue?

bhousel:
> Also, the addition of the ' to the end of amenity/fast_food|Bojangles' causes an Uncommon entry warning because the entry in keepNames.json does NOT have the ' at the end. How do we resolve this issue?

replace the `"count": nn` property with `"nocount": true`
(this is like saying, we don't care how many of these are in OpenStreetMap)

bhousel:
Thanks!

Issue #1492, opened 2018-10-17
Research needed for: shop/hairdresser|Franck Provost


bhousel:
#### :thinking: We need help deciding what to do with "shop/hairdresser|Franck Provost"

:point_right: Read the [CONTRIBUTING.md](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) guide for more info, and feel free to ask questions on this issue!

### What's needed

First, do some research on `shop/hairdresser|Franck Provost` by searching Google and Wikipedia.

#### If it is an actual brand:

* Edit `config/canonical.json`:
  * Find the entry for `shop/hairdresser|Franck Provost`
  * Add tags for `brand:wikidata` and `brand:wikipedia`
  * Optionally add `"countryCodes"` or other tags
* `npm run build`
* Commit your changes and submit a pull request

#### If it is not an actual brand:

* Edit `config/filters.json` to add an expression that discards it
* Edit `config/canonical.json` to remove the entry for `shop/hairdresser|Franck Provost`
* `npm run build`
* Commit your changes and submit a pull request

#### If the situation is complicated:

* Leave a comment on this issue with what you find

:sparkles: Thanks!

_This is an autogenerated ticket_

guilhermemalfatti:
Hi, what is the process to submit a pull request? creating a new branch e push it? or there is another way to do so?

bhousel:
I think this GitHub guide is a pretty good overview of the process:
https://guides.github.com/activities/forking/

You fork the repo, make changes on your local copy, and send a "pull request" back to this repo for me to review.

bhousel:
> Hi, what is the process to submit a pull request? creating a new branch e push it? or there is another way to do so?

I'm really sorry to have to reject your pull request - you did everything right!

bmillemathias:
Can be closed now.

PR #1933, opened 2018-10-18
fix issue #1492

guilhermemalfatti:
issue #1492

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -25843,6 +25843,8 @@
     "tags": {
       "brand": "Franck Provost",
       "name": "Franck Provost",
+      "brand:wikidata": "Q2355121",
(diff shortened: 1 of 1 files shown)


bhousel:
Ah sorry - this is one where we shouldn't accept it. 😢 

Wikidata "Q2355121" is for Franck Provost the _human being_, not Franck Provost the _business_.
Also the wiki sitelink "en:Franck Provost" doesn't seem to actually exist.  (there is a "fr:Franck Provost" but again, that's for the human).

There are actually a bunch of these cases to pick through, and it's common for hairdressers and fashion designers.  (for example Louis Vuitton, Calvin Klein, etc).

I'll go back to #1492 and tag it as "needs wikidata"

PR #1943, opened 2018-10-18
Warning cleaning (and registering/deletion of warning causing instances

good-praxis:
Decided to clean up a bunch of warnings (7)

No issue for bancpost
No wikipedia page for Banco Unión (but it seems to be a [real brand](http://www.bancounion.com.bo/index.php))

Café Central's wikidata ID is actually just the ID of a single specific historical café.

Kroger is the correct name, even though Kroger Fuel had a higher count (according to wikipedia / image search, too)

[Liberty](http://www.libertypetroleum.com/) does not yet have a wikipedia page, but is mentioned [here](https://en.wikipedia.org/wiki/List_of_automotive_fuel_retailers)


good-praxis:
### Western Union
Well, technically Western Union is neither a bank nor a bureau de change, but they offer services similar to both. (this isn't entirely true, in Germany, Austria and France WU is actually registered as a bank, this is probably because of local regulation demanding this for their services), in reality, WU is closer to something like Paypal -- do we have a financial services tag?

### Esso (Express)
[Esso Express](https://nl.wikipedia.org/wiki/Esso_Express) similar to other "Express" brands appear in the Netherlands, usually because of localization. It may be better to add Netherland localized version with "Express" within the brand, however, Esso Express isn't a distinct version of Esso, and it may be that not all Esso's in the Netherlands are Esso Express' -- they are, however, not distinct gas stations

### Kroger
[Korger](### Western Union
Well, technically Western Union is neither a bank nor a bureau de change, but they offer services similar to both. (this isn't entirely true, in Germany, Austria and France WU is actually registered as a bank, this is probably because of local regulation demanding this for their services), in reality, WU is closer to something like Paypal -- do we have a financial services tag?

### Esso (Express)
[Esso Express](https://nl.wikipedia.org/wiki/Esso_Express) similar to other "Express" brands appear in the Netherlands, usually because of localization. It may be better to add Netherland localized version with "Express" within the brand, however, Esso Express isn't a distinct version of Esso, and it may be that not all Esso's in the Netherlands are Esso Express' -- they are, however, not distinct gas stations

### Kroger
Kroger gas stations are usually called just that. Kroger websites are also devoid of any mentioning of "Kroger Fuel" as a distinct brand

matkoniecz:
> Western Union

Are there locations operating under Western Union offering only currency exchange, without banklike services?

> Esso (Express)

In that case I think that keeping both names may be a better idea (maybe with the same brand, brand:wikipedia, brand:wikidata values).

> Kroger

In that case I am OK with that.

matkoniecz:
Sorry for misclick, I wanted to comment only.

good-praxis:
### Western Union:
I find it highly unlikely, however not impossible. It'd be against their buisness model, which is mainly focused on international transactions without a bank account. I can revert that commit, too, if necessary.

### Esso
How about that?

matkoniecz:
> It'd be against their buisness model, which is mainly focused on international transactions without a bank account.

It makes sense.

Let's merge it - thanks for your fixes!

matkoniecz:
Feel free to propose reverted ones as separate PRs if you think that I was too cautious.

bhousel:
> No issue for bancpost
> No wikipedia page for Banco Unión (but it seems to be a [real brand](http://www.bancounion.com.bo/index.php))
> 
> Café Central's wikidata ID is actually just the ID of a single specific historical café.
> 
> Kroger is the correct name, even though Kroger Fuel had a higher count (according to wikipedia / image search, too)
> 
> [Liberty](http://www.libertypetroleum.com/) does not yet have a wikipedia page, but is mentioned [here](https://en.wikipedia.org/wiki/List_of_automotive_fuel_retailers)

This is good research - but make sure to link back to the original issues so it doesn't get lost!

PR #1947, opened 2018-10-18
Add wiki tags for the German supermarket 'Nahkauf'

### LINKED ISSUE #1723: Research needed for: shop/supermarket|Nahkauf
#### :thinking: We need help deciding what to do with "shop/supermarket|Nahkauf"

### What's needed

First, do some research on `shop/supermarket|Nahkauf` by searching Google and Wikipedia.

  * Find the entry for `shop/supermarket|Nahkauf`
* `npm run build`

* Edit `config/canonical.json` to remove the entry for `shop/supermarket|Nahkauf`
* `npm run build`

:sparkles: Thanks!

ENT8R:
Fixes #1723

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -30740,9 +30740,12 @@
   },
   "shop/supermarket|Nahkauf": {
     "count": 254,
+    "countryCodes": ["de"],
     "match": ["shop/supermarket|nahkauf"],
     "tags": {
       "brand": "Nahkauf",
+      "brand:wikidata": "Q573266",
+      "brand:wikipedia": "en:REWE Group",
       "name": "Nahkauf",
       "shop": "supermarket"
     }


bhousel:
Thanks @ENT8R !
I'm reading up on https://en.wikipedia.org/wiki/REWE_Group - it looks like they operate several sub-brands that don't have their own wikidata entries (and maybe should!)

The en: page suggests that Nahkauf is more of a convenience store than a supermarket.  Is this the case?  I'll merge the wiki tag, but we might want to revisit this entry later..

ENT8R:
> The en: page suggests that Nahkauf is more of a convenience store than a supermarket. Is this the case?

I would say yes, but I'm not really sure about it because I live in a region where Nahkauf does not exist that much and I don't think that I ever visited one 😄
But from reading the German wiki page, I think that Nahkauf is not a real supermarket but rather a convenience store, so you are right 👍
So those shops should be rather tagged as `shop=convenience` and not as `shop=supermarket`, right?

bhousel:
> So those shops should be rather tagged as `shop=convenience` and not as `shop=supermarket`, right?

Probably - are there some German mappers that you can reach out to?
I think they mostly talk on the osm forum:  https://forum.openstreetmap.org/viewforum.php?id=14

ENT8R:
I will open a new topic and post the link here...

ENT8R:
Opened a new topic: https://forum.openstreetmap.org/viewtopic.php?id=64138 I will post results here 😄

ENT8R:
> > So those shops should be rather tagged as shop=convenience and not as shop=supermarket, right?

> Probably - are there some German mappers that you can reach out to?

The conclusion is that if a store (in Germany) has shopping carts, it should be tagged as `shop=supermarket` and if that is not the case it should rather be `shop=convenience`...
But "Nahkauf" is rather some medium-sized store which offers shopping carts, so in this case `shop=supermarket` should be the right tag to use.
Actually there is no real convenience store chain in Germany right now because people might prefer bigger store 😄

bhousel:
> But "Nahkauf" is rather some medium-sized store which offers shopping carts, so in this case `shop=supermarket` should be the right tag to use.

Thanks for checking!  🙇

ENT8R:
BTW: I will open a new PR to add the correct wikidata item after I created it 😄

PR #1960, opened 2018-10-18
Add tags to Norfa XL

diamentowybazant:
Cheers!


matkoniecz:
And next step closer to researching all!

PR #1963, opened 2018-10-18
Add tags to Bijou Brigitte

diamentowybazant:
Cheers!


matkoniecz:
Thank you! This pull request is now merges so your work is now part of name-suggestion-index project.

Issue #1934, opened 2018-10-18
What's missing?


bhousel:
I'm really excited by the response to this repository!  We're building a really useful link between OpenStreetMap and Wikidata.  Today we've merged over 70 pull requests, which has got to be a record for me.  Thank you for all your help!  🙇 

This issue is just a placeholder for **anything else that you notice is missing from the index**.  

The index was seeded by names which exist more than 50x in OpenStreetMap.  I know from hanging out in malls, shopping centers, and airport terminals that we are missing a lot of brands.  

No Auntie Anne's 🥨 ?
No Nathan's 🌭?
No Blue Bottle ☕️ ?
No Chop't 🥗 ?
No problem!

Just add it anyway, following the [contributor guidelines](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md).  Copy an existing similar entry in the index, add the wiki tags, and make sure to include `"nocount": true` to suppress the warning that it's a new thing.

When submit your pull request, tag (but don't close!) this issue.
You can add "re: #1934" in the commit message or pull request.

matkoniecz:
Also, I would encourage to look at location you know at OSM ( https://www.openstreetmap.org/ ) and add missing shops! (there is an "Edit" button on top left leading to built-in editor with tutorial for new users).

1ec5:
* [PetScan query of articles about chains of all sorts at the English Wikipedia](https://petscan.wmflabs.org/?psid=6157573)
* [Rough diff of English Wikipedia articles listed in the name suggestion index versus English Wikipedia articles in the PetScan query above](https://gist.github.com/1ec5/77ab9bfbaed39aa4524bda049511b1c9)

1ec5:
[This spreadsheet](https://docs.google.com/spreadsheets/d/14yhHtAtKjqUL5fo5m1CqkBTlE4whzXBH91J3HRweEL0/edit?usp=sharing) lists hidden chains (of the non-Markov variety). Those in the first tab are included in the name-suggestion-index but not explicitly identified as chains via English Wikipedia categories. Those in the second tab are the other way around: explicitly identified as chains via Wikipedia categories but omitted from the name-suggestion-index.

Steps to reproduce:

1. Install [`jq`](https://stedolan.github.io/jq/).
1. Clone the [name-suggestion-index](https://github.com/osmlab/name-suggestion-index/) repository.
[...]

PR #1975, opened 2018-10-19
Add Coop Prix to index, re #1934.

maxerickson:
A bit of a corner case as the individual stores often include the [city they are located in in the name](http://overpass-turbo.eu/s/CVc), but they do operate under the brand.


bhousel:
Thanks @maxerickson 👍

PR #1988, opened 2018-10-19
Added info for Days Inn and removed motel entry

eigenbrot:
Wikipedia lists industry as Hotel, not motel, and there were more matches for
hotel.


bhousel:
Thanks @eigenbrot 👍

PR #1989, opened 2018-10-19
Add the correct wikidata item for 'Nahkauf'

ENT8R:
Follow-up of #1947
I added the correct wikidata item and removed the link to the wikipedia entry of the "Rewe Group" because this is too generic and is not the actual brand...


bhousel:
Nice! Thanks @ENT8R ..

PR #2001, opened 2018-10-19
Remove shop/bakery|نانوایی بربری

wmateam:
According to the [Wikipedia](https://en.wikipedia.org/wiki/Barbari_bread), the Barbari bakery (نانوایی بربری) is a kind of bakery in my country (Iran) and  United States and Canada.
This kind of bakery is not a brand and does not have `"brand:wikidata"`.


matkoniecz:

matkoniecz:
What $نان filter was supposed to do? نانوایی بربری was not filtered out and would be readded by `npm run build`. I fixed that. But I am not sure whatever $نان filter was supposed to be نانوایی بربری filter (I assumed that in my fix) or is it supposed to filter out something else.

PR #2003, opened 2018-10-19
removing generic: Tabacchi

good-praxis:
Found a german [wikipedia site](https://de.wikipedia.org/wiki/Tabacchi), these are generic italian tabacco shops


good-praxis:
solves #1813

PR #2005, opened 2018-10-19
Add Fourchette tags

y0nd3rb0y:
solves #1789


good-praxis:
_Meta_: if you just write "solves #1789" or "closes #1789" Github will automatically close the issue once this PR request is merged!

PR #2015, opened 2018-10-19
remove generic:Victoria, fixes #1869

### LINKED ISSUE #1869: Research needed for: tourism/hotel|Victoria
#### :thinking: We need help deciding what to do with "tourism/hotel|Victoria"

### What's needed

First, do some research on `tourism/hotel|Victoria` by searching Google and Wikipedia.

  * Find the entry for `tourism/hotel|Victoria`
* `npm run build`

* Edit `config/canonical.json` to remove the entry for `tourism/hotel|Victoria`
* `npm run build`

:sparkles: Thanks!

good-praxis:


config/filters.json
```diff
@@ -190,6 +190,7 @@
         "^trafik(a)?$",
         "^urząd pocztowy$",
         "^venezia$",
+        "^victoria$",
```
ENT8R: Maybe this should go into the `discardKeys` section of `filters.json` because there might be other stores named "Victoria" which are not hotels...

good-praxis: In which case we'll get an error once one pops up which hasn't been removed form the list, right? This only filters items which have explicitly been removed form the canonical list, right?

That's at least the way I understood it (No additionally errors arose when I ran the build, btw)

bhousel: yea this is kind of a judgement call whether we're discarding the word "victoria" for all businesses or the word "victoria" just for hotels.

In this case it probably doesn't matter, but I think I'd rather see it in the "discardKeys" section, in case there is some convenience store or gas station chain named Victoria that we don't find out about until later.

PR #2016, opened 2018-10-19
fixing coral bookmaker entry

good-praxis:


ENT8R:
Oh thanks 👍 I overlooked something 😄

PR #2019, opened 2018-10-19
correcting kyriad entry, wikidata now more specific

### LINKED ISSUE #1862: Research needed for: tourism/hotel|Kyriad
#### :thinking: We need help deciding what to do with "tourism/hotel|Kyriad"

### What's needed

First, do some research on `tourism/hotel|Kyriad` by searching Google and Wikipedia.

  * Find the entry for `tourism/hotel|Kyriad`
* `npm run build`

* Edit `config/canonical.json` to remove the entry for `tourism/hotel|Kyriad`
* `npm run build`

:sparkles: Thanks!

good-praxis:
fixes #1862 
The Kyriad hotels _do_ belong to the [Groupe du Louvre](https://en.wikipedia.org/wiki/Groupe_du_Louvre), however they do have their own [wikidata entry](https://www.wikidata.org/wiki/Q11751808) with a [corresponding wikipedia article](https://pl.wikipedia.org/wiki/Kyriad)

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -33568,6 +33568,7 @@
   },
   "tourism/hotel|Kyriad": {
     "count": 76,
+    "countryCodes": ["fr"],
     "tags": {
       "brand": "Kyriad",
       "brand:wikidata": "Q11751808",
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -33568,11 +33568,10 @@
   },
   "tourism/hotel|Kyriad": {
     "count": 76,
-    "countryCodes": ["fr"],
     "tags": {
       "brand": "Kyriad",
-      "brand:wikidata": "Q529212",
-      "brand:wikipedia": "fr:Louvre Hotels Group",
+      "brand:wikidata": "Q11751808",
+      "brand:wikipedia": "pl:Kyriad",
       "name": "Kyriad",
       "tourism": "hotel"
     }


config/canonical.json
```diff
@@ -33568,11 +33568,10 @@
   },
   "tourism/hotel|Kyriad": {
     "count": 76,
-    "countryCodes": ["fr"],
     "tags": {
       "brand": "Kyriad",
-      "brand:wikidata": "Q529212",
-      "brand:wikipedia": "fr:Louvre Hotels Group",
+      "brand:wikidata": "Q11751808",
+      "brand:wikipedia": "pl:Kyriad",
```
ENT8R: Why do you link to the Polish wikipedia page if this hotel does only exist in France?

ENT8R: See also https://overpass-turbo.eu/s/CWh

ENT8R: Mmhh.. that is some kind of suboptimal 😞 But as far as I can see, this hotel does only exist in France, so a Polish wikipedia article does not make that much sense... Maybe we should just remove it. What you you think @bhousel @matkoniecz ?

matkoniecz: IMHO linking weird language version is preferable over not linking wikipedia at all. For example some motorway crossings in Poland have German-language wikipedia links, as there is no Polish language article about them.

bhousel: > IMHO linking weird language version is preferable over not linking wikipedia at all. For example some motorway crossings in Poland have German-language wikipedia links, as there is no Polish language article about them.

yeah I agree with @matkoniecz - linking a weird wiki language page is better than linking no language at all 

In fact, the `brand:wikipedia` links are known to be pretty unstable.  However the `brand:wikidata` link is the much more stable id.

We can always go thru later and run a bulk edit on OpenStreetMap to cleanup the tags if better ones become available.  This is only possible once the editor start to assign the tags though - so yeah, imperfect tag is better than no tag 😄

PR #2023, opened 2018-10-19
Holiday inn

good-praxis:
These don't correspond to an issue,
Both belong to the same company but are distinct brands with distinct wikidata entries

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -33411,6 +33411,8 @@
     "count": 436,
     "tags": {
       "brand": "Holiday Inn",
+      "brand:wikidata": "Q2717882",
+      "brand:wikipedia": "en:Holiday Inn",
       "name": "Holiday Inn",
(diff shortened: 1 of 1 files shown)


eigenbrot:
Thanks @good-praxis. There's also an entry for "Holiday Inn Express & Suites". Could you please address this as well to put the whole Holiday Inn question to rest. Thanks!

good-praxis:
@eigenbrot Actually I describe why I didn't do that in the "Holiday Inn Express & Suites" Issue #1850

eigenbrot:
OK, I think this will be taken care of in #2063. This is a similar situation to #1905; the "Express & Suites" might be hard to find online as a unique entity, but there are clearly businesses on the ground that are branded in this way. Having "Express & Suites" link to the the "Express" wiki info seems OK because that's how the Holiday Inn website does it anyway.

PR #2030, opened 2018-10-19
Removing generic: Kassa

y0nd3rb0y:
solves #1809


config/filters.json
```diff
@@ -217,6 +217,7 @@
         "^золото(й)?$",
         "^зоо(магазин|товары)$",
         "^канцтовары$",
+        "^касса$"
```
ENT8R: There needs to be a comma after the regex. Otherwise the build does not work.

y0nd3rb0y: my bad, comma is there now

matkoniecz:
Thanks!

Issue #1880, opened 2018-10-17
Figure out how to handle: different brands, different countries, identical name


andrewharvey:
Should we add a country filter to each item to limit that name suggestion to a certain country only.

eg. at the moment we have

name=Target
shop=department_store
brand:wikipedia=en:Target Corporation

But this is very different to

name=Target
shop=department_store
brand:wikipedia=en:Target Australia

https://en.wikipedia.org/wiki/Target_Australia

bhousel:
Ah yeah this is where it gets complicated 😄 

I _think_ it's ok to have two entries like this:

```
  "shop/department_store|Target-AU": {
    "countryCodes": ["au"],
    "nocount": true,
    "nomatch": ["shop/supermarket|Target-US"],
    "tags": {
      "brand": "Target",
      "brand:wikidata": "Q7685854",
      "brand:wikipedia": "en:Target Australia",
      "name": "Target",
      "shop": "department_store"
    }
  },
  "shop/department_store|Target-US": {
    "countryCodes": ["us"],
    "match": ["shop/supermarket|Target"],
    "nocount": true,
    "nomatch": ["shop/supermarket|Target-AU"],
    "tags": {
      "brand": "Target",
      "brand:wikidata": "Q1046951",
      "brand:wikipedia": "en:Target Corporation",
      "name": "Target",
      "shop": "department_store"
    }
  },
```

But I have to check the code for if there are assumptions about how everything works... So don't add it just yet.

matkoniecz:
I think that the only way to solve that is to have two different entries with different countryCodes.

Trademark law should prevent from two companies with the same name operating the same shop type.

PR #2039, opened 2018-10-20
Resolve duplicate name warning for Giant

xeluna:
The bicycle shop and supermarket are different.

But for the supermarket, there's definitely multiple brands in different countries. 
Possibly for the bicycle shops as well. 

Those are not resolved here.
Will need to do something like suggested in #1880


matkoniecz:
Thanks!

PR #2050, opened 2018-10-20
Research completed for: shop/supermarket|Tesco Metro

### LINKED ISSUE #1751: Research needed for: shop/supermarket|Tesco Metro
#### :thinking: We need help deciding what to do with "shop/supermarket|Tesco Metro"

### What's needed

First, do some research on `shop/supermarket|Tesco Metro` by searching Google and Wikipedia.

  * Find the entry for `shop/supermarket|Tesco Metro`
* `npm run build`

* Edit `config/canonical.json` to remove the entry for `shop/supermarket|Tesco Metro`
* `npm run build`

:sparkles: Thanks!

AlbyIanna:
closes #1751

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -31384,6 +31384,7 @@
       "brand": "Tesco Metro",
       "brand:wikidata": "Q487494",
       "brand:wikipedia": "en:Tesco",
+      "countryCodes": ["gb"],
       "name": "Tesco Metro",
       "shop": "supermarket"
     }
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -31382,8 +31382,7 @@
     "count": 152,
     "tags": {
       "brand": "Tesco Metro",
-      "brand:wikidata": "Q487494",
-      "brand:wikipedia": "en:Tesco",
+      "brand:wikidata": "Q57551648",
       "countryCodes": ["gb"],
       "name": "Tesco Metro",
       "shop": "supermarket"
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -31382,6 +31382,8 @@
     "count": 152,
     "tags": {
       "brand": "Tesco Metro",
+      "brand:wikidata": "Q487494",
+      "brand:wikipedia": "en:Tesco",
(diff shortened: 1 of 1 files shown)


ENT8R:
According to https://overpass-turbo.eu/s/CWR this seems to be restricted to Great Britain. Is that true?

AlbyIanna:
it seems so, but I couldn't find anything that confirmed the hypothesis. If you want me to add it i'll do it though

config/canonical.json
```diff
@@ -31382,6 +31382,8 @@
     "count": 152,
     "tags": {
       "brand": "Tesco Metro",
+      "brand:wikidata": "Q487494",
```
ENT8R: I also think that we should use Q57551648 instead of the generic Tesco entry...

config/canonical.json
```diff
@@ -31382,6 +31382,8 @@
...
+      "brand:wikipedia": "en:Tesco",
```
ENT8R: ...and maybe it is better to remove the `brand:wikipedia` tag because the actual brand is Tesco Metro and not Tesco...

ENT8R:
> If you want me to add it i'll do it though

Yes, please. I read it up on wikipedia and it seems to be GB-only

config/canonical.json
AlbyIanna: right

config/canonical.json
AlbyIanna: okey dokey

config/canonical.json
```diff
@@ -31382,8 +31382,8 @@
     "count": 152,
     "tags": {
       "brand": "Tesco Metro",
-      "brand:wikidata": "Q487494",
-      "brand:wikipedia": "en:Tesco",
+      "brand:wikidata": "Q57551648",
+      "countryCodes": ["gb"],
```
ENT8R: This value does not belong into the tags object. You have to add it below the value of count:*

good-praxis:
Too bad `Q57551648` doesn't have a wikipedia article attached, but thanks for your contribution!

PR #2053, opened 2018-10-20
Update Deutsche Post entries

lucaswerkmeister:
Merge the two post_office entries, and add country codes and Wikipedia and Wikidata links to the remaining Deutsche Post entries.

---

Also, apparently I added the missing trailing newline to this file when I saved it in Emacs. Should I remove it?

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -12928,6 +12928,7 @@
   },
   "amenity/post_box|Deutsche Post": {
     "count": 158,
+    "countryCodes": ["de"],
     "match": [
       "amenity/post_box|Deutsche Post AG"
     ],
@@ -12937,6 +12938,8 @@
     "tags": {
       "amenity": "post_box",
       "brand": "Deutsche Post",
+      "brand:wikidata": "Q157645",
+      "brand:wikipedia": "en:Deutsche Post",
       "name": "Deutsche Post"
     }
   },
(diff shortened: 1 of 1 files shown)


ENT8R:
You just need to add `"match": ["amenity/post_office|Deutsche Post Filiale"]` to the entry of the `Deutsche Post`

And I would also add
```
"brand:wikidata": "Q157645",
"brand:wikipedia": "en:Deutsche Post",
```
to the entry of the post box. But if you feel unsure about something, just tell me and I will do it for you 😄

> Also, apparently I added the missing trailing newline to this file when I saved it in Emacs. Should I remove it?

Don't mind about the newline...

lucaswerkmeister:
Shouldn’t I alse remove the Filiale entry (and add its count to the main one)?

ENT8R:
> and add its count to the main one

Please don't modify the count by hand, this won't work. But after you added `"match": ["amenity/post_office|Deutsche Post Filiale"]` to the entry of the `Deutsche Post` you should remove the entry, that is true.

lucaswerkmeister:
Alright, new version pushed.

ENT8R:
Thank you very much for your contribution 🎉 👍

lucaswerkmeister:
Thanks for helping me out and merging it :heart:

PR #2067, opened 2018-10-21
shop jewelry pandora

eczajk1:
for #1532 

see [Wikipedia entry](https://en.wikipedia.org/wiki/Pandora_(jewelry))

see [Wikidata entry](https://www.wikidata.org/wiki/Q2241604)

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -25724,6 +25724,8 @@
     "count": 387,
     "tags": {
       "brand": "Pandora",
+      "brand:wikidata": "Q2241604",
+      "brand:wikipedia": "en:Pandora_(jewelry)",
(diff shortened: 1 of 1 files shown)


ENT8R:
@good-praxis did you looked at the source code? There is an issue with the `brand:wikipedia` value 😄
(An underscore too much...)

good-praxis:
@ENT8R whoops, the underscore went straight over my head there, was too impressed by them linking their sources and making sure they matched the entry

ENT8R:
I removed it with https://github.com/osmlab/name-suggestion-index/commit/a0065f7faed40fb0a677f48fe447d01a2be8aecc 😄

PR #2071, opened 2018-10-21
Add Swedish chain of charity shops Myrorna (re: #1934)

Vesihiisi:
Re: #1934.

This is a chain of charity shops operated by the Salvation Army in Sweden, with >30 locations.


ENT8R:
Thank you 👍

PR #2073, opened 2018-10-21
adding validator for brand:wikipedia

good-praxis:
Hey, I've added a validator to catch underscores in brand:wikipedia.
Let me know if this works for you / if this should be handled differently


good-praxis:
This may very well be too aggressive, maybe outputting a warning would be better?

ENT8R:
The output of a warning is already done with
https://github.com/osmlab/name-suggestion-index/blob/master/build_filterNames.js#L329#L336
sc

good-praxis:
Whoops, that's right! Thanks @ENT8R

PR #2074, opened 2018-10-21
Correct country code Sweden sv → se

Vesihiisi:
Change country code for Sweden from `sv` to `se` in 2 entries.


ENT8R:
Thanks for spotting these 👍

PR #2077, opened 2018-10-21
Add Swedish jewelry chain Guldfynd (re: #1934)

Vesihiisi:
Add jewelry chain Guldfynd, with 100+ locations in Sweden.


ENT8R:
Thanks 👍

PR #2082, opened 2018-10-21
Add Polish electronics chain Komputronik

Vesihiisi:
Add Polish electronics chain Komputronik.

I'm not super sure about the tags. Most of the shops I'm finding on osm are `shop=computer`, but according to their [website](https://www.komputronik.pl/) they also sell TV's, phones etc...

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -24156,6 +24156,17 @@
       "shop": "electronics"
     }
   },
+  "shop/electronics|Komputronik": {
+    "nocount": true,
+    "countryCodes": ["pl"],
+    "tags": {
+      "brand": "Komputronik",
+      "brand:wikidata": "Q11742085",
+      "brand:wikipedia": "pl:Komputronik",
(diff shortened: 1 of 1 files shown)


matkoniecz:
shop=electronis should be OK, at least I would also map it this way.

matkoniecz:
> This branch has conflicts that must be resolved

Is repasing somathing that you can do or want to learn doing that? See https://git-scm.com/book/en/v2/Git-Branching-Rebasing if you want to do the second version.
below I quoted new code to make sure it is not lost if rebase goes wrong

```
  "shop/electronics|Komputronik": {
    "nocount": true,
    "match": ["shop/computer|Komputronik"],
    "countryCodes": ["pl"],
    "tags": {
      "brand": "Komputronik",
      "brand:wikidata": "Q11742085",
      "brand:wikipedia": "pl:Komputronik",
      "name": "Komputronik",
      "shop": "electronics"
    }
  },
```

PR #2113, opened 2018-10-22
add G20 wiki data

alexbrbr:


good-praxis:
Hey @alexbrbr!
This PR could use a countryCode tag, G20 seems to be only available in France!

alexbrbr:
I added the `"countryCodes": ["fr"]`, thanks for your feedback

good-praxis:
fixes #1685

PR #1868, opened 2018-10-17
Add Tags to Price chopper

mraguso2:
issue: https://github.com/osmlab/name-suggestion-index/issues/1734 
Add tags to Price Chopper

PR #2130, opened 2018-10-23
Removed pseudo-generic Vision Plus entry

eigenbrot:
This one is really tricky so I need some advice/consensus.

There are a lot of unrelated "Vision Plus" opticians, which would suggest a generic name. That said, it wasn't hard to find at least one small chain of opticians with multiple locations in the same area (in this case Washington state).

Here is a small sample of generic Vision Pluses:
http://www.visionplustx.com/
https://www.nhs.uk/Services/opticians/Overview/DefaultView.aspx?id=16671

and the chain in Washington. Some (if not all) of the 14 locations have their own websites, which can vary wildly in style and branding:
http://www.visionpluswa.com/
http://www.vpmillcreek.com/
https://www.visionplusballard.com/

Vision Plus Washington does not have any wiki entries. There is a wikipedia entry for VisionPLUS, which is a financial software application.

So my current thought is to just make Vision Plus a generic brand (i.e., this PR). The logic here is that:

a) There is no wiki data for Vision Plus Washington
b) There are other non-affiliated Vision Pluses all over the US (and at least one in the UK)
c) People on the ground in Washington will see Vision Plus and know what it means (hopefully)

This is similar to #1868, but in this case it's a question of one brand and a bunch of generics.

edit: forgot to mention VisionPLUS financial company

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -26534,14 +26534,6 @@
       "shop": "optician"
     }
   },
-  "shop/optician|Vision Plus": {
-    "count": 59,
-    "tags": {
-      "brand": "Vision Plus",
-      "name": "Vision Plus",
-      "shop": "optician"
-    }
-  },
   "shop/optician|Visionworks": {
     "count": 59,
     "tags": {
--- a/config/filters.json
+++ b/config/filters.json
@@ -215,6 +215,7 @@
         "^vegyesbolt$",
         "^venezia$",
         "^virágbolt$",
+	"^Vision\\sPlus$",
         "^(clinica\\s)?veterinari(a|o)$",
         "^(clinique\\s)?vétérinaire$",
         "^warung$",
--- a/dist/discardNames.json
+++ b/dist/discardNames.json
@@ -4081,6 +4081,7 @@
   "shop/optician|Optica": 184,
   "shop/optician|Optical Center": 161,
   "shop/optician|Optika": 68,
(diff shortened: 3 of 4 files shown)


eigenbrot:
Yeesh, there's also a French brand called Vision Plus that has a wikidata entry (Q23752663) but no wikipedia page (?!).

good-praxis:
> Yeesh, there's also a French brand called Vision Plus that has a wikidata entry (Q23752663) but no wikipedia page (?!).
That's more common than you maybe think. Sadly

matkoniecz:
> Yeesh, there's also a French brand called Vision Plus that has a wikidata entry (Q23752663) but no wikipedia page (?!).

There is hint in history and in current page: see initial version at https://www.wikidata.org/w/index.php?title=Q23752663&oldid=319667424

Deletion log at frwiki reveals that it was deleted https://fr.wikipedia.org/w/index.php?title=Sp%C3%A9cial:Journal&type=delete&user=&page=Vision+Plus&wpdate=&tagfilter=&subtype=

Apparently it is viable wikidata entry but not a Wikipedia article (that was created almost certainly for spam purposes as most articles of small companies).

matkoniecz:
I guess that given tiny chain and substantial generic use filtering it out is the best solution.

matkoniecz:
> This one is really tricky so I need some advice/consensus.

Hopefully "checked and confirmed by another human" is good enough.

Issue #1504, opened 2018-10-17
Research needed for: shop/hairdresser|Sport Clips


bhousel:
#### :thinking: We need help deciding what to do with "shop/hairdresser|Sport Clips"

:point_right: Read the [CONTRIBUTING.md](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) guide for more info, and feel free to ask questions on this issue!

### What's needed

First, do some research on `shop/hairdresser|Sport Clips` by searching Google and Wikipedia.

#### If it is an actual brand:

* Edit `config/canonical.json`:
  * Find the entry for `shop/hairdresser|Sport Clips`
  * Add tags for `brand:wikidata` and `brand:wikipedia`
  * Optionally add `"countryCodes"` or other tags
* `npm run build`
* Commit your changes and submit a pull request

#### If it is not an actual brand:

* Edit `config/filters.json` to add an expression that discards it
* Edit `config/canonical.json` to remove the entry for `shop/hairdresser|Sport Clips`
* `npm run build`
* Commit your changes and submit a pull request

#### If the situation is complicated:

* Leave a comment on this issue with what you find

:sparkles: Thanks!

_This is an autogenerated ticket_

jgon6:
There is not a Wikipedia page for this brand. The wikidata id is: Q7579310.

Adamant36:
Reopening because Wikidata page doesn't exist anymore for some reason (or I screwed up and put in the wrong code). But it doesn't seem to have another Wikidata entry. So I'll add the "needs Wikidata" label to it at least.

bhousel:
Thanks!  We had another situation yesterday where a Wikidata admin deleted a brand that we were using because it didn't seem to meet their [notability threshold](https://www.wikidata.org/wiki/Wikidata:Notability).

If that's what happened, I can ping them in IRC and ask for it to be restored.  They are very nice!

We believe that anything from name-suggestion-index should automatically be notable, as this is an index of branded businesses operating multiple stores in the real world.  

However if people just set up a Wikidata page that's light on details, the admins may reasonably think it is not a real thing.  So it's important to add details like this to any Wikidata page we create:
- a good `label` and `description`
- `instance_of = business`  (or [pick a better one from this list](https://gist.github.com/1ec5/8d44e94db9ac23c0b859c2718d7f82c3))
- `official_website` = paste their website url
- Facebook, Twitter, Instagram if you can find it

bhousel:
[...]

PR #2131, opened 2018-10-23
Added wikidata for Batteries Plus Bulbs

eigenbrot:
Addresses #1428

Weirdly, this brand has a wikidata entry, but no wikipedia page. A la #1504 the
entry just won't have the "brand:wikipedia" tag for now.

Also added match for "Batteries Plus". This is their web address and wikidata
name, but I'm pretty sure all stores are branded "Batteries Plus Bulbs". If I
remember correctly when they started out there were some stores just called
"Batteries Plus" so that's probably where the URL and wikidata come from. I can
remove the match to "Batteries Plus" if ya'll think it's superfluous.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -23813,8 +23813,10 @@
   },
   "shop/electronics|Batteries Plus Bulbs": {
     "count": 86,
+    "match": ["shop/electronics|Batteries Plus"],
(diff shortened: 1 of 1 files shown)


matkoniecz:
Thanks!

PR #2133, opened 2018-10-23
Updated some match entries for Lowe's

eigenbrot:
Needed match for hardware|Lowe's and to keep doityourself|Lowes


maxerickson:
Looks good.

PR #2136, opened 2018-10-24
Fix duplication of Western Union and add brand:wikidata, brand:wikipedia

NMVikings:


matkoniecz:
Is it certain that amenity=bank is always preferred over amenity=money_transfer, for all countries? For example in Africa?

NMVikings:
I think, that Western union is money transfer firm, but I didn't find `money_transfer` on https://wiki.openstreetmap.org/wiki/Map_Features. That's why i chose amenity=bank.

Quote from [Wikipedia](https://en.wikipedia.org/wiki/Western_Union)
> Western Union has several divisions, with products such as person-to-person money transfer, money orders, business payments, and commercial services.

good-praxis:
Yeah, this is kind of the same discussion that we had in #1943 about WU @matkoniecz 
Technically WU only operates with a bank license in a few countries in europe, and probably only for regulations.

However I think there will be places where it only allows money transfer. That's their main buisness, afterall. But maybe this is close enough to a bank?

matkoniecz:
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dmoney_transfer is also documented, with Western Union as an example.

matkoniecz:
I think that we may need to keep both entries - Wester Union as a bank and as a money transfer.

NMVikings:
But there is match between them. They aren't different organisations.

maxerickson:
Different faces of the same organization can need different modeling though.

In the US, Western Union usually is available at a service counter in a different business and doesn't really exist as an independent location/business.

matkoniecz:
> But there is match between them. They aren't different organisations.

Tesco operates fuel station and supermarkets. They share the same brand but are otherwise tagged differently.

There is no reason to tag Western Union as a bank in places where it is not one (not, I am not 100% sure whatever amenity=money_transfer should be promoted, but at least it is used).

bhousel:
> I think that we may need to keep both entries - Wester Union as a bank and as a money transfer.

This is what I did in b3507c7.. We can just include both.

PR #2139, opened 2018-10-24
Fix Dixy

NMVikings:


eigenbrot:
Looks good. Good call using the Russian wikipedia page. Thanks @NMVikings!

PR #2140, opened 2018-10-24
Fix Victoria duplication

NMVikings:


eigenbrot:
Looks good. Thanks @NMVikings!

PR #2142, opened 2018-10-24
Merge duplicate Marshalls and add wiki tags. 

maxerickson:
shop=department_store is less common in the data than shop=clothes but the stores are larger and offer a wider range of merchandise.


eigenbrot:
Agreed. Thanks @maxerickson!

PR #2144, opened 2018-10-24
Halfords

eigenbrot:
Two things done here, one easy and one not so cut-and-dry:

1. Halfords Autocentre is clearly a related but separate brand from Halfords. They have individual wikipedia and wikidata entries. To minimize confusion I enforced the canonical car_repair|Halfords Autocentre over car_repair|Halfords, which had slightly more matches.

2. The brand Halfords is a little trickier. According to the [wikidata](https://www.wikidata.org/wiki/Q3398786) this is a car_parts store. But from [wikipedia](https://en.wikipedia.org/wiki/Halfords) and their [website](https://www.halfords.com/) it is clear that they also sell bicycles and camping gear. Furthermore, there are more matches for bicycle|Halfords than car_parts|Halfords. In the current PR I've made car_parts the canonical category and have it match to bicycle, but clearly more OSM users view this as a bicycle shop so I'm not sure this is correct. Halfords' own website title bar even lists "Bikes, Cycling" before car parts.

I also had the thought of adding a match for shop/outdoor|Halfords as a preliminary strike on someone using that tag after they buy a tent or whatever. This is currently not in the PR because I'm worried it might be too cute or convoluted. Opinions welcome!

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -16095,14 +16095,6 @@
       "shop": "bicycle"
     }
   },
-  "shop/bicycle|Halfords": {
-    "count": 164,
-    "tags": {
-      "brand": "Halfords",
-      "name": "Halfords",
-      "shop": "bicycle"
-    }
-  },
   "shop/bicycle|サイクルベースあさひ": {
     "count": 105,
     "countryCodes": ["jp"],
@@ -16561,8 +16553,14 @@
   },
   "shop/car_parts|Halfords": {
     "count": 114,
+    "match": ["shop/bicycle|Halfords"],
+    "nomatch": [
+      "shop/car_repair|Halfords Autocentre"
+    ],
     "tags": {
       "brand": "Halfords",
+      "brand:wikidata": "Q3398786",
+      "brand:wikipedia": "en:Halfords",
       "name": "Halfords",
       "shop": "car_parts"
     }
@@ -16802,6 +16800,7 @@
   "shop/car_repair|Halfords Autocentre": {
     "count": 51,
     "match": ["shop/car_repair|Halfords"],
+    "nomatch": ["shop/car_parts|Halfords"],
     "tags": {
       "brand": "Halfords Auocentre",
       "brand:wikidata": "Q5641894",
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -16562,6 +16562,7 @@
       "brand:wikidata": "Q3398786",
       "brand:wikipedia": "en:Halfords",
       "name": "Halfords",
+      "service:bicycle:retail": true,
       "shop": "car_parts"
     }
   },
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -16562,7 +16562,7 @@
       "brand:wikidata": "Q3398786",
       "brand:wikipedia": "en:Halfords",
       "name": "Halfords",
-      "service:bicycle:retail": true,
+      "service:bicycle:retail": "yes",
       "shop": "car_parts"
     }
   },
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -16799,14 +16799,14 @@
       "shop": "car_repair"
     }
   },
-  "shop/car_repair|Halfords": {
-    "count": 56,
-    "match": [
-      "shop/car_repair|Halfords Autocentre"
-    ],
+  "shop/car_repair|Halfords Autocentre": {
+    "count": 51,
+    "match": ["shop/car_repair|Halfords"],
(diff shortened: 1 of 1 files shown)


matkoniecz:
There is (frequently proposed, rarely used ) `sells:*` tag scheme. In taginfo I also found `service:bicycle:retail` (5k uses) and `service:bicycle:sales`.

eigenbrot:
@matkoniecz, that sounds like it might be a good solution. I'm afraid I'm not familiar with any of those tags you referenced, though, and could not find any examples in `canonical.json`. Are they just added to the `tags` section for a particular brand?

matkoniecz:
> Are they just added to the tags section for a particular brand?

Yes - tags may contain any tags that will be applied - https://github.com/osmlab/name-suggestion-index/blob/b3507c762469ec1f301038afbbae4f3d94d20ed3/config/canonical.json#L6557

> I'm afraid I'm not familiar with any of those tags you referenced

https://wiki.openstreetmap.org/wiki/Key:service:bicycle:retail

sells tag family is described https://wiki.openstreetmap.org/wiki/Proposed_features/Sells and has very limited use, but in this case I see no better way of tagging that

matkoniecz:
BTW, in tricky cases like this one I would also consult local community where this shop appears frequently. Maybe they will have some good ideas/preferences.

eigenbrot:
Thanks for the links, @matkoniecz! That first one has an example that feels eerily relevant :).

I tried adding `sells: bicycles` (pluralization matters?), `sells:car_parts` (probably redundant), and `sells:outdoors` (maybe should be `outdoors:camping`?), but `npm run build` seems to only allow a single `sells` tag. Is this expected?

eigenbrot:
Oh wait, perhaps the solution is to put them all into a list:

`sells: ["bicycle", "car_parts", "outdoors"]`

Is that right? It at least gets left alone by `npm run build`. Thanks for dealing with my n00bish questions!

bhousel:
> Is that right? It at least gets left alone by `npm run build`. Thanks for dealing with my n00bish questions!

Nah, I'd skip the `sells` tags.  These aren't very well established in OpenStreetMap yet.

eigenbrot:
OK, I removed the `sells` tag and pushed a commit that just adds `"service:bicycle:retail": true`. Is that good enough?

Also, related to @matkoniecz comment about local community, I did a bunch of Googling and Halfords is often referred to as a big box store that sells car and bike stuff. That said, all of their official pages (twitter, facebook, etc.) list the car parts first.

matkoniecz:
`"service:bicycle:retail": true,`

I think it should be

`"service:bicycle:retail": "yes",`

true may end getting converted to "yes" string but I would not rely on that.

matkoniecz:
> I tried adding sells: bicycles (pluralization matters?), sells:car_parts (probably redundant), and sells:outdoors (maybe should be outdoors:camping?), but npm run build seems to only allow a single sells tag. Is this expected?

sells idea was nixed, but it would be

```
"sells:car_parts": "yes",
"sells:bicycles": "yes",
"sells:whatever_else": "yes",
"sells:yet_another_key": "yes",
```

eigenbrot:
> I think it should be
> 
> `"service:bicycle:retail": "yes",`
> 
> `true` may end getting converted to `"yes"` string but I would not rely on that.

Is this true?(hehe) I can find no instances of `"key": "yes"` in `canonical.json` but many cases of `"key": true` (e.g., `"nocount": true`). I have no experience with json so I'm very willing to be proven incorrect!

matkoniecz:
These are about a bit different things. `"nocount": true` is internal data (instruction to ignore one of checks during validation), `"service:bicycle:retail": "yes",` is definition of tag that will be added to OSM database.

>  I can find no instances of "key": "yes" in canonical.json

Probably it is the first boolean OSM tag added to the brand dataset. Note that most entries have just `name`, `brand`, `brand:wikipedia` and `brand:wikidata` in `tags` array. Additional OSM tags are quite rare.

eigenbrot:
OK, @matkoniecz thanks for the clarification! I've made that change in the most recent commit.

Phew, I think this might be ready for a merge!

matkoniecz:
One more thought: it may look confusing for user as "bicycles are also sold here" will not be well visible for example in iD as user selects object. But I am not sure how it may be fixed, keep both `shop=bicycle sells:car_parts=yes` and `shop=car_parts service:bicycle:retail=yes`?

matkoniecz:
@eigenbrot Thanks!

PR #2155, opened 2018-10-26
Wiki updated for Supersol #1749

thilinatnt:


bhousel:
are you sure about this one?   It's not clear that Supersol === Shufersal

thilinatnt:
@bhousel Ah actually Supersol has changed to Shufersal, please see the second search item here - https://en.wikipedia.org/w/index.php?search=SuperSol&title=Special:Search&fulltext=1

bhousel:
ok, I found the website for them..
http://en.supersol.es/

PR #2156, opened 2018-10-26
Wiki updated for Superama #1746

thilinatnt:


eigenbrot:
Is this information correct? The wiki info added is for Walmart of Mexico and Central America. Superama is owned by that organization, but appears to have its own [wikipedia](https://es.wikipedia.org/wiki/Superama) and [wikidata](https://www.wikidata.org/wiki/Q6135762) pages

thilinatnt:
@eigenbrot Oh yes you are correct, think I missed it, let me update PR

thilinatnt:
Will create a new PR for this

PR #2157, opened 2018-10-26
Wiki update for Suma #1744

thilinatnt:


eigenbrot:
Wiki info looks good!

I'm wondering about the `supermarket` tag, though. According to their [website](http://www.suma.coop/) and wikipedia they are a food wholesaler and it doesn't look like they sell to individuals. Maybe `wholesale` would be a better key?

thilinatnt:
@eigenbrot Yeah that seems much better key

thilinatnt:
Let me create a new PR to fix this

Issue #1741, opened 2018-10-17
Research needed for: shop/supermarket|Shaw's


bhousel:
#### :thinking: We need help deciding what to do with "shop/supermarket|Shaw's"

:point_right: Read the [CONTRIBUTING.md](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) guide for more info, and feel free to ask questions on this issue!

### What's needed

First, do some research on `shop/supermarket|Shaw's` by searching Google and Wikipedia.

#### If it is an actual brand:

* Edit `config/canonical.json`:
  * Find the entry for `shop/supermarket|Shaw's`
  * Add tags for `brand:wikidata` and `brand:wikipedia`
  * Optionally add `"countryCodes"` or other tags
* `npm run build`
* Commit your changes and submit a pull request

#### If it is not an actual brand:

* Edit `config/filters.json` to add an expression that discards it
* Edit `config/canonical.json` to remove the entry for `shop/supermarket|Shaw's`
* `npm run build`
* Commit your changes and submit a pull request

#### If the situation is complicated:

* Leave a comment on this issue with what you find

:sparkles: Thanks!

_This is an autogenerated ticket_

mds08011:
"Shaw's" and "Star Market" are regional supermarket chains which share both a management team and a Wikipedia page.  These will likely need to be merged/matched which is currently beyond my expertise.  

Wikipedia: https://en.wikipedia.org/wiki/Shaw%27s_and_Star_Market

bhousel:
In this index, it's ok for both of these brands to point to the same `brand:wikidata` and `brand:wikipedia` tags.  

We can detect this situation and circle back with the wikipedia folks on whether it makes sense to split into subbrands with separate wikidata ids.

PR #2158, opened 2018-10-26
Wiki updated with Shaw's #1741

thilinatnt:


eigenbrot:
Looks good. Thanks @thilinatnt!

good-praxis:
@eigenbrot I noticed a discrepancy in the data thanks to my latest commit, pretty much the fact that this uses the `%27` instead of just `'` in the wikipedia entry. I've cleaned that up in df076a0 though!

eigenbrot:
Great, thanks @good-praxis!

PR #2159, opened 2018-10-26
Wiki updated Santa Isabel #1740

thilinatnt:


eigenbrot:
Wiki info looks good. This brand has both English and Spanish wikipedia pages and is only located in a Spanish-speaking country (Chile). Which wikipedia entry should it use? I remember this issue coming up before but can't recall what the best practice was.

bhousel:
Good idea @eigenbrot ..
I merged it and then switched it to the `es:` wiki, and added `"countryCodes": ["cl"]` in 2b1fd87

PR #2161, opened 2018-10-26
Wiki updated for rewe city #1739

thilinatnt:


eigenbrot:
Thanks @thilinatnt. I understand why you chose to use the parent company (REWE Group) as the wiki references, but this might be confusing for people who are expecting a grocery store. A German friend of mine described REWE City as "just a small REWE", so maybe it would make sense to just have it match to `shop/supermarket|Rewe`. I'm not sure, comments welcome!

ENT8R:
As I'm a German, I would say that it would make sense to merge the entries the way @eigenbrot proposes. Rewe City is really just the name of a small Rewe, but sometimes there is no difference at all except that a Rewe City has to the in the city (and not around)...

bhousel:
Thanks! The decision to merge really depends on whether you want the editor to suggest a single “Rewe” choice or two separate “Rewe” and “Rewe City” choices.

ENT8R:
This has been fixed with https://github.com/osmlab/name-suggestion-index/commit/07a3f32d83d399e90047e94a1cd1d9b1fda0b155

PR #2110, opened 2018-10-22
tiger mart (ExxonMobil)

eczajk1:
Resolves #1303

"Tiger Marts" are associated with ExxonMobil brand gas stations, such as the one mapped in [this vicinity](https://www.openstreetmap.org/search?query=tiger%20mart#map=19/38.89707/-104.83232).

Added [Wikipedia](https://en.wikipedia.org/wiki/ExxonMobil) and [Wikidata](https://www.wikidata.org/wiki/Q156238) entries for Exxon.

Issue #272, opened 2018-10-16
Research needed for: amenity/bank|Kasa Stefczyka


bhousel:
#### :thinking: We need help deciding what to do with "amenity/bank|Kasa Stefczyka"

:point_right: Read the [CONTRIBUTING.md](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) guide for more info, and feel free to ask questions on this issue!

### What's needed

First, do some research on `amenity/bank|Kasa Stefczyka` by searching Google and Wikipedia.

#### If it is an actual brand:

* Edit `config/canonical.json`:
  * Find the entry for `amenity/bank|Kasa Stefczyka`
  * Add tags for `brand:wikidata` and `brand:wikipedia`
  * Optionally add `"countryCodes"` or other tags
* `npm run build`
* Commit your changes and submit a pull request

#### If it is not an actual brand:

* Edit `config/filters.json` to add an expression that discards it
* Edit `config/canonical.json` to remove the entry for `amenity/bank|Kasa Stefczyka`
* `npm run build`
* Commit your changes and submit a pull request

#### If the situation is complicated:

* Leave a comment on this issue with what you find

:sparkles: Thanks!

_This is an autogenerated ticket_

matkoniecz:
> Add tags for brand:wikidata and brand:wikipedia

It has no own wikipedia entry. Brand is based on https://pl.wikipedia.org/wiki/Kasa_Stefczyka - but article is not about the current company.

matkoniecz:
https://pl.wikipedia.org/wiki/Sp%C3%B3%C5%82dzielcza_kasa_oszcz%C4%99dno%C5%9Bciowo-kredytowa describes this type of company, not just this one brand.

bhousel:
I don't know, @matkoniecz - use your best judgement.
If it's not really a brand, we can just discard it from the index.

matkoniecz:
It is a brand but currently there is no wikidata or wikipedia page for it. Is it OK to just create nearly empty Wikidata entry for that brand?

good-praxis:
Ah, sorry, the wikidata isn't accurate enough?

matkoniecz:
Done!

PR #2162, opened 2018-10-26
Wiki updated Punto simply #1737

thilinatnt:


eigenbrot:
The wiki info provided is for `shop/supermarket|Simply Market`, which already exists with full data in `canonical.json`. If "Punto Simply" is just the same as "Simply Market" then a `match` entry would be a better way to deal with this. Please check that "Punto Simply" is different/the same as "Simply Market".

eigenbrot:
@thilinatnt thanks for the clarification. The question remains on how to deal with this, though. If a user on the ground wouldn't care about the distinction between the two then I think a simple `match` entry on "Simply Market" (and removing "Punto Simply") would be OK. If, however, they are different in a way that matters then it doesn't make sense for "Punto Simply" to link to the same wiki info as "Simply Market".

#272 and #2110 are relevant here. Maybe just make a really bare-bones wikidata entry?

bhousel:
I did a bit of research and it does look like Punto Simply is a different sub brand of Simply Market.  

I'm going to merge as-is, since we have other similar cases in the index where we allow supermarket subbrands ("Rewe" / "Rewe City",  "ICA"/"ICA Maxi",  "Walmart" / "Walmart Supercenter", etc.)

PR #2163, opened 2018-10-26
Wiki updated for Superama #1746

thilinatnt:


eigenbrot:
Looks good. Thanks @thilinatnt!

In the future, you can just push a new commit to your old PR instead of creating a whole new one.

PR #2164, opened 2018-10-26
Wiki update for Suma #1744

thilinatnt:


eigenbrot:
@good-praxis, agreed. Another way would be to add `supermarket|Suma` to `filters.json`. Which is more correct?

good-praxis:
Hmm, that's right, although I wouldn't know if it's better. From a user standpoint, if I expect to find that brand in supermarkets (because I don't know any better) I would probably be confused not to find it? Especially if the user doesn't know the difference of wholesale/supermarket

eigenbrot:
I think having a `match` to `supermarket|Suma` would still have it show up when searching under supermarkets. Adding the supermarket entry to `filters.json` just prevents it from being automatically added to `canonical.json` during the build. So I guess what I'm saying is that @good-praxis  is correct, but you need to also add `supermarket|Suma` to `filters.json`. I think. @bhousel is that right?

bhousel:
oh @good-praxis is correct - we want it to look like this:

```
"shop/wholesale|Suma": {
   "match": ["shop/supermarket|Suma"],
   "nocount": true,
   "tags": {
      "brand": "Suma",
      "name": "Suma",
      "brand:wikidata": "Q9348064",
      "brand:wikipedia": "en:Suma (co-operative)",
      "shop": "wholesale"
   }
}
```

@eigenbrot - things in the "match" won't be presented as options in iD.  They are just alternative names.
It's like saying "If we see `shop/supermarket|Suma`, we really mean `shop/wholesale|Suma` ".

The filter list is more like saying "If we see `shop/supermarket|veggies`, it's a mistake and we don't want it".

PR #2170, opened 2018-10-28
Research needed for: shop/bakery|Panaderia #1020

Tubaleviao:
Solution for the issue #1020


ENT8R:
Thanks again! It is awesome that you can help us with the language you speak! 👍 🎉

PR #2172, opened 2018-10-28
Ambiguous names

bhousel:
solves #1880


bhousel:
This is a possible solution for #1880  - would love to hear everyone's feedback!

It works by using the tilde character `~` in the name to disambiguate between identical names 

Here is an example of "EKO" which is two different brands of fuel stations - one in Canada and one in Greece.  (Just to make things even more fun, there is also an Eko brand of supermarket in Poland).

Overpass:  https://overpass-turbo.eu/s/DaW

```  
"amenity/fuel|EKO~(Canada)": {
    "nocount": true,
    "countryCodes": ["ca"],
    "match": ["amenity/fuel|Eko~(Canada)"],
    "nomatch": [
      "shop/supermarket|Eko",
      "amenity/fuel|EKO~(Greece)"
    ],
    "tags": {
      "amenity": "fuel",
      "brand": "EKO",
      "brand:wikidata": "Q3045934",
      "brand:wikipedia": "fr:EKO",
      "name": "EKO"
    }
  },
  "amenity/fuel|EKO~(Greece)": {
    "nocount": true,
    "countryCodes": ["gr","rs","bg","cy","me","mk"],
    "match": ["amenity/fuel|Eko~(Greece)"],
    "nomatch": [
      "shop/supermarket|Eko",
      "amenity/fuel|EKO~(Canada)"
    ],
    "tags": {
      "amenity": "fuel",
      "brand": "EKO",
      "brand:wikidata": "Q903198",
      "brand:wikipedia": "el:Ελληνικά Πετρέλαια",
      "name": "EKO"
    }
  },
```



Some things about disambiguating the names this way:
- You can put anything you want after the `~` 
- When generating the `/dist/*` files, the script replaces the `~` with a space, so the preset it will display like "EKO (Canada)" or "EKO (Greece)"
- Because the name will never literally match "EKO~(whatever)":
  - We'll probably never be able to accurately count these in the planet and
  - We need to add `"nocount"` to suppress the warning.  I guess this is ok.
- I'm pretty sure no real major names use a `~` in their name, but we should _probably_ add something to `filters.json` to make sure we never consider a name like this.  (too bad for any brands that do)


🤔  Please let me know what you all think!  It seems to work ok...

maxerickson:
Could fallback to counting the `brand:wikidata` tag in planet.

On the other hand, the existence of an entry with `brand:wikidata` in the index can almost be treated like a nocount, because the thing to look for in the planet now is common names that don't have a NSI entry.

bhousel:
> On the other hand, the existence of an entry with `brand:wikidata` in the index can almost be treated like a nocount, because the thing to look for in the planet now is common names that don't have a NSI entry.

Yes good point @maxerickson - I am expecting that `count` will be become much less important after the matches have been made and the tags assigned.  It is really only useful when we are first deciding which of several similar names to canonicalize as the "best" one.

bhousel:
ok I'm going to merge this, update the docs, and start disambiguating some brands we've found related to #1880

PR #2173, opened 2018-10-28
 Research needed for: amenity/fuel|Posto Ipiranga #539

Tubaleviao:
Solution for issue #539


ENT8R:
Thank you 👍 But please remove the underscores (`_`) next time 😄

PR #2176, opened 2018-10-29
Research needed for: amenity/fuel|Posto Shell #540

Tubaleviao:
Solution for issue #540


bhousel:
Thanks @Tubaleviao  - it looks like you found the right wiki tags for Shell, but I looked into this a bit more, is "Posto" just the Portuguese word for "station"?  

If that's the case, we should probably instead add these tags that you found to the "amenity/fuel|Shell" entry, and also add a "match" there for "amenity/fuel|Posto Shell"..  Do you agree?

Tubaleviao:
Yes, "posto" is the Portuguese word for "station". I don't know how the application is organized, but if we're looking for a "Shell" fuel station, usually we type "Posto" before "Shell".

PR #2177, opened 2018-10-29
Currys PC World research

aestevens:
Re: #1430, Currys PC World country codes and wiki tags have been added, and for good measure country codes were added to the Currys brand based on my research. While Currys PC World and Currys are owned by the same parent company, both warrant inclusion in the index since some locations uniquely match each brand when querying Overpass Turbo.


bhousel:
Thanks @aestevens !

PR #2179, opened 2018-10-29
Wiki tags for shop/furniture|Bodzio (fixes: #1469)

wmalinowski:


bhousel:
Thanks @wmalinowski 👍

PR #2180, opened 2018-10-29
Improved the filter for barber shops (fixes: #1506)

wmalinowski:


bhousel:
Thanks @wmalinowski 👍 Definitely good to get rid of this...

These probably proliferated because the old version of this index included it.

screenshot 2018-10-29 18 45 47

PR #2192, opened 2018-11-02
add bakery brand for 뚜레쥬르

depth221:


ENT8R:
https://en.wikipedia.org/wiki/Tous_Les_Jours says that this bakery operates also in other countries than South Korea. Can you confirm this?

bhousel:
Hey @ENT8R - I think the build script probably added ‘kr’ automatically because the name contains Korean characters.. The assumption here is that even though the brand may operate in other countries, it would be styled differently there , using a language that can be read by people there.  We can check if this index contains entries in another language for this bakery - it’s ok for there to be multiple.

ENT8R:
Alright! Thanks for clarification 👍

depth221:
@bhousel Sorry, I added the 'kr' index in person, because didn't know that this bakery range from Korea to other countries.

bhousel:
> @bhousel Sorry, I added the 'kr' index in person, because didn't know that this bakery range from Korea to other countries.

No worries, @depth221 - it's probably the right thing to do.  In other countries the name probably does not look like "뚜레쥬르"

PR #2087, opened 2018-10-21
fix "hu" (shop/convenience|Abc) - general hungarian word - remove (need extra check in Poland!)

ImreSamu:

PR #2193, opened 2018-11-03
attempt to handle convenience shops named abc [ready, waiting for review]

matkoniecz:
closes #2087

--

this one is a bit tricky so I am submitting it as a PR

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -20289,10 +20289,13 @@
   },
   "shop/convenience|ABC": {
     "count": 798,
+    "countryCodes": ["us"],
     "match": ["shop/convenience|Abc"],
     "nomatch": ["amenity/fuel|ABC"],
     "tags": {
       "brand": "ABC",
+      "brand:wikidata": "Q4650251",
+      "brand:wikipedia": "en:ABC Stores (Hawaii)",
       "name": "ABC",
       "shop": "convenience"
     }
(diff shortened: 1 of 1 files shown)


good-praxis:
There is also a lot of them in hungary, apparently? https://overpass-turbo.eu/s/Drg
Is this the same brand as the one in poland or is this yet another one?

good-praxis:
How does this reflect the generic hungarian ABC? It will basically ignore it because of country codes, right?

ImreSamu:
> How does this reflect the generic hungarian ABC? 

 it looks fine to me.  :)

good-praxis:
I will go ahead and merge it then, because it seems factually accurate -- however, please keep in mind that a lot of the polish `abc` are incorrectly named `ABC`, as you can see on [overpass](https://overpass-turbo.eu/s/Dri)

ImreSamu:
just a comment:

in extreme cases 
-  some native Hungarian osm contributor in Poland,  adding this - instead of a normal `shop=convenience`  -  **so a brand logo would be helpful.**   

but this is a really extremes cases.

now the hungarian "terms"  - including the "ABC" 
*  "terms": "ABC, sarki élelmiszer, kisbolt, közért"

see https://github.com/openstreetmap/iD/blob/master/dist/locales/hu.json#L5400
```json
                "shop/convenience": {
                    "name": "Kis élelmiszerbolt",
                    "terms": "ABC, sarki élelmiszer, kisbolt, közért"
                },
```

Probaly somebody need to write a simple script - to check the "brand" names in the Terms,

PR #2212, opened 2018-12-18
Improve data on United Petroleum

davidhicks:
Adds alternative match.
Adds brand Wikidata item.
Adds brand Wikipedia item.


maxerickson:
Thanks!

Try to remember to run `npm run build` before committing, to cut down on formatting changes and the like.

PR #2213, opened 2018-12-18
Add data for various brands

kreed:


bhousel:
Thank you @kreed , looks good 👍

PR #2232, opened 2018-12-31
add australian brands

andrewharvey:


andrewharvey:
Sorry I tried to follow https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md but I found it a bit confusing on how to add new entries, so if it needs changes, please let me know so I can fix that up here, and in the CONTRIBUTING guide.

SilentSpike:
I also found the contributing guidelines pretty confusing when first adding a new entry and ended up looking at `schema/canonical` to suss some things out.

The only thing you're likely missing is the `"nocount"` key. Did `npm run build` have a message about "uncommon name" at the bottom of its output? The guidelines mention it in context of changing a commonly used tag combination to a more accurate one, but I *believe* it also applies to new entries since they don't appear often enough (yet) to be counted in the process of updating `allNames.json`

This addition otherwise looks good to me 👍

Adamant36:
@SilentSpike, I don't know about nocount. Did you look into the usage on taginfo? There's 83 usage of kathmandu by itself and like 17 pages of names that contain kathmandu in them. So it might be worth doing more research on the actual usage numbers. Or is it something auto generated by the script once the entry is created? 

Also, it might be worth adding the country codes. Since according to Wikipedia the company is confined to New Zealand, Australia and the United Kingdom.

SilentSpike:
@Adamant36 See this query: http://overpass-turbo.eu/s/ES5

Only 44 instances of this tag combination currently. Need 50 to be picked up in `allNames.json`. So `nocount` is necessary to suppress the warning because we know this data is correct (basically all new entries will need this tag).

As I understand it, the `count` tag is sourced by the code which updates `allNames.json` and added by `npm run build`. I haven't looked into the code yet, but it should probably also check for entries with `nocount` to see if they have 50+ yet and remove the tag once applicable (if it isn't already doing so).

Adamant36:
Oh cool. Thanks for the information. It makes sense there would be a certain threshold for it to be included.

config/canonical.json
```diff
@@ -23620,6 +23620,16 @@
       "shop": "department_store"
     }
   },
+  "shop/department_store|Kmart": {
```
bhousel: Oh another one of these!.. I'll have to disambiguate between the two KMarts.

bhousel:
Looks good, thanks @andrewharvey ! 
I'll sort out KMart in a following commit.

PR #2234, opened 2018-12-31
United Dairy Farmers

1ec5:
A United Dairy Farmers store isn’t just a convenience store: it’s also an ice cream parlor, and neither is subordinate to the other in terms of prominence. A UDF store location is always mapped as a single POI. At a typical location, the ice cream parlor isn’t well separated from the rest of the convenience store: you order ice cream and pay for groceries at the same counter, though seating is in a separate room at some locations.

Added an entry for UDF gas stations, which are mapped as separate POIs. Both entries now have a `short_name` tag and more search terms.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -10994,6 +10994,25 @@
       "name": "Turmöl"
     }
   },
+  "amenity/fuel|UDF Fuel": {
+    "match": [
+      "amenity/fuel|UDF",
+      "amenity/fuel|UDF Fuels",
+      "amenity/fuel|United Dairy Farmers"
+    ],
+    "nocount": true,
+    "nomatch": [
+      "shop/convenience|United Dairy Farmers"
+    ],
+    "tags": {
+      "amenity": "fuel",
+      "brand": "United Dairy Farmers",
+      "brand:wikidata": "Q7887677",
+      "brand:wikipedia": "en:United Dairy Farmers",
+      "name": "UDF Fuel",
+      "short_name": "UDF"
+    }
+  },
   "amenity/fuel|Ultramar": {
     "count": 631,
     "nomatch": ["shop/convenience|Ultramar"],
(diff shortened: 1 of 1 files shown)


Adamant36:
What's with the short name thing? Seems like its the only time its used in the database. I'm wondering what the advantage of that over just going with a match exception is. There seems to be an infinite number of alternative name tags to the main one. Maybe some shouldn't be used (I'm not saying short_name shouldn't be. Its just something I've been thinking about and this gives me an opportunity to bring it up).

1ec5:
I think this happens to be a strong case for `short_name`. Typically, “United Dairy Farmers” wordmarks appear on the interior and exterior of the convenience store, while “UDF” wordmarks appear on the gas station canopy and wrapped around their ice cream cones. People are likely to search Nominatim for either name.

Currently, [69 features](http://overpass-turbo.eu/s/EYN) are tagged with both `name` and `short_name`, compared to [48](http://overpass-turbo.eu/s/EYO) tagged with only `name=United Dairy Farmers` and [10](http://overpass-turbo.eu/s/EYP) tagged with only `name=UDF`.

Adamant36:
Hhhhmmm, OK. Makes sense. At some point we should collect the possible name alternatives, along with their pros and cons, and put them in the contributing document or something. It seems like there's a lot of tags we could go with. So it would be nice to have a guide out there somewhere for new contributors to look at if need be. 

Is there somewhere that lists what specific tags Nominatim supports? I always assumed it was just name=* and addresses (or tags related to things like admin boundaries).

bhousel:
> This project’s taginfo project file was added after I opened this PR, so I’ve added `short_name` to the file in [d441d2d](https://github.com/osmlab/name-suggestion-index/commit/d441d2dc24a82b97e7b7bd4b1e78f90c7f40117d).

Thanks!  FYI `taginfo.json` will also regenerate if you do `npm run dist`

PR #2222, opened 2018-12-28
Added more banks

Adamant36:
closes #230, closes #231, closes #236, closes #245, closes #252, Closes #253, closes #263, closes #264, closes #265, closes #266, closes #267, closes #269, closes #270, closes #273, closes #276, closes #278, closes #287, closes #289, closes #292, closes #294, closes #296 

Btw, I noticed that Metrobank and Metro Bank are in the same entry. They seem to be different banks though. Metro Bank is in the United Kingdom. Whereas Metrobank resides in the Philippians. So they might need to be split. Also, there's a separate entry for PNB (issue #295), but its more than likely Punjab National Bank. Which already has an entry, including Wikipedia data.

PR #2237, opened 2019-01-01
Add contributing guidelines for new brands

SilentSpike:
As mentioned in #2232 the contributing guidelines aren't super clear when it comes to new entries.

My writing skills aren't the best, but I thought I'd take a crack at clearing things up. I think it's pretty succinct, but wasn't sure whether we're okay with adding new brands without `wikidata`/`wikipedia` information.

Let me know if anything should be changed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -454,10 +454,9 @@ _(comments added for clarity)_
 If it exists, we want to know about it!
 
 Some brands aren't mapped enough (50+ times) to automatically be added to the index so this
-is a valuable way to get ahead of inconsistent tagging.
+is a valuable way to get ahead of incorrect tagging.
 
-
-1. Before adding a new brand, the minimum information you should know is the correct tagging required for instances of the brand (`name`, `brand` and what it is - e.g. `shop=food`). Ideally you also have `wikidata` and `wikipedia` tags for the brand and any other appropriate tags - e.g. `cuisine`.
+1. Before adding a new brand, the minimum information you should know is the correct tagging required for instances of the brand (`name`, `brand` and what it is - e.g. `shop=food`). Ideally you also have `brand:wikidata` and `brand:wikipedia` tags for the brand and any other appropriate tags - e.g. `cuisine`.
 
 2. Add your new entry anywhere in `config/canonical.json` (it will be sorted alphabetically later) and using the `"tags"` key add all appropriate OSM tags. Refer to [here](#card_file_box--about-configcanonicaljson) if you're not familiar with the syntax.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,6 +8,7 @@
 * Resolve warnings - [show me](#thinking--resolve-warnings)
 * Remove generic names - [show me](#hocho--remove-generic-names)
 * Add `brand:wikidata` and `brand:wikipedia` tags - [show me](#female_detective--add-wiki-tags)
+* Add missing brands - [show me](#convenience_store--add-missing-brands)
 
 ##### :no_entry_sign: &nbsp; Don't edit the files in `dist/` - they are generated:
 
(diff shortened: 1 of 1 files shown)


maxerickson:
I don't see a problem adding entries without wikidata/wikipedia tags, but it's probably a good idea to note that the version of the index published for the editors only includes entries that have a wikidata tag.

Adamant36:
Wikidata or Wikipedia tags. It doesn't need both I don't think. There's a few entries that contain only the Wikidata tag and its the more important of the two if I remember correctly. Since Wikipedia pages come and go, but wikidata entries last forever or something (see https://github.com/osmlab/name-suggestion-index/pull/2222#issuecomment-450438029 for reference).

andrewharvey:
Thank you @SilentSpike this makes it much nicer as a beginner to get in and contribute new brand presets.

maxerickson:
There shouldn't be any wikipedia articles that don't have matching wikidata items. The output is only conditioned on wikidata: https://github.com/osmlab/name-suggestion-index/blob/master/build_nameSuggestions.js#L30

CONTRIBUTING.md
```diff
@@ -448,6 +449,27 @@ _(comments added for clarity)_
...
 &nbsp;
 
+### :convenience_store: &nbsp; Add missing brands
+
+If it exists, we want to know about it!
+
+Some brands aren't mapped enough (50+ times) to automatically be added to the index so this
+is a valuable way to get ahead of incorrect tagging.
+
+1. Before adding a new brand, the minimum information you should know is the correct tagging required for instances of the brand (`name`, `brand` and what it is - e.g. `shop=food`). Ideally you also have `brand:wikidata` and `brand:wikipedia` tags for the brand and any other appropriate tags - e.g. `cuisine`.
```
1ec5: If there isn’t a Wikidata item for the brand yet, we could suggest [creating one](https://www.wikidata.org/wiki/Special:NewItem) and maybe recommend some common values to set the [instance of](https://www.wikidata.org/wiki/Property:P31) property to. Increasingly people will want to add entries for brands that don’t warrant Wikipedia articles, but they can still have Wikidata items.

SilentSpike:
@maxerickson I think the `brand:wikidata` tag should just be listed as required then since there's little point in adding entries without it (plus PRs can be checked using it). If people agree I'll update that text to say so.

Adamant36:
I don't think including Wikidata/Wikipedia information should be requirement for a new entry. As stated above, the index sorts the ones without them out as it is. Plus, its useful to have entries with extra tags like cuisine or region even without the Wikipedia data. As it saves time and research later when there is an article for it.  

Not to mention, it gives a good indicator of which brands need Wikidata/Wikipedia information in the first place. Maybe someone will create the pages if they don't exist by looking through the index, but how will they know otherwise if no entry is there to start with? 

I could also see the process being automated at some point, where the brands without Wikidata yet are periodically checked against the website to see if they have been added. By not including the entries just because they don't have that information though something like that, or someone deciding to create a wikidata page based on the fact that the brand doesn't have one, wouldn't be possible.

1ec5:
> As it saves time and research later when there is an article for it.

I figure that it takes seconds to create a Wikidata item for the brand being added to the index – this document can eventually provide some pointers – but it’s more tedious and error-prone to find and tag existing features that are missing Wikidata tags because an item wasn’t created upfront. But this PR is already an improvement, so perhaps https://github.com/osmlab/name-suggestion-index/pull/2237#discussion_r244837333 and a requirement to add a Wikidata item can wait for a separate PR.

bhousel:
> [Here’s a gist](https://gist.github.com/1ec5/8d44e94db9ac23c0b859c2718d7f82c3) sketching out what it looks like to create a Wikidata item for a brand. But this PR is already an improvement, so perhaps [#2237 (comment)](https://github.com/osmlab/name-suggestion-index/pull/2237#discussion_r244837333) and a requirement to add a Wikidata item can wait for a separate PR.

Cool -  I agree with @1ec5 that the revised guidelines are an improvement.. Thanks @SilentSpike for putting this together!

As mentioned elsewhere, yes only entries with a `brand:wikidata` tag will actually make it into the editor presets.  But even having it in the `canonical.json` file doesn't hurt, and we're starting to attract enough contributors that people are comfortable adding Wikidata entries for those brands that need them.

PR #2238, opened 2019-01-01
add some wikidata/wikipedia + country codes

Sukkoria:
Impacted brands are the following, some of them already had wikidata and/or wikipedia tags, but no country codes : 
- Braum's
- Sibylla (already closed, #440)
- Суши Wok
- Agrola #451 
- Chevron 
- Citgo


Adamant36:
You might want to run "npm run build" before you commit the file so that it doesn't give the error at the end of the file change. Plus it will double check if anything else is wrong before you do the PR. I made the same mistake myself a few PRs back.

config/canonical.json
```diff
@@ -8335,13 +8337,16 @@
   },
   "amenity/fast_food|Суши Wok": {
     "count": 55,
+    "countryCodes": ["ru", "uk"],
```
bhousel: ukranian country code is `ua` not `uk`

bhousel:
Ah this is getting old, I'll just merge and fix it

PR #2250, opened 2019-01-08
Taco Bell is Tex-Mex

1ec5:
We currently have Taco Bell as `cuisine=mexican`, based on common usage, whereas [the wiki recommends](https://wiki.openstreetmap.org/wiki/Key:cuisine#Cuisine_.28ethnicity.29) using the more obscure `tex-mex` instead – and I’d tend to agree. Taco Bell doesn’t ever include “Mexican” in the brand name, as Chipotle and Qdoba sometimes do.

https://github.com/osmlab/name-suggestion-index/blob/418be28cd2323ec8383cec2aaab5d17bc264a3d9/config/canonical.json#L8132-L8143

[](https://commons.wikimedia.org/wiki/File:Naked_Burrito_Bowl_at_Taco_Bell_in_Iso_Omena.jpg)
_Not especially Mexican…_

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -8137,7 +8137,7 @@
       "brand": "Taco Bell",
       "brand:wikidata": "Q752941",
       "brand:wikipedia": "en:Taco Bell",
-      "cuisine": "mexican",
+      "cuisine": "tex-mex",
       "name": "Taco Bell"
     }
   },


1ec5:
I also started [a discussion on the wiki](https://wiki.openstreetmap.org/wiki/Talk:Key:cuisine#Separate_cuisine_from_food_type) about potentially restructuring the `cuisine` tag.

Adamant36:
Do you have a good personal idea of what exactly "Tex-Mex" cuisine entitles? I looked for a definition online, but it seem like there is a solid one (something about using tortillas and cheese more then normal Mexican food does). Personally, id consider all the companies you mentioned as Tex-Mex, but some other companies seem a little more ambiguous. So I'm interested to know. My guess would be that most American Mexican food chains would qualify as Tex-Mex.

1ec5:
Yeah, it’s really tough. Some chains, like the Texas-based Taco Bueno, are proudly Tex-Mex, and it was pretty easy for me to be snobby and classify Taco Bell as Tex-Mex since they don’t currently brand themselves as a Mexican restaurant per se. But Chipotle and Qdoba are both “Mexican Grills”, even though it’d be more accurate to say that they’re Mexican-inspired. El Pollo Loco is also interesting, being from Mexico but carrying a more Americanized menu in the U.S. Though I guess any chain localizes its products for each country, including McDonald’s and KFC.

todrobbins:
I love that this is an actual GitHub issue. 🥰

PR #2251, opened 2019-01-08
Fix entries

Adamant36:
Fixes a few entries that were wrong.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -14263,7 +14263,7 @@
       "brand": "Cheesecake Factory",
       "brand:wikidata": "Q1045842",
       "brand:wikipedia": "en:The Cheesecake Factory",
-      "cuisine": "dessert",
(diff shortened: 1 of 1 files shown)


bhousel:
LGTM.. Merge it if you like it 👍

Adamant36:
Btw, I noticed that both UK and GB are used for the United Kingdom's country code in the index. Do you have a preference for which to use?

bhousel:
> Btw, I noticed that both UK and GB are used for the United Kingdom's country code in the index. Do you have a preference for which to use?

I think I fixed them all in  63f8530

`uk` should not exist in the index.. The 2 letter country codes can be found [here](https://www.nationsonline.org/oneworld/country_code_list.htm).

I switched them to either `gb` for United Kingdom or `ua` for Ukraine.

Issue #2236, opened 2019-01-01
Should we favour existing tagging for names?


SilentSpike:
Have encountered a few instances of this already so it would be good to have a rule in writing to go by.

Sometimes existing tagging is a `name=` value that's close, but not exactly the same as the name presented by the brand themselves. Examples:
- #2233 (second discussion)
- #2235 (CVS Pharmacy)
- #1413

Personally I think we should be updating/adding entries with the name presented by the brand even if it goes against existing tagging and even if the difference is very minor (as the goal here is to provide "canonical" information).

Any thoughts, suggestions, exceptions, edge cases?

Adamant36:
Which name is connected to the actual tagging numbers? Otherwise, you lose all the ones that are tagged already by changing the name to something not used and iD Editor is creating a new entry (mapping point) in the data base that diverges from the original. Essentially leaving the old one behind. At least thats how I think it would work. 

Also, how would we decide if the name is "wrong" exactly since we dont have ground truth? For example if you look at pictures of CVS stores on Google, in front some say "CVS Pharmacy" while some just say "CVS." So while neither is probably wrong, id say its better to go with the existing value because its established and would therefore have more weight on the ground.

SilentSpike:
@Adamant36 That's a good point and something else I've been curious about. There's a weird issue when it comes to tagging chains in that the signage doesn't always match the name. The CVS example maybe isn't the best as I've now learned that it's also a store of some kind.

However, one I've come across is Sports Direct. All of the stores (at least here in the UK) have a big sign that says "SportsDirect.com". However, the actual store name is just "Sports Direct" which is what you'd call it in conversation (and what the entry here has it tagged as).

Another example here in Scotland is Clydesdale Bank, often branches just have a sign with "CB" on it, but surely you'd tag it with the full name still because that's what it's actually called (at least, I would).

----
As for the majority tagging question, as I see it this is a source of cannonical information and thus should show preference to correct information over existing information. We may not know ground truth, but when it comes to brands tagging should (for the most part) be consistent everywhere.

To your point about losing the existing tagging, I believe that's where the `match` key comes in. You're still recognising those entities as the same thing, just not the correct tagging.

Adamant36:
@bhousel, what do you think about the whole thing?

bhousel:
Oh yeah this is something that will probably always be full of interesting edge cases and judgement calls...

@SilentSpike said:
[...]

PR #2263, opened 2019-01-21
More brands

kreed:
Few more brands I ran into. I got a little proactive in renaming Schlotzsky's, but per the discussion in #2236 I think that's the direction we're going with?


matthijsmelissen:
Thanks! Could you specify in the PR description which issues this fixes, like here: https://github.com/osmlab/name-suggestion-index/pull/2255 ?

SilentSpike:
All look good to me (the rename of Schlotzsky's included) 👍

`countryCodes` data is always nice too (as these appear to be mostly US brands), but not as important

kreed:
That's a good point. I'll add those in.

kreed:
@Adamant36 I made those changes for Safelite and Extra Space Storage. I'm not sure about La-Z-Boy. It looks like they operate outside North America (e.g. their [UK division](https://www.la-z-boy.co.uk/)), but I'm not sure they have stores elsewhere. Even if they don't now, it seems likely that they could in the future. I'm going omit the country code to be safe.

Adamant36:
That's fine. I wasn't to sure about it anyway. Thanks for adding the other stuff.

matthijsmelissen:
These seem to be smaller brands. For example Havertys only has 19 instances in OSM.

@bhousel Would we like to add these?

SilentSpike:
@matthijsmelissen We encourage adding brands which don't have enough entries (50) to be automatically picked up yet.

However, you raise a related interesting question, should we include brands which have less than 50 locations in reality?

PR #2278, opened 2019-02-11
Add additional matches for Men's Wearhouse

tas50:
There were 9 places in OSM where it was spelled Men's Warehouse. I fixed those, but this should help prevent that in the future.


bhousel:
Thanks @tas50 👍

PR #2283, opened 2019-02-12
Add wikipedia data and resolve build warnings

tas50:
Fixup some warnings and add some wikipedia entries.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -19928,8 +19928,11 @@
   },
   "shop/clothes|Express": {
     "count": 96,
+    "nomatch": ["shop/convenience|Express"],
     "tags": {
       "brand": "Express",
+      "brand:wikidata": "Q1384784",
+      "brand:wikipedia": "en:Express, Inc.",
       "name": "Express",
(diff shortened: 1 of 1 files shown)


Adamant36:
Hi there. Just a heads up, if you do npm run build when your done before you submit the PR it will check for errors in your code and get rid of the red arrow (trailing space or whatever) at the end of your code. Its not 100% necessary, but still good to do just in case.

tas50:
I'm pretty sure that's npm run build actually nuking a new line that someone else added in.

Adamant36:
> I'm pretty sure that's npm run build actually nuking a new line that someone else added in.

Hhhmmm, I'm not sure that's how it works. A new line is created every time you save the canonical.json and its not from something before you or it wouldn't throw up the error I don't think. https://stackoverflow.com/questions/34271528/what-is-little-red-arrow-and-no-sign-in-github-desktop

bhousel:
This seems ok now.. Thanks @tas50 !   
I'm not too worried about the trailing newline 😄

Issue #2275, opened 2019-02-11
Add entry for: Cinnabon


rivermont:
[Cinnabon Website](https://www.cinnabon.com)
[Wikipedia](https://en.wikipedia.org/wiki/Cinnabon)
`brand=Cinnabon`  
`amenity=fast_food`  
`cuisine=cinnamon_roll`

PR #2295, opened 2019-02-13
Entries for Cinnabon and Old Time Pottery

mparrault:
Added entry for Cinnabon according to [#2275]( https://github.com/osmlab/name-suggestion-index/issues/2275) and Old Time Pottery according to [#2219](https://github.com/osmlab/name-suggestion-index/issues/2219).

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -1,13 +1,4 @@
 {
-  "shop/confectionery|Hemmakvall": {
-    "nocount": true,
-    "tags": {
-      "amenity": "confectionery",
-      "brand:wikipedia": "Hemmakväll",
-      "brand:wikidata": "Q10521791",
-      "name": "Hemmakväll"
-    }
-  },
   "amenity/bank|ABANCA": {
     "count": 102,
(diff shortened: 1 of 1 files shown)


Adamant36:
Hi there. It doesn't seem as though cinnamon_roll is an official cuisine tag or even has any uses. So it might be worth changing to something more accepted and in use. 

Also, if you put "closes"  before the issue number it will automatically close the issue the PR is attached to when its merged. Then someone won't have to go through and find it later to close it. 

Thanks

mparrault:
I've changed the cuisine type to cuisine=dessert.  Thanks for the tip on closing.  Closes #2275 and closes #2219.

bhousel:
Thanks @mparrault 👍

Adamant36:
@bhousel, since this was merged is brand:website a tag that can be added to entries now?

bhousel:
I think `brand:website` is not really important, but if people want to add it I don’t care. 

It’s probably better to not add a bunch of extra tags that hold data that can be fetched from the Wikidata record.

PR #2303, opened 2019-02-13
Add the Vitamin Shoppe

tas50:
Seems to exist in just about every mall in the US.


bhousel:
Thanks @tas50 👍

PR #2312, opened 2019-02-15
Disambiguate Home Hardware

mparrault:
As documented [here](https://www.homehardware.ca/about-home-hardware), there are three different types of Home Hardware stores (not counting Home Furniture), one of which is a shop=hardware, one of which is a shop=doityourself, and one is both.  I'm adding the first one as shop=hardware, and the later two as shop=doityourself.  Similar to Walmart (shop=department_store), Walmart Neighborhood Market (shop=supermarket), and Walmart Supercentre (both).

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -1,13 +1,4 @@
 {
-  "shop/confectionery|Hemmakvall": {
-    "nocount": true,
-    "tags": {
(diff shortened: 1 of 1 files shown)


tas50:
Thanks for breaking it down in the PR description. That's super helpful.

PR #2316, opened 2019-02-16
Add new banks and supermarkets now present in OSM

tas50:
After updating the allnames file from a new planet file a number of new banks and supermarkets popped up. I'm going to begin slowly adding them as I do research and add Wikipedia / Wikidata entries. I also added new matches that will prevent duplicate entries with that new allnames data.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -7336,7 +7336,8 @@
     "match": [
       "amenity/fast_food|DQ",
       "amenity/ice_cream|Dairy Queen",
-      "amenity/restaurant|Dairy Queen"
+      "amenity/restaurant|Dairy Queen",
+      "amenity/fast_food|DQ Grill & Chill"
     ],
     "tags": {
       "amenity": "fast_food",
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -3736,8 +3736,18 @@
       "name": "Seylan Bank"
     }
   },
+  "amenity/bank|Siam Commercial Bank": {
+    "count": 53,
+    "tags": {
+      "amenity": "bank",
+      "brand": "Siam Commercial Bank",
+      "brand:wikidata": "Q2038986",
+      "brand:wikipedia": "en:Siam Commercial Bank",
+      "name": "Siam Commercial Bank"
+    }
+  },
   "amenity/bank|Sicoob": {
-    "count": 82,
+    "count": 96,
     "tags": {
       "amenity": "bank",
       "brand": "Sicoob",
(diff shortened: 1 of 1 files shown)


Adamant36:
I noticed you combined Bank of the Philippine Islands and BPI Family Savings Bank. Their actually two seperate banks. So you might want to split them.

 Also, DQ Grill & Chill are different stores then Dairy Queen. They have different signs out front, different sized stores, and serve different food. So they should probably be split also.

In general you might want to be a little more conservitive on combining brands into the match thing. A lot of times companys will have stores under similar names but they will have different layouts/products etc and they are worth keeping seperate. For instance see the Walmart entry. If two store brands are owned by the same company you can always use the operator tag to show they are connected like is done with Walmart. Its better then combining entries that aren't actually matches. Thanks.

tas50:
I can back out the DQ Grill & Chill change. I was going off my first hand experience here in the Pacific Northwest. Despite the clear distinction that the Wikipedia article claims exists there's several of them here that are basically a tagline change on the sign and not much more. Apparently in other areas that name means more so I'll add a distinct entry with Dairy Queen as the operator.

As for BPI I don't think my addition is the incorrect bit in that entry. Bank of the Philippines Islands is the full name for BPI. Under BPI exists many subsidiaries one of which is the existing match in that entry of BPI Family Savings Bank. BPI Family Savings Bank shouldn't be a match since that is a different brand, but BPI and Bank of the Philippines Islands are the same brand. I went off the wikipedia entry for that one as well as their homepage which is branded throughout as BPI and titled "Welcome to the Bank of the Philippine Islands | BPI".

If you'd like I'd be more than willing to tag you on these PRs for review in the future. I'd much rather have someone double checking and questioning changes than me self merging.

Adamant36:
> I can back out the DQ Grill & Chill change

Thanks

> Despite the clear distinction that the Wikipedia article claims exists there's several of them here that are basically a tagline change on the sign and not much more.

There was a similar discussion with CVS versus CVS pharmacy. Which might be same stores, but maybe not. We decided it was best to go with the established name tag and what the sign outside the store says. Since that's how mappers usually tag the names of things. Plus, the store sign is ground truth and Bing Streetside verifiable. You don't potentially lose any properly tagged stores either like you would if you combine them. 

> If you'd like I'd be more than willing to tag you on these PRs for review in the future. I'd much rather have someone double checking and questioning changes than me self merging.

It would be cool if you tag me on PRs.  I'm totally willing to double check things for you.  Feel free to review my PRs also if you want. I usually wait a while before merging them just for that reason. If your cool with it, I'll start tagging you in cases where its a big addition or I'm not sure about something.

PR #2318, opened 2019-02-17
Add Freshii restaurant

tas50:
This is commonly found on OSM as either a restaurant or fast_food. Based
on the wikipedia page it's actually fast casual dining which we map to
restaurant based on other fast casual restaurants.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -14846,6 +14846,18 @@
       "name": "Frankie & Benny's"
     }
   },
+  "amenity/restaurant|Freshii": {
+    "match": ["amenity/fast_food|Freshii"],
+    "nocount": true,
+    "tags": {
+      "amenity": "fast_food",
+      "brand": "Freshii",
+      "brand:wikidata": "Q5503051",
+      "brand:wikipedia": "en:Freshii",
+      "cuisine": "salad",
+      "name": "Freshii"
+    }
+  },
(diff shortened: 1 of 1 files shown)


Adamant36:
I might not be exactly clear on how matching works, but wouldn't `"match": ["amenity/fast_food|Freshii"],` not be necessary if the entry is already tagged as `"amenity": "fast_food",` + `"name": "Freshii",` because it would essentially be matching against itself in that case? I thought matching was for tags that were slightly different in some way.

tas50:
It's restaurant vs. fast_food here.

config/canonical.json
```diff
@@ -14846,6 +14846,18 @@
       "name": "Frankie & Benny's"
     }
   },
+  "amenity/restaurant|Freshii": {
+    "match": ["amenity/fast_food|Freshii"],
+    "nocount": true,
+    "tags": {
+      "amenity": "fast_food",
```
Adamant36: So then this line should be `"amenity": "restaurant",` to match with line 14849 correct? Since your matching with fast_food but the actual tag being used is amenity=restaurant.

tas50: Oh yep that is a mistake. I didn't cut that over to restaurant. 🤦‍♂️

Adamant36: Ah well, that explains it then. I knew it looked off. I couldn't figure out how.

PR #2320, opened 2019-02-17
Add match for Express Oil Change & Tire Engineers

tas50:
We use shop/tyres, but this same store has popped up with shop/tires.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -34794,6 +34794,9 @@
   },
   "shop/tyres|Express Oil Change & Tire Engineers": {
     "count": 100,
+    "match": [
+      "shop/tires|Express Oil Change & Tire Engineers"
+    ],
     "tags": {
       "brand": "Express Oil Change & Tire Engineers",
       "brand:wikidata": "Q39057654",


Adamant36:
How many uses of it where tagged that way? It might be worth just re-tagging them instead since its bad tagging and not official. I don't think the database would work if there was a match exception for every odd spelling of a tag.

tas50:
There were 50+ incorrectly tagged since it was added to the list when I rebuild things from a new planet file.

Adamant36:
Hhmmm, that's not that many and I don't think we should open the flood gates to supporting none standard tags as matches like that. Since there's literally an endless amount of them and they should really be re-tagged properly instead of integrated here. So I suggest we close this one without merging it.

tas50:
I think this is exactly the usage that the match field is meant for. It dedupes all the incorrect tags so they don't get their own entries in the JSON file once 50 or more incorrect tags show up. Most of the matches are exactly this. Total bogus things people have tagged as. This one popped up when I pulled in the latest planet file and without this match running the build will result in the incorrect tag being added to the file, which we really don't want.

Adamant36:
I was under the impression that match was for when you type something that misspelled into iD editors search pain that it will populate the tags with the information from the correctly spelled name tag and its sub-tags. Like if you search for Wallmart instead of Walmart. If people search for the same exact business name though, then the proper use of tyres will still come up, not tires. 

More then likely tires is being used by another mapping tool as a preset or people are just changing after the fact. Thinking tyres is a miss spelling on iD's side. Most people outside of Europe don't know OSM prefers European names. So this wouldn't help. As its meant to encourage proper business information tagging. Not proper tagging in general (although it does help with that secondarily. Its mainly done through presets in iD editor). The only thing that might work is not allowing un established sub tags in the amenity field. Which currently isn't implemented. Match doesn't do that though. 

As a test I just put "tire" into the iD editor search and it automatically uses shop=tyres. Also if you put "Express Oil Change & Tire Engineers" into the search field it also automatically uses shop=tyres. So its not anything on iD Editors or the name-index's side. It must be what I said above. I can still manually change the tag tires though. It might be worth opening an issue on iD Editors Github page to if they can block bad entries to the sub-field. I'll do some research when I have the time also to see what application the people that are doing the miss-tagging are using. It could have been through a bad import also. All the uses of shop=tires are from the same area in eastern America. So its probably an import or a single user that doesn't know any better.

bhousel:
Yes the “match” field is for catching the undesired tag-name combinations.  

It’s ok to match `shop/tires|Name` if it is commonly mistagged that way.

Adamant36:
@tas50, I stand corrected. I'll have to keep an eye out for similar things now.

PR #2321, opened 2019-02-17
Filter out Пиццерия (pizzeria)

tas50:
This is a generic term in Russian and not a particular chain.

I'd love a confirmation on this from a Russian speaker if possible.


Adamant36:
Id go with the fact that its generic. There's a ton of generically named stores in Russia. Especially by the border with Ukraine. Checkout all the Russian generic convenience store names for instance. I think it has something to do with being a Communist country and generally low on corporate entities. Which leads to a bunch of mom and pop places that name the stores after whatever the main product they are selling is. 

If you want to go through the issues related to generic terms and deal with them that would be great. Id love to see them closed, but I'm mainly into adding Wikipedia entries at the moment.

PR #2324, opened 2019-02-19
Add Hyatt hotels

tas50:
Adds the various Hyatt brand hotels with the operator of Hyatt. The Wikipedia page covers each of these brands.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -35422,6 +35422,17 @@
       "tourism": "hotel"
     }
   },
+  "tourism/hotel|Grand Hyatt": {
+    "nocount": true,
+    "tags": {
+      "brand": "Grand Hyatt",
+      "brand:wikidata": "Q1425063",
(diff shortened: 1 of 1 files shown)


Adamant36:
Looks good I think. I doubt every hotel off shot will get its own Wikipedia page in this case. Those things are way to regional and probably not note worthy by themselves.

Adamant36:
Mind if I ask why you didn't just go with a "match" for all of them under the Hyatt entry instead though?

tas50:
They are very different locations that should get named with the correct name vs. slapping Hyatt on everything which would be incorrect. Some of those grands are very upscale hotels where as some are kinda dumpy. If you searched for Grand Hyatt you would want only those.

Adamant36:
Like you say, the hotel industry is pretty convoluted. So its better to have them then not, I think.

PR #2325, opened 2019-02-19
Add Wyndham as the operator of various motels / hotels

tas50:
These are all brands run by Wyndham. They own a lot of stuff.


Adamant36:
> They own a lot of stuff.

Yes they do. It will be interesting to see who owns what once more operators are added.

PR #2327, opened 2019-02-19
Add the Renaissance Hotels brand

tas50:
I left out the name here since they are all uniquely named based on the
city they're in, but this will be useful to get the proper brand /
operator tags in place.

@bhousel Will leaving out the name tag here preserve the unique name while applying all the other tags?

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -35660,6 +35660,18 @@
       "tourism": "hotel"
     }
   },
+  "tourism/hotel|Renaissance Hotel": {
+    "nocount": true,
+    "tags": {
+      "brand": "Renaissance Hotels",
+      "brand:wikidata": "Q2143252",
+      "brand:wikipedia": "en:Renaissance Hotels",
+      "operator": "Marriott International",
+      "operator:wikidata": "Q1141173",
+      "operator:wikipedia": "en:Marriott International",
+      "tourism": "hotel"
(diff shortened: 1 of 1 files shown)


bhousel:
> @bhousel Will leaving out the name tag here preserve the unique name while applying all the other tags?

tas50:
@bhousel Should I file a bug to make that happen in iD. Most Hotels (not so much motels) seem to add the city in their name so having a preset override those would be wrong and make searching for hotels really hard in OSM.

bhousel:
> @bhousel Should I file a bug to make that happen in iD. Most Hotels (not so much motels) seem to add the city in their name so having a preset override those would be wrong and make searching for hotels really hard in OSM.

Would you be ok with us using `official_name` for this?
I'm going to merge this for now, but with a `name` tag, but we can change it.

bhousel:
> Would you be ok with us using `official_name` for this?
> I'm going to merge this for now, but with a `name` tag, but we can change it.

^ Actually what I said here was wrong.. The names do really contain the city, and it should appear in `name`, not `official_name`:

screenshot 2019-03-02 23 07 40

But this is ok in iD today because we show both `name` and `brand` fields - the user can change the Name field if they want to.  No need to open a bug for it!

screenshot 2019-03-02 23 08 36

PR #2328, opened 2019-02-19
Filter out Księgarnia

tas50:
This shows up 51 times in the new planet data and is Polish for bookstore. Adding the filter now so can update the allnames file.

### DIFF
--- a/config/filters.json
+++ b/config/filters.json
@@ -148,6 +148,7 @@
         "^kebab(ai|\\shouse)?$",
         "^kitchen$",
         "^kisbolt$",
+        "^księgarnia$",
         "^kwiaciarnia$",
         "^ladestation$",
         "^lékárna$",


Adamant36:
Good idea. I was thinking about doing that myself with some terms that aren't in the database yet. Does it need to be in the discardNames.json file also though? It includes things like `"leisure/park|Anderson Park"`, which probably wasn't in the original database. So it probably wouldn't hurt to include Księgarnia there to, just in case.

tas50:
This won't end up in the discard names yet, but it will when I merge in the update for allnames based on the new planet data. The build will auto generate that based on this filter. I'm slowly adding everything and fiiltering everything necessary to bring in that updated data. There's a ton of brands that just crossed 5 nodes and need to be added though so it's going to take a while.

Adamant36:
> This won't end up in the discard names yet, but it will when I merge in the update for allnames based on the new planet data. The build will auto generate that based on this filter.

That will work then.

PR #2330, opened 2019-02-19
Add missing cuisine tags to restaurants

tas50:
These were throwing errors in the build. Based off their menus and wikipedia entries.


Adamant36:
Just a thought, but it appears juice isn't listed on the cuisine page in the Wiki. It might be worth adding there, since we are using it and it has like 400 uses already.

PR #2331, opened 2019-02-19
Filter out additional generic names

tas50:
Máquina de Tickets: Portuguese for Ticket Machine
Café des Sports: Sports cafe
Café de la Place: seems to be a generic cafe name in France. No single website or consistent branding
Bar Kafe: Cafe Bar
Bar Centrale: Central bar
Café de la Gare: Coffee Station


tas50:
I'm pretty horrible about ticket linking in general. I'll try to get these all cross referenced int he future since it keeps the issues fresh.

PR #2332, opened 2019-02-19
Add wiki links to additional banks

tas50:
The only one here that might be a bit controversial is FNB corporation. There are a few banks in the US that use the First National Bank name, but they append their local to the name so there's First National Bank of Omaha and such. This seems to be the only actual FNB as a full name and it's quite common through the east coast according to their website. Long term it might be a good idea to add First National of Omaha and First National of Texas so that folks don't attempt to use this tag when that's what they really want. Based on overpass queries we have a bit of that to cleanup already.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -2506,6 +2506,8 @@
     "tags": {
       "amenity": "bank",
       "brand": "First National Bank",
+      "brand:wikidata": "Q5426765",
+      "brand:wikipedia": "en:F.N.B. Corporation",
       "name": "First National Bank"
     }
   },
@@ -3087,6 +3089,8 @@
     "tags": {
       "amenity": "bank",
       "brand": "Macro",
+      "brand:wikidata": "Q2335199",
+      "brand:wikipedia": "en:Banco Macro",
       "name": "Macro"
     }
   },
@@ -3115,6 +3119,8 @@
     "tags": {
       "amenity": "bank",
       "brand": "Mercantil",
+      "brand:wikidata": "Q6818004",
+      "brand:wikipedia": "en:Mercantil Servicios Financieros",
       "name": "Mercantil"
     }
   },


Adamant36:
"they append their local to the name so there's First National Bank of Omaha and such"

The question is if they do it with actual tagging practices. Otherwise they will be wrongly associated with the wikipedia tags  when someone looks up the name or brand in iD editor. A lot of times people don't add on the small parts of the name when they are tagging. Like the location or "pharmacy." There was a discussion a while back somewhere to about how we wanted to keep the names and brands semi short and leave the extra junk out when we can. Along with not creating a bunch of new name or brand tags in the database just for the sake of accuracy. I'll try to find the discussion. I'm not sure what the answer for this particular instance is, but its still something to think about.

Adamant36:
I think it was  https://github.com/osmlab/name-suggestion-index/issues/2236#issuecomment-451346930

Here's the relevant part 

--

* We almost always want to use the name that the brand prefers.  So, if "Dunkin' Donuts" is now calling themselves "Dunkin'" then that's the name we should suggest to people, even if they are overwhelmingly already mapped as and signed as "Dunkin' Donuts".  Brands are _fluid_ - they change and that's ok.  Once the `brand:wikidata` tags are more widespread, OSM is in better shape, and we can even encourage mappers to cleanup brand names or do mass renames.
* We also need to pick a name that renders reasonably on a map.  So nothing too long or weird.
  
  * "P.F. Chang's" is better than "P. F. Chang's China Bistro"
  * "Fairfield Inn" is better than "Fairfield Inn & Suites by Marriott"
  * (I think we should shorten "Applebee’s Neighborhood Grill & Bar" to simply "Applebee's")
* And it has to be generally recognizable by people
  
  * "UPS" is more recognizable than "United Parcel Service"
  * but we wouldn't shorten "Dunkin' Donuts" to "DD", even though they write that on the cups.
* "CVS" is a tricky case.  I know they are usually signed as "CVS Pharmacy", and that's technically their brand name too, and in their signage and logo, so that's what I'd go with.

PR #2336, opened 2019-02-21
Add hot dog on a stick

tas50:
Common fast food restaurant in US malls.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -7680,6 +7680,17 @@
       "name": "Hesburger"
     }
   },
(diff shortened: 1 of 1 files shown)


Adamant36:
I asked on the wiki page if hot_dog could be added to the cuisine list. It would be good if the various spellings of it can be unified. Although I don't know if any of this particular restaurant is tagged as such, you might want to add hot_dogs as a match for it at some point. In the meantime I'll still merge it though.

tas50:
Good catch. I just used it since we already had it on Wienerschnitzel.

Issue #163, opened 2018-09-28
suspicious names for second hand clothes


matkoniecz:
> shop/clothes|Second Hand
> shop/clothes|Outlet

bhousel:
This was done

PR #2337, opened 2019-02-22
Filter out "Outlet"

tas50:
This is a generic term


tas50:
Resolves https://github.com/osmlab/name-suggestion-index/issues/1197

Partially resolves https://github.com/osmlab/name-suggestion-index/issues/163

PR #2339, opened 2019-02-22
Add wikipedia entry for Only store

tas50:
It's a European Denim brand. I created the wikidata page which is basically empty since there wasn't one.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -20868,7 +20868,10 @@
     "count": 137,
     "tags": {
       "brand": "Only",
+      "brand:wikidata": "Q61799370",
       "name": "Only",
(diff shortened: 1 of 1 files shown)


tas50:
Resolves https://github.com/osmlab/name-suggestion-index/issues/1193

Adamant36:
Thanks for creating the wikidata pages.

Adamant36:
Maybe you could relate it to the operator on the wikidata page somehow. I'm not sure how that works, but at least it would have something on it then.

tas50:
That one is already linked between the brand and the parent org. I've tried to do that to all of them.

Adamant36:
Ok cool.

PR #2342, opened 2019-02-22
Add wiki entries for evgo

tas50:
I created a wikidata entry for evgo since only the parent company existed.


tas50:
Resolves https://github.com/osmlab/name-suggestion-index/issues/414

Adamant36:
Btw, "resolves" isn't closing the issues after merging them for some reason and I thought it was suppose to. So you might want to use "closes" instead.

PR #2343, opened 2019-02-22
Add missing wiki entries to brands

tas50:
Slowly filling it all in.


tas50:
Resolves https://github.com/osmlab/name-suggestion-index/issues/393 and https://github.com/osmlab/name-suggestion-index/issues/411

PR #2344, opened 2019-02-22
Update Enel and E-WALD charger entries

tas50:
Stazione di ricarica isn't part of the brand name and therefore shouldn't be in the name tag. The same goes for Ladestation.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -6928,23 +6928,30 @@
       "name": "ChargePoint"
     }
   },
-  "amenity/charging_station|E-WALD Ladestation": {
-    "count": 61,
+  "amenity/charging_station|E-WALD": {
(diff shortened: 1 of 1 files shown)


tas50:
Resolve's https://github.com/osmlab/name-suggestion-index/issues/412 and

Adamant36:
63 are tagged as E-WALD Ladestation and there's 223 instances of Enel - stazione di ricarica. So can you at least add them as matches instead of dropping the names completely?

Adamant36:
Oh wait, never mind. It looks like that's what you did. My bad ;)

tas50:
I did a match and this is going to need a challenge eventually to clean it up like all the other matches.

Issue #2243, opened 2019-01-04
Split up canonical.json?


bhousel:
`canonical.json` is about 34000 lines now, and the file is starting to get kind of cumbersome to edit.

Should we split it up?
Maybe create a folder hierarchy like:

- `brands/`
  - `amenity.json`
  - `shop.json`

or nest another level by key/value:

- `brands/`
  - `amenity/`
    - `bank.json`
    - `fuel.json`  
    - `pharmacy.json`
    - ...
  - `shop/`
    - `car.json`
    - `car_repair.json`
    - `chemist.json`
    - `clothes.json`
    - ...

Adamant36:
It might be worth going with the second option so things are easier to organize later on when a lot more companies are inevitably added to the database. Plus, it already takes a lot of time to get through a single category like banks. So the less we have to sift through things that aren't related to a specific category that's being focused on or that already have information the better. 

Maybe it would allow for easier tagging projects based on the database at some point also.

ImreSamu:
From a local data maintainer views  -  **I prefer separated country files:**

*  countries:
   * ca.json  -   Canada - brand/name suggestions
   * us.json  -   U.S - brand/name suggestions
   * hu.json  -   Hungary -   brand/name suggestions 
   * ....
  
Pros:
* As a local maintanier  - I need to check  only my country data.   
   * less data -> easier to find data 'bugs'  -> better data quality
   * less data ->  simple to create country html/excel  lists  ->  and post to the local mail list.     ( now   ~90% is a noise )
   * the `top-bottom` limiting/cleaning  is hard / not working ( https://github.com/osmlab/name-suggestion-index/issues/2107  )
* based on local community decision  -  ( bottom-up )  
  *  moving the data problems to country community level.  
* half- solution- for figurative/unregistered/local brands: ( [amenity/cafe|Caffè Nero](https://github.com/osmlab/name-suggestion-index/issues/2116#issuecomment-433040783)  ) 
   * in this case we don't add `amenity/cafe|Caffè Nero`  - for  the hungarian names.  
* good solution for  shops with general words  ( https://github.com/osmlab/name-suggestion-index/pull/2087 )
* local language support - on wikidata tags  
   * we can use   `brand:wikipedia=hu:McDonald’s`  ;   https://wiki.openstreetmap.org/wiki/Multilingual_names is a big topic in Europe.  
[...]

PR #2345, opened 2019-02-23
fixes#2315 Updates co-op store information.

### LINKED ISSUE #2315: Fix wiki tagging for Co-op in the UK
Presets affected:
- Co-op (supermarket)
- Co-op (convenience)

Currently these are tagged `brand:wikidata=Q5440676` and `brand:wikipedia=en:Federated Co-operatives` which is a Canadian brand. In the UK they should be tagged `brand:wikidata=Q3277439` and `brand:wikipedia=en:Co-op Food` 

And yes, there is overlap with The Co-operative Food which is the old name. Not sure if it's been completely phased out yet though, so it makes sense to keep both names in the suggestion index.

Zilula:
fixes #2315 Updated co-op convenience and superstore information in config/canonical.json file. Used the information in the issue request to replace the old set of information.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -22249,8 +22249,8 @@
     ],
     "tags": {
       "brand": "Federated Co-operatives",
-      "brand:wikidata": "Q5440676",
-      "brand:wikipedia": "en:Federated Co-operatives",
+      "brand:wikidata": "Q3277439",
+      "brand:wikipedia": "en:Co-op Food",
       "name": "Co-op",
       "shop": "convenience"
     }
@@ -30724,8 +30724,8 @@
     ],
     "tags": {
       "brand": "Federated Co-operatives",
-      "brand:wikidata": "Q5440676",
-      "brand:wikipedia": "en:Federated Co-operatives",
+      "brand:wikidata": "Q3277439",
+      "brand:wikipedia": "en:Co-op Food",
       "name": "Co-op",
       "shop": "supermarket"
     }


Adamant36:
I haven't been following along with this, are the convenience stores and supermarkets essentially the same stores? If so, they should probably just be combined in the same entry with a match. 

@tas50 any thoughts?

config/canonical.json
```diff
@@ -22249,8 +22249,8 @@
     ],
     "tags": {
       "brand": "Federated Co-operatives",
-      "brand:wikidata": "Q5440676",
-      "brand:wikipedia": "en:Federated Co-operatives",
+      "brand:wikidata": "Q3277439",
+      "brand:wikipedia": "en:Co-op Food",
```
matkoniecz: Based on https://github.com/osmlab/name-suggestion-index/issues/2315 both en:Federated Co-operatives and en:Co-op Food are correct, just in different areas.

So both should be present.

matkoniecz: Doing this is a bit tricky - see https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#identical-names-multiple-brands

SilentSpike:
@Adamant36 

I'm in the UK and every co-op I've encountered I would class as a convenience store. However, as wikipedia says:

>Though often considered to be one supermarket business, The Co-operative Food is a network of supermarkets and convenience shops owned and operated by over 15 independent co-operative societies, many of which have adopted the 2008 version of The Co-operative brand.

So there should be an entry for both, but it will be hard to stop people from mistagging instances as one of the other. Personally, I've been tagging them as The Co-operative Food brand elsewhere in the list (because this is what is written on the signs where I am):

https://github.com/osmlab/name-suggestion-index/blob/548530d9ac6d34141cdccec47937fe3644d60979/config/canonical.json#L23285-L23297

Now that I think about it I think the original issue is quite misleading. The entry for "The Co-operative Food" is likely intended to represent the UK brand, whereas the Co-op entry is correctly representing the Canadian brand and should be entirely left alone.

It then comes down to a question of the correct name for the UK brand. It already has a supermarket entry too as expected:

https://github.com/osmlab/name-suggestion-index/blob/548530d9ac6d34141cdccec47937fe3644d60979/config/canonical.json#L33352-L33368

Adamant36:
Hhmmm. Have you looked at usage area for 
Co-op to make sure a lot of them aren't generic? I'm wondering if some of them are just general co-ops or not. There'd really be no way to tell which are legitimate and with 284 uses Id bet at least some of them are.

lakedistrictOSM:
I'd guess that most of them are convenience stores, but there are some larger branches which would be classed as supermarkets. They rebranded from The Co-operative Food to just Co-op a couple of years ago, but I don't think that all stores have the new signage yet.  

Wiki says that Co-operative food is the old name: https://en.wikipedia.org/wiki/Co-op_Food
New signage: https://upload.wikimedia.org/wikipedia/commons/3/39/Co-op_Old_Street%2C_Shoreditch%2C_rebranding_launch_by_The_Co-operative_Group.jpg
Old signage: https://upload.wikimedia.org/wikipedia/commons/1/1f/The_Co-operative_Food_Branch_%287795238312%29.jpg

Regardless, the Canadian Co-op preset needs to be made Canada only to prevent UK Co-ops getting the wrong wiki tags.

config/canonical.json
Zilula: Interesting. Thank you for the reference. Ill check out it out again and make the needed changes best I can.

matkoniecz: Unfortunately this issue is a bit trickier than a typical one, what is a bit unlucky for a first issue :(

I wish you a good luck and feel free to ask if documentation is unclear - or even better, open a new issue (unclear documentation is a problem that should be fixed).

SilentSpike:
So, to help you out @Zilula:

Based on the discussion on this PR it looks to me like the list needs two new entries for `"shop/supermarket|Co-op"` and `"shop/convenience|Co-op"` which both have `"brand:wikidata": "Q3277439"` and `"brand:wikipedia": "en:Co-op Food"` (see [these guidelines](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#convenience_store--add-missing-brands)).

However, as these existing entries of the same name are a valid Canadian brand you'll have to differentiate them as per those [identical name guidelines](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#identical-names-multiple-brands) (I'd suggest using `~(Canada)` and `~(UK)`).

You can also add  `"countryCodes: ["ca"]` and `"countryCodes": ["gb"]` respectively so that downstream consumers can selectively show them based on where is being edited (at some future point we hope this is possible, currently we don't actually include this data in releases).

This turned out to be a bit of a harder issue than expected 😋

Zilula:
@SilentSpike Great! Thanks for all the helpful tips. I`ll switch some things up and get back to you.

bhousel:
Because this issue is a bit complicated, I'm going to close this PR out for now..  

I'm trying to clear out the PR backlog before splitting up the `canonical.json` file per #2243.  💣 
Feel free to open up another PR after the dust settles!

Issue #2326, opened 2019-02-19
Spli Ramada up into all the sub-brands


tas50:
There are several sub-brands that should be used here instead of Ramada which isn't an actual brand.

https://en.wikipedia.org/wiki/Ramada#Hotel_tiers

Adamant36:
I know entries without Wikipedia data are filtered out of the name-suggestion file. Are they also filtered out if they have a operator:wikipedia tag or do those get added? If not, it might be worth just having them as matches to Ramada so they will be included. Since they probably aren't note worthy enough on their own for individual Wikipedia articles. Maybe Wikidata, but I think it has similar notability guidelines.

tas50:
They need to get their own unique names since the brand has a pretty big spread and each brand means something different. They would all have the wikipedia entry of Ramada and the operator of Wyndham Worldwide, which owns and operates all those sub-brands. The hotel industry is a crazy web of brands.

ArloJamesBarnes:
"Maybe Wikidata, but I think it has similar notability guidelines."
Wikidata is significantly more lax: https://www.wikidata.org/wiki/Wikidata:Notability
In the case of brands, criterion 2 is easy to use since there is usually writing about them on the 'net. Verifiability (linking to references) is much more valued by Wikidatans. You could also have a Q-number for each sub-brand, relate them all to a parent brand, which would have a sitelink to the general Wikipedia article on the various langwikis, and it would be totally kosher.

SilentSpike:
@Adamant36 I believe it's not the wikipedia tag that gatekeeps inclusion in the list, but the wikidata tag (as it is an effectively static ID compared to anything else we can reasonably tag in OSM).

As @ArloJamesBarnes says wikidata items are quite easily created as long as they are notable - which pretty much all brands are.

tas50:
I'm fine with creating unique wikidata IDs for each of these. We should probably confirm we have unique IDs elsewhere. The real blocker here though is getting ID to allow the user to change the name since most of these will have a unique name in addition to their brand tags.

Adamant36:
@SilentSpike, my bad. Your correct. Its the Wikidata entry that's important. 

> You could also have a Q-number for each sub-brand, relate them all to a parent brand, which would have a sitelink to the general Wikipedia article on the various langwikis, and it would be totally kosher.

[...]

Issue #2274, opened 2019-01-29
countryCodes lost in distribution


matkoniecz:
Compare ABANCA (the first entry) in https://raw.githubusercontent.com/osmlab/name-suggestion-index/master/dist/name-suggestions.json (dist) and https://raw.githubusercontent.com/osmlab/name-suggestion-index/master/config/canonical.json (config)

countryCodes value is lost despite that it would be useful to, for example, limit useless name suggestions

Related to #2201 (without fixing this benefit of #2201 would be lost)

bhousel:
Yes, we don't have the ability to limit the presets by `countryCode` in iD yet and I wanted to keep the `name-suggestions.json` format backward compatible for now.  (I don't think anyone else is using these files anyway). 


It probably makes more sense for iD to directly consume `canonical.json` now and just deprecate `name-suggestions.json`.

simonpoole:
@bhousel Vespucci has been using the json format since day one essentially, afaik the change wouldn't break anything. Wrt cannonical.json can you give us an idea how that stacks up against the distributed file?

bhousel:
> @bhousel Vespucci has been using the json format since day one essentially, afaik the change wouldn't break anything. Wrt cannonical.json can you give us an idea how that stacks up against the distributed file?

It wouldn't break anything to add `countryCodes` to the distributed json file, but if we add those and also the alternate `match` names (which some people suggested using as preset search terms) then we basically just have `canonical.json`.  (I'm happy to keep `name-suggestions.*` around indefinitely)

matkoniecz:
> Wrt cannonical.json can you give us an idea how that stacks up against the distributed file?

Main differences is that entries umatched with wikidata/wikipedia (and therefore possibly false) are dropped. Also `countryCodes`, `match`, `nomatch` properties are dropped.

> It wouldn't break anything to add countryCodes to the distributed json file, but if we add those and also the alternate match names (which some people suggested using as preset search terms) then we basically just have canonical.json

Dropping unmatched entries would be remaining difference.

simonpoole:
What confuses me is that it seems that cannonical.json has 3699 entries, but dist/name-suggestions.json only 2841. Shouldn't things be the other way around?

bhousel:
> What confuses me is that it seems that cannonical.json has 3699 entries, but dist/name-suggestions.json only 2841. Shouldn't things be the other way around?

[...]

PR #2350, opened 2019-02-23
Add 11 new supermarkets

tas50:
There are 50 or more of each of these stores in the new planet data except for Super One Foods, which is a smaller chain from North Dakota / Wisconsin that I added so people wouldn't mis-tag with Super 1 Foods which is from Texas.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -31784,6 +31784,16 @@
       "shop": "supermarket"
     }
   },
+  "shop/supermarket|Jan Linders": {
+    "nocount": true,
+    "tags": {
+      "brand": "Jan Linders",
+      "brand:wikidata": "Q2200982",
+      "brand:wikipedia": "nl:Jan Linders Supermarkten",
+      "name": "Jan Linders",
+      "shop": "supermarket"
+    }
+  },
   "shop/supermarket|Jewel-Osco": {
     "count": 81,
     "countryCodes": ["us"],
@@ -31995,6 +32005,26 @@
       "brand:wikidata": "Q151954",
       "brand:wikipedia": "en:Lidl",
       "name": "Lidl",
+      "shop": "supermarket"
+    }
+  },
+  "shop/supermarket|Lincolnshire Co-op": {
+    "nocount": true,
+    "tags": {
+      "brand": "Lincolnshire Co-op",
+      "brand:wikidata": "Q6551231",
+      "brand:wikipedia": "en:Lincolnshire Co-operative",
+      "name": "Lincolnshire Co-op",
+      "shop": "supermarket"
+    }
+  },
+  "shop/supermarket|Linella": {
+    "nocount": true,
+    "tags": {
+      "brand": "Linella",
+      "brand:wikidata": "Q61085990",
+      "brand:wikipedia": "ro:Linella",
+      "name": "Linella",
       "shop": "supermarket"
     }
   },
@@ -32979,6 +33009,16 @@
       "shop": "supermarket"
     }
   },
+  "shop/supermarket|Save-On-Foods": {
+    "nocount": true,
+    "tags": {
+      "brand": "Save-On-Foods",
+      "brand:wikidata": "Q7427974",
+      "brand:wikipedia": "en:Save-On-Foods",
+      "name": "Save-On-Foods",
+      "shop": "supermarket"
+    }
+  },
   "shop/supermarket|Schnucks": {
     "match": ["shop/supermarket|Schnuck's"],
     "nocount": true,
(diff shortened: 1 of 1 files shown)


Adamant36:
家乐福 is already in there as Carrefour. So is 沃尔玛 under the normal Walmart tag. I don't think it would be sustainable to have individual entries for each language of a global company. @bhousel also went as far as to say there should only be the brand name tag in the language where it resides and and also English, but no other "extra" languages. So that's not an option either. I don't think it matters though, because people can just go to the page of their language from the main Wikidata page anyway.  You could possibly add 家乐福 and 沃尔玛 as matches though. Then individual apps can do the name translation on their side if need be.

maxerickson:
I think there should be country specific branding if the store happens to have country specific branding. For the most part, Walmart has purchased retailers outside of the US, so adding Walmart-like brands where they brand stores like that is something like 5 or 10 entries (look at a few links under international):

https://corporate.walmart.com/our-story/our-business

countryCodes would of course apply.

Re sustainability, there's far more podunk regional brands than there are retailers with significant international reach.

Adamant36:
> I think there should be country specific branding if the store happens to have country specific branding.

I totally agree. Like in  El Salvador Walmart operates Despensa Familiar. Which should obviously have its own entry. There are also Walmart Supercenter's there though. Which shouldn't in my opinion. That's what country codes are for. In this Particular case 沃尔玛 is directly translated as Walmart and they share the same Wikidata entry. Therefore, its not a subsidiary or a spin off brand. Its just Walmart. So it shouldn't have its own entry. 

More generally, there can be problems with using country codes if its a global brand. Since its not realistic to list every country the brand is in. So it would be cool if there was a "global" tag or more precise would be a "not included in country" tag. So the few countries that don't contain the brand could be listed and it could be filtered that way. 

> Re sustainability, there's far more podunk regional brands than there are retailers with significant international reach.

Yeah, but in those cases they either aren't listed because its only one store or they are because its a unique brand has enough reach to matter. In the cases where they are included its a better opportunity then creating 25 different unique listings for every language Walmart's store name is name. Especially if the different languages don't even have any actual tagging usage.

maxerickson:
I meant regional brands like Super 1 and Super One, not independent stores.

If you review the international links, you'll see that Walmart only operates stores with "Wal" in the name in the US, Argentina, Canada, China and Central America. The branding is shared in Central America, so it's much less than 25 entries.

As an example, stores in Canada are branded as "Walmart Supercentre", which is spelled differently than the US stores. This should be a separate entry.

Adamant36:
Like I said though, in this case the stores still say Walmart on the signs. 沃尔玛 is a suffix. Which isn't how they are tagged. Plus its already been agreed that suffix are generally discarded. See https://github.com/osmlab/name-suggestion-index/issues/2326#issuecomment-466035521 and https://github.com/osmlab/name-suggestion-index/issues/2236#issuecomment-451346930

walmart
walmart 2

maxerickson:
The entries for US stores use "Supercenter" and "Neighborhood Market" which are similarly placed on the storefronts.

Adamant36:
And? They have separate entries because the store formats/layouts are vastly different. Its not just because they have the word "supercenter" at the end of the signs. Whereas, in this case we are talking about only the brand name. Not the store formats. So its a completely different situation.

Adamant36:
Even supercenters in China still use the English name for Walmart on their signs. Litterally everyone visiting the stores knows what Walmart is and can read the English name or they wouldn't use it over the Chinese. 
supercenter
wal-mart_3

tas50:
I'm not attached to exactly how we do this, but it seems that we've been pretty inconsistent so far. Look up the various local language entries for starbucks as an example. The Starbucks sign in China just says Starbucks in English, but users pretty consistently tag it as 星巴克 (258 times) and this repo has it's own entry for that. In fact we have Starbucks four times depending on what language it is. That seems exactly the same as this Walmart setup where the sign also includes the name Walmart in English, but users don't use that in the tag.

Adamant36:
>The Starbucks sign in China just says Starbucks in English, but users pretty consistently tag it as 星巴克 (258 times) and this repo has it's own entry for that. In fact we have Starbucks four times depending on what language it is. The argument could actually be made in that case that a second listing isn't necessary also. Since its not exclusively or even favored as the Chinese either. 

Actually there's Starbucks signs in China that exclusively have the name in Chinese. Although some have both where the English is "Starbucks Coffee" but the Chinese is just "Starbucks." They also some exclusively in Chinese also. I think the tagging reflects that also. If you do an OverPassTurbo search in China for name=星巴克, name=Starbucks, and name:en=Starbucks its about even. A lot the name:en entries don't have the Chinese either. So Starbucks isn't a good comparison because of the different signs and its even objectionable if there should be different entries. Since Chinese isn't even preferred. 

>Walmart in English, but users don't use that in the tag.

That's actually not the case either. If you do a similar search on OverPassTurbo to the one suggested above only with Walmart as the names, its slightly slanted toward English because most places tagged as name:en don't have the Chinese name.

bhousel:
> 家乐福 is already in there as Carrefour. So is 沃尔玛 under the normal Walmart tag. I don't think it would be sustainable to have individual entries for each language of a global company. @bhousel also went as far as to say there should only be the brand name tag in the language where it resides and and also English, but no other "extra" languages.

It's ok to have multiple entries in this index for the international brands so that mappers around the world can use them.  Meaning, if mappers in China want to use " 沃尔玛" instead of "Walmart" then we should make an entry for that.  Because we seed this list with actual data from OSM, we have a good insight to which brands actually have worldwide coverage and local branding - it's not many.

I would like to eventually see these entries scoped with `countryCodes` so that users are only presented with brand names that they can actually read, but since we aren't scoping them in iD yet (#2274) it doesn't matter much.

PR #2359, opened 2019-02-24
Adding brand Красный Яр

alphagamer7:
Fixes issue #1778


tas50:
Thanks @alphagamer7

Adamant36:
Just a heads up, but in the future you might want to put "closes" instead of fixes and also leave out "issue" part of it. Otherwise it won't auto close the issue the PR is connected to and we have to do it ourselves later instead. Like I did in this case.

Issue #1785, opened 2018-10-17
Research needed for: shop/supermarket|Светофор


bhousel:
#### :thinking: We need help deciding what to do with "shop/supermarket|Светофор"

:point_right: Read the [CONTRIBUTING.md](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) guide for more info, and feel free to ask questions on this issue!

### What's needed

First, do some research on `shop/supermarket|Светофор` by searching Google and Wikipedia.

#### If it is an actual brand:

* Edit `config/canonical.json`:
  * Find the entry for `shop/supermarket|Светофор`
  * Add tags for `brand:wikidata` and `brand:wikipedia`
  * Optionally add `"countryCodes"` or other tags
* `npm run build`
* Commit your changes and submit a pull request

#### If it is not an actual brand:

* Edit `config/filters.json` to add an expression that discards it
* Edit `config/canonical.json` to remove the entry for `shop/supermarket|Светофор`
* `npm run build`
* Commit your changes and submit a pull request

#### If the situation is complicated:

* Leave a comment on this issue with what you find

:sparkles: Thanks!

_This is an autogenerated ticket_

maxerickson:
Someone can create a new Wikidata entry for it and we can add that, there doesn't need to be a Wikipedia entry.

mukulkhanna:
Hi @maxerickson , I was making a Wikidata entry for the above. There was an option to choose the language. Should I choose English or Russian (because its a supermarket in Russia)?

mukulkhanna:
@maxerickson I have created the wikidata for the above.

https://www.wikidata.org/wiki/Q61875920

PR #2362, opened 2019-02-25
Added info for Светофор, Командор and Десяточка supermarkets

mukulkhanna:
Resolving #1785 

- Made a wikidata entry for Светофор supermarket
- Added wikidata, countrycodes and brand:en in config/canonical.json
- Build successful ✅

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -34163,8 +34163,11 @@
   },
   "shop/supermarket|Светофор": {
     "count": 116,
+    "countryCodes": ["ru"],
     "tags": {
       "brand": "Светофор",
+      "brand:en": "Svetofor",
+      "brand:wikidata": "Q61875920",
       "name": "Светофор",
       "shop": "supermarket"
     }
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -33914,8 +33914,11 @@
   },
   "shop/supermarket|Десяточка": {
     "count": 55,
+    "countryCodes": ["ru"],
     "tags": {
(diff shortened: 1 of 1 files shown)


mukulkhanna:
Also resolving #1776 and #1774

maxerickson:
I'm not terribly worried about Wikidata, but I guess the entry should have a Russian name and then also maybe an English name. Putting the Russian name in the English name isn't the way to do it. I've also tried to add some statements to any entries I've created (because that makes the entries useful for Wikidata).

mukulkhanna:
@maxerickson Sorry for the misunderstanding. I can make the corresponding changes.

As of now, I have entered the english spelling for the Russian brand as the "brand:en" value.
Just wanted to clear out that for the english name ("brand:en"), you would like the meaning of the Russian name in english, right?

Adamant36:
Just a heads up, but I think usually its name:en that's entered as it resolves errors, along with being more accessible. Brand:en isn't as important. Although it can still be added. Also, you should put "closes" before  the issue #. Which auto closes it the PR is merged. "Resolving" doesn't do anything. As can be seen by the fact that the issues "resolved" here are still open. 

It would be good if you or @maxerickson closed them. Probably in this case @maxerickson should do it since its apart of the basic maintainance tasks of a maintainer to close old issue. In the future though, it would be good to just put "closes" in your PR messages to save the extra steps. 

Thanks

mukulkhanna:
@Adamant36 Okay, will take care next time. Thank you.

PR #2364, opened 2019-02-27
Add missing brands from the US West Coast

tas50:
These are common brands I'm running into while mapping shopping centers on the west coast.  I also added some country codes.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -14828,7 +14828,7 @@
       "name": "Cheesecake Factory"
     }
   },
-  "amenity/restaurant|Chevys Fresh Mex": {
+  "amenity/restaurant|Chevys": {
     "match": [
       "amenity/fast_food|Chevys",
       "amenity/restaurant|Chevys"
@@ -14840,7 +14840,8 @@
       "brand:wikidata": "Q5094466",
       "brand:wikipedia": "en:Chevys Fresh Mex",
       "cuisine": "mexican",
-      "name": "Chevys Fresh Mex"
+      "name": "Chevys",
+      "official_name": "Chevys Fresh Mex"
     }
   },
   "amenity/restaurant|Chifa": {
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -14638,10 +14638,8 @@
       "name": "ASK Italian"
     }
   },
-  "amenity/restaurant|Applebee’s Neighborhood Grill & Bar": {
-    "count": 1108,
+  "amenity/restaurant|Applebee’s": {
     "match": [
-      "amenity/restaurant|Applebee's",
       "amenity/restaurant|Applebee's Bar & Grill",
       "amenity/restaurant|Applebee's Bar + Grill",
       "amenity/restaurant|Applebee's Bar and Grill",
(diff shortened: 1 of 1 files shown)


config/canonical.json
```diff
@@ -8600,6 +8600,25 @@
...
+      "amenity/restaurant|The Habit Burger",
+      "amenity/restaurant|The Habit Burger Grill"
+    ],
+    "nocount": true,
+    "tags": {
+      "amenity": "fast_food",
+      "brand": "The Habit Burger Grill",
+      "brand:wikidata": "Q18158741",
+      "brand:wikipedia": "en:The Habit Burger Grill",
+      "cuisine": "sandwich",
```
matkoniecz: are you sure that it should be `"cuisine": "sandwich",` not `"cuisine": "burger",` (question based solely on name) - `cuisine=burger` is a popular tag https://taginfo.openstreetmap.org/keys/cuisine#values

config/canonical.json
```diff
@@ -14766,6 +14803,21 @@
...
+      "amenity/restaurant|Chevys"
+    ],
+    "nocount": true,
+    "tags": {
+      "amenity": "restaurant",
+      "brand": "Chevys Fresh Mex",
+      "brand:wikidata": "Q5094466",
+      "brand:wikipedia": "en:Chevys Fresh Mex",
+      "cuisine": "mexican",
+      "name": "Chevys Fresh Mex"
```
matkoniecz: Is it actually used name? Maybe `name=Chevys`, `official_name=Chevys Fresh Mex` would be better? (Again I know nothing about this fast food).

matkoniecz:
Can you self-review using for example https://github.com/osmlab/name-suggestion-index/pull/2364/files and confirm that it is ready?

config/canonical.json
tas50: Yep this was entirely a copy/paste error on my part. I updated it to burger

config/canonical.json
tas50: The name on the sign is Chevys Fresh Mex and the name that other maps use is Chevys Fresh Mex. It seemed appropriate.

Adamant36: Personally id go with Chevys. Fresh Mex is another case of an unnecessary suffix. No one calls it that in the real world and there's almost zero chance of it being confused with another brand by naming it just Chevys. I agree it would work as an official name tag though (that's kind of what the point in the tag is).

bhousel: > Personally id go with Chevys. Fresh Mex is another case of an unnecessary suffix. No one calls it that in the real world and there's almost zero chance of it being confused with another brand by naming it just Chevys. I agree it would work as an official name tag though (that's kind of what the point in the tag is).

Yeah.. We have been inconsistent about this in the past too.. Let's just decide going forward to put the short "renderable" name in `name` and the longer name in `official_name`.   (looking at you "Applebee’s Neighborhood Grill & Bar" 👀 )

tas50: I'll update this when I'm done with work and get Applebees in there as well. As long as it's available for search I'm fine with it. There's a bunch of them like this. Five Guys is really Five Guys Burgers and Fries, which should show up if you search for it, but doesn't need to display on the map. official_name should fix that nicely.

tas50: @bhousel let me know if this is what you had in mind

matkoniecz:
@tas50 Let us know once it is ready for review!

Adamant36:
@matkoniecz It seems good to me. It looks like he updated everything that needed updating. 

So I think you can merge it if you want. Its not like he can't update something after merging if he wants, but I don't see any errors with it.

bhousel:
Thanks @tas50 - looks good 👍 
I'm going to switch the apostrophe in Applebee's too.

PR #2365, opened 2019-02-28
More Japanese Brands

sguinetti:
Hello again. This pull request contains several commits: adding translations to English of Japanese brands, in addition to new found brands in this country (20+). Also, fixes Cineplanet brand name and Cencosud operator in Metro and Paris brands. Thanks.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -18456,9 +18456,11 @@
     "countryCodes": ["jp"],
     "tags": {
       "brand": "オートバックス",
+      "brand:en": "Autobacs",
       "brand:wikidata": "Q7886426",
       "brand:wikipedia": "ja:オートバックスセブン",
       "name": "オートバックス",
+      "name:en": "Autobacs",
       "shop": "car_parts"
     }
   },
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -5304,9 +5304,11 @@
     "tags": {
       "amenity": "bank",
       "brand": "りそな銀行",
+      "brand:en": "Resona Bank",
       "brand:wikidata": "Q11281447",
       "brand:wikipedia": "ja:りそな銀行",
-      "name": "りそな銀行"
+      "name": "りそな銀行",
+      "name:en": "Resona Bank"
     }
   },
   "amenity/bank|三井住友銀行": {
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -14017,7 +14017,8 @@
       "brand:wikidata": "Q11301948",
       "brand:wikipedia": "ja:ココカラファイン",
       "healthcare": "pharmacy",
-      "name": "ココカラファイン"
+      "name": "ココカラファイン",
+      "name:en": "Cocokara Fine"
     }
   },
   "amenity/pharmacy|サンドラッグ": {
@@ -14029,7 +14030,8 @@
       "brand:wikidata": "Q11305867",
       "brand:wikipedia": "ja:サンドラッグ",
       "healthcare": "pharmacy",
-      "name": "サンドラッグ"
+      "name": "サンドラッグ",
+      "name:en": "Sundrug"
     }
   },
   "amenity/pharmacy|スギ薬局": {
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -13945,10 +13945,12 @@
     "tags": {
       "amenity": "pharmacy",
       "brand": "ウエルシア薬局",
+      "brand:en": "Welcia Yakkyoku",
       "brand:wikidata": "Q11288687",
       "brand:wikipedia": "ja:ウエルシア薬局",
       "healthcare": "pharmacy",
-      "name": "ウエルシア薬局"
+      "name": "ウエルシア薬局",
+      "name:en": "Welcia Yakkyoku"
     }
   },
   "amenity/pharmacy|カワチ薬品": {
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -18469,9 +18469,11 @@
     "countryCodes": ["jp"],
     "tags": {
       "brand": "タイヤ館",
+      "brand:en": "Taiyakan",
       "brand:wikidata": "Q11315808",
       "brand:wikipedia": "ja:タイヤ館",
       "name": "タイヤ館",
+      "name:en": "Taiyakan",
       "shop": "car_parts"
     }
   },
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -13931,12 +13931,12 @@
     "tags": {
       "amenity": "pharmacy",
       "brand": "ウエルシア",
-      "brand:en": "Welcia Holdings",
+      "brand:en": "Welcia",
       "brand:wikidata": "Q11288684",
       "brand:wikipedia": "ja:ウエルシアホールディングス",
       "healthcare": "pharmacy",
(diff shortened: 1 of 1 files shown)


config/canonical.json
```diff
@@ -18434,6 +18515,7 @@
       "brand:wikidata": "Q11315808",
       "brand:wikipedia": "ja:タイヤ館",
       "name": "タイヤ館",
+      "name:en": "Taiyakan",
```
matkoniecz: is skipping `brand:en` deliberate here?

config/canonical.json
```diff
@@ -21840,9 +21926,11 @@
     "countryCodes": ["jp"],
     "tags": {
       "brand": "ワークマン",
+      "brand:en": "Workman",
       "brand:wikidata": "Q11351660",
       "brand:wikipedia": "ja:ワークマン",
       "name": "ワークマン",
+      "name:es": "Workman",
```
matkoniecz: `name:en`, right?

config/canonical.json
```diff
@@ -21851,9 +21939,11 @@
     "countryCodes": ["jp"],
     "tags": {
       "brand": "洋服の青山",
+      "brand:en": "Aoyama Tailor",
       "brand:wikidata": "Q11661241",
       "brand:wikipedia": "ja:青山商事",
       "name": "洋服の青山",
+      "name:es": "Aoyama Tailor",
```
matkoniecz: `name:en`, right?

config/canonical.json
```diff
@@ -24587,6 +24679,7 @@
     "countryCodes": ["jp"],
     "tags": {
       "brand": "ポプラ",
+      "brand:en": "Poplar",
       "brand:wikidata": "Q7229380",
       "brand:wikipedia": "ja:ポプラ (コンビニエンスストア)",
       "name": "ポプラ",
```
matkoniecz: is skipping `name:en` deliberate?

config/canonical.json
```diff
@@ -24956,7 +25049,10 @@
     "countryCodes": ["jp"],
     "tags": {
       "brand": "ほっともっと",
```
matkoniecz: is skipping `brand:en` deliberate here?

config/canonical.json
```diff
@@ -28449,6 +28577,7 @@
       "brand:wikidata": "Q11190521",
       "brand:wikipedia": "ja:au (通信)",
       "name": "auショップ",
+      "name:en": "au shop",
```
matkoniecz: "au shop"? Not "au" or "Au"?

config/canonical.json
```diff
@@ -32570,6 +32700,8 @@
       "brand:wikipedia": "en:Tiendas Metro",
       "name": "Metro",
       "operator": "Cencosud",
+      "operator:wikidata": "Q1053372",
+      "operator:wikipedia": "es:Cencosud",
```
matkoniecz: es?

matkoniecz:
Thank you very much for your contribution!

It may be better to make more smaller PRs rather one giant one, this way it is easier to spot typos and review changes.

config/canonical.json
```diff
@@ -13901,10 +13907,12 @@
     "tags": {
       "amenity": "pharmacy",
       "brand": "くすりの福太郎",
+      "brand:en": "Kusurino FUKUTARO",
```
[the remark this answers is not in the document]
sguinetti: Yes. Because, in Japanese Wikipedia is write as 
"Kusurino FUKUTARO" (with capital letters).

config/canonical.json
sguinetti: Fixed.

config/canonical.json
sguinetti: Fixed.

config/canonical.json
sguinetti: Yes. "Shop" in japanese-language is "ショップ". For example: https://onlineshop.au.com/top/CSfTop.jsp

config/canonical.json
sguinetti: Fixed. I changed to "en".

config/canonical.json
sguinetti: Fixed

config/canonical.json
sguinetti: Yes. Fixed.

config/canonical.json
sguinetti: Yes. Fixed.

sguinetti:
@matkoniecz Ok. After this I will add some additional shops in another pull request. Meanwhile, I did a few improvements for some stores in Japan, because I could forget in add them. I hope not to cause problems for that.

Adamant36:
"It may be better to make more smaller PRs rather one giant one, this way it is easier to spot typos and review changes."

+1 Things are usually reviewed and merged pretty quickly in this repo. So there's really no need to do more then a few entries at a time. Especially when it comes to adding new brands. Since each one involves multiple lines of code. 

It might be ok to a bunch of entries for something like adding country codes were its only one line, but id still be careful not to over do it there either. 

Its also probably not a good idea to add new brands after someone does an initial code review, because its harder to keep track of in that case.

Adamant36:
@sguinetti, if Donki is what the store is normally called by people who shop there then it should just be the key for the name tag. Then the longer name can be tagged as official_name. That's what we do in other instances where the company is refereed to with a shortened form of the longer name by costumers, but the longer name is the "official" one.  

If Donki isn't what the stores are called by the costumers though and its not the official name, then it shouldn't be in the index IMHO. Like in America McDonald's is sometimes refereed to as "Mickey D's" because its shorter then McDonald's, but it shouldn't be in the index as a "short name" just because of that.

bhousel:
> I see no problems now, though it would be nice if someone else looked at it.

I just took a look and it seems ok!   Thanks @sguinetti 👍

sguinetti:
@Adamant36 I would agree you. But Don Quixote recognizes "Donki" as the official name. In this website (http://www.donki.com/en/) Don Quijote shows their products and appears at the bottom of the page "Copyright Don Quijote Co.,Ltd.". The case of McDonalds is different, because it is known by the public but it is not an official mark (the alias "Mickey D" uses "Mickey Mouse", a trademark of Disney).

sguinetti:
@bhousel @matkoniecz Thank you.

matkoniecz:
BTW, "Mickey D" may fit as `loc_name` but I would not assume that it is use everywhere.

bhousel:
I really don't want to overcomplicate things in this project by adding a bunch of unnecessary name tags.

sguinetti:
@matkoniecz I don't sure. In Wikitionary shows quotes. https://en.wiktionary.org/wiki/Mickey_D%27s Also, "D" is for Donald, I suppose, other caracther from Disney. Interesting anecdote.

matkoniecz:
@bhousel Yes, local nicknames should not be handled here as such things are local and require local knowledge.

config/canonical.json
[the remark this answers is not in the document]
ArloJamesBarnes: (since both traditional [familyname givenname] and Western [givenname familyname] name orders are used in Japan now, the caps are sometimes used to remove ambiguity [where it exists; sometimes it is obvious] about which is the family name)

PR #2366, opened 2019-03-01
Added wikidata and wikipedia data for Republic Bank

cmchumak:
Added wikidata & wikipedia info for Republic Bank


bhousel:
Thanks again @cmchumak

PR #2372, opened 2019-03-03
Update allnames.json and regenerate the world

tas50:
I regenerated this about 2 weeks ago using a new planet file. This brings in the latest new common tags being used throughout OSM. In doing that it adds a bunch of new entries that are missing wiki entries, which is expected and file since we can start filling those in as time goes on. It also required adding a bunch of new matches and filters to prevent bad tags or generics from getting their own entry.


bhousel:
This looks great @tas50  - thank you for doing this! 😄

PR #2373, opened 2019-03-03
Collapse TGP into a single entry

tas50:
I created a wikipedia entry for this and I added the official_name tag.
The signs and branding are all TGP, but the name is really The Generics
Pharmacy.


tas50:
Closes #699

bhousel:
Looks good, thanks!

PR #2374, opened 2019-03-03
Filter out generic tailors

tas50:
Mens
Womens
Mixed


Adamant36:
You might have waited on the PR until #2243 is done since its currently being worked on.

tas50:
Oh nice. I'll wait on that. I look forward to GitHub being able to display a diff, as well as Atom/Sourcetree not crashing due to the diff size.

matkoniecz:
@tas50 Thanks for noticing this generic entries!

Adamant36:
There's Juicy Couture clothing stores out there and some other ones with random "couture" names. So it might be worth keeping the generic one and adding it as a match for a new "Juicy Couture" entry if there isn't one already (it looks like the name Juicy Couture has 16 uses).

tas50:
@Adamant36 There's plenty of valid Couture names so instead of a name filter I did this one on the tailor key. It should only catch the bogus mens/womens/mixed tailors.

Adamant36:
How do you know they aren't Couture licensed or branded tailers in other branded clothing stores?

bhousel:
> How do you know they aren't Couture licensed or branded tailors in other branded clothing stores?

This is something I think about sometimes - whether there is an actual brand that has a generic word for a name.  I think it's ok to keep them excluded until we find examples of it.

tas50:
In this case they're almost entirely in 1 city in the Ivory Coast. I'd imagine they're probably all from 1 mapper or a group of mappers. Couture is part of several brand names, but that's just because it's a term for handmade fitted clothes. It's pretty safe to say these are all just tailors.

Adamant36:
Hhhmmm, I'll have to leave a note there asking what the actual names of the places are.

tas50:
We should definitely get a note on those locations for further surveying.

PR #2367, opened 2019-03-01
More Japanese Brands (part 2)

sguinetti:
This is a second pull request, the previous one was  #2365. Add business brands about books, restaurants, ice cream and fast food. Also, I add operator key to "Don Quijote" brand and fixes Wikipedia article. Thanks.

- About Baskin-Robbins: Because BR is a registered trademark in Japan, the official name key is added (31 Ice Cream). Moreover, the name tag contains the global franchise mark (in "name:ja" tag is the official name in japanese). More info in [Wikipedia](https://ja.wikipedia.org/wiki/%E3%83%90%E3%82%B9%E3%82%AD%E3%83%B3%E3%83%BB%E3%83%AD%E3%83%93%E3%83%B3%E3%82%B9).
- About Tokyu Hands: I run "npm build" and its position is changed.
- About CoCo (japanese brand): I added _"countryCodes": ["jp","cn"],_. 
- About Pronto. I sure what "Pronto" is in official_name key, from italian-language, and goes to "name" key, too. The japanese name is "プロント" but is for easy pronunciation for Japanese people, so it goes to "name:ja" key. See [website](https://www.pronto.co.jp/global/).

PR #2375, opened 2019-03-03
More Japanese Brands (part 3)

sguinetti:
This is a third pull request, the previous pulls were #2365 and #2367. Add business brands about cinema, banks, restaurants, convenience stores and clothes. Thanks.

- Also, Added Hush Puppies brand
- About 7-Eleven: Added alternative typos with "-"
- About Nishimatsuya Chain, Hotto Motto and Pronto: Add NoMatch.
- About Ito-Yokado. Add Match (writed in Wikipedia Article)
- There are different numbers in "count" from this pull request, build from npm.

### DIFF
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -25390,6 +25390,7 @@
       "brand:wikipedia": "ja:ローソン・スリーエフ",
       "name": "ローソン・スリーエフ",
       "name:en": "Lawson･Three F",
+      "name:ko": "로손·쓰리에프",
       "shop": "convenience",
       "operator": "株式会社エル・ティーエフ",
       "operator:en": "L･TF Co., Ltd."
--- a/config/canonical.json
+++ b/config/canonical.json
@@ -7070,6 +7070,24 @@
       "operator:wikipedia": "en:NRG Energy"
     }
   },
+  "amenity/cinema|109シネマズ": {
+    "countryCodes": ["jp"],
+    "nocount": true,
+    "tags": {
+      "amenity": "cinema",
+      "brand": "109シネマズ",
+      "brand:en": "109 Cinemas",
+      "brand:wikidata": "Q10854269",
+      "brand:wikipedia": "ja:109シネマズ",
+      "name": "109シネマズ",
+      "name:en": "109 Cinemas",
+      "name:ko": "109 시네마즈",
+      "operator": "株式会社 東急レクリエーション",
+      "operator:en": "Tokyu Recreation Co., Ltd.",
+      "operator:wikipedia": "ja:東急レクリエーション",
+      "operator:wikidata": "Q11527011"
+    }
+  },
   "amenity/cinema|Cinema City": {
     "count": 64,
     "tags": {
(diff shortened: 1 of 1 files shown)


Adamant36:
Whoever sees this might want to wait until #2243 is finished to merge it. As its currently being worked on. 

@sguinetti, With the McDonald's it should only have the name in the language the entry is for and English. So you need to get rid of either ko or zh. Along with two of the name entries for Bank of Japan and any other ones that aren't just the entry language and English, like LAWSON, and on the last entry id get rid of operator:en because the Wikipedia page is in English and that's good enough IMHO. 

Also, it would it be good to split your contributions into smaller PRs. 292 lines in 11 commits is a lot to go over and make sure is correct (as is evidenced by the merge conflict).

sguinetti:
@Adamant36 Yes, I notice it again. I'm sorry again. I was able to divide into 3 parts, but I had to reduce more. In each brand, I added a lot of information.

Otherwise, I saw in Openstreetmap that they added the name: ko key to the Lawson brand (via Overpass Turbo). So, add the translated name could help to other visitors in the translated name without problems.

When you say "last entry", are you mean Lawson-Three 3?

Adamant36:
YorkMart for the last entry. 

As far as the name goes, people can add whatever names they want in OSM. See https://github.com/osmlab/name-suggestion-index/pull/2365#issuecomment-468703966 though and also the last paragraph in https://github.com/osmlab/name-suggestion-index/pull/2222#issuecomment-450438029

To quote it "So far we've just been doing name - and we add name:en only if name is not readable in English. I don't think it would be really helpful to add more than those, and it could get messy if we start automatically adding 100 more tags for international names, so let's not."

sguinetti:
@Adamant36 Right. I will follow that point: Keep the names of these brands in English and Japanese only.

sguinetti:
I was wrong. The commit 27b9de3  says "Change wikipedia article to japanese, and keed title in english".

bhousel:
> Whoever sees this might want to wait until #2243 is finished to merge it. As its currently being worked on.

It's ok - I did #2243 it in a way that we can keep `canonical.json` around until everything is merged. I'll finally remove it today and update the docs.

bhousel:
@sguinetti  I noticed these entries:

```js
  "shop/baby_goods|西松屋": {
    "count": 67,
    "countryCodes": ["jp"],
    "nomatch": ["shop/clothes|西松屋"],
    "tags": {
      "brand": "西松屋",
      "brand:en": "Nishimatsuya Chain",
      "brand:wikidata": "Q11628761",
      "brand:wikipedia": "ja:西松屋",
      "name": "西松屋",
      "name:en": "Nishimatsuya Chain",
      "shop": "baby_goods"
    }
  },
...
  "shop/clothes|西松屋": {
    "count": 159,
    "nomatch": ["shop/baby_goods|西松屋"],
    "tags": {
      "brand": "西松屋",
      "brand:en": "Nishimatsuya Chain",
      "brand:wikidata": "Q11628761",
      "brand:wikipedia": "ja:西松屋",
      "clothes": "children",
      "name": "西松屋",
      "shop": "clothes"
    }
  },
```

As specified here, it looks like there is a single brand that operates both generic clothing stores and also separate baby good stores. 

If they are really different, it's ok to keep them as 2 entries like this and "nomatch" each other.

But if they are really the same thing, we should pick a single entry (probably the baby goods one) and "match" the less desired entry.   

Think about it and let me know whether you want me to change it.. thanks!

bhousel:
@sinclarius  - can you check these entries for "西松屋"  ?
We're not sure whether it should be merged into a single entry, or remain as two.

sinclarius:
@bhousel Sure thing, I'll try to get to it today. Sorry—totally missed this message last week!

sinclarius:
I'm not personally familiar with any Nishimatsuya storefronts, but I think it's safe to merge it into a single entry. From the company profile page (https://www.24028.jp/company/outline) and the executive's message page (https://www.24028.jp/company), it looks like it's **both** a baby and children's clothing store ("ベビー・子どものくらし用品専門店チェーン"). I looked at a couple of their listed stores, and I found a few cases where they noted some specific baby items that they don't carry. However, the circular ads for those stores still showed a lot of items for both babies and children (https://www.nishimatsuya.com/flier/chirashir_0228_0319/html5.html#page=1). Hope that helps!

bhousel:
Thanks for checking @sinclarius - It sounds like they sell mostly baby things, so I merged them into the `shop=baby_goods` entry in e60767e.

PR #2379, opened 2019-03-05
Add more petstores and countrycodes for several existing stores

tas50:
Add some major US pet stores and add in country codes for several of the US based stores already there.

### DIFF
--- a/brands/shop/pet.json
+++ b/brands/shop/pet.json
@@ -41,6 +41,7 @@
     }
   },
   "shop/pet|Mud Bay": {
+    "countryCodes": ["us"],
     "nocount": true,
     "tags": {
       "brand": "Mud Bay",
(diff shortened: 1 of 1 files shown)


Adamant36:
I'm not sure how much it matters, but in 99% of the cases the country is below the tag count. It seems to look better that way if nothing else. In this case you have it both above and below it. Maybe it would be good to pick one way of doing it for the sake of consistency (hopefully below it).

bhousel:
> I'm not sure how much it matters, but in 99% of the cases the country is below the tag count. It seems to look better that way if nothing else. In this case you have it both above and below it. Maybe it would be good to pick one way of doing it for the sake of consistency (hopefully below it).

It doesn't really matter.. The properties all should get sorted alphabetically when the build script runs.

bhousel:
Thanks @tas50 looks good 👍

Adamant36:
> It doesn't really matter.. The properties all should get sorted alphabetically when the build script runs.

Javascript and NPM sure are fancy. I'm frequently surprised by how much they can automate.

PR #2383, opened 2019-03-06
fixed issue 1608

abdiG:
added brand:wikidata, brand:wikidata, and countryCode for shop/shoes|FLO

### DIFF
--- a/brands/shop/shoes.json
+++ b/brands/shop/shoes.json
@@ -171,7 +171,12 @@
   },
   "shop/shoes|FLO": {
     "count": 51,
+    "countryCodes": [
+      "USA"
+    ],
     "tags": {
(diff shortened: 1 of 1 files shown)


bhousel:
Sorry, but these changes don't look right..

The wikipedia and wikidata links point to a [television sitcom](https://en.wikipedia.org/wiki/Flo)

Also, the countryCode should be the 2-letter code like `us`, not `USA`. Either way, this is wrong, because FLO the shoe store looks like a [Turkish retailer](https://overpass-turbo.eu/s/GGb).

Anyway, thanks for the PR, but it needs to be changed before we can accept it.

Adamant36:
Related to #2383

I created a Wikidata entry for it. Its Q61994802

Adamant36:
Fixed in #2417

PR #2384, opened 2019-03-06
fixed-Issue 1614

abdiG:


Adamant36:
Hi there. Good find. I looked for that one a couple of times on Wikipedia and didn't see it. Btw, if you put "closes" and the issue number in your PR message Github will auto close it when it when it's merged so someone else doesn't have to later. Thanks.

PR #2387, opened 2019-03-06
Add operators for banks

Adamant36:


tas50:
Maybe this is a more fundamental OSM question that this simple repo, but it seems like we're adding a lot of static and redundant tags for information that is actually pretty fluid in the real world. The wikidata entry if done right includes the up to date information on the parent company for any brand or company. This seems like a far better place to pull this data from rather than tags, which can't be updated easily. We're going to really quickly end up with a pile of incorrect parent company information once a company is sold vs. a single place where that data can be updated.

Adamant36:
Yeah I was thinking about that myself. There's some entries were the Wikipedia pages have said the companies gone out of business and I wonder what the value of adding them like their legitimate is. So there's a lot of chances for things to go stale. I guess it just comes down to good maintenance later. The reason I added a lot of the operators in these cases is the hope that it keeps people adding the operator information where the brand information should be. Which has happened a lot in the past. Especially with none English companies and banks in particular.

tas50:
I'll admit I misused operator quite a bit in some of my early edits. It's not understood by new editors.

Adamant36:
> It's not understood by new editors.

Yeah, totally not. I did it myself a couple of times.

SilentSpike:
Yeah OSM has an inherent problem in that without use of either automation or a data structure where you can have instances of a single entity definition you're almost never going to have full coverage of up-to-date tagging for something as simple as a brand which always has the same tags.

It doesn't necessarily matter though, because as long as the wikidata tag is present and accurate then data consumers can always query wikidata from the tag and use the more up-to-date information from there. Plus if wikidata were to disappear for whatever reason, there would still be data in OSM that's hopefully mostly accurate.

Adamant36:
I'm hopping there will at least be some rudimentary automation of this kind of stuff at some point. I don't see why something like switching the values of store brands couldn't be done responsibly with ai and someone higher up over seeing it. Id like to think we are doing the ground work for it here. The main reason I got into this was to improve name tagging quality. I don't think OSM will ever have the user base for good store mapping quality at scale, plus editors are to sillowed in their particular mapping niches for any kind of major group effort to work, but having something like this index and the quality control checks in iD Editor are good starts in that direction at least. Its not like any map is 100% accurate anyway.

PR #2389, opened 2019-03-06
Add Pottery Barn and Williams-Sonoma

tas50:
Also add some country code information while I'm here.


matkoniecz:
> "name": "PWilliams-Sonoma",

Is "P" in the name intentional?

tas50:
Nope

PR #2390, opened 2019-03-06
Add Blaze Pizza

tas50:
I've been seeing this all over US shopping centers recently.

### DIFF
--- a/brands/amenity/restaurant.json
+++ b/brands/amenity/restaurant.json
@@ -93,8 +93,19 @@
   },
   "amenity/restaurant|Blaze Pizza": {
     "count": 59,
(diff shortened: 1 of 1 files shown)


Adamant36:
What do you think about the country codes going left to right for multiple entries versus down? Down seems like a lot of extra lines to me.

tas50:
The advantage of down is diffs are clean as they open new locations in other countries. It also matches a lot of the other entries we have which go down. Github and git in general do a pretty awful job of showing long line diffs.

Adamant36:
That's kind of what I was thinking. Thought id ask though. Five entries isn't that bad either.

PR #2394, opened 2019-03-07
Update clothes.json

PrimeN:
PR for:[ Issue #1212: Research needed for: shop/clothes|Street One ](https://github.com/osmlab/name-suggestion-index/issues/1212)


tas50:
@PrimeN Can you run npm run build to make sure the ordering of the values in the json is correct. Also can you add a countrycode here after determining where these stores are located. At some point we'll be utilizing that data and it'll be great to have it here.

Adamant36:
I don't think its worth adding the country codes if there is that many. Also, the Wikipedia and Wikidata entries should be for the operator. Not the the brand. Ultimately, this brand needs a Wikidata page created for it, but the operator information shouldn't be a substitute in the meantime.

tas50:
I'm not sure at what point we say it's too many countries to list. @bhousel should say for sure but I'd imagine that whenever the country codes get shipped and utilized by editors like ID the absence of codes would be interpreted as all countries. That's probably ideal for a store that's located all through Europe or worldwide. I wouldn't want to add 100 entries for Starbucks or McDonald's when we can just assume they're virtually everywhere.

Adamant36:
I tend to lean toward a company being a "global" brand if they are either on multiple continents or at least in a bunch of different countries on the same continents. Like if a brand is in almost every European country, Russia, and China, I'd just call it good at that point and not list them individually. 

In this case it seems like 99% of them are European countries that are probably each the size of Oregon. So I wouldn't list them all here either. Plus, isn't the point in country codes in the first place to filter out results for similar things in countries where the company is or isn't? 

I doubt anyone in North America are going to be looking for a similar tag and mistake it for this one, so it doesn't need to be filtered out for them. Whereas, people in Europe who will be tagging the stores already know its there, so they don't need it filtered in for them. Since it will come up when they search for it anyway. Unless I'm missing the point of country codes.

Adamant36:
Btw @tas50, do you know what's up with your "Dedupe the mobile money entries" branch is? I couldn't find any reference to it in the PR list. If its already been merged can you delete it please?

tas50:
I'll just nuke that. I was going to collapse 2 entries into one, but I think both can just get nuked potentially. I need to research that one.

bhousel:
What's weird about the these country codes is that I don't think this brand actually is in all those places. 
 @PrimeN why did you add them this way?

[Overpass query](https://overpass-turbo.eu/s/GJE)
screenshot 2019-03-07 07 24 08

And the [wikipedia page](https://de.wikipedia.org/wiki/CBR_Fashion_Holding) says that they are based [around Germany only](https://de.wikipedia.org/wiki/CBR_Fashion_Holding#/media/File:StreetOne_Lageplan.jpg)


Even if it were correct, I also wouldn't add so many country codes - mostly because it would mean that the brand is expanding widely, and that list might be out of date soon.  For example there is no `ru` in the list, but if they open a store in Russia then we'd have to change our list.  Once they have expanded beyond a few countries, I don't see it as worth scoping it.

PrimeN:
@bhousel - In wikipedia it says they're based in germany but I double checked their website and according to their store locator they have stores in all of those countries. But what you're saying makes sense, I'll go ahead and remove all the other countries aside from germany. As far as wikidata goes since this company doesn't have a page I'll go ahead and see if I can create one and point the new wikidata link to my submission.

bhousel:
> and according to their store locator they have stores in all of those countries.

In that case it's ok to leave all the countries.  Or just leave out the countries entirely if there are so many of them.

tas50:
Thanks @PrimeN

Adamant36:
I would have gotten rid of them. Who knows how many stores they actually have in those countries or how stable they are. Talk about the risk of data rot.

Adamant36:
Btw tas50, he didn't fix the brand:wikipedia tag issue. Any chance you could do it so its not forgotten about? Since you merged it.

tas50:
@Adamant36 Fixed on master

Adamant36:
Thanks

PR #2397, opened 2019-03-08
Update allnames.json w/o man_made and filter out / match additional brands

tas50:
Update allnames.json with a planet file from today
Skip processing man_made tags since we're not getting anything from this and it adds to the already lengthy time the whole process takes to run

Update all the brand counts an in doing so new brands were added and several old ones left as they dropped below 50. For the old ones I either added a match or added the brand with data.

- Spar Express dropped off as a supermarket since it's really a convenience store. I'm adding this as a match for the entry we already have.
- The electronics shop Mediamarkt dropped off, but I created another match for the proper case + spacing of the name, which is what this entry should have been the whole time
- The bank Banco Union was added as another match for Banco Unión
- The bank Caisse d'épargne was added as a match for Caisse d'Épargne
- Chata Polska dropped off the list, but I created a wikidata page for it and added those details
- Johnny Rockets dropped off so I added wikidata. These are US style diner restaurants all over the world
- Pret a Manger was added as a match to Pret A Manger with a capital A which is the correct variant.
- Кулинария convenience stores and cafes have been removed as it was a generic and it dropped off the list for cafes. It means "cooking" in Russian.
- GPL fuel dropped off, but these look like stations selling gas di petrolio liquefatto (aka LPG) so that's a generic
- La Mie Caline dropped below 50 and became a match on La Mie Câline
- Ulta dropped below 50 and became a match on Ulta Beauty


bhousel:
Thanks for running this!  I appreciate you taking the time to do it and to summarize the results 👍 

I wonder if I should also periodically re-run the script that creates the "Needs Research" issues on GitHub.  Maybe after we close all the existing ones?

Adamant36:
> I wonder if I should also periodically re-run the script that creates the "Needs Research" issues on GitHub. Maybe after we close all the existing ones?

Id say wait until we get done closing the exiting ones or at least most of them (some of them, like the odd Russian supermarkets, probably aren't closeable). The number of open issues has been dropping pretty quickly lately and I'm in the process of creating Wikidata entries for some of the ones that need them. So it should't be to long until there aren't that many open issues left.

tas50:
Seems like we're fast approaching the end of the unclassified brands in the dataset. After that's done when we regenerate the data and get new entries it would probably be a good idea to run that script. It seems to work to get folks to jump in and pick off unclassified brands.

PR #2399, opened 2019-03-08
Add data for Retro Fitness and Basic-Fit

tas50:
Add data to those so we can get them in iD.


tas50:
Close #961

Adamant36:
apparently it needs to be "closes" to work. It seems like its picker about the keywords then it should be.

PR #2402, opened 2019-03-08
Add amenity/car_sharing

tas50:
Another handy set of brands.


brands/amenity/car_sharing.json
```diff
@@ -0,0 +1,48 @@
...
+  "amenity/car_sharing|stadtmobil": {
+    "match": [
+      "amenity/car_sharing|stadtmobil CarSharing-Station"
+    ],
+    "tags": {
+      "amenity": "car_sharing",
+      "brand": "stadtmobil",
+      "brand:wikidata": "Q2327629",
+      "brand:wikipedia": "en:Stadtmobil",
+      "name": "stadtmobil"
```
ENT8R: Should not this name start with a capital letter? And the `brand` too?

tas50: Based on their signage and website they use an all lowercase name.

ENT8R: But I thought I read somewhere that all names in OSM shoudl start with a capital letter. I just can't find the source anymore... Maybe it was also some kind of Osmose error, does anybody else know more?

bhousel: In OSM the name and brand should be as-signed. It's totally ok for them to start with a lowercase letter. Thanks for confirming @tas50

PR #2403, opened 2019-03-08
add country codes to the distributed file

### LINKED ISSUE #2274: countryCodes lost in distribution

matkoniecz:
fixes #2274 

As this is change to how data is processed, not to the data itself I am opening a PR.

Note that since files where country codes are located are now split into multiple separate ones it become more important to make this data available to whoever is using it (for example Vespucci).


bhousel:
It's probably ok.. Pretty soon I will switch iD away from using `dist/name-suggestions.json` and just use the raw files under `brands/` instead.

I think Vespucci might be the only other project that consumes this file.  It's probably ok to add a `countryCodes` property, but tagging @simonpoole as a heads up.

PR #2413, opened 2019-03-09
k kiosk: added countryCodes and nomatch

SelfishSeahorse:
* k kiosk can also be a newsagent
* Removed link to Italian Wikipedia, as also/mostly used in German speaking region

### DIFF
--- a/brands/shop/kiosk.json
+++ b/brands/shop/kiosk.json
@@ -72,6 +72,8 @@
   },
   "shop/kiosk|k kiosk": {
     "count": 64,
+    "countryCodes": ["ch"],
+    "nomatch": ["shop/newsagent|k kiosk"],
     "match": [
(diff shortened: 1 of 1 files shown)


Adamant36:
Hhmm, a link to the Italian Wikipedia page is better then no Wikipedia page. Especially since its the only one and browsers translate things now. Also, wouldn't the news agent be a match if its the same thing instead of a no match?

SelfishSeahorse:
> Hhmm, a link to the Italian Wikipedia page is better then no Wikipedia page. Especially since its the only one and browsers translate things now.

Okay, i've re-added it.

> Also, wouldn't the news agent be a match if its the same thing instead of a no match?

Not sure if i understand you correctly, but "k kiosk" is a brand of both kiosks and newsagents.

Issue #2245, opened 2019-01-05
We should add local `name:xx` tags


bhousel:
In working with KeepRight data a lot lately, I notice that it sometimes flags a warning like this:

> It would be nice if this node had an additional tag 'name:XX=Chick-fil-A' where XX shows the language of its name 'Chick-fil-A'.

<img width="300" alt="screenshot 2019-01-05 01 30 33" src="https://user-images.githubusercontent.com/38784/50721201-820d8980-1089-11e9-92f9-fcad5e872c97.png">

It seems like the warning is caused if someone adds a `name:xx` tag and a `name` and they don't match. 

<img width="300" alt="screenshot 2019-01-05 01 32 22" src="https://user-images.githubusercontent.com/38784/50721218-d57fd780-1089-11e9-885c-49e3bb3ed1f4.png">
<img width="300" alt="screenshot 2019-01-05 01 32 27" src="https://user-images.githubusercontent.com/38784/50721220-d9abf500-1089-11e9-8c0c-4d534d64d115.png">

It is only a warning , but  I can understand why it would be nice in this case to add a `name:en` tag, so that people know what language `name` is in or where it came from.

We have a lot of names like this in the index.  Our Contributing guidelines even encourage it:

```js
  "amenity/fast_food|マクドナルド": {         // Identifier like "key/value|name"
    "count": 1445,
    "countryCodes": ["jp"],                 // Optional `countryCodes` - array of countries where this entry is used
    "tags": {
      "amenity": "fast_food",
      "brand": "マクドナルド",                // `brand` - Brand name in the local language (Japanese)
      "brand:en": "McDonald's",             // `brand:en` - For non-English brands, tag the English version too
      "brand:wikidata": "Q38076",           // `brand:wikidata` - Same Universal wikidata identifier
      "brand:wikipedia": "ja:マクドナルド",   // `brand:wikipedia` - Reference to Japanese Wikipedia
      "cuisine": "burger",
      "name": "マクドナルド",                 // `name` - Display name, also in the local language (Japanese)
      "name:en": "McDonald's"               // `name:en` - For non-English names, tag the English version too
     // SHOULD WE ADD "name:ja" TOO??
    }
  },
```

SilentSpike:
I can't think of any strong argument against it 👍

matkoniecz:
from what I see there is no pending issue to remove this complaint in https://github.com/keepright/keepright/issues (I checked as Keep Right is no longer developed).

[...]

PR #2417, opened 2019-03-10
Multiple additions on shoes.json

coderbeetle:
- ABCマート: Added two more country codes, Wikidata code, and Wikipedia link.
- Besson Chaussures: Added country code. A Wikipedia shop for this page does not exist, but I managed to find their operator information - so I added these too.
- FLO: Added country codes. Wikipedia page doesn't exist.
- Gabor: Added country codes, Wikidata code, and Wikipedia link.

### DIFF
--- a/brands/shop/shoes.json
+++ b/brands/shop/shoes.json
@@ -191,6 +191,7 @@
     "countryCodes": ["tr", "al"],
     "tags": {
       "brand": "FLO",
+      "brand:wikidata": "Q61994802",
       "name": "FLO",
       "shop": "shoes"
     }
--- a/brands/shop/shoes.json
+++ b/brands/shop/shoes.json
@@ -1,10 +1,12 @@
 {
   "shop/shoes|ABCマート": {
     "count": 51,
-    "countryCodes": ["jp"],
+    "countryCodes": ["jp", "kr", "tw"],
     "tags": {
       "brand": "ABCマート",
       "brand:ja": "ABCマート",
+      "brand:wikidata": "Q11188787",
+      "brand:wikipedia": "en:ABC-Mart",
       "name": "ABCマート",
       "name:ja": "ABCマート",
       "shop": "shoes"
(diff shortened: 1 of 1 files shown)


Adamant36:
I created a Wikidata entry for FLO to deal with #2383, but it ended up getting hung up. So if you want to add it this PR that would be great. Then I can close the other one. Its Q61994802

coderbeetle:
Done!

Adamant36:
Thanks. Two more minor things. In general we only add `name:en`  in cases where the `name:` tag is in another language, but we don't add other values like `name:jp` or anything. Since its essentially a duplicate of the `name:` key and we don't want to add to many name values to database (hopefully that's clear. If not I can provide some issue numbers that discuss it better for you to read). 

also, if you run npm run build before doing the commit it will check for any errors and get rid of the trailing space at the end of the file. Its not necessary, but helpful if you add more then a few lines of code.

coderbeetle:
Thanks for letting me know! I think the duplicate names were already in the code, as far as I can recall I didn't add any names in other languages in addition to the English ones. Should I remove values like `name:ja` in the future?
As for the `npm run build`, I've been actually pretty confused as on how to run it. I'm assuming I'm supposed to run it on Git, but I haven't really been able to find any useful references on that.

Adamant36:
No, its cool. They can stay in there if they were already added. I totally thought you had added them. Sorry. At least you know about it for the future now though. 

As far as `npm run build`, you run it in the terminal before you start editing the .json files and after you save your edits to commit them. The contributing file could probably be more clear about it. If you don't understand something, feel free to open a issue about it or just ask.

bhousel:
> Thanks. Two more minor things. In general we only add `name:en` in cases where the `name:` tag is in another language, but we don't add other values like `name:jp` or anything.

No - we _do_ add tags like `name:jp` now to placate the various OSM validator tools.  See #2245

The build script will even add them automatically for languages like Japanese, where we can tell with a regex what language the name is in.

Adamant36:
Oh yeah, my bad. I forgot the policy changed.

PR #2421, opened 2019-03-11
add bank info

Adamant36:


bhousel:
Thanks! .. 
per #2245, if you are going to add `name:en`, `brand:en`, `operator:en`,  make sure you also add localized tags that match the other too... e.g. `name:es`, `brand:es`, `operator:es`.

Adamant36:
My bad. I totally spaced it again. I'll go through it and add them.

bhousel:
> My bad. I totally spaced it again. I'll go through it and add them.

No problem.. I'm going to merge as-is, and you can add them later when you have time 👍

PR #2424, opened 2019-03-12
add languages

Adamant36:
@bhousel, here you go. I think that covers it. I'd appreciate it if you gave it a quick glance over.


bhousel:
> I'd appreciate it if you gave it a quick glance over.

Looks good, thank you! 👍

PR #2425, opened 2019-03-12
Add Jared and fix Kay

tas50:
Kay and Jared are both brands run by Sterling. I created brand wikidata pages for those and moved Sterling over to the operator since they run these and independent brands that give the illusion of false choice.


bhousel:
> I created brand wikidata pages for those and moved Sterling over to the operator since they run these and independent brands that give the illusion of false choice.

Haha nice..  Harder to break up the 💎monopoly that way I guess..

PR #2427, opened 2019-03-12
Add Which Wich?

tas50:
I didn't add a country code as this is located all over the world (but somehow not on the west coast of the US).

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -1947,6 +1947,17 @@
       "name": "Whataburger"
     }
   },
+  "amenity/fast_food|Which Wich?": {
+    "nocount": true,
+    "tags": {
+      "amenity": "fast_food",
+      "brand": "Which Wich?",
+      "brand:wikidata": "Q7993556",
+      "brand:wikipedia": "en:Which Wich?",
+      "cuisine": "sandwich",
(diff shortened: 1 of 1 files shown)


matkoniecz:
Version of name without quotation mark is also quite popular - http://overpass-turbo.eu/s/GT3

I would consider adding it as `alt_name` or to `"match"` property.

Also, which one is typically used as name? Maybe version with quotation mark should end in `official_name`?

Adamant36:
I'd go with a match personally. Alt_name is to vague and none official for something like an index of companies in my opinion. Names should be kept to things that are either shortened versions of official names or official ones. But we shouldn't go down the road of introducing "locally preferred names," whatever that means. At that point its just bad name tagging that should be fixed.

matkoniecz:
Of "Which Wich" is used by people with similar popularity as "Which Wich?" then it is not important whatever it is official or not.

bhousel:
Sounds good, I merged and added a match for the name without the question mark in c574b08

tas50:
I'm slowly creating MapRoulette challenges for all of these incorrectly tagged stores. At some point I'll get one for the stores without the ? here and we can get them all cleaned up using the preset. I'm mostly just waiting on the next iD release to pull in the last 50 brands we've probably added in the last week.

PR #2428, opened 2019-03-12
Add PostalAnnex

tas50:
Shipping / post office box company in the US.

### DIFF
--- a/brands/amenity/post_office.json
+++ b/brands/amenity/post_office.json
@@ -272,6 +272,17 @@
       "name": "Poczta Polska"
     }
   },
+  "amenity/post_office|PostalAnnex": {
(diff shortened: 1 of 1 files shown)


matkoniecz:
This is mapped just 3 times at this moment :)

bhousel:
> This is mapped just 3 times at this moment :)

Awesome!  I love getting things into this index early, so things can be mapped consistently when people add them..

tas50:
I'm really confused why this one isn't mapped more because they're fairly common on the west coast at least.

tas50:
Oh it's because they're all mapped wrong. There's 37 of them, but most have a space between the two words.

Issue #2235, opened 2019-01-01
Rite Aid and CVS (pharmacies) missing.


BjornRasmussen:
Both Rite aid and CVS pharmacies are missing from the Name Suggestion Index.  They should have the following:  
  
Rite Aid:
==
  "amenity/pharmacy|Rite Aid": {
    "countryCodes": ["us"],
    "tags": {
      "amenity": "pharmacy",
      "brand": "Rite Aid",
      "brand:wikidata": "Q3433273",
      "brand:wikipedia": "en:Rite Aid",
      "healthcare": "pharmacy",
      "name": "Rite Aid"
    }
  }  
  
  CVS Pharmacy  
==
  "amenity/pharmacy|CVS Pharmacy": {
    "countryCodes": ["us"],
    "tags": {
      "amenity": "pharmacy",
      "brand": "CVS Pharmacy",
      "brand:wikidata": "Q2078880",
      "brand:wikipedia": "en:CVS Pharmacy",
      "healthcare": "pharmacy",
      "name": "CVS Pharmacy"
    }
  }   
  
I hope this is helpful.

1ec5:
Note that some CVS locations omit the “Pharmacy” from signage because they don’t offer pharmacy services. However, I suppose some pharmacy locations also omit “Pharmacy” for space reasons.

Adamant36:
@1ec5, thanks for bringing it up. I mentioned it in another issue, but I totally forgot to refrence it here. 

Maybe they should be two different entries. One as a convience store and the other as a pharmacy. Although the fact that they omit the word "pharmacy" from some signs for space reasons even if they provide pharmacy services kind of muddies things a little.

BjornRasmussen:
So, should CVS have the name and brand "CVS"?  I am not sure about additional tagging, since only some CVS locations have pharmacies.  Perhaps using convenience store on all would be a good idea.  Considering that CVS sells all sorts of junk food like Doritos and Pop tarts (to make you sick?), I would say that convenience store is quite accurate.

1ec5:
There’s some other discussion about drugstores in openstreetmap/iD#4127. Some mappers (and iD) prefer to indicate the overall store as a `shop=chemist` and only the pharmacy desk as an `amenity=pharmacy`. (The non-pharmacy CVS locations would be `shop=chemist` only.) Encoding this style in presets is problematic, though: ordinary people don’t think of the store as two POIs, so they leave out one or the other unintentionally.

[...]

PR #2430, opened 2019-03-12
Fold CVS / Rite Aid and Walgreens all into Pharamacy

tas50:
This might be a bit controversial, but I'm fairly certain this is the right change and we never had wikidata on these old entries so they were never in the dist file.

These stores are all locations where you can buy random toiletries as well as prescription drugs as they run a full drug counter. In a lot of places like Target they're just the pharmacy counter without any of the convenience store like items. Since these stores dispense prescription drugs they can't be a chemist per the OSM wiki entry for chemist:

"A shop typically selling a similar range of articles to a dispensing [W] community pharmacy: [W] over-the-counter drugs, personal hygiene (toothpastes, shampoos, soaps etc.), cosmetics (usually cheaper ones), and household cleaning products. The exact product mix will vary from country to country. "

### DIFF
--- a/brands/amenity/pharmacy.json
+++ b/brands/amenity/pharmacy.json
@@ -153,9 +153,9 @@
       "amenity/pharmacy|CVS Pharmacy",
       "amenity/pharmacy|CVS/Pharmacy",
       "amenity/pharmacy|CVS/pharmacy",
+      "shop/chemist|CVS",
       "shop/convenience|CVS"
     ],
-    "nomatch": ["shop/chemist|CVS"],
     "tags": {
       "amenity": "pharmacy",
       "brand": "CVS",
@@ -971,9 +971,10 @@
     },
     "match": [
       "amenity/pharmacy|Rite Aid Pharmacy",
-      "amenity/pharmacy|Rite-Aid"
+      "amenity/pharmacy|Rite-Aid",
+      "shop/chemist|Rite Aid",
+      "shop/convenience|Rite Aid"
     ],
-    "nomatch": ["shop/chemist|Rite Aid"],
     "tags": {
       "amenity": "pharmacy",
       "brand": "Rite Aid",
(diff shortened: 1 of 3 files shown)


Adamant36:
The question is how many of them are mapped as nodes in the back of the building where the pharmacy is versus the complete store. I've seen it done both ways before. So when they are mapped as nodes they are chemists, and when not, they probably aren't (you don't usually buy toilet paper from the pharmacist. You take it to the front counter. So, although the stores have those items, the chemists are essentially separate entities. In some places they are branded differently and just license the space also. Although I think all these are the same brand). 

I think the important one to keep separate if I remember right from an earlier conversation was CVS because their stores randomly do or don't have pharmacies. So you can't assume. Whereas, all Walgreen's and Rite Aid do, but people still map them separate sometimes. Often because they have different hours then the main store (something that combining them would potentially mess up).

Adamant36:
They also often have different contact details.

Btw, CVS and CVS Pharmacy are separate brands with different store layouts. So CVS isn't just a shortened name for CVS Pharmacy. Here's the issues discussing it #2235 and #2236

bhousel:
Ah yeah, this pharmacy/chemist thing is kind of an old issue going back several years. See 
 https://github.com/osmlab/name-suggestion-index/issues/30#issuecomment-451360887 for the latest.

Basically we let people map the main store as a `shop=chemist` and the pharmacy counter as an `amenity=pharmacy` for several reason:
- they have different `opening_hours`, contact info, management, license, etc.
- occasionally a CVS / Walgreens / etc does not have a pharmacy counter
- occasionally a CVS / Walgreens / etc operates a pharmacy counter without a storefront
- It makes sense too for places like Walmart and Costco that sometimes have a pharmacy inside them
- Things work differently in other parts of the world too, so we don't assume

Sorry for the confusion.. It would be great for someone to update the OSM wiki with this info so we can move on.

PR #2431, opened 2019-03-12
Add wikidata for cost cutters and add more country codes

tas50:
Slowly filling things in

### DIFF
--- a/brands/shop/hairdresser.json
+++ b/brands/shop/hairdresser.json
@@ -17,8 +17,10 @@
   },
   "shop/hairdresser|Cost Cutters": {
     "count": 86,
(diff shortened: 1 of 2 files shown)


Adamant36:
Closes #1489

matkoniecz:
> you should add Twitter and Facebook accounts to any Wikidata entries you create

I would phrase it "you can if you consider it useful and have time to do that". Certainly not "should".

Adamant36:
@matkoniecz, I meant "should" in the general sense. Obviously what people do or not is optional. Btw, since your around and doing things, would you mind reviewing and merging my PR?

PR #2433, opened 2019-03-13
Add official_names for restaurants and fix Popeyes

tas50:
I added the full names for various american fast food and restaurant chains to aid in searches. These are based off of their official websites and wikipedia pages.

I also corrected the name of Popeyes. It's not Popeye's despite how people often spell it. Check the website: https://popeyes.com/

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -155,7 +155,8 @@
       "brand:wikidata": "Q891163",
       "brand:wikipedia": "en:Bojangles' Famous Chicken 'n Biscuits",
       "cuisine": "chicken",
-      "name": "Bojangles'"
+      "name": "Bojangles'",
+      "official_name": "Bojangles' Famous Chicken 'n Biscuits"
     }
   },
   "amenity/fast_food|Booster Juice": {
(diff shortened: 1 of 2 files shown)


Adamant36:
There's over three times as many uses of Bojangles' spelled Bojangles. I'm not going to speculate on which one is correct, except to say that Bojangles' sounds weird, but it might be worth making Bojangles the main entry since its clearly preferred by taggers.

tas50:
I didn't actually change the Bojangles' name. I just added a comma there to support the official name entry. The name has an apostrophe in it, but clearly a lot of people don't think it does. The signs have it, wikipedia has it, and their website has it. That's been my measure for getting names correct. People on OSM are pretty awful at getting strange names right so I try to avoid their using their typos as a measure with overpass counts. There's weird tagging out there. I've been fixing a lot of Chuck E Cheeses restaurants tagged as childcare facilities recently.

Adamant36:
"People on OSM are pretty awful at getting strange names right" 

Your right about that. For every brand name in the database a bunch of screwed ways to spell it, that usually only has one entry in the database each. Its pretty horrible. 

"I've been fixing a lot of Chuck E Cheeses restaurants tagged as childcare facilities recently"

Totally not what Chuck E Cheese is for. I wonder if it was from an import or or group mapping. I can't imagine multiple people making the same obvious mistake.

PR #2434, opened 2019-03-13
Add common missing restaurants

tas50:
These were the brands we were missing from a list of the top 100 US restaurant chains by revenue.

### DIFF
--- a/brands/amenity/restaurant.json
+++ b/brands/amenity/restaurant.json
@@ -71,6 +71,19 @@
       "name": "Autogrill"
     }
   },
+  "amenity/restaurant|BJ's": {
+    "countryCodes": ["us"],
+    "nocount": true,
+    "tags": {
+      "amenity": "restaurant",
+      "brand": "BJ's",
+      "brand:wikidata": "Q4835755",
+      "brand:wikipedia": "en:BJ's Restaurants",
+      "cuisine": "american",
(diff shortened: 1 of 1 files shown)


Adamant36:
For BJ's there's 48 uses of BJ's Restaurant & Brewhouse and 23 uses of BJ's Restaurant and Brewhouse (along with 15 uses of BJ's Brewhouse). With Cheddar's there is 24 uses of Cheddar's Scratch Kitchen. So you might want to add them as matches while your here. 

For On the Border, there's more uses as On The Border and it seems to be how they officially spell it. So you might want to make the lower case instance a match instead of the main one. Its better to go with how the company spells it.

tas50:
Yeah it looks like other than the title of their website they're (incorrectly) capitalizing the T. I'll get that fixed. I can switch BJ's over too. It looks like they've been moving away from the Brewery name, which is what they have on all the signage here. The wiki page mentions they're actively using both names. I suspect this might be to get around really odd alcohol laws related to breweries in some US states. I live in Oregon and we're just fine with breweries so it's not a thing here.

Adamant36:
"I suspect this might be to get around really odd alcohol laws related to breweries in some US states."

What's the thing with breweries? You just can't have the word "brewery" in the name or something?

PR #2435, opened 2019-03-13
Add buybuy BABY

tas50:
This is one of the larger missing brands left in most shopping centers.

### DIFF
--- a/brands/shop/baby_goods.json
+++ b/brands/shop/baby_goods.json
@@ -48,6 +48,16 @@
       "shop": "baby_goods"
     }
   },
+  "shop/baby_goods|buybuy BABY": {
+    "nocount": true,
+    "tags": {
+      "brand": "buybuy BABY",
+      "brand:wikidata": "Q5003352",
+      "brand:wikipedia": "en:Buy Buy Baby",
+      "name": "buybuy BABY",
+      "shop": "baby_goods"
+    }
+  },
   "shop/baby_goods|西松屋": {
     "count": 67,
     "countryCodes": ["jp"],


Adamant36:
It looks like there's 23 instances of Buy Buy Baby, versus 6 for buybuy BABY. So id probably go with Buy Buy Baby or at least add it as a match and the official name, but it seems like there's a clear tagging preference for it versus buybuy Baby.

matkoniecz:
Especially as it is (AFAIK) preferred to treat CAPS LOCK in brands as a font design, not as a preferred name.

bhousel:
Haha this is a thing that I've been thinking about a lot lately too.

Hope this isn't too controversial but - whenever I see a brand with an unusual name, I:
- make it look less weird
- do whatever Apple Maps does

So that means:

### Buy Buy Baby
Screenshot 2019-03-13 09 00 28

### Bojangles'
Screenshot 2019-03-13 09 01 48

bhousel:
Another data point:  Yesterday I added this cosmetic shop, which should really be styled like **M∙A∙C**
https://en.wikipedia.org/wiki/MAC_Cosmetics

And I kind of doubt we can encourage mappers to stick unicode bullets in the name, but maybe when it's in iD it won't be a problem 🤷‍♂️ 

https://github.com/osmlab/name-suggestion-index/blob/3d20b923450608577863d5248ace7c8269e190f6/brands/shop/cosmetics.json#L72-L85

and yes Apple does it right:
Screenshot 2019-03-13 08 59 51

tas50:
Here's the measure I've been trying to use in order from most important to least:

- Text on the companies website. When they write a sentence referring to themselves how do they stylize that text. Bojangles' is a great example because they get the apostrophe right and they use the shortened name (most of the time).
- Wikipedia entries. I'm sure these folks have battled it all out already.
- The signage. This one is a bit tricky since things usually are pretty stylized on the signs.

I generally avoid looking at how people have been tagging things on OSM because frankly people are pretty awful at this. Until I corrected them there were more Walgreens stores spelled Wallgreens or Walgreen's than Walgreens. This is why the presets are so great because we can get people tagging names correctly.

As for buybuy BABY it's on their signs, website, twitter, and it's on their wikipedia page (although mentioned as the stylized name). Also both Bing and Google display it as buybuy BABY. It'll show up on Apple maps if you search using buybuy BABY so they clearly have some alternative name field under the hood even though they're not stylizing it.

Also MAC is a really weird one. They do use the funky styling all over their website, but no one would ever search for that since it's really hard to type.

matkoniecz:
I admit that while Wikipedia has sane version of name ( since https://en.wikipedia.org/w/index.php?title=Buy_Buy_Baby&diff=604028185&oldid=604028041 ) "buybuy BABY" horribleness is quite popular.

In that case I would probably still go with saner version.

bhousel:
Cool.. Sounds like we are all in agreement that we can override a particularly weird name, so I set it as "Buy Buy Baby" in a5dd9d3

I also match a bunch of the alternate forms.  I'm starting to think `match` should be changed to work case insensitive, since it is kind of silly to need to `match` alternate capitalizations of a name.

> I generally avoid looking at how people have been tagging things on OSM because frankly people are pretty awful at this. 

💯this.. It's why I am so excited about this project right now!
(And once they are linked to `brand:wikidata` cleanup and brand renames become so much easier..)

PR #2436, opened 2019-03-13
Add Sleep Country

tas50:
It's listed as Sleep Country Canada only because there was once a Sleep Country USA in the pacific northwest that was acquired by sleep train that was acquired by MattressFirm. I've learned way too much about the mattress industry tonight.

### DIFF
--- a/brands/shop/bed.json
+++ b/brands/shop/bed.json
@@ -49,6 +49,17 @@
       "shop": "bed"
     }
   },
+  "shop/bed|Sleep Country": {
+    "countryCodes": ["ca"],
+    "nocount": true,
+    "tags": {
+      "brand": "Sleep Country",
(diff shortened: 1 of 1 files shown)


Adamant36:
There's 42 instances of Sleep Country Canada in the database. Would you mind adding it as a match? I was going to ask before I merged it, but I totally spaced it.

tas50:
I was going to create a maproulette challenge to fix those since they don't seem to be using Canada in the name now that the US version no longer exists.

Adamant36:
That works to. The question is if people don't tag it that way anymore. It looks from tag history like they do. So it might still worth be adding as a match to catch any new instances. Although I guess if they search for Sleep Country in iD editor it will stop them from adding Canada to the name.

PR #2439, opened 2019-03-13
Update logos and the docs

tas50:
I backfilled a few on wikidata, and reran the script


bhousel:
oh whoops!  I just did the same thing

bhousel:
though this makes me realize that I should fix the generated html to be more diff-able

matkoniecz:
> oh whoops! I just did the same thing

So I guess that it can be closed as this PR is just a script rerun and data was added externally to Wikidata?

bhousel:
> So I guess that it can be closed as this PR is just a script rerun and data was added externally to Wikidata?

Yeah I think so.. I just spot checked a few of the conflicts, and it seems like master is just slightly ahead , probably from me running the script after you.. Sorry!

tas50:
I was going to ask for a PP html so the diffs will make sense. That would be super useful

PR #2440, opened 2019-03-14
add E.Leclerc (supermarket)

johanricher:
#1659


Adamant36:
This should really have its own Wikidata entry since its a unique store brand of E. Leclerc, of which E. Leclerc has many. So E. Leclerc is just the operator. I'll create a Wikidata article for it if your not up on that and then you can update the PR. I'll do the same for the fuel station also if you want. It was on my list to get to eventually anyway.

matkoniecz:
It is still E.Leclerc brand.

PR #2441, opened 2019-03-14
add E.Leclerc (fuel stations)

johanricher:


Adamant36:
Thanks the PR. Station service E.Leclerc is a gastation brand of E.Leclerc though. In cases like this the usual procedure is to add the parent as the operator with the operator_name tag instead of changing the brand. The same can done for the Wikidata/Wikipedia entries with operator:wikidata and operator:wikipedia. Station Service E. Leclerc should really have its own brand:wikiwhatever entries.

matkoniecz:
> E.Leclerc
I suspect that space may be missing - is it maybe `E. Leclerc`?

> add the parent as the operator with the operator_name tag instead of changing the brand

I am not sure is it desirable. And certainly not using operator_name tag ( https://taginfo.openstreetmap.org//search?q=operator_name ).

Adamant36:
Sorry, I meant "operator." To many tags going in my head. Thanks for catching the mistake.

Adamant36:
> I suspect that space may be missing - is it maybe E. Leclerc?

Btw, it doesn't seem to have a space. I guess its another one of those weird brand names. I guess its French. Go figure. According to Wikipedia they call it Leclerc. Maybe that would work better.

matkoniecz:

Adamant36:
@matkoniecz, how come you merged it despite the bad Wikipedia entry? There's no reason a Wikidata entry for Station service E.Leclerc couldn't have been created. Since E.Leclerc isn't the brand or the name of the things.

matkoniecz:
As far as I can see it is yet another supermarket fuel station.

See https://www.openstreetmap.org/way/73692989#map=18/48.29773/4.13150 or https://www.openstreetmap.org/way/245555000#map=19/48.37420/2.94191 for what seems a typical example.

It seems to me that supermarkets and fuel stations operate under a shared brand like it is for many other fuel stations on parking lots.

Adamant36:
"It seems to me that supermarkets and fuel stations operate under a shared brand like it is for many other fuel stations on parking lots." 

This case is a little more complicated unfortunately. Check out their Wikipedia page. 

"The chain enables semi-independent stores to operate under the Leclerc brand. Leclerc operates numerous stores and services around France, some near and in cities and towns while most of them are located just outside or in the hypermarkets and shopping centres.

L'auto (Car Garage)
Brico (DIY store)
Le Prou Construction (Residential and commercial construction)
Espace Culturel (Multimedia store/DVD store)
Jardin (Garden Centre)
Location (A Quick Car rental Service)
Manege a Bijoux (Jewelry Kiosk)
Optique (Opticians)
Parapharmacie (Pharmacy, Chemist store)
Sport (Sports shop)
Station-Service (Petrol/Gas Station)
UNE HEURE POUR SOI (Perfume store)
Bornes electriques (Electrical store)
Animalerie (Pet Shop)
Jouets (Game store mainly aimed at kids)"

So their gas stations aren't necessarily connected to their supermarkets. Any more then their other branded stores are. 

http://www.e-leclerc.com/thionville/magasins-specialises/station/info-station-service

Adamant36:
Checkout the forth picture of the store front 

https://en.pornic.com/parapharmacie-e-leclerc.html

matkoniecz:
> The chain enables semi-independent stores to operate under the Leclerc brand.

Looks like a standard franchise model.

> Checkout the forth picture of the store front
> en.pornic.com/parapharmacie-e-leclerc.html

Based on this image I would describe it as a E.Leclerc branded pharmacy. Note that it does not mean that it is owned by a E.Leclerc - in the same way as many McDonald's fast foods are not owned by McDonald's company.

Adamant36:
Station Service still has the name E. Leclerc. I never claimed otherwise. It was your assumption that they are attached to supermarkets that I'm disputing. In the case of the fuel stations the operator is E. Leclerc and the brand name is Station Service E. Leclerc. Even their own websites say that's their name. Do a Google search for just E. Leclerc and see how many fuel stations come up. Then do a search for Station-Service E. Leclerc. "Station Service" isn't just a generic Station Service isn't just a generic suffix. Its part of actual names of the places. Otherwise theirs zero way to tell them apart from other E. Leclerc store brands/types when your searching for them.

matkoniecz:
> and the brand name is Station Service E. Leclerc

OK, I think that I now understand. For me both "the brand name is Station Service E. Leclerc" and "the brand name is E. Leclerc" are correct, in your opinion only the first is correct, right?

bhousel:
Trying to follow the discussion - I think I understand..

It is a problem that we are inconsistent about how treat these sub-brands in the index.  I think we should probably tag the sub-brands with the `brand` of the parent, but the `name` can match the sign where it differs.

Costco is a similar situation:

```js
  "amenity/fuel|Costco Gasoline": {
   ...
    "tags": {
      "amenity": "fuel",
      "brand": "Costco Gasoline",     // I think this should change to "Costco"
      "brand:wikidata": "Q715583",      // because this is really for "Costco"
      "brand:wikipedia": "en:Costco",
      "name": "Costco Gasoline"       // I think this should stay as "Costco Gasoline"
    }
  },

  "shop/wholesale|Costco": {
   ...
    "tags": {
      "brand": "Costco",
      "brand:wikidata": "Q715583",
      "brand:wikipedia": "en:Costco",
      "name": "Costco",
      "operator": "Costco",
      "shop": "wholesale"
    }
  },
```

Adamant36:
"OK, I think that I now understand. For me both "the brand name is Station Service E. Leclerc" and "the brand name is E. Leclerc" are correct, in your opinion only the first is correct, right?" 

Yes correct. Its rather convoluted.

PR #2442, opened 2019-03-15
Add wiki data to BoConcept A/S

tordans:
I also added to social media profiles to https://www.wikidata.org/wiki/Q11338915

I did not add the wikipedia page, since it seems to to be well written https://en.wikipedia.org/wiki/BoConcept "This article contains content that is written like an advertisement."

### DIFF
--- a/brands/shop/furniture.json
+++ b/brands/shop/furniture.json
@@ -46,6 +46,7 @@
     "count": 52,
     "tags": {
       "brand": "BoConcept",
+      "brand:wikidata": "Q11338915",
       "name": "BoConcept",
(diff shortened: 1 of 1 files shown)


Adamant36:
> I did not add the wikipedia page, since it seems to to be well written 

It doesn't really matter how its written. A lot of them are bad, but its better then nothing and its not like they can't be written better later. The important thing is the Wikidata entry anyway, but it would still be good if you added it.

bhousel:
Thanks @tordans and welcome!
I added the `brand:wikipedia` tag in 8179bfb

PR #2444, opened 2019-03-15
add Extrafarma and more information to Drogasil

willemarcel:
All Drogasil shops are wheelchair=yes

### DIFF
--- a/brands/amenity/pharmacy.json
+++ b/brands/amenity/pharmacy.json
@@ -314,9 +314,11 @@
   },
   "amenity/pharmacy|Drogasil": {
     "count": 266,
+    "countryCodes": ["br"],
     "tags": {
       "amenity": "pharmacy",
       "brand": "Drogasil",
+      "wheelchair": "yes",
       "brand:wikidata": "Q5308185",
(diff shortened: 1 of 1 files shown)


matkoniecz:
> All Drogasil shops are wheelchair=yes

Are we 100% sure that all opened in future also will be?

I am really dubious whatever this assumption is a good idea.

willemarcel:
@matkoniecz The company has this as a principle and only allow to open shops in buildings with wheelchair accessibility.

matkoniecz:
It is still a bit dubious for me to add it as default for all new objects. What if in some place they made a mistake? Or management changes?

@bhousel What you think about adding `wheelchair=yes` via name suggestions?

bhousel:
> @bhousel What you think about adding `wheelchair=yes` via name suggestions?

That is a pretty awesome policy for the company's management to follow... 
I'm ok with it, if it really is their thing.

bhousel:
Also I added `"nocount": true` for Extrafarma, because they don't appear that way in the most recent planet that we checked.  (The record would be in `allNames.json` if so)

PR #2446, opened 2019-03-15
add GBarbosa supermarkets

willemarcel:


matkoniecz:
Just to confirm - it is "GBarbosa", not "G Barbosa" like article name at https://pt.wikipedia.org/wiki/G_Barbosa ?

willemarcel:
@matkoniecz Yes, It's "GBarbosa", I tried to fix the wikipedia article, but a collaborator reverted my edit without checking. :smile:

brands/shop/supermarket.json
```diff
@@ -1590,6 +1590,18 @@
       "shop": "supermarket"
     }
   },
+  "shop/supermarket|GBarbosa": {
+    "count": 316,
+    "countryCodes": ["br"],
+    "tags": {
+      "brand": "GBarbosa Cencosud",
```
matkoniecz: I am guessing that "GBarbosa Cencosud" is name of company, not name used for its brand - right?

I suspect that

`       "brand": "GBarbosa",`

would be better (note that I have never seen GBarbosa shop in my life, that is just a guess)

willemarcel: They have the word "Cencosud" on the logo after the name, but they don't mention it on advertising, texts... You can remove it, if you think it's more correct. I don't know well about the brand key.

bhousel: I made these changes and merged it

I also switched `website` -> `brand:website`.  I don't normally add these, since they can be fetched from Wikidata, but if people want to add them it doesn't break anything.

Issue #2457, opened 2019-03-15
Update Acme with wikidata


yoasif:
http://osmlab.github.io/name-suggestion-index/brands/shop/supermarket.html#shopsupermarketacme

https://www.wikidata.org/wiki/Q341975

bhousel:
Thanks for all the issues @yoasif - Are you interested in submitting Pull Requests also?  
If not, that is ok!   But since you are doing all the work of researching these brands, it is just a few quick extra steps to get them added to the index, and I'm happy to walk you through the process if you want.

yoasif:
I can't figure out how to do a pull request from the command line after I have edited the file and run a build. Installed "hub" and I got stuck at trying 

git push origin feature
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

after forking.

matkoniecz:
You are starting with git, right?

I remember using at first https://desktop.github.com/ (though it is redesigned now, so I am not sure is it a still good for a beginner).

----

You may also edit directly in browser - https://github.com/yoasif/name-suggestion-index/blob/master/brands/shop/supermarket.json should have edit link icon leading to in-browser editor.

----

https://guides.github.com/activities/hello-world/ also may be useful

yoasif:
It isn't fun, but I am using the GH website for this for now.

Issue #2454, opened 2019-03-15
MidFirst Bank


yoasif:
https://www.wikidata.org/wiki/Q17081131

Issue #2453, opened 2019-03-15
Cathay Bank


yoasif:
https://www.wikidata.org/wiki/Q4189507

Issue #2452, opened 2019-03-15
Simmons Bank


yoasif:
https://www.wikidata.org/wiki/Q28402389

PR #2460, opened 2019-03-15
adding apple, south state, simmons, midfirst, first midwest... etc banks

yoasif:


yoasif:
See  #2452  #2447 #2448 #2449 #2450 #2453  #2454

Adamant36:
Hi there. If you put "closes" before each issue number it will auto close them when the PR is merged. It saves us a bunch of time having to go through and close them later. 

Also, what are you using to do these? Git in the terminal or something else?

yoasif:
I can't figure out how to create pull requests from the terminal, so I am pasting the file into Github.

Adamant36:
You should probably put the time into figuring out how to do it in the terminal. This way seems to be causing problems, like changing icon links to worse ones in some cases for some reason. It shouldn't be that hard to figure out the terminal and it will give you more control over things.

yoasif:
I was told to try to edit directly in the browser. https://github.com/osmlab/name-suggestion-index/issues/2457#issuecomment-473434391

I can look into doing  that later, but I have no idea why GitHub's UI would break pasted text.

Adamant36:
I wouldn't say the advice given in #2457 was bad advice per say, but its always better to try and figure out how to fix a problem then use a maybe janky work around. Editing things in the Github UI might be good for changing small things, like doing grammar correction, but using an actual code editor and the terminal is way better for more comprehensive things like adding a bunch of new entries. 

Not the least because you can do the npm run build thing to check for errors. Whereas, Github's error checker checks for different issues and doesn't flag a lot of the stuff npm run build does. I'm not saying Github would "break pasted text" either. I'm saying they are different tools that have different purposes, advantages, and disadvantages. In this case it would just be better to figure out what your terminal problem is and do it that way if you can.

yoasif:
I ran npm run build and I copied the resulting file into the GH UI. 

If what is pasted into the GH UI is different from what was copied from the npm build command, I have zero control over that. 

I may look into figuring it out later, but frankly, I could have just opened issues instead of trying to write json here (which is what I did until someone asked me to try to write it).

Adamant36:
I'm pretty sure I didn't say it was "your problem," but whatever the case, I requested @bhousel to review it. I'm sure he can figure it out. Hopefully you can work out the issues on your end so you can just use the terminal and Github won't screw things up. Like I said, feel free to ask questions or open a new issue if need be.

bhousel:
Thanks @yoasif - it looks pretty good.  

I'm not too worried about the logo urls changing because the next time we run that logo fetching script, they will just update again with whatever is newest.

It's totally ok to 
https://github.com/openstreetmap/iD/blob/master/CONTRIBUTING.md#using-github-and-git

bhousel:
It's totally ok to have questions about GitHub!  We have created some information over on the iD project, which I think is a pretty good guide to get started.

https://github.com/openstreetmap/iD/blob/master/CONTRIBUTING.md#using-github-and-git
also
https://github.com/openstreetmap/iD/issues/4324#issuecomment-369302713

yoasif:
Closing this PR in favor of my terminal based one. https://github.com/osmlab/name-suggestion-index/pull/2467

PR #2465, opened 2019-03-16
Merge shop/interior_decoration|Casa into shop/furniture|Casa

waldyrious:


Adamant36:
Seems like they are a furniture store. When I did a search for them they are actually called "Casa Furniture" and that's what it is called on their [website](https://casafurniture.com/). So this should probably be reversed.

bhousel:
Yeah let's keep it as `shop=furniture`.  This is a situation where the previous mappers are more often wrong than right!  😆

waldyrious:
Thanks so much for double-checking. I was looking for a way to quickly contributing manpower but it turns out the issue is not the scale of the task list, but the nuance required to solve each one. This particular example seemed to be more straightforward than the others, but turns out it wasn't either. Sorry! 😳 Will fix.

waldyrious:
Btw, should the counts be summed when merging entries, or do I keep them untouched?

ENT8R:
> Btw, should the counts be summed when merging entries, or do I keep them untouched?

Just keep them untouched.

bhousel:
Thanks @waldyrious 👍

PR #2466, opened 2019-03-16
Add brand:(wikipedia|wikidata) tags to two entries

waldyrious:
- amenity/bank|Emirates NBD
- amenity/fast_food|Pollo Tropical


Adamant36:
Thanks, looks good.

Issue #275, opened 2018-10-16
Research needed for: amenity/bank|Kreissparkasse Köln


bhousel:
#### :thinking: We need help deciding what to do with "amenity/bank|Kreissparkasse Köln"

:point_right: Read the [CONTRIBUTING.md](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) guide for more info, and feel free to ask questions on this issue!

### What's needed

First, do some research on `amenity/bank|Kreissparkasse Köln` by searching Google and Wikipedia.

#### If it is an actual brand:

* Edit `config/canonical.json`:
  * Find the entry for `amenity/bank|Kreissparkasse Köln`
  * Add tags for `brand:wikidata` and `brand:wikipedia`
  * Optionally add `"countryCodes"` or other tags
* `npm run build`
* Commit your changes and submit a pull request

#### If it is not an actual brand:

* Edit `config/filters.json` to add an expression that discards it
* Edit `config/canonical.json` to remove the entry for `amenity/bank|Kreissparkasse Köln`
* `npm run build`
* Commit your changes and submit a pull request

#### If the situation is complicated:

* Leave a comment on this issue with what you find

:sparkles: Thanks!

_This is an autogenerated ticket_

ENT8R:
This is only a regional brand (around the German city of Cologne) but it has its own Wikipedia entry: https://de.wikipedia.org/wiki/Kreissparkasse_K%C3%B6ln and then of course also a Wikidata entry: https://www.wikidata.org/wiki/Q1787788. But basically it is the same as https://github.com/osmlab/name-suggestion-index/issues/274... What should be done in this special case?

bhousel:
If they have 2 separate wikidata, we can keep 2 separate entries 👍

ENT8R:
But then, theoretically we also have to add all other "Kreissparkassen" listed in the table of this Wikipedia entry: https://de.wikipedia.org/wiki/Liste_der_Sparkassen_in_Deutschland#Bestehende_Sparkassen which is IMO a bad idea because if somebody wants to add a new "Kreissparkasse" using iD, every single instance is shown to the user... "Kreissparkasse" (without the city name) is sufficient and everybody does understand which bank is meant...

bhousel:
> "Kreissparkasse" (without the city name) is sufficient and everybody does understand which bank is meant...

Ah gotcha - yes reading up on wikipedia now about the German public banking system.  
[...]

PR #2475, opened 2019-03-16
Add all German Sparkassen (ref #275)

ENT8R:
This is a pretty big change, so I rather created a pull request for additional reviews 😄
Like already said in the discussion of #275 this allows users to select a very specific entry for the "Sparkasse" they want to add. There are nearly 400 new entries and it is so much, because there is a different kind of a "Sparkasse" in nearly every single town in Germany.
I wrote a script which parses the table of https://de.wikipedia.org/wiki/Liste_der_Sparkassen_in_Deutschland which lists all German "Sparkassen", and outputs a file in the format we need it to be.
I plan to publish the source code of the script, so it can be rerun at any time because I suppose that more and more "Sparkassen" of neighbouring cities will merge in the future as this is also happening nowadays. (The amount of "Sparkassen" decreased from 769 in 1990 to 390 in 2017... So this will probably also change in the future)
In addition to the script I checked all entries manually afterwards and fixed entries were something was wrong.
I'm pretty sure that everything is right as it is because the data quality of the Wikipedia entries was very good over the whole time. But if somebody wants to check the list of added entries, any input and suggestions are welcome!


ENT8R:
I published the source code of the parser at https://github.com/ENT8R/sparkassen-parser

matkoniecz:
> I plan to publish the source code of the script

I would include it as a part of the PR.

>  parses the table of de.wikipedia.org/wiki/Liste_der_Sparkassen_in_Deutschland 

I wonder is it OK from a copyright side.

> This branch cannot be rebased due to conflicts

Can you rebase it?

ENT8R:
> I would include it as a part of the PR.

I know that this is some kind of related to this project but it is still a different project and it might be better if it stays at different repositories for a better overview...

> I wonder is it OK from a copyright side.

I also don't know that for 100%
Maybe somebody else knows whether CC-BY-SA 3.0 is compatible to BSD 3-Clause?

Adamant36:
> I know that this is some kind of related to this project but it is still a different project and it might be better if it stays at different repositories for a better overview

I agree with that. I like the fact that it has a separate readme file and different forking path.

matthijsmelissen:
As far as I know, these banks do not advertise with the name of their legal (local) body, but generically as 'Sparkasse'. See for example this shop front: http://purplesneakers.com.au/content/uploads/2017/12/AlexanderplatzSparkasse-827x552.jpg

What is the best tagging practice according to the German community here? I would be inclined to follow the name that the company uses in its own advertisements, i.e. 'Sparkasse'.

matthijsmelissen:
A quick check at the data shows that name=Sparkasse is by far the most common way these shops are currently tagged. I would be strongly against merging this pull request.

Adamant36:
> As far as I know, these banks do not advertise with the name of their legal (local) body, but generically as 'Sparkasse'. See for example this shop front: 

Did you look through the prior discussion? Sparkasse is just a type of bank. Its not a brand. The vast majority of them are tagged with the brand names already. Plus, they all have their own websites, twitter handles, facebook pages, etc. They were all already mostly tagged separately until @ENT8R wrongly consolidated them under what is essentially a type of bank. Which led to the discussion in #275 and this PR. Its the essentially the same as when people miss-tag things as cooperatives or similar miss-tagging. 

See https://github.com/osmlab/name-suggestion-index/issues/275#issuecomment-473524050

"Hamburger␣Sparkasse (155 uses), Sparkasse␣KölnBonn (90 uses), Mittelbrandenburgische␣Sparkasse (89 uses), Sparkasse␣Ansbach (83 uses), Kreissparkasse␣Köln (79 uses), Kreissparkasse␣Ludwigsburg (40 uses)" etc. etc. 

So by consolidating them your essentially losing the unique entries for all of those bank brands. Including 155 branches of Hamburger Sparkasse. There might be 5000 instances of name=Sparkasse, but there's no bank just called that and if you add up the correctly named names and fixed the miss-tagged ones its probably close to it same amount.  Sparkasse should be a ref=* key or something though. If you look at the tag history it seems to be declining to. Probably in favor of more specific, proper brand tagging. Most of the usage is between 2011 and 2014. We don't necessarily go with how things are currently tagged in some cases either. So it doesn't really matter how the Germany community might tag them. Although, its clear that they don't have a definitive way of doing it in this case. 

Your assertion that they don't advertise their name brand on the store front is just plan wrong also. 

haspa

kreissparkasse-koeln

KolnBonn

There's plenty more examples.

Adamant36:
More examples 

benningen_cashpoint

Stadtsparkasse

Kreissparkasse Tubingen 3

Sparkassen Carre 2

Sparkassen Carre

download

Stadtsparkasse 2

matkoniecz:
So some Sparkassen operate under name "Sparkassen" and some under their specific brand?

ENT8R:
All Sparkassen are operating under there own name. They are just using the generic name for advertising purposes. But if you are looking at the information printed at the door you can find the name of the Sparkasse really quick.

bhousel:
Thanks @ENT8R  - I definitely don't have a problem with this, since they all do have wikidata entries and seem to operate like small brands.

Can you just double check on the forum to make sure the German community is really ok with this?

"name": "Hohenzollerische Landesbank Kreissparkasse Sigmaringen"

matthijsmelissen:
> Can you just double check on the forum to make sure the German community is really ok with this?

This is the most important thing, if the Germany community supports this I'm fine with this too. But we need agreement from the German community.

ENT8R:
I asked on the forum now: https://forum.openstreetmap.org/viewtopic.php?id=65695
I will tell you when there are any new information

ENT8R:
There is also no real consensus in the German community about this. It actually differs strongly by the mappers own preference.

But nevertheless, the majority of said that they think the more specific name is the better solution after all. There is only one user who thinks that it would be the best to simply tag `name=Sparkasse` because there are some names which are too long in his opinion... But if it is the actual name of the bank, then it should also be added in that way IMO.

I would maybe wait another day to get some more feedback on this topic on the forum.

Adamant36:
That's the way it looked to me to. If the only objection is the possibility of a few names being to long that's not really a reason to not do it. Its not we can't use a shortened, but more correct version of the name and just alt_name or official_name for the longer one.

Adamant36:
> No one will ever say "I'll go to Mittelbrandenburgische Sparkasse in Potsdam Branch Office Lübben Branch Lessingstraße Get Money" but always: "I'll go over to Sparkasse quickly" If I look at a card, I always have "Sparkasse"! I'm looking for something, I'm looking for: "where is the nearest Sparkasse" and not where is the "Mittelbrandenburgische Sparkasse in Potsdam office Lübben branch office Lessingstraße"  All other information is then in the supplementary tags.

Just between you and me @ENT8R, I think the argument against it by stretch watchers is pretty weak. I always say "I'm going to the pharmacy" without specifying the particular pharmacy I'm going to. Everyone know's where I'm going to without specifying also. That doesn't mean the names of pharmacy's where I live should be dropped or added to "supplementary" tags instead of the name. So I don't know what he's getting at there. He seems to be mixing up the difference between a "branch" (one bank) and a brand of banks (lots of banks all called Mittelbrandenburgische Sparkasse). What do you think?

I wish I knew German so I could participate gggrr. At least there's Google Translate so I can read it though.

ENT8R:
Yes, that's true. I mean, if we can specify it more precisely, we should also do that.
So, it is indeed a pretty weak argument...

osmtools:
> There is only one user who thinks that it would be the best to simply tag `name=Sparkasse` because there are some names which are too long in his opinion...

You forget to mention, that there's another user who thinks that it would be the best to simply NOT tag `name` in the preset at all (`brand=Sparkasse` is enough), and let the mapper decide what he want to map as name.

Regards, Harald Hartmann

Adamant36:
"You forget to mention, that there's another user who thinks that it would be the best to simply NOT tag name in the preset at all" 

True. That seems like an odd way to go though. Its not like people can't change the name themselves if need be or just ignore the suggestion altogether and put in what they want. iD Editor isn't forcing anyone to use anything. I think there's a miss understanding with some people that it is.

bhousel:
I find myself using this convention elsewhere in the name-suggestion-index, and I think it works pretty well:
- `name` : what you really expect to see rendered on the map
- `short_name`:  a shorter name which would also be fine to render on the map
- `official_name`: the longer version which really nobody expects to see rendered on the map

(basically, just tag for the renderer.  sorry!)  

So the question really reduces to how silly the Germans want their map to look, and I can not provide guidance in this area.

The two options are something like:
- `name=Hohenzollerische Landesbank Kreissparkasse Sigmaringen`
- `short_name=Sparkasse`
or
- `name=Sparkasse`
- `official_name=Hohenzollerische Landesbank Kreissparkasse Sigmaringen`

I know which one I'd prefer.

Adamant36:
Luckily 99% them are pretty short and shouldn't be a problem. Hohenzollerische Landesbank Kreissparkasse Sigmaringen is really long though. I bet it would look horrible on the map. Not that it matters, but I just noticed Google Maps uses their full names, although it cuts longer ones short. At least it shows the full names are used places though.

ENT8R:
I asked again in the forum and I think we should use
- `name=Sparkasse`
- `official_name=Hohenzollerische Landesbank Kreissparkasse Sigmaringen`

This is basically the same information, just splitted into two separate tags and IMO this is still fine.
People who are looking on a map and are looking for the nearest "Sparkasse" can then find it also much faster (because they don't have to read the whole name 😄)

Adamant36:
Just for that one sure. Hopefully not the rest though. Really it should be "kreissparkesse" though. Since that's what the word in the name is. Otherwise you just tagging for the render with what should be a type tag. Which I'm totally against. Just like making all Rite Aids name=pharmacy official_name=rite_aid_pharmacy would be bad in my opionion. Banks that don't have the word "sparkesse" in there names shouldn't have it as the name tag either. It should be whatever kind of public bank it is in the businesses title. There's different ones. Landsbanken etc. You should read over the Wikipedia article about them if you haven't yet. 

Personally, in general I don't think we should disregard tagging conventions just to satisfy a local community. To me its a bad precedence. They should have taken the time to propse a bank type tag or something. Instead of missusing the name tag just because it was more convienent. The whole us rather irriating. 

I'm currently in an argument with a "local" mapper that is digging his heals in on why all the beaches in his area should be tagged as parks instead of beaches, because him and his friends decided they should be tagged that way in a coffee shop together and its "extending" osm tagging (whatever that means), not missusing it. Although, he refuses to have the areas also tagged as natural=beach. This reaks as a similar thing to me. I might be bias though.

matkoniecz:
@matthijsmelissen @ENT8R Is there some conclusion from discussion on the German forum?

ENT8R:
@matkoniecz No. It seems just like everybody has his own convention on how to tag a branch of this bank properly...
Maybe we should really keep it as it is right now and only add "Sparkassen" with an amount of more than 50 branches? All others don't have that much branches, so there also won't be probably that much users who want to add a new branch. In fact, the amount of branches steadily descreases because of online banking and stuff like that... So adding every single entry of every single "Sparkasse" might be a little bit overpowered...

Adamant36:
Personally, id go with adding any over 10. That's the number id say they are "established" at and where its worth having the extra information for people want to add more. Plus, its still currently only 10 Sparkassen in the database at that point. Your only looking at one bank in the database so for with more than 50. At 50 there probably won't be that many new ones left to map. So its semi useless. Especially if other Sparkassen aren't distinguished. At  least in my opinion.

bhousel:
Since there are so many of them, and we never really got buy-in from the local community, let's just remove them.

kjonosm:
@ENT8R 
nsi now handles brands also without the `name` tag. Is it worth reconsidering to introduce Sparkassen?

tiptoptom:
> @ENT8R nsi now handles brands also without the `name` tag. Is it worth reconsidering to introduce Sparkassen?

I would say yes.

And by the way, it's already in operators/man_made/surveillance:
https://github.com/osmlab/name-suggestion-index/blob/4e32bd36a1fe1890cda6c9a6ceb0a6567b8b78bc/data/operators/man_made/surveillance.json#L7390-L7399

PR #2480, opened 2019-03-17
added some NY based stores

yoasif:


Adamant36:
What's the difference between the brand and name tags to you? In my experience they are usually the same. As the brand name is the same as the store name. As it is, use "official_name" for cases like Century 21. So it would be brand=Century 21 + official_name=Century 21 Department Store since "department store" is just a suffix and not part of the brand name.

yoasif:
Century 21 Department Store is what is shown on  the official website: https://www.c21stores.com/about-us.html

Per https://wiki.openstreetmap.org/wiki/Key:brand#Example_Hotel "official name" may be used "if the name filed with government regulators, for example, is different" so I don't think it makes sense to put this in this db unless we can be assured that the official name filed with governments is consistent. Given franchises etc., I think this is unlikely to ever be consistent except in a small number of cases (where all locations are owned by the same entity and are filed with the same government agencies). 

"Name" to me means the name that appears on the ground: https://wiki.openstreetmap.org/wiki/Disputes#On_the_Ground_Rule whereas the "brand" would be the name that the brand is officially known as - for Century 21, the name can differ on the ground, but can (and should) be overridden if it is more like "Century 21 Store", for example. The "brand" is still what appears on the site as linked above. 

Hope that clarifies my thinking around this, but happy to discuss further.

Adamant36:
> official name" may be used "if the name filed with government regulators, for example, is different"

I highly doubt anyway is checking with government regulators before using the tag or not. To me, according to the "key:name" article on the wiki, "It has been created for country names but we need a clarification for other cases between "name", "int_name", "loc_name" and "official_name". Example: official_name=Principat d'Andorra (where "name" is name=Andorra)." That doesn't say anything about government regulations. Its just where the name needs clarification between the name key and a local name or whatever. 

In other words, its what the company officially calls the brand (In this case Century 21 Department Store) and how it is tagged in the real world with the brand/name tags (in this case Century 21).  There's 351 uses of name=Century 21 and 11 uses of brand=Century 21 in the database and none for Century 21 Department Store. So its clearly not how its tagged. Official_name is used all over the index for situations like this already also. So its more then established for things to be tagged that way if need be. 

As far as the brand tag goes, the wiki says it "represents a common identify for individually owned and operated stores" and also "typically this tag is used where the name of the brand is more recognizable (and therefore of more interest to data users) than the name of the outlet." Not all stores have the "Department Store" suffix. Whereas, all stores have "Century 21" in the title. So brand=Century 21 would serve as a more recognizable name then "Century 21 Department Stores" would. The tagging reflects that also. Whereas, the name of the outlet "Century 21 Department Stores" would be the name of the outlet (or the official name). 

As a side to that, there's also been multiple discussions about businesses with suffixes in their names and the need to keep names short. The general consensus is that they should be left out and put in the official_name tag instead so they don't make names on the map to long. That's kind of its purpose for me. 

Here's some examples of Century 21's that don't have "Department Store" in their name. 

ls

2014-04-century-21 0
images

Interestingly their official logo doesn't even have "Department Store" in it. It just says "Stores." Which is the same way its on their website. 

Century 21 Stores

There's also some stores that are just named C21. There's currently 49 uses of name=C21. Which I'm pretty sure are those stores. Who knows how many uses of "Century 21" are actually miss-tagged C21 stores. 

image

So whereas all "Century 21 Department Store" stores have "Century 21" in them, not all "Century 21" stores have "Department Store" in their title. Some don't even have the word "Century."

Adamant36:
https://www.c21stores.com/about-us.html

yoasif:
>I highly doubt anyway is checking with government regulators before using the tag or not. To me, according to the "key:name" article on the wiki, "It has been created for country names but we need a clarification for other cases between "name", "int_name", "loc_name" and "official_name". Example: official_name=Principat d'Andorra (where "name" is name=Andorra)." That doesn't say anything about government regulations. Its just where the name needs clarification between the name key and a local name or whatever.

This is just the best I could come up with after reading that page; as far as people not looking at government filings when entering official names - I don't know -- if someone is going to enter the "operator", for instance, the official name (the name the location is filed as) should be found right alongside that. 

For Century 21, see for example:

Screenshot_2019-03-17 Entity Information

Screenshot_2019-03-17 Entity Information(1)

>As far as the brand tag goes, the wiki says it "represents a common identify for individually owned and operated stores" and also "typically this tag is used where the name of the brand is more recognizable (and therefore of more interest to data users) than the name of the outlet."

Unfortunately, there is no clarification in this type of use case, where the two names are basically the same brand, but variations of them. I think the example is more like if the "name" is "cash transfer store" and the "brand" is "western union".

>There's also some stores that are just named C21. There's currently 49 uses of name=C21. Which I'm pretty sure are those stores. Who knows how many uses of "Century 21" are actually miss-tagged C21 stores.

That to me implies that we ought to not be prescriptive about the name entirely and remove the "name" from this entry, because the ground rule should dominate the name, not what appears on the websites (where I think "brand" continues to make sense").

Adamant36:
> if someone is going to enter the "operator", for instance, the official name (the name the location is filed as) should be found right alongside that.

Operators are usually found by looking at the brands Wikipedia pages, not digging through FCC fillings or whatever. I think there's an assumption with a lot of this that it has to be understandable or doable by most people and not involve a degree in how to search through government filings. 

> Unfortunately, there is no clarification in this type of use case

Exactly, at least not on the Wiki. There is here though in how we use it, because that's what @bhousel decided on and it makes sense to me. Generally speaking the Wiki is pretty unreliable and there's some places where we diverge from it. Which is totally OK. It's not a set in stone document and anyone can edit it. So sometimes there's a need to not follow it exactly. This being one of them. 

> That to me implies that we ought to not be prescriptive about the name entirely and remove the "name" from this entry, because the ground rule should dominate the name, not what appears on the websites (where I think "brand" continues to make sense").

That's exactly what we are doing by keeping the name tag as "Century 21," which is how things are actually tagged. I'm not sure how that translates into throwing the name tag out completely. We have a lot more leeway with the brand tag and its perfectly fine to have it be brand=Century 21 also. There's really no reason not to. We know the instances where the name tag is Century 21 are also the Century 21 brand. At least it seems that to me. As I said, "store" and "department store" are just suffixes. Same as we aren't including the whole "Westchester LLC" at the end in either of them.

Adamant36:
Read https://github.com/osmlab/name-suggestion-index/issues/2236#issuecomment-451346930

yoasif:
https://github.com/osmlab/name-suggestion-index/issues/2236#issuecomment-451346930 doesn't give clarification on brand, just on name, and my PR is consistent with the recommendation on name. 

So I think that you are correct that removing the name doesn't need to happen, since Century 21 appears to be the preferred name, even as C21 is used in places.

Adamant36:
Personally, I think Century 21 should be the brand also though and that Century 21 Department Store should be the official name. 

See here, https://github.com/osmlab/name-suggestion-index/issues/2326#issuecomment-466035521 (I thought the original post covered it, but I guess I was wrong).

yoasif:
Same issue with https://github.com/osmlab/name-suggestion-index/issues/2326#issuecomment-466035521 -- there is really no clarity here that says that brand name should match the name, and the name I have used already complies with the idea that the name should "render reasonably on a map". 

I don't understand why we need official_name if brand already exists, especially when official name seems to want to point to something like "legal name", not the commercially used brand name (that may not match the name on the ground).

Adamant36:
> especially when official name seems to want to point to something like "legal name"

It doesn't except for on that one Wiki page you cited that wasn't even the main article for the tag.

Adamant36:
> I don't understand why we need official_name

What about something like NAPA Auto Parts? There's NAPA (how people tag it on the ground), NAPA Auto Parts (the brand name), and then National Automotive Parts Association (the official name). I think there's a lot of good use cases for official name like that.

bhousel:
1. It should be "Century 21"
2. We really don't need `official_name` here.  We probably don't need it anywhere else either, but some people add it in cases where the `name` tag could reasonably be written several ways and we're preferring the one that looks least weird on a map.
3. Please don't overcomplicate things.. @yoasif if you find yourself looking up filings with the NYS department of state to find the brand's legal entity name, you're doing it wrong.  Just pick whatever name looks least weird next to a map pin.

bhousel:
also, this entry has a few problems
1. `shop/clothes|Coach New York` does not match the `"shop": "boutique"` in the `tags` property
2. we already have an entry for Coach in the `shop/bag.json` so we don't need this one.

```
  "shop/clothes|Coach New York": {
    "nocount": true,
    "tags": {
      "brand": "Coach New York",
      "brand:wikidata": "Q727697",
      "brand:wikipedia": "en:Coach New York",
      "name": "Coach",
      "shop": "boutique"
    }
  },
```

bhousel:
and on this one
- no "smart" apostrophes
- no "Supermarkets" in the name
- name doesn't match key   (it should be `key/value|name`)

```
  "shop/supermarket|D’Agostino Supermarkets": {
    "countryCodes": ["us"],
    "nocount": true,
    "tags": {
      "brand": "D’Agostino Supermarkets",
      "brand:wikidata": "Q20656844",
      "brand:wikipedia": "en:D'Agostino Supermarkets",
      "name": "D’Agostino",
      "shop": "supermarket"
    }
  },
```

bhousel:
I cleaned up the issues and merged..
@yoasif please check e72af7c for the changes.. The PR was mostly good!   A lot of this is my fault for not having a clear FAQ or guidelines to point to.

PR #2481, opened 2019-03-17
Evroopt supermarket chain

tohaklim:
and wikidata link

### DIFF
--- a/brands/shop/supermarket.json
+++ b/brands/shop/supermarket.json
@@ -4611,9 +4611,15 @@
   "shop/supermarket|Евроопт": {
     "count": 222,
     "tags": {
+      "alt_name": "евроопт",
       "brand": "Евроопт",
+      "brand:wikidata": "Q2565040",
       "name": "Евроопт",
-      "shop": "supermarket"
+      "name:be": "Еўраопт",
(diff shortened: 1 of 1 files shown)


Adamant36:
Hi there. We generally leave out website links from the database. Since its available on the Wikidata page already, isn't reliable, and varies by store location etc. anyway.

tohaklim:
Thanks for the comments. alt_name is how they style it in marketing/signage, but not that important. Will remove the website. Not sure why the Belorussian name should be dropped?

Adamant36:
In that case its not really an alt name. Its the actual name. Otherwise what's the difference between that and Евроопт? What's on the signs goes in the brand and name values. 

As far as dropping the Belorussian, its because of the same reason I said in the other PR. We only go with two languages in the name index. The name on the signs and English if its not in English. I don't care if you drop Belorussian or Russian, but one of them needs to be dropped. If the signs are in Belorussian, id drop Russian. Otherwise, drop the Belorussian. It doesn't really matter which one. As long its one of them English at the end of it.

bhousel:
I removed the redundant tags in 7104c83 and merged..

PR #2482, opened 2019-03-17
BY post office

tohaklim:
These are two instances of the same poi in diff languages - merge?

### DIFF
--- a/brands/amenity/post_office.json
+++ b/brands/amenity/post_office.json
@@ -390,10 +390,12 @@
     "countryCodes": ["by"],
     "tags": {
       "amenity": "post_office",
-      "brand": "Белпочта",
+      "brand": "Белпошта",
       "brand:wikidata": "Q2662336",
       "brand:wikipedia": "en:Belposhta",
-      "name": "Белпочта"
+      "name": "Белпошта",
+      "name:be": "Белпошта",
+      "name:ru": "Белпочта"
     }
   },
   "amenity/post_office|Белпошта": {
@@ -401,7 +403,11 @@
     "tags": {
       "amenity": "post_office",
       "brand": "Белпошта",
-      "name": "Белпошта"
+      "brand:wikidata": "Q2662336",
+      "brand:wikipedia": "en:Belposhta",
+      "name": "Белпошта",
+      "name:be": "Белпошта",
+      "name:ru": "Белпочта"
     }
   },
   "amenity/post_office|Казпочта": {
--- a/brands/amenity/post_office.json
+++ b/brands/amenity/post_office.json
@@ -391,10 +391,13 @@
     "tags": {
       "amenity": "post_office",
       "brand": "Белпошта",
+      "brand:by": "Белпошта",
+      "brand:en": "Belpost",
       "brand:wikidata": "Q2662336",
       "brand:wikipedia": "en:Belposhta",
       "name": "Белпошта",
       "name:be": "Белпошта",
+      "name:en": "Belpost",
       "name:ru": "Белпочта"
     }
   },


Adamant36:
It would probably be good to do the same thing here as in the other one, except get rid of the Russian and add English instead. We don't generally add more then two languages to the index for an entry. The one the tag is and English.

tohaklim:
sorry, once again not sure why the presumed limit for multilingual names?

Adamant36:
> once again not sure why the presumed limit for multilingual names?

Because there's been a bunch of discussion about it already and that's the policy. Otherwise, I wouldn't be asking you to remove one. For the latest discussion see #2483 and https://github.com/osmlab/name-suggestion-index/pull/2483#issuecomment-473661622. which just took place yesterday.

bhousel:
> These are two instances of the same poi in diff languages - merge?

I think they should be merged into a single entry, likely the Belarusian one, but confirm with local mappers... This pull request doesn't actually merge them, so let us know what you want to do.

matkoniecz:
Is `amenity/post_office|Белпошта` limited to some countries (for example, only Belarus)?

tohaklim:
> > once again not sure why the presumed limit for multilingual names?
> 
> Because there's been a bunch of discussion about it already and that's the policy. Otherwise, I wouldn't be asking you to remove one. For the latest discussion see #2483 and [#2483 (comment)](https://github.com/osmlab/name-suggestion-index/pull/2483#issuecomment-473661622). which just took place yesterday.
ahh, right, that's how you do things. Happy to accept, but just want to note that this isn't the case of "let's translate the local name into all the languages in the world in case someone needs them localised", there are genuinely two official languages in Belarus, and since nobody cares they can be in RU or BE, or a combination of both. Having both values hence helps with search and on the ground usage. In this particular case, the BE name is used on official signage [here](https://www.mapillary.com/app/?lat=53.70526978751058&lng=23.840414760656504&z=17&pKey=Rzl8rKTIoUlMC5y1R-zlQQ&focus=photo&x=0.6911426918496781&y=0.12905573187197206&zoom=3) [and here](https://www.mapillary.com/app/?lat=53.702029134196884&lng=23.818647209577094&z=17&pKey=_ACJ_TLIFlUrxb3MDB-AsQ&focus=photo&x=0.6899660715447372&y=0.1618064494035508&zoom=1.3854429943022806) (again, generally), but not a rule. It's an ongoing argument for the local community, however (generally) if signage is in BE we use name:be, and vice versa for RU.

tohaklim:
> > These are two instances of the same poi in diff languages - merge?
> 
> I think they should be merged into a single entry, likely the Belarusian one, but confirm with local mappers... This pull request doesn't actually merge them, so let us know what you want to do.

as above, I'd go with name:be based on the signage

tohaklim:
> Is `amenity/post_office|Белпошта` limited to some countries (for example, only Belarus)?

yep, it's the national postal service

Adamant36:
> his isn't the case of "let's translate the local name into all the languages in the world in case someone needs them localised", there are genuinely two official languages in Belarus, and since nobody cares they can be in RU or BE, or a combination of both.

Then it really doesn't matter if its only in one language because everyone knows both right?

> In this particular case, the BE name is used on official signage here and here (again, generally), but not a rule.

Better to go with BE in this case then. Since its on the signs. Then in other cases where the signs are in Russian, there's no reason we can't with that. _Usually_ what's on the signs is preferred.

bhousel:
In 2721554 I ended up merging both entries into the single Belarusian "Белпошта". 
I also left both `:be` and `:ru` tags there because - it really doesn't matter, and it looks like both names are used in practice.

PR #2483, opened 2019-03-17
add h mart

yoasif:


Adamant36:
Thanks. For things like this we put the language of the company's title in the name field. Then name:*=whatever for the language its in. Then name:en=* for the English name. 

Here's an example. 

"name": "Banca Popolare di Milano",
 "name:en": "Popular Bank of Milan",
 "name:it": "Banca Popolare di Milano",

So that's what you need to do here and either get rid of the Korean or Chinese name also.

yoasif:
>For things like this we put the language of the company's title in the name field.

OK. I can fix that.

>In order to keep the index fairly simple though we only use two languages. The main one and English. If people need to figure out the name of the place in another language its usually on the Wikidata page.

Why? That is not mentioned in the contributor guide, and it feels kinda dumb if we have the information available. Do we want maps to have munge data from wikidata to get a localized map?

Adamant36:
> isn't that what I have? The "main" name is english, so the name:en is identical.

Is the brand called H Mart or 한아름 마트 or 韓亞龍 on the ground?

> Why? That is not mentioned in the contributor guide, and it feels kinda dumb if we have the information available. Do we want maps to have munge data from wikidata to get a localized map?

There's a lot of things that aren't in the contributor guidelines. It doesn't mean we don't do them that way though. The current discussion around official_name is a good example. There was talk of having a FAQ at one point that would include all this stuff, but it never happened for some reason. As for why exactly the policy is to not have a bunch of name:*= values for everything, you'd have to ask @bhousel. I can't remember the exact details, except for him wanting to keep the index not to big. Plus, I've had this conversation and the other one about a bunch of times already and it gets old having to explain things repeatedly. I don't mind doing it, but its easier if he just explains it if there's need for it. I wish people were more willing to accept things, but whatever. I'll probably try to revive the FAQ issue so there's a more official source for this. I know I'm just a measly, low level maintainer ;)

yoasif:
>Is the brand called H Mart or 한아름 마트 or 韓亞龍 on the ground?

H Mart

>I can't remember the exact details, except for him wanting to keep the index not to big. 

I don't get the point of it, when if we know the name of a brand in all locations in many languages, why we wouldn't attach it to each POI. It isn't as if this is a layer on top a *map*, it is inside of a map *editor* where each entry has to contain those entries to be present on a map.

Why have *any* entries for different languages at all? Just use wikidata and have the map display the correct wikipedia pages, names, whatever. The editor doesn't need any of it (and neither do the POIs) if we're shifting the responsibility of finding and displaying this data on an end user map software.

Adamant36:
> I don't get the point of it, when if we know the name of a brand in all locations in many languages, why we wouldn't attach it to each POI.

How many languages are there out there? I think providing names in more then a few languages is just out of the scope of the project. Its meant to help improve store tagging in iD Editor and data quality in OSM more generally, not act as a translation service. Plus, most stores will never be looked at on a map by people outside the country where the store is. For instance, although you technically could have the name of every restaurant in China translated into the African ZULU language, it doesn't mean you necessarily should (no offense ZULU speakers), because you just reach a point of diminishing returns after the first few languages main languages, where its not worth the effort or extra code. That's my thought's at least. @bhousel can probably better explain it. Its his project. I usually just go along with whatever he thinks is best for it. 

> Why have any entries for different languages at all?

Uuumm, a few are better then none? I forget what error checking tool it is, but one of them throws up errors if you don't have name:en with another name:* tag. One of the reasons we do it the way we do is because of that. 

> H Mart

So where is the Korean or Chinese name used? or are they not used at all but the tags for them are just being added because it's a Korean store?

yoasif:
>Its meant to help improve store tagging in iD Editor and data quality in OSM more generally, not act as a translation service.

It isn't acting as a translation service, the users providing the data are doing the translation (or the companies who run the entities) - it is simply helping data quality in OSM, just as you stated was the goal. 

>So where is the Korean or Chinese name used? or are they not used at all but the tags for them are just being added because it's a Korean store?

It is meant for users of OSM data who primarily speak Chinese and Korean who would likely be interested in these locations. Your prior example of Zulu users in China are probably minimal, but I strongly suspect that *many* Chinese and *especially* Korean users will appreciate being able to find H Mart in their native language. 

Not sure if you are from the USA either, but there is no official language in the US, and I see zero reason to not provide this affordance for Americans who prefer to not speak English.

Adamant36:
Some other people wanted to use both the Korean and Chinese names in PRs. Is there a lot of Chinese people in Korea or Korean people in China?

bhousel:
Thanks @yoasif  - as @Adamant36 said we don't add multiple name tags in this case.

@yoasif asked
> Why have any entries for different languages at all? Just use wikidata and have the map display the correct wikipedia pages, names, whatever. The editor doesn't need any of it (and neither do the POIs) if we're shifting the responsibility of finding and displaying this data on an end user map software.

You are correct about this.. The only situations we add localized (non-English) tag to the index are:
- if the `name` is not English, we also add `name:en` - this is so people everywhere know what it is (OSM is a predominantly English-speaking project)
- if we add `name:en`, we also add `name:xx` for some value that matches the `name` tag - this is because validation tools will nag the user if this value is not present, see #2245
- (these rules also apply for `operator` and `brand` tags)

But yes, even those rules introduce unnecessary redundancy when names can be easier looked up with Wikidata.  OSM has a lot of legacy expectations built into it and these rules are there to satisfy the OSM community and tools.  Someday OSM might realize that they don't need them, and that's ok.

bhousel:
Thanks again @yoasif  -  I merged this, and removed the international tags in b430177

PR #2487, opened 2019-03-18
SFR : Added infos

Lulucmy:
Added logo, wikidata, wikipedia and corrected name (different of brand)

### DIFF
--- a/brands/shop/mobile_phone.json
+++ b/brands/shop/mobile_phone.json
@@ -158,8 +158,14 @@
   },
   "shop/mobile_phone|Espace SFR": {
     "count": 52,
-    "tags": {
-      "brand": "Espace SFR",
+    "countryCodes": ["fr"],
+    "logos": {
+      "wikidata": "https://upload.wikimedia.org%2Fwikipedia%2Fcommons%2Fthumb%2F0%2F09%2FLogo_SFR_2014.svg%2F100px-Logo_SFR_2014.svg.png"
+    },
+      "tags": {
+      "brand": "SFR",
+      "brand:wikidata": "Q218765",
+      "brand:wikipedia": "en:SFR",
       "name": "Espace SFR",
       "shop": "mobile_phone"
     }


Adamant36:
I don't know if you need to add the logo that way. It does it through running the script anyway. Also, I think the brand and the name has to be the same. Usually if there's a short and long form of the name, the short form in used more. So its used in the name/brand fields and the long form is either added as a match or with the tag official_name. In this case name=SFR has 261 uses, while name=Espace SFR only 55. So SFR is clearly the preferred name and should be in the name/brand tags. Whereas, its fine to use the official_name tag for Espace SFR.

Lulucmy:
It's not the case here : "Espace SFR" is the name of the shop, the brand name is just "SFR", like "Apple"/"Apple Store". I'll remove the picture if needed

Adamant36:
> Espace SFR

Not all of them say Espace SFR. Some just say SFR. The general rule in cases like this is that we split the difference and go with the shorter name because it shows up better on the map and make use official_name if necessary, because while as all Espace SFR's have "SFR in their names, no SFR's have "Espace" As I said, SFR is used exponentially more anyway. Trust me, this comes up a lot and that's how we do it. It just came twice up yesterday a matter of fact. 

images

images (1)

images (2)

sfr-corporate_4079556

For stores that do have "Espace" in their name its not even featured predominantly. For instance, in this one you can hardly even see it on the right. 

download

Pretty much the same with this one

raw

Adamant36:
Apple stores aren't a good comparison because they don't even have a brand name in the front of their shops. Its just a picture of an apple. We always favor what the store front says. As its ground truth verifiable. 

26186-36845-applestore-thepier-l

Lulucmy:
> > Espace SFR
> 
> Not all of them say Espace SFR. Some just say SFR. The general rule in cases like this is that we split the difference and go with the shorter name because it shows up better on the map and make use official_name if necessary, because while as all Espace SFR's have "SFR in their names, no SFR's have "Espace" As I said, SFR is used exponentially more anyway. Trust me, this comes up a lot and that's how we do it. It just came twice up yesterday a matter of fact.
> 
> images
> 
> images (1)
> 
> images (2)
> 
> sfr-corporate_4079556
> 
> For stores that do have "Espace" in their name its not even featured predominantly. For instance, in this one you can hardly even see it on the right.
> 
> download
> 
> Pretty much the same with this one
> 
> raw

Didn't know all these rules, thanks 😉

Adamant36:
> Didn't know all these rules, thanks

Yep.

bhousel:
I added "match": ["shop/mobile_phone|Espace SFR"],  to the existing SFR entry, and merged

PR #2495, opened 2019-03-20
Make popeyes a takeaway = yes

yoasif:


bhousel:
weird.. I wonder if we should just set this for all of the `amenity=fast_food`?
and `amenity=cafe`
It's kind of a silly tag when you think about it.

yoasif:
I was thinking maybe but wanted to test the waters with a smaller PR first.

tas50:
In my mind part of a place being fast food is the ability to have your food packaged to go. There may be some exceptions to that in the fast food tag, but ideally not.

bhousel:
> I was thinking maybe but wanted to test the waters with a smaller PR first.

Thats ok, but going forward, open an issue first to ask and discuss instead?
PRs are helpful, but they can introduce extra work for us.

If we decide to just set `takeaway` tags for certain types, that's a thing I can do in like 10 seconds.

PR #2497, opened 2019-03-20
Add Unleased by Petco

tas50:
These are popping up all over the place in the US and they're tagged super inconsistently. I gave it an alt_name for search since the signs use the "by Petco" bit and they have it in a lot of their branding, but not the point where I'd say it's an official_name.


bhousel:
Thanks @tas50 👍  seems fine!

PR #2499, opened 2019-03-21
 Added National Bank and First Bank

bonbon12:


brands/amenity/bank.json
```diff
@@ -3273,6 +3275,8 @@
     "tags": {
       "amenity": "bank",
       "brand": "National Bank",
+      "brand:wikidata": "Q634298",
+      "brand:wikipedia": "en:National Bank of Canada",
```
ENT8R: This seems to be the wrong entry.
The canadian bank has been added with https://github.com/osmlab/name-suggestion-index/commit/42cba58c53de61cff05e937eeec2c41078638d00
This bank actually seems to be rather a generic (?) name when looking at Overpass Turbo: https://overpass-turbo.eu/s/Hf5 Though there are actually some in Canada.
Could you maybe say which one is more often used in Canada: "National Bank" or "Banque Nationale"?

brands/amenity/bank.json
```diff
@@ -2534,6 +2534,8 @@
     "tags": {
       "amenity": "bank",
       "brand": "First Bank",
+      "brand:wikidata": "Q5452217",
+      "brand:wikipedia": "en:FirstBank Holding Co",
```
ENT8R: This should probably be split up into two different entries. [Overpass Turbo](https://overpass-turbo.eu/s/Hf7) shows that this exists in the USA, Nigeria and even some in Romania. You are referencing only the bank in the US, right?

matkoniecz: @bonbon12 Ping?

bonbon12: Yes it was just the bank in the US.

brands/amenity/bank.json
ENT8R: @bonbon12 Are you living in Canada and can answer my question? Or did you just researched it on the internet?

bonbon12: I live in Canada, and its referred to as National Bank.

brands/amenity/bank.json
ENT8R: Alright. I can also add the other entries after the merge, but there is another problem:
The wikipedia article (https://en.wikipedia.org/wiki/FirstBank_Holding_Co) says that this bank operates only in three western states of the US. But according to Overpass Turbo there are also banks with that name in eastern states and even in Puerto Rico. There is for example this local bank in Michigan: https://www.first-bank.com/
What is the best thing we can do in that case @bhousel?

bhousel: Hmm yeah "First Bank" is common enough that it's sure to have several separate brands that call themselves that.  I'll merge and add a disambiguation test after the name, like "amenity/bank|First Bank~(Colorado)" so that people in other parts of the USA know not to use it.  (Showing logos helps a lot here too).

We already have a few different ShopRite supermarket brands in the US, so this is not that unheard of.

PR #2502, opened 2019-03-23
Added wikidata to 4 British brands

ldw606:
I noticed an problem with the overpass link generated for brands with an "&" in their name while checking the B&M tagging, I'll open an issue for it.

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -373,6 +373,8 @@
     "count": 475,
     "tags": {
       "brand": "Costcutter",
+      "brand:wikidata": "Q5175072",
+      "brand:wikipedia": "en:Costcutter",
       "name": "Costcutter",
       "shop": "convenience"
     }
--- a/brands/shop/gift.json
+++ b/brands/shop/gift.json
@@ -11,8 +11,11 @@
   },
   "shop/gift|Clintons": {
     "count": 56,
+    "countryCodes": ["gb"],
     "tags": {
       "brand": "Clintons",
+      "brand:wikidata": "Q5134299",
+      "brand:wikipedia": "en:Clintons",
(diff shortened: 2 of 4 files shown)


brands/shop/convenience.json
```diff
@@ -373,6 +373,8 @@
     "count": 475,
     "tags": {
       "brand": "Costcutter",
+      "brand:wikidata": "Q5175072",
+      "brand:wikipedia": "en:Costcutter",
```
ENT8R: This also exists in the supermarket section: https://github.com/osmlab/name-suggestion-index/blob/5e6f2a4e6441f3fc74d2174d91e2239e491e8dfc/brands/shop/supermarket.json#L776-L785

Do you know whether this store is considered rather as a supermarket or a convenience store?
Judging only from the amount of mapped places, I would say convenience is the right one...

Adamant36: According to Wikipedia "It encompasses both supermarkets and convenience shops" because of mergers and whatnot. I guess they run a bunch of former SPAR locations in the United Kingdom etc. So both entries are correct.

matkoniecz: I will add nomatches after merging PR.

https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#nomatch

matkoniecz:
Thanks for the pull request!

brands/shop/convenience.json
ldw606: Most of the Costcutters I've seen mapped as supermarkets are what I would call convenience stores, though some of them are large enough to be small supermarkets so both tags should be kept . Wikipedia says they have some stores in Poland as well which might be supermarkets but none of those have been mapped so far.

Issue #2503, opened 2019-03-23
Ampersands not escaped in overpass links in docs


ldw606:
Brands with an & in their name cause incorrect overpass links to be generated in the docs.
This can be seen on the B&M Bargains entry in the [variety_store](https://osmlab.github.io/name-suggestion-index/brands/shop/variety_store.html) section where the link generated is
https://overpass-turbo.eu/?Q=%5Bout:json%5D%5Btimeout:25%5D;%0A(nwr%5B%22shop%22=%22variety_store%22%5D%5B%22name%22=%22B&M%20Bargains%22%5D;);%0Aout%20body;%0A%3E;%0Aout%20skel%20qt;&R
but should be
https://overpass-turbo.eu/?Q=%5Bout:json%5D%5Btimeout:25%5D;%0A(nwr%5B%22shop%22=%22variety_store%22%5D%5B%22name%22=%22B%26M%20Bargains%22%5D;);%0Aout%20body;%0A%3E;%0Aout%20skel%20qt;&R
with a %26 instead of the &

ENT8R:
Fixed with https://github.com/osmlab/name-suggestion-index/commit/b076a33e55626dffd25a3079c093e3efa09fd564

PR #2504, opened 2019-03-23
Overpass link fix

ldw606:
This should fix issue #2503, there were 59 links with the problem as far as I could tell.
This will change all of the overpass links to a form with more of the characters escaped, I've tested several links and they all seem to work now.

Since this is the first time I've created and solved an issue on GitHub please let me know if I've done anything wrong.

### DIFF
--- a/build_docs.js
+++ b/build_docs.js
@@ -184,7 +184,7 @@ function overpassLink(k, v, name) {
 out body;
 >;
 out skel qt;`);
-    let href = `https://overpass-turbo.eu/?Q=${q}&R`;
+    let href = `https://overpass-turbo.eu/?Q=${q.replace(`&`, `%26`)}&R`;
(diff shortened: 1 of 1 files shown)


ENT8R:
Good catch :+1:
`encodeURI()` should only be used for a whole URI and `encodeURIComponent()` a part of an URI, which is obviously what we want in this case.

matkoniecz:
> Since this is the first time I've created and solved an issue on GitHub please let me know if I've done anything wrong.

Everything looks well!

One hint: You can add "fixes #XXX" as part of commit title/message and it will close that issue automatically once PR is merged - see https://help.github.com/articles/closing-issues-using-keywords/ for a description of this trick.

ldw606:
Thanks for the hint. I thought something like that existed but didn't know the exact format it needed to be in.

PR #2507, opened 2019-03-23
Wikidata and Wikipedia tags for some supermarkets/convenience stores

hfs:
Add "shop/supermarket|abc~(Poland)" as own chain for Poland. Most are tagged as shop=convenience, but some as shop=supermarket as well. As "ABC" seems to appear for supermarkets elsewhere in the world, leave that entry as it is.

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -177,6 +177,8 @@
     "count": 303,
     "tags": {
       "brand": "CBA",
+      "brand:wikidata": "Q779845",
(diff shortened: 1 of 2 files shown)


Adamant36:
Hi there. Thanks for the PR. Looks good. When your adding new Wikidata entries if you don't mind checking that they have links to their Facebook, Twitter, and Instagram accounts on the Wikidata pages if they have any it would be much appreciated. That way the icons can be used in iD Editor once there is support added for them. Although, its not obligatory it does save us some time adding them later. Thanks.

PR #2508, opened 2019-03-23
wikidata and wikipedia for DB Reisezentrum

hfs:
Brand of railway ticket shops of the German national railway operator Deutsche Bahn found mostly in train stations.

### DIFF
--- a/brands/shop/ticket.json
+++ b/brands/shop/ticket.json
@@ -12,6 +12,8 @@
   },
   "shop/ticket|DB Reisezentrum": {
     "count": 52,
+    "countryCodes": ["de"],
+    "match": ["shop/ticket|Reisezentrum"],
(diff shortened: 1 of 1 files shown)


brands/shop/ticket.json
```diff
@@ -14,6 +14,8 @@
     "count": 52,
```
ENT8R: I would also recommend to add the following values:

```suggestion
     "count": 52,
     "countryCodes": ["de"],
     "match": ["shop/ticket|Reisezentrum"],
```

hfs:
You’re too fast for me :smile: When I just _check_ for notifications you’ve already done the work. Thanks!

bhousel:
@ENT8R is on fire 🔥

PR #2513, opened 2019-03-25
Add Orangetheory Fitness

matthewlw:


maxerickson:
Looks fine. Can you go ahead and also add countryCodes? Software using the index can use that info to limit where a given name is suggested.

matthewlw:
I added all the countries they had listed on their website; while it seems they have a much smaller presence outside the US, this should be complete.

bhousel:
Thanks @matthewlw 👍

Issue #1069, opened 2018-10-16
Research needed for: shop/books|Cultura


bhousel:
#### :thinking: We need help deciding what to do with "shop/books|Cultura"

:point_right: Read the [CONTRIBUTING.md](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) guide for more info, and feel free to ask questions on this issue!

### What's needed

First, do some research on `shop/books|Cultura` by searching Google and Wikipedia.

#### If it is an actual brand:

* Edit `config/canonical.json`:
  * Find the entry for `shop/books|Cultura`
  * Add tags for `brand:wikidata` and `brand:wikipedia`
  * Optionally add `"countryCodes"` or other tags
* `npm run build`
* Commit your changes and submit a pull request

#### If it is not an actual brand:

* Edit `config/filters.json` to add an expression that discards it
* Edit `config/canonical.json` to remove the entry for `shop/books|Cultura`
* `npm run build`
* Commit your changes and submit a pull request

#### If the situation is complicated:

* Leave a comment on this issue with what you find

:sparkles: Thanks!

_This is an autogenerated ticket_

bmillemathias:
From my experience Cultura is not only books, but also have consoles and games, music goods like CD but also instruments. They also have in my local shop a part dedicated to manual art, like paint or crafting paper, carton... They selling press and concert tickets.

Cultura are I know is french only based shops, seems to be confirmed by https://www.cultura.com/les-magasins.html/

Adamant36:
@matkoniecz, you mind if this is closed since you added it?

PR #2515, opened 2019-03-25
Add Cultura shop wikipedia and wikimedia information.

bmillemathias:
To complete issue #1069

### DIFF
--- a/brands/shop/books.json
+++ b/brands/shop/books.json
@@ -43,9 +43,12 @@
     }
   },
   "shop/books|Cultura": {
-    "count": 64,
+    "count": 91,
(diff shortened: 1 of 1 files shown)


matkoniecz:
http://overpass-turbo.eu/s/Hn3 confirms

Note that changing count manually is not necessary, it is something done automatically.

@bhousel - maybe we also should remove count field from human editable file? It is useful solely during research how many uses are out there and it is always necessary to use Overpass anyway, so it is not too useful.

And people may think that it deserves being edited by humans.

matkoniecz:

PR #2522, opened 2019-03-27
Added country code, no wiki tags found

Johnnybar:
The only addition I made was a country code, as no wiki tags could be found. 
I've cloned the forked repo, and pushed to it. As this is my first contribution, I'll be happy to know if this was the correct way of opening the pull request.

### DIFF
--- a/brands/shop/jewelry.json
+++ b/brands/shop/jewelry.json
@@ -103,7 +103,7 @@
     }
   },
   "shop/jewelry|Histoire d'Or": {
-    "countryCodes": ["fr"],
+    "countryCodes": ["fr", "nl", "be"],
     "count": 63,
     "tags": {
       "brand": "Histoire d'Or",
--- a/brands/shop/jewelry.json
+++ b/brands/shop/jewelry.json
@@ -103,6 +103,7 @@
     }
   },
   "shop/jewelry|Histoire d'Or": {
+    "countryCodes": ["fr"],
     "count": 63,
     "tags": {
       "brand": "Histoire d'Or",


matkoniecz:
> I've cloned the forked repo, and pushed to it. As this is my first contribution, I'll be happy to know if this was the correct way of opening the pull request.

Yes, that is exactly how PRs work. https://help.github.com/en/articles/about-pull-requests has some further explanation, including some Github-specific things.

matkoniecz:
From http://overpass-turbo.eu/s/Hqm it seems that this name appears also in Belgium and Netherlands.

Is it a separate brand?

Johnnybar:
Based on my research, it is the same brand with branches in these countries as well. Their FB pages lead to the same Histoire D'or website and the various branches locations can be found under this page: https://www.histoiredor.com/fr/nos-bijouteries/nos-bijouteries.
But you are right, additional country codes for these countries should be added. Thanks for the comments.

matkoniecz:
Then please, add country codes of other locations where this shop may appear.

Johnnybar:
Done. Final question before pushing - the definition for count in your contributing page is 'generated # of these that we found in OpenStreetMap", which means this was automatically generated and shouldn't be edited by me, correct?

matkoniecz:
> which means this was automatically generated and shouldn't be edited by me, correct?

Yes, exactly.

matkoniecz:
If there is no Wikipedia article one may find (or create) and link just a Wikidata entry.

If you would be interested - https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#memo--edit-wikidata should have explanation.

If it is not clear how it can be done - please comment here or create a new issue requesting documentation improvements,

matkoniecz:

Johnnybar:
Perfect, happy to hear!

ENT8R:
> If there is no Wikipedia article one may find (or create) and link just a Wikidata entry.

I just did that with https://github.com/osmlab/name-suggestion-index/commit/cf857202ed5aebc3d0094839938d5678a1081ea6. I also added another country code because they say on their website that they have even some shops in Italy...

PR #2525, opened 2019-03-28
Add funeral director company Roc-Eclerc

bmillemathias:
https://overpass-turbo.eu/s/HrU

### DIFF
--- a/brands/shop/funeral_directors.json
+++ b/brands/shop/funeral_directors.json
@@ -1,4 +1,16 @@
 {
+  "shop/funeral_directors|Roc-Eclerc": {
+    "match": [
+      "shop/funeral_directors|Roc Eclerc"
+    ],
+    "nocount": true,
+    "tags": {
+      "brand": "Roc-Eclerc",
+      "brand:wikidata": "Q62558102",
+      "name": "Roc-Eclerc",
+      "shop": "funeral_directors"
+    }
+  },
   "shop/funeral_directors|The Co-operative Funeralcare": {
     "count": 124,
     "countryCodes": ["gb"],


brands/shop/funeral_directors.json
```diff
@@ -9,5 +9,13 @@
       "name": "The Co-operative Funeralcare",
       "shop": "funeral_directors"
     }
+  },
+  "shop/funeral_directors|Roc-Eclerc": {
+    "tags": {
+      "brand": "Roc-Eclerc",
+      "brand": "Q62558102",
```
matkoniecz: It should be rather `brand:wikidata`.

bmillemathias: indeed.

matkoniecz:
"Roc Eclerc" is the same but using incorrect version of name, right? https://overpass-turbo.eu/s/HrW

In that case adding `match` would be a good idea - see https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#card_file_box--about-the-brand-files

bmillemathias:
> 
> 
> "Roc Eclerc" is the same but using incorrect version of name, right? https://overpass-turbo.eu/s/HrW
> 
> In that case adding `match` would be a good idea - see https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#card_file_box--about-the-brand-files

I did a check before submittting the PR and Roc-Eclerc is the correct name in the french registry of companies so I wanted to correct "Roc Eclerc" in "Roc-Eclerc" in OSM but sure, I can do that also.

matkoniecz:
Adding `"match"` property will help with exactly that - it is used to mark different names as an incorrect variants.

See https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#card_file_box--about-the-brand-files

`amenity=fast_food name=McDonald's` is a correct version, `amenity=fast_food name=Mc Donald's` and some other are recognized as matching variants.

matkoniecz:
Thanks for a PR!

PR #2527, opened 2019-03-28
Add Funeral Director Pompes Funèbres Générales

bmillemathias:
PFG: https://overpass-turbo.eu/s/HrX
Pompes Funèbres Générales: https://overpass-turbo.eu/s/HrY

### DIFF
--- a/brands/shop/funeral_directors.json
+++ b/brands/shop/funeral_directors.json
@@ -1,4 +1,18 @@
 {
+  "shop/funeral_directors|PFG": {
+    "countryCodes": ["fr"],
+    "match": [
+      "shop/funeral_directors|Pompes Funèbres Générales"
+    ],
+    "nocount": true,
+    "tags": {
+      "brand": "PFG",
+      "brand:wikidata": "Q3396087",
+      "brand:wikipedia": "fr:Pompes funèbres générales",
+      "name": "Pompes Funèbres Générales",
+      "shop": "funeral_directors"
+    }
+  },
   "shop/funeral_directors|The Co-operative Funeralcare": {
     "count": 124,
     "countryCodes": ["gb"],


Adamant36:
Hi there. If they are both tagged with the name tag, you should just pick whatever one has the most usage and add then other one as a match instead of using the brand. Either that or go with the shorter name, as it shows up better on a map, and use the official_name tag for the longer one. Which is what id do in this case.

matkoniecz:
>  you should just pick whatever one has the most usage

Rather the better one (almost always correct `name` tag value - one used by people to refer to the object is the most common one but in some situations it is not happening).

matkoniecz:
To confirm:

`name=PFG` should be used in OSM rather than `name=Pompes Funèbres Générales`, right?

bmillemathias:
> To confirm:
> 
> `name=PFG` should be used in OSM rather than `name=Pompes Funèbres Générales`, right?

From my experience, ads mention PFG as the name so I would use PFG as I did in my commit.

Adamant36:
"From my experience, ads on TV mentioned PFG as the name so I would use PFG as I did in my commit."

That's what I thought. In your commit it says the name tag is "Pompes Funèbres Générales". 

So you should update it to have both the name and brand tags be PFG. Then as I said above you can use official_name for Pompes Funèbres Générales or just add it as a match if there's any usage of it in OSM. 

"Rather the better one"

Which is what I said, in this situation PFG. I didn't say disregard everything else for usage numbers. It is a factor though. So the names length and how people tag it. Both of which I also mentioned.

brands/shop/funeral_directors.json
```diff
@@ -1,4 +1,18 @@
...
+    "countryCodes": ["fr"],
+    "match": [
+      "shop/funeral_directors|Pompes Funèbres Générales"
+    ],
+    "nocount": true,
+    "tags": {
+      "brand": "PFG",
+      "brand:wikidata": "Q3396087",
+      "brand:wikipedia": "fr:Pompes funèbres générales",
+      "name": "Pompes Funèbres Générales",
```
matkoniecz: It seems that it should be

```
"official_name": "Pompes Funèbres Générales",
"name": "PFG",
```

according to other data and discussion.

bhousel: I merged and did this..

PR #2528, opened 2019-03-28
Two entries added and tags added

Johnnybar:
Added two entries for the same brand in newsagent.json. Added match property to suppress warning. Added country code, the appropriate wiki tags and additional information to wikidata.


brands/shop/newagent.json
```diff
@@ -0,0 +1,32 @@
...
+    "nocount": true,
+    "tags": {
+      "brand": "Martin's",
+      "brand:wikidata": "Q7277785",
+      "brand:wikipedia": "en:RS McColl",
+      "name": "Martin's",
+      "shop": "newsagent"
+    }
+  }
+}
```
ENT8R: This file seems wrong to me. It is not called `newagent` but rather `newsagent`...

brands/shop/convenience.json
```diff
@@ -780,8 +780,15 @@
   },
   "shop/convenience|Martin's": {
     "count": 53,
+    "countryCodes": ["gb"],
+    "match": [
+      "shop/newsagent|Martin's",
+      "shop/newsagent|McColls"
+    ],
```
ENT8R: Well, if the entry is the same as `shop/newsagent|Martin's` and `shop/newsagent|McColls` then why did you created another entry for that in the `brands/shop/newsagent.json` file?

brands/shop/newsagent.json
```diff
@@ -41,6 +41,36 @@
       "shop": "newsagent"
     }
   },
+  "shop/newsagent|Martin's": {
+    "count": 53,
```
ENT8R: This entry did not exist before, so it should be
```suggestion
    "nocount": true,
```

brands/shop/newsagent.json
```diff
@@ -41,6 +41,36 @@
...
+    "tags": {
+      "brand": "Martin's",
+      "brand:wikidata": "Q7277785",
+      "brand:wikipedia": "en:RS McColl",
+      "name": "Martin's",
+      "shop": "newsagent"
+    }
+  },
+  "shop/newsagent|McColls": {
+    "count": 53,
```
ENT8R: This entry did not exist before too, so it should also be
```suggestion
    "nocount": true,
```

brands/shop/convenience.json
ENT8R: According to https://en.wikipedia.org/wiki/McColl%27s the Martin's store should be tagged as `shop/newsagent|Martin's` and not as `shop/convenience|Martin's`. So I would tend to remove this entry and leave the other one (with the correct match)

brands/shop/newsagent.json
```diff
@@ -41,6 +41,36 @@
       "shop": "newsagent"
     }
   },
+  "shop/newsagent|Martin's": {
+    "count": 53,
+    "countryCodes": ["gb"],
+    "match": [
+      "shop/convenience|Martin's",
+      "shop/newsagent|McColls"
```
ENT8R: This is probably the wrong match, as there is an additional entry for this.

brands/shop/newsagent.json
```diff
@@ -41,6 +41,36 @@
...
+    ],
+    "tags": {
+      "brand": "Martin's",
+      "brand:wikidata": "Q7277785",
+      "brand:wikipedia": "en:RS McColl",
+      "name": "Martin's",
+      "shop": "newsagent"
+    }
+  },
+  "shop/newsagent|McColls": {
```
ENT8R: According to https://en.wikipedia.org/wiki/McColl%27s this should not be tagged as `shop=convenience` but rather as `shop=newsagent`. So I would consider to remove this entry and leave the other one with the correct match.

brands/shop/newagent.json
Johnnybar: Can you specify where you saw 'newagent'?

brands/shop/newagent.json
```diff
@@ -0,0 +1,32 @@
+{
+  "shop/newagent|Martin's": {
```
bhousel: the `newagent` here in the id is causing it to generate a new file called `newagent.json`

Johnnybar: I see the new file was created, but can't find any mention of 'newagent' in the shop/newsagent file... should I delete the newagent.json file and run npm run build again?

bhousel: yes delete the file, and run `npm run build` again.

brands/shop/newsagent.json
ENT8R: I think I mixed something up in my last comment. Sorry.
I meant that `McColls` should not be tagged as `shop=newsagent` but rather as `shop=conveniece`. (The other way round as in my first comment)
At least Wikipedia mentions that they are

> trading under the trading nam[e] McColls (for convenience shops)

Johnnybar: I see what you mean and I changed Martin's to newsagent and McColl's to convenience. When I run npm run build, though, an earlier result of 
`"shop/convenience|Martin's": {
    "count": 53,
    "tags": {
      "brand": "Martin's",
      "name": "Martin's",
      "shop": "convenience"
    }
  },`

keeps coming up in addition to these two entries, even after I delete it. Should I delete it and push as-is, without run build?

ENT8R: If you can, you should always run the build script before pushing a commit, because sooner or later somebody else will do this and then there might be a commit with a change that was not intentional...
The problem in your case is that you use the wrong property in the `match` field.
For `shop/newsagent|Martin's` it should be the following:
`"match": ["shop/convenience|Martin's"],`
and for `shop/convenience|McColls` it should be:
`"match": ["shop/newsagent|McColls"],`
Try these instead, delete the old, unwanted entries and then run again the build script.

Johnnybar: Followed your change requests, then removed the duplicate "McColl's" entry and added "nocount":true to the new "McColls" entry as previously requested. However,  running the build script regenerates the deleted "McColl's" entry and changes "nocount" to "count: 98" in the new "McColls" entry.

ENT8R: Is "McColl's" or "McColls" the preferred name?
I'm not sure because the Wikipedia article mentions both names but the current data of OSM indicates that "McColl's" (286 uses) is more used than "McColls" (98 uses)...

bhousel:
Very interesting that they operate both McColl's and Martin's, and also apparently RS McColl too.  I'll merge like this and remove the duplicate "McColls" which seems like a common mistake that mappers make.

PR #2531, opened 2019-03-29
Update "Co-op" brand in UK

maingockien01:
Hi, I changed the tags data (brand, brand:wikidata and brand:wikipedia) of Co-op in 2 files supermarket.json and convenience.json. Also, I changed the name from Co-op to Co-op Food because the latter seems more better and "UK belongs to" than the previous one.

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -313,7 +313,7 @@
       "shop": "convenience"
     }
   },
-  "shop/convenience|Co-op": {
+  "shop/convenience|Co-op Food": {
     "count": 275,
     "match": [
       "shop/convenience|COOP",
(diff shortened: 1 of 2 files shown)


Adamant36:
I'm wondering why you didn't just create a new entry for the UK one and then added country codes. Instead of removing the one in Canada in the process. Which is still a legitimate business. 

Also, Co-op Food for the brand should probably just stay Co-op, as "brand" is usually the same as the name tag and it seems to be tagged as Co-op, not Co-op food. Then you can use the "official_name" tag for Co-op Food. If its the business I found on Google it should actually be "The Coperative Food" since that's what is on the store signs. Then maybe Co-op Food as a match if anyone tags them that way.

maingockien01:
Sorry, I didnt really think throughly. That's my fault. I will change it. For that, I will close this pull request.

Adamant36:
It's cool. It happens. You didn't have to close the PR. You could have just changed the files and recommitted it. Closing it works to though ;)

maingockien01:
Sorry for bothering you but I want to make my thought clear before starting. Apparently, we already have "The Co-operative Food" tag which is the "Co-op" tag in UK we talked before. So I am going to add more match tag into "The Co-operative Food" tag instead of making a new one, and add nomatch tag for "Co-op" Canada tag and "The Co-operative" tag. Is it okay?

PR #2534, opened 2019-03-30
Fix Co-op bug 

maingockien01:
The Co-op is actually the brand in Canada so I add countryCodes tag and add nomatch tag to make it not conflict with the UK The Co-operative. While, in "The Co-operative" tag, which is real Co-op in Uk, I add more match tag to receive the "Co-op" search and, similarly, I add more nomatch tag.


brands/shop/convenience.json
```diff
@@ -31,17 +31,6 @@
       "shop": "convenience"
     }
   },
-  "shop/convenience|76": {
```
ENT8R: Is there a reason why you removed this?

SilentSpike: I imagine this was just a git accident since it was recently added

maingockien01: Sorry, that was a git accident, i just merge git so now shop 76 is not removed.

brands/shop/supermarket.json
```diff
@@ -649,6 +649,7 @@
     }
   },
   "shop/supermarket|Co-op": {
+    "countryCodes":["ca"],
```
ENT8R: BTW: there is also a supermarket called `Coop` in some european countries: https://en.wikipedia.org/wiki/Coop_(Switzerland)
This probably deserves an extra entry because it is a different brand than this one...

maingockien01: hmm I see. Okay I will add them soon. However, I need to check there are how many Coop, in fact, first.

matkoniecz:

If you would be interested - further PR will be also welcomed!

Are you maybe especially interested in fixing issues specific to some specific countries?

brands/shop/supermarket.json
bhousel: I added this in 75452f4 and added disambiguation text and resolved all the remaining warnings.

Now we support:
      "amenity/fuel|Co-op~(Canada)",
      "shop/convenience|Coop~(Schweiz)",
      "shop/convenience|The Co-operative Food~(UK)",
      "shop/supermarket|Co-op~(Canada)",
      "shop/supermarket|Coop~(Schweiz)",
      "shop/supermarket|The Co-operative Food~(UK)"

PR #2536, opened 2019-03-30
Add Ralph's Italian Ices

bripmccann:
Not totally sure about naming here. Open to changes.
- **The common name is "Ralph's Italian Ices".** If asked for something more than "Ralph's", I think this is the name most people would give.
- **The official name is "Ralph's Famous Italian Ices".** (On paper, ["Ralph's Famous Italian Ices Franchise Corp."](https://opencorporates.com/companies/us_ny/2567819)) This is the name on smaller logos.
- **On larger logos, they write "Ralph's Famous Italian Ices & Ice Cream".** Again, this doesn't seem to be their legal name, but it is a [registered trademark](https://tsdr.uspto.gov/#caseNumber=77726943&caseSearchType=US_APPLICATION&caseType=DEFAULT&searchType=statusSearch).
- I also found "Ralph's Italian Ices & Ice Cream" and "Ralph's Italian Ice" (without the 's' at the end) in use on <del>JOSM</del> <ins>Overpass Turbo</ins>.

My best interpretation: I used "Ralph's Italian Ices" for `"brand"` and `"name"`, "Ralph's Famous Italian Ices" for `"official_name"`, and only added "Ralph's Famous Italian Ices & Ice Cream" under `"match"`.

### DIFF
--- a/brands/amenity/ice_cream.json
+++ b/brands/amenity/ice_cream.json
@@ -91,6 +91,24 @@
       "name": "La Michoacana"
     }
   },
+  "amenity/ice_cream|Ralph's Italian Ices": {
+    "countryCodes": ["us"],
+    "match": [
+      "amenity/ice_cream|Ralph's Famous Italian Ices",
+      "amenity/ice_cream|Ralph's Famous Italian Ices & Ice Cream",
+      "amenity/ice_cream|Ralph's Italian Ice",
+      "amenity/ice_cream|Ralph's Italian Ices & Ice Cream"
+    ],
+    "nocount": true,
+    "tags": {
+      "amenity": "ice_cream",
+      "brand": "Ralph's Italian Ices",
+      "brand:wikidata": "Q62576909",
+      "cuisine": "ice_cream",
+      "name": "Ralph's Italian Ices",
+      "official_name": "Ralph's Famous Italian Ices"
+    }
+  },
   "amenity/ice_cream|Rita's Italian Ice": {
     "countryCodes": ["us"],
     "match": [


Adamant36:
Going by usage numbers there's 41 instances of Ralph's. Whereas, there is only 6 instances of Ralph's Italian Ices. Plus, Ralph's just looks better on a map and the "Italian Ice" part looks more a sufix in their signs. So personally, id go With Ralph's for the brand/name tags and Ralph's Italian Ice for the official_name tag. That's just me though. 

In general, whatever you go with for brand/name its usually good to go with a combination of A. What's on the signs B. What has the most usage C. the one that's easiest to read on the map. 

Although, which is most important out of the three is up for debate and depends on the situation. Things like what name they give in trademarks or FFC fillings (or whatever it is)  is generally way less important.

bripmccann:
Yeah, maybe it's a little silly to look at official documents and whatnot.

Maybe I'm using Overpass wrong (I'm learning) but I'm actually only getting three results for "Ralph's Italian Ices".

I do get 41 for "Ralph's", but none of them are tagged `"amenity": "ice_cream"`, so I don't know which are part of this chain. I imagine "Ralph's" is a pretty common name.

Those results are also all from all over the place, but I think this chain is only in New York and New Jersey. (It's not a big chain. Some site said there were 72 locations. I wasn't sure if I should add this at all, but the CONTRIBUTING doc says "If it exists, we want to know about it!". That seems very much in the spirit of OSM.)

bhousel:
Thanks @bripmccann & @Adamant36 - this looks ok to me!
(There is also a [Ralph's](https://www.wikidata.org/wiki/Q3929820) supermarket, but we have this in the index already)

PR #2538, opened 2019-03-31
Add a 'wikicheck' script

### LINKED ISSUE #1990: add script check whatever wikipedia, wikidata data for links to deleted, redirected or mismatched pages
It should check whatever

- wikidata entry exists and is not a redirect
- wikipedia entry exists and is not a redirect
- wikipedia entry is linked to the same wikidata as wikidata entry
- for pages without wikipedia entry, but with wikidata entry - is wikidata entry matched to any wikipedia page (with language code other than `ceb`).

ENT8R:
I wrote a new script which does several things:
- detect wrong formatted Wikidata and Wikipedia values (`brand:wikidata`, `operator:wikidata`, `brand:wikipedia`, `operator:wikipedia`)
- detect (possible) Wikipedia articles for a Wikidata entity, if a Wikipedia value is not yet present
- detect wrong Wikidata and Wikipedia pairs (i.e. sitelinks of the Wikidata entity doesn't match the specified Wikipedia article)
- detect deleted entities or redirects of entities

Maybe this should also be mentionned in the `CONTRIBUTING.md` file?

 
 

This PR fixes #1990

### DIFF
--- a/check_wikiTags.js
+++ b/check_wikiTags.js
@@ -1,0 +1,217 @@
+const clearConsole = require('clear');
+const colors = require('colors/safe');
+const fetch = require('node-fetch');
+const fileTree = require('./lib/file_tree');
+const wdk = require('wikidata-sdk');
+
+let _brands = fileTree.read('brands');
+
+let _errors = [];
+
+let _wrongFormat = [];
+let _deleted = [];
+let _foundSitelink = [];
+let _wrongLink = [];
+
+let _wikidata = gatherQIDs(_brands);
+let _qids = Object.keys(_wikidata);
+let _total = _qids.length;
+if (!_total) {
+    console.log('Nothing to fetch');
+    process.exit();
+}
+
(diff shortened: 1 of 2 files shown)


bhousel:
This looks so awesome!   I love it, thanks @ENT8R  🎉

matkoniecz:
Thanks, that is awesome! I worried that it may be useful but it returned more entries than I expected.

> Maybe this should also be mentionned in the CONTRIBUTING.md file?

Yes, that would be a great idea!

ENT8R:
> but it returned more entries than I expected.

I was also surprised by the amount, so I took another look at the code. I actually found a bug which led to a huge amount of false-positives in the "Wrong Wikipedia article which is not linked to the Wikidata entity"-section...
But this is fixed with https://github.com/osmlab/name-suggestion-index/commit/e71dd00fc42534c69bed830c563919d138c8cee3

PR #2542, opened 2019-04-01
ran npm wikidata/docs

Adamant36:
@bhousel, was this suppose to remove all the twitter image links?


tas50:
If you don’t have the API key set it nukes it. It should probably just fail if the key isn’t set really

Adamant36:
Hhmmm..I was thinking it might be something like that. So you think I should close it and let someone else do it then? 

I agree that it would be better if it just failed without a key set.

bhousel:
@tas50 said:
> If you don’t have the API key set it nukes it. It should probably just fail if the key isn’t set really

Oh yeah, sorry!  I think the script used to do this.
@Adamant36 I'm going to close this PR and just rerun it myself tonight, and I'll fix the build script to not nuke existing twitter entries if there is no credentials.

Adamant36:
@bhousel, that works. Thanks.

PR #2546, opened 2019-04-05
Add wikidata + wikipedia info for convenience Proxi

bmillemathias:
Proxi is more a convenience shop rather than a supermarket so delete duplicate shop/supermarket entry.


bhousel:
looks good, thank you @bmillemathias 👍

PR #2547, opened 2019-04-05
Add wikidata/wikipedia information about Mobalpa

bmillemathias:


brands/shop/kitchen.json
```diff
@@ -21,6 +21,8 @@
     "count": 58,
     "tags": {
       "brand": "Mobalpa",
+      "brand:wikidata": "Q3317571",
```
bhousel: from the warnings on the `official_website` claim, it looks like there may be two entries for Mobalpa:
Q3117411 and Q3317571

Lets figure out which one is better and sort it out on Wikidata before we merge here..

bmillemathias: Q3117411 is the parent company that bought Mobalpa, it should not have `official_website` set to `http://mobalpa.com`

PR #2548, opened 2019-04-05
disambification of shop/clothes|Z with amenity/fuel|Z

bmillemathias:


brands/shop/clothes.json
```diff
@@ -2411,8 +2411,10 @@
   },
   "shop/clothes|Z": {
     "count": 57,
+    "nomatch": ["amenity/fuel|Z"],
     "tags": {
       "brand": "Z",
+      "clothes": "babies,children",
```
bhousel: In this OpenStreetMap tag, they prefer to use semicolon-delimited-lists.. Can you change it to `babies;children`  (no space)

PR #2549, opened 2019-04-05
Add info for hairdresser Franck Provost

bmillemathias:
Fixes issue #1492


bhousel:
Thanks @bmillemathias  -
If you are going to create a wikidata page for the brand this is great 👍  but make sure you also add statements for the `official_website` and statements for any social media accounts (Facebook, Twitter, Instagram, etc) that the brand might have.  We've had a few pages get deleted by Wikidata admins if they look too empty.

PR #2550, opened 2019-04-05
add Tanuki data, fixes osmlab#921

DmitriiKh:


matkoniecz:

Adamant36:
Probably should have included the Ukrainian name:whatever it is also. I think that's how we do it now if name:en=whatever is added. I guess it can be added later though.

matkoniecz:
> Ukrainian

Why? See http://overpass-turbo.eu/s/HKA

Though `name:ru` probably may be added (with the same value as `name` tag).

Adamant36:
name:ru then. Google translate said it was Ukrainian for some reason. So that's what I went with.

DmitriiKh:
Thanks!
Let me check if I get it right. 
name:ru Тануки
name:en Tanuki
name ? Should it be the same as ru or en? Do you have a preference for this?

matkoniecz:
> name:ru Тануки
> name:en Tanuki

Yes.

>  name ? Should it be the same as ru or en? Do you have a preference for this?

In this case Russian, `name` is for name actually used on the ground and this restaurant chain is in Russia. So I expect that people in Russia use "Тануки" rather than "Tanuki".

It is trickier with company operating under multiple names in different countries, we use multiple entries with country restrictions to cover that.

DmitriiKh:
I see. Thanks for the explanation :) Would you like me to make a change to add `name:ru Тануки` ?

matkoniecz:
Feel free to make a new PR!

> and maybe "countryCodes": ["ru"]

Some entries are appearing outside Russia. It may be worth checking is it actually supposed to be present there, maybe also some countries should be also in countryCodes list.

DmitriiKh:
Okay. There are also three restaurants in Ukraine. Therefore it should be `"countryCodes": ["ru", "ua"], "name": "Тануки", "name:ru": "Тануки", "name:ua": "Танукi", "name:en": "Tanuki"` I think.

Adamant36:
the country codes are correct. You only need either ru or ua for the language though. Usually whichever one the sign is in. We normally only use English and one other language, either the main language of the business or in rare cases dominate language of the area where its located.

DmitriiKh:
Thanks. I'm gonna keep `"name:ru": "Тануки"` as this is the main language of the business.

DmitriiKh:
I've checked tanuki.kz (Kazakhstan) and have found out that there is a one more restaurant of the same brand.
This restaurant is not on the map yet but someone will add it later I guess.
Therefore it could be either `"countryCodes": ["ru", "ua", "kz]` or no `countryCodes`.
What do you think would be better? 
I would go with no `countryCodes` because this brand can easily open new restaurants in other post soviet countries in the future.

Adamant36:
Using the three country codes is fine, they aren't supported in iD editor yet anyway. So having them there won't cause problems. Its always better in my opinion to have some information versus no information also.

DmitriiKh:
Okay. Then it is gonna be adding `"countryCodes": ["ru", "ua", "kz]` and `"name:ru": "Тануки"`

bhousel:
> Using the three country codes is fine, they aren't supported in iD editor yet anyway.

Good news, @quincylvania implemented this recently!  https://github.com/openstreetmap/iD/issues/6124   
The country scopes will work in the next version.

matkoniecz:
> they aren't supported in iD editor yet anyway.

countryCodes are supported in Vespucci

bhousel:
> the country codes are correct. You only need either ru or ua for the language though. Usually whichever one the sign is in. We normally only use English and one other language, either the main language of the business or in rare cases the dominate language of the area where its located.

Also about this - I think we should relax this rule a little bit for countries where they speak several languages.  It's ok to put a few languages that are spoken interchangeably there. 

For example, for the BY Post offices issue, I left both `:be`, `:ru`, and `:en` languages in there, because in that country they really do speak both, and the signs are both.  
https://github.com/osmlab/name-suggestion-index/pull/2482#issuecomment-479219884

Adamant36:
"Also about this - I think we should relax this rule a little bit for countries where they speak several languages."

I agree with that. I think there can be a few languages added in some cases were multiple languages are spoken in the place without things getting out of hand. Plus, it means a lot less explaining to do when it comes up. Which I'm all for.

bhousel:
> Plus, it means a lot less explaining to do when it comes up. Which I'm all for.

PR #2552, opened 2019-04-06
Research convenience store

StheffanyHadlich:
Made the research and the brand exists but does not have a wikipedia page. #1380

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -2505,6 +2505,7 @@
     }
   },
   "shop/convenience|Юлия": {
+    "countryCodes": ["bg"],
     "count": 68,
     "tags": {
       "brand": "Юлия",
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @StheffanyHadlich - we can leave #1380 open until it has a `brand:wikidata` tag.  If you did some research and know the brand's website, can you update that ticket with whatever you found?  Thanks!

StheffanyHadlich:
this brand does not seem to have an website, I found it based on Facebook check ins in this place, I got the adress and verify through google maps that it actually exists but there is no website, sorry.

tohaklim:
This is a personal name (Yulia/Julia/July lit.), not a brand as such. It's fairly popular to call one's shop after the proprietor, as is the case here. Recommend discarding

PR #2554, opened 2019-04-07
Use French wikipedia page for 8 à Huit convenience

bmillemathias:
As they are french based shops

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -47,7 +47,7 @@
     "tags": {
       "brand": "8 à Huit",
       "brand:wikidata": "Q2818601",
-      "brand:wikipedia": "en:8 à Huit",
+      "brand:wikipedia": "fr:8 à Huit",
       "name": "8 à Huit",
       "shop": "convenience"
     }


Adamant36:
I'm not sure if it matters. I've seem people do the exact opposite before and change the page reference to English. Its not like they aren't all linked to on Wikidata anyway.

matkoniecz:
If all are in France then linking to a local Wikipedia is preferable.

BTW, country restriction is missing - I can add it. But to confirm:

@bmillemathias Are you sure that all, including ones existing but not mapped `name=8 à Huit shop=convenience` are in the France?

Adamant36:
"If all are in France then linking to a local Wikipedia is preferable." 

Does it say that anywhere? I'm saying that its not, but I've seen it go both ways multiple times.

matkoniecz:
It is certainly consensus among mappers in Poland, up to universally supported bot edits enforcing it.

There is also https://wiki.openstreetmap.org/wiki/Key:wikipedia 

> link to the Wikipedia article in the appropriate language

> which should be to the article in the primary language for the subject

> it is preferable to link article in a local language. It is appropriate to provide links to articles in secondary languages where there is no existing article about given object in local language. For example for museums in France French Wikipedia is preferred. But linking to the English article about the particular museum in France is OK if there is no matching article in French Wikipedia.

(disclaimer: I may be author of some quoted parts - see article history, but it is not a fresh change and so far I noticed no complaints)

bhousel:
Thanks @bmillemathias 👍

bhousel:
> "If all are in France then linking to a local Wikipedia is preferable."
> 
> Does it say that anywhere? I'm saying that its not, but I've seen it go both ways multiple times.

Yes, it mentions this briefly in the Contributing guide that for a non-English brand we should try to link to their local Wikipedia page.  (You're also right that it doesn't really matter much).

Adamant36:
Ok cool. I must have missed it.

PR #2555, opened 2019-04-07
Add wikidata reference for Tchip hairdresser

bmillemathias:
Partially fill issue #1505

### DIFF
--- a/brands/shop/hairdresser.json
+++ b/brands/shop/hairdresser.json
@@ -190,6 +190,7 @@
     "count": 90,
     "tags": {
       "brand": "Tchip",
+      "brand:wikidata": "Q62871250",
       "name": "Tchip",
       "shop": "hairdresser"
     }


Adamant36:
What does "partially fills" mean? Seems like it closes it to me.

bmillemathias:
> What does "partially fills" mean? Seems like it closes it to me.

Because there is no wikipedia page. I tried to create one for another shop but some zealot deleted it instantly...

matkoniecz:
In case of no Wikipedia article it is perfectly fine to add just wikidata. Changing Wikipedia notability guidelines is not required to be done as part of fixing issues :)

Adamant36:
That sucks. Yeah, in general good luck doing much of anything on Wikipedia like that these days. Luckily Wikidata is a little loser on nobility, but there's still probably there to sometimes. Just having a Wikidata entry without a Wikipedia article attached to it is perfectly fine for our purposes though.

PR #2557, opened 2019-04-07
Wikidata for hairdresser Camille Albane

bmillemathias:
fill partially #1488.

### DIFF
--- a/brands/shop/hairdresser.json
+++ b/brands/shop/hairdresser.json
@@ -3,6 +3,7 @@
     "count": 57,
     "tags": {
       "brand": "Camille Albane",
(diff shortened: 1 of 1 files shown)


Adamant36:
Thanks. In the future for the Wikidata entry it would be good to add a tittle and a description.

bmillemathias:
> Thanks. In the future for the Wikidata entry it would be good to add a tittle and a description.

There is but in  French, do you mean an english one ?

Adamant36:
I mean on the Wikidata page. There's only one and its not really French or English. It seems like you updated it. Thanks.

PR #2560, opened 2019-04-07
Added Birds bakery

ldw606:
Added the Birds bakery chain.
The website and store signs say 'Birds Bakery' but existing tagging uses the name 'Birds' so I've used 'Birds Bakery' in the brand tag and 'Birds' in the name tag.


Adamant36:
Hi there. The name tag and brand tag should match. So in this case you can make them both "birds." Then use the official_name tag for "Birds Bakery." That's how we usually do it.

ldw606:
Ok thanks, I'll change that now.
Just to make sure: I should set official_name to "Birds Bakery" even though the company is registered as "Birds (Derby) Limited"?

PR #2562, opened 2019-04-09
add support for marking entries as gone

### LINKED ISSUE #2415: Include no-longer existing brand, especially ones popular in OSM
Obviously it should be distributed separately, not as source for suggestion but rather for validator complaints.

It would also help with maintaining index - there are some unclear case like

```
  "amenity/bank|Banco Popular": {
    "count": 704,
    "countryCodes": ["pr", "us", "vi"],
    "match": ["amenity/bank|Popular"],
    "tags": {
      "amenity": "bank",
      "brand": "Banco Popular",
      "brand:wikidata": "Q537259",
      "brand:wikipedia": "es:Banco Popular Español",
      "name": "Banco Popular",
      "name:en": "Popular Bank",
      "official_name": "Banco Popular Español",
      "official_name:en": "Spanish Popular Bank",
      "operator": "Banco Santander",
      "operator:wikidata": "Q806215",
      "operator:wikipedia": "en:Banco Santander"
    }
  },
```

that is mapped 400+ in Spain, 200+ in Dominicana, company is described as gone on Wikipedia. It would be nice to know whatever someone checked is this brand is still used in Spain or not.

Other examples: Banca Intesa - is this brand still in use or is OSM outdated?

```
{
  "tags": {
    "amenity": "bank",
    "brand": "Allied Bank",
    "brand:wikidata": "Q4732555",
    "brand:wikipedia": "en:Allied Banking Corporation",
    "name": "Allied Bank"
  },
  "countryCodes": [
    "pk",
    "ph"
  ]
}
```
is also described as defunct

matkoniecz:
Opened as a PR as it extends data structure and makes collecting no longer existing brands (that are still mapped in OSM) as an explicit goal.

This PR add support for marking brand as gone, including cases where brand is gone only in some countries. Such entries are dropped from `/dist/` files.

In future it may be possible to start distributing also list of ghost brands, still present in OSM but it is nod done for now.

fixes #2415 though some new issues needs to be created for likely no more existing brands mentioned there

------

@rjw62 I am pinging you as you maintain [entries and pretty graph](https://osm.mathmos.net/ghosts/) for UK about defunct brands. This PR add supports for marking brands as gone in this global index. Given that you have bigger experience with such data gathering I would welcome your feedback!

Thanks to looking at how your idex is generated I already added comment field alowing to add explanation what happened with a brand.

BTW, would it be OK to use your brand list and descriptions in this index? It would require releasing them under BSD 3-Clause license (see https://github.com/osmlab/name-suggestion-index/blob/master/LICENSE.md for details).

In addition to intention mentioned above this data could be used to make site similar to https://osm.mathmos.net/ghosts/ but with a global coverage.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -541,3 +541,65 @@ can show you where the Wikidata information is missing or incomplete.
 
 <img width="800px" alt="Adding information on Wikidata" src="https://raw.githubusercontent.com/osmlab/name-suggestion-index/master/docs/img/wikidata.gif"/>
 
+### Mark brands as no longer existing
+
+Brands may disappear over time. To aboid cluttering suggestion lists with defunct brands and to allow flagging likely outdated entries it is possible to mark brands as no longer existing. For example following entry has `"notExisting": true` and `notExistingDescription` fields.
+
+```js
+  "shop/toys|Toys R Us~(partial closure)": {
+    "countryCodes": ["au", "us"],
+    "match": ["shop/toys|Toys \"R\" Us"],
+    "nocount": true,
+    "notExisting": true,
+    "notExistingDescription": "There are no Toys R Us branded toy shops in USA and Australia - see https://www.cnn.com/2019/02/11/business/toys-r-us-return/index.html",
+    "tags": {
+      "brand": "Toys R Us",
+      "brand:wikidata": "Q696334",
+      "brand:wikipedia": "en:Toys \"R\" Us",
+      "name": "Toys R Us",
+      "shop": "toys"
+    }
+  },
+```
+
+Note that thanks to `countryCodes` it is possible to have also entry describing that `Toys R Us` continues to function in other countries:
+
+```js
+  "shop/toys|Toys R Us": {
+    "count": 240,
+    "countryCodes": [
+      "ae",
+      "at",
+      "bh",
+      "ca",
+      "ch",
+      "cn",
+      "de",
+      "es",
+      "fr",
+      "il",
+      "jp",
+      "kw",
+      "my",
+      "nl",
+      "no",
+      "ph",
+      "pl",
+      "pt",
+      "qa",
+      "sa",
+      "se",
+      "sg",
+      "th",
+      "za"
+    ],
+    "match": ["shop/toys|Toys \"R\" Us"],
+    "tags": {
+      "brand": "Toys R Us",
+      "brand:wikidata": "Q696334",
+      "brand:wikipedia": "en:Toys \"R\" Us",
+      "name": "Toys R Us",
+      "shop": "toys"
+    }
+  },
+```
--- a/brands/shop/supermarket.json
+++ b/brands/shop/supermarket.json
@@ -704,6 +704,7 @@
   },
   "shop/supermarket|Comercial Mexicana": {
     "nocount": true,
+    "notExisting": true,
     "tags": {
       "brand": "Comercial Mexicana",
       "brand:wikidata": "Q2985173",
--- a/brands/shop/toys.json
+++ b/brands/shop/toys.json
@@ -130,7 +130,47 @@
   },
   "shop/toys|Toys R Us": {
     "count": 240,
+    "countryCodes": [
+      "ae",
+      "at",
+      "bh",
+      "ca",
+      "ch",
+      "cn",
+      "de",
+      "es",
+      "fr",
+      "il",
+      "jp",
+      "kw",
+      "my",
+      "nl",
+      "no",
+      "ph",
+      "pl",
+      "pt",
+      "qa",
+      "sa",
+      "se",
+      "sg",
+      "th",
+      "za"
+    ],
     "match": ["shop/toys|Toys \"R\" Us"],
+    "tags": {
+      "brand": "Toys R Us",
+      "brand:wikidata": "Q696334",
+      "brand:wikipedia": "en:Toys \"R\" Us",
+      "name": "Toys R Us",
+      "shop": "toys"
+    }
+  },
+  "shop/toys|Toys R Us~(partial closure)": {
+    "countryCodes": ["au", "us"],
+    "match": ["shop/toys|Toys \"R\" Us"],
+    "nocount": true,
+    "notExisting": true,
+    "notExistingDescription": "There are no Toys R Us branded toy shops in USA and Australia - see https://www.cnn.com/2019/02/11/business/toys-r-us-return/index.html",
     "tags": {
       "brand": "Toys R Us",
       "brand:wikidata": "Q696334",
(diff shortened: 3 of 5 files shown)


matkoniecz:
I believe that it is not controversial, so I may merge it in the future by myself. But in case of self-merging I will wait at least one week.

rjw62:
I'm happy for you to use the list I've compiled at https://osm.mathmos.net/ghosts/ and the descriptions and queries for each under whatever licence you want. (The information in the descriptions has come from news reports and wikipedia, but I think the interpretation and precise words used have been mine in all cases.)

As for comments / feedback: if you might want to use the notExisting entries in the future to help people review and re-tag things, it could be useful to explicitly record whether the brand has shut down completely or just been renamed. In the latter case, having a pointer to the replacement brand (or brands) could also be useful. (It may be rare but sometimes a business may split into two or more brands. e.g. the UK bank "Lloyds TSB" split back into "Lloyds" and "TSB" a few years ago.)

Adamant36:
That makes me think, it would be cool if there was a tag for consolidated versus closed companies. Although I'm not sure what use case it would have, but I still like the idea. I guess the notExistingDescription kind of covers it though.

bhousel:
Couldn't we just fetch this from wikidata too?
in other words, mark it gone there instead.

matkoniecz:
problem 1 - different situations in different countries

What about cases like Toys R Us where brand is defunct in USA and Australia and present elsewhere? How it can be represented in Wikidata? Note that processing would require to parse this properties/qualifiers, split affected entry in parts and assign/modify `CountryCodes`.

----
problem 2 - brands and liked Wikidata entries are not always matching 1:1 

In many cases we are not linking entries that are actually about brand but about company, organization or something else. There are cases where brand is existing but its original company is defunct (or reverse).

----
problem 3 Wikidata data quality

In the past I tried running comparison of data in OSM and in Wikidata, including listing Wikidata entries with P576 property linked from OSM elements.

It turned to not be very useful as Wikidata quality in this area turned to be really low (in addition to copyright issues and less than exact matches between OSM and linked Wikidata entries).

Maybe it would be better for brands, it may be worth checking. But I expect that most of this data would be collected from scratch as part of name-suggestion-index.

The worst problem is that all updates from Wikidata of this data would need to be manually reviewed. Based on my experience with Wikidata in general and P576 specifically I would expect most changes to be either true from certain perspective but not matching our use or simply wrong.

I expect that this data is not present in Wikidata and would need to be collected. And due to potentially significant impact of mistakes any updates from Wikidata would need to be reviewed, removing benefits of not keeping it in the repository.

Adamant36:
> brands and liked Wikidata entries are not always matching 1:1

Maybe we could do more to help them match then. In most cases that I've seen they either match or when they don't its because the Wikidata entry doesn't have any information, not that they are outright wrong. Its not like we can't change the titles and descriptions where we need to though. I think we should anyway in cases where it improves things on Wikidata's side. The entries can ultimately be whatever we want them to be (within reason of course).

> problem 3 Wikidata data quality

The repeated complaints about the low quality of Wikidata/Wikipedia information by people connected to OSM is largely exaggerated in my opinion. Especially in regards to Wikidata. Since you can't take "poetic licence" with it to write things however you want them to sound like you can Wikipedia. It only needs to be accurate for a single thing in this anyway. Which from what I've seen so far it is. 


I don't think it  really matters though, because as I see it there could an npm run build check that looks for the "dissolved" tag on Wikidata and then gives a warning if its there, but the equivalent isn't present in the brands index entry or something like that. In cases like that people would still have to confirm through research what the facts are and do a PR for the change. Which would then still be double checked by whoever merges it. Which is no different then the process we already use for other things.

Doing it that way might have the advantage of only requiring an extra tag being added entry instead of it having to be deleted and moved to a whole separate file. There's some instances where a company is closed or dissolved on paper, but still in a transitional period or still taking up the building where it would be helpful to have the entry there like normal, but with a tag indicating that its closed or something. For instance there was a Sears a few towns away from where I live that closed like a year ago, but the building still said Sears on it and there was still Sears related stuff going on in it until a few months ago. The same with K-Mart when it closed a while back. It's something to think about.

matkoniecz:
> Maybe we could do more to help them match then. In most cases that I've seen they either match or when they don't its because the Wikidata entry doesn't have any information, not that they are outright wrong.

The typical problem is for `brand:wikidata` tags we link entries about companies, not bands. It typically works but there are cases where brand is existing but its original company is defunct (or reverse, company continues to exist but is now under a different brand).

bhousel:
@matkoniecz I've thought about it some more and I'd really prefer for us to pick this information from Wikidata somehow.  It would benefit everyone to have it recorded there, not just our own project.  

Also I don't really want to maintain a long list of `countryCodes` where a brand is winding down its operations.

But I'd be totally ok with adding properties like this in the `build_wikidata.js` and `dist/wikidata.json` and then showing users a warning about the brand in iD.

(Another alternative.. this is kind of stupid but it just occurred to me - we could add a tag like `opening_hours=24/7 off` or whatever the preferred way is of saying "permanently closed" until the business is bought or whatever)

codeinabox:
Has there been any more consideration of this? I've noticed that The Japanese Canteen has recently gone insolvent, so I am assuming all its shops are now permanently closed, and I am wondering how to treat this 🤔

Cj-Malone:
> problem 2 - brands and linked Wikidata entries are not always matching 1:1
> 
> In many cases we are not linking entries that are actually about brand but about company, organization or something else. There are cases where brand is existing but its original company is defunct (or reverse).

This is fairly common in GB at the moment, eg: [Topshop](https://www.wikidata.org/wiki/Q1893576) sold it's "brand"/"ip" to an online only retailer, but all the employees were fired and the shops shut.

The brand still exists, the shops don't, it's not obvious which the Wikidata entry represents.

bhousel:
Yeah, we've previously avoided doing special handling for closed brands for a few reasons:

1. It's often a complicated situation.  Sometimes they really are all closed and those stores can be deleted or marked as disused, sometimes they reopen with a new brand?  It depends.
2. If it _really_ is simple situation and all the brands can be closed - It's easier to just notify the community and perform an automated edit with Overpass and Level0 or JOSM to close them.  Like it's a lot easier to do that than to try to work through the NSI and the various validators that use it, or to make maproulette challenges or whatever.  Just close them in a single changeset 👍   You dont need NSI for this..

matkoniecz:
In many cases it is 100% known that shops belonging to a specific brand are all gone but mass edit is not viable.

For example:
- where some shops got closed and some renamed
- where shops get renamed into different names (maybe new owner has regular and premium brand for its POIs)
- where it is not clear what new owner is doing with POIs
- where new objects have all the same brand but different shop type, distinct from previous shop type
- brand is gone in some specific area (like specific country)
- and more

In such cases it would be nice to have ability to flag objects as worth resurveying.

bhousel:
We do distribute a file of the items in NSI that we suspect to be dissolved, based on the information in Wikidata.
https://github.com/osmlab/name-suggestion-index/blob/main/dist/dissolved.json

So other projects could use this however they want, either to raise an issue in the user interface, or encourage users to resurvey.

PR #2563, opened 2019-04-09
Add bestdrive car repair

bmillemathias:
https://overpass-turbo.eu/s/HPl

### DIFF
--- a/brands/shop/car_repair.json
+++ b/brands/shop/car_repair.json
@@ -25,6 +25,16 @@
       "shop": "car_repair"
     }
   },
+  "shop/car_repair|BestDrive": {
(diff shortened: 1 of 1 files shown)


brands/shop/car_repair.json
```diff
@@ -25,6 +25,16 @@
       "shop": "car_repair"
     }
   },
+  "shop/car_repair|BestDrive": {
+    "match": ["shop/car_repair|BestDrive"],
```
bhousel: Just a quick note - `match` is for less desirable taggings or spellings of `key/value|name`.  
You don't need to put the item's same key "shop/car_repair|BestDrive" into here, and this makes a warning.  I fixed the warning already, just letting you know for the future.

PR #2565, opened 2019-04-09
Add dedicated wikidata for Cache Cache

### LINKED ISSUE #1151: Research needed for: shop/clothes|Cache Cache
#### :thinking: We need help deciding what to do with "shop/clothes|Cache Cache"

### What's needed

First, do some research on `shop/clothes|Cache Cache` by searching Google and Wikipedia.

  * Find the entry for `shop/clothes|Cache Cache`
* `npm run build`

* Edit `config/canonical.json` to remove the entry for `shop/clothes|Cache Cache`
* `npm run build`

:sparkles: Thanks!

bmillemathias:
Fixes #1151

### DIFF
--- a/brands/shop/clothes.json
+++ b/brands/shop/clothes.json
@@ -293,8 +293,8 @@
     "count": 86,
     "tags": {
       "brand": "Cache Cache",
+      "brand:wikidata": "Q63064086",
       "name": "Cache Cache",
-      "operator:wikidata": "Q63064086",
       "shop": "clothes"
     }
   },
--- a/brands/shop/clothes.json
+++ b/brands/shop/clothes.json
@@ -294,9 +294,7 @@
     "tags": {
       "brand": "Cache Cache",
       "name": "Cache Cache",
-      "operator": "Groupe Beaumanoir",
-      "operator:wikidata": "Q3117333",
-      "operator:wikipedia": "fr:Groupe Beaumanoir",
+      "operator:wikidata": "Q63064086",
       "shop": "clothes"
     }
   },


brands/shop/clothes.json
```diff
@@ -294,9 +294,7 @@
     "tags": {
       "brand": "Cache Cache",
       "name": "Cache Cache",
-      "operator": "Groupe Beaumanoir",
-      "operator:wikidata": "Q3117333",
-      "operator:wikipedia": "fr:Groupe Beaumanoir",
+      "operator:wikidata": "Q63064086",
```
SilentSpike: Shouldn't this be `brand:wikidata`?

Why remove all the operator tagging?

Adamant36: Yeah, Q6306408 should be the brand:wikidata entry and Q3117333 should stay as the operator. It would also be good if you filled out the description and name for Q63064086. Otherwise, there's no way for anyone to find it. Plus, it helps in the brands list here if they have those details http://osmlab.github.io/name-suggestion-index/brands/index.html

bmillemathias: > Shouldn't this be `brand:wikidata`?
> 
Yes indeed, this is a typo.

> Why remove all the operator tagging?
Cache Cache can be franchise (https://www.toute-la-franchise.com/franchise-15291-cache-cache.html 🇫🇷) so I found ackward all future Cache Cache POI will be declared as operated by Groupe Beaumanoir

Adamant36:
Hhmmm, another instance where something like an owner or subsidary_of tag would be good. I guess it doesn't really matter that much if its there. Since the information is in the Wikidata entry. Its still good to have though.

SilentSpike:
In this case I'm personally indifferent regarding the removals.

I can see an argument for keeping the tagging since it will be applicable to most instances, but I also see a strong argument that we should only include tagging that applies to all instances.

bhousel:
> I can see an argument for keeping the tagging since it will be applicable to most instances, but I also see a strong argument that we should only include tagging that applies to all instances.

In case it wasn't clear - only include tags that apply to all instances.

Adamant36:
On my side, I said "id suggest" keeping it. As in I would, but didn't outright because I don't think we should knowingly or intentionally introduce bad data ever. Even if its only a tiny amount. There's no way to know if it is only a few instances either. Maybe if it was just like two that were outliers we could fix them manually so the rest can have the operator tags, but I don't really think that's a good idea if it was. To much of a slippery slope. Operator information isn't that important anyway in my opinion. I was originally more bullish on itt because I hadn't worked with Wikidata yet and didn't know it could be included in there. Now it seems fine just to have it on the Wikidata side in some cases like this one.

PR #2566, opened 2019-04-12
Correct information for Matsuya Foods in amenity/fast_food|松屋

ben-dyer:
According to https://en.wikipedia.org/wiki/Matsuya_Foods the number of stores is 1080 not 630 (as of 2018), so I have updated the count.

The company is Japanese, so I made the brand:wikipedia point to the Japanese wikipedia page rather than the Chinese page.

Also I have added the country code for Japan.

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -2787,7 +2787,7 @@
     }
   },
   "amenity/fast_food|松屋": {
-    "count": 630,
+    "count": 1080,
     "countryCodes": [
       "cn",
       "hk",
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -2787,7 +2787,7 @@
     }
   },
   "amenity/fast_food|松屋": {
-    "count": 1080,
+    "count": 630,
     "countryCodes": [
       "cn",
       "hk",
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -2791,6 +2791,7 @@
     "countryCodes": [
       "cn",
       "hk",
+      "jp",
       "mo",
       "sg",
       "tw"
@@ -2800,7 +2801,7 @@
       "brand": "松屋",
       "brand:en": "Matsuya Foods",
       "brand:wikidata": "Q848773",
-      "brand:wikipedia": "zh:松屋食品",
+      "brand:wikipedia": "ja:松屋フーズ",
       "cuisine": "japanese",
       "name": "松屋",
       "name:en": "Matsuya Foods",


ben-dyer:
Thank you and my apologies for the mistake - I restored the original value.

tas50:
I did this to a few when I started committing.

Adamant36:
I feel like If the name and brand is going to be in Chinese then the Wikipedia information should be also. Since people in China who are clicking on the store POI's won't be reading Japanese. 

So instead of changing the Wikipedia information, there should be a new entry for the store with the Japanese name (I assume there are some uses of it) and information with a Japan specific country code limit. We already do this in other instances. 

Also, if name brand and name both equal Chinese and have their English translations, there should also be name:zh and brand:zh to go with them.

ben-dyer:
@Adamant36 Are the name and brand specifically in Chinese? 松屋 is also Japanese so I don't understand how to distinguish this.

Adamant36:
According to Google translate the characters "松屋" are Chinese and they have 697 uses in tag info with the name tag. There's only 7 uses of "株式会社松屋フーズ" Plus, the Japanese letters or the longer version of the name with the extra Chinese characters aren't used on their signs. So its not  something you really need to distinguish in the first place and its perfectly fine to just go with the 2 Chinese characters. Generally, we go with signage and established tagging in these cases. If you still want to use the longer form of the name there's the official_name tag. We use it sometimes, but I don't think it's necessary here. 

download
MATSUYA_FOODS_in_Japan_101
松屋

ben-dyer:
I definitely think 松屋 is the more representative name, thanks for the information on conventions for this database. However, I don't think that 松屋 is Chinese. Google translate may not be able to distinguish, but kanji are also a part of the Japanese language separately from their Chinese origin and in this context that word is pronounced "matsuya" which is strictly the Japanese pronunciation of these characters.

Adamant36:
Your welcome. 

Google Translate can be wrong sometimes. Yandex says that  松屋 is Chinese for "pine house" which makes sense because the first letter looks like a pine tree and I assume the second one is a house. It translates into as 松の家. Which still uses the pine tree symbol but the house. So there is some overlap. But then if you put matsuya into Google Translate as Japanese to Chinese it translate to Chinese as 松屋. So...My guess is that Matsuya is just an English translation of a word, but it could sound different or whatever in Chinese/Japanese. So you can't say "Matsuya" is Japanese and Chinese. Its ultimately English you have to just look at the Chinese characters and disregard the English usage. At least that's how I look at it. 

Some of the restaurant seem to have the name in English and Chinese, but none seem to have it in Japanese. Which is weird since its a restaurant in Japan, selling Japanese noodles 🤷‍♂️ 

matsuya

johachi:
Hi, I just want to weigh into this discussion. I speak Japanese, and am living in Japan since several years.

Japanese uses kanji, which are Chinese characters. (Japanese also uses Hiragana and Katakana.)

The full name of "松屋" is "株式会社松屋フーズ" . 株式会社松屋フーズ uses the shortened name 松屋 for their stores in Japanese (just like Hennes & Mauritz AB uses H&M) . In other words, 松屋 is Japanese. Just check this Japanese dictionary: [Jisho.org](https://jisho.org/search/matsuya).

Lastly, "Matsuya" is NOT a translation of 松屋, it is the pronunciation of "松屋" as heard by English speaker when they hear a Japanese person read the word. A Chinese person would not read "松屋" as "Matsuya" since they have a different pronunciation.

bhousel:
Thanks @ben-dyer !   This looks fine 👍

PR #2567, opened 2019-04-12
Add wikidata for optician Lissac

bmillemathias:


Adamant36:
Thanks., looks good. btw, you might want to run `npm run build`  in the terminal before you commit the files so it gets rid of the trailing space at the end. Its not that big of a deal, but I thought id mention it.

bmillemathias:
> 
> 
> you might want to run `npm run build` in the terminal before you commit the files so it gets rid of the trailing space at the end. Its not that big of a deal, but I thought id mention it.

Oups, I normally do ...
Thanks

PR #2568, opened 2019-04-12
Add wikidata for Games Workshop

bmillemathias:


matkoniecz:
`brand:wikipedia` also may be added

bmillemathias:
> `brand:wikipedia` also may be added - https://en.wikipedia.org/wiki/Games_Workshop exists

indeed, done.

PR #2570, opened 2019-04-12
Added wikidata and country code for Marko Aldany

PrimeN:
This is to close the [Issue #1503](https://github.com/osmlab/name-suggestion-index/issues/1503)

### DIFF
--- a/brands/shop/hairdresser.json
+++ b/brands/shop/hairdresser.json
@@ -145,8 +145,10 @@
   },
   "shop/hairdresser|Marco Aldany": {
     "count": 67,
+    "countryCodes": ["es"],
     "tags": {
       "brand": "Marco Aldany",
+      "brand:wikidata": "Q63109159",
       "name": "Marco Aldany",
       "shop": "hairdresser"
     }


Adamant36:
Hi there. I did a search for the Wikidata code you provided and it doesn't seem to exist. Maybe you typed it in wrong or something?

PrimeN:
It was missing a wikidata page so I went ahead and created one for it but it looks like the wikidata page is deleted. I'm not sure if I should re-create the page and update the tag again or remove the wikidata tag entirely. Please advise.

matkoniecz:
Lets wait for reply in https://www.wikidata.org/wiki/Wikidata:Administrators%27_noticeboard#Explanation_request_for_Q57643977

matkoniecz:
Undeletion requested at https://www.wikidata.org/w/index.php?title=Wikidata:Administrators%27_noticeboard&diff=prev&oldid=931120571

matkoniecz:
After discussion with Wikidata admins I created explanation what Wikidata maintainers consider as a minimal requirements - see https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#adding-references-to-wikidata

Are you maybe interested in trying to create a new Wikidata entry, this time with references?

Adamant36:
You don't think there would be an easier qualifying references than "has part"? That seems a little unnecessarily convoluted to me. I have yet to see it on any wikidata entries without Wikipedia articles attached to them that haven't been deleted either and I've seen a lot. So there has to be an easier reference.

matkoniecz:
> You don't think there would be an easier qualifying references than "has part"?

Can you propose something better? I would happily merge PR that would add other example or replace it with something simpler and also relatively easy to find for a typical brand entry.

Adamant36:
> Can you propose something better? I would happily merge PR that would add other example or replace it with something simpler and also relatively easy to find for a typical brand entry.

No. Not currently. Like I said on the other PR, it would be good to keep an eye out for something better though. I think ultimately it should be a general list of things that might help Wikidata entries from getting deleted. Instead of a single example. As it seems to be up to the whims of particular admins anyway. So only adding one thing might only satisfy one of them. Whereas, adding as many entries as we can seems like a much more reliable tactic. 10 bucks says another admin comes along at some point and thinks that's not an adequate reference. We should cover all our bases just in case and not rely to much on one reference. I'm pretty sure single references can be deprecated/sunseted also. I've seen it happen a few times already.

Adamant36:
@matkoniecz any updates on this? It seems like it still doesn't have a Wikidata entry. If its not going to be restored maybe we should just create a new one.

matkoniecz:
> we should just create a new one.

to quote myself:

> Are you maybe interested in trying to create a new Wikidata entry, this time with references?

Adamant36:
> to quote myself:

Just wanted to double check. I was under the impression that they would restore them if we added the proper sources. If that's not the case, I guess we can create new entries. I worry that would be similar to edit warring or something though. Since we would be going against the wishes of the admins that there weren't Wikidata entries for these companies.

matkoniecz:
> I was under the impression that they would restore them if we added the proper sources.

Yes, they restored bunch of them. As I explained in https://github.com/osmlab/name-suggestion-index/pull/2572#issuecomment-500703638 

> For now I am waiting for someone to fix #2608 before requesting undeletion of further items. 

Though note that anyone may go to Wikidata admins and ask for undeleting some items.

matkoniecz:
Closing as it requires recreation of Wikidata entry (with references to match Wikidata requirements).

PR #2572, opened 2019-04-12
Update info for Якитория

PrimeN:
якитория/Yakitoria is a japanese restaurant in moscow.
should be sufficient to close [issue #922](https://github.com/osmlab/name-suggestion-index/issues/922)

### DIFF
--- a/brands/amenity/restaurant.json
+++ b/brands/amenity/restaurant.json
@@ -1566,10 +1566,14 @@
     }
   },
   "amenity/restaurant|Якитория": {
+    "countryCodes": ["ru"],
     "count": 86,
     "tags": {
       "amenity": "restaurant",
       "brand": "Якитория",
+      "brand:en": "Yakitoria",
+      "brand:wikidata": "Q63109465",
+      "cusine": "Japanese",
       "name": "Якитория"
     }
   },
--- a/brands/amenity/restaurant.json
+++ b/brands/amenity/restaurant.json
@@ -1570,8 +1570,9 @@
     "count": 86,
     "tags": {
       "amenity": "restaurant",
+      "name:en": "Yakitoriya",
       "brand": "Якитория",
-      "brand:en": "Yakitoria",
+      "brand:en": "Yakitoriya",
       "brand:wikidata": "Q63109465",
       "cusine": "Japanese",
       "name": "Якитория"


Adamant36:
Hi there. It looks like it needs name:en also.

Adamant36:
Also, this doesn't seem to be a valid Wikidata code.

PrimeN:
I went ahead and added the name:en. However it looks like the wikidata page is deleted. I'm not sure if I should re-create the page and update the tag again or remove the wikidata tag entirely. Please advise.

Adamant36:
Ok. They have been doing that if you don't put enough information when you create the entry because then its not notable enough. I don't think just creating another entry is a good way to go. As it could get you banned. In cases like this the administrator that deleted it has to be contacted and then if that doesn't work it should be escalated to the admin message board. That's really my area of expertise though. 

@matkoniecz would you mind looking into this and #2570? I assume we shouldn't just create new entries for things if they are deleted right?

Adamant36:
As a side to this, since this has happened several times and its only been due to new people not knowing to put in a lot of information, I suggest we discourage new users from creating wikidata entries items. Since once it gets deleted we are sol and its not worth the risk. There was also a few times I asked people to add more information to the entries they created and they ignored me for whatever. So I don't think simply telling them to do it or having a guide they might or might not follow is a good option.

PrimeN:
Yea I wasn't sure how much information it needed and judging from some of the smaller pages I went with name, description, country and website. We can perhaps add a guide with an example for new people to follow.

P.S. I don't mind creating a more comprehensive page if you need me to (something similar to the one for burger king and such). Let me know.

Adamant36:
> I don't mind creating a more comprehensive page if you need me to

I appreciate that. It's still unclear what the threshold for notability is unfortunately. Links to social media accounts if they have any probably helps, but I think one with social media links even got deleted. So who knows. It could just be a few overly cautious admins to, who are deleting them because they don't have Wikipedia pages or something. That happens sometimes.

Adamant36:
Whatever the case is on what we do or don't add the entries though, I don't think they be deleting them immediately after you created them. That seems odd. Especially since it was two entries that got deleted so quickly at almost the same time. Even if you going to add more details to the entries they really didn't give you much of a chance to.

matkoniecz:
Lets wait for reply in wikidata.org/wiki/Wikidata:Administrators%27_noticeboard#Explanation_request_for_Q57643977

matkoniecz:
Undeletion requested at https://www.wikidata.org/w/index.php?title=Wikidata:Administrators%27_noticeboard&diff=prev&oldid=931120571

Adamant36:
@matkoniecz, if push comes to shove maybe look for the brands on the websites mentioned in the Identifiers section of a lot of articles, like the library of congress or a similar third party database. There's a lot of them out there that Wikidata supports and I'm pretty sure they would qualify as a third-party source. Or maybe even a stock exchange symbol. It doesn't have to be anything to complicated, just as long as it's not the brands website. Like in cases where I can I'll add the brands iTunes developer id #. Its both a subject-independent source and not user-generated content. So Id think it would count.

matkoniecz:
> like the library of congress or a similar third party database

I may try, but are you sure that library of congress assigns identifiers to random brands across the world?

matkoniecz:
After discussion with Wikidata admins I created explanation what Wikidata maintainers consider as a minimal requirements - see https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#adding-references-to-wikidata

Are you maybe interested in trying to create a new Wikidata entry, this time with references?

Adamant36:
@matkoniecz, any update on this? If its not going to be restored we should just create a new Wikidata entry for it or something.

matkoniecz:
> @matkoniecz, any update on this?

It turned out that 250+ of our entries are at risk of deletion according to intepretation of Wikidata rules. For now I am waiting for someone to fix https://github.com/osmlab/name-suggestion-index/issues/2608 before requesting undeletion of further items. Though anyone may go to Wikidata admins and ask for undeleting some items.

matkoniecz:
> we should just create a new Wikidata entry for it or something.

Note

> Are you maybe interested in trying to create a new Wikidata entry, this time with references?

in https://github.com/osmlab/name-suggestion-index/pull/2572#issuecomment-497978657

matkoniecz:
Closing as it requires recreation of Wikidata entry.

PR #2574, opened 2019-04-14
Add 5àsec dry_cleaning shops

bmillemathias:
Overpass: https://overpass-turbo.eu/s/HYh
name need to be normalized in OSM.

### DIFF
--- a/brands/shop/dry_cleaning.json
+++ b/brands/shop/dry_cleaning.json
@@ -1,4 +1,20 @@
 {
+  "shop/dry_cleaning|5àsec": {
+    "match": [
+      "shop/dry_cleaning|5 a sec",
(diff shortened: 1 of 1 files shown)


matkoniecz:
>  name need to be normalized in OSM.

In that case adding `match` parameters with popular incorrect name variants would help.

bmillemathias:
> 
> 
> > name need to be normalized in OSM.
> 
> In that case adding `match` parameters with popular incorrect name variants would help to do that.

is "match" case-sensitive ? it is possible to use a regex or only globbing ?

matkoniecz:
> is "match" case-sensitive ?

Currently yes, though it was proposed to change this

> it is possible to use a regex or only globbing ?

Neither is available.

PR #2576, opened 2019-04-17
Keep shop/convenience for "8 à Huit"

bmillemathias:
remove duplicates entries

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -45,6 +45,7 @@
   "shop/convenience|8 à Huit": {
     "count": 133,
     "countryCodes": ["fr"],
+    "match": ["shop/supermarket|8 à Huit"],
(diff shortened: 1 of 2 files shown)


matkoniecz:
and there is not even single 8 à Huit that should be tagged as supermarket, right?

bmillemathias:
Normally 8 à Huit target small surfarces, that's why I put convenience, I based my judgement on this document http://www.franchise-proximite.carrefour.com/nos-enseignes/franchise-8-a-huit/chiffres-cles-8-a-huit/ (🇫🇷) issued by the mother company.
But you're right some of them could potentially be tagged as supermarket.

matkoniecz:
In that case we should probably keep both (and add `nomatch` to both to stop complaints by internal validator on `npm run build`).

PR #2578, opened 2019-04-19
Added cuisine for several restaurants in Japan

johachi:
Unsure how to handle the case when a restaurant have two profiles. I separated them with `;`, like ` "cuisine":"western;japanese"`. Please confirm if this is correct or not.

This kind of things are very common for "Family Restaurants" in Japan. It is basically restaurants catering to families serving both Western and Japanese food.

### DIFF
--- a/brands/amenity/restaurant.json
+++ b/brands/amenity/restaurant.json
@@ -1687,7 +1687,7 @@
       "brand:ja": "はなまるうどん",
       "brand:wikidata": "Q11275674",
       "brand:wikipedia": "ja:はなまるうどん",
-      "cuisine":"japanese;udon",
+      "cuisine": "japanese;udon",
       "name": "はなまるうどん",
       "name:en": "Hanamarūdon",
       "name:ja": "はなまるうどん"
@@ -1703,6 +1703,7 @@
       "brand:ja": "びっくりドンキー",
       "brand:wikidata": "Q11276815",
       "brand:wikipedia": "ja:びっくりドンキー",
+      "cuisine":"burger",
       "name": "びっくりドンキー",
       "name:en": "Bikkuri Donkey",
       "name:ja": "びっくりドンキー"
@@ -1718,6 +1719,7 @@
       "brand:ja": "やよい軒",
       "brand:wikidata": "Q11280577",
       "brand:wikipedia": "ja:やよい軒",
+      "cuisine":"japanese",
       "name": "やよい軒",
       "name:en": "Yayoiken",
       "name:ja": "やよい軒"
@@ -1745,8 +1747,9 @@
     "tags": {
       "amenity": "restaurant",
       "brand": "ガスト",
-      "brand:en": "Gusto",
+      "brand:en": "びっくりドンキー",
       "brand:ja": "ガスト",
+      "cuisine":"western;japanese",
       "name": "ガスト",
       "name:en": "Gusto",
       "name:ja": "ガスト"
@@ -1762,6 +1765,7 @@
       "brand:ja": "ココス",
       "brand:wikidata": "Q11301951",
       "brand:wikipedia": "ja:ココスジャパン",
+      "cuisine":"western;japanese",
       "name": "ココス",
       "name:en": "Coco's",
       "name:ja": "ココス"
@@ -1777,7 +1781,7 @@
       "brand:ja": "サイゼリヤ",
       "brand:wikidata": "Q886564",
       "brand:wikipedia": "en:Saizeriya",
-      "cuisine":"italian",
+      "cuisine": "italian",
       "name": "サイゼリヤ",
       "name:en": "Saizeriya",
       "name:ja": "サイゼリヤ"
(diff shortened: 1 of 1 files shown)


brands/amenity/restaurant.json
```diff
@@ -1744,8 +1747,9 @@
     "tags": {
       "amenity": "restaurant",
       "brand": "ガスト",
-      "brand:en": "Gusto",
+      "brand:en": "びっくりドンキー",
```
tas50: This tag is brand:LANG. This was brand:en before since it was in English. I think you want to update the lang for this tag value now.

johachi: Thanks for catching that. This was an unintentional edit. It should be `"brand:en": "Gusto",`. I have now pushed a the fix.

matkoniecz:
Looks OK to me, though I am unable to judge correctness of added tags.

`"cuisine":"western;japanese"` seems to be correct to me.

Adamant36:
Western isn't a cuisine tag. Which it shouldn't be in my opinion because its to vague. If its American cuisine, just use American. Otherwise, use whatever specific region it is. Also, adding Japanese to the udon entries seems redundant because udon is mainly (or only) a Japanese dish.

johachi:
Adamant36, why is "western" not a cuisine tag? Should I use "european" instead? I could alternatively add several European cuisine tags. But that seems unnecessary since both "[european cusine](https://en.wikipedia.org/wiki/European_cuisine)" and "[western cuisine](https://en.wikipedia.org/wiki/Western_cuisine)" is widely used, especially here in Asia.

As said, I am happy to change it to more specific terms, but it would then be a list with about 5-6 different cuisines. I do not think that is better.

I will happily remove "japanese" from the udon enter if it is deemed unnecessary.

Adamant36:
Its not used in OSM. It doesn't really matter what Wikipedia says. As we don't really go by that, but what's established. Same goes for European. Its to vague and there isn't really a "European" cuisine in a way that would help someone know what they are going to be eating when they visit the restuarant. Cuisine in Europe can very widely by region. If you want to see what is used to find out if something matches better there's here https://wiki.openstreetmap.org/wiki/Key:cuisine

and also https://taginfo.openstreetmap.org/keys/?key=cuisine#values

Adamant36:
I don't know about the other ones, but going by the menu for ココスジャパン they just serve American food. So that would work as the cuisine and there's no reason to have it tagged as Japanese also. You might want to check the other ones. They shouldn't have the Japanese cuisine tag if they don't actually serve Japanese food. Even if they are a Japanese company located in Japan. Not that I'm saying that's why you added it.

matkoniecz:
> Its not used in OSM.

It is perfectly fine to introduce new tags if there are no existing ones that work well.

And at least for me meaning seems clear.

Though given that for some meaning of this tag may seem unclear, I will document it on the OSM Wiki once I merge this PR.

johachi:
> (...) but going by the menu for ココスジャパン they just serve American food. So that would work as the cuisine and there's no reason to have it tagged as Japanese also. (...) They shouldn't have the Japanese cuisine tag if they don't actually serve Japanese food. (...)

They serve "American food", but they do also serve food from other countries, as well as Japanese food. Just look at [this print screen](https://screenshots.firefox.com/JesTx2uBSR9gbjyI/www.cocos-jpn.co.jp) (first column, third row), or check out their [official site](https://www.cocos-jpn.co.jp/menu/index.html).

Adamant36:
> It is perfectly fine to introduce new tags if there are no existing ones that work well.

It's not clear there aren't exiting ones that would work better though. For me in the menu's it just looks like American and Japanese food. I don't see anything "western" or "European" about it. Otherwise, what specific is European on their menu? Btw, traditional American food is technically European originally. So whatever generic dish might be European is probably also American. Unless it's specific to a region. Then it could be tagged as that. There's also already the generic "regional" catch all. Personally, I prefer more specific tags are used when we can and their Wikipedia page specifically says they are an American restaurant. It doesn't mention European though. At least not that I saw.

matkoniecz:
> It's not clear there aren't exiting ones that would work better though. 

Which one you would propose to use?

johachi:
@Adamant36 if you dislike the `western` tag so much, do you dislike  `asian` just as much? For me, it is very strange if `asian` can be accepted and not `western`.

matkoniecz:
I tracked down place to edit - for `western` it would be https://wiki.openstreetmap.org/w/index.php?title=Template:Map_Features:Cuisine_(ethnicity)&action=edit

bhousel:
This is turning into an unhelpful tagging discussion so I'm just going to merge it. 

Japanese users can use whatever `cuisine` tags they want for their restaurants.

If you want to pick "common" tag values, you can find them here:
https://taginfo.openstreetmap.org/keys/cuisine#values

If none of those work, make something up.

The worst consequence of making something up, or picking something unusual, is that software might not recognize the value.  So you might see a fork and knife icon on the pin instead of a chopsticks icon.  It's really no big deal.

If we want to change the cuisine tags later, someone can just change them, and because these brand nodes have a `brand:wikidata` tag , it is no effort at all to find and update them in OSM.  

Even if we somehow accidentally merged a change that said "Kentucky Fried Chicken is a yogurt shop", it's a simple overpass query to find and fix all the mistagged KFCs in the world.

We really don't have to get hung up on details like what `cuisine` tag to use.

PR #2582, opened 2019-04-21
Add Roblot funeral_directors

bmillemathias:
https://overpass-turbo.eu/s/I8I


bhousel:
This seems ok, though you will definitely want to add some more detail to the Wikidata page so the admins don't delete it.

bhousel:
oh whoops, it's missing a `name` tag

PR #2583, opened 2019-04-23
added wiki tags to social-facility

AjeaSmith:
I added wiki-data and Wikipedia tags in the social facility.json file. I followed the required guidelines I hope I did this correctly, this is my first time contributing. :)

### DIFF
--- a/brands/amenity/social_facility.json
+++ b/brands/amenity/social_facility.json
@@ -4,6 +4,8 @@
     "tags": {
       "amenity": "social_facility",
       "brand": "Maison de retraite",
+      "brand:wikidata": "Q22908",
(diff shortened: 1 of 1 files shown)


Adamant36:
From what I know in America safe heavens are just a general thing where abused people can go to get away from their boyfriends, husbands, whatever, more safely then somewhere like a homeless shelter. Which can actually put people at risk sometimes). So I'm pretty its generic. There might be a few places out there called "safe heaven" but its highly doubtful it's a brand. Same goes for the other one in my opinion.

AjeaSmith:
Ok I see, thanks for the reply! I'll watch out for that.

PR #2597, opened 2019-05-04
Add Australian pharmacy/chemist retail chains

davidhicks:
Note that some Australian pharmacy/chemist retail names have identical names to different chains located in other countries. Some "npm run build" warnings about possible duplicates are therefore seen.

### DIFF
--- a/brands/amenity/pharmacy.json
+++ b/brands/amenity/pharmacy.json
@@ -24,6 +24,22 @@
       "name": "Amavita"
     }
   },
+  "amenity/pharmacy|Amcal": {
+    "countryCodes": ["au"],
+    "match": [
+      "amenity/pharmacy|Amcal Pharmacy",
+      "shop/chemist|Amcal",
+      "shop/chemist|Amcal Pharmacy"
+    ],
+    "nocount": true,
+    "tags": {
+      "amenity": "pharmacy",
+      "brand": "Amcal",
+      "brand:wikidata": "Q63367373",
+      "healthcare": "pharmacy",
+      "name": "Amcal"
+    }
+  },
   "amenity/pharmacy|Apollo Pharmacy": {
     "count": 128,
     "countryCodes": ["in"],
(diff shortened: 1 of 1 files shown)


matkoniecz:
Personally I would prefer adding `amenity=pharmacy` (with `healthcare=pharmacy` as a possible addition).

Note than for example in iD "pharmacy" adds both.

davidhicks:
I see that I added one shop=chemist by mistake (but it also adds amenity=pharmacy and healthcare=pharmacy at the same time). Other than that, chemist is only mentioned in matches as there are numerous shop=chemist tagged items in Australia at the moment. OSM Wiki is not clear on a global standard to use of either and/or shop=chemist and amenity=pharmacy, hence the reason for confusion in Australia where both tagging standards have been used.

From this pull request, should I remove just the accidentally tagged shop=chemist from one brand, or should I also remove the matches that use shop=chemist as well?

matkoniecz:
> OSM Wiki is not clear on a global standard to use of either and/or shop=chemist and amenity=pharmacy

That is partially because it is a really complicated situation. See for example thread about USA pharmacies https://osmus.slack.com/archives/C029HV951/p1555963219128900 

> From this pull request, should I remove just the accidentally tagged shop=chemist from one brand, or should I also remove the matches that use shop=chemist as well?

Not sure - I have no idea at all about local situation in Australia. I would advice to consult wider local community (maybe on `talk-au` mailing list?) if you are not sure. If you prefer to not do it I may post mail on it.

Reading https://wiki.openstreetmap.org/wiki/Tag:shop%3Dchemist it seems that `shop=chemist` tag is covering shop with

> over-the-counter drugs, personal hygiene (toothpastes, shampoos, soaps etc.), cosmetics (usually cheaper ones), and household cleaning products

but it is usually not added to `amenity=pharmacy` objects

> amenity=pharmacy are likely to also be shop=chemist even if this is rarely tagged explicitly)

but as mentioned pharmacy situation is quite complicated and wiki is not some kind of oracle.

bhousel:
Thanks @davidhicks!  - Currently we are purposefully not matching `amenity/pharmacy` (the pharmacy part of the store) to `shop/chemist` (the drugstore part of the store) because they often have separate opening hours, contact info, and sometimes the pharmacy part can exist as its own section in a different store of clinic or urgentcare or whatever.  (Basically there are just a _ton_ of exceptions so we leave them as 2 things).

I'll accept the PR, but will split some of them up in a later commit.

davidhicks:
For Australia at least, it looks like amenity/pharmacy and shop/chemist should almost always be present on the same nodes/ways. If the shop is open and selling sunglasses, vitamins and shampoo then the pharmacy counter at the back of the shop will also be open and filling prescriptions.

An exception that comes to mind is a pharmacy located in a hospital which doesn't sell personal hygiene products, over-the-counter products, etc--they just provide prescription medications. In every other case I can think of, a shop/chemist and amenity/pharmacy in Australia are always together.

Issue #2604, opened 2019-05-08
detect links to Wikidata items at risk of being deleted due to missing external sources


matkoniecz:
Similar to  #2603 but covers separate deletion risk, based on my undeletion discussion at Wikidata:Administrators' noticeboard

Items on Wikidata are deletion proof if there is a linked Wikipedia article.

Items without a linked Wikipedia article may be deleted in some situations, according to https://www.wikidata.org/w/index.php?title=Wikidata:Administrators%27_noticeboard&diff=936485176&oldid=936327901 items not linked to Wikipedia articles should have either

- third party identifier

or

- independent source confirming some major claim

For example https://www.wikidata.org/wiki/Q60097339 is deletionproof thanks to https://www.wikidata.org/w/index.php?title=Q60097339&type=revision&diff=888204254&oldid=888202120 edit that added external source for one of major information (that it had 158 restaurants at some specific date)

It would be useful to list in wikicheck linked Wikidata entries that have no associate Wikipedia article and all claims are without external references - such entries may become deleted.

Issue #2603, opened 2019-05-07
detect links to Wikidata items at risk of being deleted due to missing P31 property and not linked to any Wikipedia article


matkoniecz:
Potential continuation of https://github.com/osmlab/name-suggestion-index/pull/2538 by @ENT8R 

It would be useful to add check listing Wikidata entries used in the index that are at risk of being deleted.

Items without link to Wikipedia article and without [P31 property](https://www.wikidata.org/wiki/Property:P31) may be deleted, see https://www.wikidata.org/w/index.php?title=Wikidata:Administrators%27_noticeboard&diff=prev&oldid=935716550 (Identification part)

There are also some minimal requirements for third-party sources, especially for companies and I trying to establish what is the minimum necessary to deletproof entries. Once I will get that info I will make a new issue to cover that.

Adamant36:
@matkoniecz, here's an example of a foreign company with a Library of Congress authority ID. It also has a bunch of other identifiers besides that. It might be worth keeping a running list. As the more the better in my opinion.

SilentSpike:
In the UK there is also companies house https://beta.companieshouse.gov.uk

matkoniecz:
> @matkoniecz, here's an example of a foreign company with a Library of Congress authority ID. It also has a bunch of other identifiers besides that. It might be worth keeping a running list. As the more the better in my opinion.

? It seems that link is missing, it also seems to be posted in a wrong issue

ENT8R:
> Items without link to Wikipedia article

That would be quite a lot of Wikidata entries which don't fit this requirement. According to my own experience, I would say that the `P31` claim is probably the most important claim and if an entry has this claim, the risk of it being deleted is very low.

> without P31 property

That's a very good way to find potential deletion candidates. I'm currently working on improving the wikicheck script to add another check for missing claims like P31

matkoniecz:
@ENT8R I though about listing entries without P31 and without Wikipedia link as "linked to Wikipedia article" makes it ineligible for a deletion, even if items is completely empty.

ENT8R:
Ah okay, so both conditions should be true...

matkoniecz:
> I'm currently working on improving the wikicheck script to add another check for missing claims like P31

Great! Are you potentially interested in adding also #2604 check?

ENT8R:
Yes, I will do both!

Adamant36:
> ? It seems that link is missing, it also seems to be posted in a wrong issue

[...]

PR #2605, opened 2019-05-08
Add two more checks to the wikicheck script

ENT8R:
Implements additional checks of #2603, #2604:
- check for missing sitelinks and missing instance claim:

- check for missing sitelinks and missing references to external sources (currently about 250(**!**) entries):


matkoniecz:
Thanks, looks great! Only third-part confirmation makes entry deletionproof (referencing company own website is not helping), but I see no simple way to check for that.

I created #2608 for first group.

PR #2610, opened 2019-05-11
Remove generic VR-Bank and Volksbank Raiffeisenbank

dettbarn:
Volksbank Raiffeisenbank (abbreviated VR-Bank) is technically not a unified brand of German banks. So I removed all common variants of the generic form, see code. This closes issues #322 and #323.

As has been done for the Sparkassen in Pull Request #2475, one would have to include all several hundred regional VR-Banks individually (which are, in fact, brands). But this will get complicated due to the varying spellings, since one has to allow for all variants (plus single "Volksbank" and "Raiffeisenbank", maybe) again.


matkoniecz:
> technically not a unified brand of German banks

Is in practice an unified brand of German banks?

dettbarn:
> > technically not a unified brand of German banks
> 
> Is in practice an unified brand of German banks?

No, not really. The individual VR-Banken all organize themselves separately, similar to the Sparkassen. (It's just that your bank account is available throughout all VR-Bank locations.) If you look into wikidata, you indeed will not find a unified VR-Bank brand, just the individual ones, which even sometimes are named quite nontrivially. (I've seen "VR meine Bank" and other quirks...)

Though, I think it would make sense to maintain the informal terms "VR-Bank" and "Volksbank Raiffeisenbank" (and sometimes "Volksbank" or "Raiffeisenbank") as short_name alternatives for each entry. (Or as name, and keep the full name as official_name.)

ENT8R:
> > technically not a unified brand of German banks
> 
> Is in practice an unified brand of German banks?

No, it is like @dettbarn said: similar to how the "Sparkassen" are organized. There is also a list of all "Volks- und Raiffeisenbanken": https://de.wikipedia.org/wiki/Liste_der_Genossenschaftsbanken_in_Deutschland but I agree that it doesn't make much sense to add every single entry of them...

bhousel:
Sounds good to me, thanks @dettbarn 👍 
I'm going to remove the Sparkassen too..

PR #2613, opened 2019-05-14
Add secondary cuisines to some fast food locations

tas50:
I added secondary cuisines for brands where they are distinctly known for two things and people would go there for either or both of these things. Dairy Queen for instance is a classic american burger joint, but also an ice cream place. It's a flip of the coin if a person in line is going to get a full meal or just some ice cream.


bhousel:
Thanks @tas50 👍

PR #2616, opened 2019-05-15
Add regional American chains, mostly Ohio-based

1ec5:
Added a number of regional American chains, most of them based in Ohio. Also fixed a typo in a bank name, refined cuisine tags, and added takeout information.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -2922,7 +2922,8 @@
       "brand": "Fifth Third Bank",
       "brand:wikidata": "Q1411810",
       "brand:wikipedia": "en:Fifth Third Bank",
-      "name": "Fifth Third Bank"
+      "name": "Fifth Third Bank",
+      "short_name": "5/3 Bank"
     }
   },
   "amenity/bank|Finansbank": {
@@ -2953,6 +2954,17 @@
       "brand:wikidata": "Q5452734",
       "brand:wikipedia": "en:First Citizens Bank (Trinidad and Tobago)",
       "name": "First Citizens Bank"
+    }
+  },
+  "amenity/bank|First Financial Bank": {
+    "countryCodes": ["us"],
+    "matchNames": ["1st financial bank"],
+    "tags": {
+      "amenity": "bank",
+      "brand": "First Financial Bank",
+      "brand:wikidata": "Q5453009",
+      "brand:wikipedia": "en:First Financial Bank (Ohio)",
+      "name": "First Financial Bank"
     }
   },
   "amenity/bank|First Midwest Bank": {
@@ -3356,13 +3368,15 @@
       "name": "Kasa Stefczyka"
     }
   },
-  "amenity/bank|Key Bank": {
+  "amenity/bank|KeyBank": {
+    "countryCodes": ["us"],
+    "matchNames": ["key bank"],
     "tags": {
       "amenity": "bank",
-      "brand": "Key Bank",
+      "brand": "KeyBank",
       "brand:wikidata": "Q1740314",
       "brand:wikipedia": "en:KeyBank",
-      "name": "Key Bank"
+      "name": "KeyBank"
     }
   },
   "amenity/bank|Komerční banka": {
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -225,6 +225,18 @@
       "takeaway": "yes"
     }
   },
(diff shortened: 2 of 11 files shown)


brands/shop/tyres.json
```diff
@@ -58,6 +58,17 @@
       "shop": "tyres"
     }
   },
+  "shop/tyres|Tire Discounters": {
+    "countryCodes": ["us"],
+    "matchTags": ["shop/tyres"],
```
bhousel: fyi I'm going to remove this `matchTags`, since it just matches itself

brands/shop/car_repair.json
```diff
@@ -47,6 +47,16 @@
       "shop": "car_repair"
     }
   },
+  "shop/car_repair|Car-X": {
+    "countryCodes": ["us"],
+    "matchNames": ["carx"],
```
bhousel: fyi I'm going to remove this `matchNames` because these aren't needed anymore for variations in case, spacing, punctuation.

brands/amenity/fast_food.json
```diff
@@ -609,6 +634,19 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|Graeter's": {
+    "countryCodes": ["us"],
+    "matchNames": ["greater's"],
```
bhousel: this one also not needed

brands/amenity/bank.json
```diff
@@ -3356,13 +3368,15 @@
       "name": "Kasa Stefczyka"
     }
   },
-  "amenity/bank|Key Bank": {
+  "amenity/bank|KeyBank": {
+    "countryCodes": ["us"],
+    "matchNames": ["key bank"],
```
bhousel: same

brands/amenity/bank.json
```diff
@@ -2955,6 +2956,17 @@
       "name": "First Citizens Bank"
     }
   },
+  "amenity/bank|First Financial Bank": {
+    "countryCodes": ["us"],
+    "matchNames": ["1st financial bank"],
```
bhousel: But this one is good!

brands/amenity/fast_food.json
bhousel: oh whoops, it actually is spelled differently - "ea" vs "ae" - nevermind!

bhousel:
Thanks @1ec5 👍 
I cleaned up the extra `matchTags` in fe5e6f8f and added a few new ones.

brands/shop/tyres.json
1ec5: Oops, I meant to move this to `shop=car_repair`.

bhousel: cool.. I ended up just putting both tags in the same matchGroup (so any of these tags will be recognized)

https://github.com/osmlab/name-suggestion-index/blob/18406f5f3070795ee6f3ea8856677d6f900037e1/config/match_groups.json#L69-L74

brands/amenity/fast_food.json
```diff
@@ -609,6 +634,19 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|Graeter's": {
```
tas50: If this is an ice cream place shouldn't it be amenity/ice_cream not fast_food:

https://github.com/osmlab/name-suggestion-index/blob/master/brands/amenity/ice_cream.json

brands/amenity/fuel.json
```diff
@@ -2114,6 +2114,16 @@
       "name": "Susco"
     }
   },
+  "amenity/fuel|TA": {
```
tas50: A full_name entry would be really useful here

brands/amenity/fast_food.json
1ec5: Oops, thanks! Fixed in 0fb6fe64945ed34384f12551ff108f5893288ef0.

brands/amenity/fuel.json
1ec5: I wasn’t sure about this one. The company is called “TravelCenters of America”, but no one would ever refer to an individual location by that name. It’s always “TA Centerville” or somesuch. Perhaps `operator` and `operator:wikidata`?

tas50: I always think of them as TA around here, but it would be nice for search to have the full name somewhere on the node that isn't displayed, but can be matched on

PR #2617, opened 2019-05-15
Add Vietnamese foodservice brands

1ec5:
Added several Vietnamese or Vietnamese-American foodservice brands.


bhousel:
Oh the matcher folds diacritics too - you can probably get rid of even more `matchNames`  😄 

Actually, I'm going to just merge this as-is because if we don't output a warning about match duplicates, that's something for me to fix, and these are good for testing that.

PR #2619, opened 2019-05-16
Fold the defunct Sleepy's into Mattress Firm

tas50:
Mattress Firm bought and rebranded all the Sleepy's locations. It's in the wiki we're tagging locations with. I created a challenge to fix all of these and we should see it drop out of the all names file next week since there are only 57 locations right now.

https://maproulette.org/admin/project/430/challenge/8003


bhousel:
Thanks @tas50 , looks good to me 👍

PR #2624, opened 2019-05-18
Distinguish between Checkers and Rally’s

1ec5:
I split apart [Wikidata’s item for Checkers](https://www.wikidata.org/wiki/Q3419341) into separate [Checkers](https://www.wikidata.org/wiki/Q63919315) and [Rally’s](https://www.wikidata.org/wiki/Q63919323) items, leaving the original to cover the umbrella company as opposed to the specific brands. This PR updates the `brand:wikidata` tags for these brands but also removes the `brand:wikidata` tag, since the Wikipedia article is for the pair of brands and no longer corresponds to the tagged Wikidata items. Now the index can associate each brand with its own logo.

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -243,8 +243,7 @@
     "tags": {
       "amenity": "fast_food",
       "brand": "Checkers",
-      "brand:wikidata": "Q3419341",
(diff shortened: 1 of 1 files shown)


matkoniecz:
Look good to me.

Issue #2622, opened 2019-05-18
Add Bruegger's Bagels, bagel restaurant


BjornRasmussen:
https://en.wikipedia.org/wiki/Bruegger's 
https://www.wikidata.org/wiki/Q4978656

matkoniecz:
What is the preferred tagging? `amenity=cafe` or `amenity=restaurant` or maybe `amenity=fast_food`? (I know that `restaurant` is in the title, but OSM tags are sometimes a bit tricky)

http://overpass-turbo.eu/s/J7p

BjornRasmussen:
Well, it has more of a feel of a restaurant based on what I can see from the outside (I've never been in one).  `amenity=restaurant` seems accurate, because of the style of the stores. 
There isn't a perfect match here unfortunately.

bhousel:
If people wait on you it's `amenity=restaurant`... If they don't it's `amenity=fast_food`
`amenity=cafe` is really for coffee houses.

HTH

bhousel:
Also, Wikipedia calls it a "fast casual" which in OSM means `amenity=fast_food`. 
These are places where you order from a counter and seat yourself (Panera Bread is a similar example).

iandees:
Yep, definitely fast food. Some branches are small enough that they might be called a cafe (i.e. they're primarily coffee and the bagels part is less important), but in general it's more fast food than restaurant.

BjornRasmussen:
Sounds good!

1ec5:
Oops, I added Bruegger’s in 46a822c7e46670234e4e67be3fb6589cc6ee0cd5 without noticing this issue. It’s definitely `fast_food`, not even fast-casual.

matkoniecz:
Is it maybe `name=Bruegger’s official_name=Bruegger's Bagels`?

1ec5:
The company’s name is Bruegger’s, but the restaurants I’ve been to are always signposted as Bruegger’s Bagels, with the “Bagels” set in the same type as the “Bruegger’s”.

PR #2626, opened 2019-05-18
Wahoo’s Fish Taco

1ec5:
This one is somewhat difficult to place. Wahoo’s is a fast-casual restaurant where you order at the front but have to wait a while; there’s no waitstaff, but the food comes on ceramic dishes. Existing nodes in OSM are split about half-and-half between `amenity=restaurant` and `amenity=fast_food`, but I went with `amenity=restaurant` figuring that it’s a broader classification.

As for the cuisine, `seafood` is the only one that’s for sure. It’s a fusion between `mexican`, `brazilian`, `hawaiian`, and more. Existing nodes in OSM are all over the place, but I don’t know if people would really expect to find Wahoo’s by searching for Mexican, Hawaiian, or Brazilian food per se.

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -1754,6 +1754,19 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|Wahoo's Fish Taco": {
+    "countryCodes": ["us"],
+    "matchNames": ["wahoo's"],
+    "tags": {
+      "amenity": "fast_food",
+      "brand": "Wahoo's Fish Taco",
+      "brand:wikidata": "Q7959827",
+      "brand:wikipedia": "en:Wahoo's Fish Taco",
+      "cuisine": "seafood",
+      "name": "Wahoo's Fish Taco",
+      "takeaway": "yes"
+    }
+  },
   "amenity/fast_food|Wendy's": {
     "tags": {
       "amenity": "fast_food",
--- a/brands/amenity/restaurant.json
+++ b/brands/amenity/restaurant.json
@@ -1448,18 +1448,6 @@
       "name": "Wagamama"
     }
   },
-  "amenity/restaurant|Wahoo's Fish Taco": {
-    "countryCodes": ["us"],
-    "matchNames": ["wahoo's"],
-    "tags": {
-      "amenity": "restaurant",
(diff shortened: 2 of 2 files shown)


bhousel:
> Wahoo’s is a fast-casual restaurant where you order at the front but have to wait a while; there’s no waitstaff, but the food comes on ceramic dishes. Existing nodes in OSM are split about half-and-half between `amenity=restaurant` and `amenity=fast_food`, but I went with `amenity=restaurant` figuring that it’s a broader classification.

Can you switch this to [`amenity=fast_food`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dfast_food)?

The OSM convention is to tag "fast casual" places this way.. 
If people wait on you it's `amenity=restaurant`... If they don't it's `amenity=fast_food`

(also came up today on https://github.com/osmlab/name-suggestion-index/issues/2622#issuecomment-493677645)

1ec5:
OK, I’ll change it. Wahoo’s has more service than other “fast-casual” places like Chipotle – for example, the food being served with silverware – but self-service versus full-service distinction seems like a good start, and I can at least imagine someone expecting Wahoo’s from searching for fast food.

It does make me wonder what to do about limited-service restaurants where you order at the front but they wait on you after that ([example](https://pastamarket.com/)).

Adamant36:
"It does make me wonder what to do about limited-service restaurants where you order at the front but they wait on you after that (example)."

Its an interesting question. Carl's Jr has been doing that for a while now and Mcdonald's is doing it in more restaurants. I don't think it changes the nature of them being fast food places though.

1ec5:
> Carl's Jr has been doing that for a while now and Mcdonald's is doing it in more restaurants. I don't think it changes the nature of them being fast food places though.

I haven’t seen it myself at Carl’s or McDonald’s, but I suspect what you’re referring to is more like what Fazoli’s has been doing for years. That describes Wahoo’s too, though they’re definitely higher on the spectrum of fast casual.

The bit about limited-service restaurants is a bit of a tangent that I should probably take to the mailing list or wiki. In the example I gave, the restaurant operates exactly as you’d expect of a typical American sit-down restaurant – silverware, white tablecloth, busing, frequent “how are you doing?”s – except that you order and pay upfront.

PR #2627, opened 2019-05-19
Major U.S. insurance companies with independent agents

1ec5:
Added a new `office/insurance` file containing major U.S. insurance companies that have independent agents in their own offices (typically auto and home insurance companies).

I included `name` in these entries, but the `name` should really be the agent’s name or d/b/a, which we can’t predetermine. Typically, the corporate branding is prominent but secondary to the agent’s name, and a single agent may be affiliated with two or more insurance companies.

### DIFF
--- a/brands/office/insurance.json
+++ b/brands/office/insurance.json
@@ -14,7 +14,7 @@
     "countryCodes": ["us"],
     "matchNames": ["american family", "amfam"],
     "tags": {
-      "brand": "Allstate",
+      "brand": "American Family Insurance",
       "brand:wikidata": "Q4743730",
       "brand:wikipedia": "en:American Family Insurance",
       "name": "American Family Insurance",
--- a/brands/office/insurance.json
+++ b/brands/office/insurance.json
@@ -1,0 +1,90 @@
+{
+  "office/insurance|Allstate": {
+    "countryCodes": ["us"],
+    "matchNames": ["allstate insurance"],
+    "tags": {
+      "brand": "Allstate",
+      "brand:wikidata": "Q2645636",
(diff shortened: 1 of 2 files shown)


matkoniecz:
>  but the name should really be the agent’s name or d/b/a, which we can’t predetermine.

In that case name tag probably should not be included.

bhousel:
Oh yeah - I've been thinking about these.  The index doesn't currently check the `office` tag, but maybe we should try to.  I'm not sure what kinds of things would get pulled in.

I do believe that these kinds of insurance agents definitely belong in the index (probably real estate ones too).

They should have a `name` tag, but when the preset downstream in iD shows both `name` and `brand` fields, mappers can override the `name`.  (This is similar to car dealerships and hotels, which often have a different name for each location).

brands/office/insurance.json
```diff
@@ -0,0 +1,90 @@
...
+      "brand:wikipedia": "en:Allstate",
+      "name": "Allstate",
+      "office": "insurance"
+    }
+  },
+  "office/insurance|American Family Insurance": {
+    "countryCodes": ["us"],
+    "matchNames": ["american family", "amfam"],
+    "tags": {
+      "brand": "Allstate",
```
bhousel: should be "American Family Insurance"  (this probably copy-pasted from "Allstate"?)

1ec5:
Insurance agencies are already tagged with `brand` quite frequently. That to me is a good sign that insurance brands should be in the index. In many cases, they’re signposted just as a fast food chain location might be (where the brand gets top billing), while in many other cases, they’re signposted just as a car dealership might be (where the brand is secondary but still prominent).

PR #2628, opened 2019-05-19
Add UK outdoor shops Millets and Blacks

SilentSpike:
Creating a PR since these currently don't have unique wikidata items and I'm not sure what statements I would add to the items in order to make them for this case.

https://en.wikipedia.org/wiki/Blacks_Outdoor_Retail explains the history of these brands.

### DIFF
--- a/brands/shop/outdoor.json
+++ b/brands/shop/outdoor.json
@@ -15,7 +15,7 @@
       "brand": "Blacks",
       "name": "Blacks",
       "operator:wikidata": "Q4923333",
-      "operator:wikipedia": "en:Blacks_Outdoor_Retail",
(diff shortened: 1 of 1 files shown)


Adamant36:
@SilentSpike, what about also adding Ultimate Outdoors? I guess they own that also. You might want to add the operator_name tag along with them to. 

I'll create the Wikidata entries for them if need be.

SilentSpike:
I've never actually encountered one so was just adding the brands I'm familiar with 😛

Looking at them now I'm not sure what was stopping me from creating wikidata items. I've been questioning lately whether such brands should be instances of `brick and mortar business` or just `brand` and the actual operating company entity is the business.

PR #2631, opened 2019-05-19
KinderCare

1ec5:
Added KinderCare, a U.S. childcare center chain, with a bevy of tags that I’m pretty sure are uniform across all locations. There are similar chains in other countries, but I’m not sure whether `amenity=kindergarten` or `amenity=childcare` would be better for those chains.


bhousel:
Thanks!  Seems ok 👍

PR #2639, opened 2019-05-20
Common U.S. fraternal orders

1ec5:
Added a few common U.S. fraternal orders, since a couple were already there. Existing tagging is pretty inconsistent: some bear the full organization’s name, others the nickname, still others the individual lodge name and number. It may be due to inconsistencies in signage from lodge to lodge. (We could probably expect the same situation with labor union halls.) I’m open to paring back or abandoning this PR if others share my uncertainty about how to tag the names.

### DIFF
--- a/brands/amenity/social_centre.json
+++ b/brands/amenity/social_centre.json
@@ -1,20 +1,77 @@
 {
-  "amenity/social_centre|American Legion": {
+  "amenity/social_centre|American Legion Hall": {
     "countryCodes": ["us"],
+    "matchNames": ["american legion"],
     "tags": {
+      "alt_name": "American Legion Hall",
       "amenity": "social_centre",
       "brand": "American Legion",
       "brand:wikidata": "Q468865",
       "brand:wikipedia": "en:American Legion",
-      "name": "American Legion",
+      "name": "American Legion Hall",
       "social_centre:for": "veterans"
     }
   },
-  "amenity/social_centre|Odd Fellow": {
+  "amenity/social_centre|Eagles Lodge": {
+    "countryCodes": ["us"],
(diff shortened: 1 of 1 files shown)


brands/amenity/social_centre.json
```diff
@@ -1,20 +1,77 @@
...
     "countryCodes": ["us"],
+    "matchNames": ["american legion"],
     "tags": {
+      "alt_name": "American Legion Hall",
       "amenity": "social_centre",
       "brand": "American Legion",
       "brand:wikidata": "Q468865",
       "brand:wikipedia": "en:American Legion",
-      "name": "American Legion",
+      "name": "American Legion Hall",
```
bhousel: Did you mean to add "Hall" here and also in the key above?
It's ok to be in `alt_name`.

bhousel:
I think it's ok.. In iD we'll display both `name` and `brand` so users can change the `name` to include the lodge number if that's how people want to see it.

brands/amenity/social_centre.json
1ec5: I changed it for consistency with the other entries – `name=Benevolent and Protective Order of Elks` is a real mouthful compared to `name=Elks Lodge`, which can get the mapper started for a locally appropriate lodge name and number. But maybe the entries shouldn’t be consistent with each other, because it is common to see an American Legion Hall signposted as just “American Legion”, whereas it’s quite uncommon to see “Benevolent and Protective Order of Elks” anywhere out in the field.

bhousel: Ok seems fine.. I don't really have a strong preference, and encouraging consistency makes sense.

PR #2651, opened 2019-05-22
Add Brazilian branch of Itaú bank.

vgeorge:
Itaú has branches in many countries, this change separates the Brazilian and Chilean. Also improves name matching, as it might be spelled without the accent.


bhousel:
Thanks @vgeorge  and welcome! 👍

PR #2667, opened 2019-05-26
Add Orpi estate agent

bmillemathias:
Name field not set because it seems the offices have their own name.


Adamant36:
Hhhmm, I think the name field is required. I'm not sure though.

bmillemathias:
estate agent are affilied to a network but have their own name

`
{
  "type": "node",
  "id": 2725058652,
  "lat": 45.5240294,
  "lon": 4.8759268,
  "tags": {
    "addr:city": "Vienne",
    "addr:country": "FR",
    "addr:housenumber": "9",
    "addr:postcode": "38200",
    "addr:street": "Rue Henri Jacquier",
    "contact:phone": "+33474316492",
    "name": "Orpi Pierre Olivier Immobilier",
    "office": "estate_agent",
    "opening_hours": "Mo-Fr 09:00-12:00, 14:00-18:00",
    "website": "https://www.orpi.com/vienne"
  }
},
`

Adamant36:
"estate agents are affilied to a network but have their own name"

I'm aware. It doesn't mean the index wouldn't still use the name tag. It's the same for hotels, but the name tag is still used. Personally, I think in these cases the operator tag is more appropriate for the branch name. For instance, people usually just search for RE/MAX and it's probably what should show up on the map. Using the branch operator as the name just unnecessarily over complicates things in my opinion.

bhousel:
Yes, sorry - For the purposes of this index, `name` field is required.  Consider this "what gets rendered".
Users can change the name field if they want to.

PR #2668, opened 2019-05-26
Add Foncia estate agent

bmillemathias:


Adamant36:
Sure.

Adamant36:
Hey anyone that knows, is it OK to not have the name tag on these? I probably should have checked before I merged it, but I haven't ran into it before. 

@matkoniecz any opinions?

bmillemathias:
some examples I came across
```
{
  "type": "node",
  "id": 2610803784,
  "lat": 45.8240443,
  "lon": 4.9529579,
  "tags": {
    "name": "Foncia Songival",
    "office": "estate_agent"
  }
},
...
{
  "type": "node",
  "id": 2701312181,
  "lat": 48.8962642,
  "lon": 2.1053879,
  "tags": {
    "addr:city": "Le Pecq",
    "addr:housenumber": "16",
    "addr:postcode": "78230",
    "addr:street": "Rue de Paris",
    "name": "Foncia Cézanne",
    "office": "estate_agent"
  }
},
```

matkoniecz:
> Hey anyone that knows, is it OK to not have the name tag on these?

It should be OK. Main negative effect is that in case of matching ny name tag as in Vespucci such entries will be ignored.

If such entities have commonly used names we will need to ignore then manually, but it is nothing serious or problematic.

matkoniecz:
I assume here that for each object name is separate and mapper is expected to fill it.

Adamant36:
I don't know. I feel like the branch name should use the operator tag or something instead of the name tag and the name should be name of the company. Like people just look up Re/Max. The operator isn't that important and we go for what renders best on a map. That's how its done with banks and hotels. We don't just leave the name tag out because they have regional branches. Its not like someone can't manual edit the name tag if need be anyway.

Adamant36:
@matkoniecz, see https://github.com/osmlab/name-suggestion-index/pull/2667#issuecomment-496045952. Which is what I figured.

PR #2671, opened 2019-05-26
Outback Steakhouse is a steakhouse

1ec5:
Sure, Outback Steakhouse does sell a variety of American food items from burgers to… onions. But it’s most strongly identified with steak.

### DIFF
--- a/brands/amenity/restaurant.json
+++ b/brands/amenity/restaurant.json
@@ -1012,7 +1012,7 @@
       "brand": "Outback Steakhouse",
       "brand:wikidata": "Q1064893",
       "brand:wikipedia": "en:Outback Steakhouse",
-      "cuisine": "american",
+      "cuisine": "steak_house;american",
       "name": "Outback Steakhouse"
     }
   },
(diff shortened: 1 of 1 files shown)


Adamant36:
I would almost say the American cuisine tag is pointless in this case, because it's sort of a given that steak houses are American restaurants. I don't think there are steakhouses that don't mainly serve American food. I could be wrong though. 

According to Wikipedia "The steakhouse started in the United States in the mid-19th century as a development from traditional inns and bars." Sounds uniquely American to me.

iandees:
When I think of a steakhouse I think of an `amenity=restaurant` that has waitstaff and is middle- to upper-end expensive. I would tag Outback the same way I tag Applebee's or Chili's.

1ec5:
Are there better ways than `cuisine` to differentiate sit-down restaurants from casual restaurants? Anyways, `cuisine=steak` is another commonly used option that doesn’t bring connotations about the level of service. It’s similar to `pizza`, which doesn’t necessarily connote a family-style pizzeria.

tas50:
Merge away

PR #2672, opened 2019-05-26
Jenny Craig

1ec5:
Added the Jenny Craig weight loss center chain. Is there a better set of tags for this chain (and others like it)?


Adamant36:
Seems solid to me.

1ec5:
Thanks, I added a bunch more `matchedTags` based on existing features, excluding some outlandish ones like `shop=convenience` and `shop=variety_store`.

PR #2676, opened 2019-05-27
Rename Pearle to Pearle Vision

tas50:
It's on the wikipedia page, their website, and their past and present
logos.

### DIFF
--- a/brands/shop/optician.json
+++ b/brands/shop/optician.json
@@ -237,13 +237,13 @@
       "shop": "optician"
     }
   },
-  "shop/optician|Pearle": {
-    "matchNames": ["pearle vision"],
+  "shop/optician|Pearle Vision": {
+    "matchNames": ["pearle"],
     "tags": {
-      "brand": "Pearle",
+      "brand": "Pearle Vision",
       "brand:wikidata": "Q2231148",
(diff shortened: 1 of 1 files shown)


matkoniecz:
Note OSM data - http://overpass-turbo.eu/s/JmJ http://overpass-turbo.eu/s/JmL

Austria, Belgium, Netherlands have large number of `name=Pearle shop=optican`

Is it case of two separate brands or brand using two versions? Maybe country codes limitation is necessary?

Adamant36:
@matkoniecz, it seems to be two different brands. Here's the Wikipedia article for Pearle https://nl.wikipedia.org/wiki/Pearle

tas50:
Looks like we have 2 distinct brands. Let's hold off on this PR until I figure out if it should be split.

bhousel:
This seems fine..

tas50:
I still plan to figure out the European vs. US branding of this so we can autocorrect both correctly.

bhousel:
That’s fine.. I just drove by a few of these this week and they all say “pearle vision”

tas50:
In the US they're all Pearle Vision. Apparently not so much in Europe.

PR #2680, opened 2019-05-27
Benihana

1ec5:
Added the Benihana restaurant chain. For Japanese steakhouses, is it better to couple `cuisine=japanese` with `steak_house`, or should we reserve `steak_house` for Western cuisine and use `cuisine=teppanyaki`?

### DIFF
--- a/brands/amenity/restaurant.json
+++ b/brands/amenity/restaurant.json
@@ -73,6 +73,25 @@
       "name": "Bella Italia"
     }
   },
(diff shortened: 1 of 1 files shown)


Adamant36:
> should we reserve steak_house for Western cuisine and use cuisine=teppanyaki?

I say go with cuisine=teppanyaki. A steak house is a very specific thing in my opinion.

PR #2682, opened 2019-05-27
centrakor: add wikidata and change tag

### LINKED ISSUE #2681: shop=variety_store name=Centrakor without assigned brand:wikidata (objects in France)
In case that it is an existing brand - is it one that has its own Wikipedia page or a Wikidata entry?

Or maybe it is an unwanted generic name (like amenity=restaurant name=Restaurant) rather than a brand?
Or maybe it is a valid name (like amenity=restaurant name=Bamboo), without an uniting brand?

For real brands without Wikipedia article and existing Wikidata page creating new Wikidata entry may be helpful, but for start confirmation that it is a valid brand would be helpful. See https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#memo--edit-wikidata for hints about Wikidata editing.

@Lulucmy @bmillemathias 

entry data:
```
{
  "countryCodes": [
    "fr"
  ],
  "tags": {
    "brand": "Centrakor",
    "name": "Centrakor",
    "shop": "variety_store"
  }
}
```

bmillemathias:
Looking at the website Centrakor is rather an interior_decoration shop
than a variety_store due the items and price of these items.
fix #2681


Adamant36:
You still have variety store as the tag in the interior_decoration file.

Adamant36:
Also, you need to include more external sources on the Wikidata page then just facebook. Otherwise, the entry will just get deleted.

Adamant36:
Thanks.

PR #2684, opened 2019-05-27
Include office tags in osmium filtering

bmillemathias:


Adamant36:
@bmillemathias, if you could remember to delete your branches after they get merged or closed it would be appreciated. You might want to use clearer branch names to. Otherwise, it's hard to tell the branches already merged from the ones that aren't merged yet.

PR #2686, opened 2019-05-27
Add groupama in office/insurance

bmillemathias:


brands/office/insurance.json
```diff
@@ -97,5 +97,14 @@
       "name": "State Farm",
       "office": "insurance"
     }
+  },
+  "office/insurance|Groupama": {
+    "tags": {
+      "brand": "Groupama",
```
tas50: You might want to add an "official_name" tag here as well with the full name of the company since Groupama is the shortened version.

bmillemathias: Hi @tas50 

I've look to wikipedia page but I don't really what is the `official_name` I should set,

Adamant36: Groupe des Assurances Mutuelles Agricoles or Group of Mutual Agricultural Insurances depending.

PR #2687, opened 2019-05-27
Update Clever fit fitness centre

ajaravete:
I added the wikidata and wikipedia information to the Clever fit fitness centre


Adamant36:
@ajaravete, clever fit seems to be in a lower case officially for some weird reason. For instance on their website. So that's probably not something that needs changing.

bhousel:
seems fine, thanks 👍

PR #2690, opened 2019-05-28
Sherwin-Williams is hyphenated

1ec5:
Sherwin-Williams is always written hyphenated in plain text. Although the chain’s main logo has “Sherwin” and “Williams” on two separate lines with no hyphen, the horizontal wordmark does have a hyphen.

[horizontal](https://commons.wikimedia.org/wiki/File:Sherwin-Williams_Paints_in_Gillette,_Wyoming.jpg)

### DIFF
--- a/brands/shop/paint.json
+++ b/brands/shop/paint.json
@@ -34,13 +34,13 @@
       "shop": "paint"
     }
   },
-  "shop/paint|Sherwin Williams": {
-    "matchNames": ["sherwin williams paints"],
+  "shop/paint|Sherwin-Williams": {
+    "matchNames": ["sherwin-williams paints"],
     "tags": {
-      "brand": "Sherwin Williams",
+      "brand": "Sherwin-Williams",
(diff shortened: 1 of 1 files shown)


matkoniecz:
Technical note, without covering whatever match is correct: 

AFAIK matchNames is not requiring explicit punctuation, so adding hyphen there is not necessary

In case that matchNames requires explicit punctuation - adding match for hyphenless version would be necessary

brands/shop/paint.json
```diff
@@ -34,13 +34,13 @@
       "shop": "paint"
     }
   },
-  "shop/paint|Sherwin Williams": {
-    "matchNames": ["sherwin williams paints"],
+  "shop/paint|Sherwin-Williams": {
+    "matchNames": ["sherwin-williams paints"],
```
1ec5: Fuzzy matching will work regardless of this particular change, but I made the change anyways to clarify that it’s the addition of “paints” that we want to match. 🤷‍♂

PR #2691, opened 2019-05-28
Winn✓Dixie has a hyphen

1ec5:
Winn-Dixie is written with a hyphen in plain text. The usual wordmark is stylized with a checkmark instead of a hyphen, but the hyphen can be seen inside the store and in front of some older locations.

[hyphen](https://commons.wikimedia.org/wiki/File:Winn_Dixie_Millidgeville,_GA_%287967695282%29.jpg)

### DIFF
--- a/brands/shop/supermarket.json
+++ b/brands/shop/supermarket.json
@@ -3632,13 +3632,13 @@
       "shop": "supermarket"
     }
   },
-  "shop/supermarket|Winn Dixie": {
+  "shop/supermarket|Winn-Dixie": {
     "countryCodes": ["us"],
     "tags": {
-      "brand": "Winn Dixie",
+      "brand": "Winn-Dixie",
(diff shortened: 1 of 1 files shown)


brands/shop/supermarket.json
```diff
@@ -3632,13 +3632,13 @@
       "shop": "supermarket"
     }
   },
-  "shop/supermarket|Winn Dixie": {
+  "shop/supermarket|Winn-Dixie": {
```
tas50: Does this need a match for the previous name or is - part of the fuzzy matching?

1ec5: Hyphens are elided as part of fuzzy matching, so no match is needed:

https://github.com/osmlab/name-suggestion-index/blob/0eda721266ad9cf516ee6cbe9f7d77609881aaa6/lib/simplify.js#L7

PR #2692, opened 2019-05-28
Start adding urgent care clinics in the US

tas50:
Due to our wonderfully backwards health care system there's a big
business for private care urgent care clinics. I added ZoomCare which is
a big one in Portland & Seattle. This could probably be expanded to
include some of the clinics here: https://www.healthdatamanagement.com/list/30-largest-urgent-care-center-chains-in-healthcare

### DIFF
--- a/brands/healthcare/clinic.json
+++ b/brands/healthcare/clinic.json
@@ -2,6 +2,7 @@
   "healthcare/clinic|ZoomCare": {
     "countryCodes": ["us"],
     "tags": {
+      "amenity": "clinic",
       "brand": "ZoomCare",
       "brand:wikidata": "Q64120374",
       "brand:wikipedia": "en:ZoomCare",
--- a/brands/healthcare/clinic.json
+++ b/brands/healthcare/clinic.json
@@ -1,0 +1,12 @@
+{
+  "healthcare/clinic|ZoomCare": {
+    "countryCodes": ["us"],
+    "tags": {
+      "brand": "ZoomCare",
+      "brand:wikidata": "Q64120374",
+      "brand:wikipedia": "en:ZoomCare",
+      "healthcare": "clinic",
+      "name": "ZoomCare"
+    }
+  }
+}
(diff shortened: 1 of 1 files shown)


brands/healthcare/clinic.json
```diff
@@ -0,0 +1,13 @@
+{
+  "healthcare/clinic|ZoomCare": {
```
1ec5: (Note that a `matchNames` of `zoom+care` is unnecessary because plus signs are removed in fuzzy matching.)

tas50: Yeah I figured that would just magically happen here

brands/healthcare/clinic.json
```diff
@@ -0,0 +1,13 @@
...
+    "countryCodes": ["us"],
+    "tags": {
+      "amenity": "clinic",
+      "brand": "ZoomCare",
+      "brand:wikidata": "Q64120374",
+      "brand:wikipedia": "en:ZoomCare",
+      "healthcare": "clinic",
```
1ec5: `urgent_care=yes` `healthcare:speciality=emergency` seems like an appropriate way to differentiate this kind of clinic from retail clinic and community health centers without resorting to `emergency=yes` (which is for EMS admission).

maxerickson: I think it's better to reserve healthcare:speciality=emergency for actual emergency rooms, some urgent care facilities aren't doing much more than primary care.

Glancing at their website, it appears some of their locations do provide a high level of service, but not all of them, and they still say to go to the ER if it is a real emergency.

tas50: So Zoom Care isn't just an urgent care sort of clinic. You can just use them as a normal doctor if you want. urgent_care=yes might make sense, but it's def not emergency specialty. It's people with colds and also just people getting physicals and yearly checkups.

tas50: Source: ZoomCare is my go to when I'm sick and don't want to head across town to my regular doctor.

maxerickson: urgent_care=yes is meant to indicate that walk-in type service is available, not that it is the only type of service available. I think I'm the main user of the tag (but I'm bad at documenting things on the wiki or asking tagging if they want to argue about things). https://overpass-turbo.eu/s/Jt5

tas50: I meant above that `urgent_care=yes` would be appropriate here since that is a big part of what they offer.

PR #2696, opened 2019-05-28
Split out First Citizens Bank

tas50:
These are being suggested as tag upgrades in the US and we're "fixing"
US banks with the Trinidad and Tobago when we do so. This adds the
actuual US bank, which is pretty large.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -3029,6 +3029,7 @@
     }
   },
   "amenity/bank|First Citizens Bank~(Trinidad and Tobago)": {
+    "countryCodes": ["bb", "tt"],
     "tags": {
       "amenity": "bank",
       "brand": "First Citizens Bank",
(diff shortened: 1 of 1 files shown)


matkoniecz:
I would also add country codes (especially as automated country code will skip such cases as it is unable to correctly split lists).

matkoniecz:
And thanks for catching this mistake/missed data!

brands/amenity/bank.json
```diff
@@ -3037,6 +3037,15 @@
       "name": "First Citizens Bank"
     }
   },
+  "amenity/bank|First Citizens Bank~(USA)": {
+    "tags": {
```
1ec5: While we’re at it, we could add `1st citizens bank` as a matching name for both, similar to the `1st financial bank` in the entry below.

tas50:
Yep I totally forgot the country codes. I'll get that added tomorrow.

brands/amenity/bank.json
tas50: Done

PR #2698, opened 2019-05-28
Saint Algue: Add wikidata reference

### LINKED ISSUE #2689: shop=hairdresser name=Saint Algue without assigned brand:wikidata (objects in France)
In case that it is an existing brand - is it one that has its own Wikipedia page or a Wikidata entry?

Or maybe it is an unwanted generic name (like amenity=restaurant name=Restaurant) rather than a brand?
Or maybe it is a valid name (like amenity=restaurant name=Bamboo), without an uniting brand?

For real brands without Wikipedia article and existing Wikidata page creating new Wikidata entry may be helpful, but for start confirmation that it is a valid brand would be helpful. See https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#memo--edit-wikidata for hints about Wikidata editing.

entry data:
```
{
  "tags": {
    "brand": "Saint Algue",
    "name": "Saint Algue",
    "shop": "hairdresser"
  }
}
```

@Lulucmy @bmillemathias

bmillemathias:
Saint Algue does not have a dedicated wikipedia page, this is a
subsidiary of Provaliance, a group specialized in hair-dessers salons

fix #2689

### DIFF
--- a/brands/shop/hairdresser.json
+++ b/brands/shop/hairdresser.json
@@ -156,6 +156,7 @@
   "shop/hairdresser|Saint Algue": {
     "tags": {
       "brand": "Saint Algue",
+      "brand:wikidata": "Q62973210",
       "name": "Saint Algue",
       "shop": "hairdresser"
     }


matkoniecz:
Note https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#adding-references-to-wikidata - currently without any references it is at risk of deletion

See https://www.wikidata.org/w/index.php?title=Wikidata:Administrators%27_noticeboard&oldid=941582891#Entries_that_should_be_now_fixed for discussion on Wikidata

PR #2700, opened 2019-05-28
Corrected name of The Home Depot

1ec5:
The Home Depot is very consistent in calling itself “The Home Depot” rather than just “Home Depot”. The “The” is prominent in the chain’s logo.

### DIFF
--- a/brands/shop/doityourself.json
+++ b/brands/shop/doityourself.json
@@ -204,17 +204,6 @@
       "shop": "doityourself"
     }
   },
-  "shop/doityourself|Home Depot": {
-    "countryCodes": ["ca", "mx", "pr", "us"],
-    "matchNames": ["the home depot"],
-    "tags": {
-      "brand": "Home Depot",
-      "brand:wikidata": "Q864407",
-      "brand:wikipedia": "en:The Home Depot",
-      "name": "Home Depot",
-      "shop": "doityourself"
-    }
-  },
   "shop/doityourself|Home Hardware Building Centre": {
     "countryCodes": ["ca"],
     "matchNames": ["home hardware"],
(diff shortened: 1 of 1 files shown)


Adamant36:
Uuugggh, know one calls it that though and it just extend the name by one more word on the map unnecessarily.

1ec5:
I’d say the on-the-ground rule supports including the “the”. It may be somewhat pedantic, but not to the point of causing any surprise. There are many other brands in the index with words that could be considered “unnecessary”, such as The UPS Store, The North Face, and The Salvation Army. That said, “Home Depot” is a perfectly legitimate way to refer to the store, so maybe it should go in `alt_name`.

Adamant36:
> “Home Depot” is a perfectly legitimate way to refer to the store, so maybe it should go in alt_name.

Alright. Good points. That might work. My personally preference is to use the official name tag for longer names and favor the shorter names on the map where we can. I guess "the" doesn't really make the name that much longer though. So, however you want to do it is fine. I'll just put my OCD aside on this one ;)

iandees:
Another consideration for the The problem is that, similar to the St/Street problem, map renderers can optionally shorten the name (remove the The) if they want, but it's harder to add the data back if it's not there.

tas50:
Another measure used in the past is what other providers do: both Apple and Google use the full name and both of them tend to heavily shorten names of brands.

1ec5:
c0e3f14d180d34429703fc092d6fdee018122b35 adds `alt_name=Home Depot` primarily for less sophisticated geocoders. `short_name` is another candidate, though I tend to use that tag for acronyms rather than slight shortenings.

Issue #2526, opened 2019-03-28
MRS - complete mismatch between data in name-suggestion-index and data mapped in OSM.


matkoniecz:
Countries declared in countryCodes field and where such objects are mapped in OSM is not matching.
it may be caused by:
* mistake in countryCodes field (fixable by changing it)
* by different brand using the same name, not present in the index (fixable by adding it to the index)
* mistake in OSM (fixable by editing OSM)
* suspicious data in OSM (fixable by opening an OSM note)

entry data:
````
{
  "count": 70,
  "countryCodes": [
    "ng"
  ],
  "tags": {
    "amenity": "fuel",
    "brand": "MRS",
    "brand:wikidata": "Q56274919",
    "brand:wikipedia": "en:MRS Oil Nigeria Plc",
    "name": "MRS"
  }
}
````

Country codes declared in name-suggestion-index:
* ng - 0 mapped entries

Additional countries with object of matching name and type mapped in OSM
cm - 210 mapped entries
tg - 47 mapped entries
bj - 9 mapped entries
gh - 4 mapped entries

you can use following query in the Overpas Turbo Wizard to look at this data:
'amenity'='fuel' and name='MRS' global

examples are listed below:
this is in country with code 'cm', according to name index present nearly not present outside ["ng"]
https://www.openstreetmap.org/node/3395107286

this is in country with code 'cm', according to name index present nearly not present outside ["ng"]
https://www.openstreetmap.org/node/3637104936

this is in country with code 'cm', according to name index present nearly not present outside ["ng"]
https://www.openstreetmap.org/node/4451074251

this is in country with code 'cm', according to name index present nearly not present outside ["ng"]
https://www.openstreetmap.org/node/4451102860

this is in country with code 'cm', according to name index present nearly not present outside ["ng"]
https://www.openstreetmap.org/node/4451102861

this is in country with code 'cm', according to name index present nearly not present outside ["ng"]
https://www.openstreetmap.org/node/4451102862

PR #2709, opened 2019-05-28
Finnish brand names

crimsondusk:
Added some notable Finnish brands


brands/leisure/fitness_centre.json
```diff
@@ -76,6 +76,16 @@
       "name": "Fitness First"
     }
   },
+  "leisure/fitness_centre|Fitness24Seven": {
+    "countryCodes": ["fi", "sv"],
```
matkoniecz: It is supposed to be Finland, Sweden, right? Note that `sv` is a language code for Swedish language, and country code for El Salvador.

Sweden has `se` country code.

brands/shop/sports.json
```diff
@@ -228,6 +228,26 @@
       "shop": "sports"
     }
   },
+  "shop/sports|Stadium": {
+    "countryCodes": ["de", "dk", "fi", "sv"],
```
matkoniecz: Sweden has `se` country code, El Salvardor `sv`.

brands/shop/sports.json
```diff
@@ -228,6 +228,26 @@
...
+    "tags": {
+      "brand": "Stadium",
+      "brand:wikidata": "Q4993863",
+      "brand:wikipedia": "sv:Stadium (detaljhandelskedja)",
+      "name": "Stadium",
+      "shop": "sports"
+    }
+  },
+  "shop/sports|XXL": {
+    "countryCodes": ["fi", "no", "sv"],
```
matkoniecz: Sweden has `se` country code, El Salvardor `sv`.

brands/leisure/fitness_centre.json
crimsondusk: That is supposed to be Sweden, yes. I'll fix that.

matkoniecz: Thanks for fixing!

brands/shop/electronics.json
```diff
@@ -146,6 +146,15 @@
       "shop": "electronics"
     }
   },
+  "shop/electronics|Gigantti": {
```
tas50: should there be a country code here?

matkoniecz: My script will autoadd country codes, manually specifying country codes. It will work well in cases where OSM data is describing real world well (each country where this brand is mapped has some elements mapped). Assuming that this brand actually appears only in Finland, then it will work here - http://overpass-turbo.eu/s/Jw9

matkoniecz: @crimsondusk It is Finland only brand, right? (if not - adding country code is certainly useful, for Finland only you can either add country code or wait for my script to do it).

matkoniecz:
I will merge it - in case Gigantti `shop=electronics` is not Finland-only brand then followup adding country codes will be nice.

Review from local community would be nice but waiting for someone to appear here and make review is unlikely to happen, so I will just hope that once someone notices some problem there will report an issue (asking on their mailing list would probably not help significantly with this).

(I have an idea about listing on talk ml open issues waiting with https://github.com/osmlab/name-suggestion-index/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3A%22check+local%22 but for now there are still issues with provided answers and waiting for commits)

Adamant36:
@matkoniecz, how does it work when someone wants to map a brand in a country where it's not mapped yet but still has a store there? Wouldn't it filter the store out of search research results in that case because your script didn't auto add the country code due to no stores being mapped there? 

It seems like it would be better to pick country codes based more official sources than the map. Since the index is as much about improving information for stores mapped in the future then it is basing what is yet mapped off of what already is (if that makes any sense).

matkoniecz:
> It seems like it would be better to pick country codes based more official sources than the map.

Yes, but the difference is that map is machine-readable and official sources are not. If someone is interested I may generate list of entries without `countryCodes` specified or where `countryCodes` were added automatically based on the map data.

Though for now we have multiple known significant mismatches waiting for research - see https://github.com/osmlab/name-suggestion-index/search?q=mismatch+between+data+is%3Aissue&state=open&type=Issues and processing them first is probably a good idea (as in this case there is certainly something wrong)

> Wouldn't it filter the store out of search research results in that case because your script didn't auto add the country code due to no stores being mapped there?

Yes, first mapped shops will be affected. But my script is also monitoring for cases where brand is mapped and index entry is missing (see for example https://github.com/osmlab/name-suggestion-index/issues/2526 ).

Given that for very significant majority map data matches reality well and too many unrelated entries in search results is also bad. I think that balance here - significantly reduced number of unwanted entries in search results vs initial shops mapped in a country will not benefit from nsi entry is strongly positive.

There are also secondary effects like time necessary to develop script vs many mistakes found as byproduct like `se`/`sv` or `ie`/`ir` poblems and other issues found by my script.

brands/shop/electronics.json
crimsondusk: Looks like I forgot to add the country code here.

PR #2712, opened 2019-05-29
malangligis kelkajn "brand:wikipedia"

verdulo:
Mi ŝanĝis kelkajn etikedojn "brand:wikipedia" por ligi al artikoloj en Pola aŭ en Esperanto. Anglalingvanoj ne havas rajton altrudi sian lingvaĉon al la resto de mondo.

Zmieniłem kilka etykiet "brand:wikipedia", aby łączyły się z artykułami w języku polskim lub Esperanto. Anglicy nie mają prawa narzucać swojego języka reszcie świata.

### DIFF
--- a/brands/shop/clothes.json
+++ b/brands/shop/clothes.json
@@ -166,7 +166,7 @@
     "tags": {
       "brand": "Bershka",
       "brand:wikidata": "Q827258",
-      "brand:wikipedia": "en:Bershka",
+      "brand:wikipedia": "pl:Bershka",
       "name": "Bershka",
       "shop": "clothes"
     }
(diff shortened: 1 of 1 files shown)


Adamant36:
English isn't "imposing" their language on the rest of the world. These aren't mainly Polish brands and English is spoken in a lot more places where they are located than Polish is. If they were confined to Polish countries I'd zero problem with them being linked to Polish Wikipedia articles. As it is though, I suggest we reject this. It's not like people can't go to the Polish Wikipedia articles pretty easily from the Wikidata entries anyway. Or why not make it any other language?

matkoniecz:
At this moment, unfortunately, only one specific `brand:wikipedia` tag may be set for each entry. Using local wikipedia versions for given area is preferable during editing and it may be a good idea to somehow provide info that would make it easier.

I am closing this pull request as this PR is not actually helping to do this.

PR #2716, opened 2019-05-30
Aladdin Bail Bonds

1ec5:
Added Aladdin Bail Bonds. Several tags are currently being used for bail bond agents; I went with `office=bail_bond_agent`, the second most popular tag behind `amenity=bail_bonds` and ahead of `shop=bail_bonds`. I’ve never been inside one, but I think they feel more like insurance offices (`office=insurance`) than payday loan shops (`shop=money_lender`) and bank branches (`amenity=bank`) where you can quickly complete some transactions without sitting down. But `office=bail_bonds` seems reasonable too.

[aladdin](https://www.mapillary.com/map/im/Vlp_1SwJR2K80imWlq9w6g)

### DIFF
--- a/brands/office/bail_bond_agent.json
+++ b/brands/office/bail_bond_agent.json
@@ -1,0 +1,26 @@
+{
+  "office/bail_bond_agent|Aladdin Bail Bonds": {
+    "countryCodes": ["us"],
+    "matchNames": [
+      "aladdin",
+      "aladin bail bonds",
+      "alladin bail bonds"
+    ],
+    "matchTags": [
+      "amenity/bail_bond",
+      "amenity/bail_bonds",
+      "office/bail_bond",
+      "office/bail_bonds",
+      "shop/bail_bond",
+      "shop/bail_bonds",
+      "shop/money_lender"
+    ],
+    "tags": {
+      "brand": "Aladdin Bail Bonds",
+      "brand:wikidata": "Q64166257",
+      "name": "Aladdin Bail Bonds",
+      "office": "bail_bond_agent",
+      "opening_hours": "24/7"
+    }
+  }
+}


matkoniecz:
`office=bail_bond_agent` - can you consider adding it to https://wiki.openstreetmap.org/wiki/Key:office#Office ?

Or give me definition of such object?

there is https://en.wikipedia.org/wiki/Bail_bondsman

> A bail bondsman, bail bond agent or bond dealer is any person, agency or corporation that will act as a surety and pledge money or property as bail for the appearance of a defendant in court.

but it a bit unclear, and what worse I have no good idea for something better

Adamant36:
"I’ve never been inside one, but I think they feel more like insurance offices (office=insurance)"

I have, they have the look/feel of an insurance place and that's essentially what they are. Their putting money toward the possible risk imposed on soceity by letting someone out of jail before their sentancing. Seems like an insurance thing to me. 

Personally id go with office=insurance + insurance=bail_bonds. 

Generally, I like the idea of implementing an insurance=* tag more to help add details. Insurance=health is already being used in a few cases. So why not add it to index.

1ec5:
@matkoniecz, certainly. This PR is just me thinking out loud about the best way to tag these offices. In the past, I’ve thrown my hands up and used the “all of the above” tagging scheme, but Aladdin is big enough that it’s worth standardizing on a tag, including documenting it on the wiki and getting presets added to editors.

> Personally id go with office=insurance + insurance=bail_bonds.

I hadn’t thought of `insurance=*`, thanks. Currently, the key seems to be used primarily as an unstructured `brand` tag for `office=insurance`, but `insurance=private` and `insurance=health` are good precedents.

bhousel:
What you did seems fine..
Calling a bail bond place an "insurance" agent seems really mistaken to me.

1ec5:
> can you consider adding it to https://wiki.openstreetmap.org/wiki/Key:office#Office ?

I added it to the list and [documented it here](https://wiki.openstreetmap.org/wiki/Tag:office%3Dbail_bond_agent).

PR #2722, opened 2019-05-30
Add Generali insurance

bmillemathias:


tas50:
Should there be a country code for this?

bmillemathias:
> 
> 
> Should there be a country code for this?

I don't know where to get the list where Generali is present.

Adamant36:
Seems like they are mostly tagged in Europe. Although from their site it looks like they are world wide.

Have you thought about adding the insurance=health tag for the health related insurance PRs you are doing? It might be worth adding.

bmillemathias:
> Seems like they are mostly tagged in Europe. Although from their site it looks like they are world wide.
> 

> Have you thought about adding the insurance=health tag for the health related insurance PRs you are doing? It might be worth adding if it fits.

Well, the companies I added - at least the french ones - are selling lots of kind of insurances: car, house, ... so that's would be easy, but Generali is worldwide and adding insurance=health would implies that they sell that in all countries, which I don't know.

Adamant36:
> but Generali is worldwide and adding insurance=health would implies that they sell that in all countries, which I don't know.

Yeah, when I looked them up it seemed like they had a bunch of subsidiaries all over selling different kinds of insurance. So it probably wouldn't be doable in this case. Unless we split them up by brand. Which might be worth doing at some point. But probably not for now.

bhousel:
yea, this is good enough for now..

ImreSamu:
just some info:

in Hungary : web: https://www.generali.hu/eng  
* https://www.wikidata.org/wiki/Q16519988  + https://hu.wikipedia.org/wiki/Generali_Biztos%C3%ADt%C3%B3  


Global:
*  https://www.generali.com/who-we-are/global-positioning

bhousel:
If `insurance=*` is not the same everywhere, we should not add it here.  Let mappers add it if they want to.

ImreSamu:
sorry if I was not clear enough.
* the current "merged" version - is OK for me.
* just wanted to add some extra info - for future improvements.

PR #2723, opened 2019-05-30
Add Aviva Insurance

bmillemathias:


Adamant36:
@matkoniecz any objections to adding the insurance=* tag to these going forward so we can tell them apart by insurance type? If not, I'll probably go through and add it to the existing entries. I think it would be helpful.

tas50:
If we know the type of insurance offered we should start adding that. That sort of tag information is why this repo is great. It auto fills in the things that very few casual mappers know to do.

matkoniecz:
> `insurance=* tag` 

only in cases where you are at least 99,9% sure that it actually applies to all of them

PR #2725, opened 2019-05-30
Add tag for Yves Rocher

bmillemathias:
as suggested by `npm run build`


matkoniecz:
Matches current tagging http://overpass-turbo.eu/s/Jyj and my vague knowledge

PR #2726, opened 2019-05-30
Removed shop/hairdresser|Шарм

tevez07b9:
Since this is not a brand. I think it should be removed


matkoniecz:

PR #2727, opened 2019-05-31
remove unwanted 'undefined' polluting distributed files

matkoniecz:
done as PR as it is changing not just data but rather code, and maybe there is a better way to fix this issue.

note to self: after this problem is fixed (maybe by merging this) make PR to update nsi data in Vespucci - it is really old (see https://github.com/MarcusWolschon/osmeditor4android/commit/4cb4a0872f19dd8e2ed0cdd3a24add7f3e641170#diff-0ebe795d71a85fa6a5d427a48a1448b9 )


dist/taginfo.json
```diff
@@ -11,12 +11,14 @@
     "contact_email": "bryan@mapbox.com"
   },
   "tags": [
+    {"key": "access", "value": "customers"},
```
bhousel: ugh, how did this sneak in 😕

dist/taginfo.json
```diff
@@ -312,10 +345,14 @@
     {"key": "short_name:ja"},
     {"key": "social_centre:for", "value": "veterans"},
     {"key": "sport", "value": "golf"},
+    {"key": "sport", "value": "yoga"},
+    {"key": "supervised", "value": "no"},
     {"key": "takeaway", "value": "only"},
     {"key": "takeaway", "value": "yes"},
     {"key": "tourism", "value": "hotel"},
     {"key": "tourism", "value": "motel"},
+    {"key": "trade", "value": "building_supplies"},
```
bhousel: another weird one 👀

bhousel:
thanks @matkoniecz

dist/taginfo.json
matkoniecz: https://github.com/osmlab/name-suggestion-index/commit/0288e553c7debac15de8de616577bad6c4887ea7#diff-87db3b2155a8ad976aeda84255eabc22

dist/taginfo.json
matkoniecz: https://github.com/osmlab/name-suggestion-index/blob/94f7cf527f4570efb85f6556f3223ccc6700e2ae/brands/leisure/playground.json - hard to say is it correct, McDonalds in Poland sometimes have playgrounds but not branded ones

bhousel: I took them out in 2782d7b6

dist/taginfo.json
bhousel: I guess this one is alright.  I didn't realize we had `shop/trade` now.

PR #2735, opened 2019-06-01
Australian brands

andrewharvey:


brands/shop/bakery.json
```diff
@@ -82,6 +82,17 @@
       "shop": "bakery"
     }
   },
+  "shop/bakery|Breadtop": {
+    "countryCodes": ["au"],
+    "tags": {
+      "brand": "Breadtop",
+      "brand:wikidata": "Q4959217",
+      "brand:wikipedia": "en:Breadtop",
+      "cuisine": "chinese",
```
matkoniecz: I am not an expert but from looking at their website they use asian-related styling but products seems to not be limited to a Chinese cuisine.

Wikipedia article describes it as "selling a wide range of bread, cakes, buns and pastries".

Are you sure that `cuisine=chinese` tag fits here? (again, just checking is it intentional)

andrewharvey: Yes the products themselves are mostly Chinese cuisine (with the exception of the odd croissant), the cuisine they serve is different to an European or typical Australian (which would mostly be european) bakery.

Adamant36:
What exactly about 85C Bakery Cafe makes it Chinese cuisine? Nothing on their menu looks that way. The only thing Chinese about them that I can tell is that they are a Chinese company.

matkoniecz:
@Adamant36 If you think it is wrong/suspect and PR is merged - please open an issue, comment in closed PR may be easily missed.

Adamant36:
> @Adamant36 If you think it is wrong/suspect and PR is merged - please open an issue, comment in closed PR may be easily missed.

I can do that. I figured if you had a quick answer, due to something I missed, that it would be useless to open a new issue for it.

brands/shop/furniture.json
```diff
@@ -182,6 +182,15 @@
       "shop": "furniture"
     }
   },
+  "shop/furniture|Freedom Furniture": {
```
1ec5: A `matchNames` of “Freedom Furniture” could be added here, since the `name` is only “Freedom”.

brands/shop/mall.json
```diff
@@ -0,0 +1,11 @@
+{
+  "shop/mall|Westfield": {
+    "countryCodes": ["au"],
+    "tags": {
+      "brand": "Westfield",
+      "brand:wikidata": "Q2036076",
+      "brand:wikipedia": "en:Westfield Group",
```
1ec5: This article and Wikidata item pertain to a former company (two companies ago, actually). The closest thing left is [this disambiguation page of lists of malls](https://en.wikipedia.org/wiki/List_of_Westfield_shopping_centres), which points to the need for a separate Wikidata item covering the brand that’s in use by two different companies (one in 🇦🇺🇳🇿, the other in 🇬🇧🇺🇸).

1ec5: → #2743

Adamant36:
@andrewharvey, how come you tagged 85C Bakery Cafe cuisine=chinese?

andrewharvey:
> @andrewharvey, how come you tagged 85C Bakery Cafe cuisine=chinese?

Maybe taiwanese is more accurate, or just a general asian. The drinks are asian style, the bread is mostly asian style. The cakes are more western but some with more asian flavours.

The cuisine/style is vastly different to say Bakers Delight.

I'm only talking about the AU stores, if other regions are different we should confirm that and use different tags.

brands/shop/mall.json
andrewharvey: Ok thanks, I'm happy to just remove the brand:wikidata and brand:wikipedia tags then. Should I do a new PR which removes them as a quick fix?

brands/shop/furniture.json
bhousel: good catch - fixed this in 00bc22fd

brands/shop/mall.json
bhousel: I just removed them in ea284122

Adamant36:
> Maybe taiwanese is more accurate, or just a general asian. The drinks are asian style, the bread is mostly asian style.

If a similar tag is being kept, probably Asian would be better. Green tea could be from any Asian region and I also saw that they sell a Japanese style bread. Which isn't Chinese either. Thanks for the explanation.

andrewharvey:
> If a similar tag is being kept, probably Asian would be better. Green tea could be from any Asian region and I also saw that they sell a Japanese style bread. Which isn't Chinese either. Thanks for the explanation.

I'm okay if you want to change it to `asian`.

PR #2736, opened 2019-06-01
Correct ExtraMile wiki tags

1ec5:
`operator` should be reserved for the franchisee of a gas station or convenience store, not the franchise’s corporate parent. Wikidata is better suited for detailing corporate structures. I created [a new Wikidata item](https://www.wikidata.org/wiki/Q64224605) to represent the ExtraMile brand, which makes it possible for the index to feature the store’s own logo instead of the Chevron logo. If a data consumer does need to determine the brand’s corporate parent, it can refer to the [parent organization](https://www.wikidata.org/wiki/Property:P749) claim in the Wikidata item.

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -426,10 +426,8 @@
     "countryCodes": ["us"],
     "tags": {
       "brand": "ExtraMile",
+      "brand:wikidata": "Q64224605",
       "name": "ExtraMile",
-      "operator": "Chevron",
-      "operator:wikidata": "Q319642",
-      "operator:wikipedia": "en:Chevron Corporation",
       "shop": "convenience"
     }
   },


matkoniecz:
Yes, operator tag is about operator of a specific shop, not about who owns entity operating it.

Adamant36:
"operator should be reserved for the franchisee of a gas station or convenience store, not the franchise’s corporate parent."

Its used that way all over in the index though. Sometimes the corporate parent is the operator or at least in places where its not known if it is, it keeps people from mistakenly adding the corporate parents Wikipedia information as the brand wikipedia information. Which happens more  often then really should. A lot of times the "true" operator isn't known but having the information there is better than not. There was some discussion about it in the past and that seemed to be what we agreed on. There's no reason in this case there couldn't have just been both. In a lot of cases, like with fast food restaurants, its not clear which stores are corporate run versus not or how much actual control the parent has over the franchise. In a lot of cases the franchise owner might have practically zero say in store operations outside of being an intial investor and filing paperwork. I wouldn't call them the operator though. In those cases it serves nothing to remove the operator information and we don't know which are which anyway. So its not as simple as saying "operator isn't about who owns the entity", because a lot of times it is. Even in cases where it might seem like it. 

Unfortunately there's no "owner" or similar tag in OSM. So I don't know what else would work for that. There was a lot of discussion about it in the past and it was generally agreed that operator was ok usage in absence of anything better tp show that information. While Wikidata is technically better for showing that kind of information, its still useful to show it in OSM and hopefully a better way to represent the data will be used instead of it just being removed outright. If nothing else, it at least saves us on redundant PRs by ignorant people wanting to "fix" wikipedia references. I still think corprate structure is useful information to map viewers that wont go to Wikidata to find it though. In the same way that its useful to know what government entity maintains a trail or who runs a park etc. That's just my opinion. So hopefully it operator information doesn't removed out of some false rational that it's not useful or better served somewhere else. At least not without a thorough discussion of it first.

1ec5:
To rephrase, a franchise store’s `operator` should be the franchisee if known, but not the franchise’s corporate parent. Otherwise, we’d be tagging `operator=Yum! Brands` on every KFC store and Pepsi vending machine, which would be pedantic at best and misleading at worst.

> Unfortunately there's no "owner" or similar tag in OSM.

In fact, there is [`owner`](https://wiki.openstreetmap.org/wiki/Key:owner). Perhaps editors could add support for it, but it might just create more confusion.

> its still useful to show it in OSM and hopefully a better way to represent the data will be used instead of it just being removed outright

To be clear, this PR would only prevent the editor from “upgrading” a feature to have a misleading `operator` tag. (There are no corporate-owned ExtraMile stores as far as I know.) This PR doesn’t remove `operator` from any existing feature, even those that had previously gained `operator` tags due to this index entry. Instead, it _adds_ a brand with its Wikidata QID, enabling an editor or renderer to use the ExtraMile logo instead of the Chevron logo, and deincentivizing mappers from adding `website=https://www.chevronextramile.com/`.

Adamant36:
> we’d be tagging operator=Yum! Brands on every KFC store and Pepsi vending machine, which would be pedantic at best and misleading at worst.

Maybe. I use to care about that stuff more, but I keeping things simple and concise is just such an uphill battle when it comes to OSM. That's probably just me though, but I agree. The less pedanticness the better and there's no reason to mislead people. 

> In fact, there is owner.

True. I meant owner:wikipedia. Since it would be replacing operator:wikipedia in cases like this. My bad. It looks like it only has 4 uses and there isn't a wiki article for it or anything. Maybe I'll create one. 

> it adds a brand with its Wikidata QID, enabling an editor or renderer to use the ExtraMile logo instead of the Chevron logo, and disincentivizing mappers from adding 

I have zero problem with that. I was just speaking to the use of operator:wikipedia more generally. I fully agree it shouldn't be used in a misleading way. It's a muddy tag in the first place and there should really be a better one for multinational, franchisable businesses like Chevron. I don't really think it was created for that. I almost don't think owner works in those cases either.

bhousel:
>> "operator should be reserved for the franchisee of a gas station or convenience store, not the franchise’s corporate parent."
> 
> Its used that way all over in the index though. 

If you see it misused in the index, can you remove the `operator` that are wrong?  

The only time I've ever used the tag was to tag Wal-Mart as the `operator` for all Sam's Clubs. 

> Sometimes the corporate parent is the operator or at least in places where its not known if it is, it keeps people from mistakenly adding the corporate parents Wikipedia information as the brand wikipedia information.

We don't want to add tags that are "maybe wrong" just to prevent other people from entering data that's "maybe wrong".  The tags in this index really should be "always right".

Adamant36:
> If you see it misused in the index, can you remove the operator that are wrong?

Yeah sure. 

> We don't want to add tags that are "maybe wrong" just to prevent other people from entering data that's "maybe wrong". The tags in this index really should be "always right".

True.

PR #2738, opened 2019-06-01
Add some vehicule_inspection brands

bmillemathias:


brands/amenity/vehicule_inspection.json
```diff
@@ -0,0 +1,40 @@
+{
+  "amenity/vehicule_inspection|Autosur": {
+    "countryCodes": ["fr"],
+    "tags": {
+      "amenity": "vehicule_inspection",
```
matkoniecz: rather `amenity=vehicle_inspection`, without u (see https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dvehicle_inspection )

Adamant36: @matkoniecz, that's just us forcing our English on the rest of the world ;)

PR #2744, opened 2019-06-02
Remove discard filters

tohaklim:
I believe names from this index are flagged as generic - the ones removed are not, per se


bhousel:
Could you explain more the intent behind this?
What are the names you are removing?

tohaklim:
they are what I'd refer to as "semi generic", as in they are not brand names, yet should not be discouraged as are still valid names

bhousel:
ok - can you instead move them up to the `discardKeys` section so that they don't creep back into the index?

tohaklim:
oh, that's what I was missing. Moved them, hopefully the syntax is ok

bhousel:
> oh, that's what I was missing. Moved them, hopefully the syntax is ok

Thanks!
If you `npm run build` and the unwanted entries don't return - the syntax is ok 😄

PR #2745, opened 2019-06-03
add event and hoyts au/nz cinemas

andrewharvey:


Adamant36:
Can you add Event Cinemas as the official name? It seems like it's called that in a lot of places.

andrewharvey:
> Can you add Event Cinemas as the official name? It seems like it's called that in a lot of places.

That sounds okay, I've updated the PR.

Adamant36:
Thanks

PR #2747, opened 2019-06-03
Distinguish Ulta cosmetics store from beauty studios

1ec5:
Copied the Ulta Beauty entry as a `shop=cosmetics` entry. An Ultra Beauty store is primarily a cosmetics store, but there are salon studios inside it. This index considers a drugstore to be separate from the pharmacy counter within it, so by analogy the cosmetic store should be separate from the salon studios within it.

### DIFF
--- a/brands/shop/beauty.json
+++ b/brands/shop/beauty.json
@@ -49,8 +49,9 @@
   "shop/beauty|Ulta Beauty": {
     "countryCodes": ["us"],
     "matchNames": ["ulta"],
+    "nomatch": ["shop/cosmetics|Ulta Beauty"],
     "tags": {
-      "beauty": "cosmetics;eyebrow;eyelash;hair;skin_care;supplies;waxing",
+      "beauty": "eyebrow;eyelash;hair;skin_care;waxing",
       "brand": "Ulta Beauty",
       "brand:wikidata": "Q7880076",
       "brand:wikipedia": "en:Ulta Beauty",
--- a/brands/shop/cosmetics.json
+++ b/brands/shop/cosmetics.json
@@ -132,6 +132,18 @@
       "shop": "cosmetics"
     }
   },
+  "shop/cosmetics|Ulta Beauty": {
+    "countryCodes": ["us"],
+    "matchNames": ["ulta"],
+    "nomatch": ["shop/beauty|Ulta Beauty"],
+    "tags": {
+      "brand": "Ulta Beauty",
+      "brand:wikidata": "Q7880076",
+      "brand:wikipedia": "en:Ulta Beauty",
+      "name": "Ulta Beauty",
+      "shop": "cosmetics"
+    }
+  },
   "shop/cosmetics|Vente de Cosmétique": {
     "countryCodes": ["ci"],
     "tags": {


matkoniecz:
> This index considers a drugstore to be separate from the pharmacy counter within it

Note that it is done solely for USA ones where such distinction may make sense (due to law encouraging to expand pharmacy into a large shop with everything else). It is certainly not done for ones where such distinction would make no sense - for example all pharmacies in Poland.

matkoniecz:
> so by analogy the cosmetic store should be separate from the salon studios within it

I am not convinced that this is a desirable tagging, sounds like violation of https://wiki.openstreetmap.org/wiki/One_feature,_one_OSM_element to me.

I would recommend tagging primary function - is it salon studio where one may buy some of products? Or is it tiny salon studio with some basic services within a cosmetic store?

I would not tag large freezer with ice cream in supermarket as `amenity=ice_cream` and it sounds a bit like doing this.

tas50:
This seems like a wrong direction to me. The drugstore vs. pharmacy tagging in OSM is a complete mess and no one gets it right. This seems like just another version of that, and without a doubt no one will get it right. We'll just have a mix of cosmetics or beauty shops depending on which mapper touched the store. It's totally a cosmetics store first, but the beauty shop tag does cover the situation where you can buy supplies there and that seems to get us a lot closer than 2 tags, especially if those tags are applied incorrectly.

matkoniecz:
I would tag it as a `shop=cosmetics` with `beauty` tag indicating that services are also available. Is there something wrong with such solution?

tas50:
@matkoniecz is that a recognized solution to combine tags that way?

matkoniecz:
Not sure, but it seems to be logical and preferable over ignoring "One feature, one OSM element".

I can ask on tagging mailing list and document `beauty` tag (now a redirect, see https://wiki.openstreetmap.org/w/index.php?title=Key:beauty&redirect=no ).

1ec5:
> The drugstore vs. pharmacy tagging in OSM is a complete mess and no one gets it right.

I agree. The only reason I hesitated to remove the existing `shop=beauty` entry was because (without really having looked into it) it seemed like the salons had been added to the index before the cosmetics store, rather than simply being mistagged.

> I would recommend tagging primary function - is it salon studio where one may buy some of products? Or is it tiny salon studio with some basic services within a cosmetic store? In case that both functions are well developed - which is the main one?

The salons are just a store amenity; the primary function is as a cosmetics store.

> We'll just have a mix of cosmetics or beauty shops depending on which mapper touched the store. It's totally a cosmetics store first, but the beauty shop tag does cover the situation where you can buy supplies there and that seems to get us a lot closer than 2 tags, especially if those tags are applied incorrectly.

The salons aren’t actually called “Ulta Beauty” per se. It’s “The Salon at Ulta Beauty” (or informally “Ulta Salon”). The skin bar and brow bar have their own names too. So I’ll just rename the `shop=beauty` preset to prevent mappers from mistakenly using it to represent the whole store.

1ec5:
As of c8fc622fb203b25e3108041ee3318391adc5ffa1, there are separate entries for the cosmetic store, salon, skin bar, and brow bar (under a separate brand). It may seem like the height of exuberance to tag each individual department of this cosmetic store, but Ulta stores are often big box stores the size of supermarkets, and `shop=cosmetics` doesn’t really capture the range of services on its own. So if a mapper really wants to micromap the store, this set of presets is unambiguously labeled for doing so. Otherwise, they’ll map only the cosmetics store. In my opinion, that’s OK and not a theoretical public health issue like people mapping drugstores without mapping the pharmacy counters inside.

Adamant36:
Personally, I don't think it's the height of exuberance to tag each individual department in this case, because they seem to have separate brand names and apparently different social media accounts in some places. Plus, likely their own numbers for scheduling appointments. So, it's not the same case as a random cosmetic counter in a department store IMHO.

brands/shop/beauty.json
```diff
@@ -46,15 +62,16 @@
       "shop": "beauty"
     }
   },
-  "shop/beauty|Ulta Beauty": {
+  "shop/beauty|The Skin Bar at Ulta Beauty": {
     "countryCodes": ["us"],
-    "matchNames": ["ulta"],
+    "matchNames": ["ulta", "ulta skin bar"],
```
tas50: the ulta match probably shouldn't be there with this so we don't match on the whole stores which are usually tagged just ulta

bhousel: I removed the extra matchNames in 71809d11

PR #2749, opened 2019-06-04
Added BEA 東亞銀行 bank

zbc:


Adamant36:
Hi there. It would be good to add brand:zh and name:zh.

bhousel:
or if it is in classical, `brand:lzh` and `name:lzh`  (I don't know, I can't read it)
[details on Chinese multilingual names](https://wiki.openstreetmap.org/wiki/Multilingual_names#China)

zbc:
Because of 東亞銀行 is Hong Kong-based bank, so it is written by traditional Chinese. 东亚银行 is the simplified Chinese for it.  `brand:lzh` is for the old ancient Chinese I think. How about I add simplified Chinese it to `brand:zh` and `name:zh`, since `zh` usually stands for simplified Chinese?

zbc:
Or I can add `name:zh-Hant` and `name:zh-Hans` to differentiate traditional or/and simplified Chinese if you prefer?

bhousel:
> Or I can add `name:zh-Hant` and `name:zh-Hans` to differentiate traditional or/and simplified Chinese if you prefer?

I don't know.. Just try to do whatever it says to do on the thing I linked 👍

zbc:
My prettier messed up the format before, so I just fixed that issue too. Check it out. Thanks.

bhousel:
Thanks @zbc looks good 👍

PR #2751, opened 2019-06-05
add two fitness centre's

andrewharvey:


Adamant36:
Thanks. Looks fine.

Adamant36:
Quick side note, it would be worth adding social media accounts to Wikidata entries in the future when you create them so iD Editor can use the brands icon in searches. Usually the company links to them on their main page. 

Thanks

Issue #2728, opened 2019-05-31
Create preset for Kilwins


HubMiner:
brand=Kilwins
brand:wikidata=Q38250832
brand:wikipedia=en:Kilwins
name=Kilwins

# optional 
alt_name=Kilwin's
cuisine=chocolate;ice_cream;popcorn

HubMiner:
Add:
 "shop": "confectionery" or "amenity": "ice_cream"

zbc:
Can I work on this?

bhousel:
> Can I work on this?

Yes, go ahead! 👍

bhousel:
done in #2754

PR #2754, opened 2019-06-05
Added brand for Kilwins

zbc:


brands/shop/confectionery.json
```diff
@@ -39,6 +39,19 @@
       "shop": "confectionery"
     }
   },
+  "shop/confectionery|Kilwins": {
+    "countryCodes": ["us"],
+    "matchTags": ["amenity/ice_cream"],
+    "tags": {
+      "alt_name": "Kilwin's",
```
bhousel: We currently can't support having the alternate names very similar to the main names (different only in spaces, case, punctuation), so I am going to remove the `alt_name`, but other than that the PR looks ok 👍 

There haven't been any other brands where this been an issue, but I could probably change the matching code to support it if it matters a lot to people.  We don't use `alt_name` much so it hasn't been a problem yet.

brands/shop/confectionery.json
```diff
@@ -39,6 +39,19 @@
...
+      "brand": "Kilwins",
+      "cuisine": "chocolate;ice_cream;popcorn",
+      "name": "Kilwins",
+      "shop": "confectionery",
+      "wikidata": "Q38250832",
```
bhousel: oh!  also these need to be `brand:wikidata` and `brand:wikipedia`.. I'll fix that too.

brands/shop/confectionery.json
zbc: Got it.

brands/shop/confectionery.json
zbc: Sorry about that. Thanks.

bhousel:
Oh one other thing - if you add a text like "closes #2728" to the commit message, it will automatically close the issue once the commit is merged..  Pretty convenient!

https://help.github.com/en/articles/closing-issues-using-keywords

zbc:
> Oh one other thing - if you add a text like "closes #2728" to the commit message, it will automatically close the issue once the commit is merged.. Pretty convenient!
> 
> https://help.github.com/en/articles/closing-issues-using-keywords

Wow!! Good to know. Thanks!! Will do next time.

PR #2757, opened 2019-06-06
add St George Bank and Muji department store

andrewharvey:
Muji is also in other countries, but there should be a different entry for their other stores because name might be in a different default language, and I don't have the local knowledge outside australia.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -4489,6 +4489,16 @@
       "name": "Sparda-Bank"
     }
   },
+  "amenity/bank|St George": {
+    "countryCodes": ["au"],
+    "tags": {
+      "amenity": "bank",
+      "brand": "St George",
+      "brand:wikidata": "Q1606050",
+      "brand:wikipedia": "en:St.George Bank",
+      "name": "St George"
+    }
+  },
   "amenity/bank|Stanbic Bank": {
     "tags": {
       "amenity": "bank",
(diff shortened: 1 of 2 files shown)


Adamant36:
I'll merge this one. Since it's only mapped in Australia at the moment. So the country code thing shouldn't be a problem at the moment. Can you add a match for St.George Bank first though? It looks like it has a few uses.

Adamant36:
Btw, Muji is located in like 20 countries. So it might be worth removing the country code on that one so the places where it's at don't get filtered out of searches.

andrewharvey:
### St George

I've actually decided to change this to the full "St George Bank" and updated the matchNames to include a few variants.

## Muji.

My reasoning for only matching to au is because I've set name in English. I'm not sure if in other countries, the name should be in the local language, so I'd rather play it safe.

Someone with more local knowledge of other countries can then include there country when they verify that the main name is in English.

Adamant36:
> I've actually decided to change this to the full "St George Bank" and updated the matchNames to include a few variants.

How come you did that? It just St George on the signs and usually that's what's best to go with.

andrewharvey:
> How come you did that? It just St George on the signs and usually that's what's best to go with.

St George has 32 uses at the moment
St George Bank has 52 uses at the moment

To be honest I'm fine with either.

bhousel:
I cleaned up these and merged them

b877596e 2019-06-06 Remove countryCode from Muji, as they operate in many countries
99fceee6 2019-06-06 Fix brand:wikipedia for Australian Red Cross
265aa364 2019-06-06 Update "st.george", as this is how the brand calls themself

PR #2758, opened 2019-06-06
add australian red cross blood service

andrewharvey:


Adamant36:
Going by their signs etc it might be better to go with name=Blood service, brand=blood service, and operator=Australian Red Cross. Since it seems like the "Australian Red Cross" is more the operator and tiny on their signs. Plus, it's always good to go with the shorter name when possible. Since it shows up on a map better. 

You could also add short_name=ARCBS or even go with name=ARCBS + official_name=blood service (or official_name=Australian Red Cross Blood Service). Personally, I think name=Blood Service is best, but I'd like to see what others think.

andrewharvey:
Their sinage at the one near me looks like the logo on https://mytransfusion.com.au/about-us.

Although Blood Service is in larger text, I still feel we should include the full name "Australian Red Cross Blood Service" as the name tag value.

I would not abbreviate without that abbreviation being in common use, I've never heard or seen it abbreviated in that way.

I think `Blood Service` is an okay short_name.

I've updated the tags to better reflect what I think the brand and operator tags should be.

bhousel:
Seems ok, thanks @andrewharvey 👍

PR #2760, opened 2019-06-06
muji 無印良品 in jp, tw

andrewharvey:
follow up to #2757 @bhousel @Adamant36 

sorry I was in the middle of updating my PR when it was merged.

I've tried to follow the international example used at https://github.com/osmlab/name-suggestion-index/blob/b877596e309f95b35dcf1027dbac4defc75ebe73/CONTRIBUTING.md#card_file_box--about-the-brand-files

In Japan and Taiwan, possibly Singapore and other asian countries, the main name is known in the native language not English. This PR tried to address that to these stores in JP and TW are hopefully given better tags.

### DIFF
--- a/brands/shop/department_store.json
+++ b/brands/shop/department_store.json
@@ -334,10 +334,27 @@
   "shop/department_store|Muji": {
     "tags": {
       "brand": "Muji",
+      "brand:en": "Muji",
+      "brand:jp": "無印良品",
       "brand:wikidata": "Q708789",
       "brand:wikipedia": "en:Muji",
       "name": "Muji",
-      "name:zh": "無印良品",
+      "name:en": "Muji",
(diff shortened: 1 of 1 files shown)


brands/shop/department_store.json
```diff
@@ -334,10 +334,27 @@
   "shop/department_store|Muji": {
     "tags": {
       "brand": "Muji",
+      "brand:en": "Muji",
+      "brand:jp": "無印良品",
```
1ec5: FYI, `ja` is the ISO&nbsp;639 language code for Japanese. Fixed in 6a5cd967c138e255a40e44562b811eec321e1c07.

andrewharvey: Thanks for picking up on that!

PR #2766, opened 2019-06-07
Remove wrong Dekra declaration

bmillemathias:
... and correct the right one

### DIFF
--- a/brands/amenity/vehicle_inspection.json
+++ b/brands/amenity/vehicle_inspection.json
@@ -18,12 +18,12 @@
     }
   },
   "amenity/vehicle_inspection|Dekra": {
-    "countryCodes": ["fr"],
+    "countryCodes": ["de", "fr"],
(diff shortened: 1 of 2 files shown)


Adamant36:
Seems they have both vehicle inspection places and car repair shops. So it might be worth worth keeping both entries somehow. I'm not going to speculate on the best way to do it though.

matkoniecz:
> Seems they have both vehicle inspection places and car repair shops.

Separate ones? Or is it a single object that fulfills both functions?

bhousel:
AFAICT from their wikipedia page, they just do vehicle inspection services, so this seems ok.

Adamant36:
On their website it says they do oil changes and sell basic items like windshield wipers and light bulbs. I don't know if that would count as "car repair" per say. I guess they do more car repair related stuff with their other brands Autocontrol and Norisko.

PR #2767, opened 2019-06-07
Add brand:wikidata to shop/hairdresser Jean Louis David

bmillemathias:


Adamant36:
Thanks. Btw, apparently you can also use "instance of" + "hairdresser" for this type of business.

Issue #2732, opened 2019-06-01
Official name tag should automatically become a matchname


tas50:
There's a lot of brands out there with short names we use in OSM and then very long official_names that are used often. We've added a good number of those official names in the data set, but they should automatically become matchNames for the brands.

Example:

Total Wine: official name of "Total Wine & More" would become the match name of "total wine more"

matkoniecz:
The question is whatever it should appear in https://raw.githubusercontent.com/osmlab/name-suggestion-index/master/dist/name-suggestions.json and other dist files or is it something that `npm run build` would add to human-editable files?

Adamant36:
> We've added a good number of those official names in the data set, but they should automatically **become** matchNames for the brands.

Hopefully they would stay as official_names in the index also. Instead of being removed. As that would prevent people from unnecessary name re-tagging while mapping or opening pointless PR's based on a perception that the name should be different. It's also useful to have the official name on the map if someone is looking for something and doesn't know the shortened name in my Opinion. As sometimes they can be quit different.

bhousel:
I just did this!

Now the matcher will automatically match on these tags too
    - `alt_name`
    - `official_name`
    - `short_name`

So we don't need to add extra `matchNames` for these anymore.  

These all remain on all the entries as tags, wherever people have added them, so that hasn't changed.

PR #2768, opened 2019-06-08
add daiso japan

andrewharvey:
daiso existing in Japanese for japan, this adds another entry for Australia.

### DIFF
--- a/brands/shop/variety_store.json
+++ b/brands/shop/variety_store.json
@@ -52,8 +52,7 @@
     }
   },
   "shop/variety_store|Daiso Japan": {
-    "countryCodes": ["au"],
-    "matchNames": ["Daiso"],
+    "countryCodes": ["au", "us"],
     "tags": {
       "brand": "Daiso",
       "brand:wikidata": "Q866991",
--- a/brands/shop/variety_store.json
+++ b/brands/shop/variety_store.json
@@ -51,6 +51,17 @@
       "shop": "variety_store"
     }
   },
+  "shop/variety_store|Daiso Japan": {
+    "countryCodes": ["au"],
+    "matchNames": ["Daiso"],
+    "tags": {
+      "brand": "Daiso",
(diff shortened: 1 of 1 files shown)


brands/shop/variety_store.json
```diff
@@ -51,6 +51,17 @@
       "shop": "variety_store"
     }
   },
+  "shop/variety_store|Daiso Japan": {
+    "countryCodes": ["au"],
+    "matchNames": ["Daiso"],
```
bhousel: You can remove the matchName here since it is the same as the `short_name`. 
(we automatically match `alt_name`, `short_name` and `official_name` tags, see #2732)

brands/shop/variety_store.json
```diff
@@ -51,6 +51,17 @@
       "shop": "variety_store"
     }
   },
+  "shop/variety_store|Daiso Japan": {
+    "countryCodes": ["au"],
```
matkoniecz: From quick check at https://www.openstreetmap.org/search?query=Daiso+Japan#map=19/34.96736/136.61972 it seems to appear also in USA. Is it the same brand?

brands/shop/variety_store.json
matkoniecz: I think that this kind of magic may be included in a build script.

matkoniecz: done in https://github.com/osmlab/name-suggestion-index/pull/2772

brands/shop/variety_store.json
andrewharvey: I presume so, I've updated this.

For the record, I'm just adding what I know from local knowledge which is AU only, nothing stopping others following up later to expand the country codes with their local knowledge.

brands/shop/variety_store.json
andrewharvey: Thanks I didn't realise short_name was included.

PR #2771, opened 2019-06-08
require clothes tag for shop=clothes

matkoniecz:
I think that it would be desirable to expect `clothes` tag for every `shop=clothes`.

Making PR as it would add many new warnings - see https://gist.github.com/matkoniecz/79cdaa913b71a49fe564890db3bd43be

### DIFF
--- a/build_brands.js
+++ b/build_brands.js
@@ -273,6 +273,9 @@ function checkBrands() {
             case 'shop/beauty':
                 if (!obj.tags.beauty) { warnMissingTag.push([kvnd, 'beauty']); }
                 break;
+            case 'shop/clothes':
+                if (!obj.tags.clothes) { warnMissingTag.push([kvnd, 'clothes']); }
+                break;
         }
     });
 


Adamant36:
"I think that it would be desirable to expect clothes tag for every shop=clothes." 

What about places that sell a bunch of different types of clothes? I think there's a point where adding to many "type" keys is useless. Most clothing stores aren't specific. The tag should only used on speciality stores and adding a requirement/check would incourage people to use the tag as a way to doccument clothing inventory and not type of store in my opinion. Its already been leaning that way with the cuisine tag and individual menu items getting their own cuisine tag when they really shouldn't. Neither tag should be used to list inventory.

A lot of stores could be tagged as clothes=children because they sell childrens cloths but that doesn't mean they should be because they aren't children's clothing stores. Adding a check might lead to that type of tagging though.

matkoniecz:
> What about places that sell a bunch of different types of clothes?

https://wiki.openstreetmap.org/wiki/Key:clothes suggests `women;men;children`

> The tag should only used on speciality stores

If you think that it is not desirable to tag "this shop sells all types of clothes", then please - give your view at https://github.com/westnordost/StreetComplete/issues/1419

Adamant36:
> https://wiki.openstreetmap.org/wiki/Key:clothes suggests women;men;children

It doesn't suggest that specifically. It just says several types can be used to specify what type of store it is. In the case of your example it's not really "specifying" anything anymore. As that's essentially every type of cloths. There should really be something like a "specialty_store=no" or similar tag instead. 

> If you think that it is not desirable to tag "this shop sells all types of clothes", then please - give your view at westnordost/StreetComplete#1419

If I bring up something about an issue here, I should be able to discuss it here. I don't need to talk about it in another Repo for unrelated software. Especially when your the one that opened the issue there. I didn't say I had a problem with tagging places as "this shop sells all types of clothes" anyway. I said I have a problem with this particular way of doing it where your listing inventory instead of what type of store it is. 

Every clothing store might have a small area in the back with clothes that aren't their main specialty, especially seasonally, but I don't think it helps map users to know that. Because 1. it changes sporadically 2. It's not the point in the tag IMHO. Anymore then the cuisine tag is meant to list the type of cuisine of every item in a restaurant. 3. It doesn't help map users in any way. Not to mention it also hampers specific rendering based on these things. 

For instance, now it's much harder to render American Denny's as such, because it has the cuisine=Italian tag on it due to one item the menu, spaghetti, that's Italian. When people search for Italian restaurants they don't expect Denny's to come up and it's miss leading to treat it like it is. The same would go for this. The purpose of the clothing tag is to be able to find a specific clothing store on the map. Not to create an inventory list of every item sold at a store. Others use a different tag like sells:womens_cloths=yes or use something like the specialty_store=no tag that I suggested above. Otherwise, your just negating the purpose of the tag and diluting its usefulness.

bhousel:
That really is a lot of warnings, and for a most of these I'm not even sure how I'd resolve them.  Some places just sell all kinds of clothes.. Others really should be `shop=boutique`, or `shop=fashion` but those tags are discouraged lately.  I dunno. 😕

bhousel:
Closing for now.. I thought about this more and this is just too many warnings with no clear guidance on using the `clothes=*` tag to resolve them.

PR #2772, opened 2019-06-08
remove unneded name matches in npm run build

matkoniecz:
see https://github.com/osmlab/name-suggestion-index/issues/2732

### DIFF
--- a/build_brands.js
+++ b/build_brands.js
@@ -147,6 +147,19 @@ function mergeBrands() {
             if (!obj.tags.healthcare) obj.tags.healthcare = 'pharmacy';
         }
 
+        // remove unneded names, matched automatically - see #2732
+        if (obj.matchNames !== undefined) {
+            if (obj.tags.alt_name) {
+                obj.matchNames = obj.matchNames.filter(item => item !== obj.tags.alt_name)
+            }
+            if (obj.tags.official_name) {
+                obj.matchNames = obj.matchNames.filter(item => item !== obj.tags.official_name)
+            }
+            if (obj.tags.short_name) {
+                obj.matchNames = obj.matchNames.filter(item => item !== obj.tags.short_name)
+            }
+        }
+
         // Force `countryCode`, and duplicate `name:xx` and `brand:xx` tags
         // if the name can only be reasonably read in one country.
         // https://www.regular-expressions.info/unicode.html


matkoniecz:
After implementing and running it as a test I have two doubts

- it is violation of DRY to list `official_name`, `alt_name`, `short_name` in two places.

- I think that it may be confusing for user to see entries disappearing. Maybe it would be better to simply warn user and complain?

bhousel:
You are kidding right?

bhousel:
Oh sorry, read this too quickly.. I thought you were removing the matching code completely..

bhousel:
> Maybe it would be better to simply warn user and complain?

We already do this I think?

matkoniecz:
> We already do this I think?

Technically yes, but in a really cryptic way:

```
  "amenity/bank|ABN AMRO": {
    "countryCodes": ["nl"],
    "matchNames": ["ABN AMRO Bank N.V."],
    "tags": {
      "amenity": "bank",
      "brand": "ABN AMRO",
      "brand:wikidata": "Q287471",
      "brand:wikipedia": "en:ABN AMRO",
      "name": "ABN AMRO",
      "official_name": "ABN AMRO Bank N.V.",
      "operator": "ABN AMRO Group",
      "operator:wikidata": "Q17061287",
      "operator:wikipedia": "en:ABN AMRO Group"
    }
  },
```

generates

```
Warning - Brands match other brands:
--------------------------------------------------------------------------------
If the brands are the different, add a disambiguator after the name, like `~(USA)` vs `~(Canada)`
If the brands are the same, remove extra `matchTags` or `matchNames`.  Remember:
- Name matching ignores letter case, punctuation, spacing, and diacritical marks (é vs e). 
  No need to add `matchNames` for variations in these.
- Tag matching automatically includes other similar tags in the same match group.
  No need to add `matchTags` for similar tags.  see `config/match_groups.json`
--------------------------------------------------------------------------------
  "amenity/bank|ABN AMRO" -> matches? -> "amenity/bank|ABN AMRO ("abnamrobanknv")"
total 1
```

that is not clear at all. Anyway, it would be probably better to improve `Warning - Brands match other brands` for case where it matches itself.

bhousel:
> that is not clear at all. Anyway, it would be probably better to improve `Warning - Brands match other brands` for case where it matches itself.

I agree we should improve the warning..

PR #2773, opened 2019-06-09
move short 'browse the index' higher

matkoniecz:
"How it's used" section is quite long and may hide browsable index
that is a good call to action for a typical user browsing and
reduce chance that it will be missed

posted as PR to make it prominent and give opportunity for comments, I plan on merging it in the near future


Adamant36:
sounds like a sensible move to me.

PR #2774, opened 2019-06-09
Add stronger link between readable .html list file and real data files, be more explicit what is welcomed

matkoniecz:
Small chance that is intended to increase chance that someone who looks through this files will contribute.

I plan on merging it by myself in near future (assuming that there is nothing wrong with this change).


Adamant36:
> Is a brand name missing or something is incorrect?

Might sound better written as "Is a brand name missing or is something else incorrect?"

matkoniecz:
> "Is a brand name missing or is something else incorrect?"

I would not describe missing brand name as "incorrect"

Adamant36:
> I would not describe missing brand name as "incorrect"

True

PR #2776, opened 2019-06-10
AAA

1ec5:
Added American Automobile Association travel agencies and insurance agencies. I didn’t add the AAA car care centers, because each AAA affiliate runs its own small chain of owned-and-operated centers (not to be confused with AAA-approved mechanics), each with a similar name but a distinct `operator:wikidata`. Until we start filtering brands by state, the risk of a mapper getting confused and choosing the wrong chain is too high.


bhousel:
Thanks @1ec5 👍

PR #2777, opened 2019-06-10
home timber and hardware australia

andrewharvey:
see also https://github.com/openstreetmap/iD/issues/6513 where iD suggested the wrong brand tags

### DIFF
--- a/brands/shop/doityourself.json
+++ b/brands/shop/doityourself.json
@@ -225,6 +225,19 @@
       "shop": "doityourself"
     }
   },
+  "shop/doityourself|Home Timber & Hardware": {
+    "countryCodes": ["au"],
+    "matchNames": ["home hardware"],
+    "tags": {
+      "brand": "Home Timber & Hardware",
+      "brand:wikidata": "Q17006984",
+      "brand:wikipedia": "en:Home Timber & Hardware",
+      "name": "Home Timber & Hardware",
+      "alt_name": "Home Hardware",
+      "short_name": "Home",
+      "shop": "doityourself"
+    }
+  },
   "shop/doityourself|Homebase": {
     "countryCodes": ["gb", "ie"],
     "tags": {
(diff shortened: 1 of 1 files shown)


Adamant36:
You don't need a match for "home hardware" if it's already used as the alt name. The name index accounts for capitalizing and searches alt name for matches now.

andrewharvey:
> You don't need a match for "home hardware" if it's already used as the alt name. The name index accounts for capitalizing and searches alt name for matches now.

Sorry you're right.

> I'm kind of wondering about the usage of "home" to. It seems a little to general to me. Even if the business goes by that somewhere. There's a bunch of places named "home" all over the world.

According to Wikipedia "Home" is a common name, and based on my experience Home is a valid short name.

_Home Timber & Hardware (commonly known as "Home" or "Home Hardware")_

I'm wondering if it matter if there are other places also named Home? Upstream consumers of name-suggestion-index know that this should only match in `au`, and a user should still review any tag suggestions to make sure they are correct to add to the feature.

matkoniecz:
Looking at http://overpass-turbo.eu/s/JPR and its Wikipedia page it seems that `shop=hardware` may also match.

Is `shop=doityourself` better for all its locations?

bhousel:
Looks good thanks @andrewharvey 👍

andrewharvey:
> Looking at http://overpass-turbo.eu/s/JPR and its Wikipedia page it seems that shop=hardware may also match.
> 
> Is shop=doityourself better for all its locations?

I did review the wiki page for shop=hardware and shop=doityourself. The distinction is blury, but from my understanding hardware was for small shops selling screws, nails etc. whereas doityourself was for larger shops selling a wider range of products, for the kitchen, bathroom, etc.

Adamant36:
I agree the difference is blurry. In the description section for shop=hardware it says its for a "shop which sells timber, tools and other building products." That sounds more like a diy store to me. The specific mention of timber is what makes me thing this should be tagged as a hardware store. Either that or the related wiki pages should be updated to be more clear about what exactly the tags apply to.

PR #2778, opened 2019-06-10
add mitre 10 australia

andrewharvey:


Adamant36:
Hi there. Two things, can you add the country for New Zealand since they are also located there and can also add a separate entry for Mitre 10 Mega since they are different stores and it would be good if they had  unique entries? Meitre 10 Mega seems to only be located in New Zealand. 

Thanks

andrewharvey:
The NZ shops would have different wikipedia/wikidata tags https://en.wikipedia.org/wiki/Mitre_10_(New_Zealand), so I think nz should be a separate brand entity in the index.

I'd rather leave that to someone else who has more local knowledge.

Adamant36:
> The NZ shops would have different wikipedia/wikidata tags 

Hmmm, well I guess that's on them for creating a different company with the same name etc. You don't necessarily need local knowledge to add an entry to the index and it's generally better to have it then not. Especially in this case where there's the almost same brand somewhere else. Although I might be wrong, I don't think iD Editor filters based on country code yet. So without the extra entry there might be  a lot of false positives until it does. It's your choose though. I can always just add it myself after this is merged.

Adamant36:
Thanks 👍

Adamant36:
You know, are you sure they aren't the same company or something? Both Wikipedia pages mention that they also have "Mitre 10 Mega" stores. I highly doubt two different companies would have the same name/format for two different types of stores.

andrewharvey:
> Are you sure they aren't the same company? Both Wikipedia pages mention that both have "Mitre 10 Mega" stores. I highly doubt two different companies would have the same name/format for two different types of stores. More then likely it's just the same company with a different logo for the New Zealand stores or something.

To be honest I'm not sure, I haven't seen any Mega stores, and I don't have any knowledge of the NZ stores, so I'll leave that to someone else with more local knowledge to address.

PR #2779, opened 2019-06-10
Add PA Wine & Spirits

jleedev:


bhousel:
Thanks @jleedev 👍 
I also added `name` and `operator:wikipedia` tags in 963c93a9 and b48d94eb

Issue #2678, opened 2019-05-27
MOD Pizza is fast food not restaurant


roadgeek99:
I've noticed MOD Pizza is in the `amenity=restaurant` category. It should be in the `amenity=fast_food` category because it's basically like the Subway (the sandwich shop) for pizzas. Customers build their own pizzas, pay, wait for it to be baked, and get the pizza themselves. It is not a sit-down restaurant at all. Not sure why people think it's `restaurant`, but it really should be `fast_food`.

tas50:
Fixed by fcf9969ea0025eee4286285df2283732e7c8f2f5

Adamant36:
>  It is not a sit-down restaurant at all.

I'm not sure how that relates. What makes something a restaurant is if an employee brings the pizza to your table for you when it's done or not. It seem's like in this case that they do. Just because the pizza is made by pointing at fresh ingredients, like Subway, versus picking them from a menu doesn't really matter. The speed with which the food is made doesn't matter either. Subway is a fast food place only because you take the sandwich at the counter when your done having it made and the employees don't serve you. That's not the case here. In MOD pizza places you order, sit down, and they bring it to you when it's done. Really, it's no different than any other pizza restaurant. It's just made quicker. 

There was a discussion about this a few days ago, but I can't remember where.

tas50:
One of the big measures was where you order. A lot of fast food places bring the food to you, but there’s no full table service. If you wait in a line and order pizza like a burger at McDonalds but they bring it to you that seems pretty fast food to me. Fast casual blurs the line, but we have a lot of restaurants in the index that should be fast food locations.

PR #2782, opened 2019-06-11
Add Zoës Kitchen

tas50:
I'm opening a PR so we can discuss if this should be a restaurant or
fast_food. I've never been to one, but it seems like a fast food place.

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -2160,6 +2160,18 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|Zoës Kitchen": {
+    "tags": {
+      "alt_name": "Zoes Kitchen",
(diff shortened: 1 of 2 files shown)


Adamant36:
In the meantime, how is it cuisine=juice?

bhousel:
Looks like fast-casual from the pictures on their website and from the [wikipedia page](https://en.wikipedia.org/wiki/Zo%C3%ABs_Kitchen).

So similar to #2678 and #2622, OSM convention is to tag these as `fast_food`.

"If people wait on you it's `amenity=restaurant`... If they don't it's `amenity=fast_food`"

(I agree that `cuisine=juice` looks like a mistake)

bhousel:
Thanks, merged and switched cuisine from `juice` to `mediterranean` in 3f2707c1

PR #2783, opened 2019-06-11
Major credit unions of Canada

1ec5:
Added Canadian credit unions with 50 or more branches.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -3177,8 +3177,8 @@
   },
   "amenity/bank|First West Credit Union": {
     "countryCodes": ["ca"],
-    "matchNames": ["first west"],
     "tags": {
+      "alt_name": "First West",
       "amenity": "bank",
       "brand": "First West Credit Union",
       "brand:wikidata": "Q5454073",
@@ -3835,8 +3835,8 @@
   },
   "amenity/bank|Meridian Credit Union": {
     "countryCodes": ["ca"],
-    "matchNames": ["meridian"],
     "tags": {
+      "alt_name": "Meridian",
       "amenity": "bank",
       "brand": "Meridian Credit Union",
       "brand:wikidata": "Q6819303",
@@ -4520,8 +4520,8 @@
   },
   "amenity/bank|Servus Credit Union": {
     "countryCodes": ["ca"],
-    "matchNames": ["servus"],
     "tags": {
+      "alt_name": "Servus",
       "amenity": "bank",
       "brand": "Servus Credit Union",
       "brand:wikidata": "Q2274396",
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -119,6 +119,17 @@
       "name": "Addiko Bank"
     }
   },
+  "amenity/bank|Affinity Credit Union": {
+    "countryCodes": ["ca"],
+    "matchNames": ["affinity"],
+    "tags": {
+      "amenity": "bank",
+      "brand": "Affinity Credit Union",
+      "brand:wikidata": "Q4688969",
+      "brand:wikipedia": "en:Affinity Credit Union",
+      "name": "Affinity Credit Union"
+    }
+  },
   "amenity/bank|Agrani Bank Limited অগ্রণী ব্যাংক লিমিটেড": {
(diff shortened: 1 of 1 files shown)


brands/amenity/bank.json
```diff
@@ -2547,6 +2558,21 @@
       "name": "Clydesdale Bank"
     }
   },
+  "amenity/bank|Coast Capital Savings": {
+    "countryCodes": ["ca"],
+    "matchNames": [
+      "coast capital",
```
matkoniecz: is "Coast Capital" a good fit for `short_name` tag?

brands/amenity/bank.json
```diff
@@ -3149,6 +3175,17 @@
       "name": "First State Bank"
     }
   },
+  "amenity/bank|First West Credit Union": {
+    "countryCodes": ["ca"],
+    "matchNames": ["first west"],
```
matkoniecz: Is it a good fit for `short_name`?

brands/amenity/bank.json
```diff
@@ -3796,6 +3833,17 @@
       "name": "Mercantil"
     }
   },
+  "amenity/bank|Meridian Credit Union": {
+    "countryCodes": ["ca"],
+    "matchNames": ["meridian"],
```
matkoniecz: Is it a good fit for `short_name`?

brands/amenity/bank.json
```diff
@@ -4470,6 +4518,17 @@
       "name": "Security Bank"
     }
   },
+  "amenity/bank|Servus Credit Union": {
+    "countryCodes": ["ca"],
+    "matchNames": ["servus"],
```
matkoniecz: Is it a good fit for `short_name`?

brands/amenity/bank.json
```diff
@@ -4841,6 +4900,21 @@
       "name": "UCPB"
     }
   },
+  "amenity/bank|UNI": {
+    "countryCodes": ["ca"],
+    "matchNames": [
+      "caisse populaire",
+      "caisse populaire acadienne"
```
matkoniecz: Just sanity check - is this quite different name intended to match?

matkoniecz:
Looks nice, thanks for this PR!

brands/amenity/bank.json
1ec5: I put this match in because the logo kind of deemphasizes the word “savings”, but it looks like it’s always referred to as “Coast Capital Savings”. At most it could be an `alt_name`. (In my opinion, `short_name` is best suited for acronyms and the like, as opposed to shortenings of unwieldy phrases into slightly less wieldy phrases.)

brands/amenity/bank.json
1ec5: `alt_name`, perhaps.

brands/amenity/bank.json
1ec5: This should be `alt_name`, because the logo on signage says only “Meridian”, whereas it’s always “Meridian Credit Union” in plain-text contexts.

brands/amenity/bank.json
1ec5: This can be an `alt_name`; it appears this way on some branches’ signs.

brands/amenity/bank.json
1ec5: This was the credit union federation’s name until recently.

brands/amenity/bank.json
matkoniecz: Wiki at has

1ec5: I didn’t mean to say that `short_name` is _only_ for acronyms, but there’s a point at which a somewhat shortened name isn’t a “short name”. An extreme example would be that the `short_name` of “Doctor Martin Luther King Junior Park” should be `MLK Park`; `Doctor Martin Luther King Park;Martin Luther King Park` should go in `alt_name` rather than `short_name`. This case is not as extreme, so I don’t feel as strongly about it. In fact, I’d completely omit this name if not for the three POIs currently mistagged as `name=Coast Capital`.

By the way, the wiki mentions searching, but `short_name` is also used by some renderers in space-constrained situations.

PR #2784, opened 2019-06-11
Major U.S. credit unions

1ec5:
Added U.S. credit unions with 50 or more branches. (A bunch more credit unions just barely miss that threshold.)

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -245,6 +245,17 @@
       "operator:wikipedia": "en:Hellenic Financial Stability Fund"
     }
   },
+  "amenity/bank|America First Credit Union": {
+    "countryCodes": ["us"],
+    "tags": {
+      "amenity": "bank",
+      "brand": "America First Credit Union",
+      "brand:wikidata": "Q4742758",
+      "brand:wikipedia": "en:America First Credit Union",
+      "name": "America First Credit Union",
+      "short_name": "AFCU"
+    }
+  },
   "amenity/bank|Andhra Bank": {
     "countryCodes": ["in"],
     "tags": {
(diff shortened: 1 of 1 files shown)


brands/amenity/bank.json
```diff
@@ -4133,6 +4165,18 @@
       "name": "Pekao SA"
     }
   },
+  "amenity/bank|PenFed Credit Union": {
+    "countryCodes": ["us"],
+    "matchNames": ["penfed"],
```
matkoniecz: sounds like `short_name` to me.

brands/amenity/bank.json
```diff
@@ -4530,6 +4575,28 @@
...
+      "amenity": "bank",
+      "brand": "Security Service Federal Credit Union",
+      "brand:wikidata": "Q7444993",
+      "brand:wikipedia": "en:Security Service Federal Credit Union",
+      "name": "Security Service Federal Credit Union",
+      "short_name": "SSFCU"
+    }
+  },
+  "amenity/bank|Service Credit Union": {
+    "countryCodes": ["de", "us"],
```
tas50: Are you sure this is in both europe and the US?

mds08011: Surprising but per [Wikipedia](https://en.wikipedia.org/wiki/Service_Credit_Union) (emphasis mine):
>As of 2018, Service Credit Union has 50 locations (35 in New Hampshire, 1 in Massachusetts, 1 in North Dakota **and 13 on US military installations in Germany**)

tas50: That actually makes perfect sense. I'm sure USAA probably has a few branches in Germany for the same reason.

bhousel:
Looks great, thanks!

PR #2785, opened 2019-06-11
Add missing brand tags.

jleedev:
¶ cat $(git ls-files brands/) | jq -c '.[].tags|select(has("name")|not)'
    ¶ cat $(git ls-files brands/) | jq -c '.[].tags|select(has("brand")|not)'
    {"amenity":"restaurant","brand:wikidata":"Q5331211","brand:wikipedia":"en:Eat'n Park","cuisine":"american","name":"Eat'n Park"}
    {"brand:wikidata":"Q3331028","brand:wikipedia":"fr:Mutuelle d'assurance des artisans de France","name":"MAAF","office":"insurance"}
    {"brand:wikidata":"Q10521791","brand:wikipedia":"sv:Hemmakväll","name":"Hemmakväll","shop":"confectionery"}

### DIFF
--- a/brands/amenity/restaurant.json
+++ b/brands/amenity/restaurant.json
@@ -512,6 +512,7 @@
     "countryCodes": ["us"],
     "tags": {
       "amenity": "restaurant",
+      "brand": "Eat'n Park",
       "brand:wikidata": "Q5331211",
       "brand:wikipedia": "en:Eat'n Park",
       "cuisine": "american",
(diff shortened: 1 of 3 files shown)


bhousel:
oh wow thanks!   This really needs to be a build warning..

bhousel:
I've updated the schema in 10fc1672 so that it will fail validation if `name` or `brand` tags are missing.

PR #2788, opened 2019-06-12
Docs for each country - fixes #2733 [WIP]

### LINKED ISSUE #2733: nsi docs - include ability to limit results to a specific country code
To make data review easier it would be useful to add ability to show all entries with, for example, country code `au` and entries without `countryCodes` field.

Filtering can be either dynamic with some JS or static version of http://osmlab.github.io/name-suggestion-index/brands/ can be generated for every country

matkoniecz:
Allowing to list entries from just a specific country should make much easier for local mappers to look over our list and spot mistakes or serious omissions.

But during and after implementing it I started to consider migrating docs to a separate repository, 36k files is a bit of overkill.

And the same time "single page for each country" was not working anyway, with browser managing to lag on table with currently listed US brands - that was my initial implementation.

Limiting it to "serious" countries or ones where nsi has some special handling may help, but I would prefer to avoid marginalizing potential mappers from already underrepresented countries.

Assuming that so many html files in this repo will not kill it - I still need to review code and retest navigation, I fixed worst things but something buggy is certainly still lurking.


matkoniecz:
I plan on testing filtering out small countries - for example, with population below one million, but except ones with some countryCode rule applying to them.

Also, I will switch to using OSM data instead of introduced dependency (I need population data anyway...).

bhousel:
Yeah I think this is really too many files.. I'd prefer to rebuild the browser in client side code.
I'm sorry, but I can't accept this..

SilentSpike:
🤔 Would it not be easier to provide a means of dynamically... yeah that 👍

matkoniecz:
In case that somebody was interested in code of this design failure - it is at https://github.com/matkoniecz/name-suggestion-index/tree/docs_for_each_country

PR #2795, opened 2019-06-14
Add Longs Drugs

mds08011:
Longs is a pharmacy which was acquired by CVS in 2008.  Stores in the western continental US were rebranded as CVS but the approximately 60 stores in Hawaii are still branded as “Longs Drugs”.

Due to the CVS acquisition, Longs no longer has an independent Facebook or Twitter ~~so NSI will not be able to show a logo~~ but the Wikidata contains a logo.


brands/amenity/pharmacy.json
```diff
@@ -710,6 +710,18 @@
...
   },
+  "amenity/pharmacy|Longs Drugs~(Hawaii)": {
+    "countryCodes": ["us"],
+    "tags": {
+      "amenity": "pharmacy",
+      "brand": "Longs Drugs",
+      "brand:wikidata": "Q16931196",
+      "brand:wikipedia": "en:Longs Drugs",
+      "healthcare": "pharmacy",
+      "name": "Longs Drugs"
```
tas50: missing a comma here

tas50:
Thanks for the backstory in the PR. I miss Longs. The family was super nice.

brands/amenity/pharmacy.json
mds08011: Thanks!

brands/amenity/pharmacy.json
```diff
@@ -1810,4 +1822,4 @@
       "name": "屈臣氏"
     }
   }
-}
\ No newline at end of file
+}
```
tas50: if you run npm run build this should go away as well. We need an editor config file to prevent these getting added back in every time.

mds08011: I don't actually know javascript and have just been hacking these together in a browser.  I just tried to manually remove that bracket but that caused it to fail a check.  How should I proceed?

tas50:
It's the line ending vs. no line ending. If you're just doing it in the browser then don't worry about these. They'll just get added back when someone runs npm run build.  You'll want to revert that last commit though.

mds08011:
Gotcha.  Thanks for the help!

brands/amenity/pharmacy.json
```diff
@@ -710,6 +710,18 @@
       "name": "London Drugs"
     }
   },
+  "amenity/pharmacy|Longs Drugs~(Hawaii)": {
```
1ec5: Is the disambiguating suffix needed? Any Longs Drugs location outside of Hawaii would’ve changed its name by now, so there’s only one “Longs Drugs” chain to include in the index. Then again, 2fc39b458d5ec02f6f91443f892ee2909feee402 added a disambiguator to Noah’s Bagels even though there’s only one Noah’s Bagels chain, so I’m no longer certain about the criteria for disambiguators.

brands/amenity/pharmacy.json
```diff
@@ -710,6 +710,18 @@
...
+    "countryCodes": ["us"],
+    "tags": {
+      "amenity": "pharmacy",
+      "brand": "Longs Drugs",
+      "brand:wikidata": "Q16931196",
```
1ec5: > Due to the CVS acquisition, Longs no longer has an independent Facebook or Twitter so NSI will not be able to show a logo.

I added an SVG logo to the Wikidata item, so we’ll show that as the logo. The image was originally on the English Wikipedia under fair use, but a text-only wordmark is ineligible for copyright, so I moved it to Wikimedia Commons for use in Wikidata.

brands/amenity/pharmacy.json
mds08011: It’s probably overkill and I have no opposition to removing it.  I was being extra cautious because while the company was acquired by CVS in 2008, I don’t know when the rebranding was completed.  I was worried about the confusion around potential old POIs which have long since rebranded but were not updated in OSM (I haven’t checked if any actually exist).  My use of the disambiguator was meant as an extra precaution because we don’t have a US “stateCode” as we do “countryCode”.

brands/amenity/pharmacy.json
mds08011: Excellent!  I didn’t know that about text-only wordmarks and got scared off by the fair use and “non-free use rationale” language.

brands/amenity/pharmacy.json
tas50: I actually hand cleaned up all the ones left on the west coast a while back. They should all be CVS at this point.

mds08011: That works for me.  I agree with this reasoning and reverted the disambiguation.

brands/amenity/pharmacy.json
Adamant36: > I moved it to Wikimedia Commons for use in Wikidata.

@1ec5, is it generally OK for us add logos to Wikimedia Commons for usage on Wikidata items? I ran into a couple of situations where the logo was on a Wikipedia page, but I couldn't get it to load on Wikidata and was thinking about doing it that way, but I didn't want to cause problems. I guess they all have different copyright rules or something. I wasn't really clear about it.

ArloJamesBarnes: There isn't a general rule, it's a judgment call whether the logo is 'original' enough or not to qualify for copyright restriction. Most text logos are pretty simple. And Best Western's was decided by a judge to not be original enough, so there's a waypoint.

1ec5: Wikimedia Commons has stricter licensing rules than the English Wikipedia. The English Wikipedia allows some copyrighted images to be uploaded under a fair use rationale. This is very common for organizational logos, where Wikipedia has a justifiable reason for including them in articles. But sometimes Wikipedians assume all logos are copyrighted – in fact, logos that consist only of some text set in a particular font (even an obscure font) are ineligible for copyright protection and in the public domain, even as they enjoy trademark protection.

For incorrectly tagged public domain images, there are a few steps:

1. On the file description page, replace templates like `{{Non-free use rationale logo}}` and `{{Non-free logo}}` with [`{{Information}}`](https://en.wikipedia.org/wiki/Template:Information), [`{{PD-logo}}`](https://en.m.wikipedia.org/wiki/Template:PD-logo), and [`{{Trademark}}`](https://en.wikipedia.org/wiki/Template:Trademark).
2. Also add the [`{{Copy to Wikimedia Commons}}`](https://en.m.wikipedia.org/wiki/Template:Copy_to_Wikimedia_Commons) template to request a move to Commons then wait for someone to get around to it. Or if you’re sure it’s a public domain image, just do it yourself using the “Export to Wikimedia Commons” tab.
3. If you use that export tool, make sure the name is descriptive, like “Longs Drugs logo”.
4. Once the file is on Commons, go to the original file on the English Wikipedia and replace `{{Copy to Wikimedia Commons}}` with the [`{{Now Commons}}`](https://en.m.wikipedia.org/wiki/Template:Now_Commons) template, which will flag the original file for deletion. Now you can add the new file to the Wikidata item.

Follow these steps if the logo on Wikimedia Commons is the best logo there is for the brand. On the other hand, if Wikipedia has a grainy JPEG logo and a better PNG or SVG logo is available somewhere else – and it’s in the public domain – then upload the better image directly to Commons. Then you can replace the old logo with the new one in the Wikipedia article.

Only export or upload a logo to Commons if you’re sure it’s in the public domain. The Longs Drugs logo is a great example because it’s literally just some cursive text. A public domain logo can also have lines and basic shapes, such as squares and circles with solid fills. It can also contain elements that themselves are in the public domain, such as a standard American flag or an unstylized shape of Texas.

Much more complex logos have been uploaded to Commons as public domain works, but there’s a gray area, because to be in the public domain for Commons’ purposes, the logo must fall short of the [threshold of originality](https://commons.m.wikimedia.org/wiki/Commons:Threshold_of_originality) in the U.S. and the country where it was first used. The threshold of originality varies by country. For example, Chinese calligraphy is copyrightable in China but not the U.S. If in doubt, feel free to ask for a second opinion in Slack or apply that `{{Copy to Wikimedia Commons}}` template for someone else to make the call.

1ec5:
https://github.com/osmlab/name-suggestion-index/pull/2795#discussion_r294060312 got collapsed as part of a resolved discussion, but it has some guidance on how to deal with logos that are on the English Wikipedia but not Wikimedia Commons and thus unusable by Wikidata or this project.

brands/amenity/pharmacy.json
matkoniecz: > then wait for someone to get around to it

Note that https://en.m.wikipedia.org/wiki/Category:Copy_to_Wikimedia_Commons have many files.

"bot-assessed" subcategory alone has over 180 000 files.

bhousel:
What about this Longs Drugs in NC/SC?  https://longsrx.com

mds08011:
I re-added the "~Hawaii" disambiguating suffix.  The Longs Drugs brand @bhousel found has ~20 locations in the southeast and this distinction seems relevant to prevent confusion.

brands/amenity/pharmacy.json
mds08011: I unresolved this comment for better visibility but it may be worth adding this to or at linking from the CONTRIBUTING.md instructions.

brands/amenity/pharmacy.json
mds08011: Per the discovery of a "Longs Drugs" with ~20 locations in the Southeast US, I re-added the "~Hawaii" (comment linked below).  Even if there is currently only one Longs Drugs in the NSI, this will help to prevent confusion for mappers in SC, NC, and GA.

https://github.com/osmlab/name-suggestion-index/pull/2795#issuecomment-502680399

bhousel:
> The Longs Drugs brand @bhousel found has ~20 locations in the southeast and this distinction seems relevant to prevent confusion.

Thanks! 
Usually one of the first things I do is search for the brand on Facebook, and if another one is prevalent there, it is probably worth adding the disambiguation string.

PR #2797, opened 2019-06-15
Major boba tea chains

1ec5:
Added major international boba tea chains. Each chain originated in Taiwan with a Chinese name, so there are different ways to tag a given chain in each country that uses a CJKV language, in addition to the international tag set. I used `name:en` instead of `int_name` for Chatime and Quickly because those names seemed to be English-inspired names rather than ordinary transliterations.

### DIFF
--- a/brands/amenity/cafe.json
+++ b/brands/amenity/cafe.json
@@ -537,6 +537,21 @@
       "takeaway": "yes"
     }
   },
+  "amenity/cafe|Quickly": {
+    "tags": {
+      "amenity": "cafe",
+      "brand": "Quickly",
+      "brand:en": "Quickly",
+      "brand:wikidata": "Q3771463",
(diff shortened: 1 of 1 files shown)


brands/amenity/cafe.json
```diff
@@ -1005,6 +1099,87 @@
       "takeaway": "yes"
     }
   },
+  "amenity/cafe|貢茶 Gong Cha": {
+    "countryCodes": ["hk"],
```
1ec5: I just realized Nominatim considers Hong Kong and Macau to have a country code of `cn`, so an editor like iD would ignore this preset entirely. Yet the tagging would definitely differ between Hong Kong and mainland China due to the presence of English and also the differences between traditional and simplified Chinese. Would it be OK to put both entries under `cn`? Or should we begin distinguishing Hong Kong from China for brand purposes?

PR #2799, opened 2019-06-16
Avalon

1ec5:
Added Avalon, a large American chain of apartment complexes. Each complex carries the same Avalon logo (distinct from the AvalonBay corporate logo) and a name that begins with “Avalon”. AvalonBay also owns a few other apartment chains, but those chains are too small for the index.

### DIFF
--- a/brands/landuse/residential.json
+++ b/brands/landuse/residential.json
@@ -1,0 +1,15 @@
+{
+  "landuse/residential|Avalon": {
+    "countryCodes": ["us"],
+    "tags": {
+      "brand": "Avalon",
+      "brand:wikidata": "Q64665938",
+      "landuse": "residential",
+      "name": "Avalon",
+      "operator": "AvalonBay Communities",
+      "operator:wikidata": "Q4827537",
+      "operator:wikipedia": "en:AvalonBay Communities",
+      "residential": "apartments"
+    }
+  }
+}


Adamant36:
The ` "residential": "apartments"` and `"landuse": "residential"` tagging seems a little weird. Wouldn't a reality manager be better tagged as a reality office or something similar since it is where the daily business of managing the places takes place? Instead of at the apartment complexes which they usually aren't involved with in person most of the time (especially the residential area). Even in cases where there is an onsite manager, it would still be in an office that in my opinion should be tagged with their information instead of the landuse area.

1ec5:
`landuse=residential` `residential=apartments` is how an apartment complex is tagged; there’s a preset for it in iD. In the case of a luxury apartment chain such as Avalon, it’s the overall complex that’s branded, not just the management office. The logo appears prominently at the gates to the complex, the gates to the pool area, the gates to the fitness area, and on every apartment building.

[avalon-alameda](https://www.mapillary.com/map/im/Screa18hMrMXiefzyzNwyw)

Adamant36:
> it’s the overall complex that’s branded, not just the management office.

I don't know. It seems like a weird edge case. My inclination is to think map consumers would be looking for the management office and not necessarily the specific apartment complexes if they do a search for "Avalon." I could be completely wrong though. Maybe there should be an entry for both at least. It would still be useful to have an entry for their offices.

1ec5:
> My inclination is to think map consumers would be looking for the management office and not necessarily the specific apartment complexes if they do a search for "Avalon."

There’s already a well-established practice of naming landuse areas that represent coherent planned developments, including gated apartment complexes. Brand tags are just an extension of that for the more corporate apartment complexes.

Searching is only one use case for OSM. Another is browsing a rendered map; well-considered landuse area labels can make it easier to navigate the map without having to zoom in to where individual offices become visible. (Now I’m tempted to make a “Corporate America” map style that uses NSI to label the map exclusively with logos. 😏)

> Maybe there should be an entry for both at least. It would still be useful to have an entry for their offices.

I suppose we could add entries for Avalon’s leasing offices. After all, one could reasonably expect the leasing office to be given the same name as the complex it’s in. `office=apartments` is used a handful of times; not sure if there’s a better option for that.

Beyond that, we can’t expect mappers to spend their time applying the same redundant names and brands to every micromapped feature in a complex. That’s what spatial queries or site relations are for. Once a search engine or router becomes sophisticated enough to find the unnamed leasing office or clubhouse of an apartment complex, it isn’t much of a stretch to find the `entrance=main` of an arbitrary building – a pretty useful feature, in my opinion.

Adamant36:
> Searching is only one use case for OSM. Another is browsing a rendered map; well-considered landuse area labels can make it easier to navigate the map without having to zoom in to where individual offices become visible. (Now I’m tempted to make a “Corporate America” map style that uses NSI to label the map exclusively with logos. 😏)

True. You should totally do that. I had a similar idea awhile back. Except where everything would be faces of celebrities (or mainly Nicolas Cage's face). You got to have dreams.

> we can’t expect mappers to spend their time applying the same redundant names and brands to every micromapped feature in a complex.

True.

PR #2800, opened 2019-06-18
More Japanese Brands (part 5)

sguinetti:
In this pull request adds new coffee brands and update some info with Wikidata and alt_name.


bhousel:
Thanks @sguinetti 👍

PR #2801, opened 2019-06-20
HSBC UK

jnicho02:
This is deliberately a tiny change just to get things started. OSMUK would like to send through more details of presets for the UK after this one.

In the UK, HSBC banks are now branded "HSBC UK" and have that wording on their external signage. They would still have the main HSBC Brand and Wikidata:id applied. Please tell me if my config change is wrong or misguided.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -3353,6 +3353,16 @@
       "name": "HSBC"
     }
   },
+  "amenity/bank|HSBC UK": {
+    "countryCodes": ["gb"],
+    "tags": {
+      "amenity": "bank",
+      "brand": "HSBC",
+      "brand:wikidata": "Q190464",
+      "brand:wikipedia": "en:HSBC",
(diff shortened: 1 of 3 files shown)


jnicho02:
gah! Probably shouldn't have checked in changes from docs

Adamant36:
Hi. I think when you do it next time that you should delete the original HSBC entry and add it as a `"matchNames": ["HSBC"],` line to the new one. That 's my only suggestion. I think there might be something else to modify when you remove entries to, but I can't remember what it is right now 🤷‍♂

bhousel:
1. It looked pretty ok.. Just use Q5413890 for HSBC UK instead
2. You can't add `"matchNames": ["HSBC"]`, or list "HSBC" as any kind of alternate name while the global "HSBC" exists.  It would introduce a name conflict.
3. It's no big deal to change and check in doc files, but not really necessary, since they get regenerated pretty frequently.

Issue #2787, opened 2019-06-12
Consistent handling of formats


SilentSpike:
What is a format?
-

According to the OSM wiki this is one of a few "well established terms used in the British Retail Trade". I've been unable to find it defined elsewhere, but for our purposes the wiki definition is good:
>An example of different formats are Tesco Extra, Tesco Metro and Tesco Express. At the moment we do not tag individual formats.

How should we handle them?
-

At the moment we're not consistent in their handling. Personally I think different store formats should all share the same `brand` tag at least and potentially the same `brand:wikidata` tag too. However, we should come to some consensus either way and stick with it.

Previous Discussion
-
https://github.com/osmlab/name-suggestion-index/pull/2231#discussion_r244643936
https://github.com/osmlab/name-suggestion-index/issues/2740#issuecomment-498042862
https://github.com/osmlab/name-suggestion-index/issues/38#issuecomment-421456527
https://github.com/openstreetmap/iD/issues/6443#issuecomment-497713875

Adamant36:
Where exactly is that quote on the wiki? I didn't see it in the Supermarket article. 

As you see it, what's the harm in not coming to some consensus on it and what would be the benefit of coming to one? Also, why exactly do you think different store formats should all share the same brand/brand:wikidata tags? (I'm only asking to understand the question/problem better).

SilentSpike:
It's from here: https://wiki.openstreetmap.org/wiki/Retail_chains_in_the_United_Kingdom#Terminology

As for the motivation behind this issue, I've just seen a few people bring it up lately and it's always good to improve our tagging practise if there's a possibility to. I think we should reach a consensus because it sets a precedent for future cases where someone is unsure and means we're all following the same rule which makes the data more homogeneous.

I suppose the reason I think they should share the same tags is because by definition (at least, following this idea of a format) they are the same brand. It also establishes that there is some relation between the features of different formats.

bhousel:
In cases where each store format has its own Wikidata, we link to the sub-brand QIDs. This is great when the businesses use different branding, like in the case with Carrefour:

<img width="1363" alt="Screenshot 2019-06-12 08 21 28" src="https://user-images.githubusercontent.com/38784/59350854-19d36b80-8ceb-11e9-96f9-9f54c5856092.png">

But where it hasn't been split up into sub-brands, we just assign them all the same Wikidata QID:

[...]

Issue #2740, opened 2019-06-02
Tesco Express - is it shop=supermarket or shop=convenience? (United Kingdom, with some minor usage in Ireland, Czechia, Hungary, Slovakia, Thailand, South Korea, Sri Lanka)


matkoniecz:
Is shop=supermarket sometimes preferable tagging for Tesco Express?

entry data:
```
{
  "tags": {
    "brand": "Tesco Express",
    "brand:wikidata": "Q487494",
    "brand:wikipedia": "en:Tesco",
    "name": "Tesco Express",
    "shop": "supermarket"
  }
}
```

entry data:
```
{
  "nomatch": [
    "amenity/fuel|Tesco"
  ],
  "tags": {
    "brand": "Tesco Express",
    "brand:wikidata": "Q487494",
    "brand:wikipedia": "en:Tesco",
    "name": "Tesco Express",
    "shop": "convenience"
  }
}
```

appears in: United Kingdom [gb] 95%, Ireland [ie] 1%, Czechia [cz] 0%, Hungary [hu] 0%, Slovakia [sk] 0%, Thailand [th] 0%, South Korea [kr] 0%, Sri Lanka [lk] 0%

@SilentSpike, @jeisenbe

SilentSpike:
In my experience Tesco Express is always what I would call a convenience store (plus I think that's the whole purpose of differentiating the name).

I think the issue is that they can be quite big, which makes people think of them as a supermarket. However, they don't have everything a Tesco store usually would (clothing section, kitchenware, etc.) which is why I would say they're a convenience store.

ImreSamu:
> appears in:  **Hungary** [hu] 0%,

in Hungary - we have `name="Tesco Expressz"` ---->  https://taginfo.openstreetmap.org/tags/name=Tesco%20Expressz  ( count = 41  )
We have cleaned ( ~ 7 months ago ; discussed in the local mail list!  current example: https://www.openstreetmap.org/way/353281883 )  
and the  tagging now:
```
    "shop": "convenience",            <-----  because mostly small shops.
    "name": "Tesco Expressz",         <-----   en:Express == hu:Expressz  ; just a different writing.
    ...
    "brand": "Tesco",                 <----- we use this - because it is a "Tesco" Brand! 
    "brand:wikidata": "Q487494",     
    "brand:wikipedia": "hu:Tesco",    <-----  hungarian wikipedia page - for local users.
```

in Hungary we have:

[...]

PR #2803, opened 2019-06-20
Hsbc uk

jnicho02:
Let's try this again then :)


brands/amenity/bank.json
```diff
@@ -3353,6 +3353,16 @@
       "name": "HSBC"
     }
   },
+  "amenity/bank|HSBC UK": {
+    "countryCodes": ["gb"],
+    "tags": {
+      "amenity": "bank",
+      "brand": "HSBC",
+      "brand:wikidata": "Q5413890",
+      "brand:wikipedia": "en:HSBC",
```
1ec5: [Q5413890](https://www.wikidata.org/wiki/Q5413890) and “[HSBC Bank (Europe)](https://en.wikipedia.org/wiki/HSBC_Bank_%28Europe%29)” are for the UK-based European subsidiary, which is still named “HSBC”.

[Q64767453](https://www.wikidata.org/wiki/Q64767453) has been created for the UK-based UK retail banking subsidiary, which apparently has long been officially named HSBC UK but which now brands itself as such. The English Wikipedia only has a section on this sub-subsidiary, not a full article. So if `brand:wikidata` is changed to `Q64767453`, then `brand:wikipedia` should be removed for consistency.

There are Wikidata items for other national subsidiaries, but there’s no need to create separate entries for those subsidiaries, because their retail branches don’t carry national branding like the UK branches do.

brands/amenity/bank.json
```diff
@@ -3353,6 +3353,16 @@
       "name": "HSBC"
     }
   },
+  "amenity/bank|HSBC UK": {
+    "countryCodes": ["gb"],
+    "tags": {
+      "amenity": "bank",
+      "brand": "HSBC",
```
1ec5:
```suggestion
      "brand": "HSBC UK",
```

SilentSpike:
> and brand should correspond to name

This is not true, `brand` and `name` can have different values. See https://github.com/osmlab/name-suggestion-index/issues/2740#issuecomment-498071118 and related issue #2787.

bhousel:
Thanks, I'll clean this up in a followup commit..

PR #2804, opened 2019-06-21
Add Israeli burgers chains

yotamN:


Adamant36:
Are you the one creating the Wikidata entries for a lot of these?

yotamN:
Yes, why?

Adamant36:
I thought so. I was looking them and some of them don't have secondary/external references. We have gotten a lot of Wikidata items deleted for not add them. So it would good if you could double check all of the Wikidata entries you created in this PR, along with the other one, and added secondary/external references where they don't exist. Otherwise, its hard to impossible to get back deleted Wikidata entries and the only way we have to optain brand information. There's an example of what you can add in the contributing file.

brands/amenity/fast_food.json
```diff
@@ -251,6 +251,37 @@
...
+  "amenity/fast_food|Burgerim": {
+    "countryCodes": ["il"],
+    "tags": {
+      "amenity": "fast_food",
+      "brand": "Burgerim",
+      "brand:en": "Burgerim",
+      "brand:he": "בורגרים",
+      "brand:wikidata": "Q64760354",
+      "cuisine": "burger",
+      "name": "Burgerim",
```
1ec5: `name` and `brand` for some of these entries is in English but for other entries is in Hebrew. Is that consistent with how these POIs are mapped (or should be mapped) in Israel in OSM?

yotamN: Some names are obviously in English, for example Burgerim is from the English word burger and their logo is in English (https://d2pnvdnftoj8zk.cloudfront.net/wp-content/uploads/sites/9/2018/01/17034317/blogo1.svg).

Some other brands are more confusing and I tried to use what make more sense. It seems that I was wrong with at least one brand, I will try to look how they are mapped in OSM again

yotamN:
Would it be enough to reference their website? I mean most of the info that I added is either from their website or from physically visiting there.

brands/amenity/fast_food.json
bhousel: Thanks @yotamN  - if their logo is in English you are correct to use English in the name tag 👍

bhousel:
> Would it be enough to reference their website? I mean most of the info that I added is either from their website or from physically visiting there.

Wikidata roughly follows Wikipedia in preferring [secondary sources](https://en.wikipedia.org/wiki/Wikipedia:Identifying_and_using_primary_sources) to establish notability - so while stuff you find on the company's own website can be useful, it should ideally be backed up by something that you find elsewhere.

I have seen plenty of Wikidata pages that don't contain any sourcing.  Whenever I add a page, I try to include at least one secondary source - this is easy to do if the company has a publicly traded stock symbol, or is mentioned in a trade magazine, or in a newspaper saying that they are opening a new branch location somewhere.

PR #2806, opened 2019-06-21
Moving Claire's from jewelry

SelfishSeahorse:
See also #2805

### DIFF
--- a/brands/shop/fashion_accessories.json
+++ b/brands/shop/fashion_accessories.json
@@ -20,6 +20,15 @@
       "shop": "fashion_accessories"
     }
   },
+  "shop/fashion_accessories|Claire's": {
+    "tags": {
+      "brand": "Claire's",
+      "brand:wikidata": "Q2974996",
+      "brand:wikipedia": "en:Claire's",
+      "name": "Claire's",
+      "shop": "fashion_accessories"
+    }
+  },
   "shop/fashion_accessories|Jimmy Choo": {
     "tags": {
       "brand": "Jimmy Choo",


SilentSpike:
Should this not be [`shop=accessories`](https://taginfo.openstreetmap.org/tags/?key=shop&value=accessories) (found [here](https://wiki.openstreetmap.org/wiki/Retail_chains_in_the_United_Kingdom#Accessories)). It has a higher usage, but the wiki page has been deleted 🤔

SelfishSeahorse:
> Should this not be [`shop=accessories`](https://taginfo.openstreetmap.org/tags/?key=shop&value=accessories) (found [here](https://wiki.openstreetmap.org/wiki/Retail_chains_in_the_United_Kingdom#Accessories)). It has a higher usage, but the wiki page has been deleted 🤔

``shop=accessories`` has a higher usage, but it has been used for different accessories shops – not only fashion accessories –, e.g. home accessories or accessories for mobile phones, cars, motorcycles or computers. ``shop=fashion_accessories`` is unambiguous and has been approved.

tas50:
This seems like the right place overall for Claires

tas50:
That being said can you make a single PR that handles the move not a PR for each file?

SilentSpike:
I think it's because the files were edited via github directly. We can just merge both PRs 👍

tas50:
Yeah that’s obvious now that I look at the branch names.
Jet leg has me this week.

SelfishSeahorse:
> That being said can you make a single PR that handles the move not a PR for each file?

I tried to make a single pull request, but i haven't found out how to do this on GitHub.

tas50:
Thanks for bringing this up @SelfishSeahorse. I added the necessary match tag in b0cfd721b4477ee160613f9fe2ed5ddb575ceda7. That way when we run `npm run build` the script doesn't add back the jewelry store.

PR #2807, opened 2019-06-22
Replace Thomson with TUI UK

SilentSpike:
Thomson became TUI UK in 2017, the wiki tagging already reflected this and this just updates the name and brand to match. Unsure if any signage in the UK would still have Thomson on it so opening as a PR for discussion purposes.

For reference:
https://en.wikipedia.org/wiki/TUI_UK

### DIFF
--- a/brands/shop/clothes.json
+++ b/brands/shop/clothes.json
@@ -2379,16 +2379,6 @@
       "shop": "clothes"
     }
   },
-  "shop/clothes|Slaters": {
-    "countryCodes": ["gb"],
-    "tags": {
-      "brand": "Slaters",
-      "brand:wikidata": "Q7538912",
-      "clothes": "suits",
-      "name": "Slaters",
-      "shop": "clothes"
-    }
-  },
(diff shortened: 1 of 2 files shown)


SilentSpike:
Accidentally pushed a commit to the wrong branch and have now reverted. 

Suggest whoever merges this (if it gets merged) does a squash merge to hide the ugly git history.

tas50:
@SilentSpike On that branch you can force push over it with this:

```
git reset 1a23ef5e80e4e7671c99e08d866ce400e6a4e5fa --hard
git push -f
```

SilentSpike:
Yeah but it's best to avoid force pushing a remote branch

1ec5:
GitHub now records an event on the PR page when you force-push, so at least the operation has a paper trail in case others do need to recover the old version or work downstream from the PR branch.

bhousel:
> Yeah but it's best to avoid force pushing a remote branch

It's totally ok to force push to a working branch.. Just considered taboo to force push to `master`.

PR #2811, opened 2019-06-23
Add Hudson Convenience Stores

tas50:
So read this one all the way before thinking it's wrong. Hudson News is the traditional newsagent we're used to at every airport. According to Wikipedia in 2013 they added the Hudson brand, which is what this tag is for. Since 2013 they're been slowly (not so slowly) converting their stores over to the new Hudson brand. I've noticed that everywhere I fly for work lately only has the new Hudson format, which does not sell newspapers. It's just a convenience store, with a much smaller magazine and book section than the previous stores had.

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -570,6 +570,16 @@
       "shop": "convenience"
     }
   },
+  "shop/convenience|Hudson": {
+    "nomatch": ["shop/newsagent|Hudson News"],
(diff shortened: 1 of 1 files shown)


Adamant36:
If it retains the news agent entry until they all transfer over I don't see what the issue would be. It seems they have bookstores and cafes also. It would be cool if they had their own entries to at some point if they don't already.

PR #2812, opened 2019-06-23
More Japanese Brands (part 6)

sguinetti:
Continuation of to #2800. In this pull request adds new bakery shops and edits some restaurant brands.


bhousel:
thank you @sguinetti 👍

PR #2815, opened 2019-06-25
Fitness Centre Updates

mparrault:
-add brand data for GoodLife Fitness
-add two chains called Fit4Less
-remove "au" country code for F45: it exists worldwide


Adamant36:
Do you happen to know which countries F45 are specifically located in?

mparrault:
According to https://f45training.com/change-your-country/ they are located in 25 countries, including the US, UK, China, India, and Canada.

Adamant36:
Hhhmm, thanks. 25 is to many to add individual country code entries for in my opinion.

Adamant36:
Btw, I noticed you created the Wikidata entry for Fit4less. It would really be good if you added an external link reference somewhere in it so it doesn't get deleted by the admins. There's an example in the contributing file. Although, there's probably other ones you could use also. It's a problem we keep having. So it's good to do.

SilentSpike:
I don't think 25 is too many 🤔 If the data is accurate then why not?

Adamant36:
> I don't think 25 is too many 🤔 If the data is accurate then why not?

I guess. The only place sit doesn't seem to be are in Africa and South America. At that point it doesn't seem worth it to me at least, but that doesn't mean they can't be added.

PR #2817, opened 2019-06-25
Canadian Mobile Phone Brands

mparrault:
add Chatr, Virgin Mobile (Canada), WirelessWave and TBooth Wireless


bhousel:
thanks @mparrault  👍

Issue #2734, opened 2019-06-01
generate docs for individual entries, with full information presented


matkoniecz:
It would make review of data easier for local communities. JSON is relatively readable, but not for all and multiple big JSON files are not improving situation.

It would make possible to include full data, with matches and country code list included, including default matches.

Individual entries would be linked from tag lists like http://osmlab.github.io/name-suggestion-index/brands/amenity/bank.html

matkoniecz:
I see three ways how this goal can be achieved

- expand existing lists (undesirable, would break current usecases)
- add individual page for every single entry (thousands of new .html files may have negative effects)
- create list with entries by category (like the current list), each entry with full data, (may be hard to browse)

First seems to be a poor idea, I am not sure which of two remaining ones is better. Anyway, I want to implement it to make #2733 possible (#2733 would be confusing without making country code limitations visible)

Adamant36:
Hhhhmmm, neither of the three options seem optimal. Maybe there's another way of doing it that just hasn't been thought of yet.

matkoniecz:
This is not feasible with current documentation model based on static htmls.

Adamant36:
Bummer. Closing then.

Issue #2438, opened 2019-03-13
Browsable name suggestion index ideas


bhousel:
I generated some summary pages for all the categories here:
 http://osmlab.github.io/name-suggestion-index/brands/index.html 

Try it out - anyone who is interested in improving the logo coverage on Wikidata will find this very useful!

This shows you:
- which brands are missing a `brand:wikidata` tag and 
- which brands with a `brand:wikidata` tag have no logo, or unsuitable logo.

`npm run docs` will regenerate these pages.

I'm opening this issue to get some feedback on what other things we can add to the generated pages to make this even more useful.

bhousel:
- [ ] Add a link to overpass-turbo for each row, so you can see where it is
- [ ] Add anchors so people can link to any row

todrobbins:
One quick thought: ability to filter the tags on the main page:
<img width="791" alt="Screen Shot 2019-03-13 at 10 08 07 PM" src="https://user-images.githubusercontent.com/158590/54330737-8485e580-45dc-11e9-820c-396998e5d9ae.png">

matkoniecz:
404 page at http://osmlab.github.io/name-suggestion-index/ is a bit ugly

matkoniecz:
There are some problems with handling country-limited entries, though impact is not big.

See Agribank at http://osmlab.github.io/name-suggestion-index/brands/amenity/bank.html

Overpass turbo links are misleading (should be country limited).

"< 50" is result of `nocount` that was necessary due to name collisions, but it is not implying "< 50" usage. Maybe "no data" would be better?

matkoniecz:
What is the point of Facebook and Twitter columns? As far as I know it is not something that should be added to OSM or is useful in OSM editors.

Adamant36:
"What is the point of Facebook and Twitter columns"

They are just another source for an icon in case Wikimedia doesn't have it. Also, the icon links are obtained through the Wikimedia entries. So if a Twitter or Facebook entry isn't in the Wikidata page the index won't use those sources. From what I've seen so far they can have better quality icons then Wikimedia in some cases also. Especially Twitter. Twitter icons might be more up to date also. 

Twitter and Facebook isn't used other then that though (that I know of).

bhousel:
> What is the point of Facebook and Twitter columns? As far as I know it is not something that should be added to OSM or is useful in OSM editors.

Yes, we aren't adding the info to OSM, but I'm going to use those links to fetch their logo images to display in iD.  

[...]

PR #2819, opened 2019-06-25
Replace generated docs with a client-side React app

bhousel:
This PR replaces the generated static html doc pages with a React.js app that runs in the client.  It looks exactly the same as the existing doc pages.

The app pulls data from the files in `dist/` so we won't need to generate a bunch of doc pages anymore. 

It also paves the way for some useful features like client-side filtering mentioned in #2734, #2733, #2438


README.md
```diff
@@ -16,7 +16,7 @@ in OpenStreetMap.
 ### Browse the index
 
 You can browse the index at
-http://osmlab.github.io/name-suggestion-index/brands/index.html
+http://osmlab.github.io/name-suggestion-index/brands
```
tas50: this is a 404

bhousel: Yeah I see that now - gh-pages isn't quite compatible with react router, so I need to change a bit how the app works.  I'll fix it tonight..

bhousel: It works now!  I switched from using path-based routing to using querystrings instead..

Adamant36: It appears that clicking on any of the links just loops back to the main brand page.

bhousel: Thanks for reporting @Adamant36  -  I fixed it in 2ec4117f0, but let me know if you find any other issues.

PR #2821, opened 2019-06-26
add coco and sharetea bubble tea

andrewharvey:
following on from #2797

### DIFF
--- a/brands/amenity/cafe.json
+++ b/brands/amenity/cafe.json
@@ -180,6 +180,17 @@
       "takeaway": "yes"
     }
   },
+  "amenity/cafe|CoCo": {
+    "countryCodes": ["au"],
+    "tags": {
+      "amenity": "cafe",
+      "brand": "CoCo",
+      "brand:wikidata": "Q64827025",
+      "cuisine": "bubble_tea",
+      "name": "Coco",
+      "takeaway": "yes"
+    }
+  },
   "amenity/cafe|Coffee Bean": {
     "tags": {
       "amenity": "cafe",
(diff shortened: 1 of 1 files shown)


brands/amenity/cafe.json
```diff
@@ -622,6 +633,17 @@
       "takeaway": "yes"
     }
   },
+  "amenity/cafe|Sharetea": {
+    "countryCodes": ["au"],
```
1ec5: Sharetea is also found in Taiwan, the U.S., and other countries. However, a separate entry would have to be added for Chinese-speaking countries, because the English name would be secondary to the Chinese name.

brands/amenity/cafe.json
```diff
@@ -622,6 +633,17 @@
...
+    "tags": {
+      "amenity": "cafe",
+      "brand": "Sharetea",
+      "brand:wikidata": "Q64827032",
+      "cuisine": "bubble_tea",
+      "name": "Sharetea",
```
1ec5: Even in English-speaking countries like the U.S., the Chinese name 歇腳亭 is part of the chain’s logo on signage, so I think `name:en` and `name:zh` could be added.

andrewharvey: :+1: .

brands/amenity/cafe.json
andrewharvey: >  English name would be secondary to the Chinese name.

Yep, that's why I didn't include `tw` in the countryCodes, since it will probably need to be a separate entry.

I've updated to include US.

bhousel:
Thanks @andrewharvey 👍

PR #2828, opened 2019-06-28
Papyrus & American Greetings

jleedev:
Worth pointing out that Papyrus has 17 of `shop=stationery` and 7 of `shop=gift`, but it seems clear to me both from the OSM wiki and the remainder of this brands index that this is incorrect.

### DIFF
--- a/brands/shop/gift.json
+++ b/brands/shop/gift.json
@@ -1,4 +1,15 @@
 {
+  "shop/gift|American Greetings": {
+    "countryCodes": ["us"],
+    "matchTags": ["shop/stationery"],
+    "tags": {
+      "brand": "American Greetings",
+      "brand:wikidata": "Q464767",
+      "brand:wikipedia": "en:American Greetings",
+      "name": "American Greetings",
+      "shop": "gift"
+    }
+  },
   "shop/gift|Card Factory": {
     "countryCodes": ["gb", "im"],
     "matchTags": ["shop/stationery"],
--- a/brands/shop/gift.json
+++ b/brands/shop/gift.json
@@ -87,6 +87,17 @@
       "shop": "gift"
     }
   },
+  "shop/gift|Papyrus": {
+    "countryCodes": ["us"],
+    "matchTags": ["shop/stationery"],
+    "tags": {
+      "brand": "Papyrus",
+      "brand:wikidata": "Q28222692",
+      "brand:wikipedia": "en:Papyrus (company)",
+      "name": "Papyrus",
+      "shop": "gift"
+    }
+  },
   "shop/gift|Spencer's": {
     "countryCodes": ["ca", "us"],
     "tags": {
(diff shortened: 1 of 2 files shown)


brands/shop/gift.json
```diff
@@ -1,4 +1,15 @@
 {
+  "shop/gift|American Greetings": {
```
1ec5: Based on [this Carlton Cards page](http://carltoncards.com/#CA), I was under the impression that there are only a few store locations, but it looks like [third-party sites](https://www.localstore.net/stores/735537/american-greetings/) list quite a few more. Is it a case of the company only listing owned-and-operated stores, or is it that the third-party sites are outdated and there are only a few left?

jleedev: 9 feels like the right number of proper AG stores remaining; I don't think they ever had a huge number. That third-party site looks outdated.

Unlike Hallmark, AG stores aren't franchised.

1ec5: This project’s documentation doesn’t say so explicitly, but I think so far we’ve been hewing to 50 locations or more as a rule of thumb for inclusion. Should we make an exception for smaller sister brands of brands already listed here? I can think of a few hotel chains that would be included by that exception.

jleedev: Yes, that was my thinking. For instance, the AG store near me used to be a Papyrus, so at some point it would have to be "rebranded", as it were. The UI of iD is in a certain sense hostile to this, and the schemas here consisting of 3+ tags doesn't help.

Imagining if Name Brand Hotel changes a small number of locations to Name Brand Express Hotel, I think that would be something this index would endeavor to gracefully support and set the rule aside.

Adamant36: > Should we make an exception for smaller sister brands of brands already listed here?

My opinion is that we shouldn't focus on them at the expensive of more widespread brands, but if someone wants to add a few here and there, I don't see why it's a problem. It might actually be useful in cases where the brand has the same name as a more popular one like @jleedev says. For instance, there's  some smaller regional banks with the same names as national brands that I've been wanting to add just so they are filtered properly. Although, I know the discussion here is confined to smaller sister brands, I think its similar. I don't think there's a hard and fast rule there has to be 50 more stores to for the brand to be included anyway. It's mostly a cost versus benefit thing with time, added code complexity, etc etc if we add a bunch of brands with only a few stores. 50 is rather arbitrary IMHO though.

bhousel: yeah 50 is an arbitrary number... That's the number that our `build_all_names_ script uses to collect for _automatic_ inclusion, but it's totally ok to manually add brands that have fewer than 50 locations.

It's mostly to cut down on the number of generic names that the script collects.  We know that POI coverage in OSM is not good, and that there are many brands with hundreds of real locations but only a handful mapped so our script will not find them.

PR #2829, opened 2019-06-28
Canadian clothes chains

mparrault:
-Dynamite, Fairweather, Garage, Laura, Melanie Lyne, Moores, Tip Top Tailors
There might be a problem with Garage, since Garage is listed as a generic name


tas50:
Garage was on my todo list. Thanks for getting that one handled.

PR #2830, opened 2019-06-29
Add Wild Birds Unlimited

tas50:
This seems a bit odd as a pet store, but it's as close as it gets and
it's a common way to tag these stores. I'm open to a better tag.


brands/shop/pet.json
```diff
@@ -180,6 +180,16 @@
...
     }
   },
+  "shop/pet|Wild Birds Unlimited": {
+    "countryCodes": ["us"],
+    "tags": {
+      "brand": "Wild Birds Unlimited",
+      "brand:wikidata": "Q8000542",
+      "brand:wikipedia": "en:Wild Birds Unlimited",
+      "name": "Wild Birds Unlimited",
+      "shop": "pet"
```
1ec5: There’s minuscule usage of [`shop=pet_supplies`](https://taginfo.openstreetmap.org/tags/shop=pet_supplies) and [`pet=supplies`](https://taginfo.openstreetmap.org/tags/pet=supplies), but yeah, no great options for clarifying what a “pet store” means yet.

tas50: There's some random non-standard tag usage, but it's pet for the most part.

Adamant36: Shop=pet makes it seem like its a place that sells pets. This place doesn't seem to sell brids. So, shop=pet_supplies seems better, but still not ideal for some reason. There's also shop=pet_food which has 66 uses. It seems like it would fit good since the stores specifically sell bird feeders and bird food. Generally, with 24 ,535 uses of shop=pet I'm sure a lot of them could be more specific. Personally, I'm not a big fan of using pet=* with shop=pet to more clearly tag them.

1ec5: For better or worse, [`shop=pet`](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dpet) is very commonly used for stores that mainly sell pet supplies and not infrequently used for stores that only sell pet supplies. As it is, it’s basically up to data consumers to account for this usage by not equating `shop=pet` with pets on sale.

Before we can confidently use `shop=pet_supplies`, `shop=pet_food`, or `shop=animal_feed` in this index, these tags would need more usage and at least a modicum of software support and documentation.

PR #2835, opened 2019-06-30
Add Leslie's Pool Supplies

tas50:
Add new category of shop/swimming_pool which is listed as a defacto
standard tag. There's about 900 of these stores total and most are just
tagged as stores from my experience. Getting it in the index would
really help us cleanup bad data.

### DIFF
--- a/brands/shop/swimming_pool.json
+++ b/brands/shop/swimming_pool.json
@@ -1,11 +1,15 @@
 {
   "shop/swimming_pool|Leslie's Pool Supplies": {
     "countryCodes": ["us"],
+    "matchNames": [
+      "leslie's swimming pool supplies"
+    ],
     "tags": {
       "brand": "Leslie's Pool Supplies",
       "brand:wikidata": "Q6530568",
       "brand:wikipedia": "en:Leslie's Poolmart",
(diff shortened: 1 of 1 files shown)


brands/shop/swimming_pool.json
```diff
@@ -0,0 +1,12 @@
+{
+  "shop/swimming_pool|Leslie's Pool Supplies": {
```
mds08011: The full name is Leslie's Swimming Pool Supplies.  I'm not sure if we want that as a `name` or `official_name`

tas50: I don't think that's actually their full name anymore. It used to be in the logo, but the new logos say "pool supplies service & repair", which is what the website uses when they refer to their full name:

This:

image


vs this:

image

mds08011: Good catch!  Carry on

tas50: It must be pretty recent because I swear the last one I went to had the old name.

1ec5:
Opened openstreetmap/iD#6599 to request a preset corresponding to the new category.

tas50:
Thanks @1ec5

PR #2837, opened 2019-07-02
Added Circuit Électrique & FLO charging networks, St-Hubert restaurant and St-Hubert Express fast food & other brands and fixes.

mguaylambert:


mparrault:
Thanks for adding St-Hubert.  I agree that St-Hubert & St-Hubert Express should be separate entries.

mguaylambert:
Yeah @mparrault it was a little bit tricky since St-Hubert claim to have 3 branches. That would also mean "Resto-bar St-Hub" but they seem to be always integrated to the restaurant for now so I’m omitting it. For now I see two branches -> The restaurant and the fast food chain. 🙂

Adamant36:
Hi. Would you mind adding a US country code entry for BMO Harris Bank and also a BMO match name to it since your taking it away from the main BMO entry? It seems like there's a few straight BMO entries in America. They might be BMO Capital Markets and not BMO Harris Bank, but either way it would be good if BMO still filtered in America. [This one](https://www.openstreetmap.org/node/6414377838) in Missouri is actually called Bank of Montreal for the official name. So I don't know. Maybe it should be looked into and a separate entry could be created for BMO Capital Markets to keep it from being tagged as just BMO since it doesn't seem to have one.

mguaylambert:
Hey @Adamant36 good point! I removed it because being a BMO & BMO Harris customer iv’e learned the hard way that BMO is only in Canada and same for the later. 😅

bhousel:
Looks pretty good @mguaylambert thank you!
Can you add `"countryCodes": ["ca"]`  to the restaurants that should have it?

PR #2839, opened 2019-07-02
add wikidata for CVS and Rite Aid

jleedev:


SilentSpike:
I think the reason these don't have wikidata is because there's possibly some branched tagging going on (i.e. they mean the same thing) with `amenity=pharmacy` and `shop=chemist`.

I've noticed this while looking as some UK chains too (Boots for example).

tas50:
I think this is the right thing to do if we want people to tag the outer store as the drugstore in the US and the actual pharmacy counter as the pharmacy counter. The sort of practice is particularly helpful when you think of Target locations that include a CVS pharmacy counter, but not a full drugstore.

tas50:
Walgreens is already setup this way btw and folks are using it to correctly tag the drugstore vs. the counter here in Oregon: https://github.com/osmlab/name-suggestion-index/blob/17663a8c274fe048bd1ff4a93a5988c63ecd665c/brands/shop/chemist.json#L200

maxerickson:
They are probably convenience/grocery stores more than they are chemists.

tas50:
https://wiki.openstreetmap.org/wiki/Tag:shop%3Dchemist

Chemist is actually spot on for CVS, Rite Aid and Walgreens. In the US we'd call it a Drug Store, but OSM uses the British wording for tags so it's chemist.

maxerickson:
The type of shop implied by shop=chemist is much smaller than the type of shop that an American drugstore actually is (especially large chains).

PR #2841, opened 2019-07-02
Add TownePlace Suites (Marriott Brand)

mds08011:
TownePlace Suites is a brand of specialized hotel in owned by Marriott.  I am a bit unsure if the name and brand should be “TownePlace Suites” or “TownePlace Suites by Marriott” but the logo states the latter.

According to the website there are “over 400 locations across the US and Canada”

### DIFF
--- a/brands/tourism/hotel.json
+++ b/brands/tourism/hotel.json
@@ -520,6 +520,18 @@
       "tourism": "hotel"
     }
   },
+  "tourism/hotel|TownePlace Suites": {
+    "countryCodes": ["ca", "us"],
+    "matchNames": [
+      "TownePlace Suites by Marriott"],
+    "tags": {
+      "brand": "Courtyard by Marriott",
+      "brand:wikidata": "",
+      "brand:wikipedia": "en:TownePlace Suites",
+      "name": "Courtyard by Marriott",
+      "tourism": "hotel"
+    }
+  },
   "tourism/hotel|Travelodge": {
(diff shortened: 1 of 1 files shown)


brands/tourism/hotel.json
```diff
@@ -520,6 +520,18 @@
       "tourism": "hotel"
     }
   },
+  "tourism/hotel|TownePlace Suites by Marriott": {
```
tas50: I think we want to just make this TownePlace Suites and then have a matchname for TownePlace Suites by Marriot and TownePlace Suites Marriot.

mds08011: I changed the name and added "TownePlace Suites Marriott" and "TownePlace Marriott" as match names.  "TownePlace Suites by Marriott" will be matched via `official_name`

brands/tourism/hotel.json
```diff
@@ -520,6 +520,21 @@
       "tourism": "hotel"
     }
   },
+  "tourism/hotel|TownePlace Suites": {
+    "countryCodes": ["ca", "us"],
+    "matchNames": [
+      "TownePlace Suites Marriott"
```
tas50: There should be a comma at the end of this line. It's causing travis failures

mds08011: Thanks!  Just added it

tas50:
Thanks @mds08011

PR #2842, opened 2019-07-02
Add Aloha Petroleum

mds08011:
Aloha Petroleum is a subsidiary of Sunoco LP which operates gas stations under both their own brand and Shell stations in Hawaii.


bhousel:
Thanks @mds08011 👍

PR #2850, opened 2019-07-05
Planned Parenthood

1ec5:
Added an entry for Planned Parenthood, since they have a number of locations across the U.S. and existing features in OSM have a wide variety of tags. Not sure if there are any other [healthcare tags](https://wiki.openstreetmap.org/wiki/Key:healthcare) that I should add besides the counseling ones. Note that I didn’t include `healthcare:speciality=abortion`, because not all locations offer that service.

### DIFF
--- a/brands/amenity/clinic.json
+++ b/brands/amenity/clinic.json
@@ -52,6 +52,21 @@
       "name": "Fresenius Medical Care"
     }
   },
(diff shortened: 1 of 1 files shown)


bhousel:
thanks @1ec5  👍

PR #2860, opened 2019-07-08
Add the two Bealls department stores in the US

tas50:
So from what I can tell there's Bealls in just Florida and then Bealls
based out of Texas, but with stores throughout the US (except Florida).
I'm opening this as a PR to discuss how to name the geographic
restriction. (Texas) doesn't seem right since there's stores in Oregon
too. Would (non-Florida) be appropriate.

### DIFF
--- a/brands/shop/department_store.json
+++ b/brands/shop/department_store.json
@@ -10,6 +10,26 @@
       "shop": "department_store"
     }
   },
+  "shop/department_store|Bealls (Florida)": {
+    "countryCodes": ["us"],
+    "tags": {
+      "brand": "Bealls",
+      "brand:wikidata": "Q4876153",
+      "brand:wikipedia": "en:Bealls (Florida)",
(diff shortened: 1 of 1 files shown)


bhousel:
Hmm yeah it's weird but I'd go with `Bealls~(Florida)` and `Bealls~(non-Florida)`

1ec5:
`(National)` could work too for the Texas-based one. Though this reminds me I need to clean up some of the disambigators I added for banks, because some of them are just the state the bank is based in, even when they cross state lines. (For example, there’s a bank based in Puerto Rico but I don’t know if their customers in California would know to use a preset labeled as such.)

bhousel:
yeah it's tricky... for "Price Chopper" I went with `~(New York)` and `~(Kansas City)` even though both of them have spread to nearby states.. I think people who know the brands can figure it out.

PR #2862, opened 2019-07-09
Add brand for VFW Post

### LINKED ISSUE #2845: Add brand for VFW post
I see that brands are being added for [social centers](https://osmlab.github.io/name-suggestion-index/index.html?k=amenity&v=social_centre) now. A brand for US [VFW](https://en.wikipedia.org/wiki/Veterans_of_Foreign_Wars) posts would be useful, I've seen these very inconsistently tagged. Note that the names typically vary and include a number, like "VFW Post 3944".

quincylvania:
Closes #2845.

I added the codes for countries with more than just one or two posts. 

https://oms.vfw.org/findapost/foreign.aspx

### DIFF
--- a/brands/amenity/social_centre.json
+++ b/brands/amenity/social_centre.json
@@ -62,5 +62,26 @@
       "name": "Odd Fellows Hall",
       "official_name": "Independent Order of Odd Fellows"
     }
+  },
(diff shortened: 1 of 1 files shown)


tas50:
Thanks @quincylvania

Issue #2846, opened 2019-07-03
Add brand for Masonic Temple


quincylvania:
Similar to #2845. These are also tagged inconsistently.

https://en.wikipedia.org/wiki/Masonic_Temple

1ec5:
Masonic Temples are sort of getting into a gray area… from what I’ve seen, the level of branding consistency is similar to most church denominations, which would call for a different tag like `operator`. I suppose the same could be said for most of these social centers though.

quincylvania:
@1ec5 True. I still think it'd still be useful to suggest consistent tags for these kind of social centers in iD. I think the main difference between these and most brands is that the name could vary?

That said, I don't care strongly about adding them.

1ec5:
There’s a lot of brands in this index where the name can vary, so that alone wouldn’t determine whether it gets included. But I guess it comes down to the meaning of a “brand” versus an operator, which is a distinction I’m less and less sure about these days.

I also added several organizations as social center “brands” with the reasoning that they’re tagged very inconsistently. We haven’t even begun adding labor unions, and they all have lots of local halls too. But I think we’re at the very limit of the index’s current scope, and I’m wondering if we should actually remove some social center entries from the index.

There is a need for more consistent tagging of things that clearly don’t warrant `brand` tags, like public transportation facilities and places of worship. Public transportation already has a `network` tag for clarifying affiliation, and places of worship have `denomination`. Even when a denomination has [an abundance of common symbolism](https://wiki.openstreetmap.org/wiki/Tag:denomination%3Dla_luz_del_mundo#How_to_identify) from location to location, the locations don’t get `brand` tags. I do think we should expand the index’s scope at some point, but it needs to be deliberate, because clients are probably using this index in ways that are specific to commercial branding.

bhousel:
I think these are all ok and we should include them... Currently in iD, if the preset includes both `brand` and `name` fields, users are allowed to customize the `name` field.   (Anyone who really knows what they are doing can change the `name` tag by editing the raw tags or using a different editor, we just encourage users to leave the name field alone in most cases when dealing with branded stuff).

PR #2863, opened 2019-07-09
Add Greyhound bus lines (US)

quincylvania:
Inspired by discussion in #2846, I'd like to see what people think about adding branded transit stops/stations. I think it'd be useful, especially for data consumers focused on public transport routing.

The main obstacle I see is that a single transit station can sometimes be used by multiple carriers. Another obstacle is that this index can only filter geographically on a country-wide basis, while most transit brands are just city-wide.

I'm starting with Greyhound since they're national and tend to run their own stations.

### DIFF
--- a/brands/public_transport/station.json
+++ b/brands/public_transport/station.json
@@ -1,0 +1,16 @@
+{
+  "public_transport/station|Greyhound Station": {
+    "countryCodes": ["us"],
+    "matchNames": ["greyhound terminal"],
+    "tags": {
+      "amenity": "bus_station",
+      "brand": "Greyhound Lines",
+      "brand:wikidata": "Q755309",
+      "brand:wikipedia": "en:Greyhound Lines",
+      "bus": "yes",
+      "name": "Greyhound Station",
+      "name:en": "Greyhound Station",
+      "public_transport": "station"
+    }
+  }
+}


brands/public_transport/station.json
```diff
@@ -0,0 +1,16 @@
...
+    "matchNames": ["greyhound terminal"],
+    "tags": {
+      "amenity": "bus_station",
+      "brand": "Greyhound Lines",
+      "brand:wikidata": "Q755309",
+      "brand:wikipedia": "en:Greyhound Lines",
+      "bus": "yes",
+      "name": "Greyhound Station",
+      "name:en": "Greyhound Station",
+      "public_transport": "station"
```
tas50: You might want to add a short name of just Greyound here so we get that as a matcher. Also thanks for doing this one. I have a note with all the brands I run across that need an entry and this was in it.

tas50: Also there's Greyhound Canada that might be worth adding while you're in this file.

quincylvania: The name matcher only matches against features with the base tags, right? There could be various bars or other places called just "Greyhound".

And yes, we should add the other Greyhound subsidiaries once this is merged.

brands/public_transport/station.json
```diff
@@ -0,0 +1,16 @@
+{
+  "public_transport/station|Greyhound Station": {
+    "countryCodes": ["us"],
+    "matchNames": ["greyhound terminal"],
+    "tags": {
+      "amenity": "bus_station",
+      "brand": "Greyhound Lines",
+      "brand:wikidata": "Q755309",
```
1ec5: OSM uses `network` and `operator` tags for public transportation facilities, both of which come with `:wikidata` subkeys as well. I suppose `brand` would indicate a relationship such as Ford sponsoring the GoBike program in the San Francisco Bay Area, but that’s much less important than capturing the network and operator.

brands/public_transport/station.json
```diff
@@ -0,0 +1,16 @@
+{
+  "public_transport/station|Greyhound Station": {
+    "countryCodes": ["us"],
```
1ec5: Luckily, Greyhound happens to be a national carrier. But as we continue to add public transportation providers, we’re going to run into lots of conflicts across states. Unlike with brands and even banks, most public transportation operators are confined to a single state and their state coverage is quite stable. So we’ll need a way to track [ISO 3166](https://en.wikipedia.org/wiki/ISO_3166) subnational codes, not just country codes, even if we don’t happen to use it for Greyhound.

quincylvania:
> Given the difference in tagging, coverage tracking, and well, purpose, what do you think about moving this listing to a new networks/ folder?

@1ec5 Interesting idea, but I don't see harm in continuing to use the `brand` tag (and folder) for these for consistency. When I walk into a BART station I'd certainly say it's "branded" as a BART station.

Adding `network` and `network:wikidata` sounds fine since they're already in use. The `operator` and `operator:wikidata` could be okay if there aren't any exceptions for the given transit provider.

> I think clients would need to handle these entries differently than brands anyhow.

Some clients might, but not all of them, and probably not iD. My own preference is to avoid fragmentation.

What do you think, @bhousel?

brands/public_transport/station.json
quincylvania: Agreed. We don't want Los Angeles bus stops to be recommended in Boston.

1ec5:
> Interesting idea, but I don't see harm in continuing to use the brand tag (and folder) for these for consistency. When I walk into a BART station I'd certainly say it's "branded" as a BART station.

I definitely see similarity between BART station signage and retail branding. But at least in the U.S., that isn’t often or even usually the case with public transportation operators:

* In Cincinnati, some bus stops have [Metro signs](https://www.mapillary.com/map/im/6DQHLosgw7K6Z2_zu1sdwQ), but other bus stops are marked by nothing but a red line painted on a telephone pole. [This stop outside a Greyhound station has both](https://binged.it/2LLFvqu), but you get the idea.
* Blazers were the norm across the U.S. until 10–15 years ago. San Francisco Muni mostly switched over to signs [in 2015](https://sf.streetsblog.org/2015/04/23/munis-yellow-pole-markings-at-transit-stops-will-be-replaced-by-real-signs/), but some [lone blazers](https://www.mapillary.com/map/im/loopKnLzud51mJorb4LJjQ) remain.
* In San José (which will have a BART station someday 🤞), VTA bus stations do have retail-style branding, but all the bus stop signs just have [generic “Bus Stop” signs](https://www.mapillary.com/map/im/s5odVBZABBAsBHYDPyQEYQ). Unless there’s a VTA-branded shelter, you practically have to walk up and squint to see any mention of VTA.

Meanwhile, Greyhound shuttered its San José station in favor of sharing facilities at a VTA bus station. You only find the Greyhound logo on the buses themselves. It would be appropriate to include Greyhound in Diridon Station’s `network` tag, but it would be inappropriate to add a `brand` tag for it. I wouldn’t be surprised if discount intercity lines like Megabus are forcing them to colocate in other cities too. (Megabus stops are far less expensive to operate. The San Francisco stop consists of nothing more than the driver holding up a sign when he isn’t busy helping customers.)

I don’t see much harm in tagging the more built-out Greyhound stations with `brand`, or those of any private intercity line for that matter. But I’m eager to see NSI grow beyond branding, so that it can be more useful to data consumers. Going back to the original post:

> I think it'd be useful, especially for data consumers focused on public transport routing.

Maybe `brand` could distinguish between signposted and unsignposted stops, but that distinction would only maybe be meaningful to renderers. It’d be much more important for routers and search engines that `network` be consistent across all the stops in a network, regardless of build quality, and even more important that the routes are properly tagged.

Given that [`network` contains abbreviations](https://wiki.openstreetmap.org/wiki/Key:network#Public_transit_routes), some of which are nonobvious and few of which are unique, NSI would have an important role to play in helping mappers and routers make sense of it all. If I were building public transportation routing, I’d want to support `network:wikidata` rather than have to perform spatial queries just to figure out what “Metro” means in a given city. A network/ folder could also be a place to compile presets for road and cycle routes, few of which have any branding per se.

Admittedly, adding actual support for public transportation networks would be well beyond the original scope of this PR. Aside from adding build support for a network/ folder, we’d also need to implement suggestions for the Network field, similar to what exists for Name today. So I’m not sure how you’d all like to proceed, but I hope we can continue this conversation.

bhousel:
> What do you think, @bhousel?

I think OSM convention for transit is to use `operator`/`operator:wikidata` and `network` tags like @1ec5 said.  We have to follow it, even if it makes more work for us.  I thought about this a bit in https://github.com/osmlab/name-suggestion-index/issues/2243#issuecomment-451356517

It would be cleanest to start a new tree for them outside of `brands/`.  It would also require some code changes to the project, but nothing too onerous.  For example we're excluding `operator` tags from the brands, but we'd include that tag in the public transit tree, so we'd need a different json schema to validate the entries there.  And like mentioned above we'd need to handle regions better - country isn't granular enough.

quincylvania:
@1ec5 @bhousel Sounds good. Thanks for taking the time to respond and for bearing with me. I've only been tangentially involved in this project so I've missed plenty of discussion like #2243.

Closing this so we can focus on a more comprehensive solution 👍

tas50:
Oh man I'm that user responding to a closed user now, but would Greyhound be considered a public transit system that should be tagged like one? There's no public funding there. It's just a private business that has private business locations, which seems like it's a brand, which would fit nicely into the existing tagging structure.

1ec5:
Yeah, that’s why I mentioned that it’s OK to use `brand` for Greyhound. It’s a retail establishment specializing in intercity transportation, just as the various `amenity=kindergarten` entries in this index are retail establishments specializing in early childhood education. As far as transportation goes, think Greyhound is an edge case that we can include in the index, but I don’t think it should set a precedent for other bus operators.

1ec5:
#2864 tracks expanding the scope of this project to include public transportation networks.

PR #2867, opened 2019-07-10
Canadian Clothes Chains

mparrault:
Bluenotes, La Vie en Rose, Le Château, Stitches, Suzy Shier, and West 49


bhousel:
thanks @mparrault 👍

PR #2871, opened 2019-07-11
3 Cafes and a bank

mparrault:
-Aroma Espresso Bar, Bridgehead, and Timothy's cafes
-Alterna Savings bank

Aroma Espresso Bar already has an entry for Israel, but it also has locations elsewhere.


brands/amenity/cafe.json
```diff
@@ -35,6 +35,24 @@
       "takeaway": "yes"
     }
   },
+  "amenity/cafe|Aroma Espresso Bar~(International)": {
```
tas50: You can leave off the International bit here

Adamant36: I just merged it and fixed it in 124d64d. Thanks @mparrault :+1:

PR #2873, opened 2019-07-11
Remove duplicate bank (EastWest Bank)

seav:
Note: Name of bank as seen in their signages is "EastWest Unibank" although the Wikipedia article is under "EastWest Bank".


bhousel:
Thanks @seav 👍 
In f887e98b3 I added a `matchName` property for "eastwest bank" so that the less preferred name will still get matched if it's seen anywhere.

Issue #2876, opened 2019-07-12
Add CityMD


yoasif:
eg https://www.citymd.com

Adamant36:
Wikidata Q22295471

mds08011:
Added: https://github.com/osmlab/name-suggestion-index/commit/03a24a1a974c02a631bb44502183d079b63274c2

PR #2880, opened 2019-07-12
Add CityMD

mds08011:
Adds CityMD as requested in Issue #2876


mds08011:
Is having both `amenity=clinic` and `healthcare=clinic` appropriate redundant?

tas50:
Health things are double tagged

PR #2890, opened 2019-07-15
Update redbox to vend video_games and movies

tas50:
They stopped selling just DVDs a long time ago.


Adamant36:
What percent of kiosks rent video games though?

tas50:
All the ones I've ever been to have games as well. I'll see if I can figure out a breakdown from their site.

tas50:
I *think* that they don't all have games, but their support folks won't answer that for me. They gave me the physical address of their corporate office and told me to write them a letter asking. So movies it is. I'll update the PR

SilentSpike:
Ah never mind, I see this was discussed 👍

PR #2891, opened 2019-07-15
Add Kroger Marketplace

tas50:
This is Krogers new Fred Meyer like supercenter store. They have over
100 over them in the US and it looks like they're converting other
brands to this new format. I'm sure in a few years they'll finally kill
of Fred Meyer and replace it with this :(

This uses the same wiki IDs like how we did it with Walmart. I did make the brand Kroger Marketplace since that seems like the correct thing, but let me know if that should be something different.

### DIFF
--- a/brands/shop/supermarket.json
+++ b/brands/shop/supermarket.json
@@ -1940,6 +1940,16 @@
       "shop": "supermarket"
     }
   },
+  "shop/supermarket|Kroger Marketplace": {
+    "nomatch": ["amenity/fuel|Kroger"],
+    "tags": {
+      "brand": "Kroger Marketplace",
+      "brand:wikidata": "Q153417",
+      "brand:wikipedia": "en:Kroger",
+      "name": "Kroger Marketplace",
+      "shop": "supermarket"
+    }
+  },
   "shop/supermarket|Kvickly": {
     "countryCodes": ["dk"],
     "tags": {


1ec5:
> I did make the brand Kroger Marketplace since that seems like the correct thing, but let me know if that should be something different.

Either way’s OK I guess. If you think of `brand` as the human-readable version of `brand:wikidata`, then they should match. But ideally there would be a Wikidata entity for the Marketplace format anyways.

bhousel:
This looks ok to me too, thanks!
I agree it would be a little better to have the store formats each have their own brand and wikidata qid, but it's also ok to share them if that's all wikidata has right now.

tas50:
I may go and create wikidata entries for these later since the stores sharing the codes really messes up how iD displays them. Here's what happens with the hotels that share IDs:

image

1ec5:
Oh so that’s what’s going on with the hotel presets. Yeah, if QIDs are going to be treated as primary keys in the database, as it were, then they should be unique. In the meantime, maybe iD could match on a combination of `brand` and `brand:Wikidata`?

bhousel:
The QIDs are not really treated as primary keys.  iD matches presets on all the tags that are present on the feature.  I'm not really sure why the icons are missing, (not really working on iD these days), but I can dig in and see what's going on.

1ec5:
Oh, the screenshot above isn’t really about the missing image – it’s about the preset name being “Comfort Inn” even though the brand is Quality Inn. iD seems to have chosen the first brand that has the same `brand:wikidata` tag, regardless of the `brand` tag.

bhousel:
> iD seems to have chosen the first brand that has the same `brand:wikidata` tag, regardless of the `brand` tag.

AFAIK - iD should not do that.
Because there are 2 presets - one with `"brand": "Comfort Inn & Suites"` and one preset with `"brand": "Quality Inn & Suites"` - the one that matches more tags should be the one with the higher matchScore, and the one iD picks.

1ec5:
Ah, that’s good to know. Maybe we should document somewhere how clients match (or should match) brands in edge cases like this.

PR #2895, opened 2019-07-16
Canadian Shop Brands

mparrault:
-Expedia CruiseShipCenters, Long & McQuade, and Vidéotron


bhousel:
Thanks @mparrault, these look great!

PR #2897, opened 2019-07-17
Add Toolstation

peternewman:
First time at this, hopefully it's correct

### DIFF
--- a/brands/shop/doityourself.json
+++ b/brands/shop/doityourself.json
@@ -447,6 +447,16 @@
       "shop": "doityourself"
     }
   },
+  "shop/doityourself|Toolstation": {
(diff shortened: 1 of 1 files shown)


bhousel:
This looks great @peternewman, thanks again 👍 

I removed the redundant `matchTags` property in ad814110a
because there is a "hardware" match group like we were talking about in https://github.com/osmlab/name-suggestion-index/commit/8902efdb752d703c70ed8f0b7775be41cd29697f#commitcomment-34338485

(these tags are automatically considered equivalent for matching):
```js
    "hardware": [
        "shop/doityourself",
        "shop/hardware",
        "shop/trade"
    ],
```

PR #2898, opened 2019-07-17
Also catch Iceland as convenience or frozen_food

peternewman:
Personally I think it should be frozen_food as per:
https://wiki.openstreetmap.org/wiki/Tag:shop%3Dfrozen_food

But Overpass Turbo suggests supermarket is far more popular.

CC @maxerickson @bhousel

### DIFF
--- a/brands/shop/supermarket.json
+++ b/brands/shop/supermarket.json
@@ -1728,6 +1728,7 @@
     }
   },
   "shop/supermarket|Iceland": {
+    "matchTags": ["shop/convenience", "shop/frozen_food"],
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks and welcome @peternewman 👋 

I agree `shop=frozen_food` is probably better for this brand, so switched it in 8902efdb7

There is a lot of overlap in tagging sometimes because it's impossible for casual OSM contributors to know what all the tags are.  By adding brands to this index, we can help nudge mappers towards using better tags to describe these businesses!

PR #2899, opened 2019-07-17
Remove a redundant matchTags as it's already in a group

peternewman:


bhousel:
Awesome - you got it!  I beat you to it in ad814110a 😄

PR #2903, opened 2019-07-18
VA

1ec5:
Added entries for the VA Medical Center and Vet Center systems run by the U.S. Veterans Health Administration. VA Medical Centers go by a variety of names by region.

### DIFF
--- a/brands/amenity/clinic.json
+++ b/brands/amenity/clinic.json
@@ -103,5 +103,25 @@
       "healthcare": "clinic",
       "name": "Terveystalo"
     }
+  },
+  "amenity/clinic|VA Clinic": {
+    "countryCodes": ["us"],
+    "matchNames": [
+      "department of veterans affairs clinic",
+      "veterans administration",
+      "veterans administration clinic",
+      "veterans affairs",
+      "veterans affairs clinic"
+    ],
+    "tags": {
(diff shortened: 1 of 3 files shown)


brands/amenity/clinic.json
```diff
@@ -103,5 +103,25 @@
...
+    "matchNames": [
+      "department of veterans affairs clinic",
+      "veterans administration",
+      "veterans administration clinic",
+      "veterans affairs",
+      "veterans affairs clinic"
+    ],
+    "tags": {
+      "amenity": "clinic",
+      "brand": "VA",
```
mds08011: I would recommend either using the full name or the acronym VHA to prevent confusion between the Department of Veterans Affairs (overall cabinet-level department) and the Veterans Health Administration (subdivision/administration).

1ec5: “Department of Veterans Affairs” and “Veterans Health Administration” only appear on signage etc. the way an `operator` would. But “VA” figures prominently in the logo and the names of many facilities.

mds08011: I just checked some photos and had forgotten how prominent the VA is on signs.

bhousel:
Thanks @1ec5 👍

PR #2904, opened 2019-07-18
changed brand to Herfy

DavidHunterJS:
I corrected the brand and change it to Herfy.

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -853,7 +853,7 @@
     "countryCodes": ["sa", "bh", "ae", "kw", "bd"],
     "tags": {
       "amenity": "fast_food",
-      "brand": "fast_food",
+      "brand": "Herfy",
       "brand:wikidata": "Q5738371",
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks for the correction!  But I already did this in 37bdda1f1

DavidHunterJS:
I see that now, thank you.

DavidHunterJS:
Hey @bhousel, git question: What's next? 
I've done the following locally
`git fetch upstream`
`git rebase upstream/master`
`git push origin master`
Now, should I do `git brach -D add-herfy` and then do a `git push origin --delete add-herfy` ?
Is that any different from clicking the 'Delete branch' button above?
thanks

bhousel:
> Now, should I do `git brach -D add-herfy` and then do a `git push origin --delete add-herfy` ?
> Is that any different from clicking the 'Delete branch' button above?

Sorry I missed this question before!  I _think_ these are the same..

Sometimes I go through GitHub and delete old branches - there are always a bunch hanging around.  
They don't really cause any problem.

Issue #2894, opened 2019-07-16
CVS branded vending machines


bhousel:
As [reported here](https://www.reddit.com/r/openstreetmap/comments/cdpfln/any_ideas_on_how_to_map_these_newish_cvs_vending/)

My recommendation:
```
amenity=vending machine
brand=CVS
brand:wikidata=Q2078880
brand:wikipedia=en:CVS Pharmacy
vending=chemist
```

DavidHunterJS:
May I give this one a try?

bhousel:
Yes, thank you!

PR #2905, opened 2019-07-19
added CVS vending machine

DavidHunterJS:
I've added the CVS vending machine using the recommend code form bhousel  in #2894 as well as these tags

`payment: yes`
`weelchair: yes`
`covered: yes`

taken from [this press release](https://cvshealth.com/newsroom/press-releases/cvs-pharmacy-thinks-outside-box-introduction-health-and-wellness-vending) from CVS Health about the machines.

### DIFF
--- a/brands/amenity/pharmacy.json
+++ b/brands/amenity/pharmacy.json
@@ -140,7 +140,10 @@
   },
   "amenity/pharmacy|CVS": {
     "matchNames": ["cvs pharmacy"],
-    "nomatch": ["shop/chemist|CVS"],
+    "nomatch": [
+      "amenity/vending_machine|CVS",
+      "shop/chemist|CVS"
+    ],
     "tags": {
       "amenity": "pharmacy",
       "brand": "CVS",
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks and welcome @DavidHunterJS 🎉 

I'm not sure how I feel about the other tags `payment` `wheelchair` and `covered` - we don't use these tags anywhere else in the index and so I'd prefer for mappers to supply them.  (Anything that appears in this index should be considered "universally true" for the brands).

I removed the extra tags and resolved the merge conflicts in 2cb700d1f

I also changed the names of all the CVS entries to have `"name": "CVS Pharmacy"` and `"short_name": "CVS"`.  This is something someone mentioned on different ticket a while ago #2236, but we never got around to doing it.  (Reasoning for this decision is because the storefronts are signed as "CVS Pharmacy" and it is what the brand calls itself, so that can be the value for the `name`.  However it's totally ok to use the "CVS" to label these features on a map, so that can be a `short_name`).

DavidHunterJS:
I get the universality aspect of that, thanks for the update!

PR #2906, opened 2019-07-20
fixed norauto dupe error

DavidHunterJS:
I added the nomatch tag
`shop/car_parts|Norauto`
to brands>shop>car_repair|Norauto
this fixed the dupe error when running `npm run build`.
edit: it's actually a warning, not an error.

### DIFF
--- a/brands/shop/car_repair.json
+++ b/brands/shop/car_repair.json
@@ -288,6 +288,7 @@
       "pt",
       "ro"
     ],
+    "nomatch": ["shop/car_parts|Norauto"],
     "tags": {
       "brand": "Norauto",
       "brand:wikidata": "Q3317698",


bhousel:
I'm wondering should we match them instead?

Are they the same thing? (in which case we should add a `matchTags` property)

Or are there some Norauto shops that are car_parts and others that are car_repair? (in which case we should add `nomatch` property)

DavidHunterJS:
The latter is true, " *Mobivia* (previously *Norauto*) is a French
<https://en.wikipedia.org/wiki/France> based company group which focuses on
car repairs, car accessories and car parts."
taken from their wiki page <https://en.wikipedia.org/wiki/Mobivia_Groupe>

DavidHunterJS:
> (in which case we should add nomatch property)

What I've added resolves the error in npm run build, is there another `nomatch` property to be added elsewhere?

bhousel:
Thanks for the info - it sounds like they are probably the same thing.
Instead of adding nomatch, I've removed the duplicate entry in eb7e8b8

DavidHunterJS:
Thanks!

Issue #2908, opened 2019-07-20
Add MedPlus


txtsd:
```
amenity=pharmacy
healthcare=pharmacy
dispensing=yes
brand=MedPlus Pharmacy
brand:wikidata=Q65684234
brand:wikipedia=en:MedPlus
name=MedPlus
```

Store locator: https://www.medplusindia.com/store-locator.htm

PR #2910, opened 2019-07-21
Add MedPlus (Pharmacy)

mds08011:
Adds MedPlus per Issue #2908

### DIFF
--- a/brands/amenity/pharmacy.json
+++ b/brands/amenity/pharmacy.json
@@ -778,6 +778,18 @@
       "name": "Marc's Pharmacy"
     }
   },
+  "amenity/pharmacy|MedPlus Pharmacy": {
+    "countryCodes": ["in"],
+    "tags": {
+      "amenity": "pharmacy",
+      "brand": "MedPlus",
+      "brand:wikidata": "Q65684234",
+      "brand:wikipedia": "en:MedPlus",
+      "healthcare": "pharmacy",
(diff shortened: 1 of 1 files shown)


matkoniecz:
> "dispensing": "yes",

Is it certainly applying to all locations?

bhousel:
Thanks @mds08011  👍 
I made the name consistent and removed the `dispensing` tag in 1875a2cb5
We don't include this tag because most people don't use it.

1ec5:
Isn’t there a field for `dispensing` in iD?

bhousel:
> Isn’t there a field for `dispensing` in iD?

Yes but I don't understand it.  We probably added it because someone asked us to.

I read these pages but I still don't understand what the field is for:
https://wiki.openstreetmap.org/wiki/Key:dispensing
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dpharmacy

`dispensing=yes` seems like every pharmacist.  I'm not even sure `dispensing=no` is a real thing.
It even says "The default value is country specific."

Why even tag it then?  Who wants this data or can even make use of it?

mds08011:
The talk page seems to indicate it's based on a very technical UK distinction: https://wiki.openstreetmap.org/wiki/Talk:Tag:amenity%3Dpharmacy

txtsd:
Fair enough. And thanks!

PR #2914, opened 2019-07-23
Add Arabic name and brand to Herfy (fast food)

mds08011:
Herfy is a Saudi Arabia based fast food chain with both English and Arabic signs and logos.  The English branding is already in the NSI and this PR adds `brand:ar` and `name:ar`.

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -876,6 +876,27 @@
       "takeaway": "yes"
     }
   },
+    "amenity/fast_food|هرفي": {
+    "countryCodes": [
+      "ae",
+      "bh",
+      "kw",
+      "sa"
+    ],
+    "tags": {
+      "amenity": "fast_food",
+      "brand": "هرفي",
+      "brand:en": "Herfy",
+      "brand:ar": "هرفي",
+      "brand:wikidata": "Q5738371",
+      "brand:wikipedia": "ar:هرفي",
+      "cuisine": "burger",
+      "name": "هرفي",
+      "name:en": "Herfy",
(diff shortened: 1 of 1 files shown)


brands/amenity/fast_food.json
```diff
@@ -865,10 +865,14 @@
     "tags": {
       "amenity": "fast_food",
       "brand": "Herfy",
```
1ec5: While we’re at it, there should be a separate entry for Saudi Arabia, the UAE, Kuwait, and Bahrain with `brand` and `name` set to `هرفي`.

mds08011: I’m new to localization tagging.  Is that because the `name:ar` will present the POI information in Arabic to Arabic users but the POI will get tagged with `name=Herfy` regardless?

1ec5: Right, and multilingual renderers like the one on osm.org will just show “Herfy”, despite the usual practice of using local languages in that tag. If you search Nominatim for “Herfy”, most existing results are tagged with Arabic in the `name` tag. These features would get “upgraded” to English.

mds08011: But do we know the multilingual branding is broken up by country?

I see some `name=herfy` POIs in Saudia Arabia too: https://overpass-turbo.eu/s/L0A

@waso99 do you have any insight into this?

bhousel: > Edit: I added a separate entry for هرفي with Saudi Arabia, the UAE, Kuwait, and Bahrain country codes.

Looks good to me, thanks!

PR #2917, opened 2019-07-24
Adding more information to Russian Автомир

bkissin:
Found the wikidata and wikipedia information

### DIFF
--- a/brands/shop/car.json
+++ b/brands/shop/car.json
@@ -324,11 +324,10 @@
     "tags": {
       "brand": "Автомир",
       "name": "Автомир",
-      "name:en": "Autoworld"
-      "shop": "car"
-      "brand:wikidata": "Q4056321"
-      "brand:wikipedia": "ru:Автомир (автодилер)"
-      "shop": "car"
+      "name:en": "Autoworld",
+      "shop": "car",
+      "brand:wikidata": "Q4056321",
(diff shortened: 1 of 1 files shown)


brands/shop/car.json
```diff
@@ -324,6 +324,10 @@
     "tags": {
       "brand": "Автомир",
       "name": "Автомир",
+      "name:en": "Autoworld"
+      "shop": "car"
```
mds08011: This line is not needed as there is already a `"shop": "car"` on line 331.

bkissin: Happy to fix. How do I do it? I'm still new to github.

bkissin:
OK, I think I fixed it.

brands/shop/car.json
```diff
@@ -324,7 +324,10 @@
     "tags": {
       "brand": "Автомир",
       "name": "Автомир",
-      "shop": "car"
+      "name:en": "Autoworld",
+      "shop": "car",
+      "brand:wikidata": "Q4056321",
+      "brand:wikipedia": "ru:Автомир (автодилер)",
```
mds08011: This last line should not have a comma.  I think you'll be set after that.

bkissin: LOL thanks!

PR #2920, opened 2019-07-26
Canadian Mall Brands

mparrault:
add Bentley, Alia N TanJay, Bootlegger, Ricki's, Cleo, and Northern Reflections
add wikidata for Roots


mds08011:
Thank you!

Issue #2864, opened 2019-07-09
Public transportation networks


1ec5:
Public transportation infrastructure, such as stations, stops, routes, vending machines, and ticket validators, are tagged with `network` and `operator` tags. name-suggestion-index could help mappers and data consumers work with the `network` tag, which currently poses a couple challenges:

* The `network` tag is quite overloaded, being used for everything from ATMs to road routes. iD can’t rely on taginfo to populate a list of suggestions, because those suggestions can be polluted with popular values for other kinds of features.
* [By convention](https://wiki.openstreetmap.org/wiki/Key:network#Public_transit_routes), `network` often contains an acronym, which may or may not be unique even within a single state. At the same time, it’s usually the name of the agency, not a colloquial name like “RTA” or “Metro” that most people would initially turn to.

By pairing `network` values with [ISO&nbsp;3166](https://en.wikipedia.org/wiki/ISO_3166) subnational codes, friendly names and disambiguators, and `network:wikidata` tags, name-suggestion-index could both help mappers choose the right values and give data consumers machine-readable data to work with. For example, a mapper in Los Angeles could search iD’s presets for “metro” and choose between “<img src="https://upload.wikimedia.org/wikipedia/commons/b/b2/LAMetroLogo.svg" height="16" alt="">&nbsp;Metro (Los Angeles)” and “<img src="https://upload.wikimedia.org/wikipedia/commons/9/95/Muni_worm_logo.svg" height="16" alt="">&nbsp;Muni Metro”, both of which operate in California. A router could display a route number in an agency specific style by keying off the `network:wikidata` tag instead of performing point-in-polygon or line-in-polygon tests to distinguish one `VTA` from another.

Per https://github.com/osmlab/name-suggestion-index/pull/2863#issuecomment-509768110, we could maintain the canonical networks in a networks/ folder. Since many public transportation agencies offer multiple modalities (like bus and rail) but most are limited in geographical scope, it might make sense to structure this folder by geography, similar to editor-layer-index. There would also need to be changes to the schema and build system to support this new tree.

Editors like iD would need to handle networks a little differently than brands. Network presets shouldn’t match features based on their names, because the `name` tags on public transportation features aren’t usually supposed to refer to the network. It may make some sense to match relations based on relation types, but the `type` tag is also used on non-relations for various other purposes.

In the future, similar infrastructure could be repurposed to generate presets for road and cycle route networks.

[...]

PR #2921, opened 2019-07-28
Transit networks

quincylvania:
This is an incomplete draft addressing #2864. Feedback is encouraged as we figure out how to best implement this feature.

### DIFF
--- a/networks/north-america/us/ma/mbta.json
+++ b/networks/north-america/us/ma/mbta.json
@@ -1,0 +1,67 @@
+{
+  "public_transport/platform/bus|MBTA Bus Stop": {
+    "countryCodes": {
+        "us": {
+            "ma": true
+        }
+    },
+    "tags": {
+      "bus": "yes",
+      "highway": "bus_stop",
+      "network": "MBTA",
+      "network:wikidata": "Q6714772",
+      "operator": "MBTA",
+      "operator:wikidata": "Q171985",
+      "public_transport": "platform"
+    }
+  },
+  "public_transport/station/ferry|MBTA Boat Terminal": {
+    "countryCodes": {
+        "us": {
+            "ma": true
+        }
+    },
+    "tags": {
+      "amenity": "ferry_terminal",
+      "ferry": "yes",
+      "network": "MBTA",
+      "network:wikidata": "Q4345769",
+      "operator": "MBTA",
+      "operator:wikidata": "Q171985",
+      "public_transport": "station"
+    }
+  },
+  "public_transport/station/subway|MBTA Subway Station": {
+    "countryCodes": {
+        "us": {
+            "ma": true
+        }
+    },
+    "tags": {
+      "subway": "yes",
+      "railway": "station",
+      "network": "MBTA",
+      "network:wikidata": "Q15831678",
+      "operator": "MBTA",
+      "operator:wikidata": "Q171985",
+      "public_transport": "station"
+    }
+  },
+  "public_transport/station/train|MBTA Commuter Rail Station": {
+    "countryCodes": {
+        "us": {
+            "ma": true,
+            "ri": true
+        }
+    },
+    "tags": {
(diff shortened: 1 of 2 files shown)


networks/north-america/us/ma/mbta.json
```diff
@@ -0,0 +1,67 @@
+{
+  "public_transport/platform/bus|MBTA Bus Stop": {
```
1ec5: If we have a separate entry for each kind of feature that can be associated with the network, then there will be a lot of duplication, not just between the various entries in this file but also among each file. It would also become tedious to add support for associating a network with, say, ticket vending machines with networks after the fact.

Instead of orienting these files around presets, how about focusing on networks and indicating the applicable modes of transportation as fields on each network?

networks/north-america/us/ma/mbta.json
```diff
@@ -0,0 +1,67 @@
...
+    "countryCodes": {
+        "us": {
+            "ma": true
```
1ec5: This is a clever way to avoid duplicates and make life easier for clients. Do you think we should change the existing brand entries to also structure `countryCodes` as an object? Each country code could be set to just `true` if it applies nationally but to an object if we want to get more granular.

Alternatively, it’d be simpler for contributors (less verbose syntax) if we stick to a flat array of codes that could be [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2) codes. But that could be more difficult for a client to use Nominatim for, so maybe the build script could transform into this object structure.

networks/north-america/us/ma/mbta.json
```diff
@@ -0,0 +1,67 @@
+{
```
1ec5: What folder should interstate agencies or multi-state private carriers go into?

quincylvania: I figure it'd be similar to the [osm-community-index](https://github.com/osmlab/osm-community-index/tree/master/resources) structure where interstate carriers terminate under the country folder, like `networks/north-america/us/greyhound.json`.

I suppose I already introduced a case where a local agency has a service that crosses state lines with "MBTA Commuter Rail" also reaching Rhode Island. We'll have to decide how strict to be with the structure.

networks/north-america/us/ma/mbta.json
quincylvania: Looking into it more, I suppose using the combined codes like `ad-07` in an array might be easier for most people. My developer brain wanted to keep the country and subdivision codes separate and hierarchical but it's no big deal to parse them out later.

networks/north-america/us/ma/mbta.json
quincylvania: Could you go into more detail? Do you mean like defining "this is the network and then these are the things that can be part of this network"?

1ec5: Yes. For example, most county transportation agencies would be defined by a single entry describing the network. There would be a `bus` property set to true, and elsewhere NSI or iD would take that to mean that the network should apply to the existing Bus Stop, Bus Station, and Bus Route presets. If the same agency operates a bus rapid transit line that’s part of the same network, with ticket vending machines along the route, then setting `vending_machine` to true would automatically associate the network with that preset too.

That’s the simple case, anyways. I’m sure there are edge cases that maybe we can address by specifying overrides of individual tags.

bhousel:
Going to close this for now.  Soon I'm going to swap out the countryCodes for [location-conflation](https://github.com/ideditor/location-conflation)-style locationSets like we use on osm-community-index and imagery-index, which will let us define better regions for things like transit agencies.

The other restructuring stuff to allow `operator` or `network` to be the the defining tag (instead of `brand`) is in the back of my mind and will come eventually too.

quincylvania:
@bhousel Sounds good. This was pretty stale, I'd nearly forgotten about it.

PR #2923, opened 2019-07-29
Adding wikidata for Russian chain

bkissin:


bhousel:
Hmm these entries are already under `shop/convenience` and `shop/supermarket`..
Isn't `shop/department_store` wrong?

bkissin:
Not sure. Was looking at the index for items missing wikidata. If we don't want to add it because it's already there that's definitely OK.

bhousel:
Yeah I think these are a mistake..

It looks like they snuck in a few months ago with ec824f0, but since then they've been re-tagged and aren't in newer planet files, so they are not in `names_keep.json`.  I'm going to remove them.

PR #2929, opened 2019-08-02
Only ru is applicable

tohaklim:
Detskij Mir branding is Russia-only AFAIK


tohaklim:
Yep, but that is just a generic soviet-era name for toy shops, lit. "Kids' world".
Just checked and they have branched out to KZ, so added that

bhousel:
Nice @tohaklim thanks for double-checking 👍

PR #2930, opened 2019-08-02
Removing the most generic name

tohaklim:
ie shop


bhousel:
filtering shop is ok, but did you mean to change the filters for `amenity/restaurant` names?

tohaklim:
hmm, no, not sure how that happened. I'll do a separate pr

PR #2936, opened 2019-08-05
Move to discardKeys

tohaklim:
These should not be discouraged as are perfectly valid names (yes, as it turns out there are "Builder" branded hotels)


ENT8R:
@tohaklim can you confirm that the values are now really correct after running `npm run build` in https://github.com/osmlab/name-suggestion-index/commit/3b1588f295115fa2c505ab7f78ff5b0c2f51da8a
There is now a new entry for `shop/seafood|Рыба` (I doubt that this is correct as this translates simply to "fish") but no entry for `shop/fishing|Рыба` (which is fine and correct). Or is there really a seafood store called "Рыба"?

tohaklim:
> @tohaklim can you confirm that the values are now really correct after running `npm run build` in [3b1588f](https://github.com/osmlab/name-suggestion-index/commit/3b1588f295115fa2c505ab7f78ff5b0c2f51da8a)?
> There is now a new entry for `shop/seafood|Рыба` (I doubt that this is correct as this translates simply to "fish") but no entry for `shop/fishing|Рыба` (which is fine and correct). Or is there really a seafood store called "Рыба"?

Was trying to catch a misspelling, but it's unlikely to come up, so easier to just limit this to "^shop/fishing\\|рыболов$", removing the generic Fish name

PR #2943, opened 2019-08-07
Did the "dirty work" and added some more brands

bkissin:
Using the browser editor, added 3 new brands

### DIFF
--- a/brands/shop/erotic.json
+++ b/brands/shop/erotic.json
@@ -9,6 +9,25 @@
       "shop": "erotic"
     }
   },
+  "shop/erotic|Amazing Intimate Essentials": {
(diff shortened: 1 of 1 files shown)


brands/shop/erotic.json
```diff
@@ -9,6 +9,25 @@
       "shop": "erotic"
     }
   },
+  "shop/erotic|Amazing Intimate Essentials": {
```
bhousel: name here should match `name` tag below... and should be whatever these stores are signed as.

bkissin: Thanks @bhousel! I took care of that.

bhousel:
Thanks @bkissin 👍

Issue #2932, opened 2019-08-03
Red Cross Blood Service issues


tastrax:
Red Cross and the blood bank are actually separate parts of the same organisation and have different logos and websites

https://www.facebook.com/redcrossbloodau/
https://www.donateblood.com.au/

PR #2944, opened 2019-08-08
Change AU Red Cross Blood Service Wiki tags

mds08011:
As identified in #2932 the Australian Red Cross Blood Service is unique enough to be considered its own brand.  This PR changes the Wikidata and Wikipedia tags of the existing entry to refer to the AU Red Cross **Blood Service** specifically.


bhousel:
Thanks @mds08011 👍

PR #2947, opened 2019-08-08
Not a brand

tohaklim:
This literally translates as "Home use chemicals"


tohaklim:
Followed by [this filter PR](https://github.com/osmlab/name-suggestion-index/pull/2948)

PR #2949, opened 2019-08-08
Fix spelling of Ŝkoda

sb12:


tohaklim:
I wonder if this needs a matchname with the basic S to match Skoda spelling

bhousel:
> I wonder if this needs a matchname with the basic S to match Skoda spelling

The matcher _should_ catch it..

Names are compared with diacritic marks folded, (but we've seen cases where people mixup Latin and Cyrillic characters and it won't catch that.)

brands/shop/car.json
```diff
@@ -253,12 +253,12 @@
       "shop": "car"
     }
   },
-  "shop/car|Skoda": {
+  "shop/car|Ŝkoda": {
```
ENT8R: @sb12 I did a short research and every source that I found (official website of the company in multiple languages and Wikipedia) are suggesting that the right spelling would be `Škoda` instead of `Ŝkoda`. Are you really sure, that the current spelling is correct?

sb12: Of course, you are right. 
As I am not familiar with Czech language I didn't notice the difference between Škoda and Ŝkoda.
Thanks for spotting it!

bhousel:
done in #2952

PR #2950, opened 2019-08-08
EvrMarket not a supermarket

tohaklim:
Already included under convenience


bhousel:
thanks @tohaklim !
In this situation that means that the _other_ more correct entry needs a `matchTags` property, so the build script doesn't keep trying to put the extra `shop/supermarket` entry back.   I've added it in e40504502.

Issue #2847, opened 2019-07-04
Handling of Combined Brands


mparrault:
There are a lot of cases where two brands (usually commonly owned) combine in one store.  Examples I am familiar with include KFC/Taco Bell, Wendy's/Tim Hortons, Carter's / OshKosh.  I would estimate there at least 50 of each of these combos.  Sometimes the two brands have separate cash registers, but they often don't.  Should we create new entries for the combined brands?  Should the two brands be separated into two nodes?

I think it should depend on the case, but some combined brands should have entries.  For example, Wendy's/Tim Hortons should be split into separate nodes (since they have separate cashes), and Carter's / OshKosh should have a new entry.  KFC/Taco Bell could have a new entry (because the two brands share a cash), but splitting might make more sense.

1ec5:
Another reason to combine some locations is that one always comes first and the other second in signage, menu layout, and how the store is referred to. You’d want that order to be preserved even if you rotate the map.

Adding entries for the more common combination seems reasonable. Alternatively, mappers can already add two points in iD and combine them, which turns the name and brand tags into semicolon-delimited lists. Admittedly this workaround isn’t particularly discoverable.

1ec5:
Car dealerships are often affiliated with multiple brands as well. I don’t think it would be very practical to include all possible combinations of car brands, but we could include some common combinations like Chrysler/Jeep/Dodge, BMW/Mini, Jaguar/Land Rover, etc.

bhousel:
This is a really hard problem...

The index already does [filter out "KFC/Taco Bell"](https://github.com/osmlab/name-suggestion-index/blob/67d845e7ee67731cf8690edf2d350c6edf67a8a8/config/filters.json#L46), which is the most common instance in OSM currently.

I'm a big proponent of doing the simplest thing possible, which in this case is _just map 2 pins_ - one for KFC and one for Taco Bell.  A great thing about OSM is that we can map the world in as much detail as the contributors want to.  However it becomes a real problem for software (like name suggestion index) when people use creative tagging that breaks the one-feature-one-element rule.  I really don't want to maintain an index (or an editor for that matter) that needs to deal with semicolon-delimited names/brands/cuisines/ and everything else.

I took a look at what other maps do <sup>†</sup>, and none of them seem to handle the "combined brand" case at all.

### Google:
[...]

PR #2955, opened 2019-08-09
Adding Chrysler, Proton, Saipa

bkissin:


brands/shop/car.json
```diff
@@ -340,6 +359,18 @@
       "shop": "car"
     }
   },
+  "shop/car|سایپا": {
+    "countryCodes": ["ir"]
```
mds08011:
```suggestion
    "countryCodes": ["ir"],
```

mds08011:
Thank you!

Adamant36:
Are there actually "Chrysler" stores out there? I know there's places that sell sell Chryslers, but do they have their own self branded stores just called "Chrysler"?

1ec5:
It’s the same as with any other American car brand that has a dealership network. Chrysler is usually combined with other brands owned by the company (Jeep, Dodge, and Ram), but we have entries for those brands too.

[](https://commons.wikimedia.org/wiki/File:JEEP_Akita.jpg)

Adamant36:
It's not really the name of the car dealership though (in this case I assume it would be "Something Something Chrysler Jeep Whatever"). I feel like in cases like this there should be a `carbrand=*` tag that's used instead and then stores are tagged with the local name. It shouldn't be assumed that all car dealerships with "Chrysler" signs out front are corporate though or shouldn't be tagged with the local dealerships name. Which you usually can't tell just with drive by sign tagging.

bkissin:
I definitely get where @Adamant36 is coming from, which is why I was hesitant to add the other brands under the same umbrella corporation. With the exception of VW group brands (VW, SEAT, Skoda, Audi, etc.) many multi-brand automakers (partially because they often group brands together at dealerships) are not listed in the index. I don't know whether it would make sense to create individual entries for Dodge, Jeep, etc. unless there was an option to add multiple brands under the brand= section.

That being said, there is certainly a way to tag the brand, but keep the local name. Usually when I'm in iD, I'll update the tag, and then change the name from "Rimouski Honda" to "Honda" in order to activate the brand info, and then go back into the name section and re-add the local name information.

bhousel:
per #2847 I'd prefer to avoid combining brands in the index for now.  

It's totally ok if mappers want to add several points and either leave them separate or merge them together (creating semicolon-delimited values),  or if mappers want to adjust the name tag to say "Whatever Honda".  iD supports both of these things.

1ec5:
Car brands were one of the original use cases for the `brand` tag. The dealerships I have in mind are the ones whose monument signs are emblazoned with the Chrysler, Jeep, and Dodge logos and nothing else. I’m not referring to, say, used car dealerships that happen to carry Chrysler cars but are unaffiliated with that company. Whether the dealership is corporate or franchised is a concern for `operator`, not `brand`.

It is the case that car dealerships usually have something else in the name, signage notwithstanding. We should encourage mappers to add this detail, but location-specific names are never by themselves a reason to omit entries from this index.

PR #2956, opened 2019-08-10
Fix name tag of Bridgehead

mparrault:
This fixes an error I made in a previous pull request.


Adamant36:
Sure. Thanks for fixing it 👍

PR #2961, opened 2019-08-12
Velcom rebranding

tohaklim:
Now A1 nationwide, signage being replaced. 
Added name:en to differentiate from other A1 in the EU

### DIFF
--- a/brands/office/telecommunication.json
+++ b/brands/office/telecommunication.json
@@ -10,11 +10,13 @@
   },
   "office/telecommunication|Velcom": {
     "countryCodes": ["by"],
+    "matchNames": ["А1", "Велком", "velcom | A1", "velcom"],
     "tags": {
-      "brand": "velcom | A1",
+      "brand": "A1",
       "brand:wikidata": "Q1888809",
-      "brand:wikipedia": "en:Velcom",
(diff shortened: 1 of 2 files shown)


bhousel:
I don't understand this rebranding.. 
If the new A1 Belarus is a subsidiary of the "other A1 in Europe", can we just use that one instead?

https://en.wikipedia.org/wiki/A1_Telekom_Austria_Group 
https://www.wikidata.org/wiki/Q688755

Meaning - can I just make a A1 entry that has those things, and operates in a bunch of countries (including `by`) and matches the older "velcom" names?

tohaklim:
I am not sure if that would work universally considering the only thing they have in common is branding. This is a separate legal entity, but same brand they seem to be using in a couple of countries. 
I don’t think I’ve seen the other A1 entries in the index though?

bhousel:
> am not sure if that would work universally considering the only thing they have in common is branding. This is a separate legal entity, but same brand they seem to be using in a couple of countries.

It seems ok - this project is really more an index of brands not legal entities.

tohaklim:
Thanks for implementing that!
I am wondering if the local sub Wikipedia page should be inserted somewhere, or just the wiki data entries linked

PR #2962, opened 2019-08-12
Add Wiki tags to four brands

mds08011:
Adds Wikipedia and Wikidata tags to four brands from https://github.com/osmlab/name-suggestion-index/commit/6d76a546dc38764a4815aa9266689aa73787267e


bhousel:
Thanks @mds08011 👍

PR #2969, opened 2019-08-16
Add Tostao' cafe brand

quincylvania:
https://www.wikidata.org/wiki/Q60632476
http://tostaocafeypan.com/

The name stylization  is a bit funky. The name on the website is `TOSTAO’ Café & Pan`, so I used that for `official_name`. I'm not sure what the index standards are for all-caps and apostrophes (`'` vs. `’`), but `Tostao’` seemed more reasonable for the `brand` and `name`.

### DIFF
--- a/brands/amenity/cafe.json
+++ b/brands/amenity/cafe.json
@@ -802,6 +802,19 @@
       "takeaway": "yes"
     }
   },
+  "amenity/cafe|Tostao’": {
+    "countryCodes": ["co"],
+    "tags": {
+      "amenity": "cafe",
+      "brand": "Tostao’",
+      "brand:wikidata": "Q60632476",
+      "cuisine": "coffee_shop",
+      "name": "Tostao’",
+      "name:es": "Tostao’",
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks!  I usually try to avoid curly apostrophes, but there may be a few in the index.
Also the matching code now ignores punctuation so we don't need to specifically match the different variants anymore..

1ec5:
Someday I’ll convince OSM to embrace curly apostrophes. 😉

In case anyone sees this and decides to straighten out apostrophes throughout the index, note that some `wikipedia` tags need to retain curly apostrophes due to the actual page names.

/ref https://github.com/osmlab/name-suggestion-index/commit/074e98125bc857e00c286ffb20c61132d9d31f38#commitcomment-33896996

bhousel:
> In case anyone sees this and decides to straighten out apostrophes throughout the index, note that some `wikipedia` tags need to retain curly apostrophes due to the actual page names.

Ok.. In this case I left them in anyway.  I have no idea how they would look to a native Columbian speaker.

PR #2971, opened 2019-08-16
Various Mall Brands

mparrault:
-Stokes, Think Kitchen, Charm Diamond Centres, and Coles
I also added various nomatches for other brands with "Coles" in the name.


Adamant36:
Thanks.

PR #2972, opened 2019-08-18
Kampgrounds of America

1ec5:
Added the KOA Kampground, KOA Journey, KOA Holiday, and KOA Resort campground chains, plus the Kamp K9 dog park found at each campsite.

This PR introduces the following warnings because each of the four campground formats has a `short_name` of `KOA`, which I think is appropriate in each case. I couldn’t suppress these warnings using `nomatch`, but maybe that’s a bug.

```
  "tourism/caravan_site|KOA Holiday" -> matches? -> "tourism/caravan_site|KOA Journey ("koa")"
  "tourism/caravan_site|KOA Holiday" -> matches? -> "tourism/caravan_site|KOA Journey ("koa")"
  "tourism/caravan_site|KOA Holiday" -> matches? -> "tourism/caravan_site|KOA Journey ("koa")"
  "tourism/caravan_site|KOA Holiday" -> matches? -> "tourism/caravan_site|KOA Kampground ("koa")"
  "tourism/caravan_site|KOA Holiday" -> matches? -> "tourism/caravan_site|KOA Kampground ("koa")"
  "tourism/caravan_site|KOA Holiday" -> matches? -> "tourism/caravan_site|KOA Kampground ("koa")"
  "tourism/caravan_site|KOA Holiday" -> matches? -> "tourism/caravan_site|KOA Resort ("koa")"
  "tourism/caravan_site|KOA Holiday" -> matches? -> "tourism/caravan_site|KOA Resort ("koa")"
  "tourism/caravan_site|KOA Holiday" -> matches? -> "tourism/caravan_site|KOA Resort ("koa")"
```

### DIFF
--- a/brands/leisure/dog_park.json
+++ b/brands/leisure/dog_park.json
@@ -1,0 +1,13 @@
+{
+  "leisure/dog_park|Kamp K9": {
+    "countryCodes": ["ca", "us"],
+    "matchTags": ["leisure/park"],
+    "tags": {
+      "brand": "KOA",
+      "brand:wikidata": "Q16988635",
+      "brand:wikipedia": "en:Kampgrounds of America",
+      "leisure": "dog_park",
+      "name": "Kamp K9"
+    }
+  }
+}
--- a/brands/tourism/caravan_site.json
+++ b/brands/tourism/caravan_site.json
@@ -1,0 +1,64 @@
+{
+  "tourism/caravan_site|KOA Holiday": {
+    "countryCodes": ["ca", "us"],
+    "matchTags": [
+      "leisure/park",
+      "tourism/camp_site"
+    ],
+    "tags": {
+      "brand": "KOA",
+      "brand:wikidata": "Q16988635",
+      "brand:wikipedia": "en:Kampgrounds of America",
+      "name": "KOA Holiday",
+      "short_name": "KOA",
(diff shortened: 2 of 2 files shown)


bhousel:
I guess I should change the matcher to not index ambiguous secondary names.

We definitely want to see a warning if two brands try to index the same `name` but maybe not if they have the same `short_name`/`alt_name` etc.  In that situation we could just not index them.  That has the downside that iD won't be able to suggest a replacement if it sees `name=KOA` but that's alright.

1ec5:
Maybe in that case we could specify “KOA” as a matching name to ensure it matches a particular entry (in this case, “KOA Kampground”)?

bhousel:
> I guess I should change the matcher to not index ambiguous secondary names.

I did this in e639c2797 !

> Maybe in that case we could specify “KOA” as a matching name to ensure it matches a particular entry (in this case, “KOA Kampground”)?

I think so.. The code now differentiates between primary names (`name`, `official_name`, and whatever strings appear in `matchNames`) and secondary names (`alt_name`, `short_name`).  Primary name collisions will trigger a warning and secondary name collisions will not.

bhousel:
I'm just going to hit the green merge button and see what happens 🤞

PR #2975, opened 2019-08-19
Remove duplicate of Sinopec and join both

danielsjf:


bhousel:
Thanks @danielsjf!  We try to provide as many variations in brand entries as there are variations in on-the-ground signage.

So if the real world locations of this brand always say "中国石化" then this consolidation is ok.  If they sometimes say "中国石化" and sometimes say "Sinopec" then we should keep them as 2 different entries.  I'd love to hear thoughts from someone with local knowledge of how this brand operates in China?

danielsjf:
I've only ever seen them with both names together as can be seen on this picture: https://img.caixin.com/2018-07-25/1532521097999679.jpg

However, after Googling I can also find versions with just one of both names. Does this mean we actually have to create all three of can we go for one version with both names as the default.

bhousel:
> However, after Googling I can also find versions with just one of both names. Does this mean we actually have to create all three of can we go for one version with both names as the default.

I usually leave it up to what local mappers want to do.

danielsjf:
Since the combination seems by far the most common, I propose to change it to that one (this is from experience in multiple provinces in China).

bhousel:
> Since the combination seems by far the most common, I propose to change it to that one (this is from experience in multiple provinces in China).

Sounds good to me.. thanks!

PR #2978, opened 2019-08-19
Remove `recycling_type` from Salvation Army

mds08011:
Removes `recycling_type=container` from The Salvation Army entry so that it can be used for both `recycling_type=container` and `recycling_type=centre`


bhousel:
Thanks @mds08011 👍

Issue #2966, opened 2019-08-15
Add Londis (UK)


doublah:
**Brand Name**
Londis

**Brand Wikidata (if there is one)**
https://www.wikidata.org/wiki/Q21008564

**Brand Wikipedia (if there is one)**
https://en.wikipedia.org/wiki/Londis_(United_Kingdom)

**OSM Tags**
shop=convenience

**Country Codes (if any)**
gb

**Anything else?**
Londis also exists in Ireland but it's seperately operated and owned by a different company (see https://en.wikipedia.org/wiki/Londis_(Ireland) & https://www.wikidata.org/wiki/Q21015800), so keeping it to UK only is important.

mds08011:
We have Londis entries for both `shop=convenience` and `shop=supermarket`.  Does Londis operate both types of stores, or do you think `shop=supermarket` is a mistake?

mds08011:
Both the UK and Ireland Londis were added with the appropriate country codes: https://github.com/osmlab/name-suggestion-index/commit/552f35d4b0124a50aed043ed055f2380291245cd , https://github.com/osmlab/name-suggestion-index/commit/4dc33c8e1c5d4ca6498b2fc75d0f7ed491d9093f , https://github.com/osmlab/name-suggestion-index/commit/768838811fac98eb6f470331166595e98791b0e5

The `shop=supermarket` entry was removed: https://github.com/osmlab/name-suggestion-index/commit/09ed685402ff61313296c913d9ff08c3b2688588

PR #2980, opened 2019-08-21
Add Londis (Ireland)

mds08011:
Adds Londis (Ireland) and adds Wikidata and Wikipedia tags to the existing Londis (UK) entry.  Per: #2966

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -828,6 +828,8 @@
     "countryCodes": ["gb"],
     "tags": {
       "brand": "Londis",
+      "brand:wikidata": "Q21008564",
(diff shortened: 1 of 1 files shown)


mds08011:
Are the `~Ireland` and `~UK` needed or are the country codes sufficient to prevent these brands from getting confused?  I added them because I was seeing warnings about duplicate entries.

bhousel:
> Are the `~Ireland` and `~UK` needed or are the country codes sufficient to prevent these brands from getting confused? I added them because I was seeing warnings about duplicate entries.

Yes, for now it's required to prevent the duplicate warnings..
Looks good, thanks!

PR #2985, opened 2019-08-22
Add French Banque Populaire and disambification for Banque Populaire du Maroc

bmillemathias:


bhousel:
thanks @bmillemathias , looks good 👍

PR #2986, opened 2019-08-23
Add CZ countyCode for Czech branch

lgommans:
Q24282825 is only for the bank's Czech branch. I was prompted to upgrade the tags in iD for a German bank and initially accepted: it doesn't show the tags it wants to add beforehand, so I didn't see that the tags were wrong. Then I edited its other tags and noticed it now referred to a Wikipedia article about the Czech branch, only available in Czech. I'm not sure how this error could slip in and if maybe Wikidata is wrong, but this solution was suggested to me in the OpenStreetMap Telegram chat, and sounds like it would at least be a workaround.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -4502,7 +4502,8 @@
       "brand:wikidata": "Q24282825",
       "brand:wikipedia": "cs:Raiffeisenbank",
       "name": "Raiffeisenbank"
-    }
+    },
+    "countryCodes": ["cz"]
   },
   "amenity/bank|Regions Bank": {
     "countryCodes": ["us"],


claudiush:
...and there's also Q22093483 for Austria and actually a bunch of Raiffeisen subsidiaries across central and eastern european countries, see the list here: https://de.wikipedia.org/wiki/Raiffeisen_Bank_International#T%C3%A4tigkeit

bhousel:
Yep, I'm looking at this list now:  https://en.wikipedia.org/wiki/Raiffeisenbank

When the situation gets very ambiguous we can add text after each brand like "Raiffeisenbank~(Czech Republic)" or "Raiffeisenbank~(Bulgaria)", and that's probably what I will end up doing.  Any brands with a disambiguation string like that will be excluded from iD's validator, so users won't be prompted to do an upgrade.

PR #2987, opened 2019-08-25
fix wikipedia reference for Swiss "Landi" shops

das-g:
Link to https://de.wikipedia.org/wiki/Landi_(Unternehmen) instead of the non-existing https://en.wikipedia.org/wiki/Landi_(Unternehmen)

(As of now, there seems to be no English Wikipedia article about the company.)


tas50:
Thanks for the fix

PR #2993, opened 2019-09-05
add pub "Ezītis miglā"

richlv:


mds08011:
Will the name be `name=Ezītis miglā` for all locations?  The website seems to include locations in the name.  For example, would the Ezītis miglā on Palasta street use `name=Ezītis miglā` or `name=Ezītis miglā Palasta Ielā`?

richlv:
Yes, the official and signed name is always "Ezītis miglā" - see a few examples here:

https://www.mapillary.com/map/im/Bi5-UffW3unsLPwbW6lsJQ
https://www.mapillary.com/map/im/vxaky1ul7XFhOrGPzYsz4Q
https://www.mapillary.com/map/im/1RakvBmx1vfJGYTE5-pcPA

Don't have full name on file for this one, but it can be deduced to be the same from these photos:
https://www.mapillary.com/map/im/FZ7I-Ohv_EO6fyDLZbwqPA
https://www.mapillary.com/map/im/HYsi89zouCJPqLsz94CIgQ

While the way to identify a particular location is to use a nearby street name, that can be thought as an alt_name (to be added manually as needed), like "Laboratorijas ezītis", "Hospitāļu ezītis" and the like (note that title case is not used in Latvian).

You can also think of it as place identifier for supermarkets like Walmart or Rimi - the name is still "Walmart" or "Rimi", we don't add the street name in the main name tag.

mds08011:
That makes sense, thanks!  Good to know about the title case as well.

richlv:
Great, thank you for the sanity check and merging :)

PR #2999, opened 2019-09-08
Added brand:wikidata for 101 Drogerie

RickeyRichards:
Hey,
I just added brand:wikidata attribute for 101 Drogerie.
Did I do it right?
And how long does it takes, to see this brand change in OpenStreetMap?
Thanks!


mds08011:
@RickeyRichards welcome and thank you!  Your entry looks good.

The NSI gets packaged up and incorporated into the editor on openstreetmap.org (called iD) at periodic intervals (typically monthly). You can check the “releases” page to see when this last happened (13 days ago): https://github.com/osmlab/name-suggestion-index/releases

So around the end of the month this should go live on openstreetmap.org

Issue #3000, opened 2019-09-09
Supermarkets in Iceland


Stalfur:
Netto was branded with the foreign chain, www.netto.is for context.

Supermarket Iceland in Iceland (yes they recently opened) is a subsidiary of the British chain but should be its own entity.

1ec5:
> Netto was branded with the foreign chain, www.netto.is for context.

Fixed in #3001 by adding a Nettó entry and limiting the existing Netto entry to specific countries.

> Supermarket Iceland in Iceland (yes they recently opened) is a subsidiary of the British chain but should be its own entity.

Are you referring to [this Iceland chain](https://en.wikipedia.org/wiki/Iceland_%28supermarket%29) opening locations in Iceland? The index tracks brands, but not all brands are subsidiaries and not all subsidiaries have their own brands. Does Iceland’s Iceland subsidiary use unique branding or anything that would require tags that differ from the existing Iceland entry?

https://github.com/osmlab/name-suggestion-index/blob/8902efdb752d703c70ed8f0b7775be41cd29697f/brands/shop/frozen_food.json#L17-L26

mds08011:
The Iceland chain in Iceland uses a different website from the UK locations and has a different owner but it's not immediately clear whether the branding is unique: https://icelandbudir.is/

PR #3001, opened 2019-09-10
Icelandic supermarket chains

1ec5:
Added a few Icelandic supermarket chains based on [this list](https://en.wikipedia.org/wiki/List_of_supermarket_chains_in_Iceland). Differentiated between the three Netto (Nettó) chains.

/ref #3000

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -1,4 +1,14 @@
 {
+  "shop/convenience|10-11": {
+    "countryCodes": ["is"],
+    "tags": {
+      "brand": "10-11",
+      "brand:wikidata": "Q65336720",
+      "brand:wikipedia": "is:10-11",
+      "name": "10-11",
+      "shop": "convenience"
+    }
+  },
   "shop/convenience|1st Stop": {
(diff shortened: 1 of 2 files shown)


1ec5:
The validator complains about Nettó matching Netto:

```
  "shop/supermarket|Netto~(Salling)" -> matches? -> "shop/supermarket|Nettó"
```

It’s somewhat unfortunate that diacritic folding would force us to introduce a disambiguating suffix. It’s unlikely that Nettó would be spelled Netto in Iceland. A similar situation would arise between the chain Bónus (in Icelandic) and any future chain named “Bonus” (in another language like English).

bhousel:
> It’s somewhat unfortunate that diacritic folding would force us to introduce a disambiguating suffix.

Yes I see your point, and it has happened a few times where the diacritic folding causes an issue with the matcher.  However I don't view disambiguating suffixes as bad really - it lets mappers to know to be extra careful when working with these, and it excludes these entries from being upgraded by the iD validator.  I think it's pretty ok to take a "when in doubt, be conservative" approach to them.

PR #3010, opened 2019-09-17
Fix Souyuzpechat wikidata refs

tohaklim:
No relation between the Soviet brand and modern entities in BY to keep the old wikidata page


bhousel:
Thank you @tohaklim! 👍

PR #3014, opened 2019-09-17
Move Ross, TJ Maxx, TK Maxx to department_store

### LINKED ISSUE #3006: Move TJ Maxx to shop=department_store?
The NSI currently lists TJ Maxx as `shop=clothes`.  However, [Wikipedia](https://en.wikipedia.org/wiki/TJ_Maxx) and the [OSM Wiki](https://wiki.openstreetmap.org/wiki/Tag:shop%3Ddepartment_store) specifically refer to TJ Maxx as a "department store".

What are people's thoughts on `shop=department_store` vs `shop=clothes`?

https://github.com/osmlab/name-suggestion-index/blob/a052e67a011f9820f1a15e370b6e172b3f16a9ae/brands/shop/clothes.json#L2763-L2772

mds08011:
Moves Ross, TJ Maxx, and TK Maxx from `shop=clothes` to `shop=department_store`

Resolves #3006

### DIFF
--- a/brands/shop/clothes.json
+++ b/brands/shop/clothes.json
@@ -2750,15 +2750,6 @@
       "shop": "clothes"
     }
   },
-  "shop/clothes|TK Maxx": {
-    "tags": {
-      "brand": "TK Maxx",
-      "brand:wikidata": "Q23823668",
-      "brand:wikipedia": "en:TK Maxx",
-      "name": "TK Maxx",
-      "shop": "clothes"
-    }
-  },
   "shop/clothes|Takko": {
     "countryCodes": [
       "at",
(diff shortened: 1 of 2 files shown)


tas50:
I think it's a stretch to call Ross a department store, but it's referred to as a department store in places like Wikipedia so 👍

mds08011:
I agree but it also sells much more than just clothes 🤷‍♂

1ec5:
`department_store=discount` 🤷‍♂

tas50:
Sometimes you just have to find the thing that fits best. We tag walmart as supermarkets 🤷‍♂

1ec5:
That’s a bit of a counterexample actually. Walmarts and other hypermarkets are tagged as supermarkets because of what they primarily sell, but these discount department stores are being tagged as department stores because of what they happen to also sell. By the same reasoning as Walmart, we’d be tagging them as clothing stores.

PR #3016, opened 2019-09-17
Charging Stations in Luxembourg

DavidMoraisFerreira:
Hello,

I've added "Chargy" Charging Stations in Luxembourg, added the availability of Bike-Energy in Luxembourg and improved the generic name of the Bike-Energy brand.

Regards
David


bhousel:
Thanks!   In 402d611 I also added a `matchName` property, so that the index would recognize the previous name.    (without that, the scripts would try to create another entry for the old name)

PR #3019, opened 2019-09-18
Various banks and Aral in Luxembourg

DavidMoraisFerreira:
Hello
I've added banks in Luxembourg. Aral is also available here :)

Greetings
David


bhousel:
Thanks @DavidMoraisFerreira!

PR #3020, opened 2019-09-18
Added CTT Portugal

United600:


bhousel:
Thanks @United600 and welcome!

PR #3024, opened 2019-09-20
Add F. Hinds

doublah:
Add F Hinds jewellery shop.


SilentSpike:
Should this be `F.Hinds` as per [their website](https://www.fhinds.co.uk)?

I think I changed H.Samuel at some point to match their official branding the same way.

bhousel:
Thanks, I merged and dropped the space in 5b195c2b2 to keep it like how the brand spells it.

PR #3026, opened 2019-09-20
Update japanese brands

sguinetti:
Greetings. This update adds some brands from Japan and update the information of some pharmacy stores.


bhousel:
Thanks for these @sguinetti  - looks great! 👍

PR #3027, opened 2019-09-21
Add three Malian Banks

mds08011:
Adds three banks local to Mali


tas50:
Thanks as always @mds08011

PR #3033, opened 2019-09-23
Add supermarket and bank brands

sguinetti:
Several brands, including some additions of existing tags. By the way, there is a build in the Bank section, I apologize any inconvenience that may occur.


bhousel:
looks good to me, thanks @sguinetti  👍

PR #3034, opened 2019-09-23
Add matchname for usmile

danielsjf:
Ironically, uSmile doesn't match uSmile

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -1722,7 +1722,7 @@
   },
   "shop/convenience|uSmile 昆仑好": {
     "countryCodes": ["cn"],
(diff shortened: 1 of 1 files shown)


brands/shop/convenience.json
```diff
@@ -1722,7 +1722,7 @@
   },
   "shop/convenience|uSmile 昆仑好": {
     "countryCodes": ["cn"],
-    "matchNames": ["uSmile 便利店", "昆仑好"],
+    "matchNames": ["uSmile 便利店", "昆仑好", "uSmile"],
```
tas50: These can just be lower case

danielsjf: Fixed

tas50:
Thanks @danielsjf

PR #3035, opened 2019-09-24
Add Wings Financial Credit Union and edit other entries

TishG:


tas50:
@TishG Thanks for the fix, but you don't actually want to modify the dist file directly. That's the file that is built and shipped with editors. The file you want to edit is here: https://github.com/osmlab/name-suggestion-index/blob/master/brands/amenity/bank.json

TishG:
My apologies, I didn't mean to do that. I will go back and make those edits in the correct file.

PR #3037, opened 2019-09-25
Add Ixina kitchen shop

bmillemathias:
Update also the list of coutries based on http://www.ixina.com/find-a-store/


bhousel:
Thanks @bmillemathias  👍

PR #3040, opened 2019-09-27
Furniture and pastry shops in LU, Fixed broken build

DavidMoraisFerreira:
:)


tas50:
Thanks @DavidMoraisFerreira

PR #3041, opened 2019-09-27
Update Pret a Manger

doublah:
Adds more country codes and short_name.


tas50:
@doublah theres a few formatting issues with this that will get cleared up if you run ‘npm run build’ in the repo

SilentSpike:
The `matchName` is redundant since it's in the `short_name` tag already (this will automatically become a matching name)

bhousel:
Thanks @doublah  👍

PR #3044, opened 2019-09-28
Add supermarket and bank brands (part 2)

sguinetti:
This updates complements the addition of brands in Japan. It seems that these are the most relevant brands in the country, specifically Tokyo. I added other brands like book stores and restaurants.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -7609,6 +7609,20 @@
       "name:en": "China Merchants Bank"
     }
   },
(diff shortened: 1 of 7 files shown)


tas50:
Thanks @sguinetti

tohaklim:
Are 7170 and 7185 lines supposed to be both referring to SMBC?

bhousel:
Good catch @tohaklim , I just fixed that `name:en` in 75a8fb208

PR #3048, opened 2019-09-30
add cash withdrawal tags to Rewe supermarket

Discostu36:
* [The new cash_withdrawal tag has been approved](https://wiki.openstreetmap.org/wiki/Proposed_features/Cash_withdrawal)
* [Every Rewe supermarket has cash withdrawal under the same conditions](https://www.rewe.de/service/bargeld-abheben/)


bhousel:
Sorry but I'm going to have to say no to this..  I'm not sure that I'm ready to have tags like this in the NSI.  Let's just get all the Rewes of the world mapped first before we start thinking about tagging what kind of cash transactions are allowed in them.

PR #3049, opened 2019-10-01
Adding cineplex canada

vphilot:
Addded Cineplex Canada to brands list

### DIFF
--- a/brands/amenity/cinema.json
+++ b/brands/amenity/cinema.json
@@ -173,8 +173,7 @@
       "cineplex entertainment",
       "cineplex odeon",
       "cineplex theatre",
-      "cineplex theatres",
-      "galaxy cinemas"
+      "cineplex theatres"
     ],
     "tags": {
       "amenity": "cinema",
--- a/brands/amenity/cinema.json
+++ b/brands/amenity/cinema.json
@@ -165,7 +165,7 @@
       "name": "Cineplanet"
     }
   },
-  "amenity/cinema|Cineplex Entertainment": {
+  "amenity/cinema|Cineplex Entertainment": { 
     "countryCodes": ["ca"],
     "matchNames": [
       "cineplex cinema",
--- a/brands/amenity/cinema.json
+++ b/brands/amenity/cinema.json
@@ -181,7 +181,7 @@
       "brand": "Cineplex",
       "brand:wikidata": "Q2972807",
       "brand:wikipedia": "en:Cineplex Entertainment",
-      "name": "AMC"
+      "name": "Cineplex"
     }
   },
   "amenity/cinema|Cineplexx": {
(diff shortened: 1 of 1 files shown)


brands/amenity/cinema.json
```diff
@@ -165,6 +165,25 @@
...
+      "cineplex theatre",
+      "cineplex theatres",
+      "galaxy cinemas"
+    ],
+    "tags": {
+      "amenity": "cinema",
+      "brand": "Cineplex",
+      "brand:wikidata": "Q2972807",
+      "brand:wikipedia": "en:Cineplex Entertainment",
+      "name": "AMC"
```
bhousel: Should this be changed to Cineplex also?

brands/amenity/cinema.json
```diff
@@ -165,6 +165,25 @@
       "name": "Cineplanet"
     }
   },
+  "amenity/cinema|Cineplex Entertainment": {
```
bhousel: This should match the `name` tag..

vphilot: @bhousel resolved!

brands/amenity/cinema.json
```diff
@@ -165,6 +165,25 @@
...
+  "amenity/cinema|Cineplex Entertainment": { 
+    "countryCodes": ["ca"],
+    "matchNames": [
+      "cineplex cinema",
+      "cineplex cinemas",
+      "cineplex entertainment",
+      "cineplex odeon",
+      "cineplex theatre",
+      "cineplex theatres",
+      "galaxy cinemas"
```
bhousel: Adding "Galaxy Cinemas" here means that iD will prompt users to upgrade those features to "Cineplex".  Is this what you want?    (If not, they need to be 2 separate entries).

brands/amenity/cinema.json
bhousel: It should be "Cineplex" right?   not "Cineplex Entertainment" ?

brands/amenity/cinema.json
vphilot: This is an old name for the company. Thought of adding as a match just in case. Does it make sense?

bhousel: I don't know whether it makes sense or not.  `matchNames` is for alternate names that are less preferred. 

iD will encourage users to upgrade them to the preferred name.  If the theater still operates some locations as the "Galaxy" brand, then this would be wrong.  But if they have changed all the locations to have "Cineplex" signs, then it would be correct.

vphilot: @bhousel  thanks for clarifying. It makes more sense to remove it then.

brands/amenity/cinema.json
vphilot: @bhousel Cinepelx Entertainment is the official company name

bhousel:
Thanks @vphilot 👍

Issue #2679, opened 2019-05-27
Split Danske Bank


Adamant36:
It appears that Danske Bank is two different companies. This [Dankse Bank](https://en.wikipedia.org/wiki/Danske_Bank_(Northern_Ireland)) is in Ireland. Whereas, the index's entry is for a bank in Denmark. Although, both come up in OverpassTurbo. So they should probably be split into separate entries.

ENT8R:
I don't think these are separate brands: The wikipedia article of [Danske Bank](https://en.wikipedia.org/wiki/Danske_Bank) says:
> The Danske Bank group operates a number of local banks around the Nordic Region as well as across Ireland.

Also judging from the logo, I think this is actually the same. Though it might be nevertheless an option to add an additional entry for the bank in Northern Ireland because wikipedia also says:
> Danske Bank UK is a standalone business unit within the Danske Bank Group and operates under a UK banking licence. 

"[Danske Bank](https://en.wikipedia.org/wiki/Danske_Bank)" is the parent company of "[Danske Bank (Northern Ireland)](https://en.wikipedia.org/wiki/Danske_Bank_(Northern_Ireland))"

Adamant36:
"Though it might be nevertheless an option to add an additional entry for the bank in Northern Ireland"

That's what I was thinking since they have different Wikipedia articles and people in Ireland would mainly want information about the banks history there. As compared to the other article about the parent company that's mainly information about them in another country.

bhousel:
Yeah I agree with @ENT8R  - they aren't separate brands.

PR #3050, opened 2019-10-01
Split Danske Bank into Danske Bank and Danske Bank (Northern Ireland)

PavlosIsaris:
Fixed issue [https://github.com/osmlab/name-suggestion-index/issues/2679](https://github.com/osmlab/name-suggestion-index/issues/2679) by adding missing amenity.


bhousel:
oh - just seeing #2679 now..  They look like they are the same bank.. Let's not do this - sorry.

PR #3051, opened 2019-10-01
Adding universities in toronto

vphilot:
Added 2 universities in Toronto


bhousel:

Issue #2883, opened 2019-07-13
Add support for templated/generated categories - use this for ATMs, post_boxes, etc.


yoasif:
I recently ran into a free-standing Chase ATM - you can find a lot here: https://locator.chase.com/search?q=New+York%2C+NY%2C+USA&l=en&type=ATM

When I type in Chase in iD, I was presented with an entry for branch locations, but none for ATMs.

There should be a new entry specifically for free standing ATMs.

bhousel:
Hmm we don't currently do ATMs in this index - maybe we should!  
I'm curious what people think, and also want to take a look at with other maps do (Google/Apple/Bing)

1ec5:
Yes, we should have entries for ATMs, particularly for banks that have freestanding ATM locations (whether drive-through or in a mall or supermarket). We should also take care to tag these ATMs with `network` as appropriate (often but not always redundantly to `brand`).

bhousel:
> We should also take care to tag these ATMs with `network` as appropriate (often but not always redundantly to `brand`).

Awesome - this is interesting to know..
Are you sure `network` is the right tag for this?  Would like to hear more.

I think of `network` as something else, more specifically like the formats used by road and transit networks.

1ec5:
Yep, `network` is typically [tagged on ATMs](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Datm) and other amenities like bike repair stations. This practice happens to predate `brand` (which used to just be for saying that a particular car repair shop services Volvos, say), but it does express a slightly different kind of affiliation than `brand`.

For better or worse, the same `network` key has always been used for multiple things that English speakers call networks: “route network”, “ATM network”, etc. That’s one reason NSI will be a good complement to taginfo for public transportation networks (#2864).

It’s unclear to me whether [interbank networks](https://en.wikipedia.org/wiki/Interbank_network) can also be tagged in `network`, but we can cross that bridge when we get there. The most obvious NSI entries would be simple cases where `brand` and `network` are the same.

mds08011:
> It’s unclear to me whether interbank networks can also be tagged in `network`.

What is an example of a `network` that isn't an interbank network?  Wouldn't that just be a `brand`?

Assuming interbank networks are tagged as `network`, it will be common for credit union ATMs to have different `brand` and `network` tags.  For example, Navy Federal Credit Union participates in the [Plus](https://en.wikipedia.org/wiki/Plus_(interbank_network)) interbank network.

[...]

PR #3056, opened 2019-10-01
Add new ATM category

mds08011:
This is a draft pull request related to adding ATMs to the NSI as discussed in #2883 .  Still a work in progress.  Feedback is welcome as are any changes.


andrewharvey:
Since most banks with a store would have an ATM, it's it worth bootstrapping this by copying amenity=bank and then any tweaks can be made from there?

mds08011:
The initial intent was to focus on banks known for having freestanding ATMs.  While it’s true that most banks would have ATMs attached to the bank location, not all have them have a network of separate ATMs.  I’m hesitant that bootstrapping banks will bring too many excess entries.

andrewharvey:
No worries, looks good to me, which part is still a WIP? I'd like to add Australian ATMs, but would be best if this get merged first to avoid conflicts..

mds08011:
@andrewharvey overall we are trying to work through how to best use `name`, `brand`, `operator`, and `network` tags.

You can see more discussion here: https://github.com/osmlab/name-suggestion-index/issues/2883

brands/amenity/atm.json
```diff
@@ -0,0 +1,127 @@
...
+      "Harris Bank"
+    ],
+    "tags": {
+      "amenity": "atm",
+      "brand": "BMO Harris Bank",
+      "brand:wikidata": "Q4835981",
+      "brand:wikipedia": "en:BMO Harris Bank",
+      "name": "BMO Harris Bank",
+      "network": "Allpoint",
+      "network:wikidata": "Q4733264"
```
anandfresh: I feel the amenity property is redundant. Can we do some JSON refactoring to avoid redundancy and to keep the common values ahead 
"amenity": "atm",

bhousel: These are redundant only because we group them into files by this property key=value pair.  The property is important as an OpenStreetMap tag, so it must stay.

anandfresh: But do we have any plan to refractor this JSON/app in near future?

bhousel: No plans, I’m pretty happy with the JSON structure.

UKChris-osm:
Not sure how this draft is going, or if this entry would be suitable, but below is an entry for Sainsbury's Bank ATM's.

"amenity/atm|Sainsbury's Bank": {
    "countryCodes": ["gb"],
    "tags": {
      "amenity": "atm",
      "brand": "Sainsbury's Bank",
      "brand:wikidata": "Q7400525",
      "brand:wikipedia": "en:Sainsbury's Bank",
      "name": "Sainsbury's Bank",
      "network": "Link",
      "network:wikidata": "Q6458752"
    }
}

Sainsbury's doesn't have any conventional bank locations, just ATM's at Sainsbury's and Argos stores, as well as some travel money counters, so that might make this entry more aligned with the freestanding ATM concept mentioned above.

bhousel:
Going to close this for now.  

Restructuring the code to allow `operator` or `network` to be the the defining tag (instead of `brand`) is in the back of my mind and will happen eventually.

PR #3058, opened 2019-10-02
To the post_office module added chilexpress with with wikidata and wi…

Nuthi-Sriram:
To the post_office module added chilexpress with with wikidata and wikipedia and also added correo OCA
Added Animal_boarding with brand:wikipedia

### DIFF
--- a/brands/amenity/animal_boarding.json
+++ b/brands/amenity/animal_boarding.json
@@ -6,6 +6,7 @@
       "animal_boarding": "dog;cat",
       "brand": "PetsHotel",
       "brand:wikidata": "Q67141961",
+      "brand:wikipedia": "en:PetSmart#PetsHotel",
       "name": "PetsHotel"
     }
   }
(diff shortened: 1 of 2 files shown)


Adamant36:
Hi there. It seems that the Wikipedia article for Telex-Chile is a diffunct brand that is a different company then the entry. Also with the pet hotels from what I've seen we don't usually use the Wikipedia tag to link to sections of articles. So it would probably be better to just create a new Wikidata entry for it instead. Although, someone else might know differently.

Nuthi-Sriram:
Ohh I get it

Nuthi-Sriram:
> Hi there. It seems that the Wikipedia article for Telex-Chile is a diffunct brand that is a different company then the entry. Also with the pet hotels from what I've seen we don't usually use the Wikipedia tag to link to sections of articles. So it would probably be better to just create a new Wikidata entry for it instead. Although, someone else might know differently.

I have made some changes I have removed the two erroneous code which I had written

bhousel:
Thanks for the quick fix @Nuthi-Sriram and welcome! 👋

Issue #3054, opened 2019-10-01
Altar’d State


1ec5:
`shop=clothes`

https://www.altardstate.com/help/?page=store-locator

No Wikidata item yet. A public domain logo can be found in SVG format in the above webpage’s source code – search for `as-header-logo` – and uploaded to Wikimedia Commons for use in a new Wikidata item.

aestevens:
Hi @1ec5, I just entered #3124 after creating the Wikidata Item and adding the tag to the index. I'm happy to add the logo to the Wikidata Item, but my only concern is the legality of uploading logos to Wikimedia Commons.

My understanding is that logos constitute copyrighted material, and as such are not eligible for upload to the Commons. However, I'm seeing numerous logos that were uploaded to Commons that are listed variously as public domain or with a form of CC license. Can you clarify on your understanding of uploading logos to Commons? Thanks in advance!

PR #3060, opened 2019-10-03
Added Altar'd State brand to shop json file.

rsuttles58:
As noted, added Altar'd State brand information.  Was not able to locate a wikipedia page for this smaller brand.  This in reference to Issue #3054


bhousel:
Thanks @rsuttles58

PR #3069, opened 2019-10-04
Add book store "Libro"

hikemaniac:
This PR adds the Austrian book retail chain "Libro" to the index. See https://www.wikidata.org/wiki/Q1823138

Because this is my first contribution to this repository, could you please check if everything is right?


Adamant36:
Looks good. Thanks 👍

PR #3070, opened 2019-10-04
Add Cinemas UGC

bmillemathias:
ssia

initially I did not want to set the tag "name" because each cinema has its own name - usually "UGC $something", but this tag is mandatory.

https://overpass-turbo.eu/s/MRM

### DIFF
--- a/brands/amenity/cinema.json
+++ b/brands/amenity/cinema.json
@@ -347,6 +347,16 @@
       "name": "The Space Cinema"
     }
   },
+  "amenity/cinema|UGC": {
+    "countryCodes": ["fr"],
+    "tags": {
+      "amenity": "cinema",
+      "brand": "UGC",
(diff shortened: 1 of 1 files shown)


brands/amenity/cinema.json
```diff
@@ -347,6 +347,17 @@
       "name": "The Space Cinema"
     }
   },
+  "amenity/cinema|UGC": {
+    "countryCodes": ["fr"],
+    "tags": {
+      "amenity": "cinema",
+      "brand": "UGC",
+      "brand:website": "http://ugc.fr",
```
bhousel: Can you remove this tag?  We don't include `brand:website` in the index because brands can change it.  Anyone who wants this data should probably get the current website from the brand's wikidata page.

bmillemathias:
> remove `brand:website` please +1

done

PR #3071, opened 2019-10-04
Add info for La vie claire

bmillemathias:
This is a organic only shop and use French Wikipedia as the article is more complete and La Vie Claire is french based afaik


bhousel:
Looks good, thanks 👍

PR #3073, opened 2019-10-05
Removed unnecesary matchNames

himanshudigrase:


bhousel:
Some of them seem necessary 🤔

himanshudigrase:
But the ones that do not match are to be removed . Please correct me if I
am wrong.

bhousel:
So for this entry:

```js
  "tourism/hotel|Baymont": {
    "countryCodes": ["us"],
    "matchNames": [
      "baymont inn",
      "baymont inn & suites",
      "baymont inn and suites"
    ],
    "tags": {
      "brand": "Baymont",
      "brand:wikidata": "Q4874634",
      "brand:wikipedia": "en:Baymont Inn & Suites",
      "name": "Baymont",
      "official_name": "Baymont by Wyndham",
      "tourism": "hotel"
    }
  },
```

This is like saying:
The name of this thing is "Baymont", but we will also recognize "Baymont Inn", "Baymont Inn & Suites", "Baymont Inn and Suites", "Baymont by Wyndham" as all meaning the same thing.  We ignore case, whitespace, and punctuation when we look at names.  If we see a feature in OpenStreetMap tagged as `tourism=hotel` (or `tourism=motel` because of match groups) with any of these names, we consider it a match for this official "Baymont" entry and can check whether its tags match the supplied tag object.

So.. The "Baymont Inn & Suites" name is indeed redundant because we automatically expand the `&` into `and`.  However the other `matchNames` are not redundant and should stay.   

I know it is tricky and this project has evolved a bit to be this way, but I'm always looking for ways to improve it so let me know if this explanation makes sense.  I probably need to update the contributing guide a bit.

himanshudigrase:
Thank you for your response. Yes this is tricky but it needs further
description. The description you provided was enough for me to understand
the mistake.

bhousel:
This has gone a bit stale so it's easier for me to close it right now.. 
@tas50 will be working through some changes to the hotel names soon which will change `matchNames` properties for a bunch of them.

PR #3082, opened 2019-10-07
Retag Honey Baked Ham

1ec5:
Like Boston Market, Honey Baked Ham is a fast food restaurant chain that also sells larger portions of meat to go. Like Arby’s, Honey Baked Ham is not a butcher shop, though it shares a lot in common with butcher shops.

Previously the stores were branded HoneyBaked Ham (or, more fully, The HoneyBaked Ham Company). Recently they adopted a new logo and tweaked the name to The Honey Baked Ham Company, but HoneyBaked still appears in some of their marketing materials, perhaps only as the name of the prepared ham product.

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -1001,6 +1001,22 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|Honey Baked Ham": {
+    "countryCodes": ["us"],
+    "matchNames": ["honey baked ham company"],
+    "matchTags": ["shop/butcher", "shop/deli"],
+    "tags": {
+      "alt_name": "HoneyBaked Ham",
+      "amenity": "fast_food",
+      "brand": "Honey Baked Ham",
+      "brand:wikidata": "Q5893363",
+      "brand:wikipedia": "en:The Honey Baked Ham Company",
+      "cuisine": "american",
+      "name": "Honey Baked Ham",
+      "official_name": "The Honey Baked Ham Company",
+      "takeaway": "yes"
+    }
+  },
   "amenity/fast_food|Hot Dog on a Stick": {
     "countryCodes": ["us"],
     "tags": {
--- a/brands/shop/butcher.json
+++ b/brands/shop/butcher.json
@@ -35,16 +35,6 @@
       "shop": "butcher"
     }
   },
-  "shop/butcher|HoneyBaked Ham": {
-    "countryCodes": ["us"],
-    "tags": {
-      "brand": "HoneyBaked Ham",
-      "brand:wikidata": "Q5893363",
-      "brand:wikipedia": "en:The Honey Baked Ham Company",
-      "name": "HoneyBaked Ham",
-      "shop": "butcher"
-    }
-  },
   "shop/butcher|M&M Food Market": {
     "countryCodes": ["ca"],
     "tags": {


bhousel:
Calling it a fast food seems odd to me.  I don't think people eat-in there.  

FWIW, Google calls it a deli:

Screenshot 2019-10-07 01 09 21

Adamant36:
> I don't think people eat-in there.

honeybaked cafe 2

Honeybaked express

Honeybaked

Honeybaked menu

bhousel:
Oh cool.. I guess the “and Cafe” and “express” formats have some dining options..  I think the one near me really is just a ham store, but I haven’t checked it out  yet.

1ec5:
> OSM's definition of deli kind of fits

OSM’s definition of `shop=deli` has always been very clear about applying to what Americans call “gourmet food”. It’s unfortunate that there’s no tag for American-style delis other than `shop=convenience` or `cuisine=sandwich`. Honey Baked Ham is higher-end than most American delis, but I don’t think it fits `shop=deli` any more than Boston Market would.

> I don't think people eat-in there.

We have Papa Murphy’s as `amenity=fast_food` too, even though it sells ready-to-bake pizzas. If some Honey Baked Ham stores don’t have dine-in, perhaps we should leave out `takeaway=yes` and give mappers the opportunity to set it to `takeaway=only`?

tas50:
Seems like if they have a cafe model we should tag those as Honeybaked Ham & Cafe. They shouldn't get lumped into one. It's no worse than Dairy Queen vs. DQ Grill & Chill, which are unique in the index and are probably 90% the same.

1ec5:
> It might not be a "butcher" but it's definitely not "fast food" in any sense of the word. It's more like an Omaha Steaks or something.

Ideally there would be a tag that covers meat stores in general without being as specific as `shop=butcher`, which connotes freshness/rawness. Or a tag for a store that sells take-home meals that aren’t frozen.

tas50:
butcher seems to cover it pretty well based on the description in the wiki https://wiki.openstreetmap.org/wiki/Tag:shop%3Dbutcher

1ec5:
> I don't think this brand has anything in common with Auntie Annes or Arbys or Boston Market, or any of the examples you gave.

Obviously I was straining for analogies. 😉 Where I’m from, Honey Baked Ham and Boston Market are where you go for a full-fledged family meal in a bag on holidays. One does ham and the other does chicken. But both also serve individual lunches dine-in or carry-out. I had never seen a standalone Honey Baked Ham ham store before.

> butcher seems to cover it pretty well based on the description in the wiki

To me, the [`shop=butcher`](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dbutcher) page (even without [my recent edit](https://wiki.openstreetmap.org/wiki/Special:Diff/1911061)) seems to align `shop=butcher` pretty closely to the common meaning of “butcher”: a store that sells uncooked steaks, ground pork, and the like. Prepared meats such as baked hams and sausages are standard fare for a delicatessen, for Americans and Europeans alike.

The only reason I hesitated about `shop=deli` in https://github.com/osmlab/name-suggestion-index/pull/3082#issuecomment-539079805 is that I’ve always associated Honey Baked Ham with the deli-café format. If we end up splitting the entry in two, one for deli-cafés and the other for standalone ham stores, then I agree with @bhousel about `shop=deli` being a good fit for the latter.

bhousel:
Ok - so I went on a fact finding mission to the Honey Baked Ham Company the other day... 

This store near me does not have an "and Cafe" in the name.  They did have a few tables inside and they did have a one page menu, so I guess people could eat sandwiches and salads there, but the focus really seemed to be on selling the hams and mustards and stuff.  I really think `shop=deli` is a better fit for the place near me, but I'm just going to merge this anyway.  I think calling it a fast food is a stretch but ok.

1ec5:
> They did have a few tables inside and they did have a one page menu, so I guess people could eat sandwiches and salads there, but the focus really seemed to be on selling the hams and mustards and stuff. I really think shop=deli is a better fit for the place near me

Yeah, I agree it isn’t clear-cut. I think different people are going to look at the same stores and draw the line (or slice the ham) differently. In the extreme case, Meijer stores have tables near the deli and bakery sections, but it’s a stretch to call those sections fast food restaurants, because the tables are mostly for show.

bhousel:
> Yeah, I agree it isn’t clear-cut. I think different people are going to look at the same stores and draw the line (or slice the ham) differently. In the extreme case, Meijer stores have tables near the deli and bakery sections, but it’s a stretch to call those sections fast food restaurants, because the tables are mostly for show.

Yep, I was thinking about this.. Bakeries, delis, butchers, small groceries sometimes have tables to eat at, and that's fine.  Really I think the main reason to choose one tag over another is just so the map marker is drawn with a good icon and the map doesn't look weird.   (Honey Baked Ham tagged as a fast food will probably show a burger icon on the map pins, and yes that is weird).

PR #3083, opened 2019-10-07
Add indonesian store gramedia

irhamdz:
add some bookstore in indonesian

### DIFF
--- a/brands/shop/books.json
+++ b/brands/shop/books.json
@@ -121,6 +121,20 @@
       "shop": "books"
     }
   },
+  "shop/books|Gramedia": {
+    "countryCodes": ["id"],
+    "tags": {
+      "brand": "Gramedia",
+      "brand:en": "Gramedia",
+      "brand:id": "Gramedia (toko buku)",
+      "brand:wikidata": "Q19646130",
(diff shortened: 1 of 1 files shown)


brands/shop/books.json
```diff
@@ -121,6 +121,20 @@
...
+    "countryCodes": ["id"],
+    "tags": {
+      "brand": "Gramedia",
+      "brand:en": "Gramedia",
+      "brand:id": "Gramedia (toko buku)",
+      "brand:wikidata": "Q19646130",
+      "brand:wikipedia": "id:Gramedia (toko buku)",
+      "name": "Gramedia",
+      "name:en": "Gramedia",
+      "name:id": "Gramedia (toko buku)",
```
bhousel: Does it really need `brand:en`, `brand:id`, `name:en` and `name:id` tags?   I don't understand them..
It looks like this brand is just "Gramedia" always?

irhamdz: ah my bad, i will fix it

irhamdz:
i already update it, maybe you can check it @bhousel

bhousel:
Thanks @irhamdz looks great ! 👍

PR #3084, opened 2019-10-07
fix and add some indonesian supermarket

irhamdz:
fix and add some indonesian supermarket


bhousel:
Thanks and welcome @irhamdz  👋 
This looks good to me!

Issue #3053, opened 2019-10-01
Foot Solutions


1ec5:
`shop=shoes` and maybe something in `healthcare` to indicate that they specialize in “holistic foot analysis”.

https://footsolutions.com/store-locator/

No Wikidata item yet.

bhousel:
done in #3085

PR #3085, opened 2019-10-07
Foot Solutions #3053

adzika:
Added foot solutions shop and created Wikidata entry, there's no Wikipedia for this shop. I'm not sure how to add this to "healthcare" as I haven't found this label.


adzika:
Thank you for your reply, it's one of my first pull requests, hence the mistakes, even though I tried to read docs thoroughly. :)

Ad 1. Yes, it's a WebStorm file and sure, I will add it to .gitignore.
Ad 2. Ok, I just wasn't sure how to do it because I didn't see that info in the docs. Maybe after I'll change the code, I can also add one sentence about leaving the Wikipedia entry empty in the docs if the website is not there?
Ad 3. Sure! Again I was trying to follow the docs but the animation looked quite different from what I saw on the screen and it was my first time doing this, so I got a bit confused.

I'm going to change all that now.

adzika:
I've pushed the second commit, not sure if I need to anything else now?

bhousel:
> I've pushed the second commit, not sure if I need to anything else now?

I still see the files under `.idea` on the pull request, but the other items look ok..

I think you can probably remove them with something like `git rm -R .idea`

adzika:
Thanks, I deleted .idea files.

bhousel:
Thank you, looks great @adzika 👍

adzika:
Thanks for guiding me along the way. :)

brands/shop/shoes.json
```diff
@@ -294,6 +294,14 @@
       "shop": "shoes"
     }
   },
+  "shop/shoes|Foot Solutions": {
+    "tags": {
+      "brand": "Foot Solutions",
+      "brand:wikidata": "Q69982369",
+      "name": "Foot Solutions",
+      "shop": "shoes"
```
1ec5: > I'm not sure how to add this to "healthcare" as I haven't found this label.

You can add any tags you need to as properties in this JSON object, alongside `shop`, `name`, etc. I was suggesting adding a `healthcare=something` or `healthcare:speciality=something` yah but wasn’t sure exactly what to set it to. Later on, as I fleshed out the Foot Solutions item on Wikidata to save it from deletion, I learned of the word “pedorthist”, which would work as a value for the `healthcare:speciality` tag.

1ec5: Added in 6a25dbc04ff82a35fd49f0d73b247dc099851d82 along with `shoes=orthopaedic`, which is [documented on the wiki](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dshoes).

PR #3088, opened 2019-10-08
Remove "by Marriott" from Courtyard hotel chain

mds08011:
Changes the `name` and `brand` of "Courtyard by Marriott" to "Courtyard" to be consistent with other Marriott brands in the NSI (e.g. Fairfield Inn, Residence Inn, TownePlace Suites).

### DIFF
--- a/brands/tourism/hotel.json
+++ b/brands/tourism/hotel.json
@@ -125,10 +125,11 @@
       "courtyard marriott"
     ],
     "tags": {
-      "brand": "Courtyard by Marriott",
+      "brand": "Courtyard",
       "brand:wikidata": "Q1053170",
       "brand:wikipedia": "en:Courtyard by Marriott",
-      "name": "Courtyard by Marriott",
+      "name": "Courtyard",
+      "official_name": "Courtyard by Marriott",
       "tourism": "hotel"
     }
   },


bhousel:
SGTM, thanks @mds08011  👍

tas50:
So some of the brands operate with 'by Marriott' on all their signage and in references on their website. I was trying to be consistent with how Marriott refers to the brands. This is also how they were mapped in the majority of cases in OSM and it's how both Apple Maps and Google Maps handles it.

bhousel:
Sorry!  I probably merged a bit too quickly before giving people a chance to comment.. I'm really ok with this being either "Courtyard" or "Courtyard by Marriott".   (or using `short_name` or `official_name` for the alternate variant).  Anyway feel free to talk it over for a few days and revert if this is what people want.

tas50:
I'm not a huge fan of marriott putting their name all over all their brands. I mostly fall back to what the industry giants have chosen to do in terms of branding. Courtyard is better IMO, but is that correct 🤷‍♂. Either way we should make sure we're consistent with it and update all the other Marriot brands to use one or the other.

mds08011:
Likewise, I don’t feel strongly one way or another as long as we are consistent.  I knew “by Marriott” was going on the marketing logos but I didn’t realize it was also being added to physical signs.

Adamant36:
According to taginfo there's like a thousand uses of name=Courtyard. I assume a lot of those are generic. If the name index went with Courtyard here would it also be possible have Courtyard as a flagged generic term or would that cause issues?

mds08011:
I’ve changed my opinion and now think this should be reverted, and Marriott Brands in NSI should include “by ____”  as applicable.

I recently stayed in a “Four Points by Sheraton” (owned by Marriott) and the “…by Sheraton” was included on everything from the exterior signs, to floor mats, and stationary.

As @tas50 mentioned, I don’t like this branding but it does appear to be the direction Marriott is going.

bhousel:
Haha I think it's kind of silly too to add the "by Marriott" but it is what the brand wants and does everywhere, and for what it's worth, Google is on board:

Screenshot 2019-10-15 15 49 01

We can put "Courtyard" in the `short_name` tag, for anyone who just wants to make more minimalist maps.


>  If the name index went with Courtyard here would it also be possible have Courtyard as a flagged generic term or would that cause issues?

Yeah it would cause issues to have it as a generic name.

PR #3090, opened 2019-10-08
add University of leicester

klswcz:


bhousel:

see also #3051, #3068 

May I ask what prompted you to make this pull request?  We've been getting a bunch of university pull requests lately and feel bad about closing them.

klswcz:
I think it might be due to start of Hacktoberfest, in my case university was one of the first places that came to my mind when thinking about new places I want to add to map

bhousel:
Thanks!  Yes you can still add the wikidata tags to your university in OpenStreetMap, and that's a good thing to do.

This project is really more for businesses with a lot of locations, where we want them all mapped consistently.

PR #3093, opened 2019-10-09
Break Microtel up into 2 brands

tas50:
This is another case of different brands under the same sub brand. Hotels are crazy complex brand wise, but removing the "suites" bit from a name in North America is wrong they do the real rebrand. For now there's only 3 Microtels in the US and Canada.

### DIFF
--- a/brands/tourism/hotel.json
+++ b/brands/tourism/hotel.json
@@ -489,23 +489,33 @@
   },
   "tourism/hotel|Microtel": {
     "countryCodes": [
-      "ar",
       "ca",
+      "cn",
       "mx",
       "ph",
       "us"
     ],
-    "matchNames": [
-      "microtel by wyndham",
-      "microtel inn",
-      "microtel inn & suites",
-      "microtel inn and suites"
+    "matchNames": ["microtel inn"],
+    "nomatch": [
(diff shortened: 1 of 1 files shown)


brands/tourism/hotel.json
```diff
@@ -489,23 +489,33 @@
   },
   "tourism/hotel|Microtel": {
     "countryCodes": [
-      "ar",
```
tas50: This location is not on their website.

brands/tourism/hotel.json
```diff
@@ -489,23 +489,33 @@
...
       "ca",
+      "cn",
```
tas50: China is on their website

brands/tourism/hotel.json
```diff
@@ -489,23 +489,33 @@
...
       "brand": "Microtel",
       "brand:wikidata": "Q6840402",
       "brand:wikipedia": "en:Microtel Inn and Suites",
       "name": "Microtel",
+      "official_name": "Microtel by Wyndham",
+      "tourism": "hotel"
+    }
+  },
+  "tourism/hotel|Microtel Inn & Suites": {
+    "countryCodes": ["ca", "mx", "us"],
```
tas50: The Inn & Suites variety is only in North America

PR #3094, opened 2019-10-09
Add Austrian furniture stores

hikemaniac:
This PR adds the following Austrian furniture stores:

- Kika ([Q256456](https://www.wikidata.org/wiki/Q256456))
- Leiner ([Q1661472](https://www.wikidata.org/wiki/Q1661472))
- Mömax (a subsidiary of _Möbelix_)
- Rutar ([Q22910157](https://www.wikidata.org/wiki/Q22910157))

### DIFF
--- a/brands/shop/furniture.json
+++ b/brands/shop/furniture.json
@@ -285,6 +285,16 @@
       "short_name": "Jerome's"
     }
   },
+  "shop/furniture|Kika": {
+    "countryCodes": ["at"],
+    "tags": {
+      "brand": "Kika",
+      "brand:wikidata": "Q256456",
+      "brand:wikipedia": "de:Kika (Möbelhaus)",
+      "name": "Kika",
+      "shop": "furniture"
+    }
+  },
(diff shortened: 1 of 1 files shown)


brands/shop/furniture.json
```diff
@@ -354,6 +374,14 @@
       "shop": "furniture"
     }
   },
+  "shop/furniture|Mömax": {
```
tas50: Without a wikidata value this one won't actually show up in editors.

hikemaniac: Okay, thank you. The problem is, that (like _XXXLutz_) it belongs to the company "Möbelix". It has in fact got two Wikidata pages - one individual one with barely any data and a redirect to "Möbelix" (in the same way as _XXXLutz_).

hikemaniac:
I've added the missing Wikidata items, everything should be working now. 😁

bhousel:
> I've added the missing Wikidata items, everything should be working now.

Great, thanks for the quick additions! 👍

PR #3095, opened 2019-10-09
Add Ubaldi Electronic shop.

bmillemathias:
Ubaldi is a electronic (mainly) shops chain which has few shops in south of France.
Not a lot are currently reported in OSM though.

https://overpass-turbo.eu/s/MYx

### DIFF
--- a/brands/shop/electronics.json
+++ b/brands/shop/electronics.json
@@ -395,6 +395,14 @@
       "shop": "electronics"
     }
   },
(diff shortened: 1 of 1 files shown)


tas50:
@bmillemathias can you add a wikidata entry so this shows up in the index?

bmillemathias:
I think have this brand is not really interesting.

PR #3099, opened 2019-10-09
university.json update

max180643:
add Thailand Universities in university.json


bhousel:

PR #3103, opened 2019-10-10
Add popular British pubs

klswcz:


brands/amenity/pub.json
```diff
@@ -6,5 +6,25 @@
...
+      "name": "Wetherspoons"
+    }
+  },
+  "amenity/pub|Walkabout": {
+    "countryCodes": ["gb"],
+    "tags": {
+      "amenity": "pub",
+      "brand": "Walkabout",
+      "brand:wikidata": "Q7962149",
+      "brand:wikipedia": "en:Walkabout",
```
bhousel: looks like the actual wikipedia page for this should be `"en:Walkabout (pub chain)"`

brands/amenity/pub.json
```diff
@@ -6,5 +6,25 @@
...
       "name": "Ezītis miglā"
     }
+  },
+  "amenity/pub|Wetherspoons": {
+    "countryCodes": ["gb"],
+    "tags": {
+      "amenity": "pub",
+      "brand": "Wetherspoons",
+      "brand:wikidata": "Q6109362",
+      "brand:wikipedia": "en:J D Wetherspoon",
```
bhousel: looks like the actual Wikipedia page for this one should be `"en:Wetherspoons"`

klswcz: done

brands/amenity/pub.json
klswcz: done

brands/amenity/pub.json
bhousel: It looks like you set it to `"en:Wetherspoon"` instead of the correct  `"en:Wetherspoons"`.

Can you try again?  Please pay attention to the details..

klswcz: done, sorry about that, everything should work fine now

bhousel:
Thanks for fixing! 👍

tohaklim:
Would it be name=Wetherspoons? Almost none of their pubs I’ve been to have the brand in the name

klswcz:
The brand's name is Wetherspoons, but almost every local has also its own name. If you go inside there's Wetherspoons branding on menus, staff's shirts etc.

tohaklim:
Yeah, each pub would usually have the “canon” local name, often after the old pub or the building it’s in. Wetherspoons is occasionally shoehorned into the name (often in brackets after the actual name).

bhousel:
ok, I've removed it in b9c20873f
This brand is just too much trouble for this index if they all use really different names.

PR #3104, opened 2019-10-10
Add two British popular chain restaurants

klswcz:


bhousel:
Thanks 👍 
They won't actually show until someone assigns Wikidata IDs, but there's no harm in having them there.

PR #3106, opened 2019-10-11
added tempero manero

Tubaleviao:
Brazilian restaurant

### DIFF
--- a/brands/amenity/restaurant.json
+++ b/brands/amenity/restaurant.json
@@ -1943,6 +1943,15 @@
       "name": "Ted's Montana Grill"
     }
   },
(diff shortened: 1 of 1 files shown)


bhousel:
Is it this place?  http://www.temperomanero.com.br/
Is this really a brand with multiple locations - or just a single restaurant?

Tubaleviao:
Yes, that is the place.
It's a brand with multiple locations in Brazil.

bhousel:
Sounds good to me, thanks @Tubaleviao  👍 
It won't actually get included without a `brand:wikidata` tag, but there's no harm in having it here.

PR #3108, opened 2019-10-12
Add a new entry to amenity/cafe

iamjackslayer:


brands/amenity/cafe.json
```diff
@@ -113,6 +113,19 @@
       "takeaway": "yes"
     }
   },
+  "amenity/cafe|BreadTalk": {
+    "countryCodes": ["sg"],
+    "matchNames": ["Bread Talk"],
```
bhousel: This `matchName` is unnecessary and can be removed - the matching code ignores capitalization and spaces..

brands/amenity/cafe.json
```diff
@@ -113,6 +113,19 @@
...
+    "tags": {
+      "amenity": "cafe",
+      "brand": "BreadTalk",
+      "brand:wikidata": "Q1106640",
+      "brand:wikipedia": "en:BreadTalk",
+      "cuisine": "coffee_shop",
```
bhousel: I'm not familiar with this brand.. Is it definitely an `amenity=cafe` + `cuisine=coffee_shop` ?

From looking around https://www.breadtalk.com.sg/   - It looks like maybe more of a `amenity=fast_food` or `shop=bakery` ...  Cafes are really for coffee places like Starbucks.

iamjackslayer: I stand corrected. You are right that BreadTalk is primarily a bakery. Thank you for the correction.

bhousel:
Thank you for the quick fix, and for your first contribution!  👋

brands/shop/bakery.json
```diff
@@ -101,6 +101,17 @@
       "shop": "bakery"
     }
   },
+  "shop/bakery|BreadTalk": {
+    "countryCodes": ["sg"],
+    "tags": {
+      "amenity": "cafe",
```
bhousel: oh I just noticed this extra `"amenity": "cafe"` still here so I removed it in eeeff003b

Issue #2877, opened 2019-07-12
Add "Perfect Brows NYC"


yoasif:
https://www.perfectbrowsnyc.com/make-appointment/

amous4822:
wikipedia and wikidata pages were not found

Adamant36:
That's why it has the "needs wikidata" tag. So one can be made for it ;)

amous4822:
oh. my bad. sorry.

PR #3112, opened 2019-10-13
added new brand to amenities

amous4822:
#2877
added perfect brows nyc 
wikipedia and wikidata pages were not found

### DIFF
--- a/brands/amenity/salon.json
+++ b/brands/amenity/salon.json
@@ -1,0 +1,13 @@
+{
+  "amenity/salon|Perfect Brows NYC": {
+    "countryCodes": ["us"],
+    "matchNames": ["Perfect Brows"],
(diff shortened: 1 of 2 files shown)


Adamant36:
Hi  there. We try to stick to established tags. So I don't really think the healthcare tags are correct here. Since there doesn't seem to be a healthcare=salon tag on the wiki and it just seems like an amenity alone anyway, not a healthcare thing. "threading" probably isn't a  thing tagging wise either. Although I could be wrong.

amous4822:
thank you for your response. can it be added under shops ?

bhousel:
> thank you for your response. can it be added under shops ?

I'd add it to `shop/beauty.json` - there are a few other eyebrow places in there.

PR #3114, opened 2019-10-13
Move to discardKeys

tohaklim:
Is the syntaxis right here?


bhousel:
Yes this seems ok, thanks @tohaklim 👍

PR #3115, opened 2019-10-13
Keep amenity/hospital and amenity/clinic tags

tas50:
We have both of these in the index. These are going to need a lot more filtering for generic names, but I added a few obvious english ones.


bhousel:
This is great!  Thanks for including these categories and working through the filters 🎉

PR #3116, opened 2019-10-13
hotel

danielatellez:


tas50:
It looks like this wikidata / wikipedia data is for a single hotel building and not the chain.

Adamant36:
Is it even a chain? They are spread all over Europe and seems like all the Hotel Belvederes in internet searches are separate entities. Belvedere means summerhouse. So maybe it's a generic.

bhousel:
Ah yeah, it doesn't look like a brand.. @danielatellez can you remove this entry from `hotel.json`?

It looks like it should already be filtered by this line:
https://github.com/osmlab/name-suggestion-index/blob/cdc12ad632d805d887b929ccf94ff2e038ac97e8/config/filters.json#L99

So it shouldn't return back into the index if we detect them again.

danielatellez:
@bhousel this had been removed.

PR #3117, opened 2019-10-13
Add UK Post Office brand data

GurjitSangha:


bhousel:
I guess we have to think about this more.. It looks like they just call it "Post Office", not "Post Office Ltd".  Usually in OSM the `name` is the name on the sign.

We are filtering out "Post Office" as a generic term, so we would have to change how that works.  All the other post offices around the world have the name of the country in them, like "United States Post Office", "Australia Post", "Canada Post", so there is no issue or ambiguity for the existing entries.   

AFAIK this is the only entry in the index where a single country uses a very generic name to refer to a specific brand.

SilentSpike:
I would agree (as someone in the UK) that the `name` should just be `Post Office`. There are some other brands like this where the name on the sign differs from the company name where I tend to use `official_name` for the latter.

bhousel:
Cool thanks @SilentSpike - is good to hear a local perspective!
Going to merge this and remove "post office" from the `filters.json` file.

PR #3119, opened 2019-10-14
Added Strike Bowling Bar

GenericError:
Created brands/leisure/bowling_alley.json and added an entry for Strike Bowling Bar

### DIFF
--- a/brands/leisure/bowling_alley.json
+++ b/brands/leisure/bowling_alley.json
@@ -1,0 +1,13 @@
+{
+  "leisure/bowling_alley|Strike Bowling Bar": {
+    "countryCodes": ["au"],
+    "matchTags": ["leisure/bowling_alley"],
+    "tags": {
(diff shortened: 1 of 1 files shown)


GenericError:
Related issue #3119

brands/leisure/bowling_alley.json
```diff
@@ -0,0 +1,13 @@
+{
+  "leisure/bowling_alley|Strike Bowling Bar": {
+    "countryCodes": ["au"],
+    "matchTags": ["leisure/bowling_alley"],
```
bhousel: This `matchTags` property can be removed,  it already matches that one.

GenericError: Thanks for that; looking at the other files it looks unnecessary

bhousel:
Thanks! looks good 👍

Issue #3113, opened 2019-10-13
Bell Canada


Shermapper1:
Right now, Bell stores are treated as mobile phone stores, and sometimes they are. But Bell Canada also offers Satellite TV and Landline services, and as such, many Bell Canada stores offer all 3.

tohaklim:
This is similar to A1, which offers TV and cable services. Unless office=telecom is a better fit, I'd say we map those based on what is (perceived) to be the primary function

bhousel:
Yeah, a handful of these have ended up as `office=telecom`.  I'm kind of surprised people want that to be the tag.. `shop=something` seems more appropriate for a storefront business.

Shermapper1:
I ended up making up shop=telecom, because I've gotten tired of store/business tags that fit poorly (or not at all).

bhousel:
Oh yeah, see #2900 too - I think we're kind of moving towards `shop=telecommunication`.  I agree with @tas50 .  I think we should start retagging these.

tohaklim:
Yeah, that would work well, and uses the same term as the office tag. Would be good to put it on the wiki as “in use”, if avoiding the proposal process

Issue #2900, opened 2019-07-17
Add Spectrum stores


yoasif:
See https://www.spectrum.com/locations/index.html

mds08011:
@yoasif thank you for adding these!

As you find more brands to include in the index, it is very helpful if you can also include whether or not it has an existing Wikidata ID.  In this case, Spectrum does: [Q7805197](https://www.wikidata.org/wiki/Q7805197).

As a quick check, if the brand has a Wikipedia page the Wikidata ID will often be linked on the left side under “Tools”:

<img src="https://user-images.githubusercontent.com/8010709/61414441-c4b6e680-a8a2-11e9-808e-f8fdc7365aba.png" width="70%">

Adamant36:
I know if adding the existing Wikipedia/wikidata pages would be good in this instance because its less about the stores themselves and more about the services the company provides. We should probably create unique Wikidata entry just for the stores instead. Which would also work better for the stuff on Wikidata's side to like "Instance of." Although, I don't know how it would work with social media accounts or website links. Since they would have to be unique.

mds08011:
You have a good point overall but I don't see what sets Spectrum apart from any other entries.  For the majority of other restaurants and shops in the index, the corporation's Wikidata page is linked rather than one specific to the physical stores.  Ultimately, we are tagging `brand:wikidata=` and the Spectrum wiki pages seem to adequately to convey the brand.

mds08011:
I'm also not sure how to tag these stores.  `shop=electronics` seems to be the most applicable official tag.  Someone on the OSM-US Slack also suggested `shop=internet` which is not official but has [moderate use](https://taginfo.openstreetmap.org/tags/shop=internet)

bhousel:
I'd use either `office=telecommunication` or `shop=mobile_phone` based on similar brands in the index already..

mds08011:
`office=telecommunication` seems more applicable as these locations sell a lot more than just mobile phones.  That does seem consistent with other index brands.

tas50:
I tag these as `shop=telecommunication`. It's clearly not 100% right, but it feels like the least wrong. There really needs to be a `shop=telecommunication` as well to cover the retail comcast, spectrum, centurylink type offices vs. their CO type locations.

PR #3120, opened 2019-10-14
Migrate "office/telecommunication" -> "shop/telecommunication"

bhousel:
The "office=telecommunication" tag has been causing confusion.
PRing this one for visibility.  I think switching to a "shop" tag is the right thing to do.

These are telecom companies that do more than just sell mobile phones, but they operate storefronts where people can go in and shop, so they are really a better fit for "shop" than "office".

re: #3113 #2900 and probably others.

### DIFF
--- a/brands/shop/telecommunication.json
+++ b/brands/shop/telecommunication.json
@@ -1,5 +1,5 @@
 {
-  "office/telecommunication|A1": {
+  "shop/telecommunication|A1": {
     "countryCodes": [
       "at",
       "bg",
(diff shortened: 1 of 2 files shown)


tas50:
I always assumed office/telecom was for the COs based on https://wiki.openstreetmap.org/wiki/Telecoms

bhousel:
> I always assumed office/telecom was for the COs based on

I agree - they are a real thing but just not the sort of place a consumer would go to.
It might even make sense to have these in NSI too - I'm not sure.

1ec5:
> It might even make sense to have these in NSI too - I'm not sure.

Central offices should be tagged with `operator` but not necessarily `brand`. By analogy, Kroger has distribution warehouses and marketing offices that would be tagged `operator=Kroger`, not `brand=Kroger`.

Issue #2901, opened 2019-07-17
Add House of Hoops by Foot Locker


yoasif:
eg https://news.nike.com/news/nike-and-foot-locker-unwrap-dream-basketball-store
https://sneakernews.com/2015/01/14/foot-locker-house-hoops-open-largest-location-ever-nyc/
https://stores.footlocker.com/us/ny/new-york/11-west-34th-street.html

This is not exactly Foot Locker, so I think it needs a new entry.

LuisMBarrera:
@yoasif Hey already added this to shop/shoes, waiting for reviews.

LuisMBarrera:
Hey @doublah sorry for dragging you here but I would like to take on your offer of reviewing future PRs, the only problem is that I couldn't add or find how to add reviewers. I'm sorry for the hassle.

PR #3121, opened 2019-10-14
Added Footaction brand

### LINKED ISSUE #2854: Add Footaction
https://stores.footaction.com/index.html

eigenbrot:
Hopefully resolves #2854. 

This is a sub-brand of Foot Locker so I hope I did this correctly. The wiki* entries are both for the parent company (Foot Locker).

If this is the correct way to do this it could also be applied to #2901

### DIFF
--- a/brands/amenity/pub.json
+++ b/brands/amenity/pub.json
@@ -7,16 +7,6 @@
       "name": "Ezītis miglā"
     }
   },
-  "amenity/pub|Wetherspoons": {
-    "countryCodes": ["gb"],
-    "tags": {
-      "amenity": "pub",
-      "brand": "Wetherspoons",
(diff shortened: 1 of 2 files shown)


bhousel:
We do sometimes use the parent entry for a sub-brand.. (like the various Walmart formats).

But because this brand really does have its own logo and looks different, it probably should have its own wikipedia/wikidata identifiers.  It would be kind of weird to display Footaction map markers with the Foot Locker logo.

eigenbrot:
Thanks, @bhousel, I just made a quick-and-dirty wikidata entry for this brand and removed the Foot Locker wikipedia reference.

bhousel:
Looks great, thanks @eigenbrot  👍

Issue #2798, opened 2019-06-15
Burger King has been changed the logo to  "Stranger King"  ( "upside down" marketing campaign )


ImreSamu:
( via - OSM-Talk
"When I try to tag something as the fast-foot chain, Burger King in ID, the preset Burger King logo pops up as an upside down "Stranger King" symbol. "
https://lists.openstreetmap.org/pipermail/talk/2019-June/082725.html )

and it is true  -- the Burger King  logo changed. 
-  https://twitter.com/BurgerKing
-  https://www.facebook.com/burgerking
https://www.facebook.com/burgerking/photos/a.160996107247067/2765720566774595/?type=3&theater

## id Now:
![image](https://user-images.githubusercontent.com/2217320/59557087-c43fdd00-8fd0-11e9-9ac2-52acead9db63.png)

## Twitter - now:
![image](https://user-images.githubusercontent.com/2217320/59557011-da4c9e00-8fce-11e9-83d0-49f1dbcf9de1.png)

## Facebook now:
![image](https://user-images.githubusercontent.com/2217320/59557016-f9e3c680-8fce-11e9-827a-3bfe6670e0b4.png)

bhousel:
Oh yeah, this isn't really wrong, it's just Burger King being playful with their social media accounts.  

This is really  https://github.com/openstreetmap/iD/issues/6361
(on the iD side, we should probably just whitelist the few brands that do this to use their commons logo instead)

1ec5:
> on the iD side, we should probably just whitelist the few brands that do this to use their commons logo instead

We’ll never be able to keep up with these companies’ brand marketing departments. #2641 should make it possible to avoid special-casing individual brands in iD.

PR #3122, opened 2019-10-15
Resolve duplicate brand for CBA

karolyipeter:
Resolve duplicate of Hungarian food retail store CBA, classifying it as a supermarket.

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -223,6 +223,7 @@
     }
   },
   "shop/convenience|CBA": {
+    "nomatch": ["shop/supermarket|CBA"],
     "tags": {
       "brand": "CBA",
       "brand:wikidata": "Q779845",
--- a/brands/shop/supermarket.json
+++ b/brands/shop/supermarket.json
@@ -469,6 +469,7 @@
     }
   },
   "shop/supermarket|CBA": {
+    "nomatch": ["shop/convenience|CBA"],
     "tags": {
       "brand": "CBA",
       "brand:wikidata": "Q779845",


SilentSpike:
To confirm, all locations are definitely supermarkets?

We have other brands in the index which have instances of both supermarkets and convenience stores - in which case it's fine to have both entries and it's up to the end user to pick which is appropriate.

bhousel:
Thanks @karolyipeter  - since it sounds like it can be both a convenience or a supermarket, can you leave both entries and add a `nomatch` property to each one?

See https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#thinking--resolve-warnings  for an example.

ImreSamu:
just documenting: 
* now the CBA facebook logo is  a "marketing campaign"  :(    ( related: https://github.com/osmlab/name-suggestion-index/issues/2798 )
* imho:  we need a flag ;  if the wikidata logo is good quality  or not;  or which logo is the primary. 
   *  if the wikidata logo is good quality - we don't have to use the facebook logo.

## now - iD editor
image

## now: name suggestion index - 
image

bhousel:
> * now the CBA facebook logo is  a "marketing campaign"  :(    ( related: #2798 )

I'll just add this brand to the list on the iD side.  There are only a handful like this.

https://github.com/openstreetmap/iD/blob/56775fe492934e263b33e97d63424b5668f4ec12/build_data.js#L332-L337

bhousel:
Thanks @karolyipeter  the `nomatch` looks good 👍

PR #3123, opened 2019-10-15
brand added in shop/beauty

amous4822:
#2877 
added perfect brows nyc in shop/beauty

### DIFF
--- a/brands/shop/beauty.json
+++ b/brands/shop/beauty.json
@@ -121,7 +121,7 @@
     "countryCodes": ["us"],
     "matchNames": ["Perfect Brows"],
     "tags": {
-      "beauty": "eyebrows;waxing;eyelash;henna tattoo;facials;men's services;",
+      "beauty": "eyebrow;eyelash;waxing",
       "brand": "Perfect Brows NYC",
       "name": "Perfect Brows NYC",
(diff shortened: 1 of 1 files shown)


brands/shop/beauty.json
```diff
@@ -116,5 +116,15 @@
       "name": "Yves Rocher",
       "shop": "beauty"
     }
+  },
+  "shop/beauty|Perfect Brows NYC": {
+    "countryCodes": ["us"],
+    "matchNames": ["Perfect Brows"],
+    "tags": {
+      "beauty": "eyebrows;waxing;eyelash;henna tattoo;facials;men's services;",
```
tas50: We probably want to stick with the common known beauty tags here for this to be useful in search / display later

bhousel: I agree w @tas50 ..

@amous4822 can you check this list here for commonly used `beauty=*` tags:
https://taginfo.openstreetmap.org/keys/beauty#values

amous4822: @tas50 @bhousel I did check. Some of the tags I've added have never been used 😅. I will edit it. Thank you.

bhousel: Thanks! looks good 👍

PR #3124, opened 2019-10-15
Add Wikidata tag for Altar'd State brand

aestevens:
I created a Wikidata Item for the Altar'd State brand and added the newly created Wikidata tag to this index.

Addresses the other part of #3054.


tas50:
Thanks @aestevens

Issue #2856, opened 2019-07-05
Add Mandee


yoasif:
https://www.mandee.com

tas50:
Added to the index in #3125

PR #3125, opened 2019-10-15
Add Mandee brand to index with Wikidata tag

aestevens:
Created a Wikidata Item for this brand and added to the index.

Addresses #2856.


tas50:
Thanks @aestevens

PR #3129, opened 2019-10-16
DRY up the list with additional match_groups

tas50:
Avoid repetition in the match tags by using the match groups.


bhousel:
Thanks @tas50  I think these look ok!

PR #3131, opened 2019-10-16
Added Bank of Melbourne to amenity/bank

orangegrove1955:


brands/amenity/fast_food.json
```diff
@@ -676,6 +669,18 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|Donut King": {
+    "matchNames": ["donut king"],
```
tas50: if this is also only in Australia can you add the country code so we don't show it world wide

bhousel:
Thanks again @orangegrove1955  👍

PR #3135, opened 2019-10-16
Split up Travelodge

doublah:
Change current brand to seperate brands for regions which are owned and operated seperately.


bhousel:
Seems fine 👍

tas50:
TIL. Thanks @doublah

PR #3140, opened 2019-10-17
Update Dressbarn

doublah:
Switches to specific brand wiki pages.

Looking for feedback on name as newer stores and website seem to use both 'dressbarn' and 'Dressbarn', while older stores seem to use 'Dress Barn' on signage.

### DIFF
--- a/brands/shop/clothes.json
+++ b/brands/shop/clothes.json
@@ -869,12 +869,12 @@
       "shop": "clothes"
     }
   },
-  "shop/clothes|Dressbarn": {
+  "shop/clothes|Dress Barn": {
     "tags": {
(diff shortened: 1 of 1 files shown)


Adamant36:
According to [this article](https://www.nytimes.com/2019/05/21/business/dressbarn-closing.html) they are closing all their stores. So the spelling might not matter all that much. It's interesting to note though that even in the picture in that article of a single store front it uses both spellings. Know wonder they had problems. 

21xp-dressbarn-jumbo

doublah:
Switched back to old name as it's not that important and brand still uses that signage, merging new wiki tags.

PR #3144, opened 2019-10-18
Add Reiss clothing brand.

doublah:
International clothing retailer

https://en.wikipedia.org/wiki/Reiss_(brand)

### DIFF
--- a/brands/shop/clothes.json
+++ b/brands/shop/clothes.json
@@ -2600,6 +2600,15 @@
       "shop": "clothes"
     }
   },
+  "shop/clothes|Reiss": {
+    "tags": {
+      "brand": "Reiss",
+      "brand:wikidata": "Q7310479",
+      "brand:wikipedia": "en:Reiss (brand)",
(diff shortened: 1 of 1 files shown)


bhousel:
Hey @doublah - you can just make edits like this on the master branch and push directly to GitHub.. No need to do pull requests for simple stuff. 😁  

Adding each change in a pull request is ok (and some projects prefer people work that way), but it does create a bunch of notifications and extra merge commits in the history.

Most of the time I work on `master` like this:
```bash
git pull origin              #  (sync with latest stuff)
# .. edit stuff.. 
npm run build                #  (make sure the edits are ok)
git diff                     #  (just to double check what was modified)
git add . && git commit      #  (commit the work)
git push origin              #  (push it to github)
```

If the `push` command fails because someone else has pushed - you can `git pull origin --rebase`  which will append your local work onto the end of the commit history.  Then try the `git push origin` again and it will work.

Thanks for all the additions you are making! 🙇

PR #3148, opened 2019-10-19
Add Naturgy, fix Quiksilver, delete duplicates

ArenaL5:
Also removed English names for exclusively Spanish banks (no one uses their English translation), removed some generic names, and fixed some country codes.

I had to create a new file for energy suppliers; please check it's okay.


bhousel:
Thanks!  There are a lot of changes in here, but they overall look ok to me, so I'm just going to merge them.

About the `name:en` and `brand:en` tags - we mostly include these tags because 1. they help us maintain the index, and 2. they help English speakers work with OSM data.  

These tags aren't strictly necessary, so I don't mind if you want to remove the `name:en` from "Banco Sabadell".   But anyway that is the reason that the tag is there, and someone may put it back someday.

The English name matters a lot more for entries in languages that English speakers can't read.. For example:
```js
  "brand": "スターバックス",
  "brand:en": "Starbucks",
  "brand:ja": "スターバックス",
```

Without them, we wouldn't know what brand it is.  So in situations like that, I'd prefer that they stay...

ArenaL5:
Thank _you_. I'll keep it in mind for future commits. I was mostly annoyed at `Shepherd Bank` instead of `Banco Pastor`.

Now that we mention it, is there any way to highlight `brand:en=Shepherd Bank` as "non-standard tag"?

bhousel:
> Now that we mention it, is there any way to tag brand:en=Shepherd Bank and name:en=Shepherd Bank as non-standard?

It would be easier if you just discuss the issue with your local community and then remove all those tags with JOSM or level0 if people decide they don’t want them.  There can’t be more than a few hundred right?

ArenaL5:
You're probably right. It's an endangered brand anyways, so the problem will solve itself sooner or later. (it was technically absorbed by Banco Santander a few months ago, but some places still have the old signage and different ATMs)

PR #3149, opened 2019-10-20
update country codes for amenity/ice_cream|Amorino

amous4822:
the country codes for Amorino brand were updated.
[source of information](https://www.amorino.com/us/shops.html)
Please check.


bhousel:
LGTM! 👍

PR #3150, opened 2019-10-20
Add Red Ribbon Bakeshop to amenity/fast_food

CarlJimenez:


bhousel:
Looks like we already have this brand in the `shop/bakery.json` file..
https://github.com/osmlab/name-suggestion-index/blob/05fb73372755a9d3cd3032fe43bb4d70c32eed98/brands/shop/bakery.json#L545-L554

Adamant36:
cuisine=bakery is a miss use of the cuisine tag anyway. The same way cuisine=coffee_shop is IMHO, but that's another discussion.

CarlJimenez:
Oh. I tried to search it on the site but there was no result. My bad. will delete the branch.

PR #3155, opened 2019-10-24
added serbian sport shop and made some banks avaivable for serbia

MrKovalski:


Adamant36:
I feel like sports themed clothing stores like this, that don't seem to sell any sports equipment, would be better tagged as shop=clothes + clothes=sports. Although, I see Adidas  are already in the sports.json file and their pretty similar.

bhousel:
Thanks @MrKovalski and welcome! 👋 
If they sell some sports equipment, `shop=sports` is fine.

PR #3153, opened 2019-10-23
Assorted changes

ArenaL5:
Like the last pull request: I added some Spanish brands, and removed some generic names. Not speaking actual Russian, I resorted to heavy use of Wikipedia. (The names I dared to remove were direct translations of «polyclinic» and «clinic», or similar.) This removes some bogus warnings about probable duplicate entries.

I also added `cuisine=swedish` to IKEA Restaurant.

PR #3156, opened 2019-10-25
Added University in Indonesia and Fixed Missing Tags

adzkar:
I have added an university brand in indonesia and fixed some missing tags


Adamant36:
@adzkar see issue #3153 about adding the cuisine tag to the Ikea Restaurant entry. Also, see https://github.com/osmlab/name-suggestion-index/pull/3090#issuecomment-539713396 in relation to adding universities to the index.

bhousel:
Thanks but we don't add traditional universities to this index, only those that operate as branded businesses.

PR #3159, opened 2019-10-25
Update massage.json

mintdawn:
added hand and stone massage to shop/massage


bhousel:
Thanks @mintdawn and welcome! 👋

PR #3177, opened 2019-10-28
Expand Hilton and Wyndham brand hotels

tas50:
More hotel updates based on company brand lists and their full "by whatever" names.


Adamant36:
Hhhmmm, I really feel like merging this......You think it's good to go @tas50 ?

tas50:
Oh yep I totally forgot about this. It's ready

PR #3180, opened 2019-10-28
Add Livraria Saraiva

marceloabk:
https://pt.wikipedia.org/wiki/Livraria_Saraiva

### DIFF
--- a/brands/shop/books.json
+++ b/brands/shop/books.json
@@ -215,7 +215,7 @@
   "shop/books|Livraria Saraiva": {
     "countryCodes": ["pt"],
     "tags": {
-      "brand": "Livraria Saraiva",
+      "brand": "Saraiva",
       "brand:wikidata": "Q10319337",
       "brand:wikipedia": "pt:Livraria Saraiva",
       "name": "Livraria Saraiva",
--- a/brands/shop/books.json
+++ b/brands/shop/books.json
@@ -212,6 +212,16 @@
       "shop": "books"
     }
   },
+  "shop/books|Livraria Saraiva": {
+    "countryCodes": ["pt"],
+    "tags": {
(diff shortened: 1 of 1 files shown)


brands/shop/books.json
```diff
@@ -212,6 +212,16 @@
       "shop": "books"
     }
   },
+  "shop/books|Livraria Saraiva": {
```
tas50: From their website and signage itit looks like this is just be Saraiva. Does that seem right? I don't see any references to the full name.

marceloabk: Livraria translates into bookstore. Call it just Saraiva sounds kinda loose. Normaly when we go to the mall we see it as Livraria Saraiva, but I think that at the brand line I should change to just Saraiva. 

We can found some references to the full name at the wikipedia title, at a online phone list and a search at google for Saraiva we can see some malls calling it Livraria Saraiva.

But I don't have a problem changing it to Saraiva if you found that is better that way haha

mds08011: My own search for photos of the store shows mostly "Saraiva" signs.  I think it should be entered as "Saraiva" unless you know that it is regularly shown as "Livraria Saraiva" in the stores.

bhousel:
Thanks @marceloabk!  

It really looks like they are committed to the "Saraiva" branding by looking at images of stores: 
https://www.google.com/search?q=livraria+saraiva&source=lnms&tbm=isch&sa=X&ved=0ahUKEwj9qJnThMTlAhWHMd8KHSwYB2kQ_AUIFygA&biw=1437&bih=766

So I ended up using "Saraiva" for the `name` and `brand` but preserved "Livraria Saraiva" in the `alt_name` tag, since it sounds like locals also know it as that.

PR #3187, opened 2019-10-28
Added brand proof of car manufacturer "Jaguar" from wikipedia

Tck13:


brands/shop/car.json
```diff
@@ -164,6 +164,7 @@
     "tags": {
       "brand": "Jaguar",
       "brand:wikidata": "Q26742231",
+      "brand:wikipedia": "en:Jaguar_Cars",
```
bhousel:
```suggestion
      "brand:wikipedia": "en:Jaguar Cars",
```

bhousel:
Looks great @Tck13 , thanks for the quick fixes!

brands/shop/car.json
```diff
@@ -164,6 +164,7 @@
     "tags": {
       "brand": "Jaguar",
       "brand:wikidata": "Q26742231",
+      "brand:wikipedia": "en:Jaguar Cars",
```
1ec5: 0dd5929e5753422f29f7f503707b153458eb9b30 removes this tag, because the “Jaguar Cars” article is about the formerly independent manufacturer, whereas Q26742231 is about the marque that continues to be used by Jaguar Land Rover.

PR #3189, opened 2019-10-28
Update Microsoft Stores in electronics.json

ritika99:


tas50:
Thanks @ritika99. I removed the extra { in the JSON that was causing failures.

PR #3190, opened 2019-10-28
Updated Dr. Martens shoes 

ritika99:


bhousel:
Thanks @ritika99 but I'm going to accept #3191 since it doesn't have the syntax error.

Issue #3175, opened 2019-10-27
Add Dr. Martens shoe stores


tas50:
Worldwide shoe store chain: https://en.wikipedia.org/wiki/Dr._Martens

denisolvr:
I added a pull request for that, can you have a look =)

PR #3191, opened 2019-10-28
Add Dr. Martens

denisolvr:


denisolvr:
Pull request for issue #3175

brands/shop/shoes.json
```diff
@@ -963,5 +963,16 @@
       "name:ja": "東京靴流通センター",
       "shop": "shoes"
     }
+  },
+  "shop/shoes|Dr. Martens": {
+    "countryCodes": ["uk"],
```
bhousel: I'd remove this line, since they operate retail stores in places other than just the `uk`.
(also, a thing that we get wrong sometimes, the [ISO country code](https://www.nationsonline.org/oneworld/country_code_list.htm) for uk is actually `gb`)

tas50: These shops are actually all over so you can just remove it entirely

bhousel:
I removed the countrycodes and added a few more reasonable matchNames in 6932af1c7

Issue #3171, opened 2019-10-27
Add Microsoft Store


tas50:
Electronics store with 116 stores around the world: https://en.wikipedia.org/wiki/Microsoft_Store

bhousel:
this merged in 297565b35

PR #3192, opened 2019-10-28
Add Microsoft Store #3171

FabioMassaretto:
Please let me know if everything is ok because it's my first PR on a project


bhousel:
Sorry  `"countryCodes": ["en"],` isn't correct.. 

The 2 letter ISO 3166 codes can be found here:
https://www.nationsonline.org/oneworld/country_code_list.htm
If the brand operates worldwide, we don't really need to add it.

FabioMassaretto:
> Sorry `"countryCodes": ["en"],` isn't correct..
> 
> The 2 letter ISO 3166 codes can be found here:
> https://www.nationsonline.org/oneworld/country_code_list.htm
> If the brand operates worldwide, we don't really need to add it.

My bad, Thank you for the tip! I saw that the code is already correct, should I need to do something?

Issue #3165, opened 2019-10-27
Add Motherhood Maternity


tas50:
Very common maternity clothing store. Part of https://en.wikipedia.org/wiki/Destination_Maternity brand portfolio

bhousel:
this landed in #3194

PR #3194, opened 2019-10-28
add brand wikidata and wikipedia for Motherhood Maternity

AlbyIanna:
solves #3165

### DIFF
--- a/brands/shop/clothes.json
+++ b/brands/shop/clothes.json
@@ -2093,6 +2093,8 @@
     "countryCodes": ["ca", "us"],
     "tags": {
       "brand": "Motherhood Maternity",
+      "brand:wikidata": "Q5265143",
+      "brand:wikipedia": "en:Destination Maternity",
       "clothes": "maternity",
       "name": "Motherhood Maternity",
       "shop": "clothes"


bhousel:
Realizing now that for this one we should probably push for separate wikidata for
* Motherhood Maternity
* A Pea in the Pod
* Destination Maternity

instead of just using the parent company's wikidata..

AlbyIanna:
ok, but in that case we'd have wikidata and wikipedia for Destination Maternity only

tas50:
@AlbyIanna Creating the wikidata for Motherhood Maternity is probably the best bet here. We can just leave the wikipedia line out since that would just point to the parent company.

AlbyIanna:
> Creating the wikidata for Motherhood Maternity is probably the best bet here.
@tas50 How can I do that?

matkoniecz:
See https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#creating-wikidata-entries

Adamant36:
Motherhood Maternity: Q79051139
A Pea in the Pod: Q79051737
Destination Maternity: Q79051770

@AlbyIanna, you want to update the PR?

bhousel:
> @AlbyIanna, you want to update the PR?

Thanks @Adamant36  - I'll merge as-is and apply the updates 👍

Issue #3173, opened 2019-10-27
Add Abercrombie Kids


tas50:
Offshoot of Abercrombie. We already have all the wikidata for that.

PR #3195, opened 2019-10-28
add brand for Abercrombie Kids

AlbyIanna:
solves #3173


bhousel:
This looks great, thanks so much @AlbyIanna  👍

PR #3210, opened 2019-10-29
Added Van Heusen clothing store

rashmiap:


brands/shop/clothes.json
```diff
@@ -1572,6 +1572,15 @@
       "shop": "clothes"
     }
   },
+  "shop/clothes|Jockey": {
+    "tags": {
+      "brand": "Jockey International",
```
bhousel: I would change `brand` and `name` here to just say "Jockey".  These are more for the name that that they use on the storefronts, not really the official name of the company.

Adamant36:
I decided just to merge it and fixed the name/brand in https://github.com/osmlab/name-suggestion-index/commit/bc30ca983732670fc406b8452c2eb723a7c7f150. Thanks @rashmiap

PR #3211, opened 2019-10-29
Add BonWorth clothing brand

### LINKED ISSUE #3204: Add BonWorth women's clothing store
https://en.wikipedia.org/wiki/BonWorth

tomanistor:
Added BonWorth brand.

### DIFF
--- a/brands/shop/clothes.json
+++ b/brands/shop/clothes.json
@@ -346,7 +346,6 @@
   },
   "shop/clothes|BonWorth": {
     "countryCodes": ["us"],
-    "nomatch": ["shop/clothes|BonWorth"],
     "tags": {
       "brand": "BonWorth",
       "brand:wikidata": "Q4941034",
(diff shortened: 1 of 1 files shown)


brands/shop/clothes.json
```diff
@@ -344,6 +344,18 @@
       "shop": "clothes"
     }
   },
+  "shop/clothes|BonWorth": {
+    "countryCodes": ["us"],
+    "nomatch": ["shop/clothes|BonWorth"],
```
bhousel: Can you please remove this `nomatch` line - it matches this entry!

The "nomatch" property is used to suppress duplicate warnings.  It would be used if there were a different entry with the same name and we need to let the matcher know that they are two different things.  (For example a gas station brand also named "BonWorth").

tomanistor: Sorry about that, I'll get it sorted!

tomanistor:
I've removed the nomatch line. Thanks!

bhousel:
Thanks for the quick fix @tomanistor, and welcome!  👋

Issue #3168, opened 2019-10-27
Add Nestlé Toll House Café


tas50:
Cafe with a good number of locations in the US https://nestlecafe.com/locations https://en.wikipedia.org/wiki/Nestl%C3%A9_Toll_House_Caf%C3%A9

PR #3216, opened 2019-10-29
added in Nestlé Toll House Café fast_food restaurant

TomEWilkinson:
as suggested in #3168

links to restaurant 
https://nestlecafe.com/locations https://en.wikipedia.org/wiki/Nestl%C3%A9_Toll_House_Caf%C3%A9


TomEWilkinson:
good shout! just changed that over

tas50:
Thanks @TomEWilkinson

PR #3217, opened 2019-10-29
Add Little Burgundy shoe store

### LINKED ISSUE #3201: Add Little Burgandy shoe store
Shoe store in Canada: https://www.littleburgundyshoes.com

josephkmh:
Fixes #3201

### DIFF
--- a/brands/shop/shoes.json
+++ b/brands/shop/shoes.json
@@ -461,6 +461,14 @@
       "shop": "shoes"
     }
   },
+  "shop/shoes|Little Burgundy" : {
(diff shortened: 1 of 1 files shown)


bhousel:
Looks good @josephkmh, thanks and welcome! 👋 

This won't show up in iD until we add a Wikidata identifier, but it looks like there isn't one yet.

This page lists the sub-brands owned by Genesco, so would be a good place to start for anyone who wants to add the info to Wikipedia/Wikidata: https://en.wikipedia.org/wiki/Genesco#Brands

PR #3221, opened 2019-10-30
Add shoes store Dr. Martens

ghousshah:


bhousel:
Thanks @ghousshah but we already have #3191 for this

PR #3224, opened 2019-10-30
Added Perfumania

quetcodesfire:
I added Perfumania to the list here. They do not have a wikipedia so I have left the wikidata and wikipedia fields as empty strings.

### DIFF
--- a/brands/shop/perfumery.json
+++ b/brands/shop/perfumery.json
@@ -71,5 +71,19 @@
       "name": "O Boticário",
       "shop": "perfumery"
     }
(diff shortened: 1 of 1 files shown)


Adamant36:
Thanks @quetcodesfire. I added a Wikidata entry in https://github.com/osmlab/name-suggestion-index/commit/1cd6c09a8df4d3b4c30c4a6b5df7e049eb849398. For future reference if there is no Wikipedia or Wikidata pages you just leave those sections out of the code all together.

PR #3225, opened 2019-10-30
Added Van Heusen

quetcodesfire:
Added data for Van Heusen. There was no wikipedia data to add.

### DIFF
--- a/brands/shop/clothes.json
+++ b/brands/shop/clothes.json
@@ -3299,6 +3299,14 @@
       "shop": "clothes"
     }
   },
+  "shop/clothes|Van Heusen": {
+    "countryCodes": ["us"],
+    "tags": {
+      "brand": "Van Heusen",
+      "name": "Van Heusen",
+      "shop": "clothes"
+    }
+  },
   "shop/clothes|V.I.M.": {
     "countryCodes": ["us"],
     "tags": {


brands/shop/clothes.json
```diff
@@ -3299,6 +3299,14 @@
       "shop": "clothes"
     }
   },
+  "shop/clothes|Van Heusen": {
+    "countryCodes": ["us"],
+    "tags": {
+      "brand": "Van Heusen",
+      "name": "Van Heusen",
+      "shop": "clothes"
```
tas50: This will need a wikidata entry created so it will show up in the editors. You can usually look at one of the entries for a similar business to see useful keys to enter. One of the big ones for us is the social media entries since that powers the logos.

quetcodesfire: I'm not quite sure if I'm understanding wikidata 100%. I looked at a few other brands and it seemed like they had Wikipedia pages and wikidata. Any brand that did not have a Wikipedia page I was not able to find wikidata for. I can take a look at Van Heusen's social media logos to see if there is any kind of wikidata associated with them if that is a source for brands without Wikipedia pages.

Adamant36: @tas50, Q67551865 wouldn't work? The link to their website on their Wikidata entry has a section for locating stores. So, I assume its the same thing and that we can just use the normal Wikidata entry.

Adamant36: Btw @quetcodesfire, I noticed that a lot of the things your doing PR's for already had ones open by other people. You should really check to make sure there aren't already PR's open for things. Otherwise, we might miss the earlier ones like I did when I merged all yours and that's not so good for the people that technically did them first who don't get theirs merged now. 

Thanks

Adamant36:
Since this already had PR #3210 open for it a few days ago I am closing this one. As it's redundant. I suggest the discussion about if to use the Wikidata entry or not be moved there. Thanks.

brands/shop/clothes.json
bhousel: Q67551865 looks ok to me

PR #3226, opened 2019-10-30
Added Stuart Weitzman Shoe Store

quetcodesfire:
Added Stuart Weitzman Shoe Store. There was a wikipedia page on the owner, but not the business itself and it was not added.


Adamant36:
Added in #3227

PR #3228, opened 2019-10-31
add windsor

rbezerra:


brands/shop/clothes.json
```diff
@@ -3455,6 +3455,14 @@
       "shop": "clothes"
     }
   },
+  "shop/clothes|Windsor": {
+    "countryCodes": ["us"],
+    "tags": {
+      "brand": "Windsor",
+      "name": "Windsor",
```
tas50: For this to appear in iD and other editors it will also need a wikidata entry. If that doesn't' exist you can create one. Usually the best thing is to find a similar brand to see what tags you might want to add in wikidata. Let me know if you need any help with that. I'm around here on Github and also on OSM US Slack.

Adamant36:
Created and added in https://github.com/osmlab/name-suggestion-index/commit/fd29b0d609593e8790c61b909796bcb10f2f661a. Thanks @rbezerra 👍

PR #3229, opened 2019-10-31
Rename amenity/bank/OTP to OTP Bank

KAMI911:
Hi!

Based on Hungarian OSM Group discussion
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/openstreetmap-hungary/Op0Cju-f0mM/2hsN7ldNCQAJ
, real name of the bank and Wikidata: We would like to change OTP to OTP Bank in the name tag.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -4348,7 +4348,7 @@
       "brand": "OTP",
       "brand:wikidata": "Q912778",
       "brand:wikipedia": "en:OTP Bank",
-      "name": "OTP"
+      "name": "OTP Bank"
     }
   },
   "amenity/bank|Oberbank": {


Adamant36:
Seems like its in small letters on the signs and there is no space between the words. What about that? (usually we go with what's on the sign). 

33_default_article

bkil:
Thanks for the quick response. It is true for the graphic logo design, but if you also look at the website, the textual type setting goes as "OTP Bank".

See at the end of
https://www.otpbank.hu/
`© Copyright 2019 OTP Bank`

Adamant36:
> See at the end of
https://www.otpbank.hu/
© Copyright 2019 OTP Bank

Going by legal title (which I is what I would consider a copyright mark), over signage in the index was already discussed and rejected. The important thing is that it matches what's on the ground. Per the OSM ground verifiability rule. As such, there's lots of entries where the official company name doesn't fit with the indexes entry. I see no reason why this should be different. It's not like matters anyway since it essentially still says the same thing either way, but at least what I'm suggesting fits the on the ground rule. 

Btw, they also go by OTP Bank Nyrt [here](https://www.otpbank.hu/static/portal/sw/file/XS0274147296_OC.pdf) in official documents and are also referenced as such [here](https://www.bloomberg.com/profile/company/OTP:GR). Along with OTP Bank Group in other places...So, using the ground avoids having to figure out which one is best. Otherwise, someone could just as easily come along later and say it should be another way "just because." At least that's how I see it and like I've said going by the legal entity name over signage has already been rejected anyway. Maybe another collaborator will disagree though.

bhousel:
> So, using the ground avoids having to figure out which one is best.

I would usually go with what the signs say, but in this case the local community has discussed it, and they can decide how they want their banks to look, so we should just go with whatever they want..

PR #3245, opened 2019-10-31
delek brand  added 

Bitha007:


Adamant36:
Thanks @Bitha007. Looks good. Except we usually only use the English name or the name of the company in whatever language it is in. So, if you could remove all the languages and just leave `"name": "Delek"` that would be great. Thanks.

bhousel:
thanks @Bitha007 and welcome! 👋

Issue #3246, opened 2019-10-31
Add Murco


Adamant36:
UK based gas station chain 

https://en.wikipedia.org/wiki/Murco_Petroleum
https://www.wikidata.org/wiki/Q16998281

Adamant36:
Closed in #3252

PR #3252, opened 2019-10-31
added brand Mohawk

KungFuryKeyboard:


Adamant36:
Thanks. Looks fine. Except you don't need a match name for Mohawk since its what is already in the name tag.

KungFuryKeyboard:
ok I've removed the extra match tag, however i've accidently also added code for murco #3246

Adamant36:
That's fine. Thanks for fixing it and for the contributions 👍

PR #3254, opened 2019-10-31
Update CONTRIBUTING.md

Surya1231:


Adamant36:
Ummm, no thanks. Why not try contributing something that's actually helpful?

Issue #3244, opened 2019-10-31
Add Huck's


Adamant36:
American convenience store chain

https://en.wikipedia.org/wiki/Huck_Food_Fuel
https://www.wikidata.org/wiki/Q56276328

theonlymikeever:
I can help with this!

PR #3262, opened 2019-10-31
#3244 Add Huck's

theonlymikeever:
Added data for Huck Food Fuel

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -1,4 +1,15 @@
 {
+  "shop/convenience|Huck Food Fuel": {
+    "countryCodes": ["us"],
+    "matchNames": ["Huck's"],
+    "tags": {
+      "brand": "Huck Food Fuel",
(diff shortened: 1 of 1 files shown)


brands/shop/convenience.json
```diff
@@ -1,4 +1,15 @@
 {
+  "shop/convenience|Huck Food Fuel": {
+    "countryCodes": ["us"],
+    "matchNames": ["Huck's"],
+    "tags": {
+      "brand": "Huck Food Fuel",
+      "brand:wikidata": "Q56276328",
+      "brand:wikipedia": "en:Huck_Food_Fuel",
```
bhousel: Looks like this should actually point at..
```suggestion
      "brand:wikipedia": "en:Huck's Food & Fuel",
```

bhousel:
Thanks @theonlymikeever 👍 
We ended up with 2 of them between #3262 and #3267, so I cleaned things up in f59603bc3

Issue #3249, opened 2019-10-31
Add RaceWay gas stations


Adamant36:
American company. Also has convenience stores.
https://www.racewaystores.com/

PR #3269, opened 2019-10-31
Add RaceWay to brands

madesch:
Solves #3249

### DIFF
--- a/brands/amenity/fuel.json
+++ b/brands/amenity/fuel.json
@@ -2061,6 +2061,7 @@
     "tags": {
       "amenity": "fuel",
       "brand": "RaceWay",
+      "brand:wikidata": "Q73039084",
       "brand:wikipedia": "en:Raceway",
       "name": "RaceWay"
     }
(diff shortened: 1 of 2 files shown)


madesch:
I see - I've created a [wikidata page](https://www.wikidata.org/wiki/Q73039084) which I can add to the json, will that suffice?

bhousel:
> I see - I've created a [wikidata page](https://www.wikidata.org/wiki/Q73039084) which I can add to the json, will that suffice?

Yes that's excellent!   I'll try to add some more information to that page to ensure that the Wikidata admins keep it around 👍 

Could you also delete the `brand:wikipedia` tags - they still point to that disambiguation page and that's not completely correct.  Someday if somebody creates a wikipedia page like `"en:Raceway (brand)"`  we can point them to the correct page.  The most important thing is that the `brand:wikidata` field has a good value, so we can identify them all later...

madesch:
Sure thing, thanks for the input! :)

bhousel:
Thanks @madesch for the quick fixes and for adding the Wikidata page.. Welcome! 👋

Issue #3235, opened 2019-10-31
Add Azpetrol


Adamant36:
Azerbaijani fuel station company

https://en.wikipedia.org/wiki/Azpetrol
https://www.wikidata.org/wiki/Q4034661

PR #3273, opened 2019-10-31
Add Azpetrol

cmppoon:
ref #3235


bhousel:
Thanks @cmppoon and welcome! 👋 
This looks pretty good -  in 9cf7349d6 I switched it to point to the Azerbaijani wikipedia page, also removed the `official_name` since it wasn't substantially different from the `name`.

PR #3274, opened 2019-10-31
Added to amenity/bank: Banco de Poupança e Crédito

### LINKED ISSUE #3231: Add BPC
Angolan bank

https://en.wikipedia.org/wiki/Banco_de_Poupança_e_Crédito
https://www.wikidata.org/wiki/Q4854132

roseg43:
Resolves #3231

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -876,6 +876,16 @@
       "official_name": "Bank of Queensland"
     }
   },
+  "amenity/bank|BPC": {
+    "countryCodes": ["ao"],
+    "tags": {
+      "amenity": "bank",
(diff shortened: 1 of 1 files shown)


Adamant36:
Hi there. Thanks for the PR. It's probably better to go with BPC for the brand and name fields so it shows up on the map good. Then Banco de Poupança e Crédito can be used with the official_name tag if needed. That's usually how we do it for excessively long brand names like this one.

roseg43:
Not a problem @Adamant36 ! I've amended the entry to make the name and brand the short name, and added an official_name.

Adamant36:
Looks good. Thanks @roseg43 👍

Issue #3280, opened 2019-10-31
Add Din Tai Fung


Adamant36:
Global Chinese restaurant chain

https://en.wikipedia.org/wiki/Din_Tai_Fung
https://www.wikidata.org/wiki/Q699142

Grimlock257:
I'll take this!

trevorspecht:
working on this one

mds08011:
https://github.com/osmlab/name-suggestion-index/commit/f022136bb1da7888a4415cfb896d077fe7d7d428

PR #3290, opened 2019-10-31
Added Din Tai Fung to restaurants

roseg43:
For #3280

### DIFF
--- a/brands/amenity/restaurant.json
+++ b/brands/amenity/restaurant.json
@@ -619,6 +619,34 @@
       "name": "Dickey's Barbecue Pit"
     }
   },
+  "amenity/restaurant|Din Tai Fung": {
+    "countryCodes": [
+      "ae",
+      "au",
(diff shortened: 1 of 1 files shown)


Adamant36:
Not sure about the name tags on this one. Usually we just go with English and whatever language the sign is in. So probably it would be best to get rid of name:jp since the signs look like they are in Chinese characters.

roseg43:
@Adamant36 I added the names that were included on the wikipedia entry. name:zh is using Simplified Chinese characters, and name:jp is using Kana (Kanji is available as well). If you think these should be changed or removed, just let me know and I'm happy to adjust.

Adamant36:
It's fine. I just merged it and fixed it in https://github.com/osmlab/name-suggestion-index/commit/f022136bb1da7888a4415cfb896d077fe7d7d428. Thanks for the PR.

PR #3292, opened 2019-11-01
Added Beechworth Bakery

BenR31415:
Added Beechworth Bakery, a bakery chain in Victoria AU


bhousel:
Looks great, thanks @BenR31415  👍

PR #3296, opened 2019-11-01
Added Supabarn Supermarkets, added information to four square

BenR31415:
Added wikidata, wikipedia and country codes to four square.

Added Supabarn to list of new supermarkets.

### DIFF
--- a/brands/shop/supermarket.json
+++ b/brands/shop/supermarket.json
@@ -1421,8 +1421,11 @@
     }
   },
   "shop/supermarket|Four Square": {
+    "countryCodes": ["nz"],
     "tags": {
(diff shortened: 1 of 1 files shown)


Adamant36:
Thanks. You could probably get rid of the "Supermarkets" parts in the Supabarn entry though and just add Supabarn Supermarkets with the official_name tag instead. That's usually what we do with long names where the last word is more descriptive. That way it will show up better on the map.

Adamant36:
I just did it in https://github.com/osmlab/name-suggestion-index/commit/f92ae0a9b76a9649c511f478207e2015c1009646. Thanks for the PR @BenR31415 👍

PR #3324, opened 2019-11-01
Adding a gym to the list

beermandev:


Adamant36:
Hi there. Thanks for the PR and welcome. Three things, 

1. It looks like you have an error from deleting the } symbols that are above the entry. So that needs to be fixed 

2. I noticed you made a Wikidata entry for Movement Climbing Gym but your using the Wikipedia entry for The Gym Group. In cases where there the Wikipedia article is for a parent company, which I assume is the case here, and the Wikidata entry is the brand we just leave out the Wikipedia entry altogether. It's not really necessary anyway. The Wikidata entry is the important thing. 

3. Do you know what the website for Movement Climbing Gym is, where they are located, or how many gyms they have? I ask because all I could find with an internet search that looked semi similar is a place in Boulder with 3 gyms called Movement Climbing + Fitness. If that's the place, it's not really enough locations to be included in the index, let alone to meet Wikidata's guidelines of notability. As far as the index goes, we usually go with brands with over 50 locations (although sometimes its lower, but not that low). 

Thanks

bhousel:
I took a look at this more, and I agree that we shouldn't include a gym with only 3 locations around Denver.  (However it would be totally ok to add this stuff to Wikipedia and Wikidata, and then map the 3 locations in OSM and add the `brand:wikidata` and `brand:wikipedia` tags by hand).

PR #3325, opened 2019-11-01
added new car.json file

Romilj012:
added new car.json file


tas50:
This probably belongs here https://github.com/osmlab/name-suggestion-index/blob/master/brands/shop/car.json

bhousel:
Thanks @Romilj012 for the quick fix, and welcome! 👋

PR #3328, opened 2019-11-01
Fix issue where countryCodes property could be called on an undefined match

quincylvania:
Ran into this while working on https://github.com/openstreetmap/iD/issues/6479. This fix will need to be released before I can address the other issue.


bhousel:
Thanks @quincylvania 👍

PR #3329, opened 2019-11-01
Adding Oregon convenience store (Dari Mart) with about 40 stores.

b-jazz:


brands/shop/convenience.json
```diff
@@ -507,6 +507,16 @@
       "shop": "convenience"
     }
   },
+  "shop/convenience|Dari Mart": {
+    "tags": {
+      "brand": "Dari Mart",
+      "brand:website": "https://www.darimart.com/",
```
mds08011:
```suggestion
```

mds08011: We haven't been including website tags with other entries so I recommend removing this.

b-jazz: It seems like a perfect way to add some extra data to the objects, but I think following convention is more important, so I've removed the line.

mds08011: Agreed.  I’d say this is worth discussing in the #poi channel of the OSM-US Slack.

mds08011:
Thank you!

Issue #3320, opened 2019-11-01
Add The Counter


Adamant36:
Global burger restaurant brand

 https://en.wikipedia.org/wiki/The_Counter
https://www.wikidata.org/wiki/Q7727763

Adamant36:
Closed in #3330

PR #3330, opened 2019-11-02
Add The Counter Restaurant

adzkar:
i have added the counter restaurant related on this issues #3320


Adamant36:
Thanks. Looks good.

Issue #3318, opened 2019-11-01
Add Harry Ramsden's


Adamant36:
Seafood fast food restaurant chain in the United Kingdom and Ireland

https://en.wikipedia.org/wiki/Harry_Ramsden%27s
https://www.wikidata.org/wiki/Q5671849

lyndic:
Can I work on this?

Adamant36:
Hi, @lyndic. Sure, go for it. If you have any questions about anything feel free to ask.

Adamant36:
Closed in #3335

PR #3335, opened 2019-11-04
Add Harry Ramsden's to fast_food.json

lyndic:
For #3318

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -1020,6 +1020,18 @@
       "takeaway": "yes"
     }
   },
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @lyndic that looks great - and welcome! 👋 

I switched the cuisine to `fish_and_chips` in 09246c121 since it looks like this is what they serve, and that is the best fitting OSM tag I could find on [taginfo](https://taginfo.openstreetmap.org/keys/cuisine#values).

lyndic:
Thank you for letting me know where I could improve and for sharing the proper cuisine!

PR #3336, opened 2019-11-04
add Grill'd burger restaurant

andrewharvey:


bhousel:
Thanks @andrewharvey 🍔 - but it looks like we already have this entry under `amenity/fast_food`:

https://github.com/osmlab/name-suggestion-index/blob/3c02052322b485b947a8ae169972acc114b2e3ab/brands/amenity/fast_food.json#L984-L995

andrewharvey:
Oh yeah, that entry is fresh off the grill..

PR #3337, opened 2019-11-05
Creating police section for major Canadian police forces

bkissin:
The RCMP (Royal Canadian Mounted Police) is the police force in most of Canada, except for Ontario and Quebec, which both have their own provincial police forces. Edits were made in the online client, so apologies in advance for possible syntax errors. Getting back on this project after a while away.

### DIFF
--- a/brands/amenity/police
+++ b/brands/amenity/police
@@ -1,0 +1,33 @@
+{
+  "amenity/police|RCMP": {
+    "countryCodes": ["ca"],
+    "tags": {
+      "amenity": "police",
+      "brand": "Royal Canadian Mounted Police",
+      "brand:wikidata": "Q335175",
+      "brand:wikipedia": "en:Royal Canadian Mounted Police",
+      "official_name": "Royal Canadian Mounted Police",
+      "official_name:fr": "Gendarmerie royale du Canada,
+    }
+  }
+  "amenity/police|OPP": { 
+    "countryCodes": ["ca"],
+    "tags": {
+      "amenity": "police",
+      "brand": "Ontario Provincial Police",
+      "brand:wikidata": "Q1046157",
+      "brand:wikipedia": "Ontario Provincial Police",
+      "official_name": "Ontario Provincial Police",
+     }
+   } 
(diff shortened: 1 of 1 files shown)


bhousel:
This is a good idea, but I'm not quite ready for this.. 

Would police be more of a `brand` or `operator` ?  We're currently trying to stick to brands, but I want to expand to operators (to better handle things like public transport and hospital networks - police are probably in here too).  I know the distinction in OSM is really thin.

doublah:
`operator` seems more appropriate for Police and other non-retail/business tags in my opinion

bkissin:
No worries! If we want to hold off on making this change until ready, I am fine with that.

brands/amenity/police
```diff
@@ -0,0 +1,33 @@
+{
+  "amenity/police|RCMP": {
+    "countryCodes": ["ca"],
+    "tags": {
+      "amenity": "police",
+      "brand": "Royal Canadian Mounted Police",
+      "brand:wikidata": "Q335175",
+      "brand:wikipedia": "en:Royal Canadian Mounted Police",
+      "official_name": "Royal Canadian Mounted Police",
```
anandfresh: Shall we stick to camelCase for property names ?

bhousel: Yes

brands/amenity/police
```diff
@@ -0,0 +1,33 @@
+{
+  "amenity/police|RCMP": {
+    "countryCodes": ["ca"],
+    "tags": {
+      "amenity": "police",
+      "brand": "Royal Canadian Mounted Police",
```
matkoniecz, line 6: Is it really a brand? I would consider switch to `operator`

bhousel: well yes, that is the whole reason this PR has sat since November

matkoniecz: I know, I wanted to ping author.

bhousel:
Going to close this for now.  

Restructuring the code to allow `operator` or `network` to be the the defining tag (instead of `brand`) is in the back of my mind and will happen eventually.

PR #3338, opened 2019-11-05
Adding wikidata information for America's Best Value Inn

bkissin:
Added the Wikidata/Wikipedia entry for America's Best Value Inn. The Wikipedia page redirects to its parent company, which is the target of the wikidata entry.


bhousel:
Oh yeah, I'd rather see separate brand pages set up for the sub-brands of Vantage Hospitality.  It would be weird to show the Vantage logo for these hotels.  Let's hold off for now until the wikidata pages are available.
https://en.wikipedia.org/wiki/Vantage_Hospitality

PR #3339, opened 2019-11-05
Adding Boys & Girls Club

bkissin:
If consensus suggests that this should be located in Amenity="social_facility", I can go ahead and put it there.


Adamant36:
> If consensus suggests that this should be located in Amenity="social_facility", I can go ahead and put it there.

That's already where it's at. Btw, it looks like Travis didn't like it for some reason. So you might want to double check the code. Thanks.

bhousel:
Looks great @bkissin thank you 👍 
I fixed the JSON syntax error in 9972684f4

Issue #3344, opened 2019-11-06
Hampton By Hilton


Djminisite06:
Hampton should have country tags removed as it is now a worldwide brand

Adamant36:
Well, not technically world wide. According to Wikipedia it's in the United States, Canada, United Kingdom, Latin America and Europe. It really depends on how many European and Latin American countries its in as to if the country codes should be removed completely or not IMHO. We have a few brands in the index with like 20 country codes. There's no reason we couldn't do it here so it come up in searches done in Africa, Australia, or the middle east. Which is still a huge part of the world.

bhousel:
I'd like to start using [country-coder](https://github.com/ideditor/country-coder) in more projects.  It some support for regions, so we wouldn't need to specify all the countries.  For now let's just remove the `us` countryCode, and we can improve on this later.

Adamant36:
Alright. I wasn't aware of country-coder. Sounds like it will be awesome. Closed in  #3346.

PR #3346, opened 2019-11-07
fix: remove country Hampton issue #3344

arszh:
issue #3344


Adamant36:
I'm not sure if it should removed or expanded instead. See https://github.com/osmlab/name-suggestion-index/issues/3344#issuecomment-550991055

Adamant36:
I guess removel is fine. Thanks for the PR.

Issue #3348, opened 2019-11-07
Add QuickChek gas stations


Adamant36:
It already has a convenience store entry. So the no matches will probably have to be figure out. 

https://en.wikipedia.org/wiki/QuickChek
https://www.wikidata.org/wiki/Q7271689

bhousel:
done in #3352

Issue #3347, opened 2019-11-07
Add TOP


Adamant36:
Irish gas station chain 

https://en.wikipedia.org/wiki/Tedcastles_Oil_Products
https://www.wikidata.org/wiki/Q7693933

bhousel:
done in #3352

PR #3352, opened 2019-11-08
update: add  gas stations

arszh:
update gas stations data from issues #3347 and  #3348

### DIFF
--- a/brands/shop/gas.json
+++ b/brands/shop/gas.json
@@ -38,5 +38,27 @@
       "name": "Ferrellgas",
       "shop": "gas"
     }
+  },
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @arszh  - this is pretty good.. I moved the entries to `amenity/fuel` in 5a39acf53
`shop/gas` is really more for places that sell gas bottles / propane tanks.

Top Oil is an interesting one because they also do heating fuel delivery, and some of their locations [example](https://www.top.ie/depot/?loc=17) double as a fuel depot.  I don't know what the OSM tag for this kind of place would be, but I think at least getting the fuel stations mapped is a good first step, and we can figure out the fuel depots later.

Issue #2620, opened 2019-05-17
create wikidata item for name-suggestion-index itself and add it to items used by us


matkoniecz:
Based on [Wikidata discussion](https://www.wikidata.org/w/index.php?title=Wikidata:Administrators%27_noticeboard&diff=941582891&oldid=941573611#Please_undelete_this_entries_about_brands) - it appears that it may help making wikidata entries used by us less likely to be suddenly deleted and distinguish them from spam

As first step it would be necessary to create Wikidata entry about NSI itself and mention it in the linked discussion. I plan to do this, but I am unsure when I will do that - if someone creates proper Wikidata entry, please link it here and I will push that discussion forward.

> I don't really understand what this "OSM lab name suggestion index" really is. Would it be appropriate to consider it a work by its own which deserves an item? It could then perhaps be used with stated in (P248) in references, or as main value with described by source (P1343). I am not yet sure about that, however.

> In my opinion, the "name suggestion index" should have a separate item. Can you please create one and link it from here? We could then use it in references with the [[Help:Sources#Web page|web page]] source model (carefully, e.g. only on P31 claims). —[[User:MisterSynergy|MisterSynergy]]

https://www.wikidata.org/w/index.php?title=Wikidata:Administrators%27_noticeboard&diff=941582891&oldid=941573611

Adamant36:
This would be a cool idea. Although, we would have to be sure to remove the wiki reference if the brand it's attached to is deleted from the index.

1ec5:
@ArloJamesBarnes created [Q62108705](https://www.wikidata.org/wiki/Q62108705) for NSI a while back. There’s even a bot that automatically updates it whenever we publish a new release.

SilentSpike:
I personally don't think it's a good idea to try and use NSI as a source, but can't eloquently explain why currently (it feels like bootstrapping notability). If you're able to find information on a brand/company, then why not just add the sources you used directly?

However, there are potentially other ways the NSI item could be linked to brand items via a property.

1ec5:
[...]

PR #3358, opened 2019-11-09
Add id= URL query parameter

1ec5:
Added `id` as an alternative URL query parameter to `k` and `v` that takes an entry identifier. For example, `?id=amenity/fast_food|McDonald's` not only goes to the `amenity/fast_food` category but also highlights the McDonald’s entry, just as `?k=amenity&v=fast_food#McDonald's` would. This unified query parameter would make it possible for [a proposed Wikidata external identifier property](https://www.wikidata.org/wiki/Wikidata:Property_proposal/name-suggestion-index_identifier) (#2620) to automatically link to NSI.

I considered also filtering on the name part of the identifier, which would be quite nice for discovering internationalized variants of a brand, but I couldn’t figure out how to get the name to show up in the “Tag text” field.

`nom run docbuild` also made tons of changes in docs/src.a2b27638.js due to a React version mismatch. I can commit that too if desired, but it’s probably better for someone who’s been building the app lately to take care of that before release.

### DIFF
--- a/app/src/App.js
+++ b/app/src/App.js
@@ -37,7 +37,7 @@ export default function App() {
     <Switch>
       <Route path="/" render={ routeProps => {
         const params = parseParams(routeProps.location.search);
-        if (params.k && params.v) {
+        if ((params.k && params.v) || params.id) {
           return (
             <Category {...routeProps} {...params} tree='brands' data={appData} />
           );
--- a/app/src/Category.js
+++ b/app/src/Category.js
@@ -9,11 +9,13 @@ import Filters from "./Filters";
 export default function Category(props) {
   const tree = props.tree;
   const data = props.data;
-  const k = props.k;
-  const v = props.v;
+  const id = props.id && props.id.match(/^(\w+?)\/(\w+?)\|(.+)$/);
+  const k = id ? id[1] : props.k;
+  const v = id ? id[2] : props.v;
   const kv = `${k}/${v}`;
   const entries = data.dict && data.dict[k] && data.dict[k][v];
   const hash = props.location.hash;
+  const slug = id ? id[3] : (hash && hash.slice(1)); // remove leading '#'
 
   let message;
   if (data.isLoading()) {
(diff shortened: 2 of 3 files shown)


bhousel:
This seems ok.. I'm still kind of wary of considering these NSI keys as stable, but we can try it.  You might need to include the disambiguation text too.  (I really consider the keys like `key/value|name~disambiguator`).

> `nom run docbuild` also made tons of changes in docs/src.a2b27638.js due to a React version mismatch. I can commit that too if desired, but it’s probably better for someone who’s been building the app lately to take care of that before release.

Yes this is ok to commit.. We change this pretty infrequently, and it needs to be committed because the site is hosted on a GitHub page.

app/src/Category.js
```diff
@@ -9,11 +9,13 @@ import Filters from "./Filters";
...
   const data = props.data;
-  const k = props.k;
-  const v = props.v;
+  const id = props.id && props.id.match(/^(\w+?)\/(\w+?)\|(.+)$/);
+  const k = id ? id[1] : props.k;
+  const v = id ? id[2] : props.v;
   const kv = `${k}/${v}`;
   const entries = data.dict && data.dict[k] && data.dict[k][v];
   const hash = props.location.hash;
+  const slug = id ? id[3] : (hash && hash.slice(1)); // remove leading '#'
```
1ec5: We need to prefer the hash over the slug in the ID. In fact, when the user clicks one of the “#” links, we should erase that part of the slug in the `id` parameter.

app/src/App.js
```diff
@@ -37,7 +37,7 @@ export default function App() {
     <Switch>
       <Route path="/" render={ routeProps => {
         const params = parseParams(routeProps.location.search);
-        if (params.k && params.v) {
+        if ((params.k && params.v) || params.id) {
           return (
```
anandfresh: What if params are null or undefined?
Can we add a null check for params as well, something like below
let paramObj = params ? params : {} 
if ((paramObj.k && paramObj.v) || paramObj.id) {

app/src/Category.js
```diff
@@ -9,11 +9,13 @@ import Filters from "./Filters";
 export default function Category(props) {
   const tree = props.tree;
   const data = props.data;
-  const k = props.k;
-  const v = props.v;
+  const id = props.id && props.id.match(/^(\w+?)\/(\w+?)\|(.+)$/);
+  const k = id ? id[1] : props.k;
```
anandfresh: Again check if props are not undefined

app/src/App.js
bhousel: ParseParams will not return null or undefined. It returns Array.reduce seeded with an empty Object.

PR #3369, opened 2019-11-10
Ukraine edits

skfd:
* fix some country codes
* add puzata hata


Adamant36:
Hi there and welcome to the project. If you don't mind me asking, what is a work in progress about this? I ask because usually its best to break things down into separate PRs if you can so they are easier to review and your already on 8 files being modified. So, maybe it would be best to leave things where they are currently and open a new PR after this one is merged to make more changes if need be. Thanks.

skfd:
Hi! I meant WIP as in:
* please review if it looks correct.
* should I cover more\all verification on the ground from this country in single PR.

I'm ok with accepting PR as it is. Will continue adding brands in further PRs.

Just one question, I'm concerned that `ru` and `ua` have banks with same name but different wikidata entries. Will this be handled correctly?

Adamant36:
Oh OK. Cool. Thanks for the clarification. I'll try to review and merge it as soon as I can if know one else beats me to it. 

As far as banks with the same names go, in America we can disambiguate them by state so that they only come up in searches when mapping in those specific states, but I'm not sure how that would work outside of the U.S. @bhousel would probably know though.

bhousel:
Thanks @skfd !  I appreciate all the work you're doing and it looks pretty good.. I have a few comments that I'll add as a review..

> As far as banks with the same names go, in America we can disambiguate them by state so that they only come up in searches when mapping in those specific states, but I'm not sure how that would work outside of the U.S. @bhousel would probably know though.

When the names are exactly the same, we need to add a disambiguation text, and it looks like you've done it already for some of them.
The convention we use is to put it in parenthesis like  "~(Украина)".

I'd like to come up with something a little better for the future, but for now this is what we've got.

brands/amenity/bank.json
```diff
@@ -6122,6 +6122,21 @@
       "name:ru": "Альфа-Банк"
     }
   },
+  "amenity/bank|Альфа-Банк~Украина": {
```
bhousel: We add parenthesis for these, it helps to make them more readable in a list.
```suggestion
  "amenity/bank|Альфа-Банк~(Украина)": {
```

bhousel: And you'd also have to add a "~(something)" to the other entry.

brands/amenity/bank.json
```diff
@@ -6214,6 +6229,19 @@
       "name:en": "VTB Bank"
     }
   },
+  "amenity/bank|ВТБ~Украина": {
```
bhousel: Again parenthesis here (and add something to the other ВТБ)
```suggestion
  "amenity/bank|ВТБ~(Украина)": {
```

brands/amenity/bank.json
```diff
@@ -6214,6 +6229,19 @@
...
+    "countryCodes": ["ua"],
+    "matchNames": ["втб банк"],
+    "tags": {
+      "amenity": "bank",
+      "brand": "ВТБ",
+      "brand:en": "VTB Bank",
+      "brand:wikidata": "Q16628820",
+      "brand:wikipedia": "ua:ВТБ#ПАТ_«ВТБ_Банк_(Україна)»",
```
bhousel: For this one it looks like it actually links to a different Wikipedia page.  Also make sure to replace the underscores with spaces in the interwiki link
```suggestion
      "brand:wikipedia": "ru:ВТБ Банк (Украина)",
```

brands/amenity/restaurant.json
```diff
@@ -2554,13 +2554,18 @@
...
     "tags": {
       "amenity": "restaurant",
-      "brand": "Евразия",
+      "brand": "Євразія",
+      "brand:ru": "Евразия",
+      "brand:ua": "Євразія",
       "brand:wikidata": "Q66088028",
-      "cuisine": "international",
-      "name": "Евразия"
+      "cuisine": "western,japanese",
```
bhousel: OSM uses semicolon-delimited lists for `cuisine` tag..
```suggestion
      "cuisine": "western;japanese",
```

brands/amenity/bank.json
skfd: `ua` wiki does not have a sepparate page for this bank, just a paragraph in a page of a parent bank. I think a link to subsection should be parseble?

bhousel: > I think a link to subsection should be parseble?

I think it is parsable (but replacing the underscores with spaces).  I would prefer things to have their own pages  - ideally someone would make a "ua:ВТБ Банк (Украина)".   But if you really like the section link with the "#" anchor, that's cool too.

Also, I don't really see these `brand:wikipedia` tags as all that important (or stable for that matter).  They are a convenience for humans.

As long as the more important `brand:wikidata` tag is there, people can follow that to view a wikipedia page in whatever languages are available.

bhousel:
This all looks great to me - thanks again @skfd and welcome! 👋 
Looking forward to more Ukraine improvements from you!

Issue #3368, opened 2019-11-10
Add Biscuitville


joecaruso:
https://www.wikidata.org/wiki/Q4917274
https://en.wikipedia.org/wiki/Biscuitville

I'm not sure exactly how it should be tagged.  `amenity=fast_food` and then `cuisine=southern` and/or some other way to denote breakfast?

Brent017:
Hello, Is this issue open?  I would like to help if so.

bhousel:
Yes, go for it @Brent017  👍

Brent017:
Thanks Brian, this is my first contribution.  I'm just basically adding a key value for Biscuitville to the brands file correct?
________________________________

bhousel:
> Thanks Brian, this is my first contribution. I'm just basically adding a key value for Biscuitville to the brands file correct?

Great!  We love first time contributors ❤️ 

The file to change is `brands/amenity/fast_food.json`.

If you open it you'll see a lot of entries for fast food places.  You can copy one of those to a new entry, and edit it to have the details for Biscuitville which are in this ticket.   

You'll need to set the key of the entry, and also `name`, `brand`, `brand:wikidata`, `brand:wikipedia` values, and make sure it has `"cuisine": "american;breakfast",`

If you are doing this on a local development environment with git and node, you can then run `npm run build` and it will check the syntax and make the file look pretty.  (If you are doing this in the GitHub file editor, it won't do that, but that's ok).  

Check the [README](https://github.com/osmlab/name-suggestion-index/blob/master/README.md) and [CONTRIBUTING](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) files for lots more details, and let me know if you have any questions.

rajatmw1999:
I have added the Biscucitville brand to brands/amenity/fast_food.json. As per the information available on the and from Biscuitville's offical website, the company is primarily involved in selling biscuits made from their own recipe. It was involved in the pizza and bread business but biscuits is what they are known for. So, "cuisine" is given the value of "biscuit". Other better options can be "cuisine" : "snacks". I am a first time contributor, so, if there's anything that i missed, please inform about it!

https://github.com/osmlab/name-suggestion-index/pull/3373

Brent017:
I have also submitted a pull request for this issue.  It looks like the contributor above forgot to change the name at top from A&W to Biscuitville.

bhousel:
done in #3376

PR #3373, opened 2019-11-14
Added Biscuitville as mentioned in issue #3368

rajatmw1999:
I have added the Biscucitville brand to brands/amenity/fast_food.json. As per the information available on the and from Biscuitville's offical website, the company is primarily involved in selling biscuits made from their own recipe. It was involved in the pizza and bread business but biscuits is what they are known for. So, "cuisine" is given the value of "biscuit". Other better options can be "cuisine" : "snacks".


bhousel:
Thanks! However I ended up accepting #3376 because @brent017 did ask first...
(Also this PR still has "amenity/fast_food|A&W~(USA)" as the identifier)

Issue #3378, opened 2019-11-14
Nextbike: should be replaced by sub-brands


sb12:
At least for Germany, nextbike has a lot of sub-brands e.g. KVV.nextbike, VRNnextbike, metropolradruhr, I don't know any city in Germany where only `nextbike` is used as the brand. It is very important to differ between this brands as there can be differences in pricing especially regarding annual rates. Because of the increasing power of this index and the iD editor many of these subbrands have disappeared from the database.
`name=nextbike` also does not make any sense as the brand name is used on the rental stations, sometimes combined with the name of the specific location.
Therefore please remove nextbike from this index for Germany and add the subbrands instead if needed.

Adamant36:
" I don't know any city in Germany where only nextbike is used as the brand. Therefore please remove nextbike from this index for Germany and add the subbrands instead if needed."

Doing a basic Google search comes up with a bunch of results. In particular there's [this](https://www.nextbike.de/en/giessen/) interesting one where the company itself seems to use OpenStreetMaps as a way for people to find NextBike locations. There's also images on that page things bikes etc labeled just as "NextBike" and the page is specific to Germany. So, they must use it alone as branding there. 

Anyway, removing NextBike from the database in Germany might impact if they use us a map source or not. Since a lot of those places would be re-tagged and they clearly think the objects currently being tagged that way are correct as is or they wouldn't be showing them as part of just the NextBike brand if their wasn't such a thing. 

Also, there's a bunch of other website sites and images out there  with bikes and stations in Germany just showing NextBike as the branding, including a few buildings. Maybe adding the Sub brands would be OK, but it clearly shouldn't come at the cost of removing the main brand from the index. 

[NextBike sign in Berlin](https://www.gettyimages.com/detail/news-photo/rent-by-app-nextbike-is-written-on-a-street-sign-in-berlin-news-photo/1039485728) (I couldn't just upload it)

NextBike branding from images in random articles about Nextbike in Germany. Including from their own website. 

![NextBike 1](https://user-images.githubusercontent.com/30259065/68902607-a4caee80-06ed-11ea-8a14-f9bd0669a007.jpg)

![Nextbike 2](https://user-images.githubusercontent.com/30259065/68902615-a98fa280-06ed-11ea-93f5-a572a10cf361.jpg)

![nextbike_1](https://user-images.githubusercontent.com/30259065/68902732-f2475b80-06ed-11ea-8afa-015dc2b1b34a.jpg)

[...]

PR #3381, opened 2019-11-15
Add sub-brands in DE, AT and CH for nextbike

sb12:
See #3378

### DIFF
--- a/brands/amenity/bicycle_rental.json
+++ b/brands/amenity/bicycle_rental.json
@@ -44,7 +44,7 @@
       "name": "Swapfiets"
     }
   },
-  "amenity/bicycle_rental|nextbike": { // FIXME: 
+  "amenity/bicycle_rental|nextbike": {
     "countryCodes": ["at","ch","de"],
     "tags": {
       "amenity": "bicycle_rental",
--- a/brands/amenity/bicycle_rental.json
+++ b/brands/amenity/bicycle_rental.json
@@ -44,13 +44,143 @@
       "name": "Swapfiets"
     }
   },
-  "amenity/bicycle_rental|metropolradruhr": {
+  "amenity/bicycle_rental|nextbike": { // FIXME: 
+    "countryCodes": ["at","ch","de"],
+    "tags": {
+      "amenity": "bicycle_rental",
+      "brand": "nextbike",
+      "brand:wikidata": "Q2351279",
+      "brand:wikipedia": "de:Nextbike",
+      "operator": "nextbike GmbH"
+    }
+  },
+  "amenity/bicycle_rental|nextbike - Deezer nextbike (Berlin)": {
+    "countryCodes": ["de"],
+    "tags": {
+      "amenity": "bicycle_rental",
+      "brand": "Deezer nextbike",
+      "brand:wikidata": "Q67874410",
+      "network": "Deezer nextbike",
+      "operator": "nextbike GmbH"
+    }
+  },
+  "amenity/bicycle_rental|nextbike - Frelo": {
+    "countryCodes": ["de"],
+    "tags": {
+      "amenity": "bicycle_rental",
+      "brand": "Frelo",
+      "network": "Frelo",
+      "operator": "nextbike GmbH"
+    }
+  },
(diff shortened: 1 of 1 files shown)


bhousel:
We do need `name` tags for all the things in this index, even for situations where convention is for the user to modify the name tag.

sb12:
Do we really need a `name` tag? According to [the wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_rental) `name=*` should be used for the name of the station (e.g. 'Main Station`) and `network=*` for the name of the station network (`nextbike` in this case). The station name is usually printed in small letters at the station and refers to the locality. 
Adding the name tag violates the approved description in the wiki and might lead to the deletion of correct tags, when the suggestion is used careless by mappers.

bhousel:
This particular index currently needs one

bhousel:
At some point we will probably support other tags like `network`, but we're not currently there yet, sorry..

Adamant36:
@sb12, It's not like people can't just edit the name field if need be. It's currently that way with other things in the index that have local names along with the more general name, like post offices or hotels. Given that would you be willing to just add the name tags so this can be merge? Because it's better to have the presets then not. Especially with the Nextbike preset being removed in #3378. Which, IMO and how you made it sound there, was semi-contingent on this PR happening.

sb12:
> It's not like people can't just edit the name field if need be.

In theory yes, in praxis `name` will be overwritten in iD-Editor and using the JOSM preset, also name will be locked in the iD-Editor after adding the brand. So mappers, which do not read the wiki, will think, `name` is used for the brand name and not for the station name, but this is a problem of all `bicycle_rental` brands and can be solved as soon as it is possible to add brands without names.

I might add the `name` tag in order to add the brands again, but I am not very happy about it.

@bhousel: Travis also complains about `operator`, do I need to remove this tag?

bhousel:
> I might add the `name` tag in order to add the brands again, but I am not very happy about it.
> @bhousel: Travis also complains about `operator`, do I need to remove this tag?

I think if you're not happy about it, we should just leave it out of the index for now.

For now the index is really designed around brands with consistently named features.  It is after all called the "**name** suggestion index", and we are seeding it with the data that we find in the `name` tag.  

We know that we do have a handful of things in the index with name variability (for example, hotels and car dealerships) but in those cases it's not the worst thing if a mapper adds a dealership with the name "Honda", and then forgets to rename it to "Planet Honda Route 22". 

* We don't currently support features that are keyed off `operator`/`operator:wikidata` or `network`/`network:wikidata` but there are some tickets where we've said that this would be good to do.  
* We also don't have a great solution for the name variability, but we've discussed maybe adding a property to some kinds of things to say whether the name should be strictly applied or not.

Until people are happy with how iD and NSI are working together, we don't have to try to stick everything into the NSI that doesn't really fit.


Here are some related tickets
https://github.com/openstreetmap/iD/issues/6890
https://github.com/openstreetmap/iD/issues/6561  (and dup https://github.com/openstreetmap/iD/issues/6951)

Adamant36:
> I think if you're not happy about it, we should just leave it out of the index for now.

Well, I'm disappointed. @sb12 could you at least discuss this on the German tagging mailing list before we completely leave it like out (like you should have done before requesting NextBike's removal)?

Adamant36:
@ENT8R, your from Germany right? Do you have any opinions on this?

ENT8R:
Actually I fully agree with @bhousel. Currently this type of data does not really fit to the goals of the index, maybe in the future there will be another chance to add this. Maybe we need an **network-suggestion-index** beside this index... :smile:
So I would propose to add just one entry for nextbike and rather keep it more simple for the moment.

Adamant36:
> So I would propose to add just one entry for nextbike and rather keep it more simple for the moment.

That's how it originally was, but unfortunately it got removed because of #3378. I guess it can be re-added though. That's what should happen IMO.

Adamant36:
I think I'm going to close this since the consensus seems to be that it doesn't fit with the index. Maybe someone can add the Nextbike entry back at some point if that's what the community wants so there's at least something related to in iD. If not though, that's fine also.

PR #3387, opened 2019-11-17
Update outdoor with Go Outdoors

SK53:
Add Go Outdoors, UK chain with 70 stores (owned by JD Sports) https://www.gooutdoors.co.uk/google/store-locator

### DIFF
--- a/brands/shop/outdoor.json
+++ b/brands/shop/outdoor.json
@@ -50,6 +50,15 @@
       "shop": "outdoor"
     }
   },
+  "shop/outdoor|Go Outdoors": {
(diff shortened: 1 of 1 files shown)


SilentSpike:
Thanks, have been meaning to add this one for a while!

SK53:
maxerickson mentioned inline editing on GitHub the other day, & trigpoint said it wasn't there, so thought it was a good time to try out the editing: definitely easier than the desktop for this kind of thing. Probably need to add a url to the wikidata entry to stop it getting deleted

SK53:
Just seen that you've done all that. Excellent.

Issue #3371, opened 2019-11-12
Mace needs splitting up.


doublah:
Mace is 3 brands, current wikidata info only refer to chain in GB and wikipedia page refers to all of them.

Mace (Ireland)
https://mace.ie/ - https://www.facebook.com/MACEIreland
ie
Site claims 160 stores.

Mace (Northern Ireland)
http://www.maceni.co.uk/ - https://www.facebook.com/MACENorthernIreland/
gb (no used country code for NI)
Facebook claims 88 stores.

Mace (Great Britain)
http://www.macestores.co.uk/ - https://www.facebook.com/maces.stores
gb
Site claims 764 stores.

Adamant36:
Mace (Ireland) [Q75236910](https://www.wikidata.org/wiki/Q75236910)
Mace (Northern Ireland) [Q75238536](https://www.wikidata.org/wiki/Q75238536)
Mace (Great Britain) [Q75240338](https://www.wikidata.org/wiki/Q75240338) (I guess there was already a Wikidata entry for this one and I accidentally created a duplicate for it. Oh well 🤷‍♂) 

Now someone just needs to create the entries for them.

doublah:
Fixed in #3371

PR #3389, opened 2019-11-19
Add entries for Mace IE, GB, NE

lyndic:
For issue #3371


bhousel:
This looks ok.. I was confused why Q17089386 code goes away but I can see from reading https://en.wikipedia.org/wiki/Mace_(shop) that they really are operating as 3 separate entities with 3 separate brand identities.

PR #3394, opened 2019-11-21
add new brand Blue Bottle Coffee

yukijina:
Hi!

Add a new brand blue bottle coffee. Blue Bottle is a cafe but also coffee roaster and retailer. Should I add matchTags to brands/shop/coffee as well?  Thanks for your advice in advance!

### DIFF
--- a/brands/amenity/cafe.json
+++ b/brands/amenity/cafe.json
@@ -103,6 +103,17 @@
       "takeaway": "yes"
     }
   },
+  "amenity/cafe|Blue Bottle Coffee": {
+    "countryCodes": ["jp", "kr", "us"],
+    "tags": {
+      "amenity": "cafe",
+      "brand": "Blue Bottle Coffee",
+      "brand:wikidata": "Q4928917",
+      "cuisine": "coffee_shop",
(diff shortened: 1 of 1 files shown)


bhousel:
> Add a new brand blue bottle coffee. Blue Bottle is a cafe but also coffee roaster and retailer. Should I add matchTags to brands/shop/coffee as well? Thanks for your advice in advance!

Thanks @yukijina this looks really good 👍 
You don't need to add a `matchTags` for the `amenity/cafe` and `shop/coffee` because this match happens automatically. (Similar tags are automatically matched according to the [`match_groups.json` file](https://github.com/osmlab/name-suggestion-index/blob/master/config/match_groups.json))

The only other thing I would add is a tag `"brand:wikipedia": "en:Blue Bottle Coffee"` to link to the wikipedia tag.

yukijina:
Thanks, @bhousel ! I added `"brand:wikipedia": "en:Blue Bottle Coffee"` now.

bhousel:
> Thanks, @bhousel ! I added `"brand:wikipedia": "en:Blue Bottle Coffee"` now.

Awesome!  That was fast haha  🚀

PR #3395, opened 2019-11-21
Chicken Express (Q5096235) dont operate in UK

SK53:
The entry for Chicken Express is allowing iD users to make incorrect edits attaching brand tags for a regional US entity "Chicken Express" to small independent operated UK fast food outlets with a similar name. Available information suggests that the wikidata entity only operates in the US. Further evidence needs to be provided for any UK based existence. Specific example: [https://www.openstreetmap.org/user_blocks/3277](https://www.openstreetmap.org/user_blocks/3277).

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -441,7 +441,7 @@
     }
   },
   "amenity/fast_food|Chicken Express": {
-    "countryCodes": ["gb", "us"],
+    "countryCodes": ["us"],
     "tags": {
       "amenity": "fast_food",
       "brand": "Chicken Express",


Adamant36:
Hi @SK53. I fixed it in https://github.com/osmlab/name-suggestion-index/commit/f9b01ba6c053a49cc1f7dc5adc11f29aa74fd9fd (look at this earlier and then thought it was an issue later on when I wanted something to do. My bad. At least it's dealt with). Lots of entries could use improvement when it comes to country code. So feel free to do more work in that area so we don't have these problems as much. I plan to go through and add or update a lot of them myself at some point. Its unfortunate it led to miss-tagging in this case, but it's probably the exception.

SK53:
Not entirely this is the third such issue I've encountered this week. I expect a few teething troubles, and it's relatively easy to propose fixes, so not a huge deal. 

Some of the issues are that trigger happy inexperienced users are editing all over the place, but I've also seen experienced users caught out.

PR #3398, opened 2019-11-25
Update KIKO Milano and ONCE, remove generic names, add Iberdrola and Endesa

ArenaL5:
Commits should be self-explanatory if I did this correctly.


bhousel:
Looks good to me, thanks @ArenaL5  👍   I'll merge as-is..

The only thing that stood out for me is, should the energy supplier offices have country codes?  If so, you can add them in a later commit..

ArenaL5:
Doesn't really seem to be worth it. I did it for Naturgy in a prior PR, but the ones I added are multinational spread over several countries.

Thank you!

PR #3399, opened 2019-11-25
Add nomacth tags to supress warnings for gortz bakery and gortz shoes

micaelalechmann:
Add nomatch tags to supress duplicate warnings, since there are two places with the same name on different categories. This is my first pull request, hope it goes well!


bhousel:
Looks perfect, thanks @micaelalechmann and welcome! 👋

PR #3401, opened 2019-11-25
brands/amenity: Added Old Country Buffet

swapnamanupati:
This adds Old country buffet restaurant with key-value pairs in restaurant.json file, ran the builds again.

### DIFF
--- a/brands/amenity/restaurant.json
+++ b/brands/amenity/restaurant.json
@@ -1638,6 +1638,16 @@
       "official_name": "Old Chicago Pizza & Taproom"
     }
   },
+  "amenity/restaurant|Old Country Buffet": {
+    "countryCodes": ["us"],
+    "tags": {
+      "amenity": "restaurant",
+      "brand": "Old Country Buffet",
+      "brand:wikidata": "Q75243208",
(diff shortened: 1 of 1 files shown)


brands/amenity/restaurant.json
```diff
@@ -1638,6 +1638,16 @@
       "official_name": "Old Chicago Pizza & Taproom"
     }
   },
+  "amenity/restaurant|Old Country Buffet": {
+    "countryCodes": ["us"],
+    "tags": {
+      "amenity": "restaurant",
+      "brand": "Old Country Buffet",
+      "brand:wikidata": "Q75243208",
+      "cuisine": "salads;veggies;desserts",
```
bhousel: We try to use "standard" cuisine values that already exist in OpenStreetMap. You can view and search for these on [taginfo](https://taginfo.openstreetmap.org/keys/cuisine#values).

I'm going to change this one to "american", which is kind of a catchall value that doesn't mean any specific kind of food. 
```suggestion
      "cuisine": "american",
```

bhousel:
Thanks @swapnamanupati  and welcome!  👋

PR #3402, opened 2019-11-26
Change 'has part (P527)' to 'has parts of the class (P2670)'

jleedev:
Creating a claim 'has part: shop' will trigger a constraint violation here: https://www.wikidata.org/wiki/Wikidata:Database_reports/Complex_constraint_violations/P527#relations_between_instances_and_classes_must_use_P2670

Update the gif to show 'has parts of the class' instead.


bhousel:
Thanks @jleedev  👍

PR #3411, opened 2019-12-01
add Back-Factory and Pneuhage

doktorpixel14:
missing German brands


bhousel:
Thanks, looks great @doktorpixel14 👍

PR #3412, opened 2019-12-02
remove Giant Eagle fuel; it's a duplicate of GetGo

jleedev:
Preserve the matchNames.

### DIFF
--- a/brands/amenity/fuel.json
+++ b/brands/amenity/fuel.json
@@ -1016,6 +1016,7 @@
   },
   "amenity/fuel|GetGo": {
     "countryCodes": ["us"],
+    "matchNames": ["giant eagle fuel"],
     "tags": {
       "amenity": "fuel",
       "brand": "GetGo",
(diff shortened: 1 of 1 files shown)


bhousel:
Does this mean that all "Giant Eagle Fuel" should be upgraded to "GetGo"?   (i.e. they've all been rebranded?)

jleedev:
Short answer: Yes, although there aren't actually any in OSM.

`nwr["amenity"="fuel"]["brand:wikidata"="Q1522721"]` turns up nothing. `nwr["amenity="fuel"][name="Giant Eagle"]` turns up two which could be upgraded, but one may have since closed or moved.

This preset appears to never have been actually used — iD won't bring it up for some reason.

Longer answer: Wikipedia suggests that the name "Giant Eagle Fuel" was only used from 2001–2003, so I would presume that these (two) examples in OSM data are not from anyone claiming that that is the name, merely an informal title as "the gas station of the grocery store".

("GetGo from Giant Eagle" might be an official_name to consider, but I'm not inclined to do so.)

bhousel:
> This preset appears to never have been actually used — iD won't bring it up for some reason.

It was only added a month ago, so probably we haven't released an iD that includes it yet - cffdf7ce9e

bhousel:
Nice!  I checked one out and it looks like even by 2015 it was already changed over to GetGo..

Screenshot 2019-12-02 15 15 12

PR #3415, opened 2019-12-03
Move Upper Crust (UK) from bakery to fast food

SK53:
Upper Crust is a chain of sandwich bars largely located in transport hubs (railway stations and airports). They sell crusty baquette style sandwiches, coffee and pastries, but are distinctly not a bakery. Mainly UK based they have outlets in Fennoscandinavia and airport outlets elsewhere (including, according to Wikipedia, one in the US). Many currently mapped outlets in Norway & Sweden are mapped as cafes and as I am not familiar with them I have not added this as a matching value. In addition the Upper Crust Pizza chain has a number of already mapped outlets in the US, so for now, because I suspect it might cause more erroneous tagging than it may avoided, I've not added US as a country code.

Although tag usage is split pretty evenly three ways: café, fast_food or bakery. Bakery is clearly inappropriate: they remain open very late, bakeries are not typically located on stations (at least in UK), they don't sell full range of bakery products and not described as such on Wikipedia. Many have seating, but most are just an enclosure on the station concourse, but café may be appropriate in other cases. Fast food will not be wrong in all cases.

Proximate cause for this PR is a inexperienced editor [changing](https://www.openstreetmap.org/changeset/77841609) an element with the fast food tags.

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -2904,6 +2904,21 @@
       "takeaway": "yes"
     }
   },
+    "amenity/fast_food|Upper Crust": {
+     "countryCodes": ["de","dk","eg","es","fi","ie","gb","no","se"],
+      "matchTags": [ 
+      "shop/bakery"
+    ],
+    "tags": {
+      "brand": "Upper Crust",
+      "brand:wikidata": "Q7898585",
+      "brand:wikipedia": "en:Upper Crust (restaurant chain)",
+      "name": "Upper Crust",
+      "amenity": "fast_food",
+      "cuisine": "sandwich",
+      "takeaway": "yes"
+    }
+  },
   "amenity/fast_food|Valentine": {
     "countryCodes": ["ca"],
     "tags": {
--- a/brands/shop/bakery.json
+++ b/brands/shop/bakery.json
@@ -687,15 +687,7 @@
       "shop": "bakery"
     }
   },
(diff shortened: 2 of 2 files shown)


brands/amenity/fast_food.json
```diff
@@ -2903,6 +2903,21 @@
       "name": "Tropical Smoothie Cafe",
       "takeaway": "yes"
     }
+  },
+    "amenity/fast_food|Upper Crust": {
+     "countryCodes": ["de","dk","eg","es","fi","ie","gb","no","se"],
+      "matchTags": [
```
bhousel: Thanks!  You can actually remove the `matchTags` property here. 

We have a file [`match_groups.json`](https://github.com/osmlab/name-suggestion-index/blob/master/config/match_groups.json) which groups together similar tags, so that "amenity/fast_food" will automatically match other tags like "bakery" "fast_food" "cafe" "ice_cream" etc.

bhousel: I removed the matchtag in b27384a31

SK53:
I don't know if it is possible, but I'd prefer an explicit match. 

Some Upper Crust outlets are enclosed with decent seating and have been classified as cafes (particularly in Fennoscandia), whereas others are just kiosks on station concourses and anything but cafes. So I'd prefer that only bakery is over-written (of course I still stand by fast_food is never wrong too).

PR #3416, opened 2019-12-03
Frankie & Bennie's cuisine change

SK53:
Frankie & Bennie's offering is distinctly American in style, and definitely not the same as other restaurants with cuisine=Italian. As Italian-American is not represented on taginfo, I've changed this from 'italian' to 'american'. Current tag values are deceptive as they represent a single individual updating multiple restaurants and not a consensus of UK mappers (see values at start of year: https://overpass-turbo.eu/s/OFK)


bhousel:
Thanks! 👍

PR #3421, opened 2019-12-06
add 3 German brands

doktorpixel14:
New brands:
* BGV
* CAP-Markt
* Württembergische

I also corrected the Edeka Wikipedia link to de: because it only operates in Germany


bhousel:
thanks @doktorpixel14  👍

PR #3427, opened 2019-12-10
Added brand entries for Waikiki Poke and Blackjack Pizza

smfuller:


bhousel:
Thanks @smfuller and welcome! 👋 
This looks good!  "Waikiki Poke" will need a `brand:wikidata` entry at some point, but we can merge as-is..

Adamant36:
@smfuller, you know how many stores Waikiki Poke has and where they are mainly located? It seems from an internet search that they only have one location in Colorado.

smfuller:
@Adamant36 Although I thought this was a franchise, after looking at this further it appears that this is indeed the only location right now.

Should I submit another pull request from my fork without this location info?

bhousel:
no problem @smfuller  - I removed it in ecda8ff2d

PR #3429, opened 2019-12-12
Updated Clarion Wikidata 

sophiabonsu:
Hi @Adamant36  I made a new pull request with the  wikidata update , please let me know if that it's acceptable.


bhousel:
Thanks @sophiabonsu - I merged this, but without the `brand:wikipedia` pointing to the "Choice Hotels" parent company.

Adamant36:
Problem now is that they don't seem to have a unique brand presence on social media. The ones in the Wikidata entry are for a specific hotel in PH. So should we just leave them out or with the social media links for Choice Hotels?

bhousel:
> Problem now is that they don't seem to have a unique brand presence on social media. The ones in the Wikidata entry are for a specific hotel in PH. So should we just leave them out or with the social media links for Choice Hotels?

If the brand doesn't have an official "brand" facebook, sometime I will just pick something that seems close enough.  

Looking through https://nsi.guide/index.html?k=tourism&v=hotel - it does look weird that all of the sub-brands link to the same "Choice Hotels" page.  We should create a separate wikidata for each operating brand.

PR #3435, opened 2019-12-20
Add Tapi Carpets

trigpoint:
Tapi Carpets is large UK carpet chain

### DIFF
--- a/brands/shop/carpet.json
+++ b/brands/shop/carpet.json
@@ -11,3 +11,15 @@
     }
   }
 }
+
+{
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @trigpoint & welcome! 👋 
There's something going on here with a missing comma, but I'll just merge and fix it.

trigpoint:
Thank you @bhousel 
Just one thing, it should be gb only.

bhousel:
> Just one thing, it should be gb only.

ok, just changed it

PR #3448, opened 2019-12-27
update of Gyomu Super name

javbw:
removal of romaji name, add proper english name found on their website.


bhousel:
Thanks @javbw and welcome! 👋

javbw:
No Prob! Glad to see it is quick and simple - I will be doing more as I find them here in Japan.

PR #3470, opened 2020-01-06
Added Cactus supermarket in Luxembourg / Wikidata for Big-A in Japan

Contrib1043:
(for info see https://en.wikipedia.org/wiki/Cactus_(supermarket) and https://www.cactus.lu/fr/magasins-horaires#&supermarche+marche+hobbi+shoppi&), for Big-A https://ja.wikipedia.org/wiki/%E3%83%93%E3%83%83%E3%82%B0%E3%83%BB%E3%82%A8%E3%83%BC

### DIFF
--- a/brands/shop/supermarket.json
+++ b/brands/shop/supermarket.json
@@ -503,6 +503,16 @@
       "shop": "supermarket"
     }
   },
+  "shop/supermarket|Cactus": {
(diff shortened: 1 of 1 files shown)


bhousel:
Looks good thanks @Contrib1043 👍

Issue #3406, opened 2019-11-26
Add hhgreg to the dissolved list


tas50:
These are still tagged all over:

https://en.wikipedia.org/wiki/H._H._Gregg

hikemaniac:
I've added an attribute to [Wikidata](https://www.wikidata.org/wiki/Q5750102), indicating that it was dissolved. 👍 If I understand the process correctly, _H. H. Gregg_ now needs to be added to the index and running `npm run dist` should add it to dissolved.json.

hikemaniac:
@tas50 I've been digging a bit deeper on this one now and it seems that all former stores have been closed but another company bought the intellectual property of hhgregg and has opened a [new store in New Jersey](http://franklinreporter.com/h-h-gregg-opens-new-retail-electronics-store-in-township/). It is also mentioned on their [homepage](https://www.hhgregg.com/hhgregg-grand-opening/). What should be done in this case?

bhousel:
Closing - it's not in the index anymore..

Issue #3154, opened 2019-10-24
Figure out process of remapping an old QID to a new one


bhousel:
There are a handful of brands in `dissolved.json` where we believe that they should have a new Wikidata QID.  

For example, from discussion on OSM-US Slack, we think that:

* [Gulf Oil Q1296860](https://www.wikidata.org/wiki/Q1296860) maybe should be remapped to [Gulf Oil LP Q5617505](https://www.wikidata.org/wiki/Q5617505)
In this case Gulf company dissolved and their IP is now owned by a successor company that is using it. 

* [Conoco Q1126518](https://www.wikidata.org/wiki/Q1126518) maybe should be remapped to [ConocoPhillips Q214763](https://www.wikidata.org/wiki/Q214763)
This case is more complicated.  Conoco is no longer an active company, but [it is still an active brand](https://www.conoco.com/) being used by ConocoPhillips.  I'm not sure what to make of it - maybe it doesn't belong in `dissolved.json` ?

When we decide that a brand should switch QIDs, we should have a way of just remapping them all en masse and maybe recording this in the brand files so that other projects can know.

1ec5:
> Conoco is no longer an active company, but it is still an active brand being used by ConocoPhillips. I'm not sure what to make of it - maybe it doesn't belong in dissolved.json ?

dissolved.json should track brands just like our other files. A brand can change hands without going defunct, and that seems to have happened with the Conoco brand: ConocoPhillips spun off Conoco as part of [Phillips Petroleum Company](https://en.wikipedia.org/wiki/Phillips_Petroleum_Company).

Over time, we should ensure that gas station brands are linked to Wikidata items that are about the brands rather than the parent companies. Those brand items will be more stable, since gas station brands change hands quite frequently.

SilentSpike:
I was going to say the way brands are modelled at wikidata could actually be improved - separating the brand entity from the company entity. 

For most businesses it hasn't really been needed (e.g. McDonald's the brand and McDonald's the company are so intertwined that using the company item is stable enough for our needs), but for brands that are changing hands it could be useful.

Would probably want to discuss how to approach creating brand entities at https://www.wikidata.org/wiki/Wikidata:WikiProject_Companies or how brands can be best modelled at https://www.wikidata.org/wiki/Wikidata:WikiProject_Ontology

wireguy:
Chase Bank is another example.
[...]

PR #3472, opened 2020-01-07
Multi .json file edit

UKChris-osm:
I thought with my second attempt I would try a pull with multiple **.json** file edits this time.

- /brands/office/**estate_agent.json** - added 6 British Estate Agent Brands.
- /brands/leisure/**bowling_alley.json** - added British Hollywood Bowl Brand (but not AMF)
- /brands/amenity/**cafe.json** - Added Boston Tea Party and Coffee#1
- /brands/shop/**garden_centre.json** - Added 2 Brands, and removed Wyevale which no longer trades.

I've used a matchNames for ght in the **estate_agent.json** file, which might not be required, and am not 100% sure if using a # for Coffee#1 within **cafe.json** is correct, but have also used a matchNames for alternatives to using the #

AMF is a brand also used by Hollywood Bowl, but only at six locations, the rest being branded "Hollywood Bowl", so I'm not sure having one for AMF was worth it.

### DIFF
--- a/brands/amenity/cafe.json
+++ b/brands/amenity/cafe.json
@@ -135,6 +135,18 @@
       "takeaway": "yes"
     }
   },
+    "amenity/cafe|Boston Tea Party": {
+    "countryCodes": ["gb"],
+    "tags": {
+      "amenity": "cafe",
+      "brand": "Boston Tea Party",
+      "brand:wikidata": "Q23461358",
+      "brand:wikipedia": "en:Boston Tea Party (café chain)",
+      "cuisine": "coffee_shop",
+      "name": "Boston Tea Party",
+      "takeaway": "yes"
+    }
+  },
   "amenity/cafe|Bridgehead": {
     "countryCodes": ["ca"],
     "tags": {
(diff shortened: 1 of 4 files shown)


bhousel:
This looks mostly good!

The only thing I would change is leaving Wyevale in for now.  When a brand shuts down, we leave it in the index for some time but we mark it as dissolved on Wikidata (with claim P576 - Dissolution date).  This adds the brand to a special `dist/dissolved.json` file which facilitates cleanup in OSM.

(More info on #3154 and #3406 - this is not perfect, but we are still thinking of how best to handle dissolved brands).

bhousel:
> When a brand shuts down, we leave it in the index for some time but we mark it as dissolved on Wikidata (with claim P576 - Dissolution date).

I did this ☝️

bhousel:
> If I've reading the above commit's correctly, you've merged my whole pull in the first instance, which included the deletion of Wyevale, then re-added Wyevale back as it was as a secondary edit?

Yes!  I opened the garden_centre file in my editor, then copied the Wyevale lines, then ran `git pull origin` which overwrote the file with your changes, then I just pasted the Wyevale lines back in and re-committed the file and pushed the update back to GitHub.

PR #3478, opened 2020-01-10
Added CGR cinemas brand

QuentinS33:
Added CGR cinemas which operates mainly in France

### DIFF
--- a/brands/amenity/cinema.json
+++ b/brands/amenity/cinema.json
@@ -80,6 +80,17 @@
       "short_name": "Century"
     }
   },
+  "amenity/cinema|CGR": {
+    "countryCodes": ["fr"],
+    "matchNames": ["cgr"],
+    "tags": {
+      "amenity": "cinema",
+      "brand": "CGR",
+      "brand:wikidata": "Q2931406",
+      "brand:wikipedia": "fr:Circuit_Georges-Raymond",
+      "name": "CGR"
+    }
+  },
   "amenity/cinema|CineStar": {
     "countryCodes": ["de"],
     "tags": {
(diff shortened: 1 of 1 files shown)


brands/amenity/cinema.json
```diff
@@ -80,6 +80,17 @@
       "short_name": "Century"
     }
   },
+  "amenity/cinema|CGR": {
+    "countryCodes": ["fr"],
+    "matchNames": ["cgr"],
```
bhousel: This line can go away, matching is case-insensitive.
```suggestion
```

brands/amenity/cinema.json
```diff
@@ -80,6 +80,17 @@
...
+    "tags": {
+      "amenity": "cinema",
+      "brand": "CGR",
+      "brand:wikidata": "Q2931406",
+      "brand:wikipedia": "fr:Circuit_Georges-Raymond",
```
bhousel: Wiki tags need to have underscore `_` replaced with space ` `..
```suggestion
      "brand:wikipedia": "fr:Circuit Georges-Raymond",
```

brands/amenity/cinema.json
QuentinS33: Ok, I didn't know the case-insensitive for matching brands. I will correct this.

brands/amenity/cinema.json
QuentinS33: Ok, i didn't know that rule. I copied the brand wikipedia from URL. I will correct this.

bhousel:
Thanks @QuentinS33 looks great!  and welcome 👋

PR #3481, opened 2020-01-11
Add Malitel and Orange Mali

mds08011:
Adds two `shop=telecommunication` brands in Mali (West Africa).  Orange Mali is a subsidiary of the French Orange.  Both brands operate stores which sell telecommunication devices beyond just mobile phones, such as modems, wireless routers, USB receivers, etc.


bhousel:
looks good, thanks @mds08011 👍

Issue #3482, opened 2020-01-11
tReds - shoe shop, but some branches also sell clothing.


UKChris-osm:
There is a brand called [tReds](http://treds.co.uk) in the South of England, which is mostly a shoe shop, but some stores also sell clothes as a secondary line. How would this best be included in the NSI, if at all?

As a shoe shop and left at that (with a matchtag for clothes maybe), or two separate entries, one for shoes and one for clothes, or as one entry just for clothes, or something else?

My initial code is:

  "shop/shoes|tReds": {
    "countryCodes": ["gb"],
    "matchTags": ["shop/clothes"],
    "tags": {
      "brand": "tReds",
      "brand:wikidata": "Q81647338",
      "name": "tReds",
      "shop": "shoes"
    }
  }

bhousel:
Yeah the sportswear brands are all over the place.

- Adidas is `shop=sports`
https://nsi.guide/index.html?k=shop&v=sports#Adidas

- Nike is `shop=clothes`
https://nsi.guide/index.html?k=shop&v=clothes#Nike

- New Balance is `shop=shoes`
https://nsi.guide/index.html?k=shop&v=shoes#New%20Balance

We picked these categories based on how popular the tags were that people were using for the brands.  And I could see any of these being any of the tags - the business is mostly shoes, but you can also buy sportswear and maybe even some equipment or balls.

My suggestion is to not over-think it.  If it's _mostly_ a shoe store, just tag it as `shop=shoes` and it will have a shoe icon on the map.  

I don't think it's good to have multiple entries unless there is a really good reason.

UKChris-osm:
I'd say it's mostly a shoe shop, so that's what I'll submit. You're right I'm probably trying to hard to have my entries exactly right, after all I'd feel very bad if I got anything wrong an interfered with an already existing and excellent project :)

Did the NSI start from automatic scans of the current OSM database? And whatever was the most used at the time is how each entry started out?

UKChris-osm:
#3484 contained tReds in a Pull Request.

PR #3484, opened 2020-01-11
Add tReds and BrightHouse

UKChris-osm:
Added "_tReds_" (with a matchTags for clothes) to **shoes.json**
Added "_BrightHouse_" (with a matchTags for department store) to **variety.json**

### DIFF
--- a/brands/shop/variety_store.json
+++ b/brands/shop/variety_store.json
@@ -58,6 +58,17 @@
       "shop": "variety_store"
     }
   },
+  "shop/variety_store|BrightHouse": {
+    "countryCodes": ["gb"],
+    "matchTags": ["shop/department_store"],
+    "tags": {
+      "brand": "BrightHouse",
+      "brand:wikidata": "Q4967357",
+      "brand:wikipedia": "en:BrightHouse (retailer)",
+      "name": "BrightHouse",
+      "shop": "variety_store"
+    }
+  },
   "shop/variety_store|D9150": {
     "tags": {
       "brand": "D9150",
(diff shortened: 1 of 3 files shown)


brands/shop/variety_store.json
```diff
@@ -58,6 +58,17 @@
       "shop": "variety_store"
     }
   },
```
tas50: Bright house really shouldn’t be a variety store since it’s a rent to own shop. OSM unfortunately doesn’t have an appropriate tag for that which is probably why it’s tagged differently throughout the UK. In the US we’ve used electronics and furniture tags for similar stores but those are also incorrect.

UKChris-osm: I did go back and forth with how to tag BrightHouse, but came to wonder would the method of payment have a bearing on the OSM tagging, as opposed to the goods being sold by the shop?

Isn't _Rent to Own_ simply another way of paying on credit / account, similar to how I could go into the John Lewis department store, buy a new bed, and pay it off in 12 month instalments, either via a credit card transaction, or a credit agreement with John Lewis??

bhousel: I'd probably just call it a `shop=department_store` if they sell furniture, electronics, appliances.  `variety_store` is mostly used for dollar stores like Poundland/Poundworld, etc.

bhousel:
closes #3482

brands/shop/variety_store.json
UKChris-osm: As recommended, I have moved BrightHouse into the department_store.json file, and removed it from variety_store.json, and I think GitHub has automatically pulled that update in to this request?

bhousel:
Looks perfect @UKChris-osm , thanks again!

PR #2977, opened 2019-08-19
Add uSmile convenience store

danielsjf:

PR #3485, opened 2020-01-13
Correct uSmile convenience stores in PetroChina fuel stations

Koverpih:
I noticed there are styling issues from #2977 not fixed before merging. The name should be "昆仑好客", and "便利店" (convenience store) is a generic suffix


bhousel:
Thanks for the updates and welcome @Koverpih  👋 
I'll adjust the `matchNames` a bit to make sure that the older less-preferred name still matches this entry.

PR #3487, opened 2020-01-13
Remove 'Horodateur' brand (is not a brand name)

QuentinS33:
'Horodateur' is not a brand. It's a commonly french name to define 'parking meter'


bhousel:
Thanks and welcome @QuentinS33 👋 
In cee938a76 I also added it to our `config/filters.json` file so that it won't appear in the index again..

Issue #3441, opened 2019-12-22
Add William H Brown


doublah:
UK estate agents
https://www.williamhbrown.co.uk/
Site claims 129 offices.

Adamant36:
Q81787271

PR #3489, opened 2020-01-13
Added to estate_agent.json

UKChris-osm:
I was initially going to add _William H. Brown_ after seeing #3441, but see @bhousel beat me to it :)

But, after seeing _William H. Brown_ was a part of _Sequence (UK) Ltd_, I've added _Sequence (UK) Ltd_ brands of Estate Agents to the estate_agent.json file.

I've also included _Sequence (UK) Ltd_ as an estate agent in itself, not because they have branded shops now, but because of the possibility of a blanket re-branding at some point, but this can be removed if need be.

### DIFF
--- a/brands/office/estate_agent.json
+++ b/brands/office/estate_agent.json
@@ -294,7 +294,7 @@
       "name": "Sequence",
       "office": "estate_agent"
     }
-  }
+  },
   "office/estate_agent|Shipways": {
     "countryCodes": ["gb"],
     "tags": {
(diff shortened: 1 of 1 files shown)


bhousel:
> I've also included _Sequence (UK) Ltd_ as an estate agent in itself, not because they have branded shops now, but because of the possibility of a blanket re-branding at some point, but this can be removed if need be.

If they don't operated branded shops now, can you remove this entry?  I'd prefer to avoid adding anything in this index that we don't want mappers adding to OSM.  (the exception is brands that are dissolved and we want to warn about).

(also another minor nitpick - you don't need to add matchnames for the "and" vs "&" in names because this is automatic).

UKChris-osm:
All done :)

bhousel:
looks great @UKChris-osm , thanks for the quick fixes 👍

PR #3490, opened 2020-01-14
Add Bolivian brands

51114u9:
this PR adds and update banks, cinemas, shops and pharmacies of Bolivia


Adamant36:
Hi there and welcome. Just a minor bit of feedback, it might help to split up the commits into different PRs next time. Otherwise, it makes it a lot harder to review everything. Also, how come you got rid of a lot of the name, name:en, name:es, brand:en, brand:es etc for a lot of the banks? From a quick glance most if not all of them are correct the way they currently are in the index. 

Thanks

51114u9:
> Hi there and welcome. Just a minor bit of feedback, it might help to split up the commits into different PRs next time. Otherwise, it makes it a lot harder to review everything.

Thanks for the comment, I will consider next time

> Also, how come you got rid of a lot of the name, name:en, name:es, brand:en, brand:es etc for a lot of the banks? From a quick glance most if not all of them are correct the way they currently are in the index.

In Bolivia there are no names and brands translated into english, these are written in spanish and people who speak in english (or another language) use the name in spanish to refer to businesses or brands, putting names and brands in english would be an invention

Adamant36:
> putting names and brands in english would be an invention

Maybe if that was what was being used in the name= tag. Name:en is a perfectly valid tag though and anyone can add it to anything. In no way does it replace the Spanish name. Nor is the english translation of the spanish name an "invention." In a lot cases its what a company calls itself outside of its country of origin or we wouldn't use it. Even in cases where they don't though, its still valid and there's QA reasons we do that way. The same goes for the name:es tag. 

You also changed BCP to Banco de Crédito de Bolivia. We don't go with long brand names like that because it doesn't show up well on a map. Anyway, BCP is what the bank has also been tagged as. Banco de Crédito de Bolivia has always had almost zero uses compared to a couple hundred for BCP. We didn't just "invent" it. Its been that that way before it was in the name index. That's how the people in Bolivia prefer to tag it. Same goes for the other entries where you did the same thing.

Adamant36:
If you want to modify the PR to just involve the brands you added, fine. That's what the PR says its about anyway. Not modifying a bunch of other brands along with it. Otherwise, this should be closed as declined IMO. 

If you think already existing tags should be systematically removed from entries because you think they are "invented", that should really be discussed in its own issue or on a mailing list somewhere. Trying to pass it through in an otherwise legitmate PR without mentioning or discussion it first isn't the proper way to do it though. 

Thanks

bhousel:
Hey @51114u9 I took a look at this.. 
1. I don't have any problem with removing redundant name tags.  This is ok.  The `:en` name tags are more important for brands we can't read, like in cyrillic, japanese, etc. But we don't really need them for Spanish, and sometimes they have been added unnecessarily.
2. My bigger concern is whether all of these additions are really notable enough to stay in Wikidata.  You've created the wikidata pages in the past few days, but it seems like some of the pharmacies and cinemas have only a handful of locations.  I'm not that we can make a case that a cinema with 5 locations is notable.

I would probably just accept the PR, but I'm worried that your wikidata additions might get deleted.   @1ec5, what do you think?

Adamant36:
Well, less lines of code in the bank file probably isn't a bad thing at least. I still don't think BCP should be converted to the full name though. But that's just me. Its not like I'm ever going to be looking at the map in Bolivia anyway.

ArenaL5:
> I still don't think BCP should be converted to the full name though.

This is a common problem and we should decide on some standard ruling.

Following normal naming conventions on the OpenStreetMap Wiki, one would normally use `name` for the full name and `short_name` for the abbreviation. I use this convention when mapping schools in Spain, where words like "High and Middle School" are part of the name, but usually abbreviated. (They tend to be longer in Spanish.)

However, for brands which technically have a full name, but no one uses it, I personally favour using `name` and `official_name`, like this:

https://github.com/osmlab/name-suggestion-index/blob/f9904a0202467fd1c2eecc40b1b50f7a98f22ab3/brands/shop/lottery.json#L24-L25

I'm not really sure if `official_name` is the right choice, but I think the full name `Banco de Crédito de Bolivia` should be added somewhere.

Adamant36:
"This is a common problem and we should decide on some standard ruling."

I'm not sure where its a common problem. There is actually a standard rule called the on the ground rule that pretty much everyone agress on. In other words, what the places sign and their logo branding says matters the most. There's already been plenty of discussion about it here to. Its not like we decide to abreviate things randomly or without considering individual examples. In this particular case, some branches go by the abreviation without the full name. Whereas others seem to use both, but none that I'm aware of only use the full name. If they did it would be an edge case. So is it better to use the full name "just because" when its only on some of the branch signs or the abrevation when its on the vast majority or all of them? Id argue the abrevation is the better one. You don't need a wider conversation for that.

ArenaL5:
My bad, I phrased my point poorly, _and_ I misunderstood the situation.

If they use BCP everywhere, of course that's one should use as `name`. What I was really arguing for was for keeping Banco de Crédito de Perú as `official_name` or some other key. (Is there any key that's more fitting than `official_name` for this cases?)

Now I checked the commit (04af49eb4f09b742b33b697a07f61dc48ef28662), and that's precisely how it was before, so I agree on reverting that too.

> Whereas others seem to use both

@51114u9's commit d122e33ffc2b8c35221586a3fffbc7fc63052645 is one of those cases - judging from their Wikipedia article and their website, it seems like they're fond of using both the full name and short name.

Should we use `name/short_name` or `official_name/name`?

Adamant36:
> My bad, I phrased my point poorly, and I misunderstood the situation.

It's cool. It happens. 

> What I was really arguing for was for keeping Banco de Crédito de Perú as official_name or some other key. (Is there any key that's more fitting than official_name for this cases?)

I don't know if its good to use the official_name tag in this situation. As its not clear what form is the official one. There's the alt_name tag. Although I don't know if that's best either. 

> judging from their Wikipedia article and their website, it seems like they're fond of using both the full name and short name.

Yeah it seems like they don't have a preference either way. In support of the full name there's other BCP banks out there. So using it as the name might be confusing. Not in favor of it though is that it only has like 4 uses compared to around 250 for BCP. So BCP is clearly the community favorite. Although, who knows how many of those are the this bank, but it has to be more then 4. 

It's worth mentioning that I've done a huge amount of work on the bank file and in the vast majority of cases abbreviations have way more uses then the full names do, if the full name is even used at all. Abbreviations, and naming in general, is something I've done a ton of research on. At least in the case of banks, it's pretty clear abbreviations are preferred by the community.

ArenaL5:
I suppose that the most proper way to use `official_name` would be to copy verbatim whatever the national company register reads. But no, there's not consensus as far as I know. `alt_name` is a good alternative, sure.

About this PR:
- I did a quick search on the cinemas and shoe shops and they're scattered in five or six cities across the countries. I can't really know if they're notable unless I live there or research them thoroughly, but chances are good.
- I also did a couple queries in https://overpass-turbo.eu/ and there are like 18 instances of Banco Nacional de Bolivia against 11 of BNB, so I wouldn't change that one.
- How would you feel about merging this after @51114u9 reverts the change to BCP? I'd rather see this dealt with than keep adding commits to resolve Git conflicts.

Adamant36:
I just merged it and fixed BCP in https://github.com/osmlab/name-suggestion-index/commit/6cea93adc2b05cd83c398d6273d5cee5b051f0c6. @51114u9 if you could add social media accounts to the Wikidata items you created it would be helpful. Otherwise, their entries aren't as useful. Thanks.

51114u9:
thanks for merging, I will add social media accounts in wikidata items

I read all comments and understand their PoV. In Bolivia this companies are notable because the quantity of population is small and disperse compare another countries, as Peru/Chile/Argentine

In the case of BCP, in Bolivia is more common use full name over abbreviations, young people use abbreviation of companies recently. BCP is ambiguous word in Bolivia, because don't represent full/official name of bank (Banco de Crédito de Bolivia), should be BCB, but BCB is abbreviation of central bank of Bolivia (Banco Central de Bolivia)...

I will discus more this case with bolivian community of mappers and user of this bank

PR #3491, opened 2020-01-14
Add and update france fuel amenities and supermarkets

QuentinS33:
Add and update some fuels and supermarkets :
- _8 à Huit :_ Add fuel amenities and add nomatch tags on 8 à Huit conveniences and supermarkets
- _Avia :_ Add countries where it is operatring (from Avia's website)
- _AS 24 :_ Add fuel amenities
- _bi1 :_ Add fuel amenities and supermarkets (only 5 in France)
- _Dyneff :_ Add fuel amenities. It is operating mainly in France. Some fuel stations in Spain

### DIFF
--- a/brands/amenity/fuel.json
+++ b/brands/amenity/fuel.json
@@ -260,6 +260,7 @@
     }
   },
   "amenity/fuel|Avia": {
+    "countryCodes": ["at", "be", "bg", "ch", "cz", "de", "es", "fr", "hu", "it", "nl", "pl", "pt", "rs", "ua"],
     "matchNames": ["station avia"],
     "tags": {
       "amenity": "fuel",
--- a/brands/amenity/fuel.json
+++ b/brands/amenity/fuel.json
@@ -215,6 +215,18 @@
       "name": "Arco"
     }
   },
+  "amenity/fuel|AS 24": {
+    "countryCodes": ["at", "be", "bg", "cz", "de", "dk", "es", "fr", "gb", "ge", "gr", "hr",
+      "hu", "ie", "it", "lt", "lu", "mk", "nl", "pl", "pt", "ro", "rs", "se", "si", "sk", "ua"],
+    "tags": {
+      "amenity": "fuel",
+      "brand": "AS 24",
+      "brand:wikidata": "Q2819394",
+      "brand:wikipedia": "fr:AS 24 (entreprise)",
+      "hgv": "only",
+      "name": "AS 24"
+    }
+  },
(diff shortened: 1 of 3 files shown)


brands/amenity/fuel.json
```diff
@@ -315,6 +338,14 @@
       "name": "BHPetrol"
     }
   },
+  "amenity/fuel|bi1": {
+    "countryCodes": ["fr"],
+    "nomatch": ["shop/supermarket|bi1"],
+    "tags": {
+      "amenity": "fuel",
+      "name": "bi1"
```
Adamant36: Seems you might have forgot the brand tag on this one.

Adamant36: Btw, I don't think only 5 stores makes them notable enough to be included.

brands/shop/supermarket.json
```diff
@@ -274,6 +274,15 @@
...
     }
+  },
+    "shop/supermarket|bi1": {
+    "countryCodes": ["fr"],
+    "nomatch": ["amenity/fuel|bi1"],
+    "tags": {
+      "brand": "bi1",
+      "name": "bi1",
+      "shop": "supermarket"
+    }
```
Adamant36: Only having 5 stores probably doesn't make them notable enough to be included. Wikidata will just delete an entry for them at that point. Plus, its a little low for the name index also.

Adamant36:
Hi there and welcome. it looks like you forgot the brand tag on bi1. Otherwise, it looks fine.

QuentinS33:
Hi @Adamant36, thx for the answer ! I understand that the brand "bi1" (supermarkets and fuel stations) . Do yout want to remove it ? I can do that
I don't write the "brand" tag because there is not a Wikipedia page for this brand. There is only a wikipedia page of the company's owner of Bi1 (it's Schiever).
See here : https://fr.wikipedia.org/wiki/Liste_d'enseignes_de_la_grande_distribution_en_France#Schiever and https://fr.wikipedia.org/wiki/Schiever

bhousel:
Thanks @QuentinS33  looks good.. 👍 

For Bi1 I do actually see a stub page for it:  https://en.wikipedia.org/wiki/Bi1
It's not linked to a Wikidata page yet, but it does seem like the brand operates in both Poland and France, so it's maybe more established than just 5 locations.

Adamant36:
>  it's maybe more established than just 5 locations.

Yeah, there's lots more then just 5 locations. I was looking at the wrong place on their site. Sorry about that.

PR #3496, opened 2020-01-16
Azerbaijanian defaulft for pharmacy added

ferdinand0101:


ferdinand0101:
removes Aptek from list since it's the name for pharmacy in Azerbaijanian

bhousel:
Looks good, thanks @ferdinand0101 👍 

Also thank you for changing the title!  When I scroll through emails on my phone, seeing "Update filters.json" doesn't really give an idea of what the PR is about..

Issue #2028, opened 2018-10-19
Should state-run 'brands' be considered brands?


good-praxis:
In particular this is about #2009 
These shops are _all_ government run because of the specific regulation [[source](https://en.wikipedia.org/wiki/Doh%C3%A1nybolt)]. They are close enough to be considered generic, but since they're all ran by the same actor, should they be considered generic? There is no tangible difference between this and a company running a franchise.

ImreSamu:
As a Hungarian - my personal  opinion

Remove
*   "shop/tobacco|Dohánybolt"
*   "shop/kiosk|Nemzeti Dohánybolt"

Keep: 
*  "shop/tobacco|Nemzeti Dohánybolt" - without a "brand"

good-praxis:
"brand" in these cases mostly don't actually mean real life brands, it's just kind if _this_ specific kind of shop (or this "brand" of shop)

ImreSamu:
@good-praxis : 

No "hungarian" (local language) version;
so I would not  add this 2 tags - to the `shop/tobacco|Nemzeti Dohánybolt`
* "brand:wikidata": "Q20639040",
*  "brand:wikipedia": "en:Dohánybolt",
imho:  the https://www.wikidata.org/wiki/Q20639040  is low quality wikidata info

the `brand=Nemzeti Dohánybolt` - maybe usefull;    
but  We (=hungarian community) need to discuss in the hungarian mail list,  
( But it takes a lot of time -  but If we have a fdifferent consensus - I  can change later )

it is ok for you ?

bhousel:
Yeah this is in some US states for state run alcohol shops too.
Virginia ABC #969  is one of these

https://www.abc.virginia.gov/
https://en.wikipedia.org/wiki/Virginia_Alcoholic_Beverage_Control_Authority
https://www.wikidata.org/wiki/Q7934236

I'm ok with it.  They often operate like a brand and have a logo and common identity.

matkoniecz:
> but We (=hungarian community) need to discuss in the Hungarian osm mail list,
> ( But it takes a lot of time

I think that it may give good results. Ask here in case of any doubts (I am unable to contribute at Hungarian mailing list due to language issues).

good-praxis:
So I guess we will put this on freeze for now? Until @ImreSamu returns with a consensus of the hu osm mail list?

ImreSamu:
@good-praxis : 
[...]

PR #3497, opened 2020-01-16
Remove default Azerbaijanian for pharmacy

ferdinand0101:


ferdinand0101:
removes Aptek from suggestion list since it's the name for pharmacy in Azerbaijanian

Adamant36:
Hi there. Are places like this just state run generics or something instead of being a "brand"? It look like there's a bunch of them with the same kind of signage etc. 

Aptek

Aptek 2

bhousel:
> Are places like this just state run generics or something instead of being a "brand"? It look like there's a bunch of them with the same kind of signage etc.

It's ok!  State-run stores with a common branding can be included.. We have a few already, check out the discussion on #2028

Adamant36:
>It's ok! State-run stores with a common branding can be included.

I know. That's why I was asking. I figured if it was a state brand (not a generic) we could keep it in the index and just create a Wikidata entry for it. Instead of removing it from the pharmacy file altogether. Which is what this PR did 🤷‍♂

bhousel:
> I know. That's why I was asking. I figured if it was a state brand (not a generic) we could keep it in the index and just create a Wikidata entry for it. Instead of removing it from the pharmacy file altogether. Which is what this PR did

oh gotcha.. Yeah I was pretty confused, as the original title didn't say what it was about.

@ferdinand0101 are you sure you want this removed?  It looks like something that we would normally include as a brand.  Ideally, someone local gets to decide.

PR #3499, opened 2020-01-17
added 7 дней (russian chemist)

ferdinand0101:
7 дней means 7 days in Russian


bhousel:
The Wikidata page and Wikipedia page for this says that 7 Days is a magazine, not a pharmacy..  Is this maybe a mismatch?

ferdinand0101:
sorry yes it is a false match

PR #3502, opened 2020-01-19
Fix English and Japanese locales for Times parking

sanecito:
Was originally:

```
name:ja=Times
name:en=タイムズ
```

and now changed to match the correct language:

```
name:ja=タイムズ
name:en=Times
```


bhousel:
Thanks @sanecito and welcome! 👋 
This is a good fix, but the place to put the fix is in [parking.json](https://github.com/osmlab/name-suggestion-index/blob/master/brands/amenity/parking.json#L60)

The `dist/name-suggestions.min.js` file is a generated file and it will get replaced when we rebuild the project.  The files under `/brand/*` are the ones to edit...

sanecito:
Thanks, will make the change+PR to the correct file. This was the only one that came up when doing a quick search w/ StartPage for "name:en=タイムズ"

PR #3504, opened 2020-01-19
Fix English and Japanese locales for Times parking

sanecito:
Was originally:

```
name:ja=Times
name:en=タイムズ
```

and now changed to match the correct language:

```
name:ja=タイムズ
name:en=Times
```

Now fixes the correct file per conversation here: https://github.com/osmlab/name-suggestion-index/pull/3502

### DIFF
--- a/brands/amenity/parking.json
+++ b/brands/amenity/parking.json
@@ -69,8 +69,8 @@
       "brand:wikipedia": "ja:パーク24",
       "fee": "yes",
       "name": "Times",
-      "name:en": "タイムズ",
-      "name:ja": "Times"
+      "name:en": "Times",
+      "name:ja": "タイムズ"
     }
   }
 }


sanecito:
Side note, when would this be reflected in iD? Would it be with a new release of name-suggestion-index or would it there also need to be a new release (and deployment) of iD to bundle the new name-suggestion-index release?

bhousel:
Looks good - can you switch the `brand:` tags also on lines 66-67?
https://github.com/osmlab/name-suggestion-index/pull/3504/files#diff-9397f13683da764b92cb8006dd82dab9R66-R67

> Side note, when would this be reflected in iD? Would it be with a new release of name-suggestion-index or would it there also need to be a new release (and deployment) of iD to bundle the new name-suggestion-index release?

It would be in the next version of iD... We release a new version of iD about once a month.

sanecito:
Only non-brand was the issue; brand tags are correct and didn't need a change unless I'm missing something syntax-wise?

```
      "brand:en": "Times",
      "brand:ja": "タイムズ",
```

bhousel:
> Only non-brand was the issue; brand tags are correct and didn't need a change unless I'm missing some pre-existing syntax issue?

Oh whoops you are right!
Sorry I was reading too quickly..

Thanks for the PR! 👍

PR #3506, opened 2020-01-20
make changes to the brands folder

Seremba:
Here are my changes


bhousel:
Sorry, I have to reject this.. Most of these `cuisine` tags look wrong.

PR #3510, opened 2020-01-22
Add M Lhuillier (shop=pawnbroker)

seav:
Add the M Lhuillier chain of pawnshops in the Philippines and remove it as a match for Cebuana Lhuillier, which is a completely different chain of pawnshops and owned by an unrelated family. (Filipinos actually incorrectly assume that the two chains are related to each other.)

This wrong match has already resulted in a mistagging of a pawnshop due to iD's suggestion to upgrade tags: https://www.openstreetmap.org/node/6827462574/history


Adamant36:
Thanks

PR #3513, opened 2020-01-22
All the banks are named Venezuela as such I added it to match name

ferdinand0101:


Adamant36:
We could probably also remove the "venezuela" entry. There's like three other banks with venezuela in the name that don't currently have entries and might be worth adding to if you feel like doing it another PR. Otherwise, its possible some of the  venezuela entries might be those banks and get miss-tagged. As it currently has 90 mapped. 

https://en.m.wikipedia.org/wiki/List_of_banks_in_Venezuela

bhousel:
> We could probably also remove the "venezuela" entry.

I merged the PR and removed the extra "venezuela" entry in 2e9fc14a6

PR #3514, opened 2020-01-23
Add some tags to fastfoods

tlewandster:


bhousel:
thanks @tlewandster and welcome 👋

PR #3525, opened 2020-02-03
amenity/college.json for SAE Institute and Akademie Deutsche POP

camelCaseNick:
I would suggest a new `./brands/amenity/college.json` for SAE Institute and Akademie Deutsche POP. I am just not sure how the discussion on `operator` tagging went, so I made it four distinct commits.

### DIFF
--- a/brands/amenity/college.json
+++ b/brands/amenity/college.json
@@ -1,7 +1,6 @@
 {
   "amenity/college|SAE Institute": {
     "tags": {
-      "brand": "SAE Institute",
       "brand:wikidata": "Q201438",
       "brand:wikipedia": "en:SAE Institute",
       "operator:wikidata": "Q201438",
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @camelCaseNick and welcome! 👋 
These look like good additions, since these universities operate as franchises in a lot of locations.

This index doesn't currently support the `operator` tags - can you remove them?

camelCaseNick:
@bhousel Okay, I have removed them now.

brands/amenity/college.json
```diff
@@ -0,0 +1,20 @@
...
+      "name": "SAE Institute",
+      "amenity": "college"
+    }
+  },
+  "amenity/college|Akademie Deutsche POP": {
+    "tags": {
+      "brand": "Akademie Deutsche POP",
+      "brand:wikidata": "Q413887",
+      "brand:wikipedia": "de:Akademie Deutsche POP",
+      "name": "SAE Institute",
```
bhousel: oh looks like this name is wrong

camelCaseNick: That was wrong, but know it shouldn't be anymore.

bhousel:
Thanks for the fixes, looks great 👍

PR #3527, opened 2020-02-04
Added Santander Cycles to amenity=bicycle_rental

rskedgell:
No value for name=* as this is the name of the individual docking station.

### DIFF
--- a/brands/amenity/bicycle_rental.json
+++ b/brands/amenity/bicycle_rental.json
@@ -35,6 +35,18 @@
       "name": "MiBici"
     }
   },
+  "amenity/bicycle_rental|Santander Cycles": {
+    "countryCodes": ["gb"],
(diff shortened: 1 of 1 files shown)


peternewman:
Hi @rskedgell ,

Any reason this didn't get merged?

rskedgell:
> Hi @rskedgell ,
> 
> Any reason this didn't get merged?

Hi @peternewman ,

Without a value for the name=* tag, the request fails testing script(s). As the name of the docking station is used rather than the name of the service/operator, it isn't something which will work with the NSI. On the other hand, OSM's id:MacLondon has added the brand:* tags to all the docking stations, so the map is in the same state it would have been had this request been suitable for NSI.
e.g. https://www.openstreetmap.org/node/4692548549

PR #3528, opened 2020-02-04
Update Mix Markt and add Mini Mix

kriswema:
Source: [mixmarkt.eu](https://www.mixmarkt.eu/en/germany/about-us/mix-markets-in-europe/)


bhousel:
Looks great! thanks @kriswema and welcome 👋

PR #3530, opened 2020-02-08
Add a link to the JOSM help page showing the presets setting dialog.

maxerickson:
The link just shows the JOSM presets dialog, it isn't specific to the NSI, so proposing the change instead of just committing it directly.


bhousel:
Thanks @maxerickson 👍

PR #3535, opened 2020-02-09
removal of dupe + incorrect entry for Q1640290

javbw:
Came to this repository to update ENEOS gas station brand name in english. found duplicate with incorrect brand:en value and deleted it. "ENEOS" is the proper brand name in english.

### DIFF
--- a/brands/amenity/fuel.json
+++ b/brands/amenity/fuel.json
@@ -3517,20 +3517,6 @@
       "name:ja": "エッソ"
     }
   },
-  "amenity/fuel|エネオス": {
-    "countryCodes": ["jp"],
-    "tags": {
-      "amenity": "fuel",
-      "brand": "エネオス",
-      "brand:en": "JXTG Nippon Oil & Energy",
-      "brand:ja": "エネオス",
-      "brand:wikidata": "Q1640290",
-      "brand:wikipedia": "en:JXTG Nippon Oil & Energy",
-      "name": "エネオス",
-      "name:en": "JXTG Nippon Oil & Energy",
-      "name:ja": "エネオス"
-    }
-  },
   "amenity/fuel|キグナス石油": {
     "countryCodes": ["jp"],
     "matchNames": ["キグナス"],


javbw:
Q1640290 is also at line 917 in this repository with the correct brand data. This is dupe at line 3520.

bhousel:
Before I merge, I just want to confirm - does the local Japanese community definitely prefer "ENEOS" on the name tag and not "エネオス" ?  I agree we should get rid of the duplicate.

javbw:
ah - the first entry doesn't have a en:  or JP tag. I will update it.

javbw:
I just added that information to the original and made another pull request for the correction
https://github.com/osmlab/name-suggestion-index/pull/3538

I am unsure of how to handle this situation, and also unsure of the grammar, so please check it! thanks.

bhousel:
> I just added that information to the original and made another pull request for the correction
> #3538

Sorry, I don't understand what you're trying to do.

ArenaL5:
I'm going to close your second pull request #3538 because it doesn't follow the preferred format for these files, and it makes this change more difficult to manage.

Your syntax is not that bad (you're only missing a comma at the end of line 924), but in this project:
- the values of `brand`, `name` and the name after `amenity/fuel|` (or whatever) are equal. Either all of them エネオス, or all of them ENEOS.
- those values _needn't_ be in English. If this is a Japanese-only company, it's very likely they map it as エネオス. In this case, we'd rather consider ENEOS the duplicate.

In any case, I'd recommend [you add a `matchNames` tag](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#matchnamesmatchtags) so that typing either ENEOS or エネオス in iD will suggest the same company.

javbw:
I was deleting a dupe and then it was pointed out the first entry was incomplete so I updated the first entry with an additional pull request for that.

I’m very inexperienced with pull requests - thanks for cleaning My mess up. 

=)

Javbw

javbw:
Javbw

> On Feb 11, 2020, at 8:48 AM, ArenaL5 <notifications@github.com> wrote:
> 
> those values needn't be in English. If this is a Japanese-only company, it's very likely they map it as エネオス. In this case, we'd rather consider ENEOS the duplicate.

The brand:en= data was the official name of the corporation rather than the proper brand name in English (which is their logo, ENEOS), so I was trying to correct that error.

Mapping the Japanese in name= and the English in the name:en= Is the documented naming policy for OSM JA tagging - most everything Japan is physically double-signed anyways - so you would expect every Japanese Brand entry to have a brand:en value associated with it. 

I discovered this odd brand:en value  on a gasoline stand by chance while mapping and came here to correct the erroneous data.  

The name linking is new to me, thanks for linking to it for me. 

Javbw.

bhousel:
> Mapping the Japanese in name= and the English in the name:en= Is the documented naming policy for OSM JA tagging - most everything Japan is physically double-signed anyways - so you would expect every Japanese Brand entry to have a brand:en value associated with it.

Thanks @javbw - I merged the PR and added a followup in aa6ec2b26 to follow this convention.

javbw:
Question: I'm still getting "brand:en=JXTG Nippon Oil & Energy" as a value when it is suggested in iD (via searching "エネオス" for the business. When I look in the directory here, it has the correct values. Does iD have an older version of these files cached? are they updated every-so-often, and it will eventually be fixed, or is this an error? 
Screen Shot 2020-04-22 at 8 23 20 PM

PR #3536, opened 2020-02-09
Jack's nomatch nomatch prop and Marquis entry

jurrish:
added nomatch property for both Jack's supermarket and Jack's fast food
added en:wikipedia for Marquis

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -1364,6 +1364,7 @@
     }
   },
   "amenity/fast_food|Jack's": {
+    "nomatch": ["shop/supermarket|Jack's"],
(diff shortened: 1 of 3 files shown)


bhousel:
thanks @jurrish, and welcome 👋 
Can you instead add these nomatch properties in the `fast_food.json` and `supermarket.json` files?

`brands.json` is a computed file that gets overwritten.  (check [CONTRIBUTING.md](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md) for more details on which files to edit).

bhousel:
also `"brand:wikipedia": "en:Maquis (restaurant)"` looks fake - there is no page there - so please do not add that one..

PR #3539, opened 2020-02-11
Fixing incorrect EN name & adding EN brand name.

javbw:
出光 = Idemitsu, which is the common name of the brand used in English signage (in Japan).


javbw:
discovered bad data when using "upgrade the tags" iD feature on https://www.openstreetmap.org/node/968835327 .

javbw:
I think I have made an error using the name tags. 
closing this.

PR #3542, opened 2020-02-13
Added Tudor's Biscuit World

pkoby:


UKChris-osm:
Are you able to add the social media (Facebook, Twitter, Youtube) entries into the WikiData article for this entry, as the Name Suggestion Index (NIS) will use these entries to download the branding logo.

pkoby:
Ah yes, that makes sense! I didn't think through how the logo would be found without being in Wikidata. I thought of adding it there, but I wasn't sure about licensing issues.

I've added the information.

UKChris-osm:
I think the use of social media links is a way of not having to worry about licensing, as these logos are published by the brand on public / social web sites :)

Regarding `"cuisine": "regional"` - is there a more specific value available that defines the cuisine available at Tudor's? You can use multiple values if need be, separated by the semicolon **;** such as `"cuisine": "burger;pizza"`.

pkoby:
I really don't know what to call it. It's centered around biscuits (that is, American biscuits that go with gravy), but I guess it could be `"cuisine": "sandwiches;burgers"` and maybe `;fried_food`. I don't think there's a `breakfast` option. If you wouldn't mind verifying (http://www.tudorsbiscuitworld.com/) and maybe suggesting other options, I'll add these tags shortly.

UKChris-osm:
As I'm from the UK I'm not sure how best to describe this cuisine, and so I'm a little hesitant to merge the pull request myself in case a better value is available. I might wait to see if any of the US admins on here can verify, but the rest of the entry you have added looks fine to me.

I'd probably go for `"cuisine": "biscuit;burgers;sandwiches"` myself.

pkoby:
I wouldn't use "biscuit", because that refers to british-style biscuits, according to the wiki (US cookies). I've added `burger;sandwich`.

UKChris-osm:
I think it would be good to get @bhousel opinion on this :)

bhousel:
> I think it would be good to get @bhousel opinion on this :)

oh yeah - I just took a look and it looks like mostly Southern-USA comfort food (like their biscuits and gravy, beans, fried steak) but with some other offerings too like sandwiches and wraps.  I'd just go with `cuisine=american` to keep it simple.  We often take it to mean - just an unspecific mix of cuisines.

bhousel:
Thanks @pkoby and welcome 👋

PR #3544, opened 2020-02-15
Update Mix Markt

kriswema:
Change the brand's Wikipedia link from German to English Wikipedia, which didn't exist at the time.


bhousel:
Since it is a German supermarket chain, let's just leave it with the German Wikipedia link.

kriswema:
Fair enough. My reasoning for this PR was that the 130 non-German Mix Markets would be better off with an English link, but this works, too.

Issue #3545, opened 2020-02-15
Add Taziki's


joecaruso:
https://www.wikidata.org/wiki/Q19849840
https://en.wikipedia.org/wiki/Taziki%27s_Mediterranean_Caf%C3%A9

> amenity=restaurant
> cuisine=mediterranean

UKChris-osm:
Are you able to add the social media entries to the Wikidata article for this entry? It would help the NSI grab the logo of the brand.

joecaruso:
I don't have a wikidata account, but here are the links 

https://twitter.com/tazikis
https://www.facebook.com/TazikisMedCafe/
https://www.instagram.com/tazikis

UKChris-osm:
Hi @joecaruso, an account isn't needed to edit a Wikidata article, you can edit it by clicking the "_+ add statement_" link roughly halfway down the page on the right.

Clicking this will bring up an entry box, which you can type in a "_property_" and in this box you can type "_Facebook_", "_Twitter_" etc and it will offer the correct property for each social media entry that is available, and in the box that appears to the right, you can add the correct username for each entry, such as:

**Facebook ID**: _TazikisMedCafe_
**Twitter Username**: _tazikis_
**Instagram Username**: _tazikis_

With each entry you add, if you also select "_+ add reference_", you can add details of where the information came from. For something like a social media entry, I tend to add "_retrieved_" which is the date I checked the data, and "_reference URL_" for the web site I found the social media link on, which is usually the home page of the brand in question.

Feel free to give it a go, as knowing how to improve Wikidata articles is a great way to help with the Name Suggestion Index (NSI) :)

Adamant36:
Closed in #3546.

PR #3546, opened 2020-02-16
Added Taziki's to fast_food.json

UKChris-osm:
Added "_Taziki's Mediterranean Cafe_" to **/brand/amenity/restaurant.json** as per #3545 but have done so as a pull request so others from the US can check it before committing.

The entry includes a matchTags for "_amenity/cafe_" which might not be needed, and the name might be better suited to just "_Taziki's_".

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -2880,6 +2880,18 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|Taziki's Mediterranean Cafe": {
+    "countryCodes": ["us"],
+    "tags": {
+      "amenity": "fast_food",
+      "brand": "Taziki's Mediterranean Cafe",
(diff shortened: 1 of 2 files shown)


bhousel:
I'd use `amenity/fast_food.json` since this place looks like a fast-casual.  

You can also skip the `matchTags` for `amenity/cafe`.  All the food entries are in the same match group [here](https://github.com/osmlab/name-suggestion-index/blob/30dc778003fb5e383531d1441a9a837e83333a36/config/match_groups.json#L77-L87) so any of them will match.

The `name` seems ok to me.  You can also add `"short_name": "Taziki's", to give renderers a shorter option.

For what it's worth, Google uses the full name.
Screenshot 2020-02-16 16 05 49

UKChris-osm:
Suggested changes made, thanks @bhousel

PR #3548, opened 2020-02-17
Add Orell Füssli and remove Switzerland from Thalia

kriswema:
It looks like Thalia and Orell Füssli have a joint venture in Switzerland, and existing Thalia shops were converted to the Orell Füssli brand. The [source](https://www.badische-zeitung.de/nachrichten/suedwest/der-name-thalia-verschwindet) listed on Wikipedia is a bit older and speaks of Thalia shops disappearing in the future, and "thalia.ch" redirects to the [Orell Füssli](https://www.orellfuessli.ch/) website, but it might still be nice to get local knowledge about this. Maybe @habi can confirm this?

### DIFF
--- a/brands/shop/books.json
+++ b/brands/shop/books.json
@@ -199,6 +199,16 @@
       "shop": "books"
     }
   },
+  "shop/books|Orell Füssli": {
+    "countryCodes": ["ch"],
+    "tags": {
+      "brand": "Orell Füssli",
+      "brand:wikidata": "Q1511140",
+      "brand:wikipedia": "de:Orell Füssli",
+      "name": "Orell Füssli",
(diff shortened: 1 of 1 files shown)


habi:
The [German Wikipedia page of the joint venture](https://de.wikipedia.org/wiki/Orell_Füssli_Thalia) mentions that since "2016, the apperance has been converted to Orell Füssli, so that the name Thalia vanished from the Swiss book market". I can confirm that one of the Thalia stores in the city I live has long been converted to Orell Füssli: https://www.orellfuessli.ch/shop/home/filialen/showDetails/1002/, but don't know about *all* the stores in Switzerland :)

Issue #3552, opened 2020-02-21
Add brand AWG-Modecenter


Bingo67:
the name of the brand AWG is AWG-Modecenter
brand:wikipedia=de:AWG-Modecenter
brand:wikidata=Q300220

bhousel:
done in #3553

Bingo67:
Thanks for adding, but now I'm not sure if the Wikipedia name AWG-Modecenter is the right name.
On https://www.awg-mode.de/ the name of the stores is AWG Mode Center. On taginfo you find a few combinations of AWG and Mode and Center.  Maybe it is a good compromise that you added AWG as short_name, because AWG has the most hits on taginfo. Can you tell me, when the change will work in iD editor?

bhousel:
> Can you tell me, when the change will work in iD editor?

I can make a name-suggestion-index release that will start working right away in the iD development mirror, but it will probably be a while before we make another official iD release so that it is available on osm.org

Bingo67:
Ok, i can wait :-) until now it is not working (see attachment)
![2020-02-25 10_42_15-Window](https://user-images.githubusercontent.com/59020063/75235933-55e92680-57bd-11ea-9308-7c8c173784ca.jpg)

PR #3553, opened 2020-02-21
corrected the AWG brandname to AWG-Modecenter and added wikipedia and…

m0hamm3d-ac:
… wikidata information

fix to issue - https://github.com/osmlab/name-suggestion-index/issues/3552

### DIFF
--- a/brands/shop/clothes.json
+++ b/brands/shop/clothes.json
@@ -21,12 +21,14 @@
       "shop": "clothes"
     }
   },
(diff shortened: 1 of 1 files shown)


bhousel:
thanks @m0hamm3d-ac and welcome! 👋

bhousel:
I also added "AWG" in the `short_name` tag in aecde4efa.  This will tell our matcher to match that name as well.

PR #3555, opened 2020-02-22
added german prepschools

### LINKED ISSUE #3554: after school tutoring brands (de)
There are Schülerhilfe (Schülerhilfe(Q2259702))
Studienkreis (Studienkreis(Q2358555))

I would like to add them but I'm unsure if I should use the unsupported but used amenity tutoring or amenity school which is documented but doesn't fit well

ferdinand0101:
fixes #3554


bhousel:
looks great, thanks @ferdinand0101  👍

PR #3562, opened 2020-02-28
added Fatz brand

valoTN:


brands/amenity/restaurant.json
```diff
@@ -842,6 +842,15 @@
       "name": "Famous Dave's"
     }
   },
+  "amenity/restaurant|Fatz": {
+    "countryCodes": ["ca", "us"],
```
bhousel: Took a quick look at their wikipedia and I don't think they operate in Canada?
```suggestion
    "countryCodes": ["us"],
```

bhousel:
Thanks for the quick fix, and welcome! 👋

bhousel:
Oh one other thing - I noticed that you added Degussa bank in this PR too, thanks!
Because it's a German bank, I switched the `brand:wikipedia` tag to point to the German Wikipedia in 0080376a8

PR #3571, opened 2020-03-01
Add brand of vending machines in Japan

sguinetti:
Greetings. I added several brand of vending machines in Japan with extra information. Also, I added some brands in katana for easy orientation. Por example: Kirin, Calpis, Asahi and others.


Adamant36:
Thanks for the PR. I think I'm going to merge it since I see nothing wrong with it and it's been a week without any objections from anyone.

Adamant36:
Btw, are you up to being an admin now so that you can directly merge your own stuff? It would make things a little easier that way.

sguinetti:
@Adamant36 You are right and I accept the request. I was thinking the reasons before because it is the responsibility topic. Instead of constantly interrupting with pull request, it would be best to use this permission and add popular brand in Japan and Peru. Thanks for the reply.

Adamant36:
Ok. @bhousel, would you mind inviting @sguinetti to be a collaborator now?

PR #3579, opened 2020-03-05
Added Croatian pharmacies

Janjko:


UKChris-osm:
Hi @Janjko, do you have any **brand:wikidata** entries to go along with your Pharmacy additions?

Adamant36:
At this point it might be worth just merging anyway and then we can create separate issues for Wikidata entries or something. At least that way they will be in the index and it keeps someone else from opening another PR for the same brands while we wait. Which seems to be happening a lot.

Adamant36:
I just decided to merge it. I'll probably create Wikidata entries for them now that they are in the index if someone else doesn't do it first.

Janjko:
I didn't know you guys were adding Wikidata entries without Wikipedia pages. Don't they remove those for lack of notability?
I'll be adding them in the future then, if you didn't do it now.

Adamant36:
Wikidata and Wikipedia are different projects with different notability standards. Luckily for us, Wikidata's standards of notability are slightly lower. Really all you have to do is provide an external link in a reference and make sure you add enough statements for the entry to be substantial enough. On the projects side, Wikipedia isn't that important anyway since articles tend to change frequently and we are getting the icons etc from Wikidata anyway.

Issue #3327, opened 2019-11-01
Add multiple Sparkasse Brands


ferdinand0101:
There are Curremtly two allready in the project. the one being from Hamburg and the other one being from brandenburg
I'd like to add the ~400 other (those with 10 more shops)

my question would be if this is allowed

Adamant36:
This was explored in #2475 and unfortunately got shot down by the German community due to the discussion [here](https://forum.openstreetmap.org/viewtopic.php?id=65695). Personally, I'd like to see them added. As its better to have some kind of unified tagging then not, but as the saying goes "The will of the people something something..Etc etc." So, it's a no go without the German communities buy in at this point 😞

PR #3583, opened 2020-03-07
Add German Sparkassen

kriswema:
> [Savings banks in German-speaking countries are called Sparkasse (pl: Sparkassen).](https://en.wikipedia.org/wiki/German_public_bank)

Two of them were already added, so I figured I might as well complete the list. The diff looks a bit strange on my end, but `amenity/bank|Mittelbrandenburgische Sparkasse` is the only entry I actually amended - all other changes are new entries.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -2317,6 +2317,16 @@
       "name": "Beobank"
     }
   },
+  "amenity/bank|Berliner Sparkasse": {
+    "countryCodes": ["de"],
+    "tags": {
+      "amenity": "bank",
+      "brand": "Berliner Sparkasse",
+      "brand:wikidata": "Q821760",
+      "brand:wikipedia": "de:Berliner Sparkasse",
+      "name": "Berliner Sparkasse"
+    }
+  },
   "amenity/bank|Berliner Volksbank": {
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @kriswema , I can tell you did a lot of work on this..

We were previously unsure of whether include these in the index, see discussion on #275, #2475, #3327  

For now the sparkassen are specifically excluded until we can get confirmation about how the German community wants them mapped.  So while I'd like to see these added to the index, I think we need to wait for a clearer consensus before merging.

kriswema:
The branches I saw on OSM were using the full names (e.g. _Kreissparkasse Köln_), so based on my experience, this is in line with how they'd appear. Anyway, once a decision is made, the changes will be waiting, and I had a nice meditation exercise either way :)

ferdinand0101:
I also recently made an issue about this but I was told to look at #2475 which is a simmerlar pull request which got closed and this [link](https://forum.openstreetmap.org/viewtopic.php?id=65695) to the german forum.

Adamant36:
Closing until there's consensus. Thanks for the effort though @kriswema. As an aficionado of banks I'd really like to see these included in the index at some point.

PR #3585, opened 2020-03-08
Added some Dutch stores

RobinLinde:
Also fixed an incorrectly formatted Wikipedia link


Adamant36:
Everything looks good. Except on OV-fiets we don't use websites. So, any chance you can remove it?

Adamant36:
Never mind. I just merged it and got rid of the website in 290259b. Thanks for the PR 👍

PR #3593, opened 2020-03-12
Add vending machines (part 2) plus several brands of Japan

sguinetti:
Extension of: #3571

Now other brands from different financial entities are added. Several of the most popular vending machines because of the nearly 5 millions that exist in 2014 were also added. For example: Acure (JR).

I am trying to avoid any conflict with pull requests. In addition to adding sources about the brands (since first web searching generates more confusion like: "list of crazy..."). Exception is a little edit of "ja" tag to "jp".

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -8095,6 +8095,20 @@
       "name:zh-Hant": "東亞銀行"
     }
   },
+  "amenity/bank|東京ベイ信金": {
+    "countryCodes": ["jp"],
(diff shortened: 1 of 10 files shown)


Adamant36:
So, not going to do it in smaller PRs huh?

sguinetti:
I will try to divide the data into two branches. One about brands and the other miscellaneous. With respective pull request.

sguinetti:
They're ready:  #3597 and  #3596

PR #3594, opened 2020-03-12
Fix Rimi supermarket entry

richlv:
* Fix brand wikipedia reference
* Remove incorrect wikidata entry (pointing at the defunct Norwegian chain entry; there does not seem to be an entry for the Baltic one)

### DIFF
--- a/brands/shop/supermarket.json
+++ b/brands/shop/supermarket.json
@@ -3539,8 +3539,7 @@
     "countryCodes": ["ee", "lt", "lv"],
     "tags": {
       "brand": "Rimi",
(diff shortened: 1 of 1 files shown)


papuass:
Rimi Baltic Wikidata entry
https://www.wikidata.org/wiki/Q3741108

richlv:
Thank you for the Wikidata info - searching for "Rimi" there turned up all kinds of weird results, just not that one...

papuass:
image
Wikidata item for Wikipedia article can be found here.

bhousel:
Thanks @richlv and @papuass !

PR #3596, opened 2020-03-12
Several of finance and shop brands of Japan

sguinetti:
Extension of: #3571

Now other brands from different financial entities are added. Fix little edit of "ja" tag to "jp". Also, added shops and ticket office.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -7780,6 +7780,21 @@
       "name": "京城商業銀行"
     }
   },
+  "amenity/bank|京葉銀行": {
+    "countryCodes": ["jp"],
+    "tags": {
+      "alt_name:en": "αBANK",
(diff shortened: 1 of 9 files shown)


Adamant36:
Thanks. This is a tad better. Although only slightly. I'd shot for maybe half that next time. There's no rush to get these merged as they aren't added to iD Editor immediately. Considering that, I see that a lot of them don't have social media links in Wikidata. It would be good if you slowed down a little on the amount that your adding now and maybe focus on adding social media links to the ones that have been merged. Otherwise, they won't be usable in the next update of iD Editor anyway.

sguinetti:
Right. I know this problem from my limited knowledge of Git. The best way I found is to split the pages into different branches. I hope to keep improving in the future.

Also, I'll see how to add data from these brands as you propose with typing corrections.

PR #3597, opened 2020-03-12
Add Japan vending machines (part 2)

sguinetti:
Extension of: #3571

Several of the most popular vending machines because of the nearly 5 millions that exist in 2014 were also added. For example: Acure (JR). In addition to adding sources about the brands for easy edition (since first web searching generates more confusion for the user like: "list of crazy...").

### DIFF
--- a/brands/amenity/vending_machine.json
+++ b/brands/amenity/vending_machine.json
@@ -153,6 +153,19 @@
       "vending": "ice_cubes"
     }
   },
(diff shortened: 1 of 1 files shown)


Adamant36:
Looks good. Please see my comment in the other PR about taking the time to add social media account details to their Wikidata entries now so that they are useful in iD Editor. Thanks 👍

PR #3601, opened 2020-03-14
removal of dupe & brand/update for Cosmo gas.

javbw:
consolidated 2 entries for cosmo, and updated brand and name tags. 
please check to make sure my syntax is correct. 
discovered the incorrect brand value(s) while mapping.


Adamant36:
Thanks. I don't speak Japanese unfortunately, but it looks fine from just giving it a quick glance over.

PR #3602, opened 2020-03-14
Add Wikidata for incomplete japanese brands

sguinetti:
After reviewing in depth. I did some tinkering for some brands. Added info for Skylark restaurant brands and fixed duplicate code info. I also contributed separately in adding social networks to Wikidata.

### DIFF
--- a/brands/amenity/pharmacy.json
+++ b/brands/amenity/pharmacy.json
@@ -1835,6 +1835,8 @@
       "amenity": "pharmacy",
       "brand": "さくら薬局",
       "brand:ja": "さくら薬局",
+      "brand:wikidata": "Q11299021",
+      "brand:wikipedia": "ja:クラフト (ドラッグストア)",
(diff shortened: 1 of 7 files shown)


Adamant36:
Looks fine. definitely easier to review this time ;) One question though, where does An3 come from and is that what the store signs say or do they say Ansuri?

sguinetti:
In some websites this shop is writed "ansuri", translating into English (beacuse there isn't number in japanese). But, the Twitter account of this brand and Wikipedia says "an3". https://ja.wikipedia.org/wiki/%E3%82%A2%E3%83%B3%E3%82%B9%E3%83%AA%E3%83%BC and https://twitter.com/an3_keihan So leave "an3" as the main and "ansuri" as the alternative name.

Adamant36:
That should be fine then. If not, someone can always open another issue about it.

PR #3606, opened 2020-03-15
Add japanese optics brands

sguinetti:
This pull request adds Zoff and Jins with Wikidata and social networks.


Adamant36:
Thanks 👍

PR #3610, opened 2020-03-15
Add insurance office of Japan

sguinetti:
This pull request adds 3 insurance brands: Nissay, Mieji Yasuda and Dai-ichi.


Adamant36:
Thanks. Looks good 👍

PR #3577, opened 2020-03-03
Created "audiologist.json" under /brands/healthcare/

UKChris-osm:
Created "audiologist.json" under "/brands/healthcare/", and added "Imperial Hearing", a UK hearing specialist, with roughly 36 branches / locations.

Created a pull request to enable it to be reviewed by others, rather than just adding it myself, as it involved creating a new file.

PR #3615, opened 2020-03-16
Create anime.json with three brands

sguinetti:
This pull request adds three brands of anime store: K-Book (with matchtag for books), Lashinbang  and Mandarake.


Adamant36:
Thanks. I'm not 100% solid on the procedure for adding new files, but it looks fine and there wasn't any complaints on Github's side. So, I'm going to merge it. Then if breaks anything, and I don't see why it would, you can blame me for it 😉

sguinetti:
Nice to comment. I think that someone created a file with json #3577. And there were no problems.

PR #3617, opened 2020-03-17
Add Koide Camera and Palette Plaza

sguinetti:
Add Koide Camera and Palette Plaza, a japanese brand of photo shop.


Adamant36:
Thanks. Looks good.

PR #3618, opened 2020-03-17
Add Takarakuji (National Autonomous Lottery)

sguinetti:
This lottery stand is used for gambing activities in the country. Maybe, the only legally recognized brand.

Note: The "~(Japan)" is because in Wikidata there are two similar names, lottery, and other is related to lotteries from several countries. This avoids a conflict of adding a non-Japan POI.

### DIFF
--- a/brands/shop/lottery.json
+++ b/brands/shop/lottery.json
@@ -26,7 +26,7 @@
       "shop": "lottery"
     }
   },
-  "shop/lottery|宝くじ~(Japan)": {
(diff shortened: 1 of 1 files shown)


Adamant36:
The Japanese disambiguation isn't necessary if it has the country code is it? I don't see any mapped outside of Japan anyway.

sguinetti:
Right. For now there is no conflict as you say. So I remove that text.

Adamant36:
OK.

PR #3619, opened 2020-03-17
Add Karada Factory

sguinetti:
Karada is a japanese massage chain with 300 shops inside country. The japanese translation is "Body Factory", but its english name is keeped as "Karada Factory".


Adamant36:
Is there a need for the ka・ra・da match name? I'm not seeing it used that way anywhere.

sguinetti:
Okay. I removed this line.

Adamant36:
OK. Thanks 👍

PR #3623, opened 2020-03-18
Add Parking brands of Japan

sguinetti:
This pull request adds three brands of parking in Japan: Apple Park, Paraca and NPC (Nippon Parking).


Adamant36:
Awesome. The more the merrier. I actually didn't know that the index covered parking. I'll have to look for some American parking brands to add.

PR #3624, opened 2020-03-19
Add Viva Home and Super Viva Home

sguinetti:
This pull request adds [Viva brands](https://www.vivahome.co.jp/tenpo_annai/) for DIY shop. One is Viva Home and other is Super Viva Home.

### DIFF
--- a/brands/shop/doityourself.json
+++ b/brands/shop/doityourself.json
@@ -681,14 +681,11 @@
   "shop/doityourself|スーパービバホーム": {
     "countryCodes": ["jp"],
     "tags": {
-      "brand": "ビバホーム",
-      "brand:en": "VIVA Home",
-      "brand:ja": "ビバホーム",
+      "brand": "VIVA Home",
       "brand:wikidata": "Q11229278",
       "brand:wikipedia": "ja:LIXILビバ",
-      "name": "スーパービバホーム",
-      "name:en": "SUPER VIVA Home",
-      "name:ja": "スーパービバホーム",
+      "name": "SUPER VIVA Home",
+      "official_name": "LIXIL SUPER VIVA Home",
       "shop": "doityourself",
       "short_name:en": "SVH"
     }
@@ -710,14 +707,12 @@
   "shop/doityourself|ビバホーム": {
     "countryCodes": ["jp"],
     "tags": {
-      "brand": "ビバホーム",
-      "brand:en": "VIVA Home",
-      "brand:ja": "ビバホーム",
+      "brand": "VIVA Home",
       "brand:wikidata": "Q11229278",
       "brand:wikipedia": "ja:LIXILビバ",
-      "name": "ビバホーム",
-      "name:en": "VIVA Home",
-      "name:ja": "ビバホーム",
+      "name": "VIVA Home",
+      "name": "SUPER VIVA Home",
+      "official_name": "LIXIL SUPER VIVA",
       "shop": "doityourself",
       "short_name:en": "VH"
     }
--- a/brands/shop/doityourself.json
+++ b/brands/shop/doityourself.json
@@ -452,6 +452,19 @@
       "shop": "doityourself"
     }
   },
+  "shop/doityourself|SUPER VIVA Home": {
+    "countryCodes": ["jp"],
+    "matchNames": ["スーパービバホーム"],
+    "tags": {
+      "brand": "VIVA Home",
+      "brand:wikidata": "Q11229278",
+      "brand:wikipedia": "ja:LIXILビバ",
+      "name": "SUPER VIVA Home",
+      "official_name": "LIXIL SUPER VIVA Home",
+      "shop": "doityourself",
+      "short_name:en": "SVH"
+    }
+  },
   "shop/doityourself|Screwfix": {
     "countryCodes": ["de", "gb"],
     "tags": {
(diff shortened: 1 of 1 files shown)


sguinetti:
Hello and excuse me, do you have any suggestions about these brand names?

Adamant36:
> Hello and excuse me, do you have any suggestions about these brand names?

IMO they should just be in English since that's what is on the signs. There's zero to have the entries in Japanese. You can manually add brand:jp and name:jp using a MapRoulette challenge though (hopefully while also preserving the English tags).
saitama-super-viva-home-plus-155143

sguinetti:
So do you mean that I should put the English name in the name tag? Because the case is similar to that of Toys R Us.

Adamant36:
> So, do you mean that I should put the English name in the name tag with japanese name in matchNames? Because the case is similar to Toys R Us shops.

The whole thing should be in English, There's zero reason to have Japanese in the entries at all IMO. That it's located in Japan doesn't really matter anymore then it did with the Toys R Us thing did. The difference between that one and these is that I assume there aren't any entries for these at all. English or otherwise. Whereas, there already was for Toys R Us. It's better to have an entry in the NSI then not. Whatever language. So, I rather not just close it like I did with the other one.

Adamant36:
BTW, aside from the reasons I've already given for why going with Japanese isn't a good idea in cases like this, another thing to think about is that it is pretty easy for a regional brand to open stores in other countries. Also, a lot of stores can have a couple of one off locations somewhere outside of their main area that we might know about. So making the entry in Japanese when it's not what is on the signs could potentially introduce errors and lead to some hassle later. Rarely is there a "Japanese" brand or an "American" brand or whatever. It's better to assume the company knew what they where doing when they went with English instead of Japanese and had a good reason for it. I highly doubt it would confuse anyone in Japan that is using OSM to find their locations.

sguinetti:
I see what you mean, you want to differentiate translation (translating names) with localization (regionalization, or something like that). Although as they are Japanese companies it is complicated to know which ones will have the international name or not, which depends on the exclusivity of the country. The clear example is from McDonalds [which is  documented in the repository](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md). Name:ja (no name:jp, jp is for country). Something else, don't you think it's strange that names in non-Latin alphabet are considered on the NSI list?

In the case of Home Viva. You are somewhat right because the logos are shown only in English. Unfortunately there is no Wikipedia article in English (which would be correct). So I will make some changes so you can check if it is right to use the English name with matchtag for Japanese name.

sguinetti:
Incidentally, I think there are few of us who talk about these issues related to Japanese brands. In fact, I'm a foreigner to that country and a bit of a novice to Japanese communities.

bhousel:
Just catching up on this thread... I'm pretty ok with having multilingual tags on the entries if it helps people make multilingual maps for multilingual communities.

Please read #2236 and #2245 for more info and discussion.

Try to follow these guidelines for the Japanese brands:
* `name` - usually what is written on the signs, but see #2236
* `name:ja` - the name, but in Japanese
* `name:en` - the name, but in English
* add international name tags (`name:xx`) if it helps local mappers, or if the name is unreadable in English
* if you start adding international name tags (`name:xx`) make sure at least one of them matches the `name` tag because validation software expects this, see #2245


So, for this issue, I would do this (comments added):
```js
  "shop/doityourself|SUPER VIVA Home": {
    "countryCodes": ["jp"],
    "tags": {
      "brand": "VIVA Home",     // how the brand calls itself
      "brand:en": "VIVA Home",
      "brand:ja": "ビバホーム",
      "brand:wikidata": "Q11229278",
      "brand:wikipedia": "ja:LIXILビバ",
      "name": "SUPER VIVA Home",      // what it says on signs
      "name:en": "SUPER VIVA Home",
      "name:ja": "スーパービバホーム",
      "shop": "doityourself",
      "short_name:en": "SVH"   // I would only add this if locals recognize it
    }
  },
```

Adamant36:
> I'm pretty ok with having multilingual tags on the entries if it helps people make multilingual maps for multilingual communities.

I thought we were still favoring what is on the shops signs though. If not that's cool, but it seemed to me like that was what came out of the conversations. Plus, it respects the on the ground rule that way and in #2236 you said the name tag should be generally recognizable. If the sign is in English and the name tag is in Japanese, it wouldn't be. Where ever it is geographically located.

bhousel:
> I thought we were still favoring what is on the shops signs though.

Yes that is what I said in the rest of my comment, and the example.

Adamant36:
Alright. Just thought id double check.

sguinetti:
Thanks for the answers. Believe me, this is a challenge to interpret the marks into English without losing the conventions of names in Japanese territory. Now I'm going to edit what @bhousel says. I look forward to continuing those discussions.

sguinetti:
Also, I edit VIVA Home brand like this (respecting the name of the VIVA Home brand):
``` "shop/doityourself|SUPER VIVA Home": {
    "countryCodes": ["jp"],
    "tags": {
      "brand": "VIVA Home",
      "brand:en": "VIVA Home",
      "brand:ja": "ビバホーム",
      "brand:wikidata": "Q11229278",
      "brand:wikipedia": "ja:LIXILビバ",
      "name": "SUPER VIVA Home",
      "name:en": "SUPER VIVA Home",
      "name:ja": "スーパービバホーム",
      "shop": "doityourself",
      "short_name:en": "SVH"
    }
  },

bhousel:
Thanks @sguinetti it looks good :+1:

PR #3629, opened 2020-03-20
added V-Markt

vianarafael:


Adamant36:
Hi there. Thanks for the contribution. Everything looks good except that Georg Jos Kaes seems to be a holding company for four different supermarket brands. Including V-Markt. So using their Wikipedia and Wikidata entries for the brand isn't correct in this case and Wikidata entries will have to be created for the different companies if there isn't ones for them already. If you want to create the Wikidata entries that would be fine, but if not, someone else (probably me) can do it.

bhousel:
I'm going to close this as stale.. It does link to the "Georg Jos Kaes" parent company entry instead of a brand specific "V-Markt" entry.  Let's revisit once the more specific brand entry has been created.

PR #3632, opened 2020-03-20
Add Karaoke box chain, create karaoke_box.json

sguinetti:
This pull request create karaoke_box.json file with Big Echo and Joysound, japanese brands. Edit: Added 3 more.


Adamant36:
On this one you don't need the match name for the lower case instance since iD Editor accounts for them.

sguinetti:
Ok. I didn't know about it.

PR #3633, opened 2020-03-21
Create Internet cafe file with 5 brands

sguinetti:
This pull request create a internet_cafe.json in amenity folder with three brands: Manboo, Media Cafe Popeye and Space Create. All from Japan. Edit: Added Kaikatsu and Aprecio.


Adamant36:
On this one I assume you can go in the cafe without paying to get in. So do you think it would it be internet_access:fee=yes instead of fee=yes?

sguinetti:
RIght. I fixed it.

PR #3634, opened 2020-03-21
Add Takaoka Toko

sguinetti:
This pull request adds a individual charging station by Takaoka Toko, a japanese brand.


Adamant36:
How are you doing this? Are you creating different branches for everything or something?

sguinetti:
I don't exactly understand your question. Perhaps you are surprised, now I use `merge upstream/ master` option to update the latest commits and avoid conflicts in the pull requests. However, I'm adding brands in little branchs. Do you have any suggestions to improve this?

sguinetti:
With the help of Atom Editor, I create new branches from already updated master (from my repository), and I set commits of brands for the new branch and publish a pull request. So, in next time, I create other branches from master without the previous commits (because I publicate the pull request) and set new commits for other brands. And so on.

Adamant36:
No worries. I was just contemplating why it is including a list of all the commits from your other PRs instead of keeping them separate. Like this PR included 12 commits when it was really just one. I guess it doesn't really matter though. Since it only changes the one files. 

It just makes the commit history for the repository longer then it would be otherwise, but that's not really a problem I don't think. I'm 100% up on the many nuances (announces?) of Github unfortunately. From what I do know, it shouldn't be keeping old commit histories in the new PRs if you are creating new branches off the master branch to do them. So, it sounds like it's nothing that I know of on your end at least.

sguinetti:
This is the [original source](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/configuring-a-remote-for-a-fork) about upstream repos. Maybe you have more knowledge about it.

PR #3635, opened 2020-03-21
ADD other Estonian supermarket brands

druzanov:


Adamant36:
Hello and welcome to the project 👋 Everything looks fine. Except, how come you didn't add the Wikipedia article for Coop Eesti?

druzanov:
Hello. There is none in English, only in Estonian. Could add that and then recommit.

Adamant36:
Oh. That's fine. It doesnt have to bw in English. Its actually better if its thw Wikipedia article in whatever language is spoken where the brand is located. So if you want to add the Estonian article and recommit it that would be great.

Adamant36:
Thanks 👍

PR #3637, opened 2020-03-21
Added supermarket chain: Smatch

DavidMoraisFerreira:
Hi,
I noticed that the supermarket chain "Smatch" does not yet exist in the NSI. "Smatch" and "Match" are two different brands owned by the same group (they have different logos as well).


Adamant36:
Thanks.

PR #3640, opened 2020-03-22
Changed TagInfo reference to "release"

UKChris-osm:
Changed TagInfo reference to "release" in place of the former "master" branch of iD, which no longer exists.


Adamant36:
Interesting. Thanks for figuring it out.

PR #3643, opened 2020-03-22
Added missing cuisine tags and resolved issue 3567

sssuthe:
My first pr to this project. Let me know if anything looks out of sorts.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -3177,16 +3177,6 @@
       "name": "Ecobank"
     }
   },
-  "amenity/bank|Emigrant Savings Bank": {
-    "countryCodes": ["us"],
-    "tags": {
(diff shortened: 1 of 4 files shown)


Adamant36:
Hello and welcome to the project. Everything looks good. Except with the cuisine tags. If the places tagged as cuisine=bar or pub are things and not restaurants they should just have their entries moved to those .json files. Whereas, if they are still restaurants but have bars the proper thing would be to add the bar=yes tag to their entries. So, more research should probably be done to figure out what the best move with them is.

sssuthe:
I couldn't see a precedent in restaurant.json where bar=yes. For now I removed the bar and pub entries. Let me know if there's anything else I can do.

bhousel:
This seems ok, thanks @sssuthe  :+1:

PR #3644, opened 2020-03-22
Add "Toys R Us" (Japan)

sguinetti:
This pull request adds the japanese brand of Toys R Us.

### DIFF
--- a/brands/shop/toys.json
+++ b/brands/shop/toys.json
@@ -230,7 +230,7 @@
       "brand": "トイザらス",
       "brand:en": "Toys R Us",
       "brand:ja": "トイザらス",
-      "brand:wikidata": "Q696334",
+      "brand:wikidata": "Q89826762",
       "brand:wikipedia": "ja:トイザらス",
       "name": "トイザらス",
       "name:en": "Toys R Us",
--- a/brands/shop/toys.json
+++ b/brands/shop/toys.json
@@ -223,5 +223,19 @@
       "name": "Кораблик",
       "shop": "toys"
     }
+  },
+  "shop/toys|トイザらス ": {
+    "countryCodes": ["jp"],
+    "tags": {
+      "brand": "トイザらス",
+      "brand:en": "Toys R Us",
+      "brand:ja": "トイザらス",
+      "brand:wikidata": "Q696334",
+      "brand:wikipedia": "ja:トイザらス",
+      "name": "トイザらス",
+      "name:en": "Toys R Us",
+      "name:ja": "トイザらス",
+      "shop": "toys"
+    }
   }
 }


bmillemathias:
Hello @sguinetti 

Toys R Us is already in the list, what's the rationale to have a country specific version of a brand we already have ?

SilentSpike:
@bmillemathias Presumably signage is different in Japan

Adamant36:
> @bmillemathias Presumably signage is different in Japan

It seems to be exactly the same. I'm not sure that means we should close this as redundant though. Although, I'm not sure it means we should merge it as not either. But I don't think we should have a unique entry for every possible language of thing. Especially if the brand isn't in that language. 

There's been a lot of removal of the English language from brands that are none English. So, I think the consensus is to just have entries for brands in their respective language if it can be helped, without creating a bunch of other "none-native" entries or whatever. Since this is an English brand and the signage is in English that's what we should stay with IMO.

sguinetti:
Do you have any suggestions on that? Because the [website keeps](https://www2.toysrus.co.jp/store/) the name of the stores in Japanese. And in the OSM Wiki, the most popular [Japanese names](https://wiki.openstreetmap.org/wiki/JA:Naming_sample) are in "name" tag.

I will see if I add a Wikidata with the brand name in Japanese, its branch, its social network and the website.

Adamant36:
> Personally, I do not agree to use only English names in Japan. Do you have any suggestions about this brand?

Usually we go with what's on the store signs. What's on the website doesn't matter as much because it's a geographical map that people are using to find things in the real world with. Maybe there are obscure edge cases in the real world where it would be necessary to have a unique translation in the index despite what the signs say, but I can't think of one and everyone knows what a Toys R Us store is when visiting one or when looking at it on the map. Otherwise, the company itself would have created Japanese specific signage just for the stores in Japan so as to not confuse the locals. 

Like in America we have Mitsuwa Marketplace. If there was an entry for it, there would be no need to have a specific entry for it that says "Peace Marketplace" Just so people in America can understand the meaning of Mitsuwa. It might confuse people who look at the map and think Peace Marketplace is a different one then Mitsuwa Marketplace because they don't know Mitsuwa means piece since it really doesn't matter when your trying to find them. Plus, we aren't a translation service anyway. Region/language specific software can translate the name key into whatever region/language its for. That saves us from having an exhaustive list of unique entries for every language or a ton of name:* whatever variations. Which would be the correct way to do this if we where going to instead of a separate entry BTW. But generally, we only stick to two languages. English and whatever the "native" language of the brand is. It wouldn't be doable to add more. Personally, I'd love to have name:es added to every entry, but sadly its not to be.

sguinetti:
Thanks for the advice. So I'm thinking of creating a new Wikidata with the Japan branch (in fact it's the [only one in Asia](https://asia.nikkei.com/Business/Companies/Toys-R-Us-sells-Asia-operations-and-keeps-Japan-business-alive) since the parent company went bankrupt).

Adamant36:
> So I'm thinking of creating a new Wikidata with the Japan branch

I'm not sure what the benefit of that would be. It's still the same company, and still wouldn't warrant a separate entry IMO. On Wikidata's side the entries aren't tied to a specific country or language anyway. So, you'd just be creating a duplicate entry that would probably get deleted. At least that's my opinion.

sguinetti:
I understand how risky it is. I added as much information as possible to make it relevant enough. Perhaps you will find another exclusive detail of the country. If it doesn't work, it is best to discard this pull request.

Adamant36:
Even with a new Wikidata item it still involve creating a new entry in the NSI where we don't need one for the reasons I laid out above. So, I'm going to close this. I appreciate the effort and discussion though. You can always do a MapRoulette challenge or something to re-tag the Toys R Us stores in Japan with the new Wikidata entry or whatever other tags you want. It should be pretty doable. Although, I'd make sure changing the tags is the census of the Japanese OSM community first. Otherwise, you'll just reverted, but at least that would be a good middle ground between nothing and us adding an entry.

PR #3655, opened 2020-03-23
Add Sagawa Express and Yamato Transport

sguinetti:
This pull requests adds courier brands, Sagawa Express and Yamato Transport.

### DIFF
--- a/brands/office/courier.json
+++ b/brands/office/courier.json
@@ -1,0 +1,30 @@
+{
+  "office/courier|ヤマト運輸": {
+    "countryCodes": ["jp"],
+    "tags": {
+      "brand": "ヤマト運輸",
+      "brand:en": "Yamato Transport",
+      "brand:ja": "ヤマト運輸",
+      "brand:wikidata": "Q6584353",
+      "brand:wikipedia": "ja:ヤマト運輸",
+      "name": "ヤマト運輸",
+      "name:en": "Yamato Transport",
+      "name:ja": "ヤマト運輸",
+      "office": "courier"
+    }
+  },
+  "office/courier|佐川急便": {
+    "countryCodes": ["jp"],
+    "tags": {
+      "brand": "佐川急便",
+      "brand:en": "Sagawa Express",
+      "brand:ja": "佐川急便",
+      "brand:wikidata": "Q7399082",
+      "brand:wikipedia": "ja:佐川急便",
+      "name": "佐川急便",
+      "name:en": "Sagawa Express",
+      "name:ja": "佐川急便",
+      "office": "courier"
+    }
+  }
+}
--- a/brands/office/post_office.json
+++ b/brands/office/post_office.json
@@ -1,7 +1,8 @@
 {
-  "office/courier|ヤマト運輸": {
+  "office/post_office|ヤマト運輸": {
     "countryCodes": ["jp"],
     "tags": {
+      "amenity": "post_office",
       "brand": "ヤマト運輸",
       "brand:en": "Yamato Transport",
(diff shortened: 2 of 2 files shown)


Adamant36:
What are these being tagged on exactly? Is it like a mail/shipping store or something?

sguinetti:
As far as I know, these are brands for transporting of private and commercial goods. For example, [Yamato](http://www.global-yamato.com/en/hands-free-travel/facilities/) shows delivery icons of goods. Japan Post is for sending mail (and I imagine it's the only one) for [the world](https://www.post.japanpost.jp/int/index_en.html).

sguinetti:
Accourding to [Wikipedia](https://en.wikipedia.org/wiki/Courier):

> Couriers are distinguished from ordinary mail services by features such as speed, security, tracking, signature, specialization and individualization of express services, and swift delivery times, which are optional for most everyday mail services. As a premium service, couriers are usually more expensive than standard mail services, and their use is normally limited to packages where one or more of these features are considered important enough to warrant the cost.

In short, the right thing to tag is "courier office" than "mail office".

Note: Logist is for big corporations. While they admit, the services are very expensive. I put courier because it is a permissive term for individuals and companies, regardless of their economic situation.

Adamant36:
Hmmm looking over it, this sounds like it would be similar to a mail/package carrier like UPS. Which is currently tagged as amenity=post_office. The Wiki says that tag applies to "A place where letters and parcels may be sent or collected." Cost and who they ship with or to doesn't really matter that much. Like both these places just look like UPS stores. 
unnamed

sguinetti:
Interesting comparison of an American courier company. However, that makes me a little doubtful because UPS is labeled as amenity=post_office. However, the issue of mail delivery and collection is unique to Japan Post. In fact, it says so [on Wikipedia](https://ja.wikipedia.org/wiki/%E9%9B%86%E9%85%8D%E9%83%B5%E4%BE%BF%E5%B1%80).

In my opinion a courier is a particular and distinctive service to the national or state mail. For example, I have seen with Olva Courier that it offers distinctions in Peru. Although reviewing, they made a proposal to use the label amenity=courier, but [they rejected it](https://wiki.openstreetmap.org/wiki/Proposed_features/courier).

sguinetti:
For now and not to be confusing, I change this proposal to amenity=post_office. I hope that in the future they will discuss the proposal of amenity=courier.

Adamant36:
> Interesting comparison of an American courier company. However, that makes me a little doubtful because UPS is labeled as amenity=post_office.

Honestly, I'm not a big fan of UPS being tagged as a post office either. I'd much prefer something like amenity=carrier myself, but the OSM community can be a harsh task master if you go against established tagging conventions. So, that's how it is. 

I'm aware that there regional differences sometimes, but generally tags in OSM are suppose to be globally applicable and based on British English standards. I'm trying to follow conventions a little more after letting amenity=karaoke_box slide, when it is still a proposal and not used much outside of Japan. While I think that was the right choice in the moment, going with a rejected proposal in this case, to respect regional differences, wouldn't be. 

You have to watch out for not creating tagging fragmentation anyway. It's a global map and we should expect tags to be extremely similar no matter where we use, no matter how locals tag things.  Like in America where I live practically every town you go has a different way of using the leisure=park tag and we can't agree on a standard for it. So, there's no way to accurately find parks in OSM here. With brands you have to be more narrow and specific about your tags IMO.

bhousel:
These entries stayed in the `office/post_office.json` file, so in 792a8df I moved them back to the `amenity/post_office.json` file and fixed the keys so they will stay there.

PR #3662, opened 2020-03-30
Add Banque Dupuy de Parseval (French bank)

johanricher:


brands/amenity/bank.json
```diff
@@ -2198,6 +2198,16 @@
       "name": "Banque Atlantique"
     }
   },
+  "amenity/bank|Banque Dupuy de Parseval": {
+    "countryCodes": ["fr"],
+    "tags": {
+      "amenity": "bank",
+      "brand": "Banque Dupuy de Parseval",
+      "brand:wikidata": "Q2882897",
+      "brand:wikipedia": "fr: Banque Dupuy de Parseval",
```
bmillemathias: I think you need to strip the white space after the :

johanricher: Right, sorry about that!

Issue #3599, opened 2020-03-13
Split First State Bank


Adamant36:
They are all located in the United States. 

Q87647395 - Florida
Q87647340 - Illinois
Q87647452 - Mississippi
Q87647479 - Michigan 
Q87647374 - Ohio
Q5453817 - Nebraska
Q87646525 - Nebraska
Q87647065 - Texas

I think that's all of them 🤣

1ec5:
Nothing from Delaware? 😮

Adamant36:
Closed in #3675

PR #3675, opened 2020-04-04
Split First State Bank

darkkrol16:
First State Bank is splitted into multiple entities. This request related to https://github.com/osmlab/name-suggestion-index/issues/3599. 

I am unsure about how to add this ones:
Q5453817 - Nebraska
Q87646525 - Nebraska.
The first one I named "amenity/bank|First State Bank~(Nebraska)"
The second one I named "amenity/bank|First State Bank Nebraska"

If I did this in a wrong way, correct me, please. I'm new here.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -3539,11 +3539,147 @@
       "name": "First National Bank"
     }
   },
-  "amenity/bank|First State Bank": {
+  "amenity/bank|First State Bank Nebraska": {
     "countryCodes": ["us"],
+    "nomatch": [
+      "amenity/bank|First State Bank~(Florida)",
+      "amenity/bank|First State Bank~(Illinois)",
+      "amenity/bank|First State Bank~(Michigan)",
+      "amenity/bank|First State Bank~(Mississippi)",
+      "amenity/bank|First State Bank~(Nebraska)",
+      "amenity/bank|First State Bank~(Ohio)",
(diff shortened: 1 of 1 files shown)


Adamant36:
Hi there. Thanks for the PR. I'm not sure about how to handle the two Nebraska either. Also, I'm not sure if you need the nomatches if they are already disambiguated because they won't come up in searches done in states where they aren't located anyway. I could be wrong on that though. @bhousel, any chance you could look at this?

bhousel:
>The first one I named "amenity/bank|First State Bank~(Nebraska)"
> The second one I named "amenity/bank|First State Bank Nebraska"

This seems reasonable.

> @bhousel, any chance you could look at this?

We should try to format the names to have whatever information mappers need to know the difference. 

Brands like this are the main reason why [I think displaying the logos is important](https://github.com/openstreetmap/iD/issues/7028). 

First State Bank~(Nebraska)
https://www.wikidata.org/wiki/Q5453817
https://en.wikipedia.org/wiki/1st_State_Bank_-_1st.BANK
https://www.1st.bank/

First State Bank Nebraska
https://www.wikidata.org/wiki/Q87646525
https://www.1fsb.bank/

The good news is both of these banks only operate a handful of actual locations, so someone really could just edit them all in OSM quickly. Then the data will be correct anyway, and the value in having them in the NSI is to show the correct logo to mappers.

Adamant36:
Alright. Thanks. I just merged it then.

PR #3682, opened 2020-04-09
Add brands of Casino Group

johanricher:
[All French retail brands of the Casino Group](https://www.wikidata.org/wiki/Q2670967#P355) are now added


Adamant36:
I'm mixed on adding on the Wikipedia article of the company that owns all of these. Especially when the specific brands aren't even really talked about in the article to any detail if at all. It's like using the Wikipedia article of a bank holding company for a bank brand they own or something.  Are there no articles for any of their specific subsidiaries?

johanricher:
Hi thanks @Adamant36 that's actually something I wondered about.

Yes, all those Casino brands don't have unique Wikipedia pages (only Wikidata items). And since this wikipedia page was already added for one brand, I made the decision to add it everywhere missing a wikipedia link. 
Also, the Wikipedia page is not about the Casino Group per se (there's a specific page about it) but the "Brands of the Casino Group".

I can remove that if you think it's not a good idea.

Finally, I've now removed on Wikidata the wrong link which existed between [this item](https://www.wikidata.org/wiki/Q1380537) (for the Géant Casino brand) and the [Wikipedia page](https://fr.wikipedia.org/wiki/Enseignes_du_groupe_Casino) ("Brands of the Casino Group"). This can further clear up the confusion.

Adamant36:
I don't know. It's still pretty ungainly. It's to bad they don't have spin off articles. It seems like they are notable enough to.  That article is total crap. If I spoke French I'd clean it up. My recommendation is to just remove the entries for it since its such a convoluted article and it really doesn't matter anyway, but maybe someone else can chime in.

johanricher:
Done.

I agree that that French article should be deleted from Wikipedia or at least merged with the main "Groupe Casino" article. But I have to say that Wikipedia contribution is a bit outside my zone of comfort, OSM/Wikidata/Github are way easier for me! :laughing:

Adamant36:
> I have to say that Wikipedia contribution is a bit outside my zone of comfort, OSM/Wikidata/Github are way easier for me!

No worries. Same here. I mainly stay away from it myself except to organize articles for the companies Wikiproject sometimes. It's starting to feel pretty outdated. Plus, Wikidata is way easier and less toxic to deal with.

Adamant36:
One last thing, is there a difference between petit casino and le petit casino? It seems like they have stores branded with both signs, that are both convenience and grocery stores for both. Apparently they are confused about what they are doing. 

petitcasino
Petit_Casino

johanricher:
"Petit Casino" (without "Le") is a deprecated brand, [not appearing on the group's website anymore](https://www.groupe-casino.fr/groupe/les-activites-du-groupe-en-france-et-international/pays/france/). But since it's a very large franchise, it's possible some shops have not been upgraded yet?

Do you have some figures regarding the current usage of the brand?

Adamant36:
> Do you have some figures regarding the current usage of the brand?

It looks like there's maybe 5 or 6 of them left from what I could find. There might be more or less though. Their listed on lepetitcasino.com. 

https://magasins.lepetitcasino.com/fr/s/petit-casino-toulon-bazeilles
https://magasins.lepetitcasino.com/fr/s/petit-casino-sas-gaelle-sasu
https://magasins.lepetitcasino.com/fr/s/petit-casino-aix-en-provence-le-luberon
https://magasins.lepetitcasino.com/fr/s/petit-casino-lille-national

Those are the ones I found that are officially referenced. There are a few more that I couldn't find an official reference for and might not be open anymore, but there's no way to tell.

johanricher:
Alright, thanks for the thorough research!
I don't think a separate Wikidata item is necessary: "Petit Casino" is already added as complementary label (https://www.wikidata.org/wiki/Q89029249) which is the correct way to model that.
What should we do on NSI's side?

bhousel:
> I don't think a separate Wikidata item is necessary: "Petit Casino" is already added as complementary label (https://www.wikidata.org/wiki/Q89029249) which is the correct way to model that.
> What should we do on NSI's side?

We handle this in NSI by adding a `matchName` property for the less preferred / obselete name.  I'll merge this, since it looks overall fine, and add the `matchName` in a followup commit.

PR #3683, opened 2020-04-10
Add Côté Nature

johanricher:
[French brand of organic supermarkets.](https://www.wikidata.org/wiki/Q90016283)

### DIFF
--- a/brands/shop/supermarket.json
+++ b/brands/shop/supermarket.json
@@ -981,6 +981,16 @@
       "shop": "supermarket"
     }
   },
+  "shop/supermarket|Côté Nature": {
+    "countryCodes": ["fr"],
+    "tags": {
+      "brand": "Côté Nature",
+      "brand:wikidata": "Q90016283",
+      "name": "Côté Nature",
+      "organic": "only",
+      "shop": "supermarket"
+    }
+  },
   "shop/supermarket|D'Agostino": {
     "countryCodes": ["us"],
     "tags": {


Adamant36:
Hi there. Everything looks fine. On the Wikidata side it would be helpful to add social media links if there are any though just in case the logo gets deleted. They can be pretty picky about them and it's better for iD Editor if there is a backup just in case.

johanricher:
I didn't know social network links could be used in place of logo images, I'll make a note to add them from now on. Do you mean NSI scripts will actually retrieve images from Twitter/Facebook, even if there's no images on Wikidata or Wikimedia Commons?

Adamant36:
> Do you mean NSI scripts will actually retrieve images from Twitter/Facebook, even if there's no images on Wikidata or Wikimedia Commons?

If the brands social media information is in Wikidata NSI will access the logos from Facebook/Twitter through there somehow. I'm not 100% sure exactly what the process is, but I know its on Wikidata's side. It doesnt matter if there is a logo on Wikidata or Wikimedia Commons for it to work. If you browse the brands on the nsi.guide site you can see a lot of the images dont have logos from Wikidata but do from social media.

johanricher:
Thanks, good to know!

A lot of times those brands don't have an online presence though (for example this one is a smallish chain, active only in one French region). The parent company might have one, or a logo, which both could retrieved from Wikidata, but that could also lead to false positives..

Adamant36:
> A lot of times those brands don't have an online presence though (for example this one is a smallish chain, active only in one French region). The parent company might have one, or a logo, which both could retrieved from Wikidata, but that could also lead to false positives..

That can be a problem. That's why it's good to also have a logo in Wikidata as a back up if you can. Just don't rely on it. They mass deleted a bunch of logos for banks I added recently that where totally within the proper copyright standards to have in Wikimedia Commons and they weren't even willing to discuss un-deleting them. So, it should never be the first or only way of accessing logos. 

In cases where the company is small and might have an official social media account, you can always add their "fan" Facebook page or one for a branch that has the same logo. Although, don't do it unless you have to. I don't recommend adding social media links to parent companies. As the logos are usually different. Plus, it's a different entity and might come off wrong to people checking over the Wikidata entries for an excuse to delete them. 

That said, creating Wikidata entries for parent companies and linking them to the Wikidata entries for brands we use is always a cool idea. Even if we don't directly benefit, it makes it way less likely the Wikidata entry for the brand we access will get deleted.

johanricher:
> They mass deleted a bunch of logos for banks I added recently that where totally within the proper copyright standards to have in Wikimedia Commons

I don't know any other way than Wikimedia Commons + Wikidata to make images open and accessible in a structured way

I have to say that I made my first uploads to Commons these last few days, in order to add them in Wikidata. Some of those images where logos that where already uploaded on the French Wikipedia, but tagged with weird copyright notices which prevented them to be transfered to Commons. So I took the libert to upload them separately but made extra sure to tag them with `{{PD-textlogo}}` ("[This logo image consists only of simple geometric shapes or text...](https://commons.wikimedia.org/wiki/Template:PD-textlogo)") like most other logos. There are a lot of French company logos on Commons which have not been removed, I hope mine won't either. :crossed_fingers:

I love Wikidata, probably more than OSM now. Both combined are just plain fucking awesome.

Adamant36:
> I don't know any other way than Wikimedia Commons + Wikidata to make images open and accessible in a structured way.

I'm sure there's other projects out there, but I doubt any of them come anywhere close. 

> I took the liberty to upload them separately but made extra sure to tag them with {{PD-textlogo}}

I did the same thing to and for whatever reason it just didn't help. In some ways people on Wikidata/Wikimedia Commons can be way more finicky then on Wikipedia. I think a lot of is because Wikipedia has a more mature vetting/resolution process. Wikidata is a lot more streamlined and there isn't as much back and forth about things. Which has it's pros and cons. 

> I love Wikidata, probably more than OSM now. Both combined are just plain fucking awesome.

I'm pretty into it myself. Especially in combination with OSM, but even on it's own it is amazing. I'm really surprised it hasn't been picked up more considering how much you can do with it. I'm glad it is at least being used in OSM the way it is though. If I hadn't of gotten involved in the NSI I wouldn't have known about Wikidata even though I edit Wikipedia a lot. I think this project and OSM are both on the forefront of what you can do with Wikidata. All three projects have an amazing amount of potential.

PR #3684, opened 2020-04-10
Add 4 brands of the Francap group

johanricher:
https://www.wikidata.org/wiki/Q17561658#P1830


Adamant36:
Thanks. Everything looks fine. It would help to add social media to the Wikidata entries if there are any since they like to delete logos.

PR #3687, opened 2020-04-12
Add Carré Blanc

bmillemathias:


bhousel:
Can we use the more standard `shop=houseware` instead of introducing a new tag `shop=household_linen`?

bmillemathias:
> Can we use the more standard shop=houseware instead of introducing a new tag shop=household_linen?

I did not check if it was used largely or not before submitting, most of the Carré Blanc shops were tagged as it so I follow the rule, which seems to come from the wiki which as I understand it acts as a  "standard" https://wiki.openstreetmap.org/wiki/Tag:shop%3Dhousehold_linen.
I've look to houseware tag, I don't know if it applies to that. 

I'll ask around.

pyrog:
For what I understand, `household_linen` is not really a subset of `houseware`.
Here in Europe, we have shops specialised in [linens](https://en.wikipedia.org/wiki/Linens#List_of_house_linens) and other specialised in [kitchen utensil](https://en.wikipedia.org/wiki/Kitchen_utensil).

`houseware` | `household_linen`
-- |--
Kitchen utensils | Bathroom linen

PS: The following values of `shop` seems to be **household_linen**, except **linen_weaving** that seem to be a fabric.
 
Value | Count 
-- | --
bedlinen | 1
household linen | 1
**household_linen** | **151**
linen | 28
linen_weaving | 1
linens | 4

bhousel:
> most of the Carré Blanc shops were tagged as it 

Ok, I'll merge as-is since it is what mappers are using, but I really don't like introducing a rarely-used very specific tag for something. I think it's bad for OSM.  These will not render properly.  There won't be a preset in iD anytime soon for them.

> so I follow the rule, which seems to come from the wiki which as I understand it acts as a "standard"

No, the OSM wiki is far from a "standard" - it represents a few people's opinions.  It's not translated into every language and nobody asks data consumers or developers what they want.

pyrog:
> No, the OSM wiki is far from a "standard"

👍
The wiki try to describe what terms/concepts are used in reality "all" over the world.

Something obvious for a french, is sometime not different for a german, american, african, asian 😄

PR #3688, opened 2020-04-12
Another round of clothes brands found in France

bmillemathias:


bhousel:
This is fine and I'm merging it, but please go back and add `"countryCodes": ["fr"],` for any entries that you have introduced that should be scoped to France.

bmillemathias-1a:
> for any entries that you have introduced that should be scoped to France.

Yes, but actually while they are found in France and even French brands, I did not know if they are exclusive to France, so if it's better to scope to France only.

PR #3693, opened 2020-04-13
Added East of Chicago Pizza, updated Gino's Pizza & Spaghetti House

pkoby:
Gino's Pizza & Spaghetti House is consistently worded on all signs and logos that I can find, including the "& Spaghetti House" that was previously omitted from tags.

I added wikidata tags for EOCpizza for Twitter, FB, and website, so logo fetch should work.

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -818,6 +818,18 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|East of Chicago Pizza": {
+    "countryCodes": ["us"],
+    "tags": {
+      "amenity": "fast_food",
+      "brand": "East of Chicago Pizza",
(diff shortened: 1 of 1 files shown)


Adamant36:
It seems like "Gino's Pizza" is tagged a lot more then "Gino's Pizza & Spaghetti House." My recommendation is to go with official_name=Gino's Pizza & Spaghetti House instead of using the longer name in the name field if you want it to be included. Since it's always good to stay with preexisting tagging conventions and to use the shorter name when possible.

pkoby:
Fair point. Changed that (and maybe made some noobie Github mistakes...).

Adamant36:
Still needs to be changed for the brand and amenity tags though.

Adamant36:
Thanks. Looks fine now 👍

PR #3698, opened 2020-04-13
Add Pepper Lunch

sguinetti:
Add Pepper Lunch, japanese franchise with 250+ restaurants. Also, added matchname.

### DIFF
--- a/brands/amenity/restaurant.json
+++ b/brands/amenity/restaurant.json
@@ -1851,6 +1851,17 @@
       "name": "Pei Wei"
     }
   },
+  "amenity/restaurant|Pepper Lunch": {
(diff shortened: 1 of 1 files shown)


bhousel:
Please add a countryCode to scope this entry to Japan.

sguinetti:
I would this action but it is an international brand. The franchise covers 16 countries more. Includes Singapore and the United States. https://www.pepperlunch.com/world/singapore/

bhousel:
> I would this action but it is an international brand. The franchise covers 16 countries more.

ok, sorry!  Just wanted to make sure it wasn't left off :+1:

PR #3708, opened 2020-04-14
Add matchname for FamilyMart

sguinetti:
This pull requests adds matchname for FamilyMart called Famima, it's colloquially called in some establishments.


bhousel:
Looks good, thanks @sguinetti  :+1:

PR #3709, opened 2020-04-14
added BankWest (US)

### LINKED ISSUE #3690: Add BankWest
American bank. Created to disambiguate with (from?) the Australian bank Bankwest. 

[Q90386673](https://www.wikidata.org/wiki/Q90386673)

ferdinand0101:
fixes #3690 
Disambiguation is not necessary cause of the US and AU country code also spelling is slightly different
with Bank**W**est and Bank**w**est

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -2173,6 +2173,15 @@
       "name": "Bankwest"
     }
   },
+   "amenity/bank|BankWest": {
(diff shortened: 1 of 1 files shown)


bhousel:
Disambiguation is still necessary because matching is done case insensitive.  `npm run build` does show a warning about this, so I added the disambiguation in c94949ce

PR #3723, opened 2020-04-17
Add Fuji and Grand Fuji

sguinetti:
This pull request adds two brands of Fuji supermarket chain.


bhousel:
looks good, thanks @sguinetti :+1:

PR #3725, opened 2020-04-17
Added Martinus

RickeyRichards:
Added slovak bookstore brand called Martinus. Hope I did it the right way.


bhousel:
Looks perfect, thanks!

PR #3727, opened 2020-04-17
Added Okay

RickeyRichards:
Network of electronics stores in Czech Republic and Slovakia.


bhousel:
BTW, it there is also a supermarket named Okay in Belgium, so I added `nomatch` properties in caa50e2 to remove the duplicate name warning.

PR #3728, opened 2020-04-17
Add matchname for Daily Yamazaki

sguinetti:
In Japanse Wikipedia and [its website](http://www.daily-yamazaki.jp/information/), Daily Yamazaki is called Dailyshop. This pull request adds to matchname.


bhousel:
Thanks, looks good :+1:

PR #3736, opened 2020-04-18
Added City National Bank (West Virginia)

pkoby:
There are a lot of City National Banks, it turns out, and there were only 2 in the database so far.

I created the Wikidata page for this version of the bank, but there is no Wikipedia page. I did find their FB profile (no Twitter) and added it to Wikidata, so at least logo population should work.

This version of the bank is based in West Virginia, but has locations in Ohio, Kentucky, and Virginia as well. I think the way I worded the identifier should be fine, but feel free to suggest other ideas (I think saying "City National Bank (WV, OH, VA, KY)" might be a bit much).


bhousel:
> I created the Wikidata page for this version of the bank, but there is no Wikipedia page. I did find their FB profile (no Twitter) and added it to Wikidata, so at least logo population should work.

Thanks @pkoby this looks good :+1: 
I updated the wikidata page to add an "instance of" property of "bank".  The more information we add to the wikidata page, the better.

PR #3738, opened 2020-04-18
Add Mega Don Quijote

sguinetti:
This pull request adds Mega Don Quijote supermarket. Sources: [YouTube video](https://www.youtube.com/watch?v=k-9LgZpIbgI) and [official website](https://www.donki.com/mega/).

### DIFF
--- a/brands/shop/supermarket.json
+++ b/brands/shop/supermarket.json
@@ -2636,7 +2636,7 @@
       "name:en": "MEGA Don Quijote",
       "name:ja": "MEGAドン・キホーテ",
       "opening_hours": "24/7",
-      "shop": "general",
+      "shop": "supermarket",
       "short_name": "MEGAドンキ",
       "short_name:en": "MegaDonki",
       "short_name:ja": "MEGAドンキ"
--- a/brands/shop/supermarket.json
+++ b/brands/shop/supermarket.json
@@ -2619,6 +2619,13 @@
   },
   "shop/supermarket|MEGAドン・キホーテ": {
     "countryCodes": ["jp"],
+    "matchNames": [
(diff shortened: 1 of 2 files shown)


bhousel:
It looks like a good addition, but we need to fix the `matchNames`.  That should just contain alternate names.  There is a different property called `matchTags` that contains the alternate tags.

I know it's a bit confusing, but I tried to improve the documentation here:  https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#matchnamesmatchtags

sguinetti:
Ok. I checked that this tag is different from the original Don Quixote variety store. Now I see the implemented matchTags and the tips makes it much better. Thanks for the note.

brands/shop/supermarket.json
```diff
@@ -2617,6 +2617,31 @@
...
+      "brand": "ドン・キホーテ",
+      "brand:en": "Don Quijote",
+      "brand:ja": "ドン・キホーテ",
+      "brand:wikidata": "Q1185381",
+      "brand:wikipedia": "ja:ドン・キホーテ (企業)",
+      "name": "MEGAドン・キホーテ",
+      "name:en": "MEGA Don Quijote",
+      "name:ja": "MEGAドン・キホーテ",
+      "opening_hours": "24/7",
+      "shop": "general",
```
bhousel, line 2639: Should be like this so it matches the `shop/supermarket` above..
```suggestion
      "shop": "supermarket",
```

sguinetti: Fixed

bhousel:
Thanks, I think this looks ok :+1:

PR #3740, opened 2020-04-18
Add pachinko salons (create leisure=adult_gaming_centre)

sguinetti:
This pull request adds pachinko salons brands of Japan (under leisure=adult_gaming_centre with gambling=pachinko). The case [is known](https://www.businessinsider.com/what-is-pachinko-gambling-japan-2018-7), as casinos have only recently [become legal in Japan](https://www.casino.org/news/japan-casinos-expected-to-embrace-slot-machines/). Also, add term filter "pachiko".


bhousel:
Looks good to me, thanks @sguinetti :+1:

PR #3747, opened 2020-04-20
Add matchgroups for adult gaming centre

sguinetti:
I think it is more of a proposal but perhaps it will be useful for recent pachinko saloons brands  (because they don't use cards). The brands taged as adult_gaming_centre are extended to amenity tags like amenity=gambling and amenity=casino. Also, that would avoid some confusion in POIs with slot machines or games that require paying money.


bhousel:
Thanks @sguinetti this seems right :+1:

PR #3757, opened 2020-04-21
[wip] Yesss electrical shops

bmillemathias:
Hello,

Yessss Group has shops (more than 300 to date) in few European countries and in each the name differs a little bit
- fr: Yesss Électrique
- uk: Yesss Electrical
- de: Yesss Elektro
- it: Yesss Electrical

I don't know what is the best way to do that ? Should I create one entry per country, or could nsi provide a way to provide the local name to the consumer (name:fr, name:it, …) ?


bmillemathias:
what the hell did I do...

PR #3758, opened 2020-04-21
Disambiguation of Eurotyre brand

Raman077:
Added some entries in the brands.json file for the brands that are named Eurotyre but based in different European countries. Please let me know in case I need to add something more.


bhousel:
Thanks @Raman077 and welcome :wave: 
Just a few questions..
- I noticed that the `brand:wikidata` values are only added to `dist/brands.json` - this is a generated file, so they shouldn't be added here, they should go in the `shop/tyres.json` file instead.
- Is there a difference between the "Eurotyre~(Netherlands)" and "Euro-Tyre BV" entry?

Raman077:
Hi. Thanks for bringing it to my notice. I will update the pull request. In regards to your query, Eurotyre~(Netherlands) should be actually Eurotyre~(Belgium) which I found out later and only made the changes in the dist/brands.json file. I was not sure which was the primary source the changes should go to.

Raman077:
I removed the entries from `dist/brands.json`. Instead created the entries in `shop/tyres.json`.

bhousel:
Thanks @Raman077 
It looks like there is still a bit of work to do for this, but I think I'll just merge it as-is and cleanup after.

- The wikidata pages don't have the properties set correctly, which puts them at risk of deletion
- It looks like some of these are just for a single shop, not for a chain of stores.  We only really want to track brands with multiple locations.

bhousel:
added some followup commits here:
d962cb72 French EuroTyre has a few locations in Somalia too
bcbd9709 The "German" EuroTyre was apparently an old webpage for EuroTyre BV
63ef9d63 Remove Eurotyre Portugal, it seems to just be 2 warehouses
fb612b0d Euro-Tyre B.V. appears to be a single shop unaffiliated with the others
931c6b3c EuroTyre Q91309121 operates in both Belgium and Netherlands

PR #3766, opened 2020-04-24
Re-edited Gino's to reflect name on the ground, specific to WV.

pkoby:
I looked into this brand some more, because it was bugging me from my last PR. I want to map what's on the ground, and every restaurant of this brand I've seen fully writes out the whole name of Gino's Pizza & Spaghetti House.

A comment from last PR was that many instances of the restaurant only use "Gino's Pizza". Here's the output of that search—there are only three in WV:
image

The search for the full name shows this:
image

Most of the ones with the full name were not added by me (6/9), so there's no personal bias here. (2 of the "Gino's Pizza" in the WV area were mine added with the current NSI tags.)

Here's the location page from their website: http://ginospizza.com/locations.aspx

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -1035,6 +1035,18 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|Gino's Pizza & Spaghetti House~(West Virginia)": {
+    "countryCodes": ["us"],
+    "tags": {
+      "amenity": "fast_food",
+      "brand": "Gino's Pizza & Spaghetti House",
+      "brand:wikidata": "Q5563205",
+      "brand:wikipedia": "en:Gino's Pizza and Spaghetti",
+      "cuisine": "pizza",
+      "name": "Gino's Pizza & Spaghetti House",
+      "takeaway": "yes"
+    }
+  },
   "amenity/fast_food|Gino's Pizza~(Canada)": {
     "countryCodes": ["ca"],
     "tags": {
@@ -1043,19 +1055,6 @@
       "brand:wikidata": "Q84029134",
       "cuisine": "pizza",
       "name": "Gino's Pizza",
-      "takeaway": "yes"
-    }
-  },
-  "amenity/fast_food|Gino's Pizza~(West Virginia)": {
-    "countryCodes": ["us"],
-    "tags": {
-      "amenity": "fast_food",
-      "brand": "Gino's Pizza",
-      "brand:wikidata": "Q5563205",
-      "brand:wikipedia": "en:Gino's Pizza and Spaghetti",
-      "cuisine": "pizza",
-      "name": "Gino's Pizza",
-      "official_name": "Gino's Pizza & Spaghetti House",
       "takeaway": "yes"
     }
   },


ArenaL5:
I can't speak for the USA or West Virginia's community so I'd rather ask @Adamant36's opinion again. 

Just to clarify: all the Gino's Pizza _outside of_ West Virginia are from a different franchise? Or is evey Gino's Pizza in the United States part of Gino's Pizza & Spaghetti House?

@Adamant36: @pkoby's last PR is #3693.

pkoby:
I'd appreciate another US opinion, but as far as I can tell from a quick web search, a lot of "Gino's Pizza"s in the US are either small franchises in separate regions, or individual restaurants. Gino's Pizzas outside of WV are (mostly) not "Gino's Pizza & Spaghetti House", which is just Western WV and a couple in Southern Ohio.

Adamant36:
I'm mostly leaning toward what pkoby said. Although there does seem to be another Ontario franchise that's just calles Gino's Pizza. https://ginospizza.ca/locations/. Unless the other Canadian franchise in OSM is a miss spelling or something.

pkoby:
I found https://www.ginospizzakingston.com/, which only shows one location, but does seem to match. Note this is not Gino's Pizza & Spaghetti _House_.

Adamant36:
Just to make it even more convoluted there seems to be https://www.facebook.com/GinosPizzaSpaghetti/ also in the same area. I'm pretty sure they are different from the other ones and it appears they have at least a few restaurants.

ArenaL5:
This is messed up.

There's not much we can do from here to fix this. Make the wikidata entries as different as possible, adding the ["different from"](https://www.wikidata.org/wiki/Property:P1889). Add Facebook and Twitter links to all of them so they'll fetch the logo. Improving the countryCodes system wouldn't help as both Canadian franchises are based in Ontario.

I'm seeing three franchises here:
- Gino's Pizza, based in Ontario, Canada, wikidata reference [Q84029134](https://www.wikidata.org/wiki/Q84029134)
- [Gino's Pizza & Spaghetti](https://www.orderginos.ca/), also based in Ontario, needs a wikidata entry yet (the Facebook page @Adamant36 is for only one of the restaurants from that franchise)
- Gino's Pizza & Spaghetti House, at West Virginia, United States, wikidata reference [Q5563205](https://www.wikidata.org/wiki/Q5563205) (whose Wikipedia article should be filled in and renamed to the full name)

Data in the map must be corrected by hand no matter what, and independent stores not owned by these franchises can be tagged with `not:brand:wikidata` as described [here](https://wiki.openstreetmap.org/wiki/Key:not#Software_support).

So I find this PR alright, but we need an extra entry when all is done.

Adamant36:
Totally agree. How does not:brand:wikidata work when there's multiple brands its not though since we can only use one Wikidata entry with it? What stops someone from not just wrongly tagging them as one of the other franchises, especially in Canada where there are two in the same area?

ArenaL5:
I worried about that. I tested it... and the iD editor behaves in a confusing way. I did these tests with Game and GameStop, in the United Kingdom:

- Creating a new point and typing "Game" will make iD suggest both
- If I remove `brand:wikidata`, click outside, and click the point back, it will raise a warning advising me to "complete my tags" (to put `brand:wikidata` back). This only works if the name is a case-insensitive exact match: `name=Game` suggests Game, `name=gamestop` suggests GameStop, and `name=Gamesto` will not raise a warning.
- If I create a point for a Game shop and change `brand:wikidata` to `not:brand:wikidata`, I won't receive any warning.
- If I add both `brand:wikidata` and `not:brand:wikidata`, no warning or error is issued. This is an error.
- Writing several refs at brand:wikidata, separated with a semicolon, _does not work_. iD will ignore the `not:brand:wikidata` and offer me to add a `brand:wikidata` nonetheless.

A quick search leads me to think that these lines should be updated: 
https://github.com/openstreetmap/iD/blob/develop/modules/validations/outdated_tags.js#L135-L149
as line 136 only checks for exact matches.

I've also done some tests with Metro (Ontario) and Metro (Quebec) in Canada, but it's not the same scenario; both of those have the exact same name and Wikidata (why are they separate entries??) and iD never raises a warning if there is more than one candidate (it considers it an "ambiguous match").

For the moment, I'd just add to `not:brand:wikidata` the one reference that matches exactly the full name: if there's an independent Gino's Pizza, just add `not:brand:wikidata=Q84029134` and don't worry about Q5563205 or the new one. It's still possible to mistakenly tag new points with the wrong franchise, but iD won't raise further suggestions to points that already have a `brand:wikidata` or `not:brand:wikidata`, so once they're fixed, they'll remain fixed.

bhousel:
> So I find this PR alright, but we need an extra entry when all is done.

Thanks for checking, everyone!  Rather than let this go more than 2 weeks old, I'm going to accept this as-is, and we can add more "Gino's Pizza" brands as we discover them.  It is probably a very common generic name for a pizza place.

PR #3779, opened 2020-04-29
Add Megane Drug

sguinetti:
This pull request adds a optician shop brand, fix "brand:en" in some brands with similar "Megane" denominations.


tas50:
Thanks @sguinetti

PR #3780, opened 2020-04-29
Add Kozosushi, add official names for 2 fast food chains

sguinetti:
This pull request adds Kozosushi japanese brand and adds two official names for fast food chains.


tas50:
Thanks @sguinetti

PR #3782, opened 2020-04-30
Add Nippon Travel Agency

sguinetti:
This pull request adds NTA japanese brand.


tas50:
Thanks for continuing to fill out the brands in Japan @sguinetti

PR #3783, opened 2020-04-30
Added several Dutch amenities

RobinLinde:
Also split Pathé and added extra countrycode to Tango


tas50:
Looks good. Thanks @RobinLinde

PR #3791, opened 2020-05-01
Add Kitchen Origin and Renge Shokudo

sguinetti:
This pull request adds [other brands](https://pkg.navitime.co.jp/toshu/) from Origin Toshu.

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -3877,6 +3877,21 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|キッチンオリジン": {
+    "countryCodes": ["jp"],
+    "tags": {
+      "amenity": "fast_food",
+      "brand": "キッチンオリジン",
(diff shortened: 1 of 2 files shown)


ArenaL5:
Your commit is alright, but these barely have any presence on Wikidata... I'd suggest to add more data to [ensure they're not deleted](https://www.wikidata.org/wiki/Wikidata:Notability). Creating a Wikipedia article is another option, but that takes much more time.

By the way, thank you for your contributions. I merged your other two PRs.

sguinetti:
Yes. But the point is in Japanese Wikipedia, those articles redirect to the company. So, the idea of creating these entries is to separate the logo from the parent product. I have added so many details and social networks.

ArenaL5:
We can actually add redirects now? That's great news. I had problems with that in the past.

If Wikidata is okay with that, I've no problem with this. Merging now.

sguinetti:
Thank for your reply.

Issue #3764, opened 2020-04-23
Add french beauty shop Carlance


pyrog:
`brand=Carlance`
`brand:wikidata=`[Q91601625](https://www.wikidata.org/wiki/Q91601625)
`shop=beauty`

Currently 23 / 70 in France https://overpass-turbo.eu/s/Td3
website: http://www.carlance.fr

MohamedBechir:
@pyrog i am new to contributing and i want to start with solving this issue can you guide me through it ? Thank you

mds08011:
@MohamedBechir please review [CONTRIBUTING.md](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#convenience_store--add-missing-brands) and let us know if you have specific questions.

RishabhBajaj97:
hey ! 
I have added this on my local repository could you please tell me how to generate pull request and submit this code
Thank you

UKChris-osm:
Hi @RishabhBajaj97,

Thanks for your PR #3810, unfortunately there is a spelling mistake in the "name" value, where "Carl**e**nce" with an "**e**" is used in the middle, compared to the rest of the entry, which uses "Carl**a**nce", were an "**a**" is used in the middle.

UKChris-osm:
Closed with merging of #3810, thank you @RishabhBajaj97

PR #3798, opened 2020-05-02
#3764 add french beauty shop carlance

alpe88:
Hello! First time doing a pull request on github. I fixed some of the warnings in addition to adding the entry that was requested.


Adamant36:
Hi there. Thanks for the PR. I'm not sure about the beauty tags. Plus it looks like you also added an entry for Krëfel and changed some other stuff. Plus some entries were removed. Just for future reference, it's always better to stick to only what the PR is for so it is easier to review and fix any mistakes you might make.

alpe88:
Yeah, that makes sense. Would you prefer I split them up?

As an aside, with the beauty tags, and the other warning cleanup - I googled each one and read their sites before making changes.

brands/amenity/parking.json
```diff
@@ -9,7 +9,11 @@
       "name": "Diamond Parking"
     }
   },
-  "amenity/parking|Indigo": {
+  "amenity/parking|Indigo~(France)": {
```
tas50, line 12: You don't want to add ~(France) unless there is more than one Indigo parking brand to add. This should just be Indigo with the country code of FR. The same goes for the no_match. That won't match since there isn't actually a brand with that entry.

alpe88: Fair, and maybe that entry for Indigo is a different Indigo, but if you go here: https://www.group-indigo.com/en/group/ and scroll to the `SHAREHOLDERS` section there is a map of worldwide locations where this Indigo Group has parking garages etc...but yeah maybe it's not the same Indigo.

bhousel:
Since this has gone stale, I'm going to close it, sorry @alpe88 .. 
Can you submit a new pull request with only the brand additions, and any stuff that you're 100% sure on?  

We don't want to silence warnings just for the sake of it, and we also don't adding tags outside OSM mainstream usage.

PR #3804, opened 2020-05-02
Add Reraku and RELXLE

sguinetti:
This pull request adds two brands of massage shops.


ArenaL5:
I wonder how did they get RELXLE from "Riracle". They look nothing alike, I can't even tell RELXLE's pronunciation.

Merging now.

PR #3806, opened 2020-05-02
Edit AOKI with matchname

sguinetti:
This pull request helps to identify the AOKI writting in Hiragana and other charaters.

### DIFF
--- a/brands/shop/clothes.json
+++ b/brands/shop/clothes.json
@@ -21,6 +21,7 @@
   },
   "shop/clothes|AOKI": {
     "countryCodes": ["jp"],
+    "matchNames": ["青木"],
     "tags": {
       "brand": "AOKI",
       "brand:wikidata": "Q11189480",


ArenaL5:
The kanji in matchNames _may_ be useful, if there are people that write that name in kanji. Judging from the Japanese article, that store name is only ever written in in uppercase letters or katakana, never kanji. (I wouldn't really know. There are some links including those two kanjis, but I can't read Japanese; I've only ever learnt kana and a couple words.)

The kana spelling is useful as a `matchName`, but not as a `name:` tag. [OpenStreetMap generally doesn't want systematic transliterations](https://wiki.openstreetmap.org/wiki/Names#Avoid_transliteration); switching from katakana to hiragana is even more predictable than that, the kind of thing that computers should automatically do. Adding it to OSM itself only pollutes their database for no benefit.

By the way, maybe we should report a bug in the iD editor, could you test it too? When editing the map with the online iD editor, I can get suggestions for McDonald's by typing `mcdonald`, but I can't get しまむら by typing シマムラ.

sguinetti:
Personally I am somewhat confused. In the Japanese version of OSM wiki there are documentation about [ja-Hira](https://wiki.openstreetmap.org/wiki/JA:Key:name:ja) and ja-Latn (with tens of thousands of uses). Although the case of transliterations (what you says) applies when the name is also named in other language (AOKI is more called "アオキ" what "あおき". (I review again about ja-kana is now is [obsolete tag](https://wiki.openstreetmap.org/wiki/JA:Key:name:ja_kana)).

For now I'm going to remove the ja-Hira tag. Instead, we would be better to notify the ID developers.

ArenaL5:
It also says `status: undefined`, if Google Translate didn't lie to me...

That's the downside to this massive project. I can never be sure about which tags are deprecated unless I'm a very active contributor there, and even deprecated tags live on for years in thousands of points... still, that only looks useful if the name uses unusual kanjis or unusual readings; this is already in katakana.

So okay, we'll do it this way. Merging your latest commit https://github.com/osmlab/name-suggestion-index/pull/3806/commits/a86bf7d176e2876fa9f5892aae45b94b9d958906 now; I'll file an issue at iD's repository linking to this PR.

sguinetti:
As for the characters, I find that technical detail interesting (し ま む ら as シ マ ム ラ). This matches with # 3804, when the company had to change the brand to RELXLE to avoid comparisons between Hira and Kata of japanese word "Relax".

ArenaL5:
So that's the reason? They couldn't have simliar names in kana or Latin letters?

It's oddly amusing to see how copyright and trademarks work in other countries. It's much of the same everywhere.

PR #3809, opened 2020-05-02
Add 100 House Lemon

sguinetti:
This pull request adds a 100yen shop from Shizuoka.

### DIFF
--- a/brands/shop/variety_store.json
+++ b/brands/shop/variety_store.json
@@ -1,4 +1,18 @@
 {
+  "shop/variety_store|100えんハウス レモン": {
+    "countryCodes": ["jp"],
+    "tags": {
+      "brand": "100えんハウス レモン",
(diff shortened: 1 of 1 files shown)


ArenaL5:
Please check your `name` and `name:ja` tags. I think you overlooked something while copypasting.

Also, the `brand:wikipedia` tag doesn't match the info on Wikidata.

(I suggest you squash everything into a single commit when you're done.)

sguinetti:
Fixed. But the name company is different to brand name. Also, in Wikidata if one company operates one brand, so there would be no point in creating another (because it would be a duplicate with the same SNS, for example).

ArenaL5:
Okay. Having the Wikipedia page named after the company doesn't look like a problem for me. If they ever create several brands, we can always update this.

Merging.

PR #3822, opened 2020-05-04
Add Johshuya

sguinetti:
This pull request adds a outdoor shop brand.


tas50:
Is there a twitter or facebook that can be linked to the wikidata to give us an icon in iD?

sguinetti:
No, but the stores have their own SNS (they are not relatively official). Instead, I put the logo on Wikidata.

PR #3826, opened 2020-05-04
Update Fielmann in optician.json

xerus2000:
Fielmann is a German chain, so it should link to German Wikipedia

### DIFF
--- a/brands/shop/optician.json
+++ b/brands/shop/optician.json
@@ -106,9 +106,10 @@
     "tags": {
       "brand": "Fielmann",
       "brand:wikidata": "Q457822",
-      "brand:wikipedia": "en:Fielmann",
+      "brand:wikipedia": "de:Fielmann",
       "name": "Fielmann",
-      "shop": "optician"
+      "shop": "optician",
+      "website": "https://www.fielmann.de/"
     }
   },
   "shop/optician|GMO": {
--- a/brands/shop/optician.json
+++ b/brands/shop/optician.json
@@ -103,13 +103,13 @@
     }
   },
   "shop/optician|Fielmann": {
+    "countryCodes": ["de", "ch", "at", "it", "pl", "nl", "lu", "lt", "lv", "by", "ua", "ru"],
     "tags": {
       "brand": "Fielmann",
       "brand:wikidata": "Q457822",
       "brand:wikipedia": "de:Fielmann",
       "name": "Fielmann",
-      "shop": "optician",
-      "website": "https://www.fielmann.de/"
+      "shop": "optician"
     }
   },
   "shop/optician|GMO": {


Adamant36:
It would probably be helpful to add the country codes also while you are at it. Since it seems to be located in only a few European countries.

xerus2000:
add country codes where?

UKChris-osm:
You can add **"countryCodes": ["??"],** between the **key/value|name** and **tags** bit of the code, for example:

"shop/optician|Fielmann": {
"tags": {

... would become:

"shop/optician|Fielmann": {
**"countryCodes": ["at","by","de","lt"],**
"tags": {

... although more / less Country codes might be more accurate when it comes to Fielmann.

But you would need to remove the website value as this only applies to the German market.

xerus2000:
what's the point though? If Fielmann expands to a new country all previous entries have to be adjusted, no? 
And what is that key even for? Couldn't find anything in the wiki: https://wiki.openstreetmap.org/w/index.php?search=key%3AcountryCodes

xerus2000:
As for the website, it could be changed to https://www.fielmann.com but that currently seems to redirect to de anyways.

UKChris-osm:
The key isn't a specific OSM key, it's part of the Name Sugestion Index (NSI) as a way of limiting suggestions to Countries that the name is currently applicable to, so as to prevent accidental usage of the name / brand somewhere it doesn't currently operate.

You're quite right that Fielmann could expand to new areas, and in this case the new Country(s) could be added to the NSI (although the previous entries wouldn't need to be adjusted), but for now it helps to limit it becoming confused with other possible places that use the same name in other Countries.

Adamant36:
The website should be removed either way. We don't really them to entries. Just like wedon't include any other contact information type stuff. Its easy easy to get other ways and vary/change by location to easily (not even country to country).

xerus2000:
> The website should be removed either way. We don't really them to entries. Just like wedon't include any other contact information type stuff. Its easy easy to get other ways and vary/change by location to easily (not even country to country).

But the website is the main website of the brand, why would that vary? I don't fancy adding that to each entry by hand, and I think that info should be there. I want OSM to be as convenient as Google Maps (which I currently want to switch away from) and a link to the company website is clearly something I expect.

xerus2000:
As for countryCodes, that does make sense. I'd go after the list here on the left sidebar then, I guess? https://www.fielmann.eu/

Though they don't seem to be in russia yet...

Adamant36:
> But the website is the main website of the brand, why would that vary?

It being the "main site" doesn't really matter. The important is thing is it being the "correct site" and there isn't one Because it's not the site people from other countries in different languages use (for instance https://www.fielmann.lu/fr/, https://www.fielmann.nl/, etc etc). Plus it's not the "correct site" for individual locations, which people like to add, I.E. https://www.fielmann.nl/vestigingen/emmen/7811-dp-picassopassage-74/ etc etc. 

> I don't fancy adding that to each entry by hand, and I think that info should be there.

As much as you don't fancy adding it by hand, people who want the "correct site" for them probably won't fancy having to remove the "main site" to add it and we shouldn't intentionally add bad information anyway. Well I agree the information should be there, it should be the correct information if it's going to be in OSM and https://www.fielmann.eu/ isn't the correct information for everyone. 

Just like we wouldn't automatically add drive_through=yes to all (or any) of the fast food entries because some (or most) locations might have a drive through or because it might make things "simpler" for certain mappers. Even something as seemingly clear cut as takeaway=yes was only added as a default after some debate and is still sorta iffy. I'd put a website in the same class as drive_through=yes. Maybe it's correct for some and make it easier for them, but it's not for everyone and makes it harder on people who it isn't correct for. Plus, it introduces bad data. Especially considering technically in all situations the website should be for the specific stores. Not the corporations website.

xerus2000:
ah okay, that does make sense, thanks - I didn't think of location-specific subpages :)

xerus2000:
it's adjusted

UKChris-osm:
You never know, "_brand:website_"could become a thing down the road, where the general branding web site is a default setting for any entry that doesn't have a more localised version available.

xerus2000:
brand:website is already a thing: https://taginfo.openstreetmap.org/keys/brand%3Awebsite#values
Could we add that then?

Adamant36:
I'm going with no. Although it could be good idea at some point, it quit there yet IMO. It doesnt even have a wiki article explaining its use or any discussion anywhere that I could find. Plus, as far as I see it there's still issues with what exactly would constitute the brands website anway. Plus, they still aren't stable enough for a "set it and forgot" project like this and can easily be found other ways. In other words, just adding the word "brand" to the tag doesnt negate the multiple and varied reasons we dont use websites in the index in the first place.

Adamant36:
BTW, any time you think we should support something new like that feel free to open a new issue about it. A merged PR isn't really the best place to discuss implementing something new like that.

PR #3827, opened 2020-05-04
Add Gourmet City

sguinetti:
This pull request adds a subrand by Daiei.

### DIFF
--- a/brands/shop/supermarket.json
+++ b/brands/shop/supermarket.json
@@ -5560,6 +5560,20 @@
       "shop": "supermarket"
     }
   },
+  "shop/supermarket|グルメシティ": {
+    "countryCodes": ["jp"],
+    "tags": {
+      "brand": "グルメシティ",
+      "brand:en": "Gourmet City",
+      "brand:ja": "グルメシティ",
+      "brand:wikidata": "Q11300647",
+      "brand:wikipedia": "ja:グルメシティ",
+      "name": "グルメシティ",
+      "name:en": "Gourmet City",
+      "name:ja": "グルメシティ",
+      "shop": "supermarket"
+    }
+  },
   "shop/supermarket|コープ": {
     "countryCodes": ["jp"],
     "tags": {


tas50:
@sguinetti Is there a twitter of facebook account you can link in the wikidata so we get that icons in iD?

sguinetti:
Well, the SNS would be from Daiei.  There isn't other social network than that company. I just put the logo on Wikidata instead.

Adamant36:
> I just put the logo on Wikidata instead.

How has your success with that been so far? It seems like all the logos I upload get deleted.

sguinetti:
If a logo lacks artistic intent, such as plain text with basic shapes and straight lines are in the public domain. This is the requirement to upload content to Commons:
> In Japanese copyright law, a work is considered eligible for protection when it is "a production in which thoughts or sentiments are expressed in a creative way and which falls within the literary, scientific, artistic or musical domain." ([Wikipedia](https://en.wikipedia.org/wiki/Threshold_of_originality#Japan))

ArenaL5:
Shouldn't the logo be for Gourmet City, even if the social network account is for all Daiei?

(On a separate note, could you double-check this PR? d962ce975ee7113f8a7f77bcf09954d41d3a5300 rewrites this with a different store, and the next commit rewrites it back to Gourmet City.)

sguinetti:
Strange. I think that I will have wrongly merged the commits. I don't know if this is the only pull request with this problem. Should I update to the last commit to review if something weird happens?

sguinetti:
I close this pull request. Now I rework in #3865.

ArenaL5:
The other PRs looked fine to me. It was only a glitch this one.

#3865 is good, I'm merging it.

(I noticed the logo is the same, but I was asking why the text at https://www.wikidata.org/wiki/Q11300647#P154 reads "daiei" instead of "gourmet city"). Anyways, it's not that important, and I can merge it anyways.)

Issue #3761, opened 2020-04-21
Add french pharmacy Anton et Willem


pyrog:
`brand=Anton et Willem`
`brand:wikidata=`[Q91330267](https://www.wikidata.org/wiki/Q91330267)
`amenity=pharmacy`
`healthcare=pharmacy`
`dispensing=yes`

Curently 4 / [~40](https://www.anton-et-willem.fr/pharmacies) in France : https://overpass-turbo.eu/s/T8l
website: https://www.anton-et-willem.fr

See #3756 for issue with names

chankailong:
I want to help this issue.

It is my first time to contribute to the Github.

After reading the instruction, should I follow the "Adding missing brands"?

Also, I don't get the meaning of #3756. So what the conclusion is for the "name"?

Thank you!

Adamant36:
> I don't get the meaning of #3756. So what the conclusion is for the "name"?

Hi there. Unfortunately, I didn't follow along with the conversation in #3756. So I don't know what it's about. @pyrog would though, or another option would be to do a PR for something else. It doesn't really matter. Whatever you feel like doing 👍

UKChris-osm:
Merged via #3832

PR #3832, opened 2020-05-07
Add french pharmacy Anton et Willem

kirandash:
Added Anton et Willem Pharmacy in France, to address issue #3761


UKChris-osm:
I think **"name/brand": "Anton _&_ Willem"** with an ampersand (**&**) would be better in this instance, as that's what it seems to say on store fronts and the logo.

kirandash:
@UKChris-osm Modified the "name" and "brand" params from "Anton et Willem" to "Anton & Willem"

Issue #3750, opened 2020-04-20
FR: Add Vapostore (e-cigarette)


gendy54:
Hello,

country code: FR
shop=e-cigarette
name=Vapostore
brand=Vapostore
brand:wikidata=Q91257968
network=Vapostore

pyrog:
Does `network` key is useful ?

gendy54:
It's a franchised business. To see if it's useful.

pyrog:
@bhousel in https://github.com/osmlab/name-suggestion-index/issues/3756#issuecomment-617181638
> We don't set network keys now for pharmacies (or anything else yet)

pyrog:
Existing tags are `shop=tobacco`, `shop=chemist`, `shop=yes`
Currently 32 / [98](https://www.vapostore.com/magasins-cigarette-electronique) in France: https://overpass-turbo.eu/s/Td9

UKChris-osm:
Merged via #3834

PR #3834, opened 2020-05-07
FR: Add Vapostore (e-cigarette)

kirandash:
Added "Vapstore", e-cigarette shop in France to address issue #3750


UKChris-osm:
Thank you @kirandash

Issue #3741, opened 2020-04-19
Add french fast-food Coeur de Blé


pyrog:
`name=Coeur de Blé` or `name=Cœur de Blé`
`amenity=fast_food`
`cuisine=sandwich`

Could be tagged as `shop=bakery` or `amenity=restaurant`

Currently 12 / ? in France https://overpass-turbo.eu/s/T1K
website: http://restaurants.casino-restauration.fr
wikidata:

BARICISSE:
is this still open ? I am available to work on it.

PR #3835, opened 2020-05-07
Add french fast-food Coeur de Blé

kirandash:
Added French restaurant Coeur de Blé to address issue #3741

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -621,6 +621,17 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|Coeur de Blé": {
+    "countryCodes": ["fr"],
+    "matchNames": ["Cœur de Blé"],
+    "tags": {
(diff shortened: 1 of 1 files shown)


ArenaL5:
I cannot merge this.

`"brand:wikipedia": "en:Coeur de Blé (restaurant)"` means that there is a WIkipedia article at https://en.wikipedia.org/wiki/Coeur_de_Blé_(restaurant)

That article doesn't exist, and I couldn't find any entry for Coeur/Cœur de Blé at English WIkipedia, or French Wikipedia, or Wikidata. Of these, Wikidata is the most important: it disambiguates between businesses with the same name, and without that certainty, tools like the iD map editor won't work.

Without Wikipedia or Wikidata links, this addition is hardly useful. However, if you can create a Wikidata page (with links and references) and tag it with `"brand:wikidata"`, then I'll have no problem accepting it.

PR #3846, opened 2020-05-08
Add SHOE･PLAZA

sguinetti:
This pull request adds a shoe shop brand.

### DIFF
--- a/brands/shop/shoes.json
+++ b/brands/shop/shoes.json
@@ -1066,6 +1066,19 @@
       "shop": "shoes"
     }
   },
+  "shop/shoes|シュープラザ": {
+    "countryCodes": ["jp"],
+    "tags": {
+      "brand": "シュープラザ",
+      "brand:en": "SHOE PLAZA",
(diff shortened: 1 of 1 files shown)


ArenaL5:
I know this is a _really_ pedantic question, but this is a worldwide database, so I'd rather be cautious.

Do Japanese users really prefer the name to be rendered as SHOE･PLAZA (all capitals, with interpunct) or is that only a stylistic choice for the logo? Searching for it in internet shows me they're not consistent with that.

Maybe SHOE PLAZA (no interpunct) or Shoe Plaza (title case) would be better?

sguinetti:
Honestly, the character "･" is more writed than the spaces in Japanese sentences, so it is common in that country. Anyway, I will opt for only capital letters without the middle point to make it international (but "Shoe Plaza", no all caps, is a generic name and used in other countries).

ArenaL5:
I'm okay with that. Merging now.

By the way, thank you for helping so much with this database. :smile:

PR #3860, opened 2020-05-11
Add Tambunting (shop=pawnbroker)

govvin:
Add an entry for the Tambunting chain of pawnshops in the Philippines.

### DIFF
--- a/brands/shop/pawnbroker.json
+++ b/brands/shop/pawnbroker.json
@@ -54,6 +54,15 @@
       "shop": "pawnbroker"
     }
   },
+  "shop/pawnbroker|Tambunting": {
+    "countryCodes": ["ph"],
+    "tags": {
+      "brand": "Tambunting",
+      "brand:wikidata": "Q93928644",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Hi @govvin,

Did you create the Wikidata article for this PR?

I only ask as it seems one already exists [Q7681088](https://www.wikidata.org/wiki/Q7681088) along with a Wikipedia article [Tambunting](https://en.wikipedia.org/wiki/Tambunting)

govvin:
My apologies, I overlooked that. I will revise the PR to point to the correct item. The new item has been requested for deletion, and i'll update the original with additional details. 

Thanks for the heads-up!

> Hi @govvin,
> 
> Did you create the Wikidata article for this PR?
> 
> I only ask as it seems one already exists [Q7681088](https://www.wikidata.org/wiki/Q7681088) along with a Wikipedia article [Tambunting](https://en.wikipedia.org/wiki/Tambunting)

PR #3864, opened 2020-05-12
Add supermarket Carrefour Express

Crashillo:


ImreSamu:
interesting ...

in PL - there are - multiple express type : 

### "Carrefour express convenience"
* https://serwiskorporacyjny.carrefour.pl/en/franchise/carrefour-express-convenience  (  up to 250 m2  )

###  "Carrefour express minimarket"
* https://serwiskorporacyjny.carrefour.pl/en/franchise/carrefour-express-minimarket (  from 100 to 500 m2 )

Crashillo:
But do they have its own logo? In France & Spain it looks like:

imagen

About the tagging.... perhaps `convenience` would fit better, but still I have my doubts, it's uncommon these shops be smaller than 200m2.

If you provide the according tags for Poland, I can include them also with country_code=pl

ImreSamu:
> If you provide the according tags for Poland, I

I don't have a local knowledge ..  
my information is the result of the random internet search ..

imho:  checking global NSI keys is hard .. need to double check every country  :(

Crashillo:
What are you suggest? To include the countryCode also? It seems to be present at least in 11 countries.
According its web, Express is a convenience store. I'll rename then.

Crashillo:
:facepalm: Uhhh I didn't see is in another category. Sorry about the noise

PR #3868, opened 2020-05-14
Add Chiyoda Sushi

sguinetti:
This pull request adds a fresh takeover food. Unlike Kyotaru, I think that more deli that fast food but this brand sells sushi preparated and packed only. So I added a matchtag.

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -3795,6 +3795,23 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|ちよだ鮨": {
+    "countryCodes": ["jp"],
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @sguinetti this seems fine :+1: 
I have noticed on https://nsi.guide that the Japanese community does seem to use `shop=deli` for takeaway/prepared sushi places (e.g. Hotto Motto), so maybe that tag is ok.  

It would be great if someone local can provide guidance, and update the [OSM wiki page for `shop=deli`](https://wiki.openstreetmap.org/wiki/Tag:shop%3Ddeli) if that's how they want to use the tag.  It's cool with me, but this tag would probably surprise a lot of international OSMers.

PR #3870, opened 2020-05-15
Update bank.json - modify entry for BPI (Global)

govvin:
Update BPI~(Global), and create separate entry for BPI Family Savings Bank

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -937,7 +937,7 @@
       "amenity": "bank",
       "brand": "BPI Family Savings Bank",
       "brand:wikidata": "Q94480105",
-      "name": "BPI Family Savings Bank",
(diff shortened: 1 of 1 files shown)


brands/amenity/bank.json
```diff
@@ -932,6 +931,15 @@
       "official_name": "Bank of the Philippine Islands"
     }
   },
+  "amenity/bank|BPI Family Savings Bank": {
+    "countryCodes": ["ph"],
+    "tags": {
+      "amenity": "bank",
+      "brand": "BPI Family Savings Bank",
+      "brand:wikidata": "Q94480105",
+      "name": "BPI Family Savings Bank",
```
ArenaL5, line 940: The only thing that's causing your PR to fail is this comma.

```suggestion
      "name": "BPI Family Savings Bank"
```

Issue #3437, opened 2019-12-20
Timpson


doublah:
https://en.wikipedia.org/wiki/Timpson_(retailer)
https://www.wikidata.org/wiki/Q7807658

[overpass-turbo](https://overpass-turbo.eu/?w=%22name%22%3D%22Timpson%22+global&R)

It's a generalistic craft shop, and existing tagging is very inconsistent with `craft=key_cutter`, `shop=shoe_repair`, `shop=locksmith`, `shop=shoes`, `craft=shoemaker`, `craft=watchmaker`, `shop=dry_cleaning` and more

Most of these tags aren't necessarily wrong, as most Timpson shops offer these services, it's just not descriptive of all they do, maybe just `craft=yes` would be best because it's so generalistic.

bhousel:
> Most of these tags aren't necessarily wrong, as most Timpson shops offer these services, it's just not descriptive of all they do, maybe just `craft=yes` would be best because it's so generalistic.

Hmm, I'm really not comfortable giving it the tag `craft=yes`, but I'll ask for some suggestions on Slack.

SK53:
Most are a shoe repair with a key cutting service. A combination of those two would be fine (shoe_repair;key_cutting). I don't think they are craft specialists either as locksmiths or as cobblers (I'd never trust my good shoes to them). All the other services (dry cleaning, watch batteries etc) are ancillary to the two main activities. They certainly don't sell shoes or make shoes so those items are just wrong.

carlinmack:
they have now branched into phone repairs in the past five years

jnicho02:
Timpson are primarily a 'cobblers', i.e. a shoe repair shop, that have added additional services, the biggest of which is key cutting. They really aren't a hardware store, or a craft store. I concur with SK53.

(opinion from the UK)

doublah:
> A combination of those two would be fine (shoe_repair;key_cutting).

The problem is a tag like this wouldn't really be supported (see [wiki page](https://wiki.openstreetmap.org/wiki/Semi-colon_value_separator)).

jnicho02:
So, maybe we need a secondary tag...shop=shoe_repair + services=key_cutting ?

SK53:
Th

> A combination of those two would be fine (shoe_repair;key_cutting).
> 
> The problem is a tag like this wouldn't really be supported (see wiki page).

[...]

PR #3872, opened 2020-05-15
Add Mister Minit and PLUSONE

sguinetti:
This pull request adds a international brand of repair of keys in Germany. Also, this is popular in Japan with more of 300 repair centers in this country.

### DIFF
--- a/brands/shop/repair.json
+++ b/brands/shop/repair.json
@@ -12,5 +12,18 @@
       "name": "Mister Minit",
       "shop": "repair"
     }
+  },
+  "shop/repair|Plusone": {
+    "countryCodes": ["jp"],
+    "tags": {
+      "brand": "Plusone",
+      "brand:en": "Plusone",
+      "brand:ja": "プラスワン",
+      "brand:wikidata": "Q94570391",
+      "name": "Plusone",
+      "name:en": "Plusone",
+      "name:ja": "プラスワン",
+      "shop": "repair"
+    }
   }
 }
--- a/brands/shop/repair.json
+++ b/brands/shop/repair.json
@@ -1,0 +1,16 @@
+{
+  "shop/repair|Mister Minit": {
+    "matchTags": [
+      "craft/electronics_repair",
+      "craft/key_cutter",
+      "craft/shoemaker"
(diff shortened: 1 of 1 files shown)


brands/craft/key_cutter.json
```diff
@@ -0,0 +1,12 @@
+{
+  "craft/key_cutter|Mister Minit": {
+    "tags": {
+      "brand": "Mister Minit",
+      "brand:wikidata": "Q1939269",
+      "brand:wikipedia": "de:Mister Minit",
+      "craft": "key_cutter",
+      "name": "Mister Minit",
+      "repair": "only"
```
bhousel, line 9: I feel a bit weird about the `repair=only` tag here.  Surely they aren't repairing keys that weren't cut properly?

sguinetti: The problem is that reviewing the [documentation](https://wiki.openstreetmap.org/wiki/Tag:craft%3Dkey_cutter), it also offers sale of things. However, this brand does not sell, so I added ["repair = only"](https://wiki.openstreetmap.org/wiki/Key:repair) (instead of "yes").

sguinetti: Edit: This only applies in shops, that kind of logic that generates some confusion, I will remove this tag.

bhousel: thanks, yeah looking here https://misterminit.co/  they seem like they are a global brand that does more than just cut keys.  I'd put them as a `shop=repair` and skip the `repair=*` tag.

sguinetti: This label of "repair shop" is the best choice. I fixed with matchTags.

sguinetti:
@bhousel In addition to this, I add another repair shop, PLUSONE. This brand is only in Japan, and I think it's its competition.

UKChris-osm:
This brand seems to offer a similar service to Timpson in the UK (#3437) so however this entry ends up in the NSI, it might be a good guide for how Timpson should finally make the cut :)

bhousel:
Thanks @sguinetti , LGTM :+1:

PR #3873, opened 2020-05-15
Add Kagi110-Key Ambulancy Association

sguinetti:
This company operates two locksmith shop brands. Two brands share the same logo, SNS and  website. Store names differ only by name.

### DIFF
--- a/brands/shop/locksmith.json
+++ b/brands/shop/locksmith.json
@@ -1,0 +1,28 @@
+{
+  "shop/locksmith|カギの110番": {
+    "countryCodes": ["jp"],
+    "tags": {
+      "brand": "カギの110番",
+      "brand:en": "Kagi110",
+      "brand:ja": "カギの110番",
+      "brand:wikidata": "Q94551271",
+      "name": "カギの110番",
+      "name:en": "Kagi110",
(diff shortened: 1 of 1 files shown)


Adamant36:
> This company operates two locksmith shop brands. Two brands share the same logo, SNS and website. Store names differ only by name.

uummmm, I sorta feel like they could have two different Wikidata entries then, but then I think Wikidata might have an issue with putting the same SNS and website on two separate pages. Although, people in OSM should be able to find what brand name they want by searching for a unique Wikidata code. So, I don't know 🤷

matkoniecz:
Seems OK to me.

ArenaL5:
This is very weird. The Wikidata entry is titled “Kagi110-Key Ambulancy Association”, and it seems like one name always goes after another in the website.

Am I right in thinking these used to be different brands, they're now managed as a single brand but didn't bother to unify the name? If that's the case, creating separate entries wouldn't help tag this business. I'd merge it as it is.

sguinetti:
> Am I right in thinking these used to be different brands, they're now managed as a single brand but didn't bother to unify the name? 

Yes. The company merged into one but the brands did not.

PR #3874, opened 2020-05-16
Add Radiowealth Finance

govvin:
Add item for Radiowealth Finance, a local chain of money lenders in the Philippines

### DIFF
--- a/brands/shop/money_lender.json
+++ b/brands/shop/money_lender.json
@@ -119,5 +119,20 @@
       "official_name:en": "SMBC Consumer Finance",
       "shop": "money_lender"
     }
+  },
+  "shop/money_lender|Radiowealth Finance": {
+    "countryCodes": ["ph"],
+    "tags": {
+      "brand": "Radiowealth Finance",
+      "brand:en": "Radiowealth Finance",
+      "brand:wikidata": "Q94572570",
+      "name": "Radiowealth Finance",
+      "official_name": "Radiowealth Finance Company",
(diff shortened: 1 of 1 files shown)


Adamant36:
Hello. Can you remove the email and website? Also, what's the name:abbr tag for?

govvin:
> Hello. Can you remove the email and website? Also, what's the name:abbr tag for?

name:abbr is a local convention to capture common abbreviations for popular features/establishments.

govvin:
> Hello. Can you remove the email and website? Also, what's the name:abbr tag for?

what's the convention for official emails and websites of chains of establishments?

Adamant36:
> name:abbr is a local convention to capture common abbreviations for popular features/establishments.

The only name tag variations we use are two digit language codes. If you want to show the abbreviation we usually go with the short_name tag for that. Or if it is used more then the official name, we will just put that in the name tag and use official_name for the official name. 

> what's the convention for official emails and websites of chains of establishments?

They are added by users on their side. They can also be added to Wikidata. There is to chances of variations based on the location, language, etc etc. Due to that, us including them would risk introducing bad data. Also, neither is stable enough anyway for a "set it and forget it" project like this. Otherwise, we would spend most of our time updating email addresses and website URLs.

govvin:
Thank you for the info @Adamant36 . I updated the item, per your suggestion.

bhousel:
Thanks @govvin :+1:

PR #3876, opened 2020-05-19
Add Dia & Go

Crashillo:


UKChris-osm:
I see that Dia already exists as a supermarket entry, so is "Dia & Go" a new brand or type of shop? Is it maybe more of a convenience store ??

Crashillo:
Actually, I agree with you, but I've transfered the question to the OSM Spain community, since in spain the concept is unclear. I'll feedback then.

UKChris-osm:
Thank you @Crashillo, I have marked this as a "check local" as you have asked the local OSM community for feedback.

If you decide to go with convenience, you can change this in your fork and it should automatically be made available in this pull request.

Crashillo:
It was close: convenience=yes 44% vs 56% convenience=no. So it's still a `shop=supermarket` :(
If it fits better to you, I may add `countryCode=es`

bhousel:
> It was close: convenience=yes 44% vs 56% convenience=no

@Crashillo , why would this be close?  Isn't it an either-or kind of decision?
I definitely wouldn't want to add something to this index that upsets 44% of the people of Spain.

Crashillo:
Our goal is be able to brand _Dia & Go_. The controversial point is to tag it as `convenience` or `supermarket`. I did a poll in the OSM Spain community about _how to tag it_, and that was the final result. Personally, I disagree, but I've to accept the majority decision :) (and that's why I suggest to scope the brand only to Spain, just in the future the brand spreads to France, Portugal, wherever... don't push them to tag it as supermarket, when it's a convenience)
But we actually need to take over the brand tagging. If in a future iteration, we decide to move to convenience, we would just update the values.

UKChris-osm:
I've done a little Googleing and do feel that convenience would be the best way to describe Dia & Go, especially after reading the [Dia & Go blog post](https://blog.dia.es/dia-and-go/)

Crashillo:
And you're right 😄 

I wrote a post on the spain mailing list encouraging the oppositors to propose their reasons. No answers yet. The poll is stucked at 47% (convenience) vs. 53% (supermarket), so the community are widely divided.

Do you mind to wait another week? I let them know we can accept convenience, insist on supermarket, or do nothing. If the latter so, we won't track this brand 🤷‍♂️

However I've to say, *Dia Market* and *La Plaza de DIA* are another labels of the group also, they're included here, and they're tagged as `shop=supermarket` 😇

UKChris-osm:
Yes this PR can wait, it's best to get it right than to rush it and fix it later I feel, especially as this won't end up in the live index on iD for a little while anyway.

I do think this brand should exist in the index, as it's clearly a real brand on a real retail outlet, and realistically if it's tagged slightly wrong (supermarket vs. convenience) it wouldn't be the end of the world, as right or wrong, all entries using it would be tagged the same :)

bhousel:
> Our goal is be able to brand _Dia & Go_. The controversial point is to tag it as `convenience` or `supermarket`. I did a poll in the OSM Spain community about _how to tag it_, and that was the final result. Personally, I disagree, but I've to accept the majority decision

It should be one of these:
* Dia & Go is always a supermarket
* Dia & Go is always a convenience store
* Dia & Go can be either thing, so we can add it to the index both ways.

(If people are unsure of the difference, they shouldn't be "voting")

Crashillo:
I think this "always" you before-mentioned is one of the most controversial matters. People distinguish between how large the place itself is.
Sometimes a franchise gets bigger enough to build a larger grocery, keeping the same brand but the property (the area) is larger. Let's say they put the tag based on a _geographical/physical_ criteria instead of a _business_ one.
Another controversial point is the word _convenience_. We don't use that word (only in a technical context) and [our wiki](https://wiki.openstreetmap.org/wiki/ES:Key:shop) translated that into _Ultramarinos_ (corner shop), so most of them cannot tag in equal way a franchise, even more when these places could have a butcher or fruit seller.

Anyway, yesterday we gave a boost to this, and the current poll turned over: 55% vs 45% (turnout 22 votes)

Crashillo:
Updated value.

bhousel:
> Updated value.

@Crashillo it's not enough to just update the `shop` tag - you need to remove the entry from the `supermarket.json` file and add it to the `convenience.json` file, and make sure the identifier and tags are all set correctly.   (It might be easiest to just back out the past few commits and start from scratch, then force push your changes to update this PR)

Crashillo:
Yep my bad, the rush to get this finally done

brands/shop/convenience.json
```diff
@@ -595,6 +595,15 @@
       "shop": "convenience"
     }
   },
+  "shop/supermarket|Dia & Go": {
```
bhousel, line 598: Thanks - make sure to change the identifier too!
```suggestion
  "shop/convenience|Dia & Go": {
```

bhousel:
Thanks for sticking with it @Crashillo - looks good now 👍 
Welcome to the project & hope you can add more of these and provide important local perspective.. 👋

PR #3882, opened 2020-05-20
use more specific Wikidata item for Rewe to Go

camelCaseNick:
[Q16968817](http://www.wikidata.org/entity/Q16968817) is about Rewe as supermarket chain in general, while [Q85224313](http://www.wikidata.org/entity/Q85224313) the specific Rewe to Go chain represents. I therefore propose to use the more specific one. This way the fetched website and the from Wikidata fetched logo (and from Facebook) are specific to this brand and chain.

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -1595,7 +1595,6 @@
     "tags": {
       "brand": "Rewe To Go",
       "brand:wikidata": "Q85224313",
-      "brand:wikipedia": "en:REWE",
       "name": "Rewe To Go",
       "shop": "convenience"
     }
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -1594,7 +1594,7 @@
     ],
     "tags": {
       "brand": "Rewe To Go",
-      "brand:wikidata": "Q16968817",
+      "brand:wikidata": "Q85224313",
       "brand:wikipedia": "en:REWE",
       "name": "Rewe To Go",
       "shop": "convenience"


Adamant36:
Is there no connection between the two companies? Like the same parent company or something?

camelCaseNick:
@Adamant36 Those are chains, not companies. They are all related to [Q573266](http://www.wikidata.org/entity/Q573266), though they don't run every store of that chain. For what reason do you need to “parent”? And I am not sure, why there are two logos.

Adamant36:
> @Adamant36 Those are chains, not companies.

A chain is still technically a company ;)

> For what reason do you need to “parent”?

A few reasons. 

Do you happen to know if Rewe to Go is a convenience store or a supermarket? From Wikipedia and pictures it looks like a supermarket, but it's in the shop=convenience file and you can't always go off of images in Google. 
Is

camelCaseNick:
@Adamant36 
> Do you happen to know if Rewe to Go is a convenience store or a supermarket? From Wikipedia and pictures it looks like a supermarket, but it's in the shop=convenience file and you can't always go off of images in Google.
I happen to know, that they are small stores in larger cities and at gas stations with a limited range of products. And according to [w:de:Convenience Shop](https://de.wikipedia.org/wiki/Convenience_Shop) it is limited to about 1200 products. (Have not checked in one, though). It's also a good sign, that it's mentioned in that Wikipedia article, right?
I am pretty certain it's a `shop=convenience` and not a `shop=supermarket`.

Adamant36:
Hhhmmm, OK. I'll take your word on that. My main thing, and why I brought up the parent company thing, is that generally we only include Wikipedia entry if it is directly about the brand. In this case it is about REWE though and not Rewe to Go. So it should be removed. 

It's not super relevant to the NSI per say, but it seems from how you setup the Wikidata article etc that Rewe to Go is owned by Rewe, but Rewe is subsidiary of the Rewe Group? Or are they both subsidiaries of Rewe Group?

camelCaseNick:
@Adamant36 Yes, the Wikipedia article link could have been removed. And yes, Rewe To Go is part of Rewe as a brand and thus owned by the Rewe Group.
(The individual stores, here have this brand and get supplied by Rewe, but might be run be Aral etc., but that would be for `operator:*` so both out of scope here and dependent on the individual stores.)
I can make another commit also removing the Wikipedia tag suggestion, if wished.

camelCaseNick:
And to clarify: Rewe To Go is a trademark of the REWE Markt GmbH, the one that is also having the plain Rewe branded stores. So it is a different chain, but legally part of the other one.

brands/shop/convenience.json
```diff
@@ -1594,7 +1594,7 @@
     ],
     "tags": {
       "brand": "Rewe To Go",
-      "brand:wikidata": "Q16968817",
+      "brand:wikidata": "Q85224313",
       "brand:wikipedia": "en:REWE",
```
bhousel, line 1598: If you are changing the `brand:wikidata` to a better QID, please remove the `brand:wikipedia` tag, which still points to the old REWE Wikipedia page.  Ideally there will eventually be a separate Wikipedia page for REWE To Go.

PR #3890, opened 2020-05-28
add hardware MR.DIY

MASangys:


brands/shop/hardware.json
```diff
@@ -130,4 +130,17 @@
       "shop": "hardware"
     }
   }
-}
\ No newline at end of file
+  "shop/hardware|Home Hardware": {
```
matkoniecz, line 133: It probably should be "shop/hardware|MR.DIY"

brands/shop/hardware.json
```diff
@@ -130,4 +130,17 @@
...
+    "countryCodes": ["my"],
+    "nomatch": [
+      "shop/doityourself|MR.DIY"
```
matkoniecz, line 136: Is "shop/doityourself|MR.DIY" added already to NSI if it is also a valid value?

matkoniecz: And if `shop=doityourself` should be replaced by `shop=hardware` then nomatch should be replaced by match

MASangys:
I am pretty not sure about this. The category of this store should be home retailling. So best match it is hardware.
https://www.mrdiy.com/page/236/About-Us/

matkoniecz:
Based on further discussion on telegram it is rather diy shop and I was asked to complete this PR.

I will do it now.

PR #3893, opened 2020-05-28
add popular

MASangys:
may also add this for taiwan https://en.wikipedia.org/wiki/Eslite_Bookstore

### DIFF
--- a/brands/shop/books.json
+++ b/brands/shop/books.json
@@ -263,6 +263,18 @@
       "shop": "books"
     }
   },
+  "shop/books|Popular": {
+    "countryCodes": ["my", "sg", "hk"],
+    "tags": {
+      "brand": "Popular",
+      "brand:zh": "大众书局",
+      "brand:wikidata": "Q4167785",
+      "brand:wikipedia": "en:Popular Holdings",
+      "name": "Popular",
+      "name:zh": "大众书局",
+      "shop": "books"
+    }
+  },
   "shop/books|Saraiva": {
     "countryCodes": ["br"],
     "tags": {


matkoniecz:
Several are `shop=stationery`.

Is it mistagging, preferred tagging for some location or a separate company?

MASangys:
dont Know but Popular bookstore do sell stationery also. Query should only in Malaysia. (or Singapore, Hong Kong etc)

MASangys:
and Eslite is in Taiwan

ArenaL5:
> dont Know but Popular bookstore do sell stationery also. Query should only in Malaysia. (or Singapore, Hong Kong etc)

I checked @matkoniecz's query. Most of them are `shop=books`, and one of them is `amenity=library`. The website advertises Popular Holdings' as a chain of different businesses, all of them related to book publishing.

If those are correctly mapped, and they only sell books, I'd like you to add an entry to `shop/books` as well. This way mappers can tag separately bookstores and stationery shops.

If you're sure _all_ of them sell stationery, then most nodes in the map are wrong. You can [add `matchTag`s](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#matchnamesmatchtags) so that mappers will be prompted to update Popular Holdings' bookstores to Popular Holdings' stationery shop.

If you're unsure, ask or research.

Also, don't we need more `countryCode`s? Their website claims to have businesses in mainland China, Canada and United Kingdom, too.

> and Eslite is in Taiwan

You know how to add it now. You can push another commit. :slightly_smiling_face:

MASangys:
So should we create 2 tags for Popular bookstore? They identify them as a bookstore but in Malaysia, commonly we can find stationery in bookstore also. Sontag they can survive and got enough earning.  So I would like to suggest just put bookstore for Popular.

ArenaL5:
> They identify them as a bookstore but in Malaysia, commonly we can find stationery in bookstore also.

Okay, I think I understand.

So the ones tagged with `shop=stationery` _are_ still bookstores, they just sell some stationery on the side. Is that right?

If so, **don't** touch `brands/shop/stationery.json`, but please [add a `matchTag`](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#matchnamesmatchtags) to 
`"shop/books|Popular"`. Otherwise, the validator won't work for those nodes tagged as `shop=stationery`.

MASangys:
yes

ArenaL5:
Okay. Push another commit then.

After that, push a third commit to add Eslite Bookstore, too.

bhousel:
> please [add a `matchTag`](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#matchnamesmatchtags) to
> `"shop/books"`

I did this in a9f46833 and merged.
I also added `"nomatch": ["amenity/bank|Popular"]`  (because there is a bank with this name too)

@MASangys you can create a separate PR for the Eslite Bookstore and additional countryCodes in a few days?

PR #3896, opened 2020-05-29
Add post office in Malaysia

MASangys:


bhousel:
Thanks @MASangys 👍

bhousel:
In 888c4b62 I also:
* added 'ph' for J&T Express, as they seem to operate in the Philippines.
* removed the redundant matchNames property (name matching is already case insensitive)

PR #3899, opened 2020-05-29
add body glove

MASangys:


matkoniecz:
Hmm, based on WIkipedia it sounds like `shop=sports` but all currently existing ones in OSM are `shop=clothes` ( http://overpass-turbo.eu/s/UvB )

PR #3904, opened 2020-05-29
ADD KK SUPER MART

MASangys:


matkoniecz:
Note https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#adding-references-to-wikidata part - without either Wikipedia article link or independent references Wikidata entries will be deleted.

MASangys:
I had created a page for this mart
https://ms.wikipedia.org/wiki/KK_Super_Mart

MASangys:
and then some of the kk super mart was tagged as name Kk or KK mart , can validator detect it?

MASangys:
e.g. 
https://www.openstreetmap.org/search?query=kk%20malaysia#map=19/3.13305/101.71147

MASangys:
https://www.openstreetmap.org/search?query=kk%20mart%20malaysia#map=19/3.12737/101.67019

MASangys:
and this also https://www.openstreetmap.org/search?query=kk%20supermart%20malaysia#map=19/3.13982/101.62823

bhousel:
> and then some of the kk super mart was tagged as name Kk or KK mart , can validator detect it?

You would need to add `matchNames` for the alternate spellings if you want the validator to suggest a tag correction.  

You can also use tags like `alt_name` or `short_name` if these names are somewhat official.  In other words, if everyone locally knows that "KK" means "KK Super Mart" or the business puts "KK" on their signage, it's probably ok to put this in the `short_name` tag.

MASangys:
No proble, you can just do it. I agree

bhousel:
> No proble, you can just do it. I agree

I'm not familiar with this brand, so I'm not really the right person to make that decision.

MASangys:
What you say are logically correct.
First:Match anme for this brand need to be add
Second:Short name for KK tag.

Adamant36:
> without either Wikipedia article link or independent references Wikidata entries will be deleted.

From what I understand we can use the NSI identifier also now that it has been accepted. Although, I don't think it should be a full replacement to good references or a Wikipedia article link if either are also available.

bhousel:
> What you say are logically correct.
> First:Match anme for this brand need to be add
> Second:Short name for KK tag.

I ended up making these changes in 272e5233 and merging.. Thanks for the PR @MASangys 👍

Issue #3905, opened 2020-05-29
Add internet tags to Premier Inn


Cj-Malone:
[Premier Inn](https://nsi.guide/index.html?k=tourism&v=hotel#Premier%20Inn) hotels have free wifi, would it be best to add the following to the template?
```
internet_access=wlan
internet_access:fee=customers
```

matkoniecz:
Is it true for every single one?

Cj-Malone:
[Yes](https://www.premierinn.com/gb/en/why/wifi.html)

They even say they use the same SSID, so that could be added as well.

```
internet_access:ssid=Premier Inn Free Wi-Fi
```

bhousel:
Seems fine, as long as we are sure that it's true for every one. 👍

PR #3910, opened 2020-05-29
Add WiFi to Premier Inn preset

Cj-Malone:
As per #3905 add standard values to Premier Inn preset.

Note, I've not tested this, but I think you have CI set up.


bhousel:
Thanks @Cj-Malone and welcome 👋 
closes #3905

Issue #3854, opened 2020-05-10
add insurance=health for some "Krankenkasse"-presets in germany


Lukas458:
In Germany, there are presets for "AOK", "DAK Gesundheit" and "Techniker Krankenkasse" which are common health-insurance brands here. The tag insurance=health (https://wiki.openstreetmap.org/wiki/Tag:insurance%3Dhealth) is used not so often, but specifies those insurance offices which are specialized (only) on health insurances. On the tagging-mailing list, people said it would be a good purpose to use an under-tag for this. So I think this tag might be added to the presets.

bhousel:
Thanks for raising this, it's a good suggestion..

We currently have 74 entries under `office/insurance` (57 of which are complete with wikidata and social links). I think it would be great to eventually have all of them tagged with the `insurance=*` tag, and raise a warning if the sub tag is missing (similar to what we do now with tags like `cuisine` and `vending` etc).

PR #3912, opened 2020-05-29
added insurance=health for some "Krankenkasse" presets in germany

Lukas458:
See #3854 - added insurance=health for specifying those insurances offices, which are specialized only on health insurances. In germany we have some brands of them and insurance=health is the proposed tag to add.

### DIFF
--- a/brands/office/insurance.json
+++ b/brands/office/insurance.json
@@ -16,6 +16,7 @@
       "brand:wikidata": "Q87070013",
       "name": "A-Plan Insurance",
       "office": "insurance"
+      "insurance": "health"
     }
   },
   "office/insurance|AAA Insurance": {
@@ -210,6 +211,7 @@
       "brand:wikipedia": "de:DAK-Gesundheit",
       "name": "DAK",
       "office": "insurance"
+      "insurance": "health"
     }
   },
   "office/insurance|DEVK": {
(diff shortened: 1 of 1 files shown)


brands/office/insurance.json
```diff
@@ -16,6 +16,7 @@
       "brand:wikidata": "Q87070013",
       "name": "A-Plan Insurance",
       "office": "insurance"
```
matkoniecz, line 18: Have you tried running build on your computer?  See https://github.com/osmlab/name-suggestion-index#installing (if it is unclear or insufficient - what is th problem with instructions?)

`       "office": "insurance"` is missing comma, as you added a new line (and in json last line and only last line is without ,)

Lukas458: Oh sorry, I just didn't notice about that. Of course, I missed the comma, yes that's true. But it will be easy to fix...

bhousel:
Thanks @Lukas458 and welcome 👋 
I fixed the syntax in fc3d0285 and merged.

bhousel:
By the way.. what is going on with the `insurance=* tag??  The commonly used values are pretty much nonsense.
https://taginfo.openstreetmap.org/keys/insurance#values

Can someone take this up with the tagging folks on the mailing list?  It would almost be best to just remove all values except for `health` and start from scratch with a schema that is useful.

PR #3913, opened 2020-05-29
Add telecomunication service company in Malaysia

MASangys:


boothym:
Is shop=telecommunication really the right tag for these places? (it's only used 371 times worldwide) 

Because the brands look like mobile phone providers, shop=mobile_phone would surely be better. There's already two Digi Store shops in Malaysia, both tagged that way, e.g. https://www.openstreetmap.org/node/4366543058

MASangys:
They dont sell handphone but sell internet data plan to people. So should be telecommunication

Adamant36:
> They dont sell handphone but sell internet data plan to people. So should be telecommunication

Generally, if a place doesn't sell physical products, but a "service", then the proper tag would be office. In this case office=telecommunication has like 19,000 uses and my guess is that most (if not all) of the uses of shop=telecommunication is miss-tagging. So I think office=telecommunication would be the proper thing to go with in this instance if the brands don't sell actual hardware. 

That said, it looks like all the brands except for TMpoint and Yes 4G sell hardware. In the case of Yes 4G though it looks like some locations do and some don't. But we shouldn't tag mobile phone stores as offices or visa versa. So we might need two entries for that one and whatever other ones can be shown to have both mobile phone shops and offices. 

Cellcom
Maxis
U Mobil

MASangys:
I agree with your opinion.

bhousel:
> Is shop=telecommunication really the right tag for these places?

Yes, this index has been using `shop=telecommunication` for retail storefronts since #3120 , see also discussion on #3113 #2900

We've actually had a lot of previous discussion about this, so I'd appreciate people not switching back to `office=telecommunication` which is being used for office buildings where people work and not for retail places where people shop.

bhousel:
@MASangys I just checked all of these and they look great 👍   
Thanks for your work in adding all of the Malaysian brands!

Just wanted to remind you that you don't need to add `matchNames` alternate names for variations in capitalization.  I'll remove the redundant names in a followup commit.

boothym:
@bhousel The reason I asked the question is because there's nothing tagged as shop=telecommunication in Malaysia, with those brands being added in this PR currently using the office and mobile_phone tags. 

So I thought there may have been a mistake with the choice of tag, especially as normally the NSI values match the current usage.

I had no idea that shop=telecommunication was now in iD and NSI (guessing that's why it's gone from 50 to ~400 uses since October!) I agree that it is a better tag than office so I have no issue with it being added.

Can I suggest that if iD and NSI are going to popularise a new shop tag then it should be documented on the wiki? Especially so that the difference to shop=mobile_phone and the office tag is made clear - it might be a better tag for some places, but people don't know about it.

bhousel:
> Can I suggest that if iD and NSI are going to popularise a new shop tag then it should be documented on the wiki? Especially so that the difference to shop=mobile_phone and the office tag is made clear - it might be a better tag for some places, but people don't know about it.

This is a reasonable request, but I don't personally edit the OSM wiki anymore.  I think there are people who watch for my edits just to revert them.

Adamant36:
> This is a reasonable request, but I don't personally edit the OSM wiki anymore. I think there are people who watch for my edits just to revert them.

That's especially true if your creating an article for a new tag that doesn't have a ton of usage yet, but even if it does people always come out of the woodwork to butcher the page and criticize the tag in any way they can.

PR #3914, opened 2020-05-29
Add cinemas in Malaysia

MASangys:


UKChris-osm:
Hi @MASangys 

All seems okay, but are you able to add any social media links (Facebook, Twitter, etc) to the Wikidata pages of these cinema's?

The NSI and the iD Editor can use social media links in Wikidata's to display the logo of each brand, making it a lot easier for people to see and recognise the brand they are looking for.

Currently, it seems only Golden Screen Cinemas has any social media links in its Wikidata article.

MASangys:
added for FB

UKChris-osm:
Thanks for these additions to the Name Suggestion Index @MASangys :)

PR #3923, opened 2020-05-31
BB&T and SunTrust merger to Truist Financial

Eric-Sparks:
BB&T and SunTrust recently merged into a single company now called Truist Financial.  The actual buildings haven't changed their branding, yet, but the wikipages have been merged.

I suggest changing the brand wikipage and wikidata to the current tags and add the Truist brand to future-proof the system.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -459,7 +459,7 @@
       "amenity": "bank",
       "brand": "BB&T",
       "brand:wikidata": "Q795486",
-      "brand:wikipedia": "en:BB&T",
+      "brand:wikipedia": "en:Truist Financial",
       "name": "BB&T",
       "official_name": "Branch Banking and Trust Company"
     }
@@ -5896,8 +5896,8 @@
     "tags": {
       "amenity": "bank",
       "brand": "SunTrust",
-      "brand:wikidata": "Q181507",
-      "brand:wikipedia": "en:SunTrust Banks",
+      "brand:wikidata": "Q795486",
+      "brand:wikipedia": "en:Truist Financial",
       "name": "SunTrust"
     }
   },
@@ -6066,6 +6066,17 @@
       "brand:wikidata": "Q3520318",
       "brand:wikipedia": "en:The Co-operative Bank",
       "name": "The Co-operative Bank"
+    }
+  },
+  "amenity/bank|Truist Financial": {
+    "countryCodes": ["us"],
+    "matchNames": ["truist bank"],
+    "tags": {
+      "amenity": "bank",
+      "brand": "Truist",
+      "brand:wikidata": "Q795486",
+      "brand:wikipedia": "en:Truist Financial",
+      "name": "Truist Financial"
     }
   },
   "amenity/bank|Türkiye İş Bankası": {


UKChris-osm:
When you say they haven't changed their branding "yet", does that mean they do intend on updating the branding of every branch to Truist, or is it possible the parent companies have merged, yet the two banking brands might remain as independent entities?

I only ask as over in the UK, Lloyds Bank owns Halifax bank (via Bank of Scotland), yet both trade independently, and so retain their own branding, and the Wikipedia page lists Truist as a "holding company".

Eric-Sparks:
@UKChris-osm According the [Truist website](https://www.truist.com/banking-with-us), it seems that it is a "yet".

> Any account you open now at BB&T or SunTrust will eventually become a Truist account. So while we’re creating new products and services for Truist, you can open an account with us by visiting BBT.comopens in a new tab, SunTrust.comopens in a new tab, or any branch location.

Adamant36:
> Any account you open now at BB&T or SunTrust will eventually become a Truist account. So while we’re creating new products and services for Truist, you can open an account with us by visiting BBT.comopens in a new tab, SunTrust.comopens in a new tab, or any branch location.

Usually we favor what is on the actual signs at the branch locations due to the on the ground rule. It sounds like in this case that it's either not certain that the signs are going to change or it will at least be awhile until they do. So, my recommendation is that we wait to merge this until the actual branch signs have been altered to reflect the new branding.

Eric-Sparks:
So, the patch has two parts: one that is a current update to BB&T and SunTrust and then one that was for the future.  I would suggest removing the future part and keeping the updates to the existing brands.

Adamant36:
> So, the patch has two parts: one that is a current update to BB&T and SunTrust and then one that was for the future. I would suggest removing the future part and keeping the updates to the existing brands.

 Normally, I'd go with that. Except if a brand is "dissolved" (the case when it changes), I think there is more to the process in the NSI then just changing the Wikipedia entry. We don't usually use the parent companies Wikipedia entry anyway. In this case, my recommendation is to leave the existing entry as is for now and add the entry for Truist. That way people can choose to tag the banks based on whatever the signs are on the ground. Although, the name etc should be Truist instead of Truist Financial. Since that seems to be what their signs say. 
Truist

Adamant36:
BTW, that image is a good example of why. The larger sign says Truist, but the one above the door still says BB&T 🤷‍♂️

bhousel:
I also prefer to leave the BB&T and Suntrust entries alone for the time being.  The QIDs and logos for the old branches should stay valid, and whenever the branches rebrand mappers can switch over to the new Truist preset.

Also, that image of the giant "Truist" sign is pretty fake.  Let's not use that as a basis for any decision making.

Adamant36:
> Also, that image of the giant "Truist" sign is pretty fake. Let's not use that as a basis for any decision making.

Oh yeah, my bad. I didn't even notice that. I totally should have though. I agree it shouldn't be used for any decision making.

bhousel:
From Wikipedia:  
> The merger was completed at midnight December 6, 2019.[64] BB&T CEO Kelly King retained the same position with the new company. BB&T customers could begin using SunTrust ATMs free of charge, and SunTrust customers could begin using BB&T ATMs free of charge.[65] The merged bank will continue to operate under the BB&T and SunTrust names until the two banks' systems are combined, a process that could take as long as two years.

Unfortunately, someone on the wiki side has renamed both the [en:wiki](https://en.wikipedia.org/w/index.php?title=BB%26T&redirect=no) and [wikidata](https://www.wikidata.org/wiki/Q795486) pages to Truist Financial.  I really wish people would leave the predecessor entities alone as their own pages and instead make a new entry for the merged entity.  We can't really change this, but it will make things messy for us for "up to two years", because now we can't track the BB&T brand as its own thing.

bhousel:
In de4b5c3e I restored SunTrust to Q181507, left Truist as Q795486, and created Q95984154 to hold the properties for the former BB&T brand.  This lets us keep track of all three brands while the merger works out.
I hope this is acceptable for the Wikipedia/Wikidata folks.  🤷

PR #3927, opened 2020-06-01
Add Treasure Bomb

sguinetti:
This pull request adds a second hand store. For now, I don't found SNS but I found this [shop list](https://www.okoku.jp/inc/shop).


bhousel:
Thanks @sguinetti 👍

Issue #3884, opened 2020-05-23
Add country codes to Yoshinoya


hanchao:
Yoshinoya (𠮷野家, TYO: 9861) is a Japanese multinational fast food chain, and the second-largest chain of gyūdon (beef bowl) restaurants. The chain was established in Japan in 1899. Its motto is "Tasty, low-priced, and quick".

Yoshinoya has a chain of stores in Japan, the United States, Hong Kong, Cambodia, China, Indonesia, Malaysia, Philippines, Singapore, Taiwan, and Thailand.

![img](https://upload.wikimedia.org/wikipedia/commons/thumb/0/00/Yoshinoya_at_Beijing_West_Railway_Station%2C_Exit_2_%2820190528070627%29.jpg/2560px-Yoshinoya_at_Beijing_West_Railway_Station%2C_Exit_2_%2820190528070627%29.jpg)
https://en.wikipedia.org/wiki/Yoshinoya
```
"amenity/fast_food|吉野家": {
    "countryCodes": ["cn"],
    "tags": {
      "amenity": "fast_food",
      "brand": "吉野家",
      "brand:en": "Yoshinoya",
      "brand:cn": "吉野家",
      "brand:wikidata": "Q776272",
      "brand:wikipedia": "ja:吉野家",
      "cuisine": "beef_bowl",
      "name": "吉野家",
      "name:en": "Yoshinoya",
      "name:cn": "吉野家",
      "takeaway": "yes"
    }
  }
```

hanchao:
![Uploading 83314488...]()

ArenaL5:
Yoshinoya already exists in the index, but it's only listed in Japan:

https://github.com/osmlab/name-suggestion-index/blob/b9c5ddec24e173b3e544148b8db360a2ad993c3f/brands/amenity/fast_food.json#L4297-L4312

You're right we should add those other countries, not only China. Why don't you create a PR to add them?

You can have a list of country codes as seen in point 3 of [:convenience_store:   Add missing brands](https://github.com/osmlab/name-suggestion-index/blob/master/CONTRIBUTING.md#convenience_store--add-missing-brands) in the manual.

There should also be a separate `"amenity/fast_food|Yoshinoya"` entry, with different `name` and `brand` values, for those countries that don't use hanzi:

```json
"amenity/fast_food|Yoshinoya": {
    "countryCodes": [ ADD COUNTRY CODES HERE ],
    "tags": {
      "amenity": "fast_food",
[...]

PR #3935, opened 2020-06-03
add country codes to Yoshinoya

hanchao:
Fix https://github.com/osmlab/name-suggestion-index/issues/3884


bhousel:
Thanks @hanchao this looks perfect 👍

PR #3936, opened 2020-06-03
Changes to amenity/fuel

luisluigi639:
Adding Wikidata for Classic Oil and changing Copec to correct Wikidata.

### DIFF
--- a/brands/amenity/fuel.json
+++ b/brands/amenity/fuel.json
@@ -696,6 +696,8 @@
     "tags": {
       "amenity": "fuel",
       "brand": "Classic",
+      "brand:wikidata": "Q33127117",
+      "brand:wikipedia": "de:Classic Tankstellen",
       "name": "Classic"
     }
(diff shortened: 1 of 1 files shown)


matkoniecz:
Is Empresas Copec completely wrong or just imprecise?

luisluigi639:
@matkoniecz Empresas Copec just imprecise, the parent company is called Empresas Copec and the gas sation chain is called just Copec.

luisluigi639:
I think I need some help with this last commit.

bhousel:
No problem @luisluigi639  I sorted out the conflict and merged.

I left the name "Amic" as it was (instead of renaming it to "Amic Energy") because it's already tagged this way [76 times](https://github.com/osmlab/name-suggestion-index/blob/master/dist/names_keep.json#L1048) and the stations show the "energy" part very small: http://amicenergy.com/

PR #3938, opened 2020-06-06
Add TCC (Colombia)

desiderantes:
Also add wikidata for Servientrega

### DIFF
--- a/brands/amenity/post_office.json
+++ b/brands/amenity/post_office.json
@@ -397,15 +397,16 @@
       "name": "Slovenská pošta"
     }
   },
-  "amenity/post_office|TCC (Colombia)": {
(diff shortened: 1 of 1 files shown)


Adamant36:
Hi there. Thanks for the contribution. It looks fine. One thing though, you don't need the "(Colombia)"  part in any of the instances of TCC (Colombia) except for the Wikipedia entry because there isn't a another TCC in the NSI that it would be confused with. Even if there was though, we usually disambiguate entries a different way.

desiderantes:
Thank you, I updated the file.

bhousel:
Thanks @desiderantes, this looks good.. Welcome! 👋

PR #3941, opened 2020-06-08
Add Sabeco group brands (spanish filial from Auchan group)

Crashillo:
Sabeco group involves also _Simply Market_, who is already included in the guide as a worldwide brand. That's OK and the establishment is the same, but the other brands are regional scoped since they belong to the Sabeco group, who operates only in spain

Further info (in spanish):
- https://es.wikipedia.org/wiki/Sabeco#Tipos_de_establecimientos
- https://www.simply.es/pagEstatica.aspx?id=7

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -1787,7 +1787,6 @@
   },
   "shop/convenience|Simply Basic": {
     "locationSet": {"include": ["es"]},
-    "matchNames": ["simply"],
     "tags": {
       "brand": "Simply Basic",
       "brand:wikidata": "Q5643014",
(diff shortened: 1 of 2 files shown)


bhousel:
Can you double check these?  They all use the id Q11893827, which points to the entry for Siwa (a difference convenience store)
https://www.wikidata.org/wiki/Q11893827

Crashillo:
Fixed ✌️

bhousel:
Oh also, they can't all `matchName` the string "Simply" .  That would insert duplicates into the match index. Those all have to be removed.

bhousel:
Thanks @Crashillo  - I think this is ok..
The effect of this change is that users in Spain will have the option of using the alternative brand names.

PR #3948, opened 2020-06-08
Set specific Wikidata for Ueshima Coffee House

sguinetti:
This pull request edits the brand to a more specific one. Since a new Wikidata was created, I added specific information about coffee shops, in addition to using exclusive social networks. This replaces the old Wikidata with UCC.


Adamant36:
BTW, if you want to add short_name=UCC at to this at some point that it might be worth doing just in case some are tagged that way.

PR #3960, opened 2020-06-09
Add PocketChange

sguinetti:
This pull request adds a money autoservice.

### DIFF
--- a/brands/amenity/bureau_de_change.json
+++ b/brands/amenity/bureau_de_change.json
@@ -38,6 +38,17 @@
       "name": "No1 Currency"
     }
   },
+  "amenity/bureau_de_change|PocketChange": {
+    "locationSet": {"include": ["jp", "tw"]},
+    "tags": {
+      "amenity": "bureau_de_change",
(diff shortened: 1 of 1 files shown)


brands/amenity/bureau_de_change.json
```diff
@@ -38,6 +38,17 @@
       "name": "No1 Currency"
     }
   },
+  "amenity/bureau_de_change|PocketChange": {
+    "locationSet": {"include": ["jp", "tw"]},
+    "tags": {
+      "amenity": "bureau_de_change",
```
tas50, line 44: Does it make sense to tag coin to bill converters as bureau_de_change? Is there potentially a better way to tag these since there isn't actual currency conversion happening here as described in the wiki.

sguinetti: It is honestly a machine that performs currency conversions (dollar, yen and euro). It is not an ATM because it is not anchored to a bank. Nor is it Western Union. So it's the label that comes closest.

PR #3972, opened 2020-06-12
Add a match for an errant plural on Costcutters

peternewman:


bhousel:
Good suggestion, thanks 👍

Issue #3959, opened 2020-06-09
FR : Add Utopia (french cinema chain)


overflorian:
Hello, could you please add:

Country code: FR (only)
"amenity": "cinema"
"brand": "Utopia"
"brand:wikidata": "Q3552766"
"brand:wikipedia": "fr:Utopia (cinéma)"

matthenschke:
I'll make a pull request asap as this is my first ever pull request on open source.

matthenschke:
Just created a pull request https://github.com/osmlab/name-suggestion-index/pull/3973

PR #3973, opened 2020-06-12
added utopia (french cinema chain) issue #3959

matthenschke:


bhousel:
Thanks @matthenschke and welcome 👋 
This looks good.. For future can you configure your editor not apply formatting to the .json file?  When we `npm run build` the code will just sort and format all the files.

PR #3974, opened 2020-06-12
Add Frisby (Colombia)

desiderantes:


UKChris-osm:
Hi @desiderantes 

I see the "name" value is set to KFC - was this because you copy & pasted the KFC entry as your starting point, or is Frisby a sub-brand of KFC in Colombia?

desiderantes:
Copy-paste error, sorry

Issue #3889, opened 2020-05-27
Add: Halfords (The Netherlands)


boothym:
Could someone please add service:bicycle:repair=yes to Halfords, as their shops do bike repairs as well.

I also noticed from an overpass turbo query that there are a number of Halfords shop in the Netherlands https://www.halfords.nl/ https://nl.wikipedia.org/wiki/Halfords

Wonder if these should have different wikidata and wikipedia to the UK & IRL shops?

SilentSpike:
Do *all* of their shops do repair?

I think the NL brand is different, looks to have a different logo.

boothym:
It seems like it - go to https://www.halfords.com/locations uncheck garage and pickup point, then search for some random locations - all the ones I've seen have bike repair listed under services.

UKChris-osm:
The Halfords web site shows "_424_" retail stores in total on the locations page, and when using the "_byicycle repair_" filter this remains at "_424_", which to me confirms they do bicycle repair in all stores.

UKChris-osm:
Halfords in The Netherlands used to be part of the Halfords UK brand, but is now a separate company, and so should have its own entry in the NSI.

Looking at it though, I get the feeling they are more of a bicycle retail chain that also does car parts, compered to the UK chain that's more of a car parts retailer with cycling as an added extra.

Adamant36:
> I've added service:bicycle:repair=yes to Halfords.

I'm not sure adding a bicycle repair tag is a good idea. Since it doesn't seem like all their locations provide the service. Otherwise, they wouldn't have a filter for it in their location search to find ones that do.

UKChris-osm:
> I'm not sure adding a bicycle repair tag is a good idea. Since it doesn't seem like all their locations provide the service. Otherwise, they wouldn't have a filter for it in their location search to find ones that do.

I'll happily remove it if it's deemed not suitable @Adamant36 

However, I am confident all the Halfords shops listed as "shop/car_parts" do have bicycle repairs, based on the filter.

The Halfords web site lists both general Halfords shops, and the "Halfords AutoCentre", which is why I think the filter exists. I first removed all "Autocentres", which listed the remaining general Halfords stores as having "424", then selecting the bicycle repair filter remained with "424" stores listed, which to me tells me all Halfords within "shop/car_parts" do have bicycle repair facilities, whereas the Autocentres do not.

Adamant36:
[...]

PR #3976, opened 2020-06-13
Excluded "nl" from GB Halfords

UKChris-osm:
Via #3889 it details that there is a Halfords brand in the Netherlands that isn't a part of the UK brand, as there is currently no separate Netherlands brand entry for Halfords.

I have added an **exclude** clause to the UK Halfords which excludes the NL variation, but as I've not come across locationSet yet, I wanted to check if doing this would be worthwhile?

As I understand it, iD will currently only offer the UK Halfords brand when adding a node or way in the UK now, as it has location data attached to it. But, for the entries in the Netherlands that are already incorrectly branded with the UK Halfords, would adding an exclude clause do anything meaningful, such as add a warning on iD that says "this brand doesn't appear to be present in this Country"?

### DIFF
--- a/brands/shop/car_parts.json
+++ b/brands/shop/car_parts.json
@@ -83,7 +83,10 @@
     }
   },
   "shop/car_parts|Halfords": {
-    "locationSet": {"include": ["gb", "ie"]},
+    "locationSet": {
+      "include": ["gb", "ie"],
+      "exclude": ["nl"]
+    },
     "matchTags": ["shop/bicycle"],
     "nomatch": [
       "shop/car_repair|Halfords Autocentre"


Adamant36:
I was thinking maybe it would go off the brand wikidata tag and the country codes. SSo if you have two brands with the same name in the same country, but different country codes and wikidata ids it would offer to upgrade the one that doesnt have country codes for that location for the one that does. I can't swear by that though, but it makes sense somehow.

bhousel:
Oh, what we actually want to do is add a separate entry for the NL brand.  It doesn't make sense to "exclude" Netherlands from Great Britain and Ireland.

Don't worry about what iD does for now.. locationSet support is not implemented there yet for the presets or validators. 

You can experiment with locationSets on this site:
https://ideditor.github.io/location-conflation/

Adamant36:
@bhousel, BTW as of a few merged PRs ago it seems there's a duplicate plugin error when running npm run docbuild now.

bhousel:
> BTW as of a few merged PRs ago it seems there's a duplicate plugin error when running npm run docbuild now.

oh thanks i'll check that - I haven't run docbuild in a while!

bhousel:
update: the `npm run docbuild` errors look fixed now 😅 
"babel-preset-nano-react-app" was doing some stuff.

PR #3978, opened 2020-06-15
add heytea wikidata

hanchao:
Hey Tea (Chinese: 喜茶) is a Chinese tea drink chain founded in 2012 and headquartered in Nanshan District, Shenzhen.By the end of Aug, 2019, Hey Tea operates 265 stores in 37 Chinese cities.
img


bhousel:
Thanks @hanchao 👍 
It looks like they have a few locations in Singapore, so I added `sg` to the include list in ca4579db

PR #3987, opened 2020-06-16
Add Cycle Times

sguinetti:
This pull request adds [Cycle Times](https://times-info.net/info/bike.html) parking service under Times brand.


bhousel:
Thanks @sguinetti  👍

PR #3996, opened 2020-06-19
Add Qingfeng Steamed Dumpling Shop

hanchao:
fast foot in china
https://zh.wikipedia.org/wiki/%E5%BA%86%E4%B8%B0%E5%8C%85%E5%AD%90%E9%93%BA
https://overpass-turbo.eu/s/VgS
image

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -4531,6 +4531,22 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|庆丰包子铺": {
+    "locationSet": {"include": ["cn"]},
+    "tags": {
+      "amenity": "fast_food",
+      "brand": "庆丰包子铺",
+      "brand:en": "Qingfeng Steamed Dumpling Shop",
+      "brand:wikidata": "Q24841010",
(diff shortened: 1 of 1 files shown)


brands/amenity/fast_food.json
```diff
@@ -4531,6 +4531,22 @@
       "takeaway": "yes"
     }
   },
+  "amenity/fast_food|庆丰包子铺": {
+    "locationSet": {"include": ["cn"]},
+    "tags": {
+      "amenity": "fast_food",
+      "brand": "庆丰包子铺",
+      "brand:en": "Qingfeng Steamed Dumpling Shop",
+      "brand:wikidata": "Q24841010",
```
tas50, line 4540: Is this the correct wikidata entry. It's lacking any fields and will most likely be deleted by the admins there.

hanchao: It is correct. I added an English name to it

bhousel:
Thanks @hanchao  - from their Wikipedia it does look like they have 140 locations, so it sounds ok for inclusion..  
I can't find an official website or social media pages, but if you can add some more details to the Wikidata page it would be very helpful!

PR #3997, opened 2020-06-22
Add Wenzel's

peternewman:
The shops just say "Wenzel's" on the front, but on their website it's Wenzel's the bakers:
https://www.wenzels.co.uk/

### DIFF
--- a/brands/shop/bakery.json
+++ b/brands/shop/bakery.json
@@ -854,6 +854,14 @@
       "shop": "bakery"
     }
   },
+  "shop/bakery|Wenzel's": {
+    "locationSet": {"include": ["gb"]},
+    "tags": {
+      "brand": "Wenzel's",
+      "name": "Wenzel's",
+      "shop": "bakery"
+    }
+  },
   "shop/bakery|Хлеб Насущный": {
     "locationSet": {"include": ["001"]},
     "tags": {


bhousel:
Thanks @peternewman 👍 
It is ok to add this to the index, but it won't be available as a preset in the OSM editors until it has a `brand:wikidata` tag assigned.

peternewman:
> Thanks @peternewman +1
> It is ok to add this to the index, but it won't be available as a preset in the OSM editors until it has a `brand:wikidata` tag assigned.

Ah okay thanks, I didn't know.

Okay I'll maybe look at sorting that later.

bhousel:
> Ah okay thanks, I didn't know.

It’s ok! I did a quick search but didn’t find an existing page. 

There are some people here who are good at writing Wikidata and Wikipedia pages, so maybe someone will do it.

peternewman:
> > Ah okay thanks, I didn't know.
> 
> It’s ok! I did a quick search but didn’t find an existing page.

I was also going to ask about thresholds for inclusion, but this looks like it's 50 entries normally:
https://github.com/osmlab/name-suggestion-index/blob/27079c7bc147c01f86c95687c7d25cc483521dc6/scripts/build_all_names.js#L27

> There are some people here who are good at writing Wikidata and Wikipedia pages, so maybe someone will do it.

Excellent.

PR #3998, opened 2020-06-22
Add Milkshop

tntchn:
https://zh.wikipedia.org/wiki/%E8%BF%B7%E5%AE%A2%E5%A4%8F
迷客夏 Milkshop is a big beverage chain in Taiwan.

### DIFF
--- a/brands/shop/beverages.json
+++ b/brands/shop/beverages.json
@@ -155,5 +155,19 @@
       "name": "茶湯會",
       "shop": "beverages"
     }
+  },
+  "shop/beverages|迷客夏": {
+    "locationSet": {"include": ["tw"]},
+    "tags": {
+      "brand": "迷客夏",
+      "brand:en": "Milkshop",
+      "brand:wikidata": "Q67934215",
+      "brand:wikipedia": "zh:迷客夏",
+      "brand:zh": "迷客夏",
+      "name": "迷客夏",
+      "name:en": "Milkshop",
+      "name:zh": "迷客夏",
+      "shop": "beverages"
+    }
   }
 }


bhousel:
Thanks @tntchn and welcome! 👋 

I have noticed that other bubble tea shops in our index (e.g. Chatime, CoCo, Gong Cha) are tagged under `amenity=cafe` with `cuisine=bubble_tea`.  Is this brand like those?

`shop=beverages` tends to be more for places that sell cans and bottles, not really that make drinks to order for customers.

tntchn:
Thanks @bhousel 

I did some data collecting and found that it might been a large scale mismapping in the map. In fact, the density of shop=beverages in Taiwan is much more than any other place.
Screenshot_2020-06-26 overpass turbo
Most of them are bubble tea/juice shop but not sell bottles. Use amenity=cafe + cuisine=bubble_tea + **takeaway=only** may disturbing mapper in Taiwan because bubble tea shop is very different from a cafe: no coffee, no snack, no seat, only tea and juice. Like the picture below is a typical bubble tea shop looks like.
IMG_4238

According to the statistic there are more than 20,000 bubble tea shop in Taiwan, many of the brand also open some shops in Australia, US, China, etc. I think is deserved a feature proposal to have a independence tag in OSM.

Adamant36:
> Most of them are bubble tea/juice shop but not sell bottles. Use amenity=cafe + cuisine=bubble_tea + takeaway=only may disturbing mapper in Taiwan because bubble tea shop is very different from a cafe: no coffee, no snack, no seat, only tea and juice. Like the picture below is a typical bubble tea shop looks like.

What about just amenity=fast_food + cuisine=bubble_tea + takeaway=only then? It doesn't seem like something that needs a unique tag and having one would just lead to tag fragmentation anyway.

bhousel:
> Use amenity=cafe + cuisine=bubble_tea + **takeaway=only** may disturbing mapper in Taiwan because bubble tea shop is very different from a cafe: no coffee, no snack, no seat, only tea and juice. Like the picture below is a typical bubble tea shop looks like.

ok, I'm fine accepting the PR as-is.   I don't want to disturb the mappers in Taiwan.

BTW, these tags aren't as important as people think they are.  If the Taiwan community later decides they want to retag all of the Milkshops to cafes to make them like Chatime, CoCo, and Gong Cha, or if a new "bubble tea" tag emerges out of consensus, then having the `brand:wikidata` tag on the object makes updating them really easy.

PR #3999, opened 2020-06-22
Add fitness centres in Taiwan

tntchn:
https://en.wikipedia.org/wiki/World_Gym
World Gym is a global fitness center chain and have more than 200 branches around the world.
But it has more than 80 branches in Taiwan, being the largest fitness center brand in Taiwan. So I want to separate them into two tagging methods just like the commit.
Note: 「世界健身俱樂部」 is the Chinese name but not a common used name, so I put it in the alt_name tag just for searching.

健身工廠 Fitness Factory is the 2nd large fitness center chain in Taiwan, will have more than 50 branches before the end of 2020, but still have no Wikipedia and Wikidata item now.

### DIFF
--- a/brands/leisure/fitness_centre.json
+++ b/brands/leisure/fitness_centre.json
@@ -359,6 +359,30 @@
       "name": "World Class"
     }
   },
+  "leisure/fitness_centre|World Gym": {
+    "locationSet": {"include": ["001"], "exclude": ["tw"]},
+    "tags": {
+      "brand": "World Gym",
+      "brand:wikidata": "Q8035794",
+      "brand:wikipedia": "en:World Gym",
+      "leisure": "fitness_centre",
+      "name": "World Gym"
+    }
+  },
+  "leisure/fitness_centre|World Gym~(Taiwan)": {
+    "locationSet": {"include": ["tw"]},
+    "tags": {
+      "brand": "World Gym健身俱樂部",
+      "brand:en": "World Gym",
+      "brand:zh": "世界健身俱樂部",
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @tntchn  - can you remove the `website` tags?  We actually don't add them here because:

1. anyone who wants it can get it from the brand's wikidata entry
2. the `website` tag is more accurately used for the specific website of that mapped location, not the overall brand.

Adamant36:
It seems weird to have a separate entry for World Gym when they are still called "World Gym" in Taiwan. The only reason for the extra entry here seems to be so there can be a link to the zh Wikipedia article. Which people can just find by using the Wikidata link. So, I IMO there shouldn't be a separate entry for it in this case.  Especially since the Taiwanese logos for it still says "World Gym in cases where they also use 世界健身俱樂部. The same might go for Fitness Factory, but at least 健身工廠 is regularly used in that case.

Adamant36:
Or at least use 世界健身俱樂部 in the name:zh entry and get of alt_name so people don't complain later that we used the name:zh tag for English or something.

tntchn:
Thanks @bhousel I removed it and thanks @Adamant36 Yeah I think that put English name in name:zh is weird. I modified the name into "World Gym健身俱樂部" which is more suitable that also appears in their Facebook page (https://www.facebook.com/worldgymtaiwan/) and let the name:zh become full-Chinese name "世界健身俱樂部". That makes users and editors less confusing.

bhousel:
Thanks @tntchn  👍

PR #4003, opened 2020-06-27
Add supermarket in Malaysia

MASangys:


Adamant36:
Hi. Did you mean to add 99 Speedmart again here?

MASangys:
No but in variety store

Adamant36:
You want to delete it then since its already in another PR and the index?

MASangys:
99 speedmart is in convenience store category, this one is another

Adamant36:
Yeah but your already adding it in the other PR. So its not needed in this one also. That one should take care of it.

MASangys:
This one is for another shop which is The Store and TF Value Mart tho

Adamant36:
I get that. It doesn't have anything to do with The Store and TF Value Mart. My point is that you have code in this PR to add 99 Speedmart to the variety store file. Which your already doing in PR #4004. So, do you want to add it in #4004 or this PR? because it can't be added in both and you told me to re-open #4004. So, I'm assuming that's how you want to do it. Which would make the code for it here redundant.

MASangys:
Both are the same right, I think there is an issue occur when I do PR

MASangys:
So just merge 1 of them and another need to be clesed

bhousel:
Thanks @MASangys , looks good to me ..

PR #4005, opened 2020-06-28
Add brands: Xiaomi, Vélez, Juan Valdez Café

desiderantes:


brands/amenity/cafe.json
```diff
@@ -761,6 +761,23 @@
...
   },
+  "amenity/cafe|Juan Valdez Café": {
+    "locationSet": {"include": ["001"]},
+    "tags": {
+      "amenity": "cafe",
+      "brand": "Juan Valdez Café",
+      "brand:wikidata": "Q6301102",
+      "brand:wikipedia": "en:Juan Valdez Café",
+      "cuisine": "coffee_shop;empanada;biscuit",
+      "internet_access": "wlan",
```
matkoniecz, line 772: Are you sure that it is true for every single cafe using this brand, worldwide?

desiderantes: Yup. Can find the url for the franchise requirements right now but, to the best of my knowledge (because I worked at one), they DO require items on the menu, services, wifi, and outdoor seating (unless not possible), even worldwide (recipes for empanadas are provided)

bhousel: I don't see anything about empanadas on their website.  It just looks like a normal coffeehouse to me, and it should probably have similar tags to Starbucks.

https://www.juanvaldezcafe.com/en-us/menu/

desiderantes: I see one as a thumbnail here https://www.juanvaldezcafe.com/en-us/snacks/

Adamant36: If you click on the image of the empanadas it's not including in the list of salty snakes. Whereas, if you click on the picture of chocolate cake it is listed with the other sweet treats. Which makes me think they don't serve empanadas despite the image. 

Anyway, it veers into "every blade of grass" mapping to include stuff like that as it is. Since it's not their main/secondary/or even third thing if they even do serve it. Otherwise, why not also list cinnamon rolls, chocolate cake, brownies, muffins, etc etc? If they do serve them, there's nothing particularly special about empanadas compared to anything else on the menu. Same goes for biscuits. So, both should be left out IMO. Otherwise, people will think it's an "empanadas and biscuit cafe." Which it isn't.

desiderantes: OK, updated the cuisine tag

Adamant36:
I'm a little sketchy on the outdoor=yes tag being used for Jaun Valdez Cafe. Since it appears some of them don't have outdoor seating. Which also makes me sketchy on the other tags like internet_access that would be treating exist in every case by merging this when they probably don't. IMO both should be removed and people can just add them on their own if they want to. I rather not risk introducing errors though. 

Juan Valdez Cafe

bhousel:
This has been open for a while.. I'm going to just merge and remove the tags that don't belong.

PR #4012, opened 2020-07-02
Update fuel.json

luisluigi639:
Adding in Loaf 'n Jug gas along with the addition of Wikidata to Liberty gas.


luisluigi639:
It seems like that line of code has already been added. It's on line 960: https://github.com/osmlab/name-suggestion-index/blob/master/brands/shop/convenience.json#L960

luisluigi639:
Ok, I have added the line of code to pull request #4021

PR #4014, opened 2020-07-03
Create animal_theme_park.json

Mrskillful:
Created an animal theme park file to store leisure activities that revolve around animal theme parks, e.g. SeaWorld.

### DIFF
--- a/brands/leisure/animal_theme_park.json
+++ b/brands/leisure/animal_theme_park.json
@@ -1,0 +1,24 @@
+{
+  "leisure/animal_theme_park|Busch Gardens": {
+    "locationSet": {"include": ["us"]},
+    "matchTags": ["tourism/attraction"],
+    "tags": {
+      "brand": "SeaWorld Parks & Entertainment",
+      "brand:wikidata": "Q1017288",
+      "brand:wikipedia": "en:Busch Gardens",
+      "leisure": "animal_theme_park",
+      "name": "Busch Gardens"
+		}
+   },
+  "leisure/animal_theme_park|SeaWorld": {
+    "locationSet": {"include": ["us"]},
+    "matchTags": ["tourism/attraction"],
+    "tags": {
+      "brand": "SeaWorld Parks & Entertainment",
+      "brand:wikidata": "Q540102",
+      "brand:wikipedia": "en:SeaWorld",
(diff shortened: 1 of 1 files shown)


bhousel:
Welcome @Mrskillful 👋 
Thanks for this pull request, however I don't think it really fits in with what this project is for.

While there are a handful of Busch Gardens and SeaWorlds in the world, I don't think there are enough of them that they should be included in an index of brands like this one.  All the other entries in this index are for branded business chains that occur frequently.  We use "around 50" as a rough threshold, but people have added smaller brands that have only a few dozen locations. 

For major theme parks like Busch Gardens or SeaWorld, mappers can manually set the tags in OSM for these things, adding `wikidata` and `brand:wikidata` as they see fit.

Mrskillful:
Ok, I'm assuming something like 50 either nationally or internationally correct? 

And if so, where would something like Disney Resorts/Parks fall into?

bhousel:
> So where would something like Disney Resorts/Parks fall into?

Good question - here's an example of a park in OSM (Disney's Animal Kingdom):
https://www.openstreetmap.org/way/297441457#map=16/28.3608/-81.5908

Theme parks like these are notable enough that they should have their own Wikidata/Wikipedia tags.  I see this particular park has these tags:
`wikidata=Q764186`
`wikipedia=en:Disney's Animal Kingdom`

I do see these tags too:
`operator=Disney Parks and Resorts`
`brand=Disney Parks`

I think if mappers want to add `operator:wikidata`/`operator:wikipedia` and `brand:wikidata`/`brand:wikipedia` tags linking to whatever pages on Wikidata describe these entities within the Disney organization, that would help too.

This name-suggestion-index project is not really for cataloging specific notable features like theme parks or universities (both Wikipedia and OpenStreetMap projects already do a great job of that).  We are really a list of branded businesses to help OpenStreetMap users tag them consistently.  (For example, Disney Stores [are in our index](https://nsi.guide/index.html?k=shop&v=gift#Disney%20Store) - we know of 63 of them so far, and they can be found in malls everywhere).

Mrskillful:
Ok, now I understand, so the name-suggestion-index project is to more prevalent brand-locations than those brand-locations that are few, and where users can just update themselves. Thanks for clarifying.

Issue #4009, opened 2020-06-30
Add brand: WIND mobile phone stores in Greece


sucrerouge123:
Hi!

I would like to request the addition of Wind mobile telecommunications chain stores in Greece.

shop:mobile_phone
name:WIND
brand:WIND Hellas
brand:wikidata:Q10711016
brand:wikipedia=el:WIND Hellas
country codes:gr

Thanks!

sucrerouge123:
Please add the tag:

website=https://www.wind.gr

Thanks!

bhousel:
> Please add the tag:
> 
> website=https://www.wind.gr

Thanks for the suggestion but we don't actually add `website` tags in this index.. 

OSM convention is that:
- tags like `website` and `contact:website` should be for a website for _that specific POI's branch location_, and
- tags like `brand:website` or `operator:website` are for the overall brand or operator - however these tags have fallen out of favor because this information can just be fetched more accurately from that brand or operator's Wikidata page.

sucrerouge123:
Thanks!

PR #4016, opened 2020-07-04
Add Wind Hellas (Greece).

darkkrol16:
This request closes issue https://github.com/osmlab/name-suggestion-index/issues/4009

Mobile shop Wind (Italy) is already presented, so I use the whole name.
image


tas50:
Thanks @darkkrol16

Issue #4007, opened 2020-06-29
FR: change wikipage for Pathé Gaumont cinemas


overflorian:
Hi,
The wikipedia page of Pathé Gaumont has changed. It's now https://fr.wikipedia.org/wiki/Les_cin%C3%A9mas_Path%C3%A9_Gaumont

My request is about an existing cinema chain: https://nsi.guide/index.html?k=amenity&v=cinema#Path%C3%A9%20Gaumont
Could you please change it for:
"amenity": "cinema"
"brand": "Pathé Gaumont"
"brand:wikidata": "Q3060526"
**"brand:website": "https://www.cinemaspathegaumont.com"**
**"brand:wikipedia": "fr:Les cinémas Pathé Gaumont"**
"name": "Pathé Gaumont"

Thanks

PR #4017, opened 2020-07-04
change wikipage pathe

sandygudie:
This request closes the issue https://github.com/osmlab/name-suggestion-index/issues/4007


bhousel:
Thanks, but please remove the `brand:website` tag. Many mappers think that tag is redundant and unstable, so we encourage anyone who needs the current website to fetch it from the brands Wikidata page.

tas50:
Thanks @sandygudie

Issue #3712, opened 2020-04-14
Remove entries with generic names


kjonosm:
Please remove/discard following entries as they tend to be in the index just because of their generic name:

- 'amenity/dentist|Zahnarztpraxis' (generic German name for dentist practice)
- 'amenity/pharmacy|Einhorn-Apotheke'
- 'amenity/vending_machine|Schlauchomat'
- 'shop/hairdresser|Hair Studio'
- 'tourism/hotel|Hotel Belvedere' (maybe someone from Italy can confirm?)
- 'tourism/hotel|Hôtel de la Gare' (maybe someone from France can confirm?)

Thanks!

sandygudie:
I think all the generic names above have been removed

matkoniecz:
No, for example https://github.com/osmlab/name-suggestion-index/search?q=Hair+Studio&unscoped_q=Hair+Studio

sandygudie:
Hi there, how about the last two?

matkoniecz:
Hotel Belvedere researched and removed as nonbrand name

bhousel:
I've done "Hotel de la Gare" - that should take care of all of them 👍

PR #4018, opened 2020-07-04
remove generic hair studio

sandygudie:
This request is part of the issue https://github.com/osmlab/name-suggestion-index/issues/3712


tas50:
Thanks @sandygudie

PR #4019, opened 2020-07-06
generic name removed

sandygudie:
Rremoved generic names Schlauchomat and Einhorn-Apotheke https://github.com/osmlab/name-suggestion-index/issues/3712


tas50:
Thanks @sandygudie

PR #4023, opened 2020-07-10
Central clean up

Cj-Malone:
I've done a few things, changed it to GB only, I think it's mainly the south of England, but I doubt you get that specific.

Split it into a newsagent and a convenience store, they also have a 3rd store format, but for tagging purposes it's the same as the convenience preset.

And I've changed the name to "Central", this matches the shop fascia.


bhousel:
Thanks @Cj-Malone , looks good 👍

PR #4024, opened 2020-07-10
Added Movado

Mrskillful:
Added Movado to watches.json

### DIFF
--- a/brands/shop/watches.json
+++ b/brands/shop/watches.json
@@ -22,6 +22,18 @@
       "shop": "watches"
     }
   },
+  "shop/watches|Movado":{
+	"locationSet": {"include": ["001"]},
+	"matchNames": ["モバード", "武嘩度"],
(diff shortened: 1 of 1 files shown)


Adamant36:
Hi there. Thanks for the PR and welcome to the project. Everything looks fine with it except for one thing. I don't think the match names are needed because the brand is just known as "Movado" everywhere, there isn't any usage of those two names in OSM for the entry to match with as it is, and we don't add match names based on what languages Wikipedia articles are translated into anyway.  Which I'm assuming is where you got the idea to add them from. So, if you want to remove them I can merge the PR for you 👍

Mrskillful:
Thanks for the clarification. I'll make the changes.

Adamant36:
> Thanks for the clarification. I'll make the changes. 

Totally. Thanks for making the changes.

PR #4025, opened 2020-07-11
Fixed the fuel brand OKQ8.

Kezxo:
OKQ8 is not the same as Kuwait Petroleum Corporation, they only own 50% of it but even still it's a separate brand. This is the correct brand.
I hope I did this correctly.


bhousel:
Looks good to me, @Kezxo 
Thanks and welcome to the project 👋

PR #4026, opened 2020-07-11
Add Sparkasse - Cassa di Risparmio di Bolzano Bank (Italy)

abonte:
Add Italian bank.


Adamant36:
Sorry, but as it currently is we're kinda not adding Sparkasse banks to the index because the mapping communities where they are located don't want us to. So, I think the best thing to do would be to discuss it in the DE or IT mailing lists first, or to just close it as won't fix.

PR #4028, opened 2020-07-11
Update fast_food.json

max-dn:
aixbrick found a typo (https://forum.openstreetmap.org/viewtopic.php?id=69996) immergün->immergrün

### DIFF
--- a/brands/amenity/fast_food.json
+++ b/brands/amenity/fast_food.json
@@ -3589,13 +3589,13 @@
   },
   "amenity/fast_food|immergrün": {
     "locationSet": {"include": ["de"]},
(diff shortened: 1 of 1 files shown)


Adamant36:
How come aixbrick didn't open an issue since they seem to know about the repository? Also, do you know if the original spelling is used anywhere? It might have been added that way because of usage at the time and if so it would help have a match name for it. I do wonder why currently the match name and name are the same.

bhousel:
Closing as #4029 did this.
`name` should contain the correctly spelled name, `matchNames` can contain the common misspellings or other less desired names.

PR #4032, opened 2020-07-15
Add Oppo

hanchao:
Oppo is a Chinese consumer electronics and mobile communications company headquartered in Dongguan, Guangdong. Since then, they have expanded to all parts of the world.
img
https://www.wikidata.org/wiki/Q2084900
https://en.wikipedia.org/wiki/Oppo


bhousel:
Thanks @hanchao looks good 👍

PR #4033, opened 2020-07-16
Add Chowtaifook

hanchao:
Chow Tai Fook Jewellery is a jewellery store and gold shop founded by Chow Chi-yuen in 1929 in Guangzhou, China. 
image
https://www.wikidata.org/wiki/Q48880993
https://en.wikipedia.org/wiki/Chow_Tai_Fook


bhousel:
Thanks @hanchao 👍

PR #4042, opened 2020-07-22
Add Courts in Malaysia

MASangys:
https://en.wikipedia.org/wiki/Courts_Malaysia
https://www.courts.com.my/

### DIFF
--- a/brands/shop/furniture.json
+++ b/brands/shop/furniture.json
@@ -170,6 +170,18 @@
       "shop": "furniture"
     }
   },
+  "shop/furniture|  "Courts Malaysia": {
+    "locationSet": {
+      "include": ["my"]
+    },
+    "tags": {
+      "brand": "Courts Malaysia",
+      "brand:wikidata": "Q5178707",
+      "brand:wikipedia": "en:Courts Malaysia",
+      "name": "Courts",
+      "shop": "furniture"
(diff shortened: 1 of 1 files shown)


brands/shop/furniture.json
```diff
@@ -170,6 +170,18 @@
       "shop": "furniture"
     }
   },
+  "shop/furniture|  "Courts Malaysia": {
```
bhousel, line 173: Whoa looks like an extra quote here is messing things up.

```suggestion
  "shop/furniture|Courts Malaysia": {
```

bhousel:
Looks good now

MASangys:
However, there are another Courts in Singapore woth different Wikidata, so I added this as Courts Malaysia to differenciate between them

bhousel:
> However, there are another Courts in Singapore woth different Wikidata, so I added this as Courts Malaysia to differenciate between them

oh weird - it probably should just be named "Courts"

MASangys:
Then we should ask the editor at Wikipedia. because they do it,  I just follow their rules,  but commonly this tag for Courts should be apply for multi countries also.

bhousel:
For now I just added 2 entries, one for Singapore and one for Malaysia - c98fc281

PR #4045, opened 2020-07-24
Add Home Inn

hanchao:
Home Inn is the largest budget hotel chain in China
image

https://www.wikidata.org/wiki/Q4980954
https://en.wikipedia.org/wiki/Home_Inn


tas50:
Thanks @hanchao

PR #4047, opened 2020-07-25
Fix Беларусбанк duplicate

danjamesyee:
Some Беларусбанк locations are just a bank and others are just a bureau de change. I added the nomatch tags for each entry. I hope its correct

### DIFF
--- a/brands/amenity/bureau_de_change.json
+++ b/brands/amenity/bureau_de_change.json
@@ -60,7 +60,7 @@
     }
   },
   "amenity/bureau_de_change|Беларусбанк": {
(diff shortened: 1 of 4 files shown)


brands/amenity/bureau_de_change.json
```diff
@@ -61,6 +61,7 @@
   },
   "amenity/bureau_de_change|Беларусбанк": {
     "locationSet": {"include": ["001"]},
```
tas50, line 63: Can you add the specific location for this?

danjamesyee: Ok just added this.

PR #4051, opened 2020-07-27
Tesoro duplicate fix

danjamesyee:
There is a duplicate for Tesoro, but I could only find evidence of Tesoro being an oil and gas company which would be under fuel. Tesoro bank appears to be a completely separate entity if it exists at all.

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -6007,6 +6007,15 @@
       "name": "Taytay sa Kauswagan"
     }
   },
+  "amenity/bank|Tesoro": {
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks for adding this 👍 

From overpass, it looks like the Tesoro bank is a regional bank around Caracas:
Screenshot 2020-07-28 13 17 20

Some of them are tagged really well, so we can probably dig up more information about it.

bhousel:
I did a bit of research and was able to add the wiki tags for the Tesoro bank in c3a6ddc

Issue #4054, opened 2020-07-31
Incorrect name:ua tag instead of name:uk


Andygol:
It was found that users got incorrect suggestion of `name:ua` key for names in Ukrainian instead of using correct `name:uk=*` key for some POIs in Ukraine. 

Places where `name:ua=*` exists is possible to see here - https://github.com/osmlab/name-suggestion-index/search?q=%22name%3Aua%22&unscoped_q=%22name%3Aua%22

It would be great to fix this issue and add some tests to avoid using `name:ua` key

PS
**uk** - Ukrainian language (https://en.wikipedia.org/wiki/Ukrainian_language)
**ua** - Ukraine (https://en.wikipedia.org/wiki/ISO_3166-2:UA)

PR #4055, opened 2020-07-31
Fix of incorrect lang code for Ukrainian

Andygol:
re #4054 

This PR fixes incorrect key `name:ua` for names in Ukrainian to correct one - `name:uk`


bhousel:
Thanks @Andygol  👍

PR #4062, opened 2020-08-10
Added gym

VJ1224:
Wikipedia link: https://en.wikipedia.org/wiki/Talwalkars
Wikidata link: https://www.wikidata.org/wiki/Q7680402


tas50:
Thanks @VJ1224

Issue #4037, opened 2020-07-18
Add Biki


TheAdventurer64:
Biki is a bike share rental system chain in Honolulu. Wikipedia page not available.

PR #4067, opened 2020-08-13
Add Biki

Identitaet:
I also made the Wikidata
solves #4037 
(as I side note thanks for welcoming me but I just deleted my old accound and took a break but now I'm back)


bhousel:
Cool, thanks again and welcome back.. 👍 
I just welcome everyone when it says they are a "first time contributor".

PR #4071, opened 2020-08-13
Panda Express is only active in the US

thibaultmol:
There are 'Panda Express' restaurants in Europe (UK, NL, FR,...) but these are all just similary named. Not related to the US brand


tas50:
Thanks @thibaultmol

PR #4073, opened 2020-08-14
PizzaExpress only active in these countries

thibaultmol:
UK + these ones: https://www.pizzaexpress.com/about-us/new-international-restaurants


bhousel:
Thanks @thibaultmol 👍

Issue #4036, opened 2020-07-18
Add Bedmart


TheAdventurer64:
Zippy's is a fast food/restaurant chain. Wikipedia page here: 
https://en.wikipedia.org/wiki/Zippy%27s

Aloha Island Mart is a convenience store for Aloha Petroleum. Wikipedia page here: https://en.wikipedia.org/wiki/Aloha_Petroleum

Costco Food Court is a food court for Costco. Wikipedia page here: https://en.wikipedia.org/wiki/Costco#Food_service

Costco Optician is a optician for Costco. Wikipedia page here:
https://en.wikipedia.org/wiki/Costco#Costco_Optical

Times Supermarket is a supermarket chain. Wikipedia page here:
https://en.wikipedia.org/wiki/Times_Supermarkets

Homeworld is a furniture store chain. Wikipedia page not available. 

Slumberworld is a mattress store chain. Wikipedia page not available. 

Hele is a gas station chain. Wikipedia page not available.

Down to Earth is a health food store chain. Wikipedia page not available.

Bedmart is a furniture store chain. Wikipedia page not available.

mds08011:
Times Supermarket added here: https://github.com/osmlab/name-suggestion-index/commit/ed6abfc0f9ff08bce1d7bf2ba1f4ce3afb5be185

Identitaet:
Zippy's now has all the things needed for NSI in wikidata

VJ1224:
TODO:

- [ ] Zippy's is a fast food/restaurant chain. Wikipedia page here:
https://en.wikipedia.org/wiki/Zippy%27s

- [ ] Aloha Island Mart is a convenience store for Aloha Petroleum. Wikipedia page here: https://en.wikipedia.org/wiki/Aloha_Petroleum

- [ ] Costco Food Court is a food court for Costco. Wikipedia page here: https://en.wikipedia.org/wiki/Costco#Food_service

- [ ] Costco Optician is a optician for Costco. Wikipedia page here:
https://en.wikipedia.org/wiki/Costco#Costco_Optical

- [ ] Times Supermarket is a supermarket chain. Wikipedia page here:
https://en.wikipedia.org/wiki/Times_Supermarkets

- [ ] Homeworld is a furniture store chain. Wikipedia page not available.

- [ ] Slumberworld is a mattress store chain. Wikipedia page not available.

- [ ] Hele is a gas station chain. Wikipedia page not available.

[...]

PR #4078, opened 2020-08-17
Added Zippys #4036

VJ1224:
Wikipedia link: https://en.wikipedia.org/wiki/Zippy%27s

Wikidata link: https://www.wikidata.org/wiki/Q8072671


bhousel:
thanks @VJ1224 👍

Issue #4034, opened 2020-07-16
Is name required?


Cj-Malone:
Hello,

Harvester (amenity/restaurant|Harvester) is currently listed in NSI, however each instance should have a different name tag in OSM. The [website](https://www.harvester.co.uk/restaurants) demonstrates this.

Is it possible to remove `"name": "Harvester"` from NSI, or is that needed else where in this project?

bhousel:
Yes #3009 has some background but for now a `name` is required.  If it's causing problems, we could just remove Harvester from the index for now - let me know if you want this and I'll remove it.

At some point we want to relax this requirement, it's on my list of things to consider for #3995

Cj-Malone:
I've not witnessed any data being lost due to it, so I wouldn't remove it just yet.

Is there anything I can help with? I'm not a node developer, but I do have a basic understanding of js.

bhousel:
Name isn't required anymore, so I removed the `name` tag from Harvester in f0b72129

Issue #3995, opened 2020-06-19
Rethinking identifiers


bhousel:
I'd like to make some changes in the index to better support more kinds of named POIs and more granular locations.  This is just a brain dump of some thoughts on how we track brands and the limitations with our current approach.

### a long time ago

When this project started many years ago, each entry in NSI was just a unique name that we picked out of the OSM planet file.
For example, here's what NSI looked like in 2017:
https://github.com/osmlab/name-suggestion-index/blob/50993d4e7e8885a3b2312a1a633514b9a33db881/name-suggestions.json#L8621-L8623
This says that Target is an entry that with the name tag set to `Target`, and it has been used about 1000 times and sits under the `shop/department_store` hierarchy.  iD would turn this into a preset that assigns the tags `name=Target` and `shop=department_store`.

Limitations of this:
- couldn't track different things named "Target" (the USA and Australia ones are different)
- couldn't add more OSM tags like `brand:wikidata`
- support for alt versions of the tag and name was just - adding them to ignore lists
- no `countryCodes` or any concept of where the brand was valid
- all the names were in a single file that was starting to get too big

So we did a lot of work on the NSI in 2018-2019 to arrive at the current format.
[I did a talk about it!](https://2019.stateofthemap.us/program/sat/mapping-brands-with-the-name-suggestion-index.html)

### today

Each entry in NSI represents a unique "brand" identified by an string like:
`key/value|name~(disambiguator)`
The `disambiguator` part is optional and used in situations where distinct entities use the same literal name.

https://github.com/osmlab/name-suggestion-index/blob/50278e5b7ff226cea96d0bcecdc15442ec0b5c01/brands/shop/department_store.json#L730-L758

We solved the limitations from long ago, and NSI has really grown!
- we can track any number of things with the same name (by using disambiguators)
- we can attach whatever OSM tags we want to each entry
- we can link each entry to `brand:wikidata` and fetch a wealth of related data from the Wikidata project (like logos)
- we can support brands with alternate tagging and naming
- we used `countryCodes` for a while, now `locationSet` which is even more flexible
- we wrote code to fuzzy match the names and their alternates, and exported that code into a library for iD's validator to use
- we made https://nsi.guide to browse it all, which is pretty rad

But now we have a new set of limitations:
- we're limited to using `brand` and `brand:wikidata` as the "key" even though other feature types might better be keyed off of `operator` or `network`
[...]

Issue #3009, opened 2019-09-17
Add Wetherspoons Pub


doublah:
Brand Name
Wetherspoons (also referred to as J D Wetherspoon, JD Wetherspoon, Wetherspoon in current tags)
amenity=pub

Country Codes:
gb, ie

https://www.jdwetherspoon.com/pubs/all-pubs
Pretty much all Wetherspoons are seperately named pubs with wetherspoon branding (going by their website, only 9 out of the almost 900 pubs have "Wetherspoon(s)" in the name).
This means most current tagging is very inconsistent with most currently using various operator= tags (see https://taginfo.openstreetmap.org/search?q=Wetherspoon#values and https://taginfo.openstreetmap.org/search?q=Wetherspoons#values )

Wiki:
https://en.wikipedia.org/wiki/Wetherspoons
https://www.wikidata.org/wiki/Q6109362

tohaklim:
Wondering if this might work with an operator matchnames string.
This would also need to preserve the current name tag

bhousel:
This was done in #3103 🍻

doublah:
Would still be nice if we had a way to matchnames for operators, plus needs an ie country code.

bhousel:
> Would still be nice if we had a way to matchnames for operators, plus needs an ie country code.

Good catch, I added the `ie` code in 21aec5792

We don't include operator tags in this index because they have historically been misused, but we will probably add something eventually.
(see #2816, but we are considering relaxing this rule per #2928)

doublah:
Reopening as was undone in b9c20873f

Probably need to think about a better way to deal with brands with obvious branding like this but with unique location names.

UKChris-osm:
I'm not sure how well linked the NSI is with the iD Editor, but for a brand like Wetherspoon, could the matching occur if someone were to add a node or way and search for the tag "JD Wetherspoon", and have the name element itself appear within iD (like a preset) yet remain empty, and hence invite the mapper to add the name?

 "amenity/pub|JD Wetherspoon": {
    "countryCodes": ["gb"],
    "matchNames": [ "JD Wetherspoons", "Wetherspoons", "Wetherspoon"],
    "tags": {
      "amenity": "pub",
      "brand": "JD Wetherspoon",
      "brand:wikidata": "Q6109362",
      "brand:wikipedia": "en:Wetherspoon",
      "name": **REQUIRED**
    }
  }

[...]

PR #4086, opened 2020-08-18
Add Quirónsalud hospital chain

Robot8A:
Without name, as every hospital may have their own name, but they are part of the same chain (it has 52 hospitals in Spain).


bhousel:
Ah sorry, we can't currently accept entries that don't have a `name` tag.
See #3009, #3995, #4034 for some more context.

PR #4087, opened 2020-08-18
Change Tesco Express wikidata item to a more specific entry

rjw62:
Other Tesco sub-brands (Tesco Extra, Tesco Metro) have their own Wikidata entries but Tesco Express was missing. I've created a Wikidata item for it at https://www.wikidata.org/wiki/Q98456772 , and would now like to update the NSI to use it.

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -2042,7 +2042,7 @@
     ],
     "tags": {
       "brand": "Tesco Express",
-      "brand:wikidata": "Q487494",
+      "brand:wikidata": "Q98456772",
       "brand:wikipedia": "en:Tesco",
       "name": "Tesco Express",
       "shop": "convenience"
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -2043,7 +2043,6 @@
     "tags": {
       "brand": "Tesco Express",
       "brand:wikidata": "Q98456772",
-      "brand:wikipedia": "en:Tesco",
       "name": "Tesco Express",
       "shop": "convenience"
     }


ImreSamu:
As I see this ( https://www.wikidata.org/wiki/Q98456772 )  wikidata only for the "UK" 
*  _"Brand for smaller convenience-style Tesco food stores in the UK."_


imho it is a global brand .. not only for the UK 
* https://github.com/osmlab/name-suggestion-index/issues/2740

bhousel:
Thanks @rjw62 & welcome to the project 👋 

I'm ok with adding a more specific `brand:wikidata` identifier for Tesco Express, but you should also remove the `brand:wikipedia` tag, as it points to the general Tesco wikipedia page.  (we expect these to match)

Also double check the Wikidata page to make sure the info there is accurate, re: what @ImreSamu said.

rjw62:
I've fixed the wikidata page -- there are Tesco Expresses in other countries too.

I'm not sure what you want to do about the Wikipedia link. https://en.wikipedia.org/wiki/Tesco_Express is just a redirect to a section of the https://en.wikipedia.org/wiki/Tesco page. Should we use the redirect (which is what the NSI entry for Tesco Extra does) that, stick with en:Tesco, or just omit the Wikipedia link altogether (which is what the NSI entry for Tesco Metro does)?

bhousel:
> I'm not sure what you want to do about the Wikipedia link.

I’d omit the `brand:wikipedia` tag for now.. Or ideally, if you are comfortable working with Wikipedia, take those sections about Tesco Express and Tesco Extra and expand them into their own Wikipedia pages, and have the `brand:wikipedia` tags point to those.

ImreSamu:
Another important issue (for me)  ... the brand logo :smile:   !

as I see - the current version ( "Tesco Express" )  has 2 logo 
imho -  we need to copy the **logo info** from [Q487494 ](https://www.wikidata.org/wiki/Q487494)  to : [Q98456772](https://www.wikidata.org/wiki/Q98456772)
- Logo image (P154)
- Twitter username  (P2002)

## now: 

https://nsi.guide/index.html?k=shop&v=convenience#Tesco%20Express
image

ImreSamu:
The logo issue - has been solved 
*  I have added the logo+twitter info to https://www.wikidata.org/wiki/Q98456772

SilentSpike:
Just as a Wikidata point (not something holding up this PR), a company should probably not have multiple P31 statements, this should really be reserved for where it is incorporated. The regions it operates in would be better captured in P2541.

That said, I think a lot of the Wikidata modelling for these sub-brand chains is all wrong anyway (they're probably not companies in their own right).

bhousel:
Thanks, sorry this got delayed!
Looks good now 👍

PR #4090, opened 2020-08-19
Add GAES, Spanish hearing aid store chain

Robot8A:


matkoniecz:
I see some cases in other countries, is it possible that coverage should be expanded? https://overpass-turbo.eu/s/Xdk

Robot8A:
Added the countries stated in the company website

PR #4093, opened 2020-08-19
Add MásQMenos, Spanish "Ham & Tapas" chain

Robot8A:


matkoniecz:
Just to confirm:

So mapping as restaurants and cafes is always wrong and should be changed to pub?

See https://www.openstreetmap.org/search?query=M%C3%A1sQMenos#map=19/41.36088/2.12964&layers=N

Robot8A:
In my own view:
- It's not a restaurant as they don't serve 'proper food plates', just tapas, sandwiches and small portions that you can eat for example during a break, brunch or similar
- It's not a cafe as its main focus is not serving coffee

In the Spanish community we already discussed the topic about how to tag typical Spanish bars, were small food (tapas) can be eaten alongside drinks (including alcoholic ones: usually beer or wine), and we decided that it was amenity=pub. Citing the wiki: A pub or public house is an establishment that sells alcoholic drinks that can be consumed on the premises. Pubs commonly sell food which also can be eaten on the premises.

I think MásQMenos tries to imitate and standardize a typical Spanish tapas bar, so therefore it should be tagged as an amenity=pub.

bhousel:
My strong recommendation is to ask around a few more viewpoints before settling on tagging MásQMenos as `amenity=pub`, especially ask some people from gb if they are ok with this..

yopaseopor:
If the option amenity=pub does not fit to a typical Spanish place where they are serving "tapas" then we will need a new value called amenity=tapas_bar (we call bars to Spanish pubs) . As we don't ask to German people any problem with biergarten I don't know why is there any problem with the classification done by an entire community as the Spanish. If that classification would be not accepted, then we will need a new specific value.

Robot8A:
The tag cuisine=tapas makes it clear that it's not a typical British pub. In the Spanish community we have had everlasting conversations about how to tag Spanish (Mediterranean) bars. Perhaps as @yopaseopor suggests it is time to open the can of worms and suggest a new tag for this kind of establishments, or even a re-haul of the whole food&drink amenity tagging, as the current one is quite north-european centric. We are currently debating in the Spanish community about this issue.

bhousel:
All I'm asking is that you check with more people.  "pub" seems weird to me, but I don't know this place.

yopaseopor:
Yes, right. You have said: especially ask some people from gb... why from gb and not from D,A,BE,FR,P,IT,NL,DK,SE,SW,UK,RU,TR... to tag an Spanish chain of restaurants?
From which major nationality would be the costumers of that chain? GB?

bhousel:
>Yes, right. You have said: especially ask some people from gb... why from gb and not from D,A,BE,FR,P,IT,NL,DK,SE,SW,UK,RU,TR... to tag an Spanish chain of restaurants?
From which major nationality would be the costumers of that chain? GB?

The pull request includes these ones:
`"locationSet": {"include": ["es","gb","fr"]},`

jandr:
> All I'm asking is that you check with more people. "pub" seems weird to me, but I don't know this place.

You are indeed right, and the 'pub' tagging normally seems weird for Spanish people regarding what we now as a "Bar" (and probably other Mediterranean people as well, as I believe they have the same concept).  We have English or Irish Pubs in Spain, and they are different in concept to our "Bars", so it feels weird to people to tag these facilities as "pub", but we are trying to be consistent with the definition and not with the name.

Spanish bars typically serve coffee (or tea) and some kind of breakfast in the morning (which can range from toasts, [churros](https://en.wikipedia.org/wiki/Churro), bakery products, omelettes, etc.); are used for coffee/tea/refreshment beverages/beer, etc. mid-morning; can serve tapas or some meal-of-the-day for lunch (normally with less variety and complexity than a 'proper' restaurant); are used for (alcoholic) beverages in late afternoon or evening (beer, wine, tapas, etc.). 

So they are not what in English is known as 'bar' (which mainly serves alcoholic drinks, and to confuse things further has been known in Spain as a 'pub' when it also plays music but is not a dance-club or night-club per se), and would be more close to the 'pub' definition in OSM, although they do not match exactly. 

Most newcomers to the Spanish OSM community start tagging these Bars as 'bar' in OSM, because that is what their logic and common sense dictates, but we have agreed as a community that, in order to standarize the tag, we will be using 'pub' to tag these (so you might have something tagged as 'ameniy=pub' and 'name=Bar Pepe', because locally it is called like that).  During the debate, different approaches were considered, from leaving them as 'amenity=bar' (since they are called 'bares' (plural for bar in Spanish)) and have people visiting Spain understand that the tag here had a different meaning (which also would confuse renders), trying to create a new tag that reflects their reality better (as has been again suggested in this thread), or tagging them as 'amenity=pub', which most people feel is not 100% accurate but is the most similar definition that matches...

Hope I have clarified [where] the discussion [comes from] a little bit...

yopaseopor:
So, we have two option: or the nowadays scheme fits and Spanish places where serve food and drink in a table o a bar can be pubs, bars, cafes , restaurants or we need a new value for amenity or a new scheme. In Spain, in OpenStreetMap there are more than 150000 places of that kind. All discussion will be welcomed.

matkoniecz:
> tag here had a different meaning

That is a really bad idea, I am glad that it was not used. BTW, Poland also has things called "bar" in Polish with meaning of "low-end restaurant / traditional fast food / kind of pub" (with `amenity=bar` not used for them, except mistakes by newbies).

yopaseopor:
How do you tag this in Poland?
May this things would fit in a new tag/scheme?

<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Libre
de virus. www.avast.com
<https://www.avast.com/sig-email?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

Robot8A:
Regarding MásQMenos, this is their menu: https://www.masqmenos.com/carta_en.pdf
They serve alcoholic drinks (beer, wine) as well as non-alcoholic wines like coffee or even smoothies.
So it may not be a pub nor a cafe nor a restaurant, but a combination of those three. I strongly think MásQMenos is a take on a typical Spanish bar, franchising it and focusing a bit more on ham and tapas.
So you can have a coffee and a croissant for breakfast, or a juice and cold cuts plate for brunch, or a sandwich and beer for dinner.

matkoniecz:
> How do you tag this in Poland?

`amenity=restaurant`, `amenity=fast_food` `amenity=bar` depending on what fits best. There are some additional confusing false friends like "bar mleczny" that is not the same thing as Australian "[milk bar](https://en.wikipedia.org/wiki/Milk_bar)" but rather fast food serving traditional meals (dating back to communist times and one of few things that were done well during that era).

PR #4102, opened 2020-08-26
Add Tchibo coffee shop

PiotrBlachnio:


brands/amenity/cafe.json
```diff
@@ -1062,6 +1062,29 @@
       "takeaway": "yes"
     }
   },
+  "amenity/cafe|Tchibo": {
+    "locationSet": {"include": [
+      "at",
+      "cz",
+      "hu",
+      "is",
+      "nl",
```
RobinLinde, line 1071: According to the [Dutch Wikipedia article](https://nl.wikipedia.org/wiki/Tchibo#Geschiedenis), the last location in the Netherlands closed in 2007.

PiotrBlachnio: Updated.

UKChris-osm:
Tchibo appears to already exists within the brand index, is this intended to be an update?

bhousel:
> Tchibo appears to already exists within the brand index, is this intended to be an update?
> 
> Edit: I see now, it's already listed but under "_shop/cafe_" as opposed to "_amenity/cafe_"

I'll just merge as-is and cleanup in a followup commit after the identifiers settle..

bhousel:
whoops - I meant to merge but clicked the wrong button.
Anyway I will move the record to `amenity/cafe` later.

PR #4107, opened 2020-09-01
Update restaurant.json

### LINKED ISSUE #4100: Loving Hut is vegan=only, not vegetarian=only
Besides, vegetarian=only is implied by vegan=only.

See https://wiki.openstreetmap.org/wiki/Key:diet:vegan

nicolasmaia:
Fixes #4100


Adamant36:
Thanks. Looks good :+1:

PR #4108, opened 2020-09-01
Added/improved pastry and bakery chains in Luxembourg

DavidMoraisFerreira:
This PR adds Wikidata identifiers for pastry/bakery chains in Luxembourg. I also added a chain that was missing.


Adamant36:
Thanks. Looks fine :+1:

PR #4110, opened 2020-09-03
Add gas station chain "Fieten Olie"

hole1988:


Adamant36:
Thanks for the PR. I don't think the match name is really needed in this case, but other then that it looks good :+1:

PR #4118, opened 2020-09-09
add details to Moldindconbank

Identitaet:


brands/amenity/bank.json
```diff
@@ -4622,10 +4622,12 @@
     }
   },
   "amenity/bank|Moldindconbank": {
-    "locationSet": {"include": ["001"]},
+    "locationSet": {"include": ["md"]},
     "tags": {
       "amenity": "bank",
       "brand": "Moldindconbank",
+      "brand:wikidata": "Q46730",
```
tas50, line 4629: This wikidata doesn't look like it's the right one

Identitaet: Good catch
I

Adamant36:
Merged since it was fixed. Thanks.

PR #4119, opened 2020-09-11
add Saint Petersburg Bank details

Identitaet:


UKChris-osm:
Thank you @Identitaet

PR #4127, opened 2020-09-11
add details to Banpais

Identitaet:
also added spelling with í cause that's what they also use (they're really inconsisten with that)


Identitaet:
it's giving me a weird warning does NSI care about special characters like í ?

tas50:
Thanks @Identitaet

PR #4146, opened 2020-09-15
improved butcher 'Renmans'

thibaultmol:


UKChris-osm:
Thank you @thibaultmol

PR #4159, opened 2020-09-17
Move Robert Dyas to hardware

peternewman:
The first line of https://wiki.openstreetmap.org/wiki/Tag:shop=hardware says "sometimes known as an ironmongers shop in the UK", some of the shops are branded as "Robert Dyas The Ironmongers".

### DIFF
--- a/brands/shop/hardware.json
+++ b/brands/shop/hardware.json
@@ -84,6 +84,20 @@
       }
     },
     {
+      "displayName": "Robert Dyas",
+      "id": "robertdyas-6506bb",
+      "locationSet": {"include": ["gb"]},
+      "matchTags": ["shop/electronics"],
+      "oldid": "shop/hardware|Robert Dyas",
+      "tags": {
+        "brand": "Robert Dyas",
+        "brand:wikidata": "Q7343720",
+        "brand:wikipedia": "en:Robert Dyas",
+        "name": "Robert Dyas",
+        "shop": "hardware"
+      }
+    },
+    {
       "displayName": "True Value",
       "id": "truevalue-a223fb",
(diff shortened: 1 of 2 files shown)


brands/shop/hardware.json
```diff
@@ -83,6 +83,20 @@
         "shop": "hardware"
       }
     },
+    {
+      "displayName": "Robert Dyas",
+      "id": "robertdyas-6506bb",
+      "locationSet": {"include": ["gb"]},
+      "matchTags": ["shop/electronics"],
+      "oldid": "shop/hardware|Robert Dyas",
```
peternewman, line 91: Should I have changed this or left it as shop/electronics?

peternewman:
I assume this will be enough to get iD to update the other instances as people find them?

brands/shop/hardware.json
```diff
@@ -83,6 +83,20 @@
         "shop": "hardware"
       }
     },
+    {
+      "displayName": "Robert Dyas",
+      "id": "robertdyas-6506bb",
```
peternewman, line 88: Do I need to change this ID at all?

bhousel: No need, the build script will take care of it.

bhousel:
> I assume this will be enough to get iD to update the other instances as people find them?

Yes 👍

brands/shop/hardware.json
bhousel: It doesn't really matter, the `oldid` only exists as a claim on Wikidata, and I'll probably end up bulk updating them anyway.

peternewman:
Great, thanks for that @bhousel

PR #4164, opened 2020-09-20
Added Milma

AnjaliGGeetha:
For the name, common name 'Milma' is used. For Brand, legal name of the firm holding the brand is used. The public is generally not aware of the legal name.

### DIFF
--- a/brands/shop/dairy.json
+++ b/brands/shop/dairy.json
@@ -1,6 +1,17 @@
 {
   "brands/shop/dairy": [
     {
+      "displayName": "Milma",
+      "id": "milma-6e9671",
+      "locationSet": {"include": ["in"]},
+      "tags": {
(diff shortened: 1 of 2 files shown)


bhousel:
> For the name, common name 'Milma' is used. For Brand, legal name of the firm holding the brand is used. The public is generally not aware of the legal name.

That's kind of weird, but ok.
Can you also add a `brand:wikipedia` tag?

I see on https://www.wikidata.org/wiki/Q3532857 that there are several wikilinks..
This one looks to be the most complete:
`"brand:wikipedia": "en:Kerala Co-operative Milk Marketing Federation"`

bhousel (changes requested):
add `brand:wikipedia` tag

bhousel:
looks like #4179 replaces this PR

Issue #4123, opened 2020-09-11
Add Spencer's 


Adamant36:
Indian convenience and supermarket chain.

https://en.wikipedia.org/wiki/Spencer%27s_Retail
https://www.wikidata.org/wiki/Q7575959

PR #4175, opened 2020-09-20
spencers added

zahra-abdullah:
https://github.com/osmlab/name-suggestion-index/issues/4123


bhousel:
Thanks @zahra-abdullah and welcome! 👋

PR #4176, opened 2020-09-20
Nomi added

zahra-abdullah:
https://github.com/osmlab/name-suggestion-index/issues/4126


bhousel:
Thanks @zahra-abdullah and welcome! 👋

PR #4183, opened 2020-09-22
Adding Wetherspoon brand, without the name requirement,

UKChris-osm:
Adding Wetherspoon pub brand, without the name requirement, although I may have misunderstood the optional nature of "name" now.


bhousel:
Haven't tested it yet, but it should work :shipit:

PR #4186, opened 2020-09-22
remove brand name

hernandez87v:
- This removes a generic name that isnt an actual brand name.
- made sure that the items I wanted to discard are gone (and no others are affected)


bhousel:
Looks good to me, thanks @hernandez87v and welcome 👋

PR #4188, opened 2020-09-22
Add wiki tags for Gogoro in charging_station.json and removed generic brand

hernandez87v:


Adamant36:
If I'm understanding this correctly Enel x is a brand of charging stations that are used with Metbus, but the brand isn't "Enel x Metbus"?

hernandez87v:
> If I'm understanding this correctly Enel x is a brand of charging stations that are used with Metbus, but the brand isn't "Enel x Metbus"
Yes that's correct.

Adamant36:
Hhhhmmm, that's what I thought. There's currently like 72 uses of name=Enel x Metbus. So, it might be to do a match  or something for it instead of just discarding the whole thing outright so they are "corrected."

bhousel:
> There's currently like 72 uses of name=Enel x Metbus. So, it might be to do a match or something for it instead of just discarding the whole thing outright so they are "corrected."

Yes, I think in this case it would be more correct to use matchNames to match the `enel x metbus` name that is commonly used.  I did this in cede9eaf

> (As an off topic side thought, this is an instance where I wish we didn't have to use the name tag at all)

As of yesterday the `name` tag isn't required anymore.  I left it in for now, but if people really want to assign unique names to all the Enel X charging stations, we can remove it.

Issue #4140, opened 2020-09-14
Jewson needs changing from shop=doityourself to shop=trade


lakedistrictOSM:
Q6190226

The chain is aimed at builders, rather than regular DIY customers.

UKChris-osm:
Would you say [Travis Perkin](https://nsi.guide/index.html?k=shop&v=doityourself#Travis%20Perkins) should also be listed as trade??

peternewman:
From:
https://www.travisperkins.co.uk/content/faqs

> Q1 - I'm new to Travis Perkins. Do you only supply to the Trade? And do I need to have an account to buy from you?
> Anyone can buy from Travis Perkins and you don't need to open an account.

I can't find something so clear cut with Jewson:
https://forums.moneysavingexpert.com/discussion/5063042/buying-from-trade-only-shops

They may be aimed at trade more so than B&Q say, I'd say that's probably true of both.

According to that MSE forum, Howdens is trade only.

matkoniecz:
https://en.wikipedia.org/wiki/Jewson 

> Jewson is one of the largest chains of British general builders' merchants, selling to small and medium building contractors.

On the other hand https://www.jewson.co.uk/tools-fixings/hand-tools lists things that would perfectly fit DIY...

peternewman:
> On the other hand https://www.jewson.co.uk/tools-fixings/hand-tools lists things that would perfectly fit DIY...

Erm yes, because the building trade is basically paid DIY, or vice-versa.

> I asked on talk-gb for more feedback.

I'd suggest it's a more general issue. From the wiki:
"A place of business that sells to a particular trade or trades, but normally also retails to normal consumers."

Then you're supposed to use trade=* to distinguish which trade it sells to. trade=building_supplies in this case.

Then later relative to DIY:
"TODO: document how to distinguish"

To me, this is a bit of a messy tagging scheme, unless it ONLY sells to people with a trade account (e.g. proof of trade employment), then if I want to buy some paint or wood, I just want the nearest store selling them, I may even favour the trade supplier as they may be cheaper. Perhaps if I'm at the more naive end of things, I might go to a DIY store to ask advice, although that can be mixed (but I think it's generally better there then in Maplins (as was)/Currys/PC World etc).

peternewman:
See also #4190

PR #4190, opened 2020-09-23
Add the type of trade to Jewson's (and others) as part of #4140

peternewman:
Add the inevitable matchname too and some matchtags

### DIFF
--- a/brands/shop/trade.json
+++ b/brands/shop/trade.json
@@ -5,6 +5,9 @@
       "id": "84lumber-4e950c",
       "locationSet": {"include": ["us"]},
       "matchNames": ["84", "eighty-four lumber"],
+      "matchTags": [
+        "shop/doityourself"
+      ],
       "tags": {
         "brand": "84 Lumber",
         "brand:wikidata": "Q4644779",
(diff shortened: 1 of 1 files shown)


bhousel:
Looks ok to me, thanks 👍 

The matchTags for `shop/doityourself` is a good suggestion, but it will get removed as redundant when I run the build script.

This is because of "match groups" - see:
https://github.com/osmlab/name-suggestion-index/blob/main/config/match_groups.json
Tags in a match group will automatically match any other tags in the same match group.

peternewman:
> The matchTags for `shop/doityourself` is a good suggestion, but it will get removed as redundant when I run the build script.
> 
> This is because of "match groups" - see:
> https://github.com/osmlab/name-suggestion-index/blob/main/config/match_groups.json
> Tags in a match group will automatically match any other tags in the same match group.

Ah I did ponder them, but didn't go looking for them. Can the match groups handle shop=trade being in multiple match groups? Wholesale should probably have it for the catering supplies aspect for example.
https://wiki.openstreetmap.org/wiki/Tag:shop=trade

bhousel:
> Ah I did ponder them, but didn't go looking for them. Can the match groups handle shop=trade being in multiple match groups? Wholesale should probably have it for the catering supplies aspect for example.

Yes the matcher code does allow a tagpair to be in to multiple match groups.

For example, `shop=wholesale` is in a few match groups.  This means that if a wholesaler like Costco is mistagged as `shop=supermarket` or `shop=department_store` or `shop=food` it will still be matched, and the iD validator will suggest it to be tagged as `shop=wholesale` to match the NSI.

PR #4191, opened 2020-09-23
Added Santander Cycles to amenity=bicycle_rental

peternewman:
A rehash of #3527 now no names are allowed.


peternewman:
@rskedgell @bhousel do either/both of you want to review this?

brands/amenity/bicycle_rental.json
```diff
@@ -159,6 +159,21 @@
         "name": "OV-fiets"
       }
     },
+    {
+      "displayName": "Santander Cycles",
+      "id": "santandercycles-fa5175b",
+      "locationSet": {"include": ["gb"]},
+      "matchNames": ["barclays bikes", "barclays cycle hire", "boris bike", "boris bikes"],
```
peternewman, line 166: Do these work on brand? Hopefully they can be made to do so if they don't currently?

bhousel: I'm not sure what you mean?

bhousel:
Thanks @peternewman 👍 
I think this looks ok, but I'll squash away a lot of the work-in-progress commits, and remove that stuff in the travis file.

brands/amenity/bicycle_rental.json
peternewman: Sorry, I mean do matchNames just work on a name field or will they also cover brands, or is that a question for iD and other consumers?

i.e. if I type in Boris Bike will if offer Santander Cycles?

peternewman:
> I think this looks ok, but I'll squash away a lot of the work-in-progress commits,

Fair enough

> and remove that stuff in the travis file.

It will only do anything if the build fails, but up to you obviously.

Unfortunately from my brief experience of JSON validation they do seem to be rather good at giving you a rather vague and incomprehensible error message when they fail.

brands/amenity/bicycle_rental.json
bhousel: > i.e. if I type in Boris Bike will if offer Santander Cycles?

Yes, the brand tag is considered an alternate name also.

PR #4193, opened 2020-09-23
add shoe chain ara

Identitaet:
yes it's lowercase in their own styelisation and it's already mapped that way


UKChris-osm:
Thanks @Identitaet

PR #4197, opened 2020-09-24
Added internet cafes

AnjaliGGeetha:
Added Akshaya & CSC.

### DIFF
--- a/brands/amenity/internet_cafe.json
+++ b/brands/amenity/internet_cafe.json
@@ -1,6 +1,30 @@
 {
   "brands/amenity/internet_cafe": [
     {
+      "displayName": "Akshaya",
+      "id": "akshaya-ec41bc",
+      "locationSet": {"include": ["in"]},
+      "tags": {
+        "amenity": "internet_cafe",
+        "brand": "Akshaya",
+        "brand:wikidata": "Q99562093",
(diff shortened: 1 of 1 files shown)


bhousel (changes requested):
Looks like the Wikidata ID for Akshaya project is Q4701821, and maybe Q99562093 is a duplicate that should be merged with the existing entry?

I see you've edited both of these on Wikidata in the past few days - what's your intent with having 2 separate QIDs?

AnjaliGGeetha:
Akshaya center is different from Akshaya project. Akshaya project started as e-literacy campaign by the government of Kerala. As part of that, private entrepreneurs were asked to establish internet kiosks where government e-services can be accessed by common people. What we are mapping is Akshaya centers which are not owned by the government but owned by private entrepreneurs. Hence added a separate wikidata entry for Akshaya Center.

bhousel:
ok, thanks for the clarification @AnjaliGGeetha  - in that case, can you remove the `"brand:wikipedia": "en:Akshaya project",` tag, since that really doesn't go with the Q99562093 id for the Akshaya Centers?

AnjaliGGeetha:
Ok, Removed wikipedia tag from Akshaya Center

Identitaet:
closed cause fixed in #4211

PR #4199, opened 2020-09-25
add country code and wiki to Bank99

Identitaet:


UKChris-osm:
Thank you @Identitaet

PR #4200, opened 2020-09-25
Added Maveli Store

AnjaliGGeetha:


bhousel:
Thanks @AnjaliGGeetha , this looks ok, but I'm not too comfortable with introducing the `owner` and `ownership` tags on the items yet.  We don't use these tags anywhere else in the index and it seems like information that could just be fetched from Wikidata.

I'll merge and remove those tags for now, but we can add them back in another time if people really want this.

Issue #4203, opened 2020-09-25
Add brands in Germany


kjonosm:
brand=Vorwerk
name=Vorwerk
shop=vacuum_cleaner
brand:wikidata=Q449852
brand:wikipedia=de:Vorwerk_(Unternehmen)

brand=Zoo&Co.
name=Zoo&Co.
shop=pet
brand:wikidata=Q219979
brand:wikipedia=de:Zoo & Co.

brand=Wempe
name=Wempe
shop=jewelry
brand:wikidata=Q1483424
brand:wikipedia=de:Gerhard D. Wempe

brand=Multipolster
name=Multipolster
shop=furniture
brand:wikidata=n/a
brand:wikipedia=n/a
website=https://www.multipolster.de

brand=Poco
name=Poco
shop=furniture
brand:wikidata=Q1714820
brand:wikipedia=de:Poco (Möbelhaus)

brand=SB-Möbel Boss
name=SB-Möbel Boss
shop=furniture
brand:wikidata=n/a
brand:wikipedia=n/a
website=https://moebel-boss.de

brand=Holzleitner
name=Holzleitner
shop=electronics
brand:wikidata=n/a
brand:wikipedia=n/a
website=https://www.holzleitner.de/

brand=BayWa
name=BayWa
shop=doityourself
brand:wikidata=Q812055
brand:wikipedia=de:BayWa

brand=Vom Fass
name=Vom Fass
shop=deli
brand:wikidata=Q2533008
brand:wikipedia=de:Vom Fass

brand=Gravis
name=Gravis
shop=computer
brand:wikidata=Q1543982
brand:wikipedia=de:Gravis Computervertriebsgesellschaft

brand=Vergölst
name=Vergölst
shop=tyres
brand:wikidata=Q1673875
brand:wikipedia=de:Vergölst

brand=Jacques’ Wein-Depot
name=Jacques’ Wein-Depot
shop=wine
brand:wikidata=Q1678150
brand:wikipedia=de:Jacques’ Wein-Depot

brand=MFO Matratzen
name=MFO Matratzen
shop=bed
brand:wikidata=Q26794529
brand:wikipedia=de:MFO Matratzen

Identitaet:
Could you help me with what the actuall label used for Möbell Boss is

I'm a bit confused there

kjonosm:
> Could you help me with what the actuall label used for Möbel Boss is
> 
> I'm a bit confused there

[...]

PR #4205, opened 2020-09-25
add German brands (Part 2)

Identitaet:
@kjonosm just a tip you only need a wikidata item for this list a wikipedia article isn't required but it's nice
add BayWa
add Holzleitner


Identitaet:
#4203

bhousel:
cool 👍

PR #4206, opened 2020-09-25
add Möbel Boss (Part 3)

Identitaet:
please check the matchNames if I have done that correctly since this entry doesn't have a set name

### DIFF
--- a/brands/shop/furniture.json
+++ b/brands/shop/furniture.json
@@ -587,10 +587,11 @@
       "locationSet": {
         "include": ["de"]
       },
-      "matchNames":["SB-Möbel Boss","Boss Möbel","Möbel Boss"],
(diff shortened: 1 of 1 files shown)


bhousel:
I know this brand is unusual, but I still think it is important to have a name tag.

Would it be ok to just add 2 entries?
- one with name "Möbel Boss" and matchName for "SB-Möbel Boss"
- one with name "Boss Möbel" and no matchname

Same wikidata Q99624217 for both of them.

Identitaet:
Boss Möbel is only the way it looks on some signs due to weird layout decisions so I'll add möbel boss as the name

bhousel:
> Boss Möbel is only the way it looks on some signs due to weird layout decisions so I'll add möbel boss as the name

Thanks, that sounds good to me 👍

PR #4209, opened 2020-09-26
Add Cinesa, Spanish cinema chain

Robot8A:


bhousel:
Thanks @Robot8A looks good 👍 
BTW, it will get a different `id` generated for it next time someone runs the build script - you don't have to supply one.   (It's not a problem if you include it, but just wanted you to know that it won't end up being the `id` that you assigned it.)

PR #4211, opened 2020-09-27
Removed wikipedia tag from Akshaya Center

AnjaliGGeetha:
Removed wikipedia tag from internet cafe


Identitaet:
Okay this one looks good you fixed the wikipedia problem

PR #4220, opened 2020-09-30
Add Carlin, stationery chain

Robot8A:


bhousel (changes requested):
Thanks @Robot8A  - It looks like Carlin is already in that file, that's why the Travis check failed.  Can you remove the existing record?  The one you added looks great 👍

bhousel:
Thanks for the quick fix @Robot8A!

PR #4221, opened 2020-10-01
Added a couple of banks

suman-nath:
Added 8 new banks to the list.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -472,6 +472,94 @@
       }
     },
     {
+      "displayName": "Assam Gramin Vikash Bank",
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @suman-nath and welcome 👍 
The Banks look great .. 

But we don't add individual universities to this index, we only add the kind of for-profit "universities" operating like a branded business in multiple locations.  (I know this is an unusual concept outside of the USA - they are more like training centers than universities).

Anyway I'm going to back out the university commit and merge just the banks.. Thanks again 👍

bhousel:
This ended up merging in 6e6357b9

PR #4222, opened 2020-10-01
added Apollo hospitals

nis130:
added new entry of apollo hospital in the hospital.json


bhousel:
Thanks @nis130 looks great 👍

PR #4224, opened 2020-10-01
Update convenience.json

debyos:
CBA is a Hungarian brand having shops in these countries.

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -403,7 +403,7 @@
     {
       "displayName": "CBA",
       "id": "cba-4d6b4b",
-      "locationSet": {"include": ["001"]},
+      "locationSet": {"include": ['hu', 'sk', 'ro', 'bg']},
       "tags": {
         "brand": "CBA",
         "brand:wikidata": "Q779845",


bhousel:
Thanks @debyos and welcome 👋 

I noticed that CBA is in both the `shop/convenience.json` and also the `shop/supermarket.json` files..  Do you know whether:
- if it should always be tagged either supermarket or convenience store? (if so, remove the unwanted entry, and add a "matchTags" property to the wanted entry)
- or can it be both things?  (if so, please update the locations in both places).

debyos:
Hi @bhousel 

thanks for the quick and positive reaction!

To be honest, I am almost stupid for coding (therefore I am very proud that I could make a real pull request) but I do not really understand your question. What I know is that CBA is a big chain which has both `shop=convenience` and `shop=supermarket` stores.

I hope that with the help of this information you can make the appropriate changes in the code. ;-)

bhousel:
> What I know is that CBA is a big chain which has both `shop=convenience` and `shop=supermarket` stores.

Great! that answers the question..
Thanks again 👍

PR #4225, opened 2020-10-01
Added support for browsers not having JS enabled

hrupesh:
If Javascript is not enabled on the browser then a text showing `Your browser does not support JavaScript!` will be displayed to the User.


bhousel:
Thanks @hrupesh and welcome 👋

PR #4232, opened 2020-10-01
Add VRR

Identitaet:
adds VRR my local operator


Identitaet:
Okay it's ready had some trouble staring but now i'm good

bhousel:
We can `"matchNames": ["vrr"]` these too, if the naming is inconsistent and we want to nudge them towards using the preferred spelled out value "Verkehrsverbund Rhein-Ruhr".  What do you think?

Identitaet:
Sounds good Verkehrsverbund Rhein-Ruhr is the official name but since it's long everybodey uses VRR
I'll add that

bhousel:
Just wanted to check this a bit more before hitting the green merge button.  I went looking in the [networks_all.json](https://github.com/osmlab/name-suggestion-index/blob/main/dist/collected/networks_all.json) file:
```js
"route/bus|VRR": 2806,
"route/bus|Verkehrsverbund Rhein-Ruhr": 21,
```

It looks like for some of these "VRR" is overwhelmingly preferred by mappers. Do you need to check with anyone before setting it up this way?

Saying that the item is "Verkehrsverbund Rhein-Ruhr" with a matchName for "VRR" is like saying that they are the same thing and the spelled out version is preferred.  iD will suggest that "VRR" is outdated and the tag be upgraded to "Verkehrsverbund Rhein-Ruhr".  (adding `network:wikidata` is the real important part of the upgrade, but people still might not want the tag changed)

Identitaet:
I mean I usually think the official name is correct but since everyone uses VRR (in OSM and irl) I'll just change it

bhousel:
> I mean I usually think the official name is correct but since everyone uses VRR (in OSM and irl) I'll just change it

Cool, thanks for understanding..  Definitely don't want to upset any OSM rail mappers!

PR #4234, opened 2020-10-01
Adding missing tags

jeuel2001:
Finished adding missing tags :) please check if it's okay, still new to coding. Thank you!


data/brands/amenity/vending_machine.json
```diff
@@ -101,7 +103,8 @@
       "tags": {
         "amenity": "vending_machine",
         "brand": "Bild",
-        "name": "Bild"
+        "name": "Bild",
+        "vending": "parcel_pickup"
```
bhousel, line 107: Are you sure about this one?  A quick google search looks like this might be a newspaper vending macine.

data/brands/amenity/vending_machine.json
```diff
@@ -383,16 +380,6 @@
         "vending": "excrement_bags"
       }
     },
-    {
-      "displayName": "Schwalbe",
```
bhousel, line 387: Isn't this one real?  They make bicycle tubes..

bhousel (approved):
This looks mostly ok, just please double check the items that are being changed or deleted to make sure everything is accurate..

data/brands/amenity/vending_machine.json
jeuel2001: Okay cool! I'll look through the missing tags now! Thank you :)!

bhousel:
Sorry @jeuel2001 , I'm not convinced that you actually know what these brands are.  You removed the Bild vending machine and turned it into a `shop=newsagent`, and removed the Schwabe tire vending machine and turned it into a `shop=tyre` store.  These kinds of changes would upset a lot of people in OpenStreetMap community.

We are ok with the build script producing warnings - this just means that we'd like someone with local knowledge to provide the answer.

PR #4239, opened 2020-10-01
add DB brands

Identitaet:


Identitaet:
The error is I guess problematic since they are all operated by the same company

bhousel:
Travis is failing because of this - It probably is because of the duplicate `operator`, so I'll need to fix that.  My fault, not yours! 😁 
Screen Shot 2020-10-01 at 5 36 01 PM

Identitaet:
I mean just a small oversight from the normal procces

bhousel:
just pushed a fix, I'll rerun Travis and see if that worked 🤞

Identitaet:
works

PR #4240, opened 2020-10-01
add French regional rail lines under sncf

Identitaet:


Identitaet:
oh forgot to say operator=SNCF is verry commmonly used

bhousel:
Looks good to me!  

One thing I'd suggest is that in the locationSet you can use:
 `{ include: ['fx'] }` - [Metropolitan France](https://ideditor.github.io/location-conflation/?locationSet=%7B%20include%3A%20%5B%27fx%27%5D%20%7D) (just the part in Europe)
instead of
`{ include: ['fr'] }` - [France](https://ideditor.github.io/location-conflation/?locationSet=%7B%20include%3A%20%5B%27fr%27%5D%20%7D) (includes overseas territories)

As far as I know, France is the only country that has their ISO codes like this, but it's kind of useful for us to keep the scope narrowed.

bhousel:
I merged as-is.. you can push a change with `fr` -> `fx` later if you think it makes sense.

PR #4241, opened 2020-10-02
Adding new entry for train lines.

3Hearts:
Added Metro-North Railroad system as an entry to list.

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -485,6 +485,8 @@
         "network:wikidata": "Q125908",
         "network:wikipedia": "en:Metro-North Railroad",
         "operator": "Metro-North Railroad",
+        "operator:wikidata": "Q125908",
+        "operator:wikipedia": "en:Metro-North Railroad",
         "route": "train"
       }
     },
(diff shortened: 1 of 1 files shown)


data/transit/route/train.json
```diff
@@ -476,6 +476,18 @@
...
     },
+    {
+      "displayName": "Metro-North Railroad",
+      "id": "mnr-44bbb3",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "network": "Metro-North Railroad",
+        "network:wikidata": "Q125908",
+        "network:wikipedia": "en:Metro-North Railroad",
+        "operator": "Metro-North Railroad",
```
bhousel, line 487: I know it's kind of redundant, but if adding `operator` we should add the wikipedia and wikidata for those also.  (I'll update the docs to suggest this also).

```suggestion
        "operator": "Metro-North Railroad",
        "operator:wikidata": "Q125908",
        "operator:wikipedia": "en:Metro-North Railroad",
```

bhousel (changes requested):
Thanks @3Hearts - this looks great, just one minor suggestion..

bhousel (approved):

PR #4242, opened 2020-10-02
fixed missing tag in restaurant and fast_food deleted generic names

Maxgit3:
first pull request to add missing cuisine tags.  removed generic name in restaurant Json file


bhousel:
Sorry, I have to close this - it would conflict with #4234 already in progress. Also there were some missing tags in there - one of them said `"cuisine": ""`, but there were several other odd choices.

PR #4243, opened 2020-10-02
Add pizzeria chains in Israel

yotamN:
Sadly both chains are not well mapped yet but Pizza Shemesh have 63 branches and Pizza Prego 21 so hopefully it will help in the future.


Identitaet:
Looks good 👍
Thanks @yotamN

PR #4249, opened 2020-10-02
Added West Somerset Railway

UKChris-osm:
Added West Somerset Railway


bhousel:
Nice @UKChris-osm , thanks! 🚂

PR #4251, opened 2020-10-03
added rail networks in the Western United States

claysmalley:
added Wikidata and Wikipedia for:

- Altamont Corridor Express
- UTA FrontRunner
- Metrolink
- NCTD Coaster
- NM Rail Runner Express
- Sounder

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -21,6 +21,18 @@
       }
     },
     {
+      "displayName": "Altamont Corridor Express",
+      "id": "altamontcorridorexpress-44bbb3",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "network": "ACE",
+        "network:wikidata": "Q434302",
+        "network:wikipedia": "en:Altamont Corridor Express",
+        "operator": "San Joaquin Regional Rail Commission",
+        "route": "train"
+      }
+    },
+    {
       "displayName": "Amtrak",
       "id": "amtrak-44bbb3",
       "locationSet": {"include": ["us"]},
@@ -265,6 +277,18 @@
       }
     },
     {
+      "displayName": "FrontRunner",
+      "id": "frontrunner-44bbb3",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "network": "UTA",
+        "network:wikidata": "Q5505779",
+        "network:wikipedia": "en:FrontRunner",
+        "operator": "UTA",
+        "route": "train"
+      }
+    },
+    {
       "displayName": "Großraum-Verkehr Hannover",
       "id": "grossraumverkehrhannover-44bbb3",
       "locationSet": {"include": ["001"]},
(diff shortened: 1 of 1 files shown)


Adamant36:
No Wikipedia/Wikidata entries for the operators?

data/transit/route/train.json
```diff
@@ -20,6 +20,18 @@
...
     },
+    {
+      "displayName": "Altamont Corridor Express",
+      "id": "altamontcorridorexpress-44bbb3",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "network": "ACE",
+        "network:wikidata": "Q434302",
+        "network:wikipedia": "en:Altamont Corridor Express",
+        "operator": "San Joaquin Regional Rail Commission",
```
1ec5, line 31: The San Joaquin Regional Rail Commission is the authority that owns and governs the ACE service, but does SJRRC operate the line? My understanding was that, in plain English, the “operator” of a commuter rail line would be the staffing company, which would be Herzog Transit Services. Should we use `owner` for SJRRC and `operator` for Herzog? (I’m not especially interested in tagging these staffing companies, but I guess it doesn’t hurt. Staffing contracts can last several years but typically change with very little fanfare.)

For comparison, [Caltrain `train` route relations](https://www.openstreetmap.org/relation/2768040) are currently tagged `operator=TransitAmerica Services`. Meanwhile, [Caltrain `railway` route relations](https://www.openstreetmap.org/relation/7916804) are tagged `owner=Peninsula Corridor Joint Powers Board` `operator=Peninsula Corridor Joint Powers Board;Union Pacific Railroad`, reflecting the fact that some trackage rights are shared with Union Pacific.

natfoot: All locomotives are marked as ACE and ACE would be the operator. If ACE has a contract with HERZOG to run these trains for ACE this would be a direct contract and does not mean they are the operator. It just means that they are contracted to run the trains and train services. Another way to look at this is that the reporting marks on the train cars and locomotives is the operator and it looks like the line is owned by Union Pacific. 
I am a little more related to Sounder Train service in Seattle.  Sounder Trains are operated by Sound Transit but they have a contract with BNSF to run the trains. I don't see a need to list that the Sounder Trains are operated by BNSF crews.  

There is a little bit of track that the Sounder Trains run on that are owned by Washington State DOT and the trains are still operated by Sound Transit with a contract with BNSF.  

In conclusion I would suggest a new term like staffed by HERZOG or staffed by BNSF as railroad operator is very specific to the regulatory operation of the line. For more detail look for who the operator is for a line on the Surface Transportation Board Website.  stb.gov

1ec5: Personally, I’m not enthusiastic about tagging staffing contractors at all in OSM. That seems to be a detail best left for Wikidata. But it does seem to align with the situation in Europe that @westnordost described in Slack.

Looking back at how bus routes and infrastructure are usually tagged, I think setting `operator` to the San Joaquin Regional Rail Commission would be OK, but it is a change from how commuter rail lines have often been tagged.

natfoot: It was late last night when I wrote that and it is not complete. I agree the staffing contractors is best left to wiki data. I hear you that you think this is different that things have often been tagged for railroad transit.  I can say that I looked at a line in Niles Canyon with ACE and the ACE relation does not list the staffing contract as operator, nor does the Sound Transit line I looked at.

bhousel: Rather than let this PR get more stale, I merged it as-is, since it looks mostly good and contains useful ids.
If you want to adjust the `operator`/`owner` etc tags, we can always revise this with a followup PR.

PR #4253, opened 2020-10-03
Added a Indian Jewelry brand

ankita2210:
Added a Indian Jewelry brand : PC: Jeweller 
**brand:wikipedia: en:PC Jeweller**


Identitaet:
Looks good

PR #4256, opened 2020-10-03
Country coder supports GB-SCT so use that directly

peternewman:
They also handle US-HI etc, so all those states could be dropped too.


peternewman:
I've just seen you asked for this in https://github.com/ideditor/country-coder/issues/26 :facepalm: 

Do we want the other stuff to switch across too?

peternewman:
Blocked by https://github.com/ideditor/country-coder/issues/29 and it not being released yet. Although nothing actually seems to be using Scotland that I can see...

bhousel:
Yeah I was going to say - I don't think country-coder actually supports this level of detail.

When we need sub-country detail, I've been just drawing a custom geojson and dumping them into the `features/` folder 
(and I think it would be wonderful for all the transit networks to have a geojson saying where they operate).

We do have a few of these in `features/` already:
* [great_britain.geojson](https://github.com/osmlab/name-suggestion-index/blob/main/features/great_britain.geojson)
* [northern_ireland.geojson](https://github.com/osmlab/name-suggestion-index/blob/main/features/northern_ireland.geojson) 
* [scotland.geojson](https://github.com/osmlab/name-suggestion-index/blob/main/features/scotland.geojson)

peternewman:
> Yeah I was going to say - I don't think country-coder actually supports this level of detail.
> 
> When we need sub-country detail

Erm England is a country:
https://en.wikipedia.org/wiki/England

> (and I think it would be wonderful for all the transit networks to have a geojson saying where they operate).

I assume you mean where they currently operate, couldn't this be extracted from OSM and then verified, rather than people having to/trying to draw custom geojson, assuming you want a particular corridor it runs in or something?

Although it may not have its own ISO code at all levels.

> We do have a few of these in `features/` already:

I'd have thought, due to the risk of errors as below, having them in just one place would be better than having to roll our own.
 
> * [great_britain.geojson](https://github.com/osmlab/name-suggestion-index/blob/main/features/great_britain.geojson)

This is currently wrong as it includes the Isle of Man (I had to look it up).

peternewman:
Are you onboard with this in principle generally @bhousel then I could delete the duplicate geojsons and potentially start fixing up the config pending the release of the new version?

bhousel:
> Are you onboard with this in principle generally @bhousel then I could delete the duplicate geojsons and potentially start fixing up the config pending the release of the new version?

Yes 👍   Anything we can do to avoid shipping geojson, and just use whatever country-coder supports is a win..
I'd hold off doing the work until it's actually released though.

peternewman:
> I'd hold off doing the work until it's actually released though.

Okay, I sort of assumed that if they support a particular thing we can guarantee it will appear eventually so we could at least delete our geojson, even if it's not worth the conflicts of editing the presets yet?

bhousel:
> Okay, I sort of assumed that if they support a particular thing we can guarantee it will appear eventually so we could at least delete our geojson, even if it's not worth the conflicts of editing the presets yet?

I wouldn't merge something like that.  The locationSets would be invalid in the interim, and links would stop working on http://nsi.guide

peternewman:
Yeah I didn't expect you to, just preemptively starting the assessment of what's duplicated and then hopefully the checks would flag the broken location sets to fix up.

peternewman:
They've released, although just looking at package.json here do we need to get this bumped too? https://github.com/ideditor/location-conflation

bhousel:
These are unblocked now with location-conflation 0.6 and country-coder 4.0  🎉

PR #4257, opened 2020-10-03
Make GB transit location sets more specific

peternewman:
Compared to a shop they're unlikely to expand across borders and we could potentially just suggest a list of possible transit operators for each area as there should be fewer of them.


peternewman:
Blocked by https://github.com/ideditor/country-coder/issues/29 and it not being released yet.

peternewman:
> Also redo the fix in #4258 before this goes in and make a similar edit in #4261.

Done these in https://github.com/osmlab/name-suggestion-index/pull/4598

PR #4258, opened 2020-10-03
Add Romney, Hythe & Dymchurch Railway

peternewman:
I didn't realise it had stopped, but up until July 2015 it was still running school services, it's more than just a tourist attraction though:
https://en.wikipedia.org/wiki/Romney,_Hythe_and_Dymchurch_Railway#Passenger_services

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -791,7 +791,7 @@
     },
     {
       "displayName": "Romney, Hythe & Dymchurch Railway",
-      "locationSet": {"include": ["gb-eng"]},
+      "locationSet": {"include": ["gb"]},
       "tags": {
         "network": "Romney, Hythe & Dymchurch Railway",
         "network:wikidata": "Q2090360",
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -1410,6 +1410,18 @@
         "route": "train"
       }
     },
+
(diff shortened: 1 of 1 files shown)


data/transit/route/train.json
```diff
@@ -789,6 +789,17 @@
...
       }
     },
+    {
+      "displayName": "Romney, Hythe & Dymchurch Railway",
+      "locationSet": {"include": ["gb"]},
+      "tags": {
+        "network": "Romney, Hythe & Dymchurch Railway",
+        "network:wikidata": "Q2090360",
+        "network:wikipedia": "en:Romney, Hythe and Dymchurch Railway",
+        "operator": "Romney, Hythe & Dymchurch Railway PLC",
```
bhousel, line 799: Any reason for the "PLC" here?  Seems like it should just be the same thing as what `network` has.

peternewman: I just went off Wikipedia: https://en.wikipedia.org/wiki/Romney,_Hythe_and_Dymchurch_Railway#Ownership_and_operation

bhousel: Gotcha - I think I'm just going to put them as the same thing for now, and we can always change it later if the railway and the owning company ever end up having separate wikidata ids.

PR #4259, opened 2020-10-03
Added Super-Pharm brand:wikidata and brand:wikipedia

Ad1-18:


Identitaet:
I'm a bit confused about you also correcting the railway ids but it looks correct and can be added
Thanks

PR #4260, opened 2020-10-03
Improve Mi Teleférico aerialway

peternewman:
I'm assuming you only want proper networks of these, not point to point ones like this:
https://en.wikipedia.org/wiki/Emirates_Air_Line_(cable_car)

### DIFF
--- a/data/transit/route/aerialway.json
+++ b/data/transit/route/aerialway.json
@@ -3,9 +3,11 @@
     {
       "displayName": "Mi Teleférico",
       "id": "miteleferico-af9569",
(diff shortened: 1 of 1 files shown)


bhousel:
> I'm assuming you only want proper networks of these, not point to point ones like this:
> https://en.wikipedia.org/wiki/Emirates_Air_Line_(cable_car)

I haven't thought about whether to include "networks-of-one-thing", but I guess I'm ok with it.  

It would be great to nudge OSM towards considering `network`/`network:wikidata` as "required" tags for any transit route.

PR #4266, opened 2020-10-03
Added UFMG and major Brazilian schools

jvbraganca:
Hello!

Added some of the Brazilian schools.

I'm trying to find a way to create a script to get the list and data of all public universities so I can build this file automatically.

### DIFF
--- a/data/brands/amenity/school.json
+++ b/data/brands/amenity/school.json
@@ -1,6 +1,71 @@
 {
   "brands/amenity/school": [
     {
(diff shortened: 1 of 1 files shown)


Adamant36:
Hello. Thanks for the PR. As it currently is though we do not add schools unless it is widespread and has many satellite schools. Which it seems like these schools do not have (at least the ones I looked into). So, I am going to have to close this. If there are any particular schools in Brazil with widespread campuses I would welcome a PR for them though.

PR #4280, opened 2020-10-04
Update index.html

manish-kumar5848:


UKChris-osm:
Hi @manish-kumar5848 

Sorry if I am being blind here, but what changes does this PR make to index.html?

All I can see is the addition of a space between **Javascript!** and **\</noscript\>**, was this the intended change?

Issue #4274, opened 2020-10-04
Add Tantalizers


Adamant36:
Nigerian fast food chain. 

Cuisine=nigerian

https://en.wikipedia.org/wiki/Tantalizers
https://www.wikidata.org/wiki/Q7683795

29parthi:
can i work on this issue

29parthi:
#4282

Adamant36:
Closed in #4282

PR #4282, opened 2020-10-04
Fix for #4274 : added Tantalizers

29parthi:


29parthi:
#4274 Fixed

UKChris-osm:
Hi @29parthi,

Just a quick note, if "_short_name_" is the same value as "_name_", then you can remove "_short_name_" from your PR. Also, if you don't have an "_id_" value, you can remove that as well and it will be generated automatically later on :)

Adamant36:
> if you don't have an "id" value, you can remove that as well and it will be generated automatically later on

What's the "id" for? I must have been out of the office that day.

Adamant36:
Fixed in be253cf. Thanks for the PR.

UKChris-osm:
> What's the "id" for? I must have been out of the office that day or something.

It's something that's now created automatically via **npm run build** for each entry, replacing the old "_amenity/fast_food|Tantalizers_" ID's that were used previously.

PR #4278, opened 2020-10-04
German transit districts

Karthoo:
added network:wikipedia + network:wikidata links, removed operator tags, duplicates and historical networks and added a few missing networks

PR #4289, opened 2020-10-04
German transit networks pt. 1

Karthoo:
added network:wikipedia + network:wikidata links, removed operator tags, duplicates and historical networks and added a few missing networks


Karthoo:
Checker has the same problem as in #4278

bhousel:
> Checker has the same problem as in #4278

Yes, Travis build is failing because you have multiple networks `AVV`, so it generates duplicate ids.  You'll need to pick unique `newtork` values for these.

bhousel:
Ok I'm a bit more caught up on what happened with #4278 ..
The build code generates a unique `id` based on the `network` and `locationSet` and `route=bus`. 

So for example, you can't have multiple bus routes called "AVV" with a locationSet of `{"include": ["de"]}`.  You can give them unique `network` or change the locationSets so they that each "AVV" operate in different parts of Germany (by supplying geojson for where they operate).  Does that make sense?

Identitaet:
After resolving the merge conflic the only question is should it be called AVV or Augsburger Verkehrs- und Tarifverbund in `operator `and `network`

PR #4291, opened 2020-10-04
Added Margin Free Market

AnjaliGGeetha:


Adamant36:
Hello. Do you think it goes by "Margin Free Market" or some variation of "MarginFree" more in OSM?

AnjaliGGeetha:
The name is used inconsistently, but the registered one as per their website is MarginFree (https://www.mfreem.com/). So I am changing the brand name to MarginFree.

bhousel (changes requested):
Can you fix the name?  It definitely shouldn't be "MarginFree Market Pvt Ltd"..

AnjaliGGeetha:
Changed file.

PR #4294, opened 2020-10-04
Solved some conflicts

jvbraganca:
Hello,

How can I find more easily where the conflicted files are? I just see the names and sometimes it's in Russian or an Oriental language, and I can't read it nor identify what is the brand of the conflicted register.

I fixed some conflicts and added missing cuisine tags of the restaurants I was able to find.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -615,7 +615,7 @@
       "tags": {
         "amenity": "bank",
         "brand": "Banca Intesa",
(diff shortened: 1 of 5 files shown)


bhousel:
Sorry, I can't merge this.. It's not clear why the bank wikidata's were changed, the cuisine changes seem like guesses, we don't add schools to this index, and I'm confused about why the wikidata.json file changed too..  There's just too much changing in this pull request and it doesn't make sense to me.

PR #4299, opened 2020-10-04
Add Kokoriko

desiderantes:


Adamant36:
I don't think we add the delivery tag with anything because we would have to be 100% sure every place does delivery and that it wont change any time soon. Otherwise, it risks introducing bad data. So, are you certain every location does delivery?

desiderantes:
> I don't think we add the delivery tag with anything because we would have to be 100% sure every place does delivery and that it wont change any time soon. Otherwise, it risks introducing bad data. So, are you certain every location does delivery? 

Yup. They offer both their own delivery, and have agreements with iFood, so you can order from Kokoriko itself and the delivery can be fulfilled by Kokoriko itself or by iFood

bhousel (changes requested):
Please remove the `delivery` tag - we don't include it for other brands in this index.  Thanks!

Adamant36:
I removed it in f8d7610.

Issue #4265, opened 2020-10-03
Add Tulsa Transit.


BalooUriza:
Metropolitan Tulsa Transit Authority operates the bus services in Tulsa, Sand Springs, and Broken Arrow, Oklahoma.  Currently operates a single network.

PR #4300, opened 2020-10-05
Added Tulsa Transit to bus routes json

AgaskarJ:
Added Tulsa Transit to bus.json based on issue #4265 
Please check id generation or let me know how I can initiate generation

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -13572,7 +13572,9 @@
       "locationSet": {"include": ["us"]},
       "tags": {
         "network": "Metropolitan Tulsa Transit Authority",
-        "operator": "City of Tulsa",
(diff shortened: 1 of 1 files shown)


bhousel (changes requested):
From looking here: https://www.openstreetmap.org/relation/10259299
I think the `network` and `operator` should both be "Metropolitan Tulsa Transit Authority"
Please also supply the `network:wikidata` and `network:wikipedia` values, thanks!

(Don't worry too much about id generation - the build script will handle it.)

AgaskarJ:
@bhousel Thanks for your feedback! I previously got the Operator listed on the Wikipedia page, my bad. Added the changes!

bhousel:
Thanks for the quick fix @AgaskarJ & welcome to the project 👋

PR #4317, opened 2020-10-05
Added HokBen resturant

markgrichanik:


Adamant36:
Thanks for the PR. A few things though. The display name should be HokBen, not Hoka Hoka Bento. Although as I said in the issue Hoka Hoka Bento could be a match name. Also, you don't need to capitalize Japanese in the cuisine tag. Other then that it looks good though 👍

Adamant36:
Plus, it's suppose to be in the fast food file. Given that, I think I'm going to close. Feel free to open another issue for it with the changes I said though. Thanks 👍

PR #4326, opened 2020-10-05
Add wikidata for 2 shops/clothes and 4 transit/bus in France

Sukkoria:
At least I should be back on the right branch, with the right file structure this time. 
If that goes right, I will add more data on transit files!


bhousel:
Looks good, thanks @Sukkoria 👍

PR #4334, opened 2020-10-05
added PC-Spezialist

londonjob:


bhousel (changes requested):
For situations like this we usually create a unique identifier for the "PC-Spezialist" brand.  It would be weird to display a logo for Synaxon, even though that might be the parent company.

londonjob:
I added the parent company as a consulting brand instead to reflect the bigger business for Synaxon

bhousel (approved):

bhousel:
Thanks for the update @londonjob 👍

PR #4335, opened 2020-10-05
Improve BART and Delhi Metro entries

peternewman:


data/transit/route/subway.json
```diff
@@ -23,10 +23,14 @@
     {
       "displayName": "Delhi Metro",
       "id": "delhimetro-907487",
-      "locationSet": {"include": ["001"]},
+      "locationSet": {"include": ["in"]},
       "tags": {
         "network": "Delhi Metro",
-        "operator": "Delhi Metro",
+        "network:wikidata": "Q271195",
+        "network:wikipedia": "hi:दिल्ली_मेट्रो_रेल",
```
peternewman, line 30: I think I'm right in thinking we should use the local language by default?

data/transit/route/subway.json
```diff
@@ -23,10 +23,14 @@
...
+        "operator": "Delhi Metro Rail Corporation Limited",
+        "operator:wikidata": "Q1029339",
+        "operator:wikipedia": "en:Delhi Metro Rail Corporation",
```
peternewman, line 33: There isn't one for this.

data/transit/route/subway.json
bhousel: > I think I'm right in thinking we should use the local language by default?

Yes, I agree 👍

data/transit/route/subway.json
bhousel: Yeah, we try to use local language unless it's not available or the local page is clearly a stub.
Really the `*:wikipedia` tags are somewhat redundant once the `*:wikidata` is there, but some people and softwares will linkify them, so we keep them around.

peternewman:
Sorry I've just landed 11bcfc3f5ee01820bccfa0f0d7566347cca84d20 after your review.

data/transit/route/subway.json
```diff
@@ -3,10 +3,14 @@
...
       "displayName": "BART",
       "id": "bart-907487",
-      "locationSet": {"include": ["001"]},
+      "locationSet": {"include": ["us"]},
       "tags": {
         "network": "BART",
-        "operator": "BART",
+        "network:wikidata": "Q610120",
+        "network:wikipedia": "en:Bay Area Rapid Transit",
+        "operator": "Bay Area Rapid Transit District",
```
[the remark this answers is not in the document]
bhousel: Interesting! Whatever you want to do is fine here.. I love that people are putting a lot more care and thought into transit tagging because of this.. 😁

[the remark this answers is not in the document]
peternewman: Like the existing brand stuff, it just makes sense, get it right once, get it right everywhere.

PR #4336, opened 2020-10-05
Check VBB transit

britiger:
The Verkehrsverbund Berlin-Brandenburg is a network only. There are different operators which can be added later.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -11307,11 +11307,12 @@
     },
     {
       "displayName": "Verkehrsverbund Berlin-Brandenburg",
-      "id": "verkehrsverbundberlinbrandenburg-a45453",
-      "locationSet": {"include": ["001"]},
+      "id": "verkehrsverbundberlinbrandenburg-3b9544",
+      "locationSet": {"include": ["de"]},
       "tags": {
         "network": "Verkehrsverbund Berlin-Brandenburg",
-        "operator": "Verkehrsverbund Berlin-Brandenburg",
+        "network:short": "VBB",
+        "network:wikidata": "Q315451",
         "route": "bus"
       }
     },
--- a/data/transit/route/subway.json
+++ b/data/transit/route/subway.json
@@ -271,11 +271,12 @@
     },
     {
       "displayName": "Verkehrsverbund Berlin-Brandenburg",
-      "id": "verkehrsverbundberlinbrandenburg-907487",
-      "locationSet": {"include": ["001"]},
+      "id": "verkehrsverbundberlinbrandenburg-447f34",
+      "locationSet": {"include": ["de"]},
       "tags": {
         "network": "Verkehrsverbund Berlin-Brandenburg",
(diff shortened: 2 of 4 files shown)


data/transit/route/bus.json
```diff
@@ -11307,11 +11307,12 @@
...
     {
       "displayName": "Verkehrsverbund Berlin-Brandenburg",
-      "id": "verkehrsverbundberlinbrandenburg-a45453",
-      "locationSet": {"include": ["001"]},
+      "id": "verkehrsverbundberlinbrandenburg-3b9544",
+      "locationSet": {"include": ["de"]},
       "tags": {
         "network": "Verkehrsverbund Berlin-Brandenburg",
-        "operator": "Verkehrsverbund Berlin-Brandenburg",
+        "network:short": "VBB",
```
peternewman, line 11314: I see this is popular in Germany but not really mentioned on the wiki aside from there, it could work well it bits of the UK too if it's semi official.

britiger: The list of the German transit networks can be found here: https://wiki.openstreetmap.org/wiki/Germany/Nahverkehr#.C3.96PNV-Verkehrsverb.C3.BCnde_in_Deutschland

A good and clean list of some operators of this network for one city: https://wiki.openstreetmap.org/wiki/Potsdam/Transportation

I know, the page https://wiki.openstreetmap.org/wiki/Verkehrsverbund_Berlin-Brandenburg needs some cleanup, I'll put it on the agenda for our next local online meetup for berlin area this friday. :smiley: 

Or should I remove the "Verkehrsverbund Berlin-Brandenburg" entity and add every single operator with this network? My idea was to add the "generic" network entity first. After this to found and check all local operators of the network and add them.

Note: In Berlin there are also ferries which are in the network: https://www.openstreetmap.org/relation/2690005

bhousel: Yeah I noticed these `network:short` and `network:guid` tags also.  `network:wikidata` is the best identifier, but we can definitely set these other tags if it's something that mappers are using.

bhousel (approved):

bhousel:
Thanks @britiger 👍 
I also added corresponding `network:wikipedia` tags in ef22b4d6.  If you can include this information when you look up the `network:wikidata` tag, that would be great too!

data/transit/route/bus.json
peternewman: Should we be doing the same thing with e.g. BART, MBTA etc then @bhousel , or only if people are already using it locally?

I guess I'm saying should `network` correspond to how it's locally signed or what? I guess also will iD and others use these other keys like matchNames or would we need to double up there?

bhousel: There's this page: https://wiki.openstreetmap.org/wiki/Key:network:guid but it doesn't give much detail.  I think we should only include `network:guid` or `network:short` if communities are using them already, but not invent values for e.g. BART, MBTA.

peternewman: Sounds sensible to me!

PR #4338, opened 2020-10-05
Added Eilles to confectionery

londonjob:


bhousel:
This points to the Arko Q673647 - was that intentional?  We normally make a different qid if the brand is different, because it would be weird to show a "Arko" logo on an Eilles map marker.

londonjob:
Eilles has no own wikipedia entry but because Eilles was bought by Arko in 2016 I was refering to them. So in this case should I rather remove the reference to the Wikipedia page from Arko?

bhousel:
> Eilles has no own wikipedia entry but because Eilles was bought by Arko in 2016 I was refering to them. So in this case should I rather remove the reference to the Wikipedia page from Arko?

The right thing to do probably depends - Is Eilles still operating as a business?
- If it was bought in 2016, and rebranded to "Arko" and the stores no longer say "Eilles", then the right thing to do is add a `"matchNames": ["eilles"]` property to Arko.
- If there are still retail storefronts and social media presence for the "Eilles" brand, then there should be a wikidata page created for it, with some supporting references, link to the website, social media, etc, and then we'd add it to our index with a link to that "Eilles" qid.

londonjob:
There are Eilles stores in Germany so I created a wikidata site here https://www.wikidata.org/wiki/Q99976174 and updated the pull request.

bhousel:
> There are Eilles stores in Germany so I created a wikidata site here https://www.wikidata.org/wiki/Q99976174 and updated the pull request.

Perfect!  Thanks for taking care of it so quickly 👍

PR #4340, opened 2020-10-05
Added "The Buses of Somerset" to bus.json

UKChris-osm:
Added "The Buses of Somerset" to bus.json


bhousel:
LGTM, thanks @UKChris-osm 👍

PR #4345, opened 2020-10-05
Add wikidata/wikipedia for transit networks + country codes

Sukkoria:
* Augsburg (Germany)
* Granada (Spain)
* Léon (Spain)
* Blois (France)
* Savoie (France)
* Bus en Seine (France)
* Cholet (France)


Identitaet:
Looks good and there are no acronym shenenigans so we can easily merge it

PR #4347, opened 2020-10-05
Add country codes + wikidata/wikipedia for trolleybus networks

Sukkoria:
I tried to find the right codes for network names in latin script (avoiding acronyms), added some wikidata/wikipedia while doing so.


bhousel:
Thanks @Sukkoria looks great 👍

PR #4350, opened 2020-10-05
Added wiki tags

camilorecce:
Added wiki tags for "Bangkok Bank", "Komercijalna banka" and "Old National Bank" in bank.json


Adamant36:
Thanks. Looks good 👍

PR #4361, opened 2020-10-06
Added Rolls to brands index

KennethYoel:


data/brands/amenity/restaurant.json
```diff
@@ -2613,6 +2613,19 @@
...
     },
+    {
+      "displayName": "Rolls",
+      "id": "rolls-fe858e",
+      "locationSet": {"include": ["fi"]},
+      "tags": {
+        "amenity": "restaurant",
+        "brand": "Rolls",
+        "brand:wikidata": "Q3195068",
+        "brand:wikipedia": "en:Rolls (restaurant chain)",
```
Identitaet, line 2624: If it's a finish brand we should use finish Wikipedia

Identitaet (changes requested):

data/brands/amenity/restaurant.json
KennethYoel: I have thought the same as well but when I searched for Rolls in Finnish Wikipedia the page doesn't exist, only within the English Wikipedia that you can find that restaurant.

bhousel: The [wikidata page for Rolls](https://www.wikidata.org/wiki/Q3195068) shows that there is one:
(this is where I usually go to look it up)

Screen Shot 2020-10-06 at 9 57 01 AM

KennethYoel: 🤦🏻‍♂️ Thank you, I’ll get that updated.

bhousel:
Thanks @KennethYoel and welcome to the project 👋

PR #4364, opened 2020-10-06
Add Cercanías networks in Spain

Robot8A:


Identitaet:
What are we waiting for isn't es=Spain a normal accepted country code ?

Identitaet:
Oh you're still working on it 🤦

Robot8A:
The plan is to add the different Cercanías (commuter rail) networks in Spain. Most may be straight-forward to add, but some I still don't know how to deal with them.

PR #4366, opened 2020-10-06
Add some fitness centers

Sukkoria:
I checked the fitness centers in the nsi file against the Wikidata query: https://w.wiki/fTh and added the missing ones.

### DIFF
--- a/data/brands/leisure/fitness_centre.json
+++ b/data/brands/leisure/fitness_centre.json
@@ -252,20 +252,26 @@
     },
     {
       "displayName": "KeepCool",
-      "locationSet": {"include": ["fr", "be", "lu"]},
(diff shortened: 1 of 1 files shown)


bhousel:
This is great, thanks @Sukkoria 👍 
I like the idea of using SPARQL queries to look up businesses that our index is missing.  We haven't done any of this yet, but I bet you can find a lot of missing brands that way.

Sukkoria:
Thanks! I tried this method for fitness centers because there seemed to be a dedicated wikidata item ("fitness center chain") but it may be possible to use the WD property "OSM tag" (P1282) to generalize to other cases… I may dig further into the subject or open an issue with some more elements one day :)

PR #4367, opened 2020-10-07
Add El Corral

desiderantes:
Not sure if it's properly tagged. Officially, it's called "Hamburguesas El Corral", but everybody knows it as "El Corral", and even the logo just says "El Corral"


bhousel:
Looks great, thanks @desiderantes 👍

PR #4370, opened 2020-10-07
added brand sakae sushi

pushkar1393:
- Added entry under fast_food.json for `Sakae Sushi`. 
- I think there was one entry apart from it which lacked an Id, it was generated when I build the code so I have kept it as is and not reverted that change.

Please let me know if anything needs to be modified.


Adamant36:
Thanks. Looks fine 👍

Issue #4353, opened 2020-10-06
Add Albaik


Adamant36:
Saudi Arabian fast food chicken chain. Goes in the fast food file.

name:en=Albaik
name:ar=البيك‎
(Same for brand tags)

cuisine=chicken

https://en.wikipedia.org/wiki/Albaik
https://www.wikidata.org/wiki/Q4703536

PR #4374, opened 2020-10-07
added brand albaik

pushkar1393:
- Added entry under fast_food.json for `Albaik`.
- Related issue - #4353 

Please let me know if anything needs to be modified.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -5026,7 +5026,7 @@
         "brand:ar": "البيك",
         "brand:en": "Albaik",
         "brand:wikidata": "Q4703536",
-        "brand:wikipedia": "en:Albaik",
+        "brand:wikipedia": "ar:البيك",
         "cuisine": "chicken",
         "name": "البيك",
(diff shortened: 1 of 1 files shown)


Adamant36:
Thank you. It looks fine. BTW, if put "closes" and the issue number in your PR comment then it will automatically close the issue the PR is related to when it's merged.

Adamant36:
> Thank you. It looks fine.

Actually, I guess it has a conflict. So, I'll wait for someone else to come around and look at it before I merge it.

pushkar1393:
Thanks, for the quick reply @Adamant36. I checked the conflict was due to malformation in the `fast_food.json` file, due to a mismatch with the upstream since the request was raised earlier. I have resolved the conflict.

pushkar1393:
Out of curiosity how does one decide what tag value should be provided for "brand:Wikipedia" if there is a wiki page in English and in Arabic (or any other language.)

Identitaet:
We usually try to use the "local language"
And if that isn't possible usually English

PR #4378, opened 2020-10-07
Update bank.json

debyos:
After the merger of the many small "takarékszövetkezet" in Hungary, the resulting bank's name is Takarékbank.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -7130,15 +7130,15 @@
       }
     },
     {
-      "displayName": "Takarékszövetkezet",
+      "displayName": "Takarékbank",
       "id": "takarekszovetkezet-60884a",
       "locationSet": {"include": ["hu"]},
       "tags": {
         "amenity": "bank",
-        "brand": "Takarékszövetkezet",
+        "brand": "Takarékbank",
         "brand:wikidata": "Q30324674",
-        "brand:wikipedia": "en:TakarékBank",
-        "name": "Takarékszövetkezet"
+        "brand:wikipedia": "en:Takarékbank",
+        "name": "Takarékbank"
       }
     },
     {


Adamant36:
How long ago was the merge?

data/brands/amenity/bank.json
```diff
@@ -7130,15 +7130,15 @@
       }
     },
     {
-      "displayName": "Takarékszövetkezet",
+      "displayName": "Takarékbank",
       "id": "takarekszovetkezet-60884a",
       "locationSet": {"include": ["hu"]},
```
peternewman, line 7135: This will also capture the old name too, if every Takarékszövetkezet became a Takarékbank?

```suggestion
      "locationSet": {"include": ["hu"]},
      "matchNames": [
        "Takarékszövetkezet"
      ],
```

Identitaet: Yes

Adamant36:
Just an FYI, we aren't really solid on changing names for entries due to mergers unless a lot of time has past and we can be sure the  signs on the ground have changed. A lot of times it either takes years for to change signs or they just don't change them and ground verifiability is core to OSM. So, if the merger just happened, or it's just a name change on paper, it would likely be best to either hold off on changing the entry for now or just leave it as is.

debyos:
This merger happened in several steps, the last one about two years ago. Actually, it was a political decision stipulated by a law, and it was implemented quickly and thoroughly. So nowadays all logos on all branches are of "Takarékbank". Nevertheless, on OSM you should look for [Takarékszövetkezet](http://overpass-turbo.eu/s/YRU) as well to find, there 

An alternative name is [Takarékpénztár](http://overpass-turbo.eu/s/YRS). Of course, there are also some entries which has nothing to do with a current bank (e.g. there are city houses built by these banks a century ago and they bear their names still today) but this may apply to all brands, I think.

Identitaet:
> So nowadays all logos on all branches are of "Takarékbank".
Sounds good so there also is value in helping change over the data

Adamant36:
> Sounds good so there also is value in helping change over the data

Totally. As long as they have changed on the ground there shouldn't be a problem with it.

bhousel:
Sounds good 👍   
I added matchNames for the older names in 2a1511ce

PR #4376, opened 2020-10-07
Resolves #4357 Add tom n toms

TheDynamicPunk:
Add tom n toms to brands/amenity/cafe

Let me know if it looks fine!

PR #4379, opened 2020-10-07
Revert "Resolves #4357 Add tom n toms"

### LINKED ISSUE #4357: Add Tom N Toms
South Korean coffee shop chain. 

name:en=Tom N Toms
name:ko=탐앤탐스
(also with the brand tags)

https://en.wikipedia.org/wiki/Tom_N_Toms
https://www.wikidata.org/wiki/Q18142418

Identitaet:
Reverts osmlab/name-suggestion-index#4376

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -1433,36 +1433,6 @@
       }
     },
     {
-      "displayName": "Tom N Toms",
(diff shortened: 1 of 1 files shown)


Identitaet:
Thought i misread something
But no it was correct

PR #4380, opened 2020-10-07
Bus transit networks of Baden-Württemberg, Germany

Karthoo:
Added network tags (:short, :guid, :wikidata, :wikipedia) and locations
Removed some duplicates

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -1169,12 +1169,17 @@
       }
     },
     {
-      "displayName": "bodo",
+      "displayName": "Bodensee-Oberschwaben Verkehrsverbund",
       "id": "bodo-a45453",
-      "locationSet": {"include": ["001"]},
+      "locationSet": {"include": ["de"]},
+      "matchNames": [
+        "bodo"
+      ],
       "tags": {
-        "network": "bodo",
-        "operator": "bodo",
+        "network": "Bodensee-Oberschwaben Verkehrsverbund",
+        "network:short": "bodo",
+        "network:wikidata": "Q889680",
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @Karthoo for doing all this work 👍

bhousel:
@Karthoo I just ran the build script again in 9f7f89e1 and it looks like a few of the removed items may have been put back.

Things like these:
```js
    {
      "displayName": "Verkehrsverbund Großraum Nürnberg",
      "id": "verkehrsverbundgrossraumnurnberg-a45453",
      "tags": {
        "network": "Verkehrsverbund Großraum Nürnberg",
        "operator": "Verkehrsverbund Großraum Nürnberg",
        "route": "bus"
      }
    },
    {
      "displayName": "Verkehrsverbund Großraum Nürnberg GmbH",
      "id": "verkehrsverbundgrossraumnurnberggmbh-a45453",
      "tags": {
        "network": "Verkehrsverbund Großraum Nürnberg GmbH",
        "operator": "Verkehrsverbund Großraum Nürnberg GmbH",
        "route": "bus"
      }
    },
```

When that happens, it's because the build script sees that `"network": "Verkehrsverbund Großraum Nürnberg GmbH"` is a commonly occurring tag that it can not match it to an item that we already know about.  So it creates a "new" item with that `network` and `operator` and a worldwide '001' locationSet.

If you want these items to go away, the values like "Verkehrsverbund Großraum Nürnberg GmbH" need to match something in the index, either via a tag or a matchName.  Then you can delete them and they won't come back again.  Make sense?

Karthoo:
Okay, thanks for the advise, I didn't know you're re-running the script regularly. I'll put duplicates into matchNames in future.

PR #4381, opened 2020-10-07
added India Post

pearlslugs:


pearlslugs:
Sorry about the uselessness of this, all the other brands were added from the ready for check in wiki thing

Adamant36:
Thanks for the PR. Nothing is useless about it. Every brand we can add improves things a lot. That said, there's no need for the match name. Outside of that its good though.

bhousel:
Thanks @pearlslugs & welcome 👋

PR #4386, opened 2020-10-08
Train networks A-C

Karthoo:
added details like location, network and operator tags and matchNames


bhousel:
Looks good to me, thanks again @Karthoo 👍

PR #4387, opened 2020-10-08
Added wiki tags to supermarket.json

camilorecce:
Added "Choppies", "Key Food" and "Natural Grocers" wiki tags to supermarket.json

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -946,6 +946,8 @@
       "locationSet": {"include": ["001"]},
       "tags": {
         "brand": "Choppies",
+        "brand:wikidata": "Q5104860",
+        "brand:wikipedia": "en:Choppies",
         "name": "Choppies",
         "shop": "supermarket"
       }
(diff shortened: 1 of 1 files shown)


data/brands/shop/supermarket.json
```diff
@@ -946,6 +946,8 @@
       "locationSet": {"include": ["001"]},
       "tags": {
         "brand": "Choppies",
+        "brand:wikidata": "Q5104860",
```
matkoniecz, line 949: "Area served | Botswana, Zimbabwe, Zambia" - https://en.wikipedia.org/wiki/Choppies 

Restricting locationSet may be nice to keep it from appearing elsewhere.

See https://github.com/osmlab/name-suggestion-index/blob/5e55b07e746c35337f38cbb20fccff1bfe64ed47/CONTRIBUTING.md#locationset

data/brands/shop/supermarket.json
```diff
@@ -2842,6 +2844,8 @@
       "locationSet": {"include": ["001"]},
```
matkoniecz, line 2844: https://en.wikipedia.org/wiki/Key_Food - it seems to be US only

PR #4392, opened 2020-10-09
Added wiki tags to supermarket.json

camilorecce:
Added "Choppies", "Key Food" and "Natural Grocers" wiki tags to supermarket.json, and fixed the locations in #4387


Identitaet:
Looks good and you addressed the issues 👍

PR #4423, opened 2020-10-09
Add Wikipedia, Wikidata and network GUID to MVV and VVS

FloEdelmann:


data/transit/route/bus.json
```diff
@@ -6234,7 +6234,12 @@
       "locationSet": {"include": ["001"]},
       "tags": {
         "network": "Münchner Verkehrs- und Tarifverbund",
+        "network:guid": "DE-BY-MVV",
+        "network:wikidata": "Q259000",
+        "network:wikipedia": "en:Münchner_Verkehrs-_und_Tarifverbund",
```
FloEdelmann, line 6239: Should this rather link to the local (German) Wikipedia page?

data/transit/route/light_rail.json
```diff
@@ -148,7 +148,12 @@
       "locationSet": {"include": ["001"]},
       "tags": {
         "network": "Verkehrs- und Tarifverbund Stuttgart",
+        "network:guid": "DE-BW-VVS",
+        "network:wikidata": "Q2516108",
+        "network:wikipedia": "de:Verkehrs- und Tarifverbund Stuttgart",
```
FloEdelmann, line 153: Or should this rather link to the English Wikipedia page? (I copied these three tags from `bus.json`.)

data/transit/route/bus.json
bhousel: Yes, German stuff should link to the German wikipedia page, if there is one.

A few other things I noticed:
- make sure to update the `locationSet` to only include "de"
- replace the underscores with spaces in the wikipedia sitelinks

bhousel (changes requested):
Welcome @FloEdelmann 👋 
Looks like you've really done your research, thank you! 
Just a few things to change:
* make sure to update the locationSet to only include "de"
* replace the underscores with spaces in the wikipedia sitelinks

Karthoo:
I would also recommend to add "network:short": "MVV" (/VVS) and remove the operator tags because the lines within the networks are operated by sub-contractors

FloEdelmann:
> I would also recommend to add "network:short": "MVV" (/VVS)

@Karthoo I considered this, as I've seen it for other networks, but the tag is not documented in the OSM wiki. I'll add it now, though.

> and remove the operator tags because the lines within the networks are operated by sub-contractors

Alright, will do.

bhousel (approved):

bhousel:
Thanks for the quick fixes @FloEdelmann 👍

PR #4424, opened 2020-10-09
Fix german newspaper vending machines

fxedel:
This is a hacktoberfest-inspired PR :tada:


bhousel:
Thanks @fxedel looks good and welcome 👋 
Happy Hacktoberfest 👻

fxedel:
Hey @bhousel, thanks for merging that fast! :rocket: 

Just one question: Why is the `dist` directory git-tracked? In `sitemap.xml`, for example, every single `<lastmod>` was updated to the current time. Doesn't this produce lots of merge conflicts?

bhousel:
> Just one question: Why is the `dist` directory git-tracked? In `sitemap.xml`, for example, every single `<lastmod>` was updated to the current time. Doesn't this produce lots of merge conflicts?

We have to put the sitemap somewhere and check it in because http://nsi.guide is served on GitHub pages.  We also check in build products so the page can see them.

On most of the projects that I maintain, I've settled into a philosophy/workflow like:
- `npm run test`
   - Automate this!  Run on every commit, includes things like schema validation, code testing
   - Important: don't change anything here. Running tests must not have side effects.
   - Tests are nice and we should have more of them
- `npm run build` 
   - Don't expect your users to do this, and don't automate it.
   - Build whatever is needed for the project to actually do something.. 
   - Can break sometimes, that's ok.  Really, don't stress about broken builds.
   - Produce lots of warnings. Warnings are not problems to be fixed necessarily - they are signal for the maintainer. 
   - Clean stuff up - This introduces churn/conflicts but makes the project mostly "self healing"
 - `npm run dist`
   - Build project outputs in folder like `dist/*` (do this infrequently).
   - Should not produce broken output!  In other words, whatever is produced here should be mostly able to `npm publish` right away
   - Project should ideally have a GitHub page on `docs/*` so people can preview it or try it out.


So on this project, tests are run by Travis, builds are run probably a few times a day, dist is run maybe once a week or more frequently if I want contributors to see their work reflected on http://nsi.guide 

Hope that's ok!

PR #4425, opened 2020-10-09
Fast food

alefuedev:
I just add Burger's Priest to /data/brands/amenity/fast_food.json
https://theburgerspriest.com/


data/brands/amenity/fast_food.json
```diff
@@ -4281,6 +4281,18 @@
         "name": "Thai Express",
         "takeaway": "yes"
       }
+    },
+     {
+      "displayName": "The Burgers's Priest)",
```
tas50, line 4286:
```suggestion
      "displayName": "The Burgers's Priest",
```

data/brands/shop/hardware.json
```diff
@@ -18,6 +18,23 @@
         "shop": "hardware"
       }
     },
+    {
+      "displayName": "Do it Best",
```
tas50, line 22: I'm not sure about the name here. All the locations around me are signed and referenced as XYZ Do it Best Hardware

PR #4432, opened 2020-10-09
Resolves #4416 Add Southern Fried Chicken

### LINKED ISSUE #4416: Add Southern Fried Chicken
Global fast food fried chicken chain. 

cuisine=chicken

https://en.wikipedia.org/wiki/Southern_Fried_Chicken
https://www.wikidata.org/wiki/Q7569934

TheDynamicPunk:
Add Southern Fried Chicken to the list

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -3861,6 +3861,20 @@
       }
     },
     {
+      "displayName": "Southern Fried Chicken",
+      "id": "southernfriedchicken-a6761c",
+      "locationSet": {"include": ["in", "uk"]},
+      "tags": {
+        "amenity": "fast_food",
+        "brand": "Southern Fried Chicken",
+        "brand:wikidata": "Q7569934",
+        "brand:wikipedia": "en:southern fried chicken",
+        "cuisine": "chicken",
+        "name": "Southern Fried Chicken",
+        "takeaway": "yes"
+      }
+    },
+    {
       "displayName": "Specialty's",
       "id": "specialtys-4d2ff4",
       "locationSet": {"include": ["us"]},


TheDynamicPunk:
@Adamant36 Should I add 001 to the locationSet for Global or the countries that I've added, are alright?

bhousel:
> @Adamant36 Should I add 001 to the locationSet for Global or the countries that I've added, are alright?

The locationSet should include only the countries where the business exists.  

`001` is the [global region code](https://en.wikipedia.org/wiki/UN_M49), so that would override everything and make the brand available everywhere.  For most items in the name-suggestion-index we probably don't want this.  

Ideally, we would specify well-scoped locationSets for all the brands.

TheDynamicPunk:
In #4416, I was suggested this

> No just put 001. I think that's location code for the globe.

by @Adamant36. So that's why I was asking, because I've not added the complete location set of Southern Fried Chicken as it's pretty vague in Wikipedia, with terms like Europe, Asia-Pac etc.

Please check @bhousel

Adamant36:
The country codes you added are fine. It's not like we can't add more later if need be. I'm still not sure when it's better to consider something a "global" brand versus it just existing in a lot of places.

bhousel:
Agree, it's always ok to change it later.
Also, if it makes more sense, you can use the [M.49 region codes](https://en.wikipedia.org/wiki/UN_M49), for example `'150'` for Europe, `'142'` for Asia, etc. instead of specifying a whole lot of country codes.

TheDynamicPunk:
Ohkay, that makes sense. I just wanted to draw your attention if everything was fine. I'd hate to break anything by my contributions😅

PR #4441, opened 2020-10-10
Add the Universite de Montreal to data/brands/amenity/university.json

alefuedev:
https://www.umontreal.ca/l-udem/

### DIFF
--- a/data/brands/amenity/university.json
+++ b/data/brands/amenity/university.json
@@ -26,6 +26,19 @@
         "short_name": "Strayer"
       }
     },
+{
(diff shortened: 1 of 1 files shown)


Adamant36:
Does this have a lot of satellite campuses or is it just the one main campus?

bhousel:
Sorry @alefuedev, but we don't add individual universities to this index - we only add the kind of for-profit "universities” that operate like a branded business in multiple locations, such as DeVry or University of Phoenix. I know this is an unusual concept outside of the USA - the “universities” in this index are really more like training centers than true universities.

Issue #4403, opened 2020-10-09
add Westfalenbahn (easy)


Identitaet:
https://de.wikipedia.org/wiki/Westfalenbahn network
https://de.wikipedia.org/wiki/Abellio_(Unternehmen) operator
Area DE only

Identitaet:
Also add matchName WFL

shubh710:
Hi @Identitaet ,
I would like to work on this but a little guidance will be required as I am new to OSS.

Identitaet:
Sure feel free to ask any questions
As a tip I'd suggest you look at #4414 which is quite similar

shubh710:
Thanks @Identitaet for tip it was quite helpful.
I have submitted PR #4446 
Please review and let me know if it is okay or anything is missing.
Thanks again.

PR #4446, opened 2020-10-11
Added Westfalenbahn to Transit

shubh710:
Related to Issue https://github.com/osmlab/name-suggestion-index/issues/4403

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -1775,6 +1775,20 @@
       }
     },
     {
+      "displayName": "Westfalenbahn",
+      "id": "westfalenbahn-da20e0",
+      "locationSet": {"include": ["de"]},
+      "tags": {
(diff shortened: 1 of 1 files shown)


Identitaet:
Almost perfect but the operator tag should only be Abelio cause what's in the brackets is disambiguation for Wikipedia

Adamant36:
> Almost perfect but the operator tag should only be Abelio cause what's in the brackets is disambiguation for Wikipedia

Plus there shouldn't be an underscore for the operator name or the Wikipedia title.

shubh710:
Hi @Identitaet @Adamant36 
my bad, since I can see you have merged previous request do you want me to resolve mentioned issues and submit another PR?

shubh710:
I can see you have rectified issues yourself. :)
Thanks.

PR #4456, opened 2020-10-11
Add ferries vbb

britiger:
Within the network VBB (Verkehrsverbund Berlin-Brandenburg) there are two operators of ferries. Cause of the conflict with the same `id` if the `network` and `locationSet` is the same. I created new features for the cities Berlin and Potsdam. The features are not depend of the transit networks so I used generic names for this.


bhousel:
> Within the network VBB (Verkehrsverbund Berlin-Brandenburg) there are two operators of ferries. Cause of the conflict with the same `id` if the `network` and `locationSet` is the same. I created new features for the cities Berlin and Potsdam. The features are not depend of the transit networks so I used generic names for this.

Cool, yes this looks great to me.. Giving the two VBB ferries their own locations where they operate is a good solution 👍

PR #4457, opened 2020-10-11
Add McLean's pub to data/brands/amenity/pub.json

alefuedev:
Add McLean's pub.
http://mcleanspub.com/

### DIFF
--- a/data/brands/amenity/pub.json
+++ b/data/brands/amenity/pub.json
@@ -59,6 +59,18 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


bhousel:
Sorry @alefuedev , as far as I can tell this is just a single pub.  This index is for the sorts of things where there are lots of them in OpenStreetMap and we want to encourage people to tag them consistently.

alefuedev:
Got it!

PR #4461, opened 2020-10-11
Add Smoke's Poutinerie to data/brands/amenity/fast_food.js

alefuedev:
Add Smoke's Poutinerie to  data/brands/amenity/fast_food.js


bhousel:
Looks great @alefuedev , thanks! 👍

PR #4471, opened 2020-10-12
Add Duck duck go as an option for the search functionallity

Identitaet:
1. Is this a slippery slope to adding all Search engines ?
I'd like to have ddg cause of privacy 
We could also search wikipedia with DDG instead of Google with the excact same syntax for even more privacy

### DIFF
--- a/app/src/CategoryRow.jsx
+++ b/app/src/CategoryRow.jsx
@@ -98,6 +98,7 @@ relation[${k}=${v}][network=${n}][network:wikidata=${qid}]
       <div className='viewlink'>
         { searchOverpassLink(n, overpassQuery) }<br/>
         { searchGoogleLink(n) }<br/>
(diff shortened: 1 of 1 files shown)


Adamant36:
DDG has semi sub par search results don't they?

Identitaet:
It's not as good as Google but in my opinion good enough most times.
That's also why I wouldn't want to remove the Google option

Adamant36:
Also, from what I've read there's no proof that DuckDuckGo is even private. It's not open source and as far as I know they just have a "take our word for it" policy. So, you might want to research it more.

Identitaet:
Yeah I just took their word for it.
I guess I'll close it for now

bhousel:
Hmm, I definitely don't want to add more links here.  I think it's fine for people to use whatever search engine they want when researching an unmatched brand - these links are just there for convenience.

PR #4480, opened 2020-10-13
Added Metro Trains Melbourne to transit

orangegrove1955:
Added Metro Trains Melbourne to transit data, including wikidata and wikipedia


bhousel:
Thanks @orangegrove1955 , looks good 👍

PR #4486, opened 2020-10-13
Add wikidata

britiger:
Add existing wikidata for Pfennigpfeifer and Adler

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -130,7 +130,7 @@
       "tags": {
         "brand": "Adler",
         "brand:wikidata": "Q358196",
-        "brand:wikipedia": "de:Adler_Modemärkte",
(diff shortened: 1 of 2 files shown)


data/brands/shop/clothes.json
```diff
@@ -129,6 +129,8 @@
       },
       "tags": {
         "brand": "Adler",
+        "brand:wikidata": "Q358196",
+        "brand:wikipedia": "de:Adler_Modemärkte",
```
bhousel, line 133:
```suggestion
        "brand:wikipedia": "de:Adler Modemärkte",
```

bhousel (changes requested):
Looks mostly good - only change is that the underscores need to be replaced by spaces in the `brand:wikipedia` tag value 👍

Issue #4490, opened 2020-10-14
Add Alsford


Adamant36:
GB based DIY chain. 

shop=doityourself

https://www.wikidata.org/wiki/Q100298902

PR #4492, opened 2020-10-14
added Alsford to DIY shops

lelelew:
Added Alsford to the list of DIY shops.

Fixes: https://github.com/osmlab/name-suggestion-index/issues/4490


Identitaet:
looks good

Issue #4489, opened 2020-10-14
Add Hirebase


Adamant36:
GB based tool hire chain. 

shop=tool_hire

https://www.wikidata.org/wiki/Q100297859

PR #4493, opened 2020-10-14
added Hirebase to tool hire shops

lelelew:
Added Hirebase to the list of tool hire shops.

Fixes: https://github.com/osmlab/name-suggestion-index/issues/4489


Identitaet:
looks good

Issue #4488, opened 2020-10-14
Add Buildbase


Adamant36:
GB based DIY chain. 

shop=doityourself 

https://en.wikipedia.org/wiki/Buildbase
https://www.wikidata.org/wiki/Q4986416

PR #4494, opened 2020-10-14
added Buildbase to DIY shops

lelelew:
Added Buildbase to the list of DIY shops.

Fixes: https://github.com/osmlab/name-suggestion-index/issues/4488


Identitaet:
looks good

PR #4496, opened 2020-10-14
Add wiki tags for in Philadelphia, including SEPTA/PATCO

bhousel:
This Pull Request adds the network and operator tags for Philadelphia transit networks.  
- SEPTA (bus, tram, trolleybus) - Q2037863
- SEPTA (light_rail, subway) - Q11158637
- PATCO Speedline (subway) - Q2043730

The tagging of SEPTA's light_rail and subway is unusual in OpenStreetMap in that it already has `network=SEPTA` but `network:wikidata=Q11158637`  'Philadelphia subway'.  This QID does not have an English language wiki, and is only used to group the light_rail and subway networks.  I don't really agree with this, but I'm going to keep it that way in NSI for now.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -8817,11 +8817,18 @@
     },
     {
       "displayName": "SEPTA",
-      "id": "septa-a45453",
-      "locationSet": {"include": ["001"]},
+      "id": "septa-a21cc9",
+      "locationSet": {"include": ["us"]},
+      "matchNames": [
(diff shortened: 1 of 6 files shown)


bhousel:
Update on this - I'm from Philadelphia myself and reached out to the original editor of this data and some other Philadelphia mappers and we've agreed that it's ok to just use Q2037863 consistently across the entire SEPTA network.

PR #4497, opened 2020-10-14
Add hotels

britiger:
- Add Dorint, NH Hotels, Motel One
 - B&B Hotels are also in other countries than France


bhousel:
Thanks @britiger , lgtm 👍

PR #4498, opened 2020-10-14
addding l.l.bean

mneko:


bhousel:
Thanks @mneko and welcome 👋 
Kind of surprised we didn't have L.L.Bean in the index already 😆

mneko:
I was shocked too, but this might explain how much trouble I was having with the iD editor.

I'll be dropping in more chains as I remember them

data/brands/shop/outdoor.json
```diff
@@ -144,6 +144,21 @@
         "shop": "outdoor"
       }
     },
+    {
+      "displayName": "L.L.Bean",
+      "id": "llbean-783431",
+      "locationSet": {
+        "include": ["jp", "us"]
```
tas50, line 151: They're in CA too

bhousel:
> I was shocked too, but this might explain how much trouble I was having with the iD editor.
> I'll be dropping in more chains as I remember them

Thanks! We're happy to add any chains that are missing..
http://nsi.guide is a good resource to check what we have.
iD is a little bit behind what we have, but we will catch it up soon.

PR #4499, opened 2020-10-14
adding a few more regional brands, fixing the reported location error

mneko:


data/brands/amenity/fast_food.json
```diff
@@ -2656,6 +2671,21 @@
         "takeaway": "yes"
       }
     },
+    {
+      "displayName": "Milio's",
```
tas50, line 2675: It seems like this should be the full name since they use Milio's Sandwiches on all the signage, social media and website.

tas50:
LGTM. Thanks @mneko

PR #4501, opened 2020-10-14
Add Alsa wikidata and wikipedia 

ItaloSa:
I've created a wikidata and wikipedia entity for Alsa at transit/route/bus


bhousel:
Looks great @ItaloSa , thanks again 👍

PR #4503, opened 2020-10-14
more chains

mneko:


data/brands/shop/outdoor.json
```diff
@@ -93,6 +93,18 @@
         "shop": "outdoor"
       }
     },
+    {
+      "displayName": "Fjällräven",
+      "id": "fjallraven-d45246",
+      "locationSet": {"include": ["001"]},
```
Identitaet, line 99: they only seem to have their own stores in the US and Canada everywhere else it's partners

mneko: i ran a build after changing it so it would be happy, and the build changed some other things, sorry

Identitaet: the build looks to be correct and you fixed that small bit fast

Identitaet:
Looks good but we should probably tell @bhousel about adding a new category of brands

bhousel:
> Looks good but we should probably tell @bhousel about adding a new category of brands

`craft=electronics_repair` seems good to me 👍 
As long as it seems like a well-established tag, I'm ok with just about anything people are using.

Issue #4478, opened 2020-10-12
Some more bus networks in Germany to add


kjonosm:
"network": "Regio Verkehrsverbund Lörrach"
"network:short": „RVL“
"network:wikidata": "Q2138145"
"network:wikipedia": "de:Regio Verkehrsverbund Lörrach"
"route": "bus"

"network": "Regensburger Verkehrsverbund"
"network:short": „RVV“
"network:wikidata": "Q2137458"
"network:wikipedia": "de:Regensburger Verkehrsverbund"
"route": "bus"

"network": "Verkehrsgemeinschaft am Bayerischen Untermain"
"network:short": „VAB“
"network:wikidata": "Q2516264"
"network:wikipedia": "de:Verkehrsgemeinschaft am Bayerischen Untermain"
"route": "bus"

"network": "Verkehrsverbund Großraum Nürnberg"
"network:short": „VGN“
"network:wikidata": "Q2516463"
"network:wikipedia": "de:Verkehrsverbund Großraum Nürnberg"
"route": "bus"

"network": "Verkehrsgemeinschaft Rottal-Inn"
"network:short": „VGRI“
"network:wikidata": "Q2516251"
"network:wikipedia": "de:Verkehrsgemeinschaft Rottal-Inn"
"route": "bus"

"network": "Verkehrsgemeinschaft Landkreis Cham"
"network:short": „VLC“
"network:wikidata": "Q1795575"
"network:wikipedia": "de:Verkehrsgemeinschaft Landkreis Cham"
"route": "bus"

"network": "Verkehrsgemeinschaft Landkreis Passau"
"network:short": „VLP“
"network:wikidata": "Q2516229"
"network:wikipedia": "de:Verkehrsgemeinschaft Landkreis Passau"
"route": "bus"

"network": "Verkehrsverbund Mainfranken"
"network:short": „VVM“
"network:wikidata": "Q2516466"
"network:wikipedia": "de:Verkehrsverbund Mainfranken"
"route": "bus"

"network": "Aachener Verkehrsverbund"
"network:short": „AVV“
"network:wikidata": "Q300763"
"network:wikipedia": "de:Aachener Verkehrsverbund"
"route": "bus"

"network": "Verkehrsverbund Rhein-Sieg"
"network:short": „VRS“
"network:wikidata": "Q896041"
"network:wikipedia": "de:Verkehrsverbund Rhein-Sieg"
"route": "bus"

"network": "Westfalentarif"
[...]

PR #4506, opened 2020-10-15
added several German bus networks

lelelew:
Added several German bus networks (& removed duplicates). 

Fixes: https://github.com/osmlab/name-suggestion-index/issues/4478


Identitaet:
Looks good

PR #4510, opened 2020-10-16
Add Hall & Woodhouse brand

Cj-Malone:
H&W is a chain of pubs in the south of England. Each pub has it's [own name](https://www.hall-woodhouse.co.uk/our-pubs/).


Identitaet:
Looks good

PR #4512, opened 2020-10-16
Added brands in New England and did npm run build

weech:
I added Aroma Joe's and Pub 99 and added tags to the NH Liquor & Wine Outlets that weren't picked up automatically. The other stuff automatically happened when I ran `npm run build`, so I'm assuming that's normal?


Identitaet:
Yes other things were added before without running npm run build so that's normal.
The brands also look good
Only thing I'm unsure about is what name pub 99 uses/is used in OSM

weech:
That's why I want it in the NSI, because it's such a mess right now. There are about a dozen "99 Restaurant", a couple "99 Restaurant & Pub", and at least 3 that are simply "Ninety Nine".  I think "Pub 99" was an old branding, or maybe it's a local name, but where I am, in speech, the restaurant is referred to as either "Pub 99" or "the 99 restaurant". "99 Restaurant" is the most common in OSM right now, but the current branding is "Ninety Nine Restaurant & Pub", so I think that'd be best to use as the `name`.

bhousel:
Thanks @weech  & welcome 👋 
Looks good to me 👍

PR #4513, opened 2020-10-16
Add Arriva wikidata and wikipedia

ItaloSa:
I've created a wikidata and wikipedia entity for Arriva at transit/route/bus


bhousel:
Thanks @ItaloSa 👍

PR #4515, opened 2020-10-17
amenity restaurant add 'La Pataterie'

Binnette:
Add french restaurant chain: 'La Pataterie'

Top 7 french chain by number of restaurants in France (given OSM data).

Note, I completed the wikidata element: https://www.wikidata.org/wiki/Q20899937


Identitaet:
Looks good

PR #4516, opened 2020-10-17
amenity fast_food add 'La Croissanterie'

Binnette:
Corresponding wikidata brand: https://www.wikidata.org/wiki/Q3207961


Identitaet:
Looks good

PR #4517, opened 2020-10-17
fix:fix the generic names

kaushikisinha06:


Identitaet:
Why would we remove stores with wikidata ?
Those seem to be valid entries

bhousel:
Sorry @kaushikisinha06 but these are valid items and should remain.

PR #4518, opened 2020-10-17
amenity fast_food add 'Pomme de Pain'

Binnette:
Wikidata page: https://www.wikidata.org/wiki/Q3276265


UKChris-osm:
Thank you @Binnette

PR #4524, opened 2020-10-18
shop supermarket add 'Intermarché Express'

Binnette:
wikidata item : https://www.wikidata.org/wiki/Q98278043

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -2666,6 +2666,18 @@
       }
     },
     {
+      "displayName": "Intermarché Express",
+      "id": "intermarcheexpress-5dc877",
+      "locationSet": {"include": ["fr", "pt"]},
+      "tags": {
+        "brand": "Intermarché Express",
+        "brand:wikidata": "Q98278043",
+        "brand:wikipedia": "fr:Intermarché",
+        "name": "Intermarché Express",
+        "shop": "supermarket"
+      }
+    },
+    {
       "displayName": "Intermarché Super",
       "id": "intermarchesuper-86bfe8",
       "locationSet": {


UKChris-osm:
Hi @Binnette, based on the "Express" name, is it possible this would be better placed in the "_convenience.json_" section??

Binnette:
Hi @UKChris-osm :sun_with_face:,

So i double-checked before submiting my PR and learned some new stuffs : in France for shops between 400 and 2 500 m², we should call them supermarket (supermarché) and not convenience (épicerie).

And I do a little "study" on existing 'Intermarché Express' in OSM - [OverpassQuery](http://overpass-turbo.eu/s/Z91)
|way id|area m2|
|-----|------|
|84136371|978,9|
|84041775|1280,8|
|320579105|1218,6|
|79098930|1033,6|
|131515418|1454,4|
|42463165|3177,0|
|84020874|1116,7|
|63636041|942,8|
|103506582|835,7|

And it seems that every 'Intermarché Express' is OSM are bigger than 400 m².

What do you think? Should I still place my item in convenience ??? Thank you :+1:

UKChris-osm:
That is some good research @Binnette.

I would suggest leave this PR as it is for now, and I will wait to see if anybody else has a view on supermarket / convenience :)

Binnette:
Ok. I'm not in a hurry :wink: Can you put a label "need-help" on the ticket?

Binnette:
Some more investigations based on OSM data:
- 4 'Intermarché Express' are taggued 'convenience'
- 46 'Intermarché Express' are taggued 'supermaket'

Adamant36:
When I first the Wikidata I seem to remember thinking they should be tagged as supermarkets because it appears they sell produce and have shopping carts. Both of which aren't really convenience store things. 

Intermarche Express

Intermarche Express 2

Binnette:
So maybe we can put it in 2 categories ?
- shop=convenience
- shop=supermarket

(I don't know if NSI support it...)

UKChris-osm:
I'm happy for this to be merged as a supermarket based on the above :)

I think the "express" name just made me thing of it being similar to a "Tesco Express".

Binnette:
@UKChris-osm, is it possible to have the same shop in convenience and supermarket in the same time?

Cause my next PR will be on the brand "Sherpa" and they do both. https://www.wikidata.org/wiki/Q3481967

bhousel:
Hey @Binnette , it is possible to have a brand in multiple categories.  Each item will get its own preset in iD or JOSM:

Screen Shot 2020-10-19 at 3 43 29 PM

PR #4525, opened 2020-10-18
add three brazilian universities

rafaelncarvalho:
Hey guys, 

This PR has the three majors brazilian universities

wikipages:

https://www.wikidata.org/wiki/Q2060678
https://www.wikidata.org/wiki/Q835960
https://www.wikidata.org/wiki/Q586904

Thanks

### DIFF
--- a/data/brands/amenity/university.json
+++ b/data/brands/amenity/university.json
@@ -38,6 +38,45 @@
         "name": "University of Phoenix",
         "short_name": "UOPX"
       }
+    },
+    {
+      "displayName": "Universidade de Campinas",
+      "id": "universityofcampinas-d62529",
+      "locationSet": {"include": ["br"]},
+      "tags": {
+        "amenity": "university",
+        "brand": "Universidade de Campinas",
+        "brand:wikidata": "Q2060678",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Hello @rafaelncarvalho and thank you for your first contribution :)

Are these three Universities individual Universities that are based at a single location?

If so, unfortunately they don't really fit in with this particular project, as the aim of the "_Name Suggestion Index_" is to target companies / businesses / education center's, etc which have multiple locations around a Country (or multiple Countries) such as McD's, Target, Ikea, so that they all have the same name / brand data.

rafaelncarvalho:
> Are these three Universities individual Universities that are based at a single location?

No, they have multiples campi

bhousel:
Sorry @rafaelncarvalho , but we don't add individual universities to this index - we only add the kind of for-profit "universities” that operate like a branded business in multiple locations, such as DeVry or University of Phoenix. I know this is an unusual concept outside of the USA - the “universities” in this index are really more like training centers than true universities.

rafaelncarvalho:
No problem, sorry if I miss understood the project

PR #4529, opened 2020-10-19
add location sets to tram networks

kjonosm:


bhousel:
Thanks @kjonosm and welcome 👋 
Would be great to have the `network:wikipedia`/`network:wikidata` tags in a next pass over this..

PR #4532, opened 2020-10-19
Added wikidatas & wikipedias for 拉亞漢堡 & 早安美芝城

yelaD:
Wikidata for 拉亞漢堡: https://www.wikidata.org/wiki/Q67933743


Identitaet:
Looks good and thanks for adding those Chinese brands

PR #4540, opened 2020-10-20
Added financial

AnjaliGGeetha:
Added KFC to office/financial


bhousel:
Thanks @AnjaliGGeetha  - I moved "KFC" to the `short_name` tag in 64c9967c
This allows the matcher code to match this brand for either the long or short names.  (It's not really able to do this if both long and short names are together in the same tag value).

PR #4546, opened 2020-10-20
Update CONTRIBUTING.md with link to installation instructions

tordans:
I was looking for those chapters to lookup the correct node version.


bhousel:
Telling people to read the README is kind of dumb but ok.

Issue #4544, opened 2020-10-20
split German insurance offices Provinzial


kjonosm:
there are two independent insurance companies with that name:

- Provinzial Nordwest, see https://de.wikipedia.org/wiki/Provinzial_NordWest
- Provinzial Rheinland, see https://de.wikipedia.org/wiki/Provinzial_Rheinland

tordans:
Reference to the json: https://github.com/osmlab/name-suggestion-index/blob/eb58fea43b697ed2662a41de090f3aa4b06e5522/data/brands/office/insurance.json#L748-L757

So the new JSON should be:
```
    {
      "displayName": "Provinzial NordWest",
      "id": "provinzial-422e71a", // TODO
      "locationSet": {"include": ["de"]},
      "tags": {
        "brand": "Provinzial NordWest",
        "name": "Provinzial NordWest",
        "brand:wikidata": "Q1778922",
        "brand:wikipedia": "de:Provinzial_NordWest",
        "office": "insurance"
      }
    },
    {
      "displayName": "Provinzial Rheinland",
      "id": "provinzial-422e71a", // TODO
      "locationSet": {"include": ["de"]},
      "tags": {
        "brand": "Provinzial Rheinland",
        "name": "Provinzial Rheinland",
        "brand:wikidata": "Q2114066",
        "brand:wikipedia": "de:Provinzial_Rheinland",
        "office": "insurance"
      }
    },
```

**Location:**
Looking at https://de.wikipedia.org/wiki/Provinzial_Rheinland#/media/Datei:Karte_%C3%96ffentliche_Versicherer_in_Deutschland.png, it should be possible to make the locationSet more specific to only match the green areas for one and the lightblue areas for the other.

Do you agree, @kjonosm ?

kjonosm:
> Do you agree, @kjonosm ?

Yes, that looks great!

bhousel:
reopening, as the latest build put the "Provinzial" item back.. caf0895f

bhousel:
The correct way to fix this is for the 2 Provinzials to have distinct locationSets (specify the regions where they operate with a geojson) and for each entry to also match the generic "Provinzial" somehow, for example in the `short_name` tag or a `matchName` property.

bhousel:
closing, see comments on #4569 
Apparently the companies have completed their merger this year and will all be called "Provinzial" anyway.

PR #4548, opened 2020-10-20
Split Provinzial into two companies

tordans:
Fixes https://github.com/osmlab/name-suggestion-index/issues/4544

Update: ID is auto-generated now.

### DIFF
--- a/data/brands/office/insurance.json
+++ b/data/brands/office/insurance.json
@@ -752,7 +752,7 @@
       "tags": {
         "brand": "Provinzial NordWest",
         "brand:wikidata": "Q1778922",
-        "brand:wikipedia": "de:Provinzial_NordWest",
+        "brand:wikipedia": "de:Provinzial NordWest",
         "name": "Provinzial NordWest",
         "office": "insurance"
       }
@@ -764,7 +764,7 @@
       "tags": {
         "brand": "Provinzial Rheinland",
         "brand:wikidata": "Q2114066",
-        "brand:wikipedia": "de:Provinzial_Rheinland",
+        "brand:wikipedia": "de:Provinzial Rheinland",
         "name": "Provinzial Rheinland",
         "office": "insurance"
       }
--- a/data/brands/office/insurance.json
+++ b/data/brands/office/insurance.json
@@ -747,25 +747,25 @@
     },
     {
       "displayName": "Provinzial NordWest",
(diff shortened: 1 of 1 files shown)


tordans:
Question: When I ran the build-script, the script re-added the previous data again. I dont know where that came from, since I removed it (twice). I did not commit the previous data, just the new one. Should I?

data/brands/office/insurance.json
```diff
@@ -746,12 +746,26 @@
...
-      "id": "provinzial-422e71",
+      "displayName": "Provinzial NordWest",
+      "id": "provinzialnordwest-422e71",
       "locationSet": {"include": ["de"]},
       "tags": {
-        "brand": "Provinzial",
-        "name": "Provinzial",
+        "brand": "Provinzial NordWest",
+        "brand:wikidata": "Q1778922",
+        "brand:wikipedia": "de:Provinzial_NordWest",
```
bhousel, line 755:
```suggestion
        "brand:wikipedia": "de:Provinzial NordWest",
```

data/brands/office/insurance.json
```diff
@@ -746,12 +746,26 @@
...
+      }
+    },
+    {
+      "displayName": "Provinzial Rheinland",
+      "id": "provinzialrheinland-422e71",
+      "locationSet": {"include": ["de"]},
+      "tags": {
+        "brand": "Provinzial Rheinland",
+        "brand:wikidata": "Q2114066",
+        "brand:wikipedia": "de:Provinzial_Rheinland",
```
bhousel, line 767:
```suggestion
        "brand:wikipedia": "de:Provinzial Rheinland",
```

bhousel (changes requested):
Thanks @tordans , looks mostly good!
The wikidata tags need underscores replaced with spaces, then good to merge..

tordans:
> The wikidata tags need underscores replaced with spaces, then good to merge..

@bhousel thank for the change-suggestions, I commited them.
Where would I normally copy this string from? Is it the title of the wikipedia-doc prefixed with the lang?

Identitaet:
You'll find the language code at the begging of the Wikipedia link something like 
`www.en.wikpedia and so on`
And the non encoded version of the name will be in the top left as the title

bhousel (approved):

PR #4549, opened 2020-10-21
amenity cafe remove cafe that are not brands

Binnette:
Hi, this PR remove french cafe that are not brands.

- "Café de la Mairie" : it's like the cafe near the town hall
- "Café de la Poste" : it's like the cafe near the post office
- "Café de Paris" : it is not a brand. Paris is france capital, so many cities have a cafe named so

As you can see, all this cafe names are really common in France and are not a brand.
Feel free to ask any question before merging my PR

### DIFF
--- a/config/filter_brands.json
+++ b/config/filter_brands.json
@@ -210,7 +210,7 @@
     "^bücherschrank$",
     "^büfé$",
     "^cabinet dentaire$",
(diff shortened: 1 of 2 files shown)


Binnette:
Here you go @bhousel :wink: :+1:

bhousel:
Nice, thanks @Binnette 👍

bhousel (approved):

Issue #4547, opened 2020-10-20
How does the autogeneration of the id field work?


tordans:
In https://github.com/osmlab/name-suggestion-index/blob/main/CONTRIBUTING.md#id-generated it says 
> id (generated)
Each item has a unique id generated for it. You can leave it out and the build script will generate one automatically.

How does this work in detail?

1. For my edit, I leave the `id: "",` empty.
2. Then I run `npm run build` => The value in my edited file is not changed
How do I get the id value set?

I am missing something – but what is it :-).

bhousel:
> 2\. Then I run `npm run build` => The value in my edited file is not changed
>     How do I get the id value set?

I don't know - `npm run build` should read all the files into the index and write them back out again cleaned up and with identifiers.  Are there build errors?

TheDynamicPunk:
>     1. For my edit, I leave the `id: "",` empty.

I generally remove the whole id tag when adding a brand and then run the command `npm run build` and it generates the id for that brand

tordans:
Thanks for the help. It worked for https://github.com/osmlab/name-suggestion-index/pull/4548, now.

I tried to clarify the docs via https://github.com/osmlab/name-suggestion-index/pull/4553/files. IMO this makes it easier to understand.

PR #4553, opened 2020-10-21
Extend contributing.md about the id-field.

tordans:
This closes https://github.com/osmlab/name-suggestion-index/issues/4547.


bhousel:
I don't see how this is any simpler, but ok.

PR #4556, opened 2020-10-21
shop bakery: remove non brand 'Le Fournil'

Binnette:
Remove "Le Fournil" cause it is not a brand. "Le Fournil" means "The Bakery" so....


tas50:
Thanks @Binnette

PR #4557, opened 2020-10-21
shop: add tags to 'La Halle'

Binnette:
The shoes brand 'La Halle aux chaussures' have been merged with a clothes brand.
The only brand remaining is "La Halle".


bhousel:
Thanks @Binnette , makes sense to me 👍

PR #4560, opened 2020-10-22
add boost juice

andrewharvey:
Unfortunately the amenity=drinks proposal at https://wiki.openstreetmap.org/wiki/Proposed_features/Takeaway_drink_shops
was not accepted by the community and so these need to be tagged as cafe.

Currently they are mostly mapped as either amenity=cafe + amenity=fast_food, because the proposal wasn't approved it's even more important that we tag the brand:wikidata so that data consumers can fix the cafe/fast_food tagging to show them as mostly takeaway drink shops.

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -262,6 +262,19 @@
       }
     },
     {
+      "displayName": "Boost Juice",
+      "locationSet": {"include": ["au"]},
+      "tags": {
+        "amenity": "cafe",
+        "brand": "Boost Juice",
+        "brand:wikidata": "Q4943789",
+        "brand:wikipedia": "en:Boost Juice",
+        "cuisine": "juice",
+        "name": "Boost Juice",
+        "takeaway": "yes"
+      }
+    },
+    {
       "displayName": "Boston Tea Party",
       "id": "bostonteaparty-a89223",
       "locationSet": {"include": ["gb"]},


Binnette (approved):
Seems good 👍 The "id" is missing but will be generated next build time. And we will need to add this "id" in wikidata item.

I edited the wikidata item: https://www.wikidata.org/wiki/Q4943789
I added facebook, twiiter and instagram, cause NSI use logo from those web sites.

bhousel:
> Unfortunately the amenity=drinks proposal at https://wiki.openstreetmap.org/wiki/Proposed_features/Takeaway_drink_shops
> was not accepted by the community and so these need to be tagged as cafe.
> 
> Currently they are mostly mapped as either amenity=cafe + amenity=fast_food, because the proposal wasn't approved it's even more important that we tag the brand:wikidata so that data consumers can fix the cafe/fast_food tagging to show them as mostly takeaway drink shops.

Yeah I know there are several juice and bubble tea places using the `amenity=cafe` tag.. This is ok for now.
One of the cool things about the `brand:wikidata` tag is that it makes it trivially easy to go through and retag all of these businesses if OSM ever does decide to change their tag to something else.

Thanks for the PR @andrewharvey 👍

andrewharvey:
Thanks for the review.

> The "id" is missing but will be generated next build time. 

This part I was confused by the CONTRIBUTING guide, because it mentioned it was automatically added and generated. If I needed to manually create one then it didn't explain how to create the id and what it should be, so I left it blank.

> And we will need to add this "id" in wikidata item.

Is there some documentation which explains why that's needed, where it's used etc?

PR #4561, opened 2020-10-22
Fix vending machines (mostly german ones)

fxedel:
I just fixed some warnings and added more data to the changed items.


bhousel:
Thanks @fxedel , these all look great 👍

PR #4563, opened 2020-10-22
amenity shop clothes: set correct wikidata item

Binnette:
Link to the correct wikidata item :wink:

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -2749,7 +2749,7 @@
       "locationSet": {"include": ["ch", "fr"]},
       "tags": {
         "brand": "La Halle",
-        "brand:wikidata": "Q62390731",
+        "brand:wikidata": "Q100728296",
         "name": "La Halle",
         "shop": "clothes"
       }


Adamant36:
It would be cool if there was an entry for the shoe store that the other Wikidata entry is about also. Since I assume they are different and otherwise it won't be in the index anymore.

bhousel:
I'd like to understand this brand better to be sure that we are doing the right thing:
* [La Halle - Q100728296](https://www.wikidata.org/wiki/Q100728296)
* [La Halle aux Chaussures - Q62390731](https://www.wikidata.org/wiki/Q62390731)
 
These don't look all that different to me, and they were both created by people on the NSI project.  I expect that wikidata admins would see this as redundant and merge them.

If the difference really is "A normal La Halle that sells all clothes" and "A special format La Halle that sells only shoes", and they use the same logo, website, social media, and management - we should probably just use one QID for the La Halle stores.

We have a few other entries in the index like this.  For example "Gap", "Gap Factory", "Gap Kids", "Baby Gap" all use Q420822.  This is ok because I don't think anybody would be confused to see the Gap logo on any of those store formats.
Gap Inc also owns many [other brands](https://en.wikipedia.org/wiki/Gap_Inc.#Brands) such as  "Janie & Jack", "Banana Republic", "Old Navy", etc, and these are unique enough that they have their own brand identities, social media, etc, so they have their own QIDs.

Binnette:
Hi guys, long story short. the 2 old brands "La Halle aux vêtements" and "La Halle aux chaussures" do not exist anymore. They have been merged in a third **new** brand "La Halle" in january 2019.

I made the choice to not delete the shoes shop "lahalleauxchaussures-c57706" yet.
As far as I understand, all the shoes shop will be "converted" IRL to clothes shoes "lahalle-5bd015". (Some shop are already converted to clothes shop IRL. But I need to study this more)

So for now, i don't want to delete "lahalleauxchaussures-c57706" but i will delete it in few month, when i'm sure that no shoes shop remain IRL.

Hope I clarify everything :wink:

bhousel:
> So for now, i don't want to delete "lahalleauxchaussures-c57706" but i will delete it in few month, when i'm sure that no shoes shop remain IRL.
> 
> Hope I clarified everything

Thanks, this seems ok!  It is no problem for us to keep 2 items in NSI.  My concern is more about what the Wikidata admins will think.  If it is all going to be a single brand eventually, this is fine.

PR #4565, opened 2020-10-23
Fix location of Norwegian Extra, add Brazilian Extra supermarket and convenience

willemarcel:


Adamant36:
It looks like you just created another entry for the Extra. What's the difference between the two?

willemarcel:
@Adamant36 The Norwegian one is totally different from the Brazilian one (the wikidata/pedia tags are different). The Brazilian company has both convenience stores (Mini Extra) and supermarkets.
If you prefer, I can create a wikidata entry for Mini Extra.

bhousel:
> The Norwegian one is totally different from the Brazilian one (the wikidata/pedia tags are different).

Cool, thanks for clarifying @willemarcel 👍 

> The Brazilian company has both convenience stores (Mini Extra) and supermarkets.
> If you prefer, I can create a wikidata entry for Mini Extra.

No need, they seem like the same.

PR #4566, opened 2020-10-23
Fix location of Minipreço and Globus supermarkets

willemarcel:
They were tagged as global and have very common names


bhousel:
Thanks! 👍 

I'm starting to feel like we should be very suspicious about any brand tagged as global.  Maybe we should have a challenge to tighten up the scope of the `include: '001'` items.

PR #4567, opened 2020-10-23
Improve Ipercoop and add Coop brand

willemarcel:


Adamant36:
I don't really get what's going on here. They seem to be the same exact brand with the same Wikipedia article. What's the difference?

willemarcel:
@Adamant36 The same company has supermarkets (Coop) and hypermarkets (Ipercoop). There are different wikidata entries, but only one wikipedia article. That part of the WP article, mentions all the brands of the company: https://it.wikipedia.org/wiki/Coop_Italia#Rete_di_vendita

Adamant36:
OK. Its better to just get rid of the Wikipedia entry for the one that's only mentioned in a section. We don't need it anyway. Honestly, in this case id get rid of it for both entries since about the holding company and only barely mentions either one. Just having the Wikidata entries is fine. Its better then links to articles that aren't actually about the brands.

willemarcel:
@Adamant36 code updated

data/brands/shop/supermarket.json
```diff
@@ -1136,6 +1136,18 @@
...
       }
     },
+    {
+      "displayName": "Coop (Italy)",
+      "id": "coop-0cb133",
+      "locationSet": {"include": ["it"]},
+      "tags": {
+        "brand": "Coop",
+        "brand:wikidata": "Q1129774",
+        "brand:wikipedia": "it:Coop_Italia",
```
bhousel, line 1146:
```suggestion
        "brand:wikipedia": "it:Coop Italia",
```

bhousel (changes requested):
Just a small change - need underscores replaced with spaces in the brand:wikipedia tag.

(I'm so tired of telling people this that I'm probably just going to add this to the list of things the build script fixes automatically for us)

bhousel:
> (I'm so tired of telling people this that I'm probably just going to add this to the list of things the build script fixes automatically for us)

I did this in dafb0423

bhousel (approved):

PR #4568, opened 2020-10-23
Add sweet jesus ice cream

alefuedev:
https://www.sweetjesusicecream.com/


Adamant36:
Hello. If there's no Wikipedia or Wikidata tags then they are usually just left out. That said, maybe someone can make a Wikidata entry for it at least.

PR #4569, opened 2020-10-23
Add matchNames to provinzial to prevent error

tordans:
Re https://github.com/osmlab/name-suggestion-index/issues/4544#issuecomment-714524865, adding the matchNames prevents the script to re-add the previous "Provinzial" data.

@bhousel I just added the `matchNames` and the previous "Provinzial" data did not come back (which it did before https://github.com/osmlab/name-suggestion-index/pull/4548#issuecomment-713498993).

To this might be enough and no need to add separate `locationSet`, right?

Update to https://github.com/osmlab/name-suggestion-index/pull/4548.
Fixes https://github.com/osmlab/name-suggestion-index/issues/4544.

### DIFF
--- a/data/brands/office/insurance.json
+++ b/data/brands/office/insurance.json
@@ -750,19 +750,10 @@
       }
     },
     {
-      "displayName": "Provinzial",
-      "id": "provinzial-422e71",
-      "locationSet": {"include": ["de"]},
-      "tags": {
-        "brand": "Provinzial",
-        "name": "Provinzial",
-        "office": "insurance"
-      }
-    },
-    {
       "displayName": "Provinzial NordWest",
       "id": "provinzialnordwest-422e71",
       "locationSet": {"include": ["de"]},
+      "matchNames": ["provinzial"],
       "tags": {
         "brand": "Provinzial NordWest",
         "brand:wikidata": "Q1778922",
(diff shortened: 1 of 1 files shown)


bhousel:
> To this might be enough and no need to add separate `locationSet`, right?

No, you will need separate locationSets for these.  Otherwise the matcher (and iD validator) will not know how to upgrade a brand with `name=Provinzial` in Germany.  It will match 2 things.

bhousel:
I researched these brands a bit today, and want to know more:

It looks like there are 2 clear clusters in the data in OSM.. Is it fair to say that these are [Provinzial NordWest](https://www.wikidata.org/wiki/Q1778922) and [Provinzial Rheinland](https://www.wikidata.org/wiki/Q2114066)
Screen Shot 2020-11-02 at 5 12 45 PM

Currently they are all just named "Provinzial" and it looks like they might be in the process of merging?
http://www.globallegalchronicle.com/provinzial-nordwests-merger-with-provinzial-rheinland/

Are we sure we definitely want these split up (per #4544 / #4548), or should will they all just end up as the same "Provinzial" anyway?

kjonosm:
Well, sorry for the confusion then. I'm happy enough switching back to one single entry.

bhousel:
> Well, sorry for the confusion then. the reason I brought it up was that I couldn't find a corresponding wikidata item. I'm happy enough switching back to one single entry.

No apology needed, the situation actually is pretty confusing (but not all that unusual for banks and insurance companies to merge).  The Provinzial NordWest website now redirects to a more generic `https://www.provinzial-konzern.de/` page (translation "Provinzial Group") which also has a link to Provinzial Rheinland, which is just branded as "Provinzial" and their angel wing logo.

Sounds like it is probably ok to switch back to a single "Provinzial" entry.
- "Provinzial" with matchNames for "Provinzial Rheinland" and "Provinzial NordWest", scoped to include all of 'de'

PR #4572, opened 2020-10-23
Update Dirk supermarket

willemarcel:
Wikipedia says that "Dirk van den Broek" is now only "Dirk": https://en.wikipedia.org/wiki/Dirk_(supermarket)
Location is only Netherlands


bhousel:
Thanks @willemarcel 👍

Issue #4443, opened 2020-10-10
add Reál


debyos:
Hi, this is will be another easy entry. Reál is a Hungarian retail chain:
- https://real.hu/
- https://hu.wikipedia.org/wiki/Re%C3%A1l_Hung%C3%A1ria_%C3%89lelmiszer_Kereskedelmi_Kft.
- http://overpass-turbo.eu/s/YUm (~270 shops in Hungary on the map)

shubh710:
Hi @debyos 
I would like to work on this, I might require some guidance though.
Thanks.

Adamant36:
> I would like to work on this, I might require some guidance though.
Thanks.

There's a guide on how to do things in the contributing file 
https://github.com/osmlab/name-suggestion-index/blob/main/CONTRIBUTING.md

If you have any specific questions feel free to ask.

shubh710:
Hi @Adamant36,
Sorry but I will not be able to work until weekend.
Please assign this to someone else if someone can work on it.
Thanks

Adamant36:
No worries. We don't really assign things to people anyway. So, feel free to work on it or anything else when you have the time. Otherwise, someone else will do it eventually.

debyos:
Hi,
thanks for doing it. I can't code, nevertheless, I can help you with information from the field.
Cheers

amous4822:
It's a Hungarian retail company. Which file should I add it in ?

amous4822:
please review and revert back. 
Thanks

amous4822:
please review and revert back.
Thanks.

PR #4573, opened 2020-10-23
added Reál

amous4822:
Added Real to department_store file. 
#4443


bhousel (changes requested):
Few things:
- From [their website](https://real.hu/), it looks more like a supermarket or convenience store (or both?)
- This long name is definitely not how they refer to themselves, nor how people would expect the points to appear on a map.  Why not just "Reál"?

amous4822:
should I put it only under supermarket ? one of the banners in [the website](real.hu) carousel mentions it's a super store.

bhousel:
> should I put it only under supermarket ? one of the banners in [the website](real.hu) carousel mentions it's a super store.

I don't know this brand, so I'm not able to really decide that.
- supermarket is more for larger stores where you push a cart around.
- convenience is more for smaller stores where you carry a basket around.

If "Reál" operates both kinds of stores, it can be in both files.

amous4822:
Should I mention Real only as a display name and use long names for brand and name in tags ?

debyos:
Hi,
sorry for being so late. Reál works mostly with `convenience` stores, mostly in rural areas. Nevertheless, they do have also `supermarket` like shops so you can put them to both categories. Moreover, they work in a franchise system, so the name of the company varies for each store, i.e. the company name is of less interest.
Cheers

amous4822:
created new PR #4588

PR #4574, opened 2020-10-23
Remove duplicate names for 'Mazda' brand

Ferveloper:
Mazda appeared as a duplicate name warning. I fixed it as follows:

- Added `"matchTags": ["shop/car_repair"]` to the (preferred) `"shop/car|Mazda"` entry.
- Deleted the (not preferred) entry for `"shop/car_repair|Mazda"`.
- Executed `npm run build` and checked that Mazda warning disappeared.

Hope this helps. Happy Hacktoberfest!

### DIFF
--- a/data/brands/amenity/clinic.json
+++ b/data/brands/amenity/clinic.json
@@ -134,7 +134,7 @@
     },
     {
       "displayName": "Posyandu",
-      "id": "posyandu-ed675b",
+      "id": "posyandu-87a279",
       "locationSet": {"include": ["id"]},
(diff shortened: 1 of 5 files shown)


bhousel:
Sorry @Ferveloper - the `shop/car` - `shop/car_repair` pairing is actually a situation where we want both items included.  It's pretty common for a car dealership to also have a repair center.  The correct thing to do is add `brand` tags to the `shop/car_repair|Mazda` item that is missing them.  Can you open another PR for that?

PR #4575, opened 2020-10-24
Update bus.json

evansiroky:
Added two more San Francisco Bay Area agencies. I was going to update VTA as well, but it looks like the overpass-turbo query shows that it is used by the Santa Clara VTA, but also some other VTA in Massachusetts. I don't know what to do about that situation, can you provide guidance?

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -1632,7 +1632,11 @@
       "locationSet": {"include": ["us"]},
       "tags": {
         "network": "Caltrain",
+        "network:wikidata": "Q166817",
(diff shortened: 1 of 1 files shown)


1ec5:
> it looks like the overpass-turbo query shows that it is used by the Santa Clara VTA, but also some other VTA in Massachusetts.

It’s expected that `network` isn’t unique globally, but `network:wikidata` would be unique.

bhousel:
Yes, this looks great, so I'll accept it as-is 👍 

You can make multiple entries for the 2 VTAs.. They can each have a unique locationSet and `network:wikidata`.

PR #4576, opened 2020-10-24
Add Mazda wiki brands in shop/car_repair list

Ferveloper:
As requested, `brand:wikipedia` and `brand:wikidata` were added to Mazda entry in `shop/car_repair`.
After running `npm run build` everything seems fine for Mazda.


bhousel:
Thanks @Ferveloper , looks good 👍

PR #4581, opened 2020-10-24
Fix displayNames of Extra supermarkets and convenience

willemarcel:
I noticed I added wrong display names on https://github.com/osmlab/name-suggestion-index/pull/4565, so fixing it now with some meaningful display names.


bhousel:
Thanks @willemarcel  - this is better 👍

PR #4582, opened 2020-10-25
add Costco Food Court

modaserMoj:


data/brands/amenity/fast_food.json
```diff
@@ -908,6 +908,19 @@
         "takeaway": "yes"
       }
     },
+    {
+      "displayName": "Costco Food Court",
+      "id": "costcofoodcourt-4d2ff4",
+      "locationSet": {"include": ["us"]},
```
tas50, line 914: Do the worldwide costcos have the food court?

modaserMoj: I can add a few of the countries that I found but I can find some sort of a complete list. Should I add those countries?

tas50: Add whatever you can find and folks can build on it from there.

modaserMoj: I added what I could find

tas50:
Thanks for the update

PR #4585, opened 2020-10-25
deletion of elements that aren't brands 

Olyon:


tas50:
@Olyon these will just come back when the file is generated again unless you add to the filters to mark them as not being brands

Olyon:
I made the additions in filter_brands.json so it doesn't come back if I understood correctly.
but I am not added "Accor" which only appears 3 times as a name in the world. But appears a lot in brand=* but it is not a brand, it's the owner of the brand...

tas50:
Thanks @Olyon

tas50:
@Olyon I can "npm run build" which regenerated the names we keep and Accor has come back, but the others are good.

PR #4586, opened 2020-10-25
Add Suzuki wiki brands in shop/car_repair list

Ferveloper:
Given that Suzuki is a brand both for car selling and car repairing, `brand:wikipedia` and `brand:wikidata` were added to Suzuki entry in `shop/car_repair` to fix the duplicates warning.
After running `npm run build` the console output seems ok for Suzuki.


bhousel:
Thanks @Ferveloper , looks perfect 👍

PR #4595, opened 2020-10-26
add Premio

kjonosm:
Add car repair franchise "Premio".
They also sell tyres but brand themselves as full service car repair shops.
According to overpass, capital letter is preferred.


bhousel:
Looks good @kjonosm , thanks!  👍

Issue #4404, opened 2020-10-09
National express


Identitaet:
National Express = network
https://en.wikipedia.org/wiki/National_Express

Operator for Germany
https://de.wikipedia.org/wiki/National_Express_Holding
I guess internationally https://en.wikipedia.org/wiki/National_Express is the best (unsure tough)

Adamant36:
I almost feel like creating a label just for adding transport entries. It seems odd to use the "add brand" label with them for some reason. I feel like it should be more specific. "add transport brand" maybe? "add brand - transport"? I don't know.

Identitaet:
I'd suggest we go with transit to be consistent but yeah sounds good

Adamant36:
> I'd suggest we go with transit to be consistent but yeah sounds good

Yeah, I hadn't noticed we were going with "transit" until after I wrote the message. Consistent is good though.

amous4822:
National express offers services in Bus, Coach, Train, Tram. Currently it's only added in bus category. Should I add it in other categories too ?

bhousel:
> National express offers services in Bus, Coach, Train, Tram. Currently it's only added in bus category. Should I add it in other categories too ?

Yes!  The script only collects networks that are used on at least 10 routes, so we're surely missing a bunch.

KasperFranz:
It might also be worth changing the wikidata from https://www.wikidata.org/wiki/Q60188 (Mobico Group) to Q13479582  (National Express Coaches)

They renamed their group to Mobico group
source: https://www.expressandstar.com/news/business/2023/05/10/change-of-name-for-national-express/

PR #4597, opened 2020-10-26
updated national express

amous4822:
added missing tags and location #4404

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -6728,23 +6728,8 @@
     {
       "displayName": "National Express",
       "id": "nationalexpress-6cf6aa",
-      "locationSet": {
-        "include": [
-          "be",
-          "bh",
-          "ca",
-          "de",
-          "es",
-          "fr",
-          "gb",
-          "ie",
-          "ma",
-          "mt",
-          "nl",
-          "pt",
-          "us"
-        ]
-      },
+      "locationSet": { 
+        "include": ["gb-eng", "gb-sct", "gb-wls"] },
       "matchNames": [
         "national express group",
         "national express group plc",
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -6727,11 +6727,36 @@
     },
     {
       "displayName": "National Express",
-      "id": "nationalexpress-8e0970",
-      "locationSet": {"include": ["gb"]},
+      "id": "nationalexpress-6cf6aa",
+      "locationSet": {
+        "include": [
+          "be",
+          "bh",
+          "ca",
+          "de",
+          "es",
+          "fr",
+          "gb",
+          "ie",
+          "ma",
+          "mt",
+          "nl",
+          "pt",
+          "us"
+        ]
+      },
+      "matchNames": [
+        "national express group",
+        "national express group plc",
+        "national express holding"
+      ],
       "tags": {
         "network": "National Express",
+        "network:wikidata": "Q1233212",
+        "network:wikipedia": "en:National Express",
         "operator": "National Express",
+        "operator:wikidata": "Q1233212",
+        "operator:wikipedia": "en:National Express",
         "route": "bus"
       }
     },
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -6752,10 +6752,10 @@
       ],
       "tags": {
         "network": "National Express",
-        "network:wikidata": "Q1233212",
+        "network:wikidata": "Q60188",
         "network:wikipedia": "en:National Express",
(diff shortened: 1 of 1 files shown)


amous4822:
please review and revert.
Thanks.

bhousel (changes requested):
I see from their wikipedia page https://en.wikipedia.org/wiki/National_Express that they might be the "operator" of a bunch of transit stuff around the world, but I think that the "national express" coaches are only a GB thing.

Also it looks like the wikidata that you added for them is wrong - Q1233212 links to A-welle.

So, this PR needs some work, and I'd like to get someone from GB to confirm how `network=National Express` should work.  (We can check overpass too)

peternewman:
They certainly do coaches in the UK. It looks like they deserve their own network tag as I don't think they are just the operator for government routes:
https://en.wikipedia.org/wiki/National_Express_Coaches

amous4822:
I tried to book tickets for their coaches around the world and it offered for places outside UK too. Should Location set contain all locations with available service or just the headquarters ? they have headquarters only in UK.

bhousel:
Doing research like this is the main reason that we built the nsi.guide site 😄 

Here's what we know about National Express: https://nsi.guide/index.html?t=transit&k=route&v=bus#nationalexpress-8e0970
Clicking the Overpass query shows the places where "National Express" is used on a bus route:  https://overpass-turbo.eu/s/Zrc

Since we only see it used in Great Britain, I'd make the locationSet like this:
`{ "include": ["gb-eng", "gb-sct", "gb-wls"] }`

amous4822:
This is a great tool !!
but I'm a bit confused, the website mentions stops at paris, amsterdam and other locations not shown in overpass-turbo. We should ignore this ?

bhousel:
> but I'm a bit confused, the website mentions stops at paris, amsterdam and other locations not shown in overpass-turbo. We should ignore these locations ?

Yes sorry, this one is confusing - my guess is that this is an unusual bus operator.  

They operate "National Express" busses on the "National Express" network in Great Britain.  They also operate other busses on other networks in other places.  

For this `bus.json` file, we're trying to pin down the extent of the "National Express" bus network.  

Put another way, if you were to make a map of where "National Express" busses run, what would it look like?  The best map we can make with data gathered from OpenStreetMap looks like this (from Overpass):  
Screen Shot 2020-10-27 at 11 54 50 AM

It's probably incomplete, and our goal is to make the data in OpenStreetMap more complete and correct.

amous4822:
updated the location set.

bhousel (approved):

bhousel:
Thanks @amous4822 for the quick fixes 👍

peternewman:
> Since we only see it used in Great Britain, I'd make the locationSet like this:
> `{ "include": ["gb-eng", "gb-sct", "gb-wls"] }`

Technically none of the currently mapped places are in Scotland.

Although they do mention Scotland on their website:
https://www.nationalexpress.com/en/destinations/edinburgh

Pre-covid they also ran routes to Northern Ireland:
https://www.nationalexpress.com/en/destinations/belfast/london-to-belfast

And even Ireland itself:
https://www.nationalexpress.com/en/destinations/dublin/london-to-dublin

bhousel:
Hm yeah the situation is definitely complicated.
Looking at their own route map (https://routemap.nationalexpress.com/), I'd be comfortable expanding it to include Scotland, Paris, and Rome: 
 
Screen Shot 2020-10-27 at 1 37 24 PM
https://routemap.nationalexpress.com/

peternewman:
Given they say temporarily suspended:
https://www.nationalexpress.com/en/destinations/dublin/london-to-dublin

Shouldn't we add the other places?

Or to ask the broader question, is NSI mapping for during Covid or before (and assuming things will come back), same thing with say restaurants which have become or added takeaway currently.

We could tag `covid:opening_hours=closed` for Ireland at least, but presumably other places in Europe it may still be travelling through even if it's not stopping.

bhousel:
> Or to ask the broader question, is NSI mapping for during Covid or before (and assuming things will come back), same thing with say restaurants which have become or added takeaway currently.
>
> We could tag `covid:opening_hours=closed` for Ireland at least, but presumably other places in Europe it may still be travelling through even if it's not stopping.

A good way to think about the NSI is that we don't want to add anything to it that isn't true everywhere.  The tags in NSI will be suggested to users in other software like iD, or might be the basis of Maproulette tasks or automated cleanup scripts.

So I think because the covid situation is constantly changing, we should avoid storing tags like this in NSI (though mappers can definitely add these tags to the individual routes or businesses in OSM).

PR #4605, opened 2020-10-27
Update tram.json

kjonosm:
WVV is no network. They operate tram lines within the "Verkehrsverbund Mainfranken" network.

### DIFF
--- a/data/transit/route/tram.json
+++ b/data/transit/route/tram.json
@@ -728,10 +728,10 @@
       "id": "wurzburgerversorgungsundverkehrsgmbh-37759c",
       "locationSet": {"include": ["de"]},
       "tags": {
-        "network": "Würzburger Versorgungs- und Verkehrs-GmbH",
-        "network:short": "WVV",
-        "network:wikidata": "Q1291006",
-        "network:wikipedia": "de:Würzburger Versorgungs- und Verkehrs-GmbH",
+        "operator": "Würzburger Versorgungs- und Verkehrs-GmbH",
+        "operator:short": "WVV",
+        "operator:wikidata": "Q1291006",
+        "operator:wikipedia": "de:Würzburger Versorgungs- und Verkehrs-GmbH",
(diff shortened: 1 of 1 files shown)


bhousel:
ok, I see it here:
https://nsi.guide/index.html?t=transit&k=route&v=tram#wurzburgerversorgungsundverkehrsgmbh-37759c

the thing to do in this case is to remove it from the index if it's not a `network`.

kjonosm:
@bhousel I could also add the network WVV operates in. But I am not sure as how to choose the correct display name in such an instance where network and operator are different from each other. In Germany busses, trains, trams etc. are mostly branded with the operators name and logo. Should I add operator tags only when they correspond with the network? What's your recommendation?

bhousel:
Thanks for the context @kjonosm  - looks like 'WVV' made it into the index because we collected it as the `network=WVV` on several `route=bus` features: 
https://github.com/osmlab/name-suggestion-index/blob/10cb83e504063a04e6de1d1fd78ff09072668bd6/dist/collected/networks_all.json#L1206
See overpass results for it:
https://overpass-turbo.eu/s/ZtG

A few possibilities:
- 'WVV' is actually a network run and operated by that company in that location (reminds me of National Express #4597)
- 'WVV' is actually a network in that location that means something other than what we think it means?  
- 'WVV' is mistagged and the network tag should be set to something else?  (reminds me of SEPTA high speed rail #4496)

Can you check the overpass results and maybe reach out to whomever added this data?

Generally the goal is to collect `network`, check them for consistency, and link a `network:wikidata` to each one.  
If every route in that `network` has the same `operator` then we can set those tags also, but I'm considering `operator` as optional and we should just leave it off when a `network` has multiple operators.

kjonosm:
It's certainly mistagged. I cleaned up the relevant OSM data and fixed the network tags. I will close this pull request and delete the entry in another one. Thanks for your further explanations.

PR #4609, opened 2020-10-28
Update train.json

kjonosm:
Update train networks in Germany

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -393,12 +393,16 @@
       }
     },
     {
-      "displayName": "Dublin Commuter",
+      "displayName": "Commuter",
       "id": "dublincommuter-73b98a",
       "locationSet": {"include": ["ie"]},
       "tags": {
-        "network": "Dublin Commuter",
-        "operator": "Dublin Commuter",
+        "network": "Commuter",
+        "network:wikidata": "Q5155093",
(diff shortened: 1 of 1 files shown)


data/transit/route/train.json
```diff
@@ -393,12 +393,16 @@
...
     },
     {
-      "displayName": "Dublin Commuter",
+      "displayName": "Commuter",
       "id": "dublincommuter-73b98a",
       "locationSet": {"include": ["ie"]},
       "tags": {
-        "network": "Dublin Commuter",
-        "operator": "Dublin Commuter",
+        "network": "Commuter",
```
bhousel, line 400: If the `network=Dublin Commuter` is already in widespread use, we probably shouldn't encourage retagging it to `network=Commuter`.  But other than that, the rest of the PR looks good 👍

bhousel (changes requested):
just a question about the Dublin Commuter

data/transit/route/train.json
kjonosm: OK. Unfortunately I don't know how to edit the pr to fix the issue...

bhousel: > OK. Unfortunately I don't know how to edit the pr to fix the issue...

thats ok, I'll merge and put it back

bhousel (approved):

PR #4612, opened 2020-10-28
Resolve cosmo duplicates

Ferveloper:
I found a warning for Cosmo, which is a brand for a British restaurant and a Japanese oil company.
https://en.wikipedia.org/wiki/Cosmo_Oil_Company
https://www.cosmo-oil.co.jp/eng/index.html

British restaurant already has its wiki tags, but when I added wiki tags and locationSet for Cosmo Oil company, after running `npm run build`, an error crashed the process.

Then I found that there was another entry for Cosmo oil company, but with its Japanese name (コスモ). Therefore, I deleted the incomplete entry for Cosmo in `amenity/fuel.json` list to fix the duplication warning.


bhousel:
From overpass it looks like there is also a cluster of "Cosmo" fuel stations in Thailand:
https://overpass-turbo.eu/s/Zu3

Screen Shot 2020-10-28 at 4 24 52 PM

I did some Google searching and it looks like these Thailand Cosmo fuel stations might have been rebranded or bought by [PT](https://nsi.guide/index.html?t=brands&k=amenity&v=fuel#pt-56d9b3).

So this is probably ok to remove the "Cosmo" item.

PR #4615, opened 2020-10-29
Add two VTA networks

evansiroky:
This PR adds the two VTA networks. I'm not sure if I did it right, so I'm open to feedback if something needs to be corrected.


bhousel:
You'll need to add custom geojsons for these features to disambiguate between the Martha's Vineyard VTA and the Santa Clara VTA.  (If they are both set up with `"include": "us"`, they will get the same id generated and it won't work).

evansiroky:
@bhousel can you please point me to some documentation or an existing example that shows me how to do this?

evansiroky:
Ok, I found the relevant docs [here](https://github.com/osmlab/name-suggestion-index/blob/main/CONTRIBUTING.md#locationset). I'll work on fixing this.

evansiroky:
I added some geojson to differentiate between the two VTA agencies. Hope this works.

bhousel:
Perfect!  Thanks for adding these @evansiroky 👍

PR #4620, opened 2020-10-29
Add Virginia Railway Express (VRE)

michaelblyons:
Also fix VIA Rail typo


bhousel:
Looks great thanks @michaelblyons 👍 

At first I was sure Keolis was a typo, but nope, they actually are the operator 😆

michaelblyons:
> At first I was sure Keolis was a typo, but nope, they actually are the operator 😆

I was likewise surprised. 😉

Issue #4231, opened 2020-10-01
Hacktoberfest challenge - name-suggestion-index has transit now!


bhousel:
Happy Hacktober! 👻 

The past few days I've been updating the core code to add another "tree" to the name-suggestion-index project for **transit**.  I've collected frequently-occuring `network` tag values on for several route features, and we're going to canonicalize all the transit networks in OpenStreetMap to be linked to a `network:wikidata` tag.

🚄  Check out the transit networks of OpenStreetMap here:  https://nsi.guide/index.html?t=transit

This is going to be **a lot** of work, so I'll be opening **a lot** of research tickets looking for volunteers to go through the lists and figure out the wikidata identifier for each transit network, and update the locationSet to say which country they operate in.  You can even now draw a GeoJSON around the bounds of where the transit network operates, if you want to be even more helpful!.  Also, make sure those wikidata pages are updated with the transit company's social media information, so we can fetch those sweet logos. 

If you haven't already, sign up for Hacktoberfest here - https://hacktoberfest.digitalocean.com/
👕 Make 4 pull requests during the month of October and you will receive a free t-shirt (or you have the option to plant a tree if you already have enough t-shirts  )

_Important: If you don't want to receive a LOT of email about the name-suggestion-index project - you should unsubscribe now for a few days at least!_

3Hearts:
Looking to cut my teeth on some work this month and happy to assist with this.

camelCaseNick:
>  You can even now draw a GeoJSON around the bounds of where the transit network operates [...]

Is it possible to extract it from OSM's boundary=public_transport? [Taginfo says, we have 263.](https://taginfo.openstreetmap.org/tags/boundary=public_transport)

bhousel:
> Is it possible to extract it from OSM's boundary=public_transport? Taginfo says, we have 263.

Interesting! I’ll look into it.. boundaries in OSM are often waaaay more complex than what we want for our needs (a fast point-in-polygon test to answer questions like “do MTA bus stops belong here?”)

In every situation I’ve needed a polygon, it’s a lot faster for me to just go to http://geojson.io and draw it myself - again keeping the point count as low as possible.

peternewman:
Cool, a few questions/thoughts.

While I'm not just here for a free T-shirt (as hopefully my history shows: https://github.com/osmlab/name-suggestion-index/pulls?q=is%3Apr+author%3Apeternewman ), currently no-one will get them:
[...]

PR #4625, opened 2020-10-30
adding wikipedia to EMTU bus services

Tubaleviao:
Adding wikipedia and wikidata values for issue #4231

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -2904,10 +2904,13 @@
     },
     {
       "displayName": "EMTU",
-      "id": "emtu-2ef076",
+      "id": "emtuempresametropolitanadetransportesurbanos-2ef076",
       "locationSet": {"include": ["br"]},
       "tags": {
-        "network": "EMTU",
+        "network": "EMTU - Empresa Metropolitana de Transportes Urbanos",
+        "network:short": "EMTU",
+        "network:wikidata": "Q5374439",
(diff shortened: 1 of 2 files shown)


data/transit/route/bus.json
```diff
@@ -2904,10 +2904,13 @@
     },
     {
       "displayName": "EMTU",
-      "id": "emtu-2ef076",
+      "id": "emtuempresametropolitanadetransportesurbanos-2ef076",
       "locationSet": {"include": ["br"]},
       "tags": {
-        "network": "EMTU",
+        "network": "EMTU - Empresa Metropolitana de Transportes Urbanos",
```
bhousel, line 2910: The effect of this is that, once we deploy this, iD validator will suggest a tag upgrade for every network=EMTU bus route in Brazil to use the long value that you have changed it to. Is this what you intended?

In my opinion there is no harm in continuing to use network=EMTU tag, but there is clear benefit in adding the network:wikidata and network:wikipedia tags.

bhousel (changes requested):
Can you undo the changes to package.json?  Also had a question about changing the `network` tag.  (This is ok, but just want to confirm that local mappers want it changed - it might affect transport maps that rely on it).

data/transit/route/bus.json
bhousel: This was getting a bit stale, so I just added the tags in 1fe62781.. Thanks @Tubaleviao 👍

PR #4626, opened 2020-10-30
Update tram.json

kjonosm:
- update location sets
- add wikidata & wikipedia
- update dissolved CTC India with WBTC
- delete operator-only entries (WVV, RVG, IVB)
- merge IDS JMK & cz:IDS JMK

### DIFF
--- a/data/transit/route/tram.json
+++ b/data/transit/route/tram.json
@@ -6,6 +6,8 @@
       "locationSet": {"include": ["it"]},
       "tags": {
         "network": "ATAC",
+        "network:wikidata": "Q530087",
+        "network:wikipedia": "it:ATAC",
         "operator": "ATAC",
         "route": "tram"
       }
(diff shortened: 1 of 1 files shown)


bhousel:
@kjonosm are you really sure that all of these communities are ok with having their `network` tags that they are using renamed?  I worry about stuff like this.

bhousel:
Take this change for example:  
Screen Shot 2020-10-30 at 10 23 44 AM

The effect of this is that, once we deploy this and it's live, iD validator will suggest a tag upgrade for every `network=ZVV` tram line in Switzerland to use the values that you have changed them to.  Is this what you intended?   

In my opinion there is no harm in continuing to use `network=ZVV` tag, but there is clear benefit in adding the `network:wikidata` and `network:wikipedia` tags.

kjonosm:
While it's good to be consistent with naming and to have both versions in the index with `network:short`, I think you're absolutely right to ask that question. We should hold back with merging for now. I brought up your question in the German speaking forum, see [https://forum.openstreetmap.org/viewtopic.php?id=71034](url)

 Let's see if there is some agreement...

kjonosm:
as suspected, there were no clear results in the discussion. i close the pull request and open a new one without the potentially disputed changes you mentioned.

PR #4635, opened 2020-10-31
Add MPREIS miniM

hikemaniac:
With this pull request I added the convenience store chain "miniM" which belongs to the MPREIS supermarket chain. I've set the `brand` tag to miniM because that's what it says on the store logos but it is very frequently referred to as MPREIS miniM (on e.g. Google Maps) so maybe MPREIS would be a better value for the brand tag.

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -1649,6 +1649,18 @@
       }
     },
     {
+      "displayName": "miniM",
+      "id": "minim-d41f40",
+      "locationSet": {"include": ["at"]},
+      "tags": {
+        "brand": "miniM",
+        "brand:wikidata": "Q873491",
+        "brand:wikipedia": "de:MPreis",
+        "name": "miniM",
+        "shop": "convenience"
+      }
+    },
+    {
       "displayName": "Ministop",
       "id": "ministop-4d6b4b",
       "locationSet": {"include": ["001"]},


bhousel:
> With this pull request I added the convenience store chain "miniM" which belongs to the MPREIS supermarket chain. I've set the `brand` tag to miniM because that's what it says on the store logos but it is very frequently referred to as MPREIS miniM (on e.g. Google Maps) so maybe MPREIS would be a better value for the brand tag.

It's really a judgement call whether this would be ok, and something that only a local mapper could decide.  

Would it be weird to see a map that uses this logo for a "miniM" convenience store?  
- If it's ok, then no problem using `brand=MPREIS` and `brand:wikidata=Q873491`
- If it's weird, then we should use `brand=miniM` and create a new Wikidata page for it, and add Wikidata properties that link it to MPREIS.

Screen Shot 2020-11-01 at 12 14 43 PM

hikemaniac:
Adding a Wikidata page for miniM would not make much sense, since they don't have different social media channels and share the same website. miniM is really just the branding MPREIS uses for their smaller convenience store variant.

Adamant36:
> Adding a Wikidata page for miniM would not make much sense since they don't have different social media channels and share the same website. miniM is really just the branding MPREIS uses for their smaller convenience store variant.

Just a suggestion, but you could forgo the social media and website since they aren't used by OSM anyway and then get the logo directly from Wikidata/Wikicommons. It's basic enough to not get deleted for copyright reasons and would then say MiniM.  Plus, it's worth creating unique Wikidata entries for subsidiaries anyway. I'm way more of a nerd about that then other people might be though, but it does help to have the logo in OSM that is actually on the stores' signs.

bhousel:
> Plus, it's worth creating unique Wikidata entries for subsidiaries anyway. I'm way more of a nerd about that then other people might be though, but it does help to have the logo in OSM that is actually on the stores' signs.

I agree!  I'm going to merge this PR as-is, but @Adamant36 if you want to create a separate Wikidata for it and update the QID later, go for it 👍

hikemaniac:
Thank you! 👍

PR #3418, opened 2019-12-04
Add Holland Blumen Mark (dissolved)

hikemaniac:
This PR adds the dissolved florist store chain "Holland Blumen Mark" to the index. There are still quite a few mapped although most of them have been closed or taken over by a different company. Wikidata already contains the dissolved information and running `npm run wikidata` should therefore also add it to dissolved.json.

PR #4636, opened 2020-10-31
Add Baumax

hikemaniac:
With this PR I added the dissolved store chain "Baumax" ([Q310499](https://www.wikidata.org/wiki/Q310499)). Once `npm run wikidata` is run it should be added to `dissolved.json`.

### DIFF
--- a/data/brands/shop/doityourself.json
+++ b/data/brands/shop/doityourself.json
@@ -70,6 +70,20 @@
       }
     },
     {
+      "displayName": "Baumax",
+      "id": "baumax-bc270a",
+      "locationSet": {
+        "include": ["at", "cz", "si", "sk"]
+      },
+      "tags": {
+        "brand": "Baumax",
+        "brand:wikidata": "Q310499",
+        "brand:wikipedia": "de:Baumax",
+        "name": "Baumax",
+        "shop": "doityourself"
+      }
+    },
+    {
       "displayName": "BayWa",
       "id": "baywa-db1fbb",
       "locationSet": {"include": ["de"]},


bhousel:
Thanks @hikemaniac - we can add this, but it won't become an iD or JOSM preset because of it's dissolved status.  But it can be useful for people who want to find them and mark them as closed in OSM.

hikemaniac:
Thank you! That's exactly why I'm adding this information, so that people can easily spot closed/replaced shops :+1:. However, for some reason dissolved brands do appear in iD even though they are listed in `dissolved.json`, e.g. Holland Blumen Mark, which I added last year in #3418. :confused:

bhousel:
> However, for some reason dissolved brands do appear as presets in iD even though they are listed in `dissolved.json`

This is on purpose.. I believe people didn't want presets generated from dissolved brands, but I can't find discussion about it now.  It's been this way for a while.  

Maybe we should relax this rule?  I think it's reasonable to map things that are dissolved, particularly now that we are in a prolonged  ["retail apocalypse"](https://en.wikipedia.org/wiki/Retail_apocalypse).  😱  All those stores are going to sit empty for a while.  

I also think mapping out the death of retail would be a fascinating topic of research, for any data scientists out there.

hikemaniac:
> This is on purpose.. I believe people didn't want presets generated from dissolved brands, but I can't find discussion about it now. It's been this way for a while.

Sorry, I think I might have misunderstood something. From what I see in the current version of iD, presets for dissolved brands are included (and not in any way flagged as dissolved).

> Maybe we should relax this rule? I think it's reasonable to map things that are dissolved, particularly now that we are in a prolonged "retail apocalypse". 😱 All those stores are going to sit empty for a while.

As far as I can see iD includes dissolved brand presets but it would be useful if a lifecycle prefix would be added to these presets.

Issue #2928, opened 2019-08-02
amenity/car_sharing should have an operator tag


tas50:
It was pointed out to me when updating nodes for Zipcar that car_sharing nodes should have an operator tag per the wiki. We should probably add this to all of them matching the brand tag. https://wiki.openstreetmap.org/wiki/Tag:amenity=car%20sharing?uselang=en

bhousel:
Yeah ......
in cf21ce5  I removed all the `operator` tags and I think it was probably a bit too heavy-handed.  There are entries in this index where `operator` tags really are appropriate  (also for post offices and post boxes).

I think I need to relax the schema rule prohibiting the `operator` tags, and maybe we just pay more attention to the index so that they don't sneak in except for certain key/value types.

DavidMoraisFerreira:
+1 for an operator tag in [amenity=charging_stations](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station#Examples) :)

bhousel:
Thanks for your patience everyone - I've moved `amenity/car_sharing` and `amenity/charging_station` to the 'operators' tree.

PR #4638, opened 2020-11-01
Created "post_depot.json"

UKChris-osm:
Created "_post_depot.json_" to house operator data for post / mail sorting depots, starting with **Royal Mail (UK)** and **An Post (Ireland)**.

This may not be a welcome addition to the NSI, but I thought rather than simply ask about it, I would create a pull request so it could be reviewed easily.

### DIFF
--- a/data/brands/amenity/post_depot.json
+++ b/data/brands/amenity/post_depot.json
@@ -11,6 +11,16 @@
       }
     },
     {
+      "displayName": "La Poste",
+      "locationSet": {"include": ["fr"]},
+      "tags": {
+        "amenity": "post_depot",
+        "operator": "La Poste",
+        "operator:wikidata": "Q373724",
+        "operator:wikipedia": "fr:La Poste (entreprise française)"
+      }
+    },
+    {
       "displayName": "Royal Mail Delivery Office",
       "locationSet": {"include": ["gb"]},
       "tags": {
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @UKChris-osm !  
We've explored the idea of using `operator`/`operator:wikidata`/`operator:wikipedia` in a few previous tickets:  #2846, #2863, #2883, #2928,  #3337, and probably more...

I think after the late summer work on identifiers (#3995), this project is _finally_ in a place where we can actually do this.  🎉 

In #4231, I added a 'transit' tree where `network:wikidata` is the main tag, so what I want to do next is add an 'operators' tree where `operator:wikidata` is the main tag...

Then we can then start moving stuff from the "brands" tree to the "operators" tree for things like post offices, mailboxes, infrastructure, utility stuff, etc. 

Are you ok if I close this PR and do this instead? ☝️  - then you can add the post depots in a better place..

By the way, items in the brand tree can still have `operator` tags and items in the operator tree can still have `brand` tags - the difference is which tag we consider the main one.

UKChris-osm:
@bhousel I'm happy for this to be closed as you seem to have a game plan. I was going to ask about post boxes (Royal Mail), phone boxes (BT), telephone cabinets (Openreach) etc but it seems you already have this in mind 👍

PR #4641, opened 2020-11-03
Update trade.json

kjonosm:
none of these brands appear in the osm database according to overpass-turbo.
do they need to be deleted manually or are they deleted automatically with the next build?

### DIFF
--- a/data/brands/shop/trade.json
+++ b/data/brands/shop/trade.json
@@ -28,36 +28,6 @@
       }
     },
     {
-      "displayName": "Bygger'n",
-      "id": "byggern-4df98d",
-      "locationSet": {"include": ["001"]},
-      "tags": {
-        "brand": "Bygger'n",
-        "name": "Bygger'n",
-        "shop": "trade"
-      }
-    },
-    {
-      "displayName": "Byggmakker",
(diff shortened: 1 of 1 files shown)


bhousel:
> none of these brands appear in the osm database according to overpass-turbo.
> do they need to be deleted manually or are they deleted automatically with the next build?

It looks like some of these are really a brand:  
I changed the Overpass query to use '~' instead of '=':  https://overpass-turbo.eu/s/ZHq

My guess is that someone recently (since the last time we processed the planet) mass edited them all to have the unique store name in the `name` tag.

Screen Shot 2020-11-03 at 2 58 01 PM

If Norwegian mappers really want to do that, it's fine, we can remove the `name` tags from NSI.  
We really also have to change our generated Overpass queries to check `brand` tag also, so we catch stuff like this.

kjonosm:
if used like that, can these brands be kept in the index without the "name" tag?

bhousel:
> if used like that, can these brands be kept in the index without the "name" tag?

Yes, this is allowed now.

kjonosm:
> Yes, this is allowed now.

OK, i will close this and open another pull request.

PR #4644, opened 2020-11-04
Update trade.json

kjonosm:
Bygger'n , Byggmakker, Byggtorget:
- update location sets
- add wikidata/wikipedia
- delete "name" Tag as local mappers differentiate names between branches


bhousel:
Looks great, thanks for following up on these 👍 
(CI failures are unrelated - my fault)

PR #4646, opened 2020-11-04
Created "post_box.json" for operator

UKChris-osm:
Created "post_box.json" (operator), starting with Royal Mail.

### DIFF
--- a/config/filter_operators.json
+++ b/config/filter_operators.json
@@ -1,5 +1,5 @@
 {
-  "keepTags": ["^amenity/post_(depot|office)$"],
+  "keepTags": ["^amenity/post_(box|depot|office)$"],
   "discardKeys": [],
   "discardNames": [
     "^(buzón\\sde\\s)?correos$",
--- a/data/operators/amenity/post_box.json
+++ b/data/operators/amenity/post_box.json
@@ -1,0 +1,14 @@
+{
+  "operators/amenity/post_box": [
+    {
+      "displayName": "Royal Mail",
+      "locationSet": {"include": ["gb"]},
+      "tags": {
+        "amenity": "post_box",
+        "operator": "Royal Mail",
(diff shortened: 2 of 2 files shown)


bhousel:
This seems pretty good..
I was actually thinking of making the post_box category "automatic" kind of like what we were discussing on the ATM ticket (#2883).   This isn't a thing yet, but I could hack it together tomorrow probably?  What do you think?

UKChris-osm:
When you say automatic, do you mean auto-generating a file with many data points for post box's around the world, in the same way you had done for transit? At which point we could add wikidata and locationSet after? I'm happy with that method and thing it would move the project along much more quickly 👍

bhousel:
> auto-generating a file with many data points for post box's around the world,

Yes sort of.. I'm thinking of a rule that generates a `amenity=post_box` item for every `amenity=post_office` item.

UKChris-osm:
Are you thinking of auto-populating the post box data (operator) with post office data (operator)?

bhousel:
> Are you thinking of auto-populating the post box data (operator) with post office data (operator)?

yes, exactly! 👍

UKChris-osm:
My only concern in that would be the population of Post Office data in to Post Boxes which are operated by other entities.

In the UK for example, the UK Post Office doesn't operate the post boxes, the Royal Mail do.

This wouldn't be an issue for the UK as I can happily update the entry pretty quickly, but I'm not sure how other Countries operate?

Would you be able to build the script to compare current operator data assigned to post boxes to see if they match post office data before populating the NSI, for example:

- check post boxes in each Country for any pre-existing operator data.
- if found, does this data match the post office operator?
- if yes, populate, if not, leave blank!
- if multiple operator data points are found, leave blank!

Does that make sense?

UKChris-osm:
Closing so @bhousel can implement automatically

PR #4663, opened 2020-11-11
Add Getränke City

Robot8A:


bhousel:
Thanks @Robot8A 👍

Robot8A:
> Thanks @Robot8A +1

Now I live in Germany, so expect to see less brands from Spain and more from Germany from now on, haha

PR #4664, opened 2020-11-11
Add a bunch of transit agencies in the Puget Sound Region

evansiroky:


data/transit/route/bus.json
```diff
@@ -2245,9 +2245,14 @@
       "displayName": "Community Transit",
       "id": "communitytransit-a21cc9",
       "locationSet": {"include": ["us"]},
+      "matchNames": ["ct"],
```
tas50, line 2248: Have you confirmed that this isn't going to be too aggressive of a match?

evansiroky: I guess I'm not sure how this is supposed to work. At my job we refer to this transit agency as "CT" sometimes. I can remove if this seems inappropriate for this project. Same thing goes for King County Metro ("KCM") in this PR I suppose.

tas50: We really only need the matchtag is there are a lot of instances on OSM right now that are tagged with that name, which we'd want to upgrade. Matching on abbreviations is generally going to be pretty problematic since there's probably a good number of systems in the US that folks we call CT (think Caltrain in the bay area)

evansiroky: Ok, I just removed the matchNames with abbreviations in 2716881297334bd70010aba350dc4119128079c6.

bhousel: Thanks!  It's definitely ok to put them back later if you narrow the scope of the locationSet to just the Puget Sound region, like with a geojson drawn around there.

PR #4669, opened 2020-11-12
Update pharmacy.json

kjonosm:
update "Rexall" to match the more appropriate Canadian pharmacy brand (fix location set and wikidata)


bhousel:
Thanks for fixing @kjonosm 👍 
I'm going to also add this to the `replacements.json` file to help remap the existing Rexall pharmacies in Canada.

There were a handful of old Rexall tagged items in the US (all historic or individual locations that happen to still be using the Rexall name) so I removed the `brand:wikidata` and `brand:wikipedia` from them: https://overpass-turbo.eu/s/102C

PR #4671, opened 2020-11-12
Add Santa Ynez Valley Transit

evansiroky:
This transit agency didn't have a Wikidata or Wikipedia page. I created a Wikidata page, but didn't create a Wikipedia page. Is that ok?


bhousel:
Thanks @evansiroky looks good to me 👍

PR #4687, opened 2020-11-14
Update charging_station.json

kjonosm:
- add operator wikidata
- merge duplicates


bhousel (changes requested):
Looks good but please add the `operator:wikipedia` as well..

bhousel (approved):

bhousel:
> Looks good but please add the `operator:wikipedia` as well..

Now we can just merge these as-is and the script will fill them in, per #4716

PR #4688, opened 2020-11-14
Update car_sharing.json

kjonosm:
- add operator Wikidata


bhousel (changes requested):
Looks good but please add `operator:wikipedia` as well..

bhousel (approved):

PR #4704, opened 2020-11-15
 wikidata on french brands

Olyon:


bhousel:
These look great, thanks @Olyon 👍

Issue #4667, opened 2020-11-12
UK bank Virgin Money rebranding


kymckay:
Became aware that some UK bank brands have been acquired by Virgin and are rebranding:

https://www.virginmoneyukplc.com/newsroom/news-and-releases/2020/virgin-money-begins-national-store-rebrand-programme

Just opening this issue to track this in the index. I believe relevant brands are:
- Clydesdale bank (I can confirm a full rebranding in progress across all stores)
- Yorkshire bank (rebranding status unknown to me)

UKChris-osm:
Is the rebranding of all Clydesdale banks happening now (or soon), because we could add a matchname for "_Clydesdale_" to the existing entry for Virgin Money and remove Clydesdale bank from the index if that's the case.

kymckay:
According to their correspondence it's happening across all branches now, but whether that matches group truth who knows (e.g. there are still "Co-operative Food" signed locations even though they rebranded to just "Co-op"

PR #4711, opened 2020-11-16
"Clydesdale Bank" becoming "Virgin Money"

UKChris-osm:
As per #4667 "_Clydesdale Bank_" is rebranding branches to become "_Virgin Money_" after Clydesdale acquired Virgin.

I've removed "_Clydesdale Bank_" and added a **matchNames** to "_Virgin Money_".

I've done this as a pull request to make sure I've followed the correct process.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -7769,6 +7769,7 @@
       "id": "virginmoney-e1e9d0",
       "locationSet": {"include": ["gb"]},
       "matchNames": ["Clydesdale", "Clydesdale Bank"],
+      "note": "osmlab#4667 - Clydesdale Bank branches have been rebranded to become Virgin Money.",
       "tags": {
         "amenity": "bank",
         "brand": "Virgin Money",
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -3471,18 +3471,6 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


bhousel:
Looks good to me 👍  
The only other thing I'd add is a `note` that links to issue #4667 in GitHub, just in case someone comes along later looking for Clydesdale.

UKChris-osm:
Added a note 👍 

Perhaps a future update could allow the ability to add something like **"rebranded": "virginmoney-e1e9d0"** to old entries, rather than deleting them right away, allowing the [nsi.guide](https://nsi.guide) web site to continue listing recently removed entries so people can see "Oh, this bank was there but is now another brand"?

    {
      "displayName": "Clydesdale Bank",
      "id": "clydesdalebank-e1e9d0",
      "locationSet": {"include": ["gb"]},
      "rebranded": "virginmoney-e1e9d0",
      "tags": {
        "amenity": "bank",
        "brand": "Clydesdale Bank",
        "brand:wikidata": "Q971680",
        "brand:wikipedia": "en:Clydesdale Bank",
        "name": "Clydesdale Bank"
      }
    }

bhousel:
> Perhaps a future update could allow the ability to add something like **"rebranded": "virginmoney-e1e9d0"** to old entries

Maybe, but I'm definitely wary about making extra work for myself.
I kind of think it would be better for someone to just update them all in OSM with a bulk edit.

PR #4714, opened 2020-11-17
Trying out operators for Police, Fire and Ambulance services

UKChris-osm:
As previously attempted (#3337) prior to the "_operator_" tag being added to the index, this PR is aimed at trying out operators for Police, Fire and Ambulance services, part filled with UK based data at the moment.

It is possible these services would be better added under automation, similar to previous additions (#4646, #4638 ).

### DIFF
--- a/data/operators/amenity/fire_station.json
+++ b/data/operators/amenity/fire_station.json
@@ -1,0 +1,34 @@
+{
+  "operators/amenity/fire_station": [
+      {
+      "displayName": "Cornwall Fire and Rescue Service",
+      "locationSet": {"include": ["gb-eng"]},
+      "tags": {
+        "amenity": "fire_station",
+        "operator": "Cornwall Fire and Rescue Service",
+        "operator:wikidata": "Q5171976",
+        "operator:wikipedia": "en:Cornwall Fire and Rescue Service"
+      }
+    },
+    {
+      "displayName": "Devon and Somerset Fire and Rescue Service",
+      "locationSet": {"include": ["gb-eng"]},
+      "tags": {
+        "amenity": "fire_station",
+        "operator": "Devon and Somerset Fire and Rescue Service",
+        "operator:wikidata": "Q5267788",
+        "operator:wikipedia": "en:Devon and Somerset Fire and Rescue Service"
+      }
+    },
+    {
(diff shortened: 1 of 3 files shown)


bhousel:
Thanks @UKChris-osm , this looks pretty good to me 👍 
I think I'd prefer if people only add these in situations where the `operator` is used a bunch of times.  In other words, if it's just a small town with a single police station, we don't need to add an item for it.  But for a larger town with a lot of police stations, it definitely makes sense.

You can browse through https://github.com/osmlab/name-suggestion-index/blob/main/dist/collected/operators_all.json to see all of the fire, police, rescue operators that we collected the last time I processed the OSM planet.  There are really a bunch of these!

UKChris-osm:
I'd agree with the non-single or small entries, but do you have a target number?

For example, from the [operators_all.json](https://github.com/osmlab/name-suggestion-index/blob/main/dist/collected/operators_all.json) file:

- "emergency/ambulance_station|East Midlands Ambulance Service": 11,
- "emergency/ambulance_station|London Ambulance Service": 15,
- "emergency/ambulance_station|North West Ambulance Service": 11,
- "emergency/ambulance_station|Yorkshire Ambulance Service": 13,
- [South Western Ambulance Service](https://overpass-turbo.eu/s/10fY) (Overpass link, not in the file)

... would you say that was to few entries to justify inclusion? Baring in mind it's unlikely all locations are currently mapped in OSM at the moment?

bhousel:
> ... would you say that was to few entries to justify inclusion? Baring in mind it's unlikely all locations are currently mapped in OSM at the moment?

Currently I have the cutoff for `operator` collection set to 10.. From what we've seen of other POI coverage, the data in OSM is pretty fragmented and incomplete, so if it finds 10, there are likely many more.

> Also, can the NSI target specific Counties (England: Somerset, Devon, Cornwall, etc) or is **gb-eng**, **gb-sct**, etc about as specific as we can be??

You would  need to supply a geojson to get more specific than what's supported by country-coder, but this is allowed now!  
I built a thing last week to browse all the supported codes:  https://ideditor.codes

bhousel:
I know you marked it as draft, but I'm going to just make it a normal PR and merge it as-is.. 
We've wanted it for a while and the underlying work is finally done to make it happen.

PR #4719, opened 2020-11-21
Ukraine amenities

skfd:
### added cafe
* Aroma Kava
### removed generic cafes
* Берёзка
* Виктория
* Лакомка
* Мечта
* Оазис
* Парус
* Сказка
### added pharmacies
* 1 Соціальна Аптека
* Аптека доброго дня
### added cafe\dairy shop
* Молоко від фермера
### added medical lab
Synevo
### misc
fixed some `locationSet`s

### DIFF
--- a/config/filter_brands.json
+++ b/config/filter_brands.json
@@ -50,7 +50,7 @@
     "^amenity/bank\\|volksbank( |-)raiffeisenbank$",
     "^amenity/bank\\|vr( |-)bank$",
     "^amenity/bar\\|уют$",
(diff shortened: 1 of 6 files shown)


bhousel:
Thanks @skfd that looks really great 👍   We can always use more help with Ukraine!

One thing I changed:  we don't actually add the `website` tag to the items in this index because the OSM convention is for that tag to be a website for that location, not for the overall brand.  Consumers who need the website for a brand can get it from wikidata.   So I removed the `website` tags in 331ab54e / b27431fa

PR #4720, opened 2020-11-23
Add Wikidata tags to MTS transit entry

mds08011:
Adds operator and network Wikidata information to the MTS entry.  Please scrutinize this PR as it's my first use of the new transit category.


bhousel:
Looks great to me @mds08011, thanks 👍 
I did a quick check on overpass and it does look concentrated around San Diego, so this seems fine.

PR #4721, opened 2020-11-23
Update toys.json

kjonosm:
add match name to merge "Smyths Toys" and "Smyths"


Identitaet:
maybe we should also add Toys r us as a matchname since they took over their stuff (in europe at least the us seems to be different)
Looks very good though.

PR #4725, opened 2020-11-24
Tiny correction in CategoryInstructions.jsx

UKChris-osm:
Tiny correction in "CategoryInstructions.jsx" where the nsi.guide web site is displaying "aoperator" (with an 'a' at the beginning) instead of "operator" within the "_Some things you can do here:_" instructions!

This minor edit is done as a pull request, just in case another file needs to be edited that I'm not aware of.

If merged, I believe **npm run docbuild** will be required to make this change live on the nsi.guide web site!


bhousel:
Oh thanks!  Good catch 👍

PR #4734, opened 2020-11-25
Update bicycle_rental.json

kjonosm:
the common German tagging scheme for bike sharing systems seems to be:

`name=*`  -  name of the actual station within the system
`brand=*`  -  branded name of the sharing system itself
`operator=*`  -  name of the company that operates the system

### DIFF
--- a/data/operators/amenity/bicycle_rental.json
+++ b/data/operators/amenity/bicycle_rental.json
@@ -18,6 +18,7 @@
       "tags": {
         "amenity": "bicycle_rental",
         "brand": "fLotte",
+        "brand:wikidata": "Q102333872",
         "operator": "ADFC",
         "operator:wikidata": "Q24349",
         "operator:wikipedia": "de:ADFC"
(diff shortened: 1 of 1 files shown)


bhousel:
Wondering whether we should move these back to the 'brand' tree? 
It seems like the logo we'd show people would be more tied to the brand than the operator..

kjonosm:
well, at least for Germany it might make sense. I don't know how it's done in other countries though...

bhousel:
> well, at least for Germany it might make sense. I don't know how it's done in other countries though...

I think bicycle_rental and car_sharing are similar and could be either considered brands or operators (but leaning more towards brands).  Post offices can also be either (but it sounds like people are learning more towards operators for those).

UKChris-osm:
Serco Group operate the London Bicycle's, but I would say the "_Santander_" brand dominates in this case as well.

UKChris-osm:
Looking at this more, I think it would be good if there was a separate Wikidata for [fLotte](https://flotte-berlin.de/), as they have a [Facebook](https://www.facebook.com/fLotteBerlin/) / [Twitter](https://twitter.com/fLotteBerlin) account where the logo could be grabbed.

kjonosm:
> a separate Wikidata for fLotte

fixed!

Issue #4735, opened 2020-11-25
Life Boat Stations


UKChris-osm:
I have been thinking that Life Boat Stations would be suitable for the NSI, for example:

- [Royal National Lifeboat Institution](https://en.wikipedia.org/wiki/Royal_National_Lifeboat_Institution) (UK & IRE) - 238 stations
- [Koninklijke Nederlandse Redding Maatschappij](https://nl.wikipedia.org/wiki/Koninklijke_Nederlandse_Redding_Maatschappij) (NL) - 46 stations
- [Société Nationale de Sauvetage en Mer](https://fr.wikipedia.org/wiki/Soci%C3%A9t%C3%A9_nationale_de_sauvetage_en_mer) (FR) - 214 stations
- [Deutsche Gesellschaft zur Rettung Schiffbrüchiger](https://de.wikipedia.org/wiki/Deutsche_Gesellschaft_zur_Rettung_Schiffbr%C3%BCchiger) (DE) - 55 stations
- [Redningsselskapet](https://no.wikipedia.org/wiki/Redningsselskapet) (No)
- [Sjöräddningssällskapet](https://sv.wikipedia.org/wiki/Sj%C3%B6r%C3%A4ddningss%C3%A4llskapet) (SE)

If suitable, would these come primarily under brand, operator, or network? I'd go with operator, but Life Boat Stations could be called a network as well.

If included, would **emergency=lifeboat_station** or **amenity=lifeboat_station** be used?

bhousel:
> If suitable, would these come primarily under brand, operator, or network? I'd go with operator, but Life Boat Stations could be called a network as well.

yes, looks suitable!.. I'd definitely go with `operator` as the tree, but no idea about the tag.  

Both `emergency=lifeboat_station` and `amenity=lifeboat_station` seem used and documented, 
and `operators_all.json` has collected a few from both tags 🤷

Adamant36:
Personally, I'd go with `emergency=lifeboat_station` since it's more consistent with similar tags like `emergency=life_ring`. Plus, there seems to be a general move going on to move tings out of the amenity tag that aren't specifically amenities and can be put in other tags.

UKChris-osm:
Merged via [2778f60](https://github.com/osmlab/name-suggestion-index/commit/2778f60992628c66fafd0a0faabd87ec4121aff8)

PR #4737, opened 2020-11-26
Life Boat Stations

UKChris-osm:
As per #4735, I have created "_lifeboat_station.json_" under **/operator/emergency**, populated with a few Rescue Services. I also (hopefully correctly) created a matchGroup.

I'm not sure if I have the use of **operator:_lang_** correct, so please double check this.


bhousel:
All looks good to me, thanks @UKChris-osm 👍

PR #4740, opened 2020-11-29
Keep table headers on top when scrolling

kymckay:
Prevents brand icons from weirdly overlapping when the page is scrolled and the table headings stick to the top of the screen

Would usually just push this, but here's a PR just in case it could cause any unintended side effects

### DIFF
--- a/app/style.css
+++ b/app/style.css
@@ -237,6 +237,7 @@ table.summary th {
   font-size: 15px;
   font-weight: bold;
   color: #000;
+  z-index: 1; /* stay on top when scrolling */
 }
 .dark table.summary th {
   background: #1a1a1a;


kymckay:
I've actually just realised it's only with the dark theme enabled that this behaviour was observable which is interesting

UKChris-osm:
When dark mode is activated, the CSS code "_filter: contrast(0.7);_" seems to be applied to the icons, which is causing the icon to then overlap the table header.

bhousel:
Cool, thanks for the fix @kymckay 
I guess this is because adding a CSS filter creates a new stacking context, [see stackoverflow for some explanation](https://stackoverflow.com/questions/25764404/why-does-stacking-order-change-on-webkit-filter-hover).

PR #4742, opened 2020-11-30
Updated Category Instructions

UKChris-osm:
Just a minor edit, I've changed the instruction text "_View on Overpass Turbo_" to "_Search Overpass Turbo_" just to match the link text we use now.


bhousel:
Thanks! You can merge stuff like this 👍

PR #4749, opened 2020-12-01
Add Thai department store chain

michaelblyons:
`name` tag intentionally missing. The stores themselves are named like "Central Rama II," "Central Rama IX," "Central Chitlom," etc.


bhousel (approved):
looks good to me , thanks @michaelblyons  👍

Issue #4751, opened 2020-12-01
Clean up the police


michaelblyons:
My favorite JSON in the project is currently

```jsonc
    {
      "displayName": "Police",
      "id": "police-d00fc1",
      "locationSet": {"include": ["001"]},
      "tags": {
        "amenity": "police",
        "operator": "Police"
      }
    },
```

but there are plenty of others similarly suspicious:

- Police Nationale
- Policía Nacional
- Polizia (maybe false-positive? Limited to Italy, but there are more-specific options for Italy right there)
- Several acronyms (e.g. PSP) suggested for `001`

michaelblyons:
Actually _now_ my favorite JSON in the project is in `post_offices.json`...

```jsonc
    {
      "displayName": "???",
      "id": "d41d8c-91a3cc",
      "locationSet": {"include": ["001"]},
      "tags": {
        "amenity": "post_office",
        "brand": "???",
        "name": "???",
        "operator": "???"
      }
    },
```

bhousel:
> ```js
>  "displayName": "???",
> ```

haha I know, I saw that too.. ❓

UKChris-osm:
>  "displayName": "???",

Is this not the Post Office brand created by the _Black Eyed Peas_ when they were asking where [all the love was](https://feminismandreligion.files.wordpress.com/2019/10/where-is-the-love-1280x720.jpg) 👍

PR #4752, opened 2020-12-01
Start cleanup of police

michaelblyons:
See #4751

### DIFF
--- a/data/operators/amenity/police.json
+++ b/data/operators/amenity/police.json
@@ -156,8 +156,8 @@
     },
     {
       "displayName": "Cuerpo de Agentes de Control Municipal",
-      "id": "cuerpodeagentesdecontrolmunicipal-d00fc1",
-      "locationSet": {"include": ["001"]},
+      "id": "cuerpodeagentesdecontrolmunicipal-210a6b",
+      "locationSet": {"include": ["ec"]},
       "tags": {
         "amenity": "police",
         "operator": "Cuerpo de Agentes de Control Municipal"
@@ -368,6 +368,7 @@
       "displayName": "Guardia Civil",
       "id": "guardiacivil-dd66e9",
       "locationSet": {"include": ["es"]},
+      "matchNames": ["ministerio de defensa"],
       "tags": {
         "amenity": "police",
         "operator": "Guardia Civil",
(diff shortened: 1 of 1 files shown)


data/operators/amenity/police.json
```diff
@@ -755,12 +775,13 @@
       "id": "policianacionaldebolivia-23a32a",
       "locationSet": {"include": ["bo"]},
       "matchNames": [
-        "policia nacional boliviana (pnb)",
-        "policia nacional de bolivia (pnb)"
+        "policia nacional boliviana",
```
bhousel, line 778: OH - I left the "(pnb)" part in the `matchName` because these were collected from the OSM planet that way?

https://github.com/osmlab/name-suggestion-index/blob/ab1b26c99b68b249a41edf8e73d07fe7fd86f88a/dist/collected/operators_all.json#L6649-L6652

Sometimes matchName needs to match the weird things that people have previously done.

michaelblyons:
Validation error message is super-misleading. It said

```
Error - Schema validation:
  ./data/operators/amenity/police.json:
  instance.operators/amenity/police[18].tags requires property "brand"
  instance.operators/amenity/police[18].tags requires property "flag:name"
  instance.operators/amenity/police[18].tags requires property "flag:type"
  instance.operators/amenity/police[18].tags requires property "operator"
  instance.operators/amenity/police[18].tags requires property "network"
  instance.operators/amenity/police[18].tags is not any of [subschema 0],[subschema 1],[subschema 2],[subschema 3]
  instance.operators/amenity/police[18] is not allowed to have the additional property "displayName"
  instance.operators/amenity/police[18] is not allowed to have the additional property "id"
  instance.operators/amenity/police[18] is not allowed to have the additional property "locationSet"
  instance.operators/amenity/police[18] is not allowed to have the additional property "tags"
  instance.operators/amenity/police[18] requires property "templateSource"
  instance.operators/amenity/police[18] requires property "templateTags"
  instance.operators/amenity/police[18] is not any of <#/definitions/normalItem>,<#/definitions/templateItem>
```

when the real problem was no default `"operator"` tag (only language-indicated operators).

bhousel:
> Validation error message is super-misleading

Ugh yeah, json-schema errors can be hard to understand.. I've struggled with this too.

bhousel (approved):
This all looks great to me 👍 
You can merge as is and push more commits later - it doesn't have to be perfect.  We're all pretty ok with the idea that the index is constantly changing and improving.

PR #4753, opened 2020-12-01
Some US state flags

michaelblyons:


michaelblyons:
Possible tweak: DC flag is both `flag:type=regional` _and_ `flag:type=municipal`. If you'd prefer the latter, we can switch it.

1ec5:
I guess it depends on the context – whether it’s in a row of state flags or city flags. Some other flags also have types that depend on the context. But I suspect it’s more common to see the D.C. flag representing the district as such.

bhousel:
> Possible tweak: DC flag is both `flag:type=regional` _and_ `flag:type=municipal`

Let's go with `regional`.  It will probably eventually be a state.

PR #4767, opened 2020-12-04
Try and improve the Santander Cycles tagging

peternewman:
https://www.santandercycles.co.uk/


data/brands/amenity/bicycle_rental.json
```diff
@@ -1524,29 +1524,72 @@
       }
     },
     {
-      "displayName": "Santander Cycles",
+      "displayName": "Santander Cycles (Brunel)",
+      "locationSet": {"include": ["gb-eng"]},
```
peternewman, line 1528: Probably needs some custom GeoJSON.

data/brands/amenity/bicycle_rental.json
```diff
@@ -1524,29 +1524,72 @@
...
+    {
+      "displayName": "Santander Cycles (London)",
       "id": "serco-b88dc4",
       "locationSet": {"include": ["gb-eng"]},
       "matchNames": [
         "barclays bikes",
         "barclays cycle hire",
         "boris bike",
-        "boris bikes",
-        "santander cycles mk"
```
peternewman, line 1535: Despite looking through the history, I couldn't make sense of who added this line, but the MK one is operated by nextbike not Serco and isn't in the tfl network.

data/brands/amenity/bicycle_rental.json
```diff
@@ -1524,29 +1524,72 @@
...
+        "network": "nextbike",
+        "operator": "nextbike",
+        "operator:wikidata": "Q2351279",
+        "operator:wikipedia": "en:nextbike"
+      }
+    },
+    {
+      "displayName": "Santander Cycles (London)",
       "id": "serco-b88dc4",
       "locationSet": {"include": ["gb-eng"]},
```
peternewman, line 1543: Probably needs some custom GeoJSON for "London".

data/brands/amenity/bicycle_rental.json
```diff
@@ -1524,29 +1524,72 @@
...
-        "name": "Serco",
         "network": "tfl_cycle_hire",
         "operator": "Serco",
         "operator:wikidata": "Q499104",
         "operator:wikipedia": "en:Serco"
       }
     },
+    {
+      "displayName": "Santander Cycles (Milton Keynes)",
+      "locationSet": {"include": ["gb-eng"]},
```
peternewman, line 1564: Probably needs some custom GeoJSON.

data/brands/amenity/bicycle_rental.json
```diff
@@ -1524,29 +1524,72 @@
...
+      ],
+      "tags": {
+        "amenity": "bicycle_rental",
+        "brand": "Santander Cycles MK",
+        "brand:wikidata": "Q807961",
+        "brand:wikipedia": "en:Santander Cycles",
+        "network": "nextbike",
+        "operator": "nextbike",
+        "operator:wikidata": "Q2351279",
+        "operator:wikipedia": "en:nextbike"
```
peternewman, line 1576: Their tagging is a bit random:
https://overpass-turbo.eu/s/10Rf

Some use name or network.

Including docking station number in the name rather than in the ref.

bhousel:
It doesn't seem like there is anything different about these?  They all point to the same Q807961 Santander Cycles wikidata and have nextbike as the operator/network?

peternewman (commented):
A few things that need thinking about/looking at.

peternewman:
> It doesn't seem like there is anything different about these? They all point to the same Q807961 Santander Cycles wikidata and have nextbike as the operator/network?

There is definitely a different operator for London versus the others. I don't know how much we should respect how local mappers have tagged the network versus trying to be consistent. And again that discussion of what the network is for in something like this.

bhousel:
> There is definitely a different operator for London versus the others. I don't know how much we should respect how local mappers have tagged the network versus trying to be consistent. 

Ok cool - yeah we do want separate items if they have different operators.  I agree with your comments inline above that custom geojsons would make this clearer too. 

> And again that discussion of what the network is for in something like this.
> Should network correspond to just the ones in that cluster, i.e. docking stations you could leave the network at. Although if they're all on the same system I wonder if you could cycle from Hillingdon to MK to undock!?!

Yeah - I am still not sure what to do about that.  Bike sharing systems  _feel_ to me like brands, not networks.  I guess we can leave them in the "brand" tree but insist that people add a `network` tag.   (like how we insist on a `cuisine` tag for restaurants).

I definitely want to try to preserve the tagging that local mappers have established, and try to follow whatever the osm wiki says today.

data/brands/amenity/bicycle_rental.json
```diff
@@ -1524,29 +1524,72 @@
...
+      }
+    },
+    {
+      "displayName": "Santander Cycles (Swansea)",
+      "locationSet": {"include": ["gb-wal"]},
```
UKChris-osm, line 1581: The "_locationSet_" for Wales would need to be  **gb-wls**

peternewman: Thanks @UKChris-osm . Good spot!

kymckay:
> Bike sharing systems feel to me like brands, not networks.

Yeah when I added the bicycle hire scheme here in Edinburgh (e213f8536074ad97a103538c60897715414916a4) it felt more natural to me for the brand to be the company sponsoring it. The brand could change at any time if a new company took over the sponsorship, whereas the underlying network would remain the same.

Similarly, the operator is a 3rd party company (Secro) acting on behalf of Transport for Edinburgh - which could change at any time.

The network, to me, would be the actual instance of a bicycle hire scheme (e.g. what is currently called "Just Eat Cycles", but could equally be referred to as "Edinburgh's bicycle hire shceme").

UKChris-osm:
I've created "_london-cycles.geojson_" inside the "_features/gb/_" folder if it helps with targeting Santander Cycles in London. I've used a generic name, but this could be changed if need be.

peternewman:
> The brand could change at any time if a new company took over the sponsorship, whereas the underlying network would remain the same.

Agreed.

> Similarly, the operator is a 3rd party company (Secro) acting on behalf of Transport for Edinburgh - which could change at any time.

The operator has survived in London, I imagine because they do the work and the brand just pay a sponsorship fee, but they could be re-contracted to someone else.

> The network, to me, would be the actual instance of a bicycle hire scheme (e.g. what is currently called "Just Eat Cycles", but could equally be referred to as "Edinburgh's bicycle hire shceme").
> 
> Edit: So I suppose the network is kind of the primary key in my mind, because it is the stable identifier - to which an unstable brand and operator is associated.

I've tweaked Edinburgh a bit @kymckay . Sorry intended to do a PR but hit the wrong button, feel free to comment and I'll fix and actually open a PR:
https://github.com/osmlab/name-suggestion-index/commit/f52b7f3061fce695af82952792605a8912f40894

@bhousel perhaps I need restricting from merging without an approved review! :laughing:

data/brands/amenity/bicycle_rental.json
peternewman: Done by @UKChris-osm

data/brands/amenity/bicycle_rental.json
peternewman: Done by me.

features/gb/london-cycles.geojson
```diff
@@ -3,6 +3,7 @@
   "features": [
     {
       "type": "Feature",
+      "id": "london-cycles.geojson",
```
peternewman, line 6: Would `npm run build` add these? Could it be done on Travis before it tries to validate if so?

Nope, they're still just failing! :confused:

UKChris-osm:
Your branch doesn't seem to have the same copy of "_featureCollection.json_" as the main branch, so your copy doesn't have the "_london-cycle.geojson_" code in it - I'm not sure if this is causing any problems.

features/gb/london-cycles.geojson
UKChris-osm: I've removed the .geojson file and replaced it with a single GPS location which will include a 25km radius of London

peternewman: That brought up a new issue covered in #4776 so I've switched back.

peternewman:
> Your branch doesn't seem to have the same copy of "_featureCollection.json_" as the main branch, so your copy doesn't have the "_london-cycle.geojson_" code in it - I'm not sure if this is causing any problems.

That was enough of a hint, having tweaked npm it now builds the feature collection before validating which was sufficient to make it pass.

data/brands/amenity/bicycle_rental.json
peternewman: Added

UKChris-osm:
Thanks for this @peternewman 🥇

Issue #4770, opened 2020-12-04
add 'avec', a Swiss convenience store chain


habi:
I wanted to add one close to my home where a `shop=convenience` rebranded and was surprised that https://avec.ch/ is neither suggested in Go Map!! nor in iD.

I'm happy to try a PR, but need some guidance.

habi:
There are some stores mapped in Switzerland already: https://osm.li/RKH

habi:
And *two* in Germany: https://overpass-turbo.eu/s/10SE

PR #4771, opened 2020-12-05
Add avec convenience stores

habi:
This PR is trying to fix https://github.com/osmlab/name-suggestion-index/issues/4770

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -241,6 +241,7 @@
       "locationSet": {"include": ["ch", "de"]},
       "tags": {
         "brand": "avec",
+        "brand:wikidata": "Q103863974",
         "name": "avec",
         "shop": "convenience"
       }
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -237,7 +237,7 @@
     },
     {
       "displayName": "avec",
-      "locationSet": {"include": ["ch"]},
+      "locationSet": {"include": ["ch", "de"]},
       "tags": {
         "brand": "avec",
         "name": "avec",
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -237,13 +237,14 @@
     },
     {
       "displayName": "avec",
+      "id": "avec-c71a8a",
       "locationSet": {"include": ["ch", "de"]},
       "tags": {
         "brand": "avec",
         "name": "avec",
         "shop": "convenience"
       }
-    }, 
+    },
     {
       "displayName": "Avia",
       "id": "avia-4d6b4b",
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -236,6 +236,17 @@
       }
     },
     {
+      "displayName": "avec",
+      "locationSet": {"include": ["ch"]},
+      "matchNames": ["Avec"],
+      "note": "Convenience store, often at gas stations.",
+      "tags": {
+        "brand": "avec",
+        "name": "avec",
+        "shop": "convenience"
+      }
+    }, 
+    {
       "displayName": "Avia",
       "id": "avia-4d6b4b",
       "locationSet": {"include": ["001"]},
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -238,7 +238,6 @@
     {
       "displayName": "avec",
       "locationSet": {"include": ["ch"]},
-      "matchNames": ["Avec"],
       "note": "Convenience store, often at gas stations.",
       "tags": {
         "brand": "avec",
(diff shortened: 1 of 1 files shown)


habi:
Open questions: How do I get or generate the `id`?

habi:
Ah: "When adding new data, don't add the id line (key and value)."

But do I now need to run `npm run build` on this on my local machine?
I.e. can I not submit a PR solely on GitHub?

UKChris-osm:
Hi @habi 

If you have made your edit via GitHub directly, such as by editing the "_data/brands/shop/convenience.json_", then you can still submit a PR (it's how I always do it) but it will mean you don't include an "_id_" value in your edit, as this "_id_" will be generated later by the build script.

If you made your edits on a local machine, you could run "_npm run build_" before your PR so the "_id_" is generated there and then.

In either case, the "_id_" will show up eventually 👍 

You can remove the "_matchNames_" in this case, as it's the same name, just capitalised.

Does your entry have a "_brand:wikidata_" that can go with it?

Adamant36:
You should get rid of `"note": "Convenience store, often at gas stations.",` also. Since the index doesn't include notes.

bhousel:
Thanks @habi it's a good start 👍 
I agree with what the other commenters said.. Also in #4770 you said there were a few in Germany, so make sure to include `de` in the locationSet too, and whatever countries this brand is in..

habi:
> If you have made your edit via GitHub directly, such as by editing the "_data/brands/shop/convenience.json_", then you can still submit a PR (it's how I always do it) but it will mean you don't include an "_id_" value in your edit, as this "_id_" will be generated later by the build script.
> 
> If you made your edits on a local machine, you could run "_npm run build_" before your PR so the "_id_" is generated there and then.

When I `npm run build` I get

````zsh
(base) ➜  name-suggestion-index git:(patch-1) npm run build

> name-suggestion-index@4.0.2 build
> run-s build:features build:index

sh: run-s: command not found
npm ERR! code 127
npm ERR! path /Users/habi/Dev/name-suggestion-index
npm ERR! command failed
npm ERR! command sh -c run-s build:features build:index

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/habi/.npm/_logs/2020-12-05T19_53_29_270Z-debug.log
````
and don't have the mental bandwidth to figure out what's wrong from the log, so I just left the `id` out for the moment.

habi:
> You can remove the "_matchNames_" in this case, as it's the same name, just capitalised.

I did https://github.com/osmlab/name-suggestion-index/pull/4771/commits/681f576ed3f0141f7f217190e57d0ae6e9f55425

habi:
> Does your entry have a "_brand:wikidata_" that can go with it?

I could not find anything relevant on Wikidata.org, so I guess no.

habi:
> You should get rid of `"note": "Convenience store, often at gas stations."`

I removed the note here: https://github.com/osmlab/name-suggestion-index/pull/4771/commits/33f81ebe936d7e92586ee0f6482e4e8910ae1d41
The index *does* include notes, though, see https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/shop/convenience.json#L253 (which I used as a template).

habi:
> make sure to include `de` in the locationSet too

I did that here: https://github.com/osmlab/name-suggestion-index/pull/4771/commits/1d9cbcfc5491ca08bbda876dcc4c2243ed0dd5e5 although I'm not sure if that's the same brand/chain.

Unfortunately, their [store finder](https://avec.ch/de/standorte/?Service=9,35,36) does not work in any browser that I tried, so I cannot confirm or falsify this :)

bhousel:
> ```shell
> sh: run-s: command not found
> ```

Oh yeah, this means you need to do `npm install` to install the dependencies..

UKChris-osm:
> I removed the note here: [33f81eb](https://github.com/osmlab/name-suggestion-index/commit/33f81ebe936d7e92586ee0f6482e4e8910ae1d41)
> The index _does_ include notes, though, see https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/shop/convenience.json#L253 (which I used as a template).

I think the comment about your note was more about it not being necessary in this case, because you're adding an entry to the file called "_convenience.json_", so we wouldn't need a note telling us it's a convenience store :)

habi:
> I think the comment about your note was more about it not being necessary in this case,

Of course, my answer sounded harsher than intended.

habi:
> you need to do `npm install` to install the dependencies..

I've just very recently started to use `npm` when working on a home improvement project with a fork of a library: https://github.com/habi/epaper.js/tree/localize
That did the trick, now there's an `id`, too.

UKChris-osm:
I've created [Q103863974](https://www.wikidata.org/wiki/Q103863974) for you to add as well :)

habi:
> I've created [Q103863974](https://www.wikidata.org/wiki/Q103863974) for you to add as well :)

Thanks, I've added it to the PR.

UKChris-osm:
Thanks for doing this @habi 👍

PR #4778, opened 2020-12-06
bus networks

Olyon:
addition of information on French bus networks and removal of networks no longer appearing in osm


bhousel:
Thanks @Olyon for adding this info for us, looks good 👍

PR #4782, opened 2020-12-07
Add Indego bikeshare

quincylvania:
Hi friends, I haven't contributed to NSI in awhile so please let me know if this needs changes.

https://www.wikidata.org/wiki/Q19876452
https://www.wikidata.org/wiki/Q104005517


UKChris-osm:
I've added the wikipedia link for Indego into this PR @quincylvania, otherwise all fine I think 👍

PR #4783, opened 2020-12-07
Change US Post Office to US Postal Service

michaelblyons:
Some parts of this may be somewhat controversial.

### Hopefully innocuous

- Changing the `brand` and `operator` to USPS.
    + This is in line with what people are actually tagging in the wild. Compare 1,800 items on Overpass with `operator=United States Postal Service` to ~70 with `United States Post Office`.
    + That's also what the operator is actually called.
- `brand:short`, `operator:short`, in line with above.

### Possibly contentious

- Changing `name` to USPS.
    * Yes, the name would be better listed as "United States Post Office" (or left to the user, e.g. "Humboldt Park Post Office") but the `post_box` tagging is templated to the same tags and a post box should not be named "United States Post Office." 
    * I'd really argue that a post box shouldn't have a name (just an operator), but that's part of the issue.

### Stuff that `npm run build` makes that I don't particularly want.

- Do we need to apply a `name` tag? (See above.)
- `name:short`: Is this a thing? I thought we were supposed to use `short_name` for that, but if not, that's fine.


tas50:
I get the operator change here since any US Post Office is operated by USPS, but the name is really confusing in my mind. That doesn't seem to match how I'd think of the post office. I realize many of them are named with the local name, but again at that point it's XYZ Post Office and not XYZ Postal Service. US Postal Service is who runs it, not what it is. It's a US Post Office.

michaelblyons:
> I get the operator change here since any US Post Office is operated by USPS,

👍 

> but the name is really confusing in my mind. That doesn't seem to match how I'd think of the post office. I realize many of them are named with the local name, but again at that point it's XYZ Post Office and not XYZ Postal Service. US Postal Service is who runs it, not what it is. It's a US Post Office.

Agreed. My ideal would be to skip `name` altogether, and the editor him/herself can choose the name. But NSI forces us to pick a name. If "have to pick a name" was the only constraint, I'd be fully with you.

The problem is that post box suggestions (for those blue mail drop boxes) are built from the same exact tags. This means that if we choose the natural "United States Post Office" for post offices, iD will try to name post boxes "United States Post Office" as well. That is kind of obnoxious at best. If someone goes out to buy some stamps only to find a mail-drop, they will be annoyed.

I'm going to leave this PR open for comment for another week. If @tas50's view is consensus, I'll merge only the changes to `brand*` and `operator*` and open a new issue about the templating frustration.

bhousel:
> Agreed. My ideal would be to skip `name` altogether, and the editor him/herself can choose the name. But NSI forces us to pick a name. If "have to pick a name" was the only constraint, I'd be fully with you.

Oh we relaxed this requirement a few months ago... I agree that we can remove `name` from post offices (and _definitely_ from post boxes - if it's still there, i'm sorry, oversight on my part).

michaelblyons:
> Oh we relaxed this requirement a few months ago... I agree that we can remove `name` from post offices (and _definitely_ from post boxes - if it's still there, i'm sorry, oversight on my part).

I'm still getting a `name` re-added each time I `npm run build`. If that's just a bug, and `name` isn't supposed to be stamped in, I think we're all on the same page.

1ec5:
> I get the operator change here since any US Post Office is operated by USPS, but the name is really confusing in my mind. That doesn't seem to match how I'd think of the post office. I realize many of them are named with the local name, but again at that point it's XYZ Post Office and not XYZ Postal Service. US Postal Service is who runs it, not what it is. It's a US Post Office.

Agreed, a brand isn’t always named the same as the organization that runs the brand. (We don’t need to tag it, but U.S. Bank versus U.S. Bancorp is another prominent example.) There are probably more post office branches that bear a plain “United States Post Office” or “U.S. Post Office” sign than branches that use the more recent USPS logo.

> The problem is that post box suggestions (for those blue mail drop boxes) are built from the same exact tags. This means that if we choose the natural "United States Post Office" for post offices, iD will try to name post boxes "United States Post Office" as well. That is kind of obnoxious at best. If someone goes out to buy some stamps only to find a mail-drop, they will be annoyed.

If this is the main reason for changing the `name` and `brand`, I’d also suggest holding off on this change until NSI or iD can potentially distinguish the two cases. It would be surprising if we effectively started naming post offices after the boxes out front in OSM.

> `name:short`: Is this a thing? I thought we were supposed to use `short_name` for that, but if not, that's fine.

Interesting, the key would be `short_name` in OSM. Especially if we keep “United States Post Office” as the name, it won’t be necessary to specify `short_name`.

bhousel:
> If this is the main reason for changing the `name` and `brand`, I’d also suggest holding off on this change until NSI or iD can potentially distinguish the two cases. It would be surprising if we effectively started naming post offices after the boxes out front in OSM.

I'd actually like to get this change in now, release NSI v5, and then send a PR to iD soonish.  The issue of iD's validator overwriting `name` has been a problem for a while and we can finally close it out now that they aren't required in NSI anymore.

bhousel (approved):
merging as-is, will adjust in a followup commit

bhousel:
Relevant commits here:
2cc5cb93 Remove names from post offices, but preserve edge cases in `alt_name`
b53616ab No name tag on post boxes
4ce1be14 Drop *:short tags from USPS  (`short_name` remains, as that is the tag)

I think this covers it..

michaelblyons:
Thank you, all.

cicku:
Although it is up to mapper to give it a local name, however, a local name will bring confusion to other users when the office name differs from the area city name.

Columbus OH 

image


If you add an office with name `Northwest Post Office`, it is weird when any user from Washington DC to search for the same thing. A more extreme case is, `Northwest Post Office` of Washington DC is named in OSM but another one in Columbus OH is unamed.

image

If `City A Post Office` is located in City B, it brings more confusion. In a nutshell, the best is either unamed or `United States Post Office`.

1ec5:
> If you add an office with name Northwest Post Office, it is weird when any user from Washington DC to search for the same thing. A more extreme case is, Northwest Post Office of Washington DC is named in OSM but another one in Columbus OH is unamed.

A decent search engine would show the Northwest Post Office in Washington, D.C, not the one in Columbus, Ohio, when you search for “post offices in Washington, D.C.” Even if it shows both because you didn’t qualify your search, a typical search engine will display search results indicating the city and state of each result. If a user is confused that multiple neighborhoods in different parts of the United States have the same name – let alone the 29 populated places named “Springfield” – then I’m not sure how they’d be helped by giving all the post offices nationwide exactly the same name.

(Incidentally, the one in Washington, D.C., is called the “Northwest Station Post Office”. In addition to a postal city, USPS location names often indicate where they fall in the distribution network.)

In any case, the discussion here was about “United States Post Office” versus “United States Postal Service” as a default name that the mapper can refine with the benefit of local knowledge. If you disagree that this local knowledge should ever be given prominence in the `name` key for a post office, then that seems to me like a broader discussion not well suited to a PR merged long ago on this preset-focused repository.

PR #4789, opened 2020-12-09
Changed the Max Computation time in vscode

Asher-MS:
Changed the Timeout in milliseconds after which diff computation is cancelled. the default Vs Code Setting was 5000


bhousel (changes requested):
Looks like the json is missing a comma `,`

UKChris-osm:
I've added the missing comma, but don't use VS Code, is an extended timeout setting necessary?

bhousel:
I'm just going to close this.. It doesn't seem to do anything.

Issue #4779, opened 2020-12-06
Add universities?


kjonosm:
How about adding universities to the index? According to [OSM-Wiki ](https://wiki.openstreetmap.org/wiki/Tag:amenity=university) they should be listed in the operator tree.

kymckay:
I'm not sure that there would be enough applicable elements for any give university to warrant including them.

UKChris-osm:
This has come up before, and I think the idea hasn't been pushed forward because (as @kymckay mentioned above) so few Universities operate multiple campuses that it wouldn't be worth having in the index.

While many Campus locations would have multiple elements within it (buildings, parking, etc) all these would likely be enclosed in a boundary marking the university campus as a whole.

kjonosm:
I did some quick overpass turbo queries browsing through Europe and there are quite a few universities that do not reside on one or multiple campuses but are scattered across the city:
- [Cologne University](https://overpass-turbo.eu/s/10Ug)
- [Berlin Free University](https://overpass-turbo.eu/s/10Uh)
- [Berlin Technical University](https://overpass-turbo.eu/s/10Ui)
- [Humboldt University Berlin](https://overpass-turbo.eu/s/10Uj)
- [Université de Paris](https://overpass-turbo.eu/s/10Uk)
- [Université Sorbonne](https://overpass-turbo.eu/s/10Uu)
- [Universidad Politécnica de Madrid](https://overpass-turbo.eu/s/10Ul)
- [Universitat de Barcelona](https://overpass-turbo.eu/s/10Um)
- [Università degli Studi di Roma](https://overpass-turbo.eu/s/10Un)
- [Warsaw University](https://overpass-turbo.eu/s/10Uo)
- [Vienna University](https://overpass-turbo.eu/s/10Up)
- [Zurich University](https://overpass-turbo.eu/s/10Uq)
- [Coimbra University](https://overpass-turbo.eu/s/10Ut)

According to OSM wiki, most individual features (institutes, departments etc.) should not be tagged with `amenity=university`. However, I think NSI should be able to support that requirement.

bhousel:
That's an interesting idea - my initial thought was to say "no" but I clicked on a few of the overpass links, and these are actually campuses that have dozens or hundreds of `building=university` combined with `operator=whatever`. 

I agree that I could see it being useful for `building=university` and `amenity=parking`, and maybe more, but not for the `amenity=university` feature.

PR #4793, opened 2020-12-10
add_university

Kify7:
##Adding Mexican Culinary University🌮


bhousel:
Sorry @Kify7 , but we don't (currently!) add individual universities to this index - we only add the kind of for-profit "universities” that operate like a branded business in multiple locations, such as DeVry or University of Phoenix. I know this is an unusual concept outside of the USA - the “universities” in this index are really more like training centers than true universities.

This index is really for tracking the types of things where there are a _lot_ of them, not just single things.  We are considering adding some university operated assets in #4779

UKChris-osm:
Thanks for trying your hand at a pull request for the Name Suggestion Index (NSI) @Kify7 👍 

As @bhousel has mentioned above, we don't currently take a single or low volume locations (such as a University with only one or two locations) in the NSI, as the project is aimed at higher volume entries at the moment.

Your pull request was edited well however, so I wanted to say thank you for that and please do feel free to submit more 👍 but if you submit any more pull requests, I just wanted to give you a heads up about wikipedia, and the need for including the language of the article within the entry, which in this case, would have been "_es_":

**"brand:wikipedia": "Instituto Culinario de México",** (missing language)
 "brand:wikipedia": " **es:** Instituto Culinario de México", (shows Spanish language, bolded and spaced)
 "brand:wikipedia": "es:Instituto Culinario de México", (shows how it would look entered in the index)

Issue #4795, opened 2020-12-12
Add flags for French regions


daniel-callejas-sevilla:
I've prepared a JSON file containing information for the flags for the French regions. 

Since my system is not able to run the required npm version, I'm posting the JSON here for someone else to append into `data/flags/man_made/flagpole.json` (followed by `npm run build` and pull request of course).

[regions_françaises.prettyprint.zip](https://github.com/osmlab/name-suggestion-index/files/5683116/regions_francaises.prettyprint.zip)

UKChris-osm:
Hi @daniel-callejas-sevilla,

I see you have already forked this repository and created a "_.json_" file under the flags folder. If you copy that data into the "_flagpole.json_" file, maybe at the end, but still within the:

{
  "flags/man_made/flagpole": [
    ... existing flag data ...
    ... your new flag data ...
  }
}

... then you can create a pull request of your fork online (no need for npm) and if we can accept it, someone here could run npm run build after the merge.

PR #4796, opened 2020-12-12
Flags for French regions

daniel-callejas-sevilla:
As suggested in #4795, submitting this PR with field 'id' still missing as my local system does not support a recent version of npm. `npm run build` should be run after merge.

List of regions taken from this page: https://fr.wikipedia.org/wiki/R%C3%A9gion_fran%C3%A7aise


bhousel:
Thanks for adding these @daniel-callejas-sevilla, & welcome to the project 👋

PR #4807, opened 2020-12-19
Add Unit Test

wafuwafu13:
Hi :)
I introduced Unit Test.
I use [jest](https://jestjs.io/en/) and [testing-library/react-testing-library](https://github.com/testing-library/react-testing-library) and [testing-library/jest-dom](https://github.com/testing-library/jest-dom)
This test just show that render GitHubIcon correctly, but anyone can add tests from now on!
You can run test by `npm run test` at root.


app/src/Header.jsx
```diff
@@ -70,7 +70,7 @@ function Title(props) {
 function TreeSwitcher(props) {
   const t = props.t;
   const others = ['brands', 'flags', 'operators', 'transit'].filter(d => d !== t);
-  const links = others.map(t => (<li><Link to={`index.html?t=${t}`}>{t}/</Link></li>));
+  const links = others.map(t => (<li key={t}><MemoryRouter><Link to={`index.html?t=${t}`}>{t}/</Link></MemoryRouter></li>));
```
wafuwafu13, line 73: I add [MemoryRouter](https://reactrouter.com/web/api/MemoryRouter) because test doesn't pass if not it.
`Invariant failed: You should not use <Link> outside a <Router>`

Identitaet:
While this is a bit outside of my skill set what would be the use of this ?
We already have a functtion to validate the data and building everything should also reaveal most errors ?

But on the other hand having more tests can't hurt

bhousel:
Thanks @wafuwafu13 but I don't want to add tests for the react nsi.guide site now.

https://nsi.guide is just a preview site, so not like anyone depends on it serving up content in a certain way. We make changes to it sometimes, and in fact have a bunch more issues planned for it.

PR #4814, opened 2020-12-25
Change Safeway to Safeway Fuel Station

cicku:
1. Avoid name confusion with Safeway supermarket
2. Official name used by https://local.fuel.safeway.com/

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -3342,7 +3342,7 @@
       }
     },
     {
-      "displayName": "Safeway",
+      "displayName": "Safeway Fuel Station",
       "id": "safeway-b3d110",
       "locationSet": {"include": ["001"]},
       "tags": {
@@ -3350,7 +3350,7 @@
         "brand": "Safeway",
         "brand:wikidata": "Q1508234",
         "brand:wikipedia": "en:Safeway Inc.",
-        "name": "Safeway"
+        "name": "Safeway Fuel Station"
       }
     },
     {


bhousel:
I'm ok with this, but I know a lot of people in OSM prefer for the `name` to be something that is written on the sign for the business.  Is this change in response to a discussion from somewhere?

1ec5:
> Avoid name confusion with Safeway supermarket

It’s very common for a gas station chain to be named the same as a supermarket or convenience store chain, but different kinds of POIs have different tags (`amenity=fuel` versus `shop=supermarket`), so that by itself wouldn’t be a reason to change the preset’s `name` tag.

Somehow I’ve never been to a Safeway with its own gas station before – is “Safeway Fuel Station” as a proper noun the most common way to refer to the gas station, such as in in-store promotions? It used to be that way with Kroger’s “Kroger Fuel Center” and Meijer’s “Meijer Gas Station”, but nowadays I think the usage is just “Kroger” and “Meijer gas station” (except in title case sometimes). Maybe the Safeway website just refers to “Safeway Fuel Station” because it would be confusing otherwise on the website. The surest way to know would be to look at a receipt.

bryceco:
Apple and Google Maps label them Safeway Fuel Station.

cicku:
I do not have a receipt right now (I drove there about 2 days ago). 

But my bank transaction shows the merchant name as SAFEWAY FUEL though.

Adamant36:
The sign at the one I go to says just "Safeway" on the building, but then it says "Safeway Gasoline" on the sign next to the street. I guess they also something called "Safeway Express" also (in the picture it's just "Express"). Given the variations it's probably better just to go with "Safeway" or we could create like 4 separate entries with the different names :shrug:


98fa4cd9be372baa945000fae22c74b1

SafewayFuelStation3-225x300

1ec5:
As far as I can tell, Safeway Express is a chain of standalone convenience stores (with gas stations) that are separate from supermarket locations. I think it would be reasonable to split out a separate brand and Wikidata item for Safeway Express. I did something similar recently for the [Speedway Express](https://www.wikidata.org/wiki/Q103980347) chain of mini convenience stores (with gas stations), though I couldn’t tell if there were enough locations to warrant an NSI entry in that case.

UKChris-osm:
> but then it says "_Safeway Gasoline_" on the sign next to the street

It looks like that sign refers to the Gasoline prices (3 in red) vs. the Diesel 2 prices (1 in green), rather than calling it a "_Safeway Gasoline_"?

Adamant36:
>It looks like that sign refers to the Gasoline prices (3 in red) vs. the Diesel 2 prices (1 in green), rather than calling it a "Safeway Gasoline"?

You could be right. Here's another example that clearly isn't. 

Screenshot 2020-12-27 135415

bhousel:
> Apple and Google Maps label them Safeway Fuel Station.

Cool, let's do this then.  It does make the data much easier for people to understand.

We also already have a few other items under amenity/fuel, like [Costco Gasoline](https://nsi.guide/index.html?t=brands&k=amenity&v=fuel#costcogasoline-73aa20), [Hy-Vee Gas](https://nsi.guide/index.html?t=brands&k=amenity&v=fuel#hyveegas-c5cf43),  [H-E-B Fuel](https://nsi.guide/index.html?t=brands&k=amenity&v=fuel#hebfuel-c5cf43), [UDF Fuel](https://nsi.guide/index.html?t=brands&k=amenity&v=fuel#udffuel-c5cf43), where mappers prefer to put something in the name to distinguish it from the store that it's next to.

PR #4818, opened 2020-12-30
french bus network and bus country

Olyon:


bhousel:
Looks good to me, thanks @Olyon 👍   happy to merge as-is.

BTW if you want to scope things to just Metropolitan France (the part in Europe), you can use the country code `FX` instead of `FR`

[View France (FR)](https://ideditor.codes/?code=Q142)
[View Metropolitan France (FX)](https://ideditor.codes/?code=Q212429)

Country Coder supports a few other regional polygons like this for other countries too, you can see them all on https://ideditor.codes

PR #4820, opened 2020-12-30
Change subject of Flag of Europe

Karthoo:
European Union -> Europe
The Flag of Europe is not only the symbol of the European Union. There is actually a separate Wikidata item for the flag of the EU (Q98055852)

### DIFF
--- a/data/flags/man_made/flagpole.json
+++ b/data/flags/man_made/flagpole.json
@@ -1419,8 +1419,8 @@
         "flag:type": "governmental",
         "flag:wikidata": "Q122482",
         "man_made": "flagpole",
-        "subject": "European Union",
-        "subject:wikidata": "Q458"
+        "subject": "Europe",
+        "subject:wikidata": "Q46"
       }
     },
     {


bhousel:
Thanks for the correction @Karthoo 👍

1ec5:
> The Flag of Europe is not only the symbol of the European Union.

True, but [Q122482](https://www.wikidata.org/wiki/Q122482) is about the European flag, which is also known as the flag of the European Union. [Q1123840](https://www.wikidata.org/wiki/Q1123840) represents the symbols of Europe writ large, and there can be another item for the symbols of the EU if that subset is notable.

> There is actually a separate Wikidata item for the flag of the EU (Q98055852)

This item exists only because the German Wikipedia created [a separate article](https://de.wikipedia.org/wiki/EU-Flagge) to discuss the European flag as it was adopted by the EU after the fact. In Wikidata, items about flags already distinguish between the [date of creation](https://www.wikidata.org/wiki/Property:P571) and the date of adoption (as [start date](https://www.wikidata.org/wiki/Property:P580) qualifiers on [represents](https://www.wikidata.org/wiki/Property:P1268) statements), so Wikidata has no intrinsic need for a separate item to represent the EU flag.

I’ve retagged [Q98055852](https://www.wikidata.org/wiki/Q98055852) as being a flag design synonymous with [Q122482](https://www.wikidata.org/wiki/Q122482). But this still leaves us with a bit of a quandry for the entry that was changed here: when a European flag flies outside a French embassy, for example, it almost certainly represents the European Union. But perhaps there are places that fly the European flag to represent Europe as a concept rather than the European Union organizationally. Would anyone be opposed to splitting this entry in two, one with `subject=European Union` and the other with `subject=Europe`? I think both should still link to Q122482, because Q98055852 is quite obscure.

bhousel:
> But perhaps there are places that fly the European flag to represent Europe as a concept rather than the European Union organizationally. Would anyone be opposed to splitting this entry in two, one with `subject=European Union` and the other with `subject=Europe`?

I'm somewhat opposed to generating several presets for the same flag differentiated only by the `subject` tag.  

We could also just remove the `subject` tag and 🇪🇺 flag mappers could enter it?  But asking a mapper to decide "does this flag represent Europe or the European Union" seems kind of nitpicky.  Europe (Q46) covers it all.

doublah:
I'd say the Flag represents the EU significantly more than just Europe, especially in places like the UK and Eastern Europe.

PR #4821, opened 2020-12-30
Add country codes for NL and DK + Add flag of Baden

Karthoo:
Add country codes for the Netherlands and Denmark + Add flag of Baden


bhousel:
Thanks 👍

PR #4831, opened 2021-01-02
Dunkin' Donuts -> Dunkin'

cicku:
Source: https://news.dunkindonuts.com/news/releases-20180925

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -1166,7 +1166,7 @@
       }
     },
     {
-      "displayName": "Dunkin' Donuts",
+      "displayName": "Dunkin'",
       "id": "dunkindonuts-70af5e",
       "locationSet": {
         "include": ["001"],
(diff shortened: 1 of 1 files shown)


UKChris-osm:
I think it would be a good idea to add the old name as a matchName considering so many are still tagged in that way. Would it be worth excluding Canada from the locationSet as well?

Also, with the rebrand, is the Spanish version still called "_Dunkin' Coffee_" or is this to now just "_Dunkin'_"??

cicku:
@UKChris-osm you are right, old tags should be upgraded when they are being edited in the future.

Dunkin’ should still use Dunkin’ Coffee in Spain, they are not changed.

bhousel:
I've been wondering about this too.  I know they "officially" rebranded, but I haven't seen the signage change except in a handful of places.  I'm ok with changing the `name`, but I think we should at least leave "Dunkin' Donuts" in the `alt_name` tag.

UKChris-osm:
Thanks @cicku, if you've added an "_alt_name_" then you don't need to include it as a "_matchName_" in this instance.

bhousel:
Thanks @cicku 👍 
Like @UKChris-osm said, we don't need to add names to `matchNames` that are already in a tag. 
I'll merge as-is and it will be removed next time I run `npm run build`.

PR #4835, opened 2021-01-03
Add Carroll Motor Fuels and Crown Central Petroleum

cicku:
One thing I'm not sure is Crown brand, it was acquired by Clark few years ago.

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -780,6 +780,16 @@
       }
     },
     {
+      "displayName": "Carroll Motor Fuels",
+      "locationSet": {"include": ["us"]},
+      "matchNames": ["carroll", "carrollfuel"],
+      "tags": {
+        "amenity": "fuel",
+        "brand": "Carroll",
+        "name": "Carroll Motor Fuels"
+      }
+    },
+    {
       "displayName": "Casey's General Store",
       "id": "caseysgeneralstore-b3d110",
       "locationSet": {"include": ["001"]},
@@ -1077,6 +1087,20 @@
         "brand": "Crodux",
         "brand:wikidata": "Q62274622",
         "name": "Crodux"
+      }
+    },
+    {
+      "displayName": "Crown",
+      "locationSet": {
+        "include": ["us"]
+      },
+      "matchNames": ["crown"],
+      "tags": {
+        "amenity": "fuel",
+        "brand": "Clark",
+        "brand:wikidata": "Q5189323",
+        "brand:wikipedia": "en:Crown_Central_Petroleum",
+        "name": "Crown"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


data/brands/amenity/fuel.json
```diff
@@ -1079,6 +1089,20 @@
         "name": "Crodux"
       }
     },
+    {
+      "displayName": "Crown",
+      "locationSet": {
+        "include": ["us"]
+      },
+      "matchNames": ["crown"],
```
UKChris-osm, line 1097: You won't need a "_matchName_" here as it's the same name as in the "_displayName_" and the "_build_script.js_" can handle names that are the same but different case 👍

cicku: Just learned, I will fix that (mobile now).

bhousel: It actually won't match the `displayName` but it will match the `name` tag, so yes "crown" is redundant in this case.

data/brands/amenity/fuel.json
```diff
@@ -1079,6 +1089,20 @@
...
+      "tags": {
+        "amenity": "fuel",
+        "brand": "Clark",
```
bhousel, line 1100: A bigger issue is it shouldn't say "Clark" here, but rather something that matches where the wikidata/wikipedia tags point to, so "Crown Central Petroleum" or "Crown" probably.

bhousel (changes requested):

data/brands/amenity/fuel.json
UKChris-osm: Ops, my mistake.

data/brands/amenity/fuel.json
UKChris-osm: It seems like Clark and Crown are almost one brand now "Clark / Crown", as the Wikipedia article has a "_merge_" tag, and the [Facebook logo](https://nsi.guide/index.html?t=brands&k=amenity&v=fuel&tt=clark) includes both the Crown and Clark Logo, so maybe in this case both "_Clark_" and "_Crown_" named entries should link to the same Wiki*?

Or, perhaps both should have unique wikidata "brand" entries that are separate from "_Clark Brands_"?

cicku: Wikipedia does not reflect reality. If these 2 wiki pages merge into one in the future, I can send another PR. Crown is not a national chain so it is very easy to upgrade tags again. However, I think redirection inside mediawiki will cover this well.

You still cannot spot any “Clark” at any Crown stations in MD, even they have already merged since 2013.

bhousel (approved):

bhousel:
Thanks @cicku this looks good 👍 
We've seen a few other similar examples in the fuel world, where one company acquires another but they still operate stores with the old branding. #3154 has a few others.

PR #4836, opened 2021-01-03
Add China Telecom (Update telecommunication.json)

LaoshuBaby:


LaoshuBaby:
I don't know how to ```npm run build``` because it give me a lot of ERROR

and it change dist json but in README it tell me not to touch that

So I only Update this ```telecommunication.json```

______________________________________

Here is the ERROR Screenshot

图片

Here is the Github Desktop tell me

图片

______________________________________

So, What should I do?

LaoshuBaby:
Sorry
It's my fault
I should use ```cn``` and ```tw``` to select those two regions, not ```zh```(I guess it means mainland, but in fact it is undefined)

PR #4837, opened 2021-01-03
Add China Telecom into telecommunication.json

LaoshuBaby:


UKChris-osm:
Thank you @LaoshuBaby 

Do you know if China Telecom have any social media links that can be added to the Wikidata article, such as Facebook or Twitter? We can use those logo's to add to the index if they are available 👍

LaoshuBaby:
> 
> 
> Thank you @LaoshuBaby
> 
> Do you know if China Telecom have any social media links that can be added to the Wikidata article, such as Facebook or Twitter? We can use those logo's to add to the index if they are available 👍

If I can I will really willing to add more information into this brand.

But the fact is that usually Chinese company won't set up Facebook or Twitter account. And I found that there are a lot of fake account of some famous Chinese company.

So if CT is OK, I will add more chain-brand which very-famous into NSI list.

Thank you for your review!

LaoshuBaby:
> 
> 
> > Thank you @LaoshuBaby
> > Do you know if China Telecom have any social media links that can be added to the Wikidata article, such as Facebook or Twitter? We can use those logo's to add to the index if they are available 👍
> 
> If I can I will really willing to add more information into this brand.
> 
> But the fact is that usually Chinese company won't set up Facebook or Twitter account. And I found that there are a lot of fake account of some famous Chinese company.
> 
> So if CT is OK, I will add more chain-brand which very-famous into NSI list.
> 
> Thank you for your review!

I will choose the brand that have Wikidata and have nation-wide chain-store.

UKChris-osm:
Thanks for the feedback about Facebook and Twitter in China @LaoshuBaby 👍

PR #4839, opened 2021-01-03
Update Hilton Garden Inn to note that it's globally available

desiderantes:


UKChris-osm:
Thank you @desiderantes 👍 

However, looking at their [website](https://www.hilton.com/en/locations/hilton-garden-inn/), I'm thinking maybe it would be better to mark Hilton Garden Inn as "_001_" as they seem to be in many places?

desiderantes:
Thanks @UKChris-osm I updated the PR with your suggested change

UKChris-osm:
Thank you for the edit @desiderantes and for highlighting that "_Hilton Garden Inn_" only had a small locationSet currently available to it 🥇

PR #4840, opened 2021-01-04
Added two brands seen in France - Saga Cosmetics - RougeGorge

nuxper:
This is my first pull request to name-suggestion-index so do not hesitate to suggest any improvement.
I started with just two brands and will add others if the pull request is accepted.
This one integrates two brands present in France : RougeGorge which is a brand of underwear shops and Saga Cosmetics which sells cosmetics and perfumes.

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -4393,6 +4393,18 @@
       }
     },
     {
+      "displayName": "RougeGorge",
+      "id": "rougegorge-4a3e29",
+      "locationSet": {"include": ["fr"]},
+      "tags": {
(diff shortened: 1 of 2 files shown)


UKChris-osm:
Thank you for your first PR @nuxper 

It looks like "_RougeGorge_" is also located in Belgium, are you able to confirm this and add the location to "_locationSet_"?

nuxper:
You are perfectly right, thank you. I added be in locationSet and checked that saga cosmetics is only in France for the moment.

UKChris-osm:
Thank you @nuxper, your first PR has been merged 🥇

PR #4844, opened 2021-01-04
Add Tamrex in Latvia

richlv:


UKChris-osm:
Thank you @richlv 👍 

Are these shops also available in Estonia?

richlv:
They are, but I was not fully sure about their classification (saw one mapped as shop=security).
I'll ping the Estonian community on this :)

PR #4845, opened 2021-01-04
For Benu, also match "benu aptieka"

richlv:
That's the same in Latvia as "benu vaistinė" in Lithuania.


UKChris-osm:
Thanks @richlv 👍

PR #4846, opened 2021-01-05
added aditional shops present in France

nuxper:
I added more brands observed in France, with wikidata tag filled in.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -186,6 +186,19 @@
       }
     },
     {
+      "displayName": "Bagel Corner",
+      "id": "bagelcorner-e4dee6",
+      "locationSet": {"include": ["fr"]},
+      "tags": {
+        "amenity": "fast_food",
+        "brand": "Bagel Corner",
+        "brand:wikidata": "Q104649272",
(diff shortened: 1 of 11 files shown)


data/brands/amenity/fast_food.json
```diff
@@ -185,6 +185,19 @@
         "takeaway": "yes"
       }
     },
+    {
+      "displayName": "Bagel Corner",
```
Identitaet, line 189: Isn't that more of a Bakery ?
The results from overpass show that it's splti 
Rest looks good 👍

nuxper: Hi @Identitaet .
I've checked : the few nodes that have shop=bakery also have amenity=fast_food. A company (Digitalbricks) has added both tags together.
Bagel Corner works a bit like subway but with a bagel instead of bread.
They don't sell bread. which at least in France definitively exclude them from the bakery notation ;)
I'll send a message to Digitalbricks but I believe it's safe to integrate them as fast_food in name-suggestion-index

Identitaet: okay if they don't sell bread I agree that's quite clear

Issue #4105, opened 2020-08-29
Add supermarket cackback policies?


Cj-Malone:
What's the current opinion on adding extra meta data like cashback? I know there was a previous pull request (#3048) that was rejected. However recently you have taken on tags like internet access, they seem similar on the face of it. Extra meta data in OSM is a good thing, and NSI seems to be a good place to recommend it.

bhousel:
I think just about any business allows you to take cash back when you use a debit card to pay.  I don't see how this is a brand-specific thing?  I'd really prefer not to add so many tags that describes almost every business out there.  (it would be like adding `lit=yes` to all the stores - it may be true but nobody needs this.)

I'm also iffy on the value of adding the wifi tags, but I guess some brands use the same wifi in all their locations.

Also you call them "meta tags" - why?  Meta tags would be like, "data about the data" - like "date added" or "source".  I think there is probably a better word to describe them, but I don't know what it would be.

Cj-Malone:
> I think just about any business allows you to take cash back when you use a debit card to pay.

I don't think you can generalise all the way to that, if you said most **supermarkets** in the UK(/EU/US/"Western world"), I'd agree, but "all" is quite inaccurate for a global project.

> I don't see how this is a brand-specific thing?

Certain chains/brands do it, certain ones don't. It's specific to the chain, not any individual location.

I guess you could argue for defaults, like all supermarkets in the UK offer cashback by default unless `cash_withdrawal=no`, but the OSM ecosystem doesn't have a standard place for data like that, nsi is the closest thing.

tas50:
In the US you'd be really hard pressed to find a Supermarket in the index that didn't allow cash back. Some of them like certain Kroger brands may charge for that cash back, but it's assumed that debit card transaction comes with cash back here.

Cj-Malone:
There are other tags related to cashback like fee and limit, not just the overall yes or no. See [the wiki](https://wiki.openstreetmap.org/wiki/Key:cash_withdrawal).

lakedistrictOSM:
> Certain chains/brands do it, certain ones don't. It's specific to the chain, not any individual location.

Not necessarily. Some branches of Tesco Express offer cashback and some branches don't.

Cj-Malone:
@lakedistrictOSM that's what Tesco said when I contacted them, so obviously it shouldn't be in the Tesco entries. On the other hand Heron Foods said none of there stores offer cash back.

bhousel:
[...]

PR #4855, opened 2021-01-10
Add diet:vegetarian to some fast food brands.

Eric-Sparks:
I've been adding the diet:vegetarian tag to some local restaurants but, according to the brand's website, these places universally have vegetarian options.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -30,7 +30,7 @@
     },
     {
       "displayName": "A&W (USA)",
-      "id": "aandw-4d2ff4",Tr
+      "id": "aandw-4d2ff4",
       "locationSet": {"include": ["us"]},
       "tags": {
         "amenity": "fast_food",
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -4409,7 +4409,7 @@
         "brand:wikidata": "Q752941",
         "brand:wikipedia": "en:Taco Bell",
         "cuisine": "tex-mex",
-        "diet:vegetarian": "yes",
+        "diet:vegan": "yes",
         "name": "Taco Bell",
         "takeaway": "yes"
       }
@@ -4993,7 +4993,7 @@
         "brand:wikidata": "Q1244034",
         "brand:wikipedia": "en:White Castle (restaurant)",
         "cuisine": "burger",
-        "diet:vegetarian": "yes",
+        "diet:vegan": "yes",
         "name": "White Castle",
         "takeaway": "yes"
       }
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -796,7 +796,7 @@
         "brand:wikidata": "Q465751",
         "brand:wikipedia": "en:Chipotle Mexican Grill",
         "cuisine": "mexican",
-        "diet:vegetarian": "yes"
+        "diet:vegetarian": "yes",
         "name": "Chipotle",
         "official_name": "Chipotle Mexican Grill",
         "takeaway": "yes"
@@ -1519,7 +1519,7 @@
         "brand:wikidata": "Q1131810",
         "brand:wikipedia": "en:Five Guys",
         "cuisine": "burger",
-        "diet:vegetarian": "yes"
+        "diet:vegetarian": "yes",
         "name": "Five Guys",
         "official_name": "Five Guys Burgers and Fries",
         "takeaway": "yes"
@@ -4309,7 +4309,7 @@
         "brand:wikidata": "Q244457",
         "brand:wikipedia": "en:Subway (restaurant)",
         "cuisine": "sandwich",
-        "diet:vegetarian": "yes"
+        "diet:vegetarian": "yes",
         "name": "Subway",
         "takeaway": "yes"
       }
--- a/data/brands/shop/furniture.json
+++ b/data/brands/shop/furniture.json
@@ -97,6 +97,15 @@
       }
     },
     {
+      "displayName": "Bedmart",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "brand": "Bedmart",
+        "name": "Bedmart",
+        "shop": "furniture"
+      }
+    },
+    {
       "displayName": "Bellona",
       "id": "bellona-12c613",
       "locationSet": {"include": ["001"]},


UKChris-osm:
Thanks @Eric-Sparks 

There seems to be a random "_Tr_" floating around within the "_A&W (USA)_" entry, any reason for this?

Identitaet:
Overall a good idea.

I think there's one problem here which is that many of these brands have different menus world wide.

For example while lots of restaurants have lots of vegetarian options for the Indian market the German market isn't as likely to have vegetarian options.
So that's something we have to keep in mind but in general a really good idea 👍

Eric-Sparks:
> Thanks @Eric-Sparks
> 
> There seems to be a random "_Tr_" floating around within the "_A&W (USA)_" entry, which is causing a build fail, any reason for this?

Umm, no?  Didn't think I changed anything with A&W.

bhousel:
I feel pretty weird about this - 
Does [`diet:vegetarian=yes`](https://wiki.openstreetmap.org/wiki/Key:diet:vegetarian) really just mean "vegetarian options are available"?

That would cover almost all restaurants and fast food, right?  
I could see tagging exceptions like `only` or `no` as interesting, but `yes` should maybe be assumed?
It reminds me of #4105/#3048 (cashback) and I really dont want to upset OSM.

data/brands/amenity/fast_food.json
```diff
@@ -795,6 +796,7 @@
         "brand:wikidata": "Q465751",
         "brand:wikipedia": "en:Chipotle Mexican Grill",
         "cuisine": "mexican",
+        "diet:vegetarian": "yes",
```
akadouri, line 799:
```suggestion
        "diet:vegan": "yes",
```
Chipotle's sofritas is vegan as well as no-protein options. This would encompass vegetarian.

data/brands/amenity/fast_food.json
```diff
@@ -4404,6 +4409,7 @@
         "brand:wikidata": "Q752941",
         "brand:wikipedia": "en:Taco Bell",
         "cuisine": "tex-mex",
+        "diet:vegetarian": "yes",
```
akadouri, line 4412:
```suggestion
        "diet:vegan": "yes",
```
Taco Bell also has an [offical path](https://www.tacobell.com/blog/how-to-eat-vegan) to modifying menu options to make them vegan.

data/brands/amenity/fast_food.json
```diff
@@ -4986,6 +4993,7 @@
         "brand:wikidata": "Q1244034",
         "brand:wikipedia": "en:White Castle (restaurant)",
         "cuisine": "burger",
+        "diet:vegetarian": "yes",
```
akadouri, line 4996:
```suggestion
        "diet:vegan": "yes",
```
White Castle has an impossible slider which is vegan.

akadouri (commented):
Made some suggestions to change `diet:vegetarian` to `diet:vegan` based on chains I know have vegan friendly options or quick substitutions. 

As to whether this is worth it, I think so. It's definitely subjective as to whether the tag is warranted, I wouldn't add it for a place that just had french fries but I would for a burger restaurant that offers a meat-free patty. I usually stick to vegan tags since vegan-friendly menus are a bit less ubiquitous than vegetarian ones and that tag encompasses vegetarian. It helps power [OpenVegeMap](https://openvegemap.netlib.re/#zoom=15&lat=40.74429&lon=-73.9852) which is an open alternative to things like HappyCow and until recently Google Maps/Yelp/etc didn't have reliable searching options for "Vegan". There's an active telegram where people talk about this as well.

data/brands/amenity/fast_food.json
Eric-Sparks: I wouldn't replace vegetarian with vegan due to the way certain searches may work but I'd certainly add it.

Eric-Sparks (commented):
Ugh, okay, I clearly do better when just using git in a command-line environment...

So, I wouldn't change vegetarian to vegan due to the way some searches may work *but* I'd totally add diet:vegan=yes to those chains that support vegan.

bhousel:
Still feeling pretty unsure about this PR.  For example this change:


Screen Shot 2021-01-16 at 12 28 22 PM

This means:
* Every White Castle in the USA gets the `diet:vegan=yes` tag.
* Users of iD would be prompted to "upgrade" all the White Castles to have this tag.

This might be _technically_ correct - I don't know - but it would surely be confusing and suspicious for users.  If iD prompted me to make this change, I'd think something was wrong.

A good way to think about the name-suggestion-index is that we shouldn't add any tags to this index that we wouldn't also be comfortable making to the OSM with a mass edit.  I'm very ok with updating tags like `:wikidata` and `cuisine` but less ok about tags like `diet:` or `payment:`.

akadouri:
That makes sense, rethinking about it as a mass edit I don't know if I'd still make the change unless it was a core part of the chain (the `diet:x=only` case). Might be something to track on the wikidata page instead and check for that way.

Eric-Sparks:
I felt this was important because it is sometimes difficult to find vegetarian options in the sea of restaurants that are out there. Not all restaurants offer good vegetarian options. The restaurants I selected specifically say they offer vegetarian options *as a brand*. I think this is a good opportunity to start adding these dietary options to the map.

bhousel:
@Eric-Sparks - I've been thinking about this more.. What if we added another property like `"impliedTags": { … }` where people could put tags like this?  They'd still be attached to the brand, but wouldn't be added to OSM automatically.

I agree with you that it's useful information to track somewhere, my biggest concern is just that adding them all en masse to OSM would be seen badly by some mappers.  

A project that wants to know which brands have vegan offering could still use this index to say "If I match White Castle, I can imply that they offer vegan food".   Or you could make/use/release your own distribution of OSM that has the implied tags already merged in?

Identitaet:
> 
> 
> @Eric-Sparks - I've been thinking about this more.. What if we added another property like `"impliedTags": { … }` where people could put tags like this? They'd still be attached to the brand, but wouldn't be added to OSM automatically.

that might be quite usefull but in my opinion more for values which conflict with that implied tag
so if a place wasn't offering vegetarian food and now is
> 
> I agree with you that it's useful information to track somewhere, my biggest concern is just that adding them all en masse to OSM would be seen badly by some mappers.
> 
Yup probably not a good idea

> A project that wants to know which brands have vegan offering could still use this index to say "If I match White Castle, I can imply that they offer vegan food". Or you could make/use/release your own distribution of OSM that has the implied tags already merged in?

Do you mean that as a general value for each catergory or for each entry on it's own ?

Eric-Sparks:
> @Eric-Sparks - I've been thinking about this more.. What if we added another property like `"impliedTags": { … }` where people could put tags like this? They'd still be attached to the brand, but wouldn't be added to OSM automatically.

I'm not sure why you wouldn't want to add this information automatically?

> I agree with you that it's useful information to track somewhere, my biggest concern is just that adding them all en masse to OSM would be seen badly by some mappers.

Why would this be seen as bad by some mappers?

What am I missing here?  Adding [diet:vegetarian](https://wiki.openstreetmap.org/wiki/Key:diet:vegetarian) to a restaurant is not new to OSM and has been used [more than 30,000 times](https://taginfo.openstreetmap.org/keys/?key=diet%3Avegetarian) around [the world](https://taginfo.openstreetmap.org/keys/?key=diet%3Avegetarian#map).  If a brand is openly advertising vegetarian options then it would seem like an easy thing to do to add this to their brand which then helps the [millions of people](https://en.wikipedia.org/wiki/Vegetarianism_by_country) that are looking for this information on the streets.  Clearly there is a problem here that I'm not understanding as this seems to be rather straight forward.

UKChris-osm:
> If a brand is openly advertising vegetarian options then it would seem like an easy thing to do to add this to their brand.

Generally I would agree with you as it relates to this, but there are times when a brand will not necessarily have an entire menu available at every restaurant, which is sometimes marked as "_at participating restaurants_" or in White Castle's case "_at participating Castles_".

That being said, I wouldn't have any objection to the NSI listing "_diet:*_" based on a brands menu, as I think that is a general indication that it's likely to be available, and I can't imagine any restaurant or franchisee would remove that type of menu selection in full from the brand menu.

ArloJamesBarnes:
I'd imagine a `source` tag could go a long way towards dispelling myths about iD meddling, since then the discussion becomes "well, argue with [restaurant brand] about whether they're offering what they say they're offering". This could be stored in each index entry, and seems easier to interpret for downstreams than an `impliedTags` as it would just be a link to the menu or part of the menu that features a full veg[eteri]an meal.

UKChris-osm:
I've noticed that #5006 was merged with **"diet:vegan": "only",** included, so would that have any bearing on this PR being accepted, or was #5006 merged incorrectly?

bhousel:
> I've noticed that #5006 was merged with **"diet:vegan": "only",** included, so would that have any bearing on this PR being accepted, or was #5006 merged incorrectly?

#5006 is fine - there's a big difference between adding `diet:vegan=only` to a vegan restaurant and adding `diet:vegetarian=yes` to Burger King like this PR does.

LaoshuBaby:
Is it necessary to keep this Pull Requests still open?

This pr has already been pending for 8 months, and still have something controversial such as adding ```diet:vegetarian=yes``` for even Burger King or Subway, Taco Bell

Eric-Sparks:
Well, this really shouldn't have been controversial.  These restaurants are advertising having vegetarian options (not necessarily vegan, by the way, as this PR morphed into).  I'm just trying to make it easier for vegetarians to find appropriate places to eat.

bhousel:
> Is it necessary to keep this Pull Requests still open?
> 
> This pr has already been pending for 8 months, and still have something controversial such as adding `diet:vegetarian=yes` for even Burger King or Subway, Taco Bell

I'm ok with this PR staying open because I think it's a reasonable request, but I think tags like this should be in an `impliedTags` structure (tags that aren't applied automatically to all the matched features in OSM), and I haven't build that yet.  [We get criticism about setting `brand:wikipedia` tags](https://github.com/osmlab/name-suggestion-index/issues/5275#issuecomment-905798587), so having a separate `impliedTags` object could also solve that problem.

bhousel:
> Well, this really shouldn't have been controversial.  These restaurants are advertising having vegetarian options (not necessarily vegan, by the way, as this PR morphed into). 

I agree with you, but if you were to ask any of the regular OSM channels (tagging mailing list, slack, discord, etc) and propose a mass edit to `diet:vegan=yes` to all the White Castles in the world, I feel like people would tell you no.  I've seen way less controversial suggestions get rejected or reverted.

LaoshuBaby:
> having a separate `impliedTags` object could also solve that problem.

That will be a better way.
We will waiting for that so let this keep open. Thank you!

bhousel:
> Sort of linked to this, and also `takeaway=yes` tagging, how do we feel about adding `food=yes` ( https://wiki.openstreetmap.org/wiki/Key:food ) on e.g. `amenity=pub` ( https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dpub ) like say Wetherspoons? Which can then trigger things like StreetComplete to ask if they have various diets available at each location?

I kind of disagree with this.. `food=yes` seems like a really unhelpful tag, and could be applied to almost anything.

`amenity=pub` should be tag that triggers StreetComplete to ask the "is there food here" question.

(as an aside, I sort of think the `takeaway=yes` tag is silly too, and we probably shouldn't have started doing it)

peternewman:
> I kind of disagree with this.. `food=yes` seems like a really unhelpful tag, and could be applied to almost anything.
> 
> `amenity=pub` should be tag that triggers StreetComplete to ask the "is there food here" question.

Yeah, we sort of got caught up with what is food, see https://github.com/streetcomplete/StreetComplete/pull/3099 and https://lists.openstreetmap.org/pipermail/tagging/2021-July/062106.html onwards...

So tagging food=yes where we know there definitely would be would work around it for some places, without having to answer the awkward question?

See also https://en.wikipedia.org/wiki/Substantial_meal !

> (as an aside, I sort of think the `takeaway=yes` tag is silly too, and we probably shouldn't have started doing it)

How so? A posh afternoon tea place is going to be `takeaway=no` or maybe `takeaway=not_unless_you_bring_your_own_reusable_cup` if it serves everything in the finest china, so therefore `takeaway=yes` is useful info...

Eric-Sparks:
Can we please get back on track with the original intent of this pull request?  There are people out there that have dietary restrictions that are looking for ways to find food easily.  When a brand openly advertises that they offer a vegetarian, or whatever, option, I think we should add that to the brand to help those with dietary restrictions find food.  These places are mostly cookie-cutter where each place is going to offer the exact same food so there is no need to have each restaurant surveyed to determine if they serve the dietary option or not.

There has been a lot of speculation as to what "people" might say if you asked on a particular communications medium.  I say that if there are people against such a proposal then they aren't thinking about the users out there that would actually use the data to obtain, you know, FOOD.  Of course, no one has actually asked these "people", not that it really matters.  This really isn't a huge deal and I'm not sure why people are making more of this than it has to be.

Georift:
Hey I was linked across to this PR as I was making a change to a local brand who when checking all regional menus would have vegetarian and vegan options.

I totally understand the point you're making here @bhousel, it's a fine line to make sure people don't discount this index. For the implied tags feature, do you imagine there will be any integration with editors to help recommend the tag be applied? Perhaps a two step approach, first apply the tags, then offer some additional tags? I understand you can't speak for every editor but just your best guess.

At least for me, the only reason it would be interesting adding these tags to this index, would be to increase the usage of these tags across the map. Where this would help [OpenVegeMap](https://openvegemap.netlib.re/) really flesh out it's data without needing significant investment or understanding by the editor about the nuance of this tag and our discussion right now.

Excited to hear your thoughts :slightly_smiling_face:

Adamant36:
I'm closing this as stale and probably not mergeable with how things currently are anyway. To quote a comment from another issue, "f it's an "attribute tag" that is true for every instance of this thing in OSM, there isn't a lot of value in actually adding it to every instance of the thing in OSM, and it just makes extra work for mappers if that attribute ever changes." Feel free to re-open this if anything changes about this in the future though 👍

PR #4856, opened 2021-01-12
Update Kwik Fill

cicku:
- Wikidata
- Wikipedia

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -2184,6 +2184,8 @@
       "tags": {
         "amenity": "fuel",
         "brand": "Kwik Fill",
+        "brand:wikidata": "Q7889091",
+        "brand:wikipedia": "en:Kwik Fill",
         "name": "Kwik Fill"
       }
     },


UKChris-osm:
Thanks @cicku 

However, it might be preferable to create a specific "_Kwik Fill_" branded Wikidata item, rather than using the parent company / operator wikidata in this case.

For example, having a Wikidata item that's specific to "_Kwik Fill_" it's web site, social media links, inception, headquater location, and the fact it's a subsidiary of the "_United Refining Company_".

bhousel:
Agree with @UKChris-osm - it would be better to have wikipedia/wikidata pages for the Kwik Fill brand.

cicku:
Sorry I was busy recently.

Yes your ideas area great. So what should I do now? Simply creating a new page at wikipedia with a redirection to parent company?

UKChris-osm:
I would recommend creating a Wikidata entry for "_Kwik Fill_" (I don't think there is enough information to create a Wikipedia entry) and add a reference to its parent company "_United Refining Company_", as well as any other useful information, such as social media links, the date the company was created, web address, headquarter locations, etc.

The "_Kwik Fill_" Wikipedia reference you have included in your PR is only a redirect to "_United Refining Company_" and so isn't suitable unfortunately.

cicku:
I'm going to close this one and work on this in the future.

PR #4859, opened 2021-01-13
Add Metro Bike Share

quincylvania:
Bikesharing in LA: https://en.wikipedia.org/wiki/Metro_Bike_Share

### DIFF
--- a/data/brands/amenity/bicycle_rental.json
+++ b/data/brands/amenity/bicycle_rental.json
@@ -2137,6 +2137,22 @@
       }
     },
     {
+      "displayName": "Metro Bike Share",
+      "id": "metrobikeshare-97ea79",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Thank you @quincylvania 👍 

Do you think ["locationSet": {"include":[[-118.33,34.05]]},](https://ideditor.github.io/location-conflation/?locationSet=%7B%22include%22%3A%5B%5B-118.33%2C34.05%5D%5D%7D) would cover the bike rental area sufficiently?

quincylvania:
> Do you think "locationSet": {"include":[[-118.33,34.05]]}, would cover the bike rental area sufficiently?

Seems like it! Unless they expand significantly. https://bikeshare.metro.net/stations/

Issue #4857, opened 2021-01-12
Brazil Brands


Playzinho:
name=Magazine Luiza
shop=department_store
wikidata=Q6729763
wikipedia=pt:Magazine Luiza
website=https://www.magazineluiza.com.br/

Playzinho:
name=Armazém Paraíba
shop=department_store
wikidata=Q9629505
wikipedia=pt:Armazém Paraíba
website=https://armazemparaiba.com.br/

Playzinho:
name=SAMU 192
emergency=ambulance_station
wikidata=Q104713679
wikipedia=pt:Serviço de Atendimento Móvel de Urgência
phone=192

Playzinho:
name=Banco24Horas
phone=+55 800 56 2400
website=https://www.banco24horas.com.br
wikidata=Q4854069
wikipedia=pt:Banco24Horas
network=Banco24Horas

Playzinho:
name=Crefisa
phone=+55 86 3221 2476
shop=money_lender
wikidata=Q10261396
wikipedia=pt:Crefisa
website=https://www.crefisa.com.br/

Heikrana:
Can I give it a go?

UKChris-osm:
@Heikrana yes of course you may give any of the above a go 👍

Playzinho:
of course, thanks don't close the issue, I intend to add more brands as soon as possible

Heikrana:
Update: I have added the above mention entries in their respective JSON files. Only problem is I don't understand where to put the website entry? I have added website in "brand:website" under "tags" like this...
![entry](https://user-images.githubusercontent.com/49754221/104811181-cba11b00-57f1-11eb-9cc0-ab4596c3d234.png)
Is it correct or should I change it? What am I missing because I have read contributing.md 3 times and checked nsi.guide also (there, website is listed but I can't find it in the json file which I'm editing).

UKChris-osm:
Hi @Heikrana, we currently don't keep an entry for "_brand:website_" in this index, so you can safely remove it 👍 

The web site listings on [nsi.guide](https://nsi.guide) are taken from the Wikidata entries for each item.

Heikrana:
Thanks a lot @UKChris-osm, you've been of great help. This is my first contribution in a substantially big project, albeit small but it's a start. Thanks for clearing my doubts, I really appreciate it. 💯

UKChris-osm:
[...]

PR #4860, opened 2021-01-16
Fix issue #4857 - Add new entries

Heikrana:
Add new entries in various files in "data" folder.

### DIFF
--- a/data/brands/shop/money_lender.json
+++ b/data/brands/shop/money_lender.json
@@ -122,7 +122,6 @@
         "brand": "Crefisa",
         "brand:wikidata": "Q10261396",
         "brand:wikipedia": "pt:Crefisa",
-        "phone": "+55 86 3221 2476",
         "shop": "money_lender"
       }
     },
--- a/data/brands/shop/money_lender.json
+++ b/data/brands/shop/money_lender.json
@@ -120,7 +120,6 @@
       "locationSet": {"include": ["br"]},
       "tags": {
         "brand": "Crefisa",
-        "brand:website": "https://www.crefisa.com.br/",
         "brand:wikidata": "Q10261396",
         "brand:wikipedia": "pt:Crefisa",
         "phone": "+55 86 3221 2476",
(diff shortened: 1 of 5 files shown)


UKChris-osm:
Thanks @Heikrana 

Just out of interest, what are the edits in "_docs/src.6ebfba02.js _" as this file powers the NSI web site, but the file itself is built via the /app/ folder?

data/brands/shop/money_lender.json
```diff
@@ -114,6 +114,19 @@
...
+    {
+      "displayName": "Crefisa",
+      "id": "crefisa-46cd18",
+      "locationSet": {"include": ["br"]},
+      "tags": {
+        "brand": "Crefisa",
+        "brand:website": "https://www.crefisa.com.br/",
+        "brand:wikidata": "Q10261396",
+        "brand:wikipedia": "pt:Crefisa",
+        "phone": "+55 86 3221 2476",
```
UKChris-osm, line 125: Just an FYI, we also tend not to include contact details in this project, so you can also remove the phone number 👍

Heikrana:
> Thanks @Heikrana
> 
> Just out of interest, what are the edits in "_docs/src.6ebfba02.js _" as this file powers the NSI web site, but the file itself is built via the /app/ folder?

It was automatically created after "npm run build", the first time I compiled the project.

data/brands/shop/money_lender.json
Heikrana: Okay, I have remove entry "phone" from the file "money_lender.json".

UKChris-osm:
In that case I better wait for someone else (such as @bhousel) to check that out, as I'm not sure what "_npm run build_" has done to that file.

Thanks for this PR though @Heikrana 👍

bhousel:
Thanks @Heikrana & @UKChris-osm  this looks ok..
It looks like some slightly different versions of dependencies, so I'll update everything to the latest and run the build again.

Also, I'm going to remove the `operator:phone=192`, that's another tag that we don't really add to this index.

PR #4861, opened 2021-01-17
Update Taiwan bank information

Supaplextw:
add name:zh, name:nan, name:hak, redirect Sinpac Bank Wikidata link to the bank branch Wikidata item

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -10414,7 +10414,12 @@
         "brand:wikidata": "Q4207208",
         "brand:wikipedia": "en:Shanghai Commercial and Savings Bank",
         "name": "上海商業儲蓄銀行",
-        "name:en": "Shanghai Commercial and Savings Bank"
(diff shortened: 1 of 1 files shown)


Identitaet:
Does the TL stand for transliterated here and the HJ for hanji ?

Looks good as far as I can tell

Supaplextw:
@Identitaet It is the [Tâi-lô system](https://en.wikipedia.org/wiki/T%C3%A2i-u%C3%A2n_L%C3%B4-m%C3%A1-j%C4%AB_Phing-im_Hong-%C3%A0n) to write Taiwanese Banlamgu. Not just transliteration.

Supaplextw:
and yes HJ is for Hanji.

Identitaet:
okay then it looks good
you probably know more about these language tags than me 😆

welcome to the Name suggestion index also

Supaplextw:
I'll add more chain stores of Taiwan to the name-suggestion-index. There are still many chain stores missing in the NSI.

PR #4862, opened 2021-01-17
Update restaurant.json

Supaplextw:
fix Din Tai Fung

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -4254,7 +4254,7 @@
     {
       "displayName": "一風堂",
       "id": "machidashouten-36ca8e",
-      "locationSet": {"include": ["jp","tw"]},
+      "locationSet": {"include": ["jp", "tw"]},
       "tags": {
         "amenity": "restaurant",
         "brand": "一風堂",
@@ -4505,7 +4505,7 @@
     {
       "displayName": "大戸屋",
       "id": "ootoya-36ca8e",
-      "locationSet": {"include": ["jp","tw"]},
+      "locationSet": {"include": ["jp", "tw"]},
       "tags": {
         "amenity": "restaurant",
         "brand": "大戸屋",
@@ -4524,7 +4524,7 @@
     {
       "displayName": "大阪王将",
       "id": "osakaohsho-36ca8e",
-      "locationSet": {"include": ["jp"]},
+      "locationSet": {"include": ["jp", "tw"]},
       "tags": {
         "amenity": "restaurant",
         "brand": "大阪王将",
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -1011,8 +1011,9 @@
       },
       "tags": {
         "amenity": "restaurant",
-        "brand": "Din Tai Fung",
+        "brand:en": "Din Tai Fung",
         "brand": "鼎泰豐",
+        "brand:zh": "鼎泰豐",
         "brand:wikidata": "Q699142",
         "brand:wikipedia": "en:Din Tai Fung",
         "cuisine": "taiwanese",
@@ -2661,7 +2662,8 @@
         "brand": "Ponderosa Steakhouse",
         "brand:wikidata": "Q64038204",
         "cuisine": "american;steak;buffet",
-        "name": "Ponderosa Steakhouse"
+        "name": "Ponderosa Steakhouse",
+        "name:zh": "龐德羅莎"
       }
     },
     {
@@ -4489,7 +4491,8 @@
         "cuisine": "japanese;chinese",
         "name": "大阪王将",
         "name:en": "Osaka Ohsho",
-        "name:ja": "大阪王将"
+        "name:ja": "大阪王将",
+        "name:zh": "大叛王將"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


data/brands/amenity/restaurant.json
```diff
@@ -1012,11 +1012,20 @@
...
         "amenity": "restaurant",
         "brand": "Din Tai Fung",
+        "brand": "鼎泰豐",
         "brand:wikidata": "Q699142",
         "brand:wikipedia": "en:Din Tai Fung",
         "cuisine": "taiwanese",
-        "name": "Din Tai Fung",
-        "name:zh": "鼎泰丰"
+        "name": "鼎泰豐"
+        "name:en": "Din Tai Fung",
```
UKChris-osm, line 1020: Hi @Supaplextw 

You'll need a comma after **"name": "鼎泰豐"** as it's currently causing a build error.

Adamant36: A lot of variations on the name tag for other entries, like name:nan-POJ, are pretty questionable to. Plus, they are just duplicates in a lot of cases. And name:zh probably shouldn't have been removed from this entry either.

Supaplextw:
@Adamant36 [POJ](https://en.wikipedia.org/wiki/Pe%CC%8Dh-%C5%8De-j%C4%AB) is the some people view it as written words for Taiwanese Hokkian. Even though the Minister of Education promote [Tâi-lô system](https://en.wikipedia.org/wiki/Taiwanese_Romanization_System), POJ is still one of the popular system to write Taiwanese Hokkian. So we in Taiwan have not had s unified system to write Taiwanese Hokkian.

data/brands/amenity/restaurant.json
```diff
@@ -1012,11 +1012,20 @@
       "tags": {
         "amenity": "restaurant",
         "brand": "Din Tai Fung",
+        "brand": "鼎泰豐",
```
1ec5, line 1015: In JSON, it’s invalid for an object to redefine the same property twice. There are some examples of `brand:en`, `brand:zh`, etc. in this repository.

data/brands/amenity/restaurant.json
```diff
@@ -1012,11 +1012,20 @@
...
         "brand:wikidata": "Q699142",
         "brand:wikipedia": "en:Din Tai Fung",
         "cuisine": "taiwanese",
-        "name": "Din Tai Fung",
-        "name:zh": "鼎泰丰"
+        "name": "鼎泰豐",
```
1ec5, line 1019: Din Tai Fung has locations in several countries (listed above), most of which do not use Chinese or Min Nan as their primary languages. As far as I can tell, this entry was originally written for the situation in countries like the United States, where `name` would be in pinyin and `name:zh` would still be tagged (since the chain does have a Chinese name). Since the situation is obviously different in Taiwan, we should add a separate entry for the same brand with `name` set to Chinese, alongside the existing one in pinyin. The two entries can have different `locationSet`s.

data/brands/amenity/restaurant.json
```diff
@@ -4205,7 +4232,14 @@
...
         "cuisine": "taiwanese",
-        "name": "三商巧福"
+        "name": "三商巧福",
+        "name:en": "Mercuries Food",
+        "name:hak": "Sâm-sông Kháu-fuk",
+        "name:zh": "三商巧福",
+        "name:nan": "Sam-siong Kháu-hok",
+        "name:nan-POJ": "Sam-siong Kháu-hok",
+        "name:nan-TL": "Sam-siong Kháu-hok",
+        "name:nan-HJ": "三商巧福"
```
1ec5, line 4242: The Min Nan name is unlikely to be relevant in Japan (which is listed as one of the locations above). This probably calls for separate entries for Japan versus Taiwan.

bryceco: The Din Tai Fung restaurants in Taiwan use both Chinese and English in the logo/signage, just as they do in the US. I don't think they need to be split apart. 

https://www.dintaifung.com.tw/index.php

https://dintaifungusa.com

1ec5: I assume you meant this in response to https://github.com/osmlab/name-suggestion-index/pull/4862#discussion_r560348748. (This thread is about a different chain in Taiwan and Japan.) Even so, the Min Nan POJ tagging would seem quite tangential for a location in an English-speaking country. (After all, we don’t tag KFC locations in France with `name:zh=肯德基`.)

Supaplextw:
@bryceco According to the Google Streetview image last June that capture the newest chain of "鼎泰豐" around Tongmen Station, Taipei, there is one sign used both Taiwanese Mandarin and English, but the name on the front is only in Taiwanese Mandarin, no English at all.

1ec5:
Right, that’s the appropriate tagging for the restaurant in Taiwan, but we won’t be able to merge this PR until it’s rewritten to add separate brands for the localized names rather than adding translations to the existing multinational brands.

Supaplextw:
@1ec5 I'll add a seperate 鼎泰豐 entry for Taiwan.

bhousel:
Thanks for updating the PR @Supaplextw 👍 
This looks mostly ok to me, so I'm just going to merge it as-is, and add a followup commit to address whatever other feedback people gave.

Issue #4848, opened 2021-01-07
How to treat name tags with combined latin and arabic alphabet?


kjonosm:
How should we treat collected brands that carry both the Arabic and Latin version together in their `name=*` Tag? There seems to have been an import of some North African banks recently with that naming style, see for example:
- [Attijariwafa Bank](https://nsi.guide/index.html?t=brands&k=amenity&v=bank#attijariwafabank-c1cae6)
- [Attijariwafa Bank التجاري وفا بنك](https://nsi.guide/index.html?t=brands&k=amenity&v=bank#2eed3c-b7a026)
- [التجاري وفا بنك](https://nsi.guide/index.html?t=brands&k=amenity&v=bank#dac1a1-b7a026)

bhousel:
Check https://github.com/osmlab/name-suggestion-index/issues/4673#issuecomment-734483414 for some advice, but this one feels like they should just be merged into a single entry with multiple language tags.

kjonosm:
How about to just merge all given entries into one and leave out the `name=*` tag to not potentially rename any current features in OSM? We only add `wikidata`, `name:en`, `brand:en`, `name:ar`, `brand:ar` tags to enrich OSM data.
Would that be a feasible approach to keep a certain consistency within the index?

bhousel:
> How about to just merge all given entries into one and leave out the `name=*` tag to not potentially rename any current features in OSM?

It seems ok to me... Or I usually just pick whatever is most common from what we've collected from the planet:
https://github.com/osmlab/name-suggestion-index/blob/main/dist/collected/names_all.json
https://github.com/osmlab/name-suggestion-index/blob/main/dist/collected/brands_all.json

I haven't looked at this one, but sometimes it's easy to tell which tags are mistakes / outliers.

PR #4867, opened 2021-01-19
merge bank with name spelled in different alphabets

kjonosm:
- add match names
- add language specific name:en and name:ar
- leave out name tag to not update existing features

(see: #4848)


bhousel:
Thanks @kjonosm 👍 
Looks good to me, I don't know what the build failure is about, but I'll merge and double check locally.

PR #4869, opened 2021-01-21
Fix japanese chains wikidata

venue-map:
Corrected Wikidata for Japanese chains.


bhousel:
Thanks for the fix @venue-map 👍

PR #4872, opened 2021-01-22
Update convenience.json

trigpoint:
Fix incorrect wikidata tag for Tesco Express

### DIFF
--- a/brands/shop/convenience.json
+++ b/brands/shop/convenience.json
@@ -1583,7 +1583,7 @@
     ],
     "tags": {
       "brand": "Tesco",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Thank you @trigpoint 

The suggested edits you have submitted were added to the index a little while ago, but thank you.

Also, the fork you are working from is very old, and your edits were made to a file structure that no longer exists. All the data is now stored in a folder called "_data_" and from here sub-folders exists for "_brands_", "_flags_", "_operators_" and "_transit_".

I would recommend deleting your current fork, and re-forking the latest version of the Name Suggestion Index :)

PR #4877, opened 2021-01-24
Remove name from Premier

peternewman:
Shops will often have a unique name (see the Weatherspoons situation).


UKChris-osm:
I agree, thank you @peternewman 🥇

PR #4885, opened 2021-01-31
Merge Europris and Eurorpris

UKChris-osm:
Removed "Eurorpris" (extra 'r') that seems to be a planet grab with a misspelling (as it's currently misspelled en-mass in OSM), and adds it as a match name to "Europris".

This was also done using GitHub Desktop to test my first usage of it!


tas50 (approved):

tas50:
Nice catch

PR #4889, opened 2021-02-03
add Gelatissimo

andrewharvey:


bhousel:
thanks again!
BTW, you're already a co-maintainer, so if you want to push small stuff like this directly to main, go for it 👍

andrewharvey:
Thanks Bryan, yes I might still go the PR route, but I can merge myself if no issue and CI passes. cheers.

PR #4897, opened 2021-02-06
Update bank.json

waso99:
Adding bank

### DIFF
--- a/brands/amenity/bank.json
+++ b/brands/amenity/bank.json
@@ -7282,6 +7282,22 @@
       "name:en": "Woori Bank",
       "name:ko": "우리은행"
     }
+     {
+      "displayName": "البنك الأهلي التجاري",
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @waso99 & welcome 👋 
The structure of our repository has changed a bit since the version that you have cloned. The main branch is called `main` and bank files are nested under `data/brands/amenity` now, so GitHub wasn't able to merge cleanly.  

But the bank data you provided was perfect so I just added it in 9a8b9fd5. 

I think it would be great if you can make a fresh clone of the repo and continue to send us Saudi Arabian brands that we don't have yet - we are always looking for experts in different parts of the world and I hope you will contribute more 👍

waso99:
Thanks @bhousel for support.
I will do my best.
when can I see this bank logo on ID editor?

PR #4899, opened 2021-02-07
Alsco

1ec5:
Added [Alsco](https://en.wikipedia.org/wiki/Alsco), a uniform rental service, as an operator. I’ve been tagging uniform rental services’ industrial laundry facilities as [`industrial=laundry`](https://taginfo.openstreetmap.org/tags/industrial=laundry) in the absence of an established tag. This tag isn’t common, but then again, neither have these facilities been mapped all that much, and `industrial=laundry` seems like a good analogue to the well-established `shop=laundry` (which this definitely isn’t). The [`industrial`](https://wiki.openstreetmap.org/wiki/Key:industrial) key itself is a bit of a free-for-all.

### DIFF
--- a/data/operators/industrial/laundry.json
+++ b/data/operators/industrial/laundry.json
@@ -1,0 +1,31 @@
+{
+  "operators/industrial/laundry": [
+    {
+      "displayName": "Alsco",
+      "id": "alsco-75a496",
+      "locationSet": {
+        "include": [
+          "au",
+          "br",
+          "ca",
+          "cn",
+          "de",
+          "gb",
+          "it",
+          "my",
+          "nz",
+          "sg",
+          "th",
+          "us"
+        ]
+      },
+      "tags": {
+        "industrial": "laundry",
+        "name": "Alsco",
+        "operator": "Alsco",
+        "operator:wikidata": "Q60752605",
+        "operator:wikipedia": "en:Alsco"
+      }
+    }
+  ]
+}


UKChris-osm:
I don't think "_Alsco_" operate as [Alsco](https://alsco.com/locations/) in the UK, and instead are called [Clean](https://www.cleanservices.co.uk/).

bhousel:
I'm a bit hesitant on this one also because of the `industrial=*` tag.  It is not really a toplevel tag, but rather it nests under `landuse=industrial`, and the `industrial=*` [page on the wiki](https://wiki.openstreetmap.org/wiki/Key:industrial) does say it "requires" a companion `landuse=industrial` tag. 

This is extra unfortunate because `landuse=*` used to be an area-implying tag until people started using it to map POI-like stuff.

LaoshuBaby:
The first I must say that I‘m not lived in US, but according to wikidata and the official website, I think …… maybe it offer a cluster of laundry? and offer services in large scales？
Maybe looks like very big and surprising, but for the part that provide end-user-oriented services in the form of stores, I think this still a kind of retail and should be tagged with ```shop=laundry```

(But for Alsco's washing factory workshop, that can be tagged as ```industrial=*```)

(Besides, Alsco is a big company, looks like it both have laundry store and textile processing bussiness? But in China don't offer service for single person, looks like only B2B)

图片

------

In short, I think the part of this company that provides services to individuals, no matter how big or automated it is, it is still a shop. Workshops and warehouses for the production of textiles and cleaning are industrial
(总之，我觉得这个企业面向个人提供服务的部分，不管再大，再自动化，洗衣机数量庞大，它还是一个shop。生产纺织品和清洗的车间、库房，那才是industrial)

bhousel:
I'm going to close this as the tagging is still not established.  (I feel bad)
- Wiki still says that `industrial=*` "requires" `landuse=*`
- `industrial=laundry` has almost no usage (only 23 times as of today)
- Without a source preset in [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema) it won't ever get used.

PR #4900, opened 2021-02-09
Update Tesla Supercharger operator

mxdanger:
Update operator of Tesla Supercharger to Tesla, Inc.


tas50:
Thanks for fixing this. It seemed a bit odd the previous way

Issue #2816, opened 2019-06-25
[bank] key:operator vs. Parent_company


ImreSamu:
Some bank operator is strange for me. ( https://osmlab.github.io/name-suggestion-index/brands/amenity/bank.html )

Imho:  the [Parent company](https://en.wikipedia.org/wiki/Parent_company)  !=  [OSM  Operator](https://wiki.openstreetmap.org/wiki/Key:operator)
 
in my sense:  operator == operating the bank == Usually visible on bills/contracts  with customers;  And has a local tax number ....

So I don't think, that the "National Bank of Greece" is visible on the "Banca Românească" bills ; 

in this case : "National Bank of Greece"  is  _"A parent company is a company that owns enough voting stock in another firm to control management and operation by influencing or electing its board of directors."_
and not the operator.

```
"amenity": "bank"
"brand": "Banca Românească"
"brand:en": "Romanian Bank"
"brand:ro": "Banca Românească"
"brand:wikidata": "Q4854052"
"brand:wikipedia": "ro:Banca Românească"
"name": "Banca Românească"
"name:en": "Popular Bank of Vicena"
"name:ro": "Banca Românească"
"operator": "National Bank of Greece"
"operator:wikidata": "Q1816028"
"operator:wikipedia": "en:National Bank of Greece"
```

![image](https://user-images.githubusercontent.com/2217320/60089193-41d9ba80-9740-11e9-924e-ad943d01f84b.png)
https://en.wikipedia.org/wiki/Banca_Rom%C3%A2neasc%C4%83

Adamant36:
If National Bank of Greece isn't the operator it can probably just be deleted. There isn't a parent_company tag though unfortunately. That said, sometimes the parent company can also be the operator in my opinion and  I still think National Bank of Greece is the operator in this case. It says on the National Bank of Greece's wiki page that it does banking services and that Banca Romaneasca is one of it's accusations. There's no mention of an intermediary firm. I don't know why there would be one either.

ImreSamu:
> That said, sometimes the parent company can also be the operator in my opinion

sometimes  "brand" - "operator" has an
* 1:1 relationships   ( no extra information ; can be added to the wikidata )
* 1:N relationships   ( we can't add to his register - only the local users  )

So I think -  We should remove all `operator=*`  tags from this repo.

[...]

PR #4902, opened 2021-02-13
Add Greene king

Cj-Malone:
[Greene King](https://www.greeneking.co.uk/) is a chain of pubs and hotels in Great Britain, they each have different names. I have found conformation that all of the hotels have free WiFi, but can't find the same confirmation for the pubs, although it's probably true. Existing OSM data supports `amenity=pub` over `amenity=restaurant` with 230 to 3 occurrences.

### DIFF
--- a/data/brands/amenity/pub.json
+++ b/data/brands/amenity/pub.json
@@ -78,7 +78,10 @@
         "amenity": "pub",
         "brand": "Hungry Horse",
         "brand:wikidata": "Q5943510",
-        "brand:wikipedia": "en:Hungry Horse"
+        "brand:wikipedia": "en:Hungry Horse",
+        "operator": "Greene King",
+        "operator:wikidata": "Q5564162",
+        "operator:wikipedia": "en:Greene King"
       }
     },
     {
--- a/data/brands/amenity/pub.json
+++ b/data/brands/amenity/pub.json
@@ -46,6 +46,17 @@
       }
     },
     {
+      "displayName": "Greene King",
+      "locationSet": {"include": ["gb"]},
+      "note": "No name for Greene King's brand as each pub is named differently.",
+      "tags": {
+        "amenity": "pub",
+        "brand": "Greene King",
+        "brand:wikidata": "Q5564162",
+        "brand:wikipedia": "en:Greene King"
+      }
+    },
+    {
       "displayName": "Hall & Woodhouse",
       "id": "hallandwoodhouse-eaed91",
       "locationSet": {"include": ["gb"]},
--- a/data/brands/amenity/pub.json
+++ b/data/brands/amenity/pub.json
@@ -20,7 +20,10 @@
         "amenity": "pub",
         "brand": "Chef & Brewer",
         "brand:wikidata": "Q5089491",
-        "brand:wikipedia": "en:Chef & Brewer"
+        "brand:wikipedia": "en:Chef & Brewer",
+        "operator": "Greene King",
+        "operator:wikidata": "Q5564162",
+        "operator:wikipedia": "en:Greene King"
       }
     },
     {
--- a/data/brands/tourism/hotel.json
+++ b/data/brands/tourism/hotel.json
@@ -567,6 +567,19 @@
       }
     },
     {
+      "displayName": "Greene King",
+      "locationSet": {"include": ["gb"]},
+      "note": "No name for Greene King's brand as each hotel is named differently.",
+      "tags": {
+        "brand": "Greene King",
+        "brand:wikidata": "Q5564162",
+        "brand:wikipedia": "en:Greene King",
+        "internet_access": "wlan",
+        "internet_access:fee": "customers",
+        "tourism": "hotel"
+      }
+    },
+    {
       "displayName": "Grupo Barceló",
       "id": "grupobarcelo-a464b1",
       "locationSet": {"include": ["es", "us"]},


Adamant36:
Hi. It appears there's some problems with this that need be worked out before it's merged.

1. Unless it's a new thing that I'm not aware of the NSI doesn't include notes in presets.
2.  Again, unless it's a new thing that I'm not aware of even if it can be confirmed that all the places have internet it's not something that is worth or should be included in the index.
3. If I am correct in how these this things are, having Greene King as the operator seems wrong, because aren't they technically  the owner and whoever runs the place is the operator?
4. Also, one of the entries, at line 51 in the pub file, is just for Greene King. Without any other information. I don't think it would work as a preset if everywhere is named differently and they therefore aren't really, technically the brand in those cases, because whatever the actual individual locations are called is.

Cj-Malone:
1. The note was copied from others without a name tag, see fullers-eaed91 and wetherspoon-eaed91.
2. `internet_access` and `internet_access:fee` has a little usage in this project already, but if it's been decided that this project is just for brand tags and not universal tags, that's fine, it can be removed. As for usefulness I don't think we can expect all OSM user agents to know `tourism=hotel` + `brand:wikidata=Q5564162` implies `internet_access=wlan`, so I feel it should be documented, even if you decide it shouldn't be documented here in nsi.
3. `operator=Greene King` was copied from wackywarehouse-6e0ce7, as far as I understand it, they are all owned and operated by Greene King. This isn't a franchise.
4. The hotel preset is the same on [line 569](https://github.com/osmlab/name-suggestion-index/blob/ede2eb65e14cd4a6e3cbb7012b920223ca69d544/data/brands/tourism/hotel.json#L569). I don't see this as being different from any of the other brands that have different names per location, like wetherspoon-eaed91 or harvester-9b5453.

Adamant36:
I'm not sure what the deal is with Wetherspoon is, because while it is already in the index it was removed at the end of issue #3103 by @bhousel because it was to much trouble for the index. I think the same problems for why it 
was removed would probably apply here. Also, issue #3009 that the PR was based on is not closed either. So my guess is that the existing  entry is a mistake  and therefore should not be used as an example of how to do things.

Anyway, things that generally shouldn't be in the index do slip by sometimes. For instance website urls. So, you can't always go on edge cases in the index to determine what the "correct" way to do things is.

Cj-Malone:
The `note` is defiantly not a child of `tags` and so isn't copied it OSM. Fully agree that a `note` shouldn't be put into OSM via NSI, but this is a note for NSI, not OSM.

UKChris-osm:
"_note_" is NSI only, and was introduced a few months back - it doesn't appear in the final build for ID / JSOM presets, and is just a way for us to add a little context to some entries that might need it, to prevent people asking about it later.

Adamant36:
Alright then. I must have missed it when they were added.

The issue is that there are pubs named "Greene King." So whatever the case is with usage of notes in NSI, the note in this instance  is simply wrong. Which is also why the entry with them as just the operator is incorrect IMO and why you can't just go off the other pub entry as an example, because they do have "Greene King" name/branded pubs. From I what can tell the different names are just location suffixes like hotels do with "whatever hotel brand wherever." These don't warrant special tagging or a note about it any more then those instances do IMO.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

UKChris-osm:
Do you mean there are Greene King branded pubs which have the same name, or non-affiliated pubs?

Adamant36:
(There's a Stag & Hounds pub not connected Greene King. So not calling the one that is "Greene King Stag and Hound" might be confusing. Plus, "Greene King" is on the sign anyway.

peternewman:
'Spoons came back in here after @bhousel made some changes so name is no longer required (precisely for situations like this):
https://github.com/osmlab/name-suggestion-index/pull/4183

So the Bury St Edmunds bit threw me, as that's their base, but your first image is actually a picture of the Two Chairmen (despite being used a s a stock photo in a story about renaming The Black Boy pubs):
https://goo.gl/maps/AjYVDp8svEXtJHQk9

I initially started in Bury and all the ones there have names:
https://www.greeneking.co.uk/map/

So do you have some other examples of Greene King pubs called just that with no name?

> (There's a Stag & Hounds pub not connected Greene King. So not calling the one that is "Greene King Stag and Hound" might be confusing. Plus, "Greene King" is on the sign anyway. Its clear they don't see themselves as just an anonymous operator of pubs that go by different names or brands.

Neither do Spoons, but everyone uses the two interchangeably but not together in my experience. There are seven Stag and Hounds in the UK:
https://www.pubnames.co.uk/pubcanon.php?canon=Stag-and-Hounds

Greene King alone had three The Black Boy's:
https://www.bbc.co.uk/news/uk-england-berkshire-56040545

We're not going to start trying to make their names fully unique are we?

I would have said we want to split stuff out as much as we can, i.e. keep the name clean, then someone could grab the brand logo from the wikidata and show the brand logo alongside the pub name for example.

Adamant36:
"So do you have some other examples of Greene King pubs called just that with no name?" 

You can look them on Google Images. There was that I saw just called Greene King without the location suffix, but I still don't we would treat it any different then we do with hotels where the preset has the brand name of the hotel since that's what is on the sign "on the ground" and then let people fill in the location suffix in the name field on their own if they want. 

For instance with Hyatt Regency Portland it would be way stupid and nonsensical if the preset for "Hyatt Regency" just filled in the operator and "Portland" was the only thing in the name field just because they call themselves "Hyatt Regency Portland." Just having the "Portland" at the end of the name doesn't we drop every other word in the before that and call the hotel chain only an operator of the hotel and not the brand or name of the hotel.  

"We're not going to start trying to make their names fully unique are we?" 

No we just put Greene King in the name field for the preset and let everyone decide on what else to add to it based what's on the ground. If the pubs signs say Greene King though, then that's what we should go with. 

"Neither do Spoons, but everyone uses the two interchangeably but not together in my experience. There are seven Stag and Hounds in the UK:" 

I had nothing to do with the Spoons preset being added to the index. If I did, I probably would done it differently. Generally, if company thinks their brand name is important enough to put on signs at actual locations then it should be in the index. Otherwise, we would be doing a lot of mind reading, guess work, and assuming what mappers would be looking for. In my first example a mapper consumer wouldn't even be able to see "Bury st Edmunds" from across the street if they were looking for the pub based just on that being the only thing on the map. They would see Greene King above it though. I don't really care if a few locals just call it Bury st Edmunds." Anymore then I would if a few locals in Portland called Hyatt Regency Portland "Portland." That's how things are tagged. I doubt anyone calls it just "Portland" anyway though.

Adamant36:
BTW, there's plenty of Wetherspoons out there that are just called "Wetherspoons." So 100% if # I was involved in that issue I would have said the same thing about it as I am here.

peternewman:
> "So do you have some other examples of Greene King pubs called just that with no name?"
> 
> You can look them on Google Images. There was that I saw just called Greene King without the location suffix, but I still don't we would treat it any different then we do with hotels where the preset has the brand name of the hotel since that's what is on the sign "on the ground" and then let people fill in the location suffix in the name field on their own if they want.

I've just had another look and failed to find anything, in reality that generic Greene King bit is a small plaque on the wall, or above/below the sign. Even the rebranded ones still have the name with more significance:
https://www.itv.com/news/anglia/2013-06-07/suffolk-brewery-accused-of-cultural-vandalism-after-scrapping-more-than-200-traditional-pub-signs

I think also if you asked a group of people what the name of any of these pubs was, the vast majority would say "Stag and Hounds". A small minority might mention the pubco that owns it too, I don't think any would just call it by the brand name.

So there are two challenges with the hotel approach, firstly with the current setup it overwrites the name if its in the preset, so if I'd already set it, it would flatten it, although @bhousel has been considering ways to workaround that during the v5 stuff. Secondly from a data perspective how do I tell if there is a name still to find using an app like StreetComplete if it's populated the brand. Also if I want accuracy I can probably remember it was a 'Spoons, but I'd want to check the exact name, with The or plural onsite.

> For instance with Hyatt Regency Portland it would be way stupid and nonsensical if the preset for "Hyatt Regency" just filled in the operator and "Portland" was the only thing in the name field just because they call themselves "Hyatt Regency Portland." Just having the "Portland" at the end of the name doesn't we drop every other word in the before that and call the hotel chain only an operator of the hotel and not the brand or name of the hotel.

Yeah fair point for the hotels, or you'd get multiple Portland's where each chain had their hotel in that area. FWIW it seems branch has been invented for that use:
https://wiki.openstreetmap.org/wiki/Key:branch

> No we just put Greene King in the name field for the preset and let everyone decide on what else to add to it based what's on the ground. If the pubs signs say Greene King though, then that's what we should go with.

But they don't, not as the primary part, or as any particularly significant bit, the name of the pub "Stag and Hounds" always takes top billing.

> I had nothing to do with the Spoons preset being added to the index. If I did, I probably would done it differently. Generally, if company thinks their brand name is important enough to put on signs at actual locations then it should be in the index. Otherwise, we would be doing a lot of mind reading, guess work, and assuming what mappers would be looking for.

Again they take second billing to the name of the pub:
https://www.jdwetherspoon.com/pubs/all-pubs/england/east-sussex/the-john-logie-baird-hastings

Should we put Free House on too?

> In my first example a mapper consumer wouldn't even be able to see "Bury st Edmunds" from across the street if they were looking for the pub based just on that being the only thing on the map.

Please take a look at the Google Streetview link I posted earlier, that pub is not called the Bury St Edmunds and it's not in Bury either, if anything it's a lazy journalist/stock photographer using an image of a pub near their office in London, but cropping out the name because the story was all about pub names and that pub doesn't even have the name they were talking about.

For the actual info about that pub, neither Wikipedia ( https://en.wikipedia.org/wiki/Two_Chairmen ), nor Greene King's own site ( https://www.greeneking-pubs.co.uk/pubs/greater-london/two-chairmen/ )  feature the brand particularly prominently. Indeed of the vintage of photos on their site, the house number is more prominent than the brand.

> Id find it extremely surprising if no one out there called the pubs Greene King, or Greene King whatever location. So there's zero wrong with having Greene King in the name and brand field for the preset and letting people on the ground decide the particulars if need be.

They don't call  their pubs Greene King on the website:
https://www.greeneking-pubs.co.uk/find-us/

The only time they mention locations are when there are two in the same county with the same name, and then it's only appended.

Interestingly Wetherspoons do have some called just their name, well sort of J D Wetherspoon versus Wetherspoons:
https://www.jdwetherspoon.com/~/media/images/pubs/0646/1.jpg

But given that’s in a shopping centre I suspect it's because there wasn't an old name to use.

peternewman:
> BTW, there's plenty of Wetherspoons out there that are just called "Wetherspoons." So 100% if # I was involved in that issue I would have said the same thing about it as I am here.

There appear to be 7:
https://www.jdwetherspoon.com/pubs/all-pubs?searchterm=Wetherspoons

They have more Red Lion's alone (11):
https://www.jdwetherspoon.com/pubs/all-pubs?searchterm=red%20lion

Adamant36:
"In reality that generic Greene King bit is a small plaque on the wall, or above/below the sign." 

Cool that your able to cherry pick an example. Good luck saying it about any of these though. Which I'm rather surprised you missed on Google Images. 


Are you seriously going to tell me that in those images that "Greene King" is just a small  plaque or on the wall compared to the actual name or above/below the actual sign? Id find it hard believe they would put Greene King in big bold letters on their doors if no one refered to them by that and if it wasn't the name of the place. 




In those last two, maybe it's "above" the name of location, but its still a part of the signs. 

"if you asked a group of people what the name of any of these pubs was, the vast majority would say "Stag and Hounds"." 

Maybe, but your ignoring the fact that we tag things based what is on signs on the ground. Not how some random person on the says to. Feel free to tag them with local_name or whatever the tag for what locals call things though. Id have zero problem with that, but that's not the point in presets. Ultimately, we could argue all day about locals call the place or not, but the signs clearly say Greene King. IMO that's what we should go with. 

"So there are two challenges with the hotel approach"

Good thing this PR isn't about hotels. 

"if I want accuracy I can probably remember it was a 'Spoons, but I'd want to check the exact name, with The or plural onsite." 

Good thing this PR isn't about Spoons. 

Im case your wondering, I'm a  not a huge fan of "other stuff exists" types arguments. If me as a calibrator who has been at this for four years and has a ton of edits/merged PR points out something I feel needs to be changed and the response is along the lines of "yeah well, blah blah blah some other entry", I tend to not take the person seriously after that. Personally, I could care less how Spoons does things or what @bhousel has planned for v5 This isn't a PR for Spoons or related to v5. Nothing personal.

"FWIW it seems branch has been invented for that use"

True, but I wouldn't personally consider a local hotel a "branch" just because the local suffix in the name. Its just the normal brand, as always, with an extra word. Otherwise, your really needlessly over complicating things. 

"But they don't, not as the primary part, or as any particularly significant bit, the name of the pub "Stag and Hounds" always takes top billing." 

Maybe your right for that one particular pub, but clearly that's the case with the other examples. Although, you are basing that on some random sample of local people would say..So...Do a Google Search for Stag and Hounds and tell me how search results down it is compared to all the other Stag and Hounds out there. Then do a search for "Greene King Stag and Hounds" and tell me where in the search results it comes up. 

"Again they take second billing to the name of the pub" 

Not in cases where the name of the pub is literally just "Wetherspoons"


"They don't call their pubs Greene King on the website"

Go to the actual website of a pub https://www.greeneking-pubs.co.uk/pubs/avon/horseshoe/

It literally says Greene King 15/1 compared to Horseshoe

Picking a random one and looking at the sign, it says Greene King twice on it and only the name of the location once..I guess that's meaningless though because Greene King is above and below "Fox & Hound." Shrug. 



"They have more Red Lion's alone"

Personally I wouldn't be so concerned in that case because "Wetherspoon" is off to side in small letters in most of those cases and not a part of the main sign. That can't be said for Greene King though. So they really aren't comparable.

Adamant36:
From far away do you think someone is going to see the crown and "Greene King" first or "Rose and Crown" hanging off the bottom of the sign?

Cj-Malone:
This is getting weirdly heated. Form a distance you are supposed to see the brand, it's the bit that's marketed, it's the known quantity. But any individual pub isn't "The Greene King", it has a name. Current OSM data backs this up, not one single `name=Greene King` or `name=The Greene King` in GB.

Adamant36:
"This is getting weirdly heated." 

Not on my end. I was just explaining my reasoning. I tend to stick to the on the ground rule then others might because it avoids these kinds of back and forth. I wouldn't call it heated though. 

Just to reiterate I have two issues  here. 

One the note says "No name for Greene King's brand as each pub is named differently." 

And that is clearly wrong because the name of the brand "Greene King" is "Greene King." Its on the signs. You even said so yourself.

" Form a distance you are supposed to see the brand, it's the bit that's marketed, it's the known quantity." 

"But any individual pub isn't "The Greene King", it has a name." 

I never said any individual pub was called "The Greene King." I said going by the signs they are called "The Greene King - name of location." Which is completely different then "The Greene King." 

"OSM data backs this up, not one single name=Greene King or name=The Greene King in GB." 

I'm aware. I looked through the current tagging and while none are technically called "Greene King" at the moment, that really doesn't matter as far as what to put in a preset. Most U-Haul places in America were originally called some odd variation of the name, but that doesn't mean we just went with operator for the preset and called it day. Presets are meant to work these types of things out and to standardize things. 

I personally have zero problem with location being in the name. I have a problem with a note saying none of the places go by Greene King when the actual signs say Greene King. The note and the preset aren't just for locals that go to the pubs. So they should both make sense to a general audience. Anyone that's not an insider looking up one of those pubs online, using the preset, or looking at the note would see that the signs say "Greene King" and think there's something wrong with either the note or the preset. You could just as well make the note clearer and leave the entry as is. I could really care less. 

That said, as I originally said  I also think the operator tag is problematic because according Wikipedia "Its pub partners division has leased, tenanted and franchised pubs" and according to the OSM Wiki "An operator isn't necessarily the owner of the map feature. Many chains (store, restaurants...) use a franchise system, where the brand does not operate the point-of-presence." So tagging all their pubs with them as the operator by way of the preset is wrong IMO.

bhousel:
Hey, just catching up on this Greene King issue.. This is a really good discussion about naming, and something that we haven't really resolved yet in NSI/iD, although we've gotten a lot better about it in the past year.  Thanks for everyone being patient as we figure it out.

_"What to do with names"_ is really the final unresolved thing about NSI that I need to figure out before making a real v5 release - you can see on https://github.com/openstreetmap/iD/pull/8305 that we're running into this same thing with Amazon Lockers (they all have unique names), and on https://github.com/openstreetmap/id-tagging-schema/issues/119 the same thing with cinema chains.

Here's the history of the issue, sorry that this goes off on a tangent from Green King, but it is related:

#### Historically:
- For years, NSI only looked at the `name` tag, and we generated iD presets and an iD validator that would overwrite the `name` tag with whatever NSI provided. 
- On the iD side, we have coded in a hack based on what fields the preset allows:
  - If both the "name" and "brand" fields are shown, we lock the "brand" and let the user change the "name"
  - If only the "name" field is shown, we lock the "name" field - user can't change it 
  - This allows the user change the name for things like Hotels and Car Dealerships, but it's not perfect - overwriting is still a problem

#### Now
- About 6 months ago, [I relaxed the name requirement in NSI](https://github.com/osmlab/name-suggestion-index/issues/3009#issuecomment-696902260).  Entries in NSI don't need a `name` anymore, so we can support pubs with unique names like Wetherspoon and Greene King, as well as features like transit and flagpoles, which share tags in common but aren't named like POIs.

#### The decision
I see three ways to handle a `name` tag, depending on what kind of a feature we have:
- "strict name" - NSI provides the correct name, and iD should enforce it (e.g. "Burger King")
- "default name" - NSI provides a name that works as a suitable default, but users can change it:  (e.g. "Amazon Locker" / "Amazon Locker - Hemlock") - _I think pubs are like this_
- "no name" - NSI doesn't contain a name tag and takes no opinion on what the user wants to do - bus routes are like this.

I think this information should live in NSI somewhere, and I should remove the iD hack mentioned above.  I also think that this should be a per-category decision (maybe items can override it), but this is tricky because NSI has quite a lot of categories.

Sorry to say that I have to think about this more, but it shouldn't hold up this PR which seems fine.

bhousel:
@Cj-Malone & @Adamant36  I removed the `operator` tags in 9df0bdd5
As mentioned in comments above, OSM has a very unintuitive notion of [what `operator` means](https://wiki.openstreetmap.org/wiki/Key:operator), and we don't want to add this tag unless it's universally true for the brand.  (e.g. a franchisee is not an operator, a parent company is not an operator).  
If I'm wrong, I apologize and we can definitely put it back later.

Cj-Malone:
Thanks, @bhousel. I don't have a strong position on `operator`, I kinda like the idea they they should all be linked, but if it doesn't fit the tag, fair enough. It should probably be removed from `wackywarehouse-6e0ce7` too.

peternewman:
> Hey, just catching up on this Greene King issue.. This is a really good discussion about naming, and something that we haven't really resolved yet in NSI/iD, although we've gotten a lot better about it in the past year. Thanks for everyone being patient as we figure it out.

Thanks for the detail @bhousel . Is it worth us having an issue to track these outstanding challenges in so we can link awkward edge cases like these to it?

> #### The decision
> I see three ways to handle a `name` tag, depending on what kind of a feature we have:
> 
> * "strict name" - NSI provides the correct name, and iD should enforce it (e.g. "Burger King")
> * "default name" - NSI provides a name that works as a suitable default, but users can change it:  (e.g. "Amazon Locker" / "Amazon Locker - Hemlock") - _I think pubs are like this_
> * "no name" - NSI doesn't contain a name tag and takes no opinion on what the user wants to do - bus routes are like this.

As I mentioned before, and maybe even more strongly in the case of the Amazon Locker, if we're storing the Hemlock bit in the name, as soon as we populate a default name, it then becomes an order of magnitude harder for other systems to realise that although there is a name, there is more data to gather and so they should still prompt for a name. I could definitely add some lockers to the map having walked passed them, but I'd need to go and explicitly check what random word is assigned to them when I'm in the vicinity.

> @Cj-Malone & @Adamant36 I merged this PR but removed the `operator` tags in [9df0bdd](https://github.com/osmlab/name-suggestion-index/commit/9df0bdd587e043c30bf3ba8084539a90a5b25dac)
> As mentioned in comments above, OSM has a very unintuitive notion of [what `operator` means](https://wiki.openstreetmap.org/wiki/Key:operator), and we don't want to add this tag unless it's universally true for the brand. (e.g. a franchisor is not an operator, a parent company is not an operator).
> If I'm wrong, I apologize and we can definitely put it back later.

Should the Greene King direct pubs, restaurants and hotels have the operator tag, as they run them (i.e. don't franchise them):
https://en.wikipedia.org/wiki/Greene_King#Operations

The Wacky Warehouse website says they're part of Spirit Pub Company:
https://www.wackywarehouse.co.uk/jobs

But it looks like they're out of date and that's just Greene King now:
https://en.wikipedia.org/wiki/Spirit_Pub_Company

But maybe I've misunderstood the operator thing? Although I see we don't populate it for some probably comparable alternatives either.

Adamant36:
"As I mentioned before, and maybe even more strongly in the case of the Amazon Locker, if we're storing the Hemlock bit in the name, as soon as we populate a default name, it then becomes an order of magnitude harder for other systems to realise that although there is a name, there is more data to gather and so they should still prompt for a name. I could definitely add some lockers to the map having walked passed them, but I'd need to go and explicitly check what random word is assigned to them when I'm in the vicinity." 

People should be walking by Amazon Lockers in real life to make sure there isn't more to the name then "Amazon Locker" instead of couch mapping anyway. Especially in case of QA tools like StreetComplete. So, I don't see how it's any different then anything else we have a preset for. It's possibly my local Safeway, Walmart, Shell gas station, bank branch that's part of a larger national banking chain etc. etc. wouldn't potentially have the same problem. Its not on to be that paranoid for every preset that we have to account for every single edge case piece of software that someone might accidently use to tag something using a preset without ground checking it first when they should be. Otherwise, there's zero point in the index in the first place. Amazon Lockers and cases like Greene King aren't particularly special in that regard.

"Should the Greene King direct pubs, restaurants and hotels have the operator tag, as they run them (i.e. don't franchise them):"

For me personally it depends on how many of each there are IRL. If there is only 10 hotels or whatever that they directly operate then it's not enough to warrant separate entries for each possible option that might be out there. Since people can just add the tags on their own if its smaller then 20 locations or whatever. The point in presets from my understanding is to help "normalize" and "standardize" the tagging on "large data sets" that might be all over the place tagging wise. If its something like 2 locations out of 50 totally of a restaurant having outdoor seating though then its not worth having a separate preset for that because people can just map the edge case themselves. In otherwards, this isn't a database of everything possible that can be adding to a brand or whatever. Its a database of the "basic" information on a brand that applies to most of them and can't realistically be added by individual mappers in a semi-standard way. At least that's how I see it.

bhousel:
from @peternewman 
> Thanks for the detail @bhousel . Is it worth us having an issue to track these outstanding challenges in so we can link awkward edge cases like these to it?

Good idea - I'll move some of my comment above to a new issue.

> As I mentioned before, and maybe even more strongly in the case of the Amazon Locker, if we're storing the Hemlock bit in the name, as soon as we populate a default name, it then becomes an order of magnitude harder for other systems to realise that although there is a name, there is more data to gather and so they should still prompt for a name. I could definitely add some lockers to the map having walked passed them, but I'd need to go and explicitly check what random word is assigned to them when I'm in the vicinity.

Yes, AFAIK people are already mapping the Amazon Lockers with the identifier in the name.  This is ok!  I think one of the challenges of this project is that it's not always up to us - we need to match the NSI and iD code to try to fit what mappers are doing and what people want their maps to look like.  There is a thread about the Amazon Lockers here:  https://osmus.slack.com/archives/C029HV951/p1607006115301000


> Should the Greene King direct pubs, restaurants and hotels have the operator tag, as they run them (i.e. don't franchise them):

If Greene King runs the pub then it's probably ok to include them as `operator`/`operator:wikidata` tags.  I'm pretty ok with leaving the tag off if unsure.  I don't think anyone really uses that tag for anything.

> But maybe I've misunderstood the operator thing? Although I see we don't populate it for some probably comparable alternatives either.

I removed them all a long time ago because many people were misunderstanding the tag - see #2816 and the handful of issues that came before it.

from @Adamant36 
> People should be walking by Amazon Lockers in real life to make sure there isn't more to the name then "Amazon Locker" instead of couch mapping anyway.

This did come up before, possibly on Slack, but I'm pretty sure the Amazon Locker identifiers are discoverable in real life.  I think it says in the app something like "Pick up your package here at Amazon Locker - Clarinet" and the word is written somewhere on the locker or at that location.

For example, here's how it looks on Google Maps around me, so users would expect something similar from an OSM map:

Screen Shot 2021-02-16 at 10 27 06 AM

peternewman:
> People should be walking by Amazon Lockers in real life to make sure there isn't more to the name then "Amazon Locker" instead of couch mapping anyway. Especially in case of QA tools like StreetComplete. So, I don't see how it's any different then anything else we have a preset for. It's possibly my local Safeway, Walmart, Shell gas station, bank branch that's part of a larger national banking chain etc. etc. wouldn't potentially have the same problem. Its not on to be that paranoid for every preset that we have to account for every single edge case piece of software that someone might accidently use to tag something using a preset without ground checking it first when they should be. Otherwise, there's zero point in the index in the first place. Amazon Lockers and cases like Greene King aren't particularly special in that regard.

I'm personally not really interested in couch mapping, so try to minimise what mapping I do sat at a computer to resolving my own notes. Because of that I probably don't survey as heavily as some people, but I might add a note about an Amazon Locker I see from a bus and then add it when I get home. However because I'm not walking I can't check the identifier of it (different sizes and clarities of branding). If we're only storing the identifier in the name, then aside from the lack of normalisation from a database perspective, making life harder for any data consumers, it's also a lot harder to QA, SC or iD or whatever can't easily flag there is some information missing (or they have to do a lot more work to spot that).

I'd suggest the same with Safeway, if they regularly had a branch name, which we wanted in the name field, perhaps we should drop the name from the preset, the map could fall back to the brand until someone had found it.

> Its a database of the "basic" information on a brand that applies to most of them and can't realistically be added by individual mappers in a semi-standard way. At least that's how I see it.

Broadly agree, although surely it needs to apply to *all* of them, or you start adding outdoor seating to pubs that don't actually have it.

Adamant36:
> I'm personally not really interested in couch mapping

Me neither. Except when I have to or feel like I need a day off from being chased by dogs or dealing with something else that makes on the ground mapping around where I live a massive pain a lot of the time. I was mostly speaking in relation to StreetComplete. Which should be done on the ground, but has an override feature for the requirement.  I don't think it uses the Name-Suggestion-Index though. Would be great if it did. 

> I'd suggest the same with Safeway, if they regularly had a branch name, which we wanted in the name field, perhaps we should drop the name from the preset, the map could fall back to the brand until someone had found it.

Naaw, in the case of Safeway the branch name 100% only serves corporate purposes and has no functional meaning outside of that. No Safeway sign says "Safeway 1954" on it and no one says they are going to "Safeway 1954" either. So the number is 100% a reference code and we shouldn't be encouraging putting them in the name field. Otherwise, it opens a whole can of worms. Walmarts are the same way. I'm pretty sure Starbucks, Barns & Nobles, Target, Homedepot, and probably most major retail chains are. They all have local store id's for corporate purposes only that no one IRL refers to them by, that shouldn't be dumped into the name field. 

> Broadly agree, although surely it needs to apply to all of them, or you start adding outdoor seating to pubs that don't actually have it.

I don't know if it does. I'm personally not a big of proponent of the whole "100% accuracy or nothing" crowd. Most of the time people that mentality are either micro-mappers with serious perfectionist issues or they mainly map in places that only already mostly filled in, where it's easy to ground confirm things 100% of the time, and they won't get chased by dogs or drug dealers when they stumble on a pot grow while trying to figure what surface a forest road is. In reality the map should be "just good enough." To me that means like 80% or closer to 90%, but it's not like there isn't notes and other QA tools to fix things in cases where they need to be. 

I rather 99% of the 4,756 Walmarts out there mapped in a standard way that people can understand then say to hell with it because the 1% has to be fixed by someone on the ground. That's no different then regular mapping without presets and there's nothing particularly special about them. Where presets should be held to a standard of perfection that isn't realistic and can't be met, or not be used at all. 

Also, I tend it's not worth mapping things that we (mappers and map consumers) are 100% sure about. For instance, it would be rather obtuse and useless IMO to tag every Walmart POI in America with toilets=yes (without other relevant details like if they are wheelchair accessible or whatever), because it's a given that there isn't going to be a Walmart in America that doesn't have a toilet. There was a recent proposal for a tag that would have likely led to every building in America and other places being tagged as electricity=yes that was shot down for the same reason. It's a given that most buildings in the world in the will have electricity. The rare cases that don't aren't worth a tag that causes the ones that do to be over tagged. So the proposal was shot down. Given that and common sense I think the community agrees with me. Or we would all be busy tagging buildings with walls=yes right now and similar nonsense.

Cj-Malone:
> Also, I tend to think it's not worth mapping things that we (mappers and map consumers) are 100% sure about. For instance, it would be rather obtuse and useless IMO to tag every Walmart POI in America with toilets=yes (without other relevant details like if they are wheelchair accessible or whatever), because it's a given that there isn't going to be a Walmart in America that doesn't have a toilet.

That's interesting, because as an English person, I wouldn't have made the assumption that every Walmart has a toilet. I could have found out via Googling, or asking an American, but are all OSM consumers supposed to do that, for every type of POI in the world?

I think it makes sense for these assumptions and defaults to be documented somewhere, maybe NSI isn't the place for that, but it is data that should be available to the OSM ecosystem.

> There was a recent proposal for a tag that would have likely led to every building in America and other places being tagged as electricity=yes that was shot down for the same reason. It's a given that most buildings in the world in the will have electricity. The rare cases that don't aren't worth a tag that causes the ones that do to be over tagged. So the proposal was shot down.

Having defined defaults could help with that. We'd just say every building (excluding building=no, building=shed, etc) in America has a default of `electricity=yes`, then if any given building doesn't a mapper could add `electricity=no` to override the default.

Adamant36:
> That's interesting, because as an English person, I wouldn't have made the assumption that every Walmart has a toilet. I could have found out via Googling, or asking an American, but are all OSM consumers supposed to do that, for every type of POI in the world?

I'd assume every major retailer in the UK would have toilets in there stores. I don't really care if those things are "documented" or tagged, I just don't it's the best thing for presets. Since there inherently has to be some things that are left to individual mappers to add. Even if they are true and can applied in 100% of the cases. Otherwise, it really risks inflating things extremely quickly and making this massively ungainly.  I take the same stance with other things also though. 

So, it's nothing against people knowing where there's a restroom or anything. I just think some things should be "out of scope" so to speak. I'm willing to admit I might be in the minority though. Luckily, my personal preferences don't carry any weight anywhere 😉 That said, the more this gets adopted, the more size of the database etc. matters. Since it can make a huge difference to mobile apps and the like. 

> Having defined defaults could help with that. We'd just say every building (excluding building=no, building=shed, etc) in America has a default of electricity=yes, then if any given building doesn't a mapper could add electricity=no to override the default.

Maybe. Things like the location conflation tool possibly could help to confine certain tags to specific geographical areas. I worry some about OSM becoming to siloed though. Where tagging is so varied from country to country or state to state that it is essentially worthless for anyone to use that doesn't live in that specific geographical area. It's already like that in some respects and it's sad to see where that is the case. 

That and if the "default" essentially becomes no default (where people are just filling in all the fields by themselves, except for in extremely small niche geographical areas) then I would seriously wonder what the point is.

peternewman:
> Me neither. Except when I have to or feel like I need a day off from being chased by dogs or dealing with something else that makes on the ground mapping around where I live a massive pain a lot of the time.

Ouch, I just occasionally feel slightly awkward/like a burglar wandering down the street staring at houses!

> I was mostly speaking in relation to StreetComplete. Which should be done on the ground, but has an override feature for the requirement. I don't think it uses the Name-Suggestion-Index though. Would be great if it did.

Yeah, I was just meaning that if the name field had been pre-populated with a partial name, it's a lot harder to add a quest to SC to ask to populate the full name.

> Naaw, in the case of Safeway the branch name 100% only serves corporate purposes and has no functional meaning outside of that. No Safeway sign says "Safeway 1954" on it and no one says they are going to "Safeway 1954" either. So the number is 100% a reference code and we shouldn't be encouraging putting them in the name field. Otherwise, it opens a whole can of worms. Walmarts are the same way. I'm pretty sure Starbucks, Barns & Nobles, Target, Homedepot, and probably most major retail chains are. They all have local store id's for corporate purposes only that no one IRL refers to them by, that shouldn't be dumped into the name field.

Ah, that's very different to the UK then. A lot of stores will have a name listed on their website and in real life (I realise this isn't a perfect example as they don't match, but I suspect the StreetView is old, I was struggling to get a pairing):
https://goo.gl/maps/uW1aLtevEVWDDYnL9
https://www.tesco.com/store-locator/uk/?bID=3482

> > Broadly agree, although surely it needs to apply to all of them, or you start adding outdoor seating to pubs that don't actually have it.
> 
> I don't know if it does. I'm personally not a big of proponent of the whole "100% accuracy or nothing" crowd. Most of the time people that mentality are either micro-mappers with serious perfectionist issues or they mainly map in places that only already mostly filled in

But if it was in the preset it would be constantly adding it back automatically even if somewhere didn't have it. Although possibly not with some of the more recent changes.

> In reality the map should be "just good enough." To me that means like 80% or closer to 90%, but it's not like there isn't notes and other QA tools to fix things in cases where they need to be.

Again broadly agree, I suspect updating a closed business is probably more use than adding whether a pub has outdoor seating (it might be full anyway)

> I rather 99% of the 4,756 Walmarts out there mapped in a standard way that people can understand then say to hell with it because the 1% has to be fixed by someone on the ground. That's no different then regular mapping without presets and there's nothing particularly special about them. Where presets should be held to a standard of perfection that isn't realistic and can't be met, or not be used at all.

Surely it's just about the info you add to the preset? I'd hope we can all agree Walmart exists and the Wikidata ID represents it etc.

> > Also, I tend to think it's not worth mapping things that we (mappers and map consumers) are 100% sure about. For instance, it would be rather obtuse and useless IMO to tag every Walmart POI in America with toilets=yes (without other relevant details like if they are wheelchair accessible or whatever), because it's a given that there isn't going to be a Walmart in America that doesn't have a toilet.
> 
> That's interesting, because as an English person, I wouldn't have made the assumption that every Walmart has a toilet. I could have found out via Googling, or asking an American, but are all OSM consumers supposed to do that, for every type of POI in the world?

> I'd assume every major retailer in the UK would have toilets in there stores.

Yeah as a Brit, a lot of smaller shops/chains may not have toilets (I don't know if you mean major as in large floor area, or retailer with many stores), or I know in France for example some large supermarkets don't have their own toilets, but the "shopping centre" they are within have them instead.

> I think it makes sense for these assumptions and defaults to be documented somewhere, maybe NSI isn't the place for that, but it is data that should be available to the OSM ecosystem.

> I don't really care if those things are "documented" or tagged, I just don't it's the best thing for presets. Since there inherently has to be some things that are left to individual mappers to add. Even if they are true and can applied in 100% of the cases. 

> So, it's nothing against people knowing where there's a restroom or anything.

Although if you're a data consumer that means you have to "enrich" your data to be able to find the answer if it's a mix of presets and raw tags. Realistically finding the nearest toilet is going to be a popular quest (for people with young children, those with disabilities/medical issues).

At least one UK supermarket even uses it as a selling point in their TV adverts and does press releases about it:
https://www.tescoplc.com/news/2020/tesco-installs-100th-changing-places-facility/

Obviously extending that data so you can find the nearest disabled loo or one with a baby changing area will help for more specialised groups, but having that base info would help, for example StreetComplete wouldn't want to ask if every corner shop had a toilet, but it could ask the question of every Tesco, or ideally if that was guaranteed it could go straight onto asking if it had a disabled toilet.

> Otherwise, it really risks inflating things extremely quickly and making this massively ungainly. I take the same stance with other things also though.

> That said, the more this gets adopted, the more size of the database etc. matters. Since it can make a huge difference to mobile apps and the like.

I don't think that's really an issue, apart from data usage if streaming the data live. Clearly your average phone isn't downloading the entire planet file, so must be filtering the data anyway. Unless you are an OSM editing app, that's fine, if you're doing navigation you might throw away a lot of detail about building levels and roofs (unless you're doing 3D), if you're a toilet finder you don't care about electricity, and if you're an app for charging electric cars or bikes (or cafe workers or nomadic tourists) you might want to keep electricity=? and WiFi tags.

As we found when looking at roofs (flat versus sloped) for StreetComplete, there isn't necessarily even consistency of that in the same country, let alone continent.

PR #4904, opened 2021-02-15
Add Farmhouse Inns and Belhaven Pubs

Cj-Malone:
These are different brands that Greene King has bough, but continues to operate separate brands. They both sit on the boundary of pub with food vs restaurant, but doing a spot check on current OSM data Belhaven Pubs gets mainly `amenity=pub`, and Farmhouse Inns gets `amenity=restaurant`. I don't disagree with those definitions.

### DIFF
--- a/data/brands/amenity/pub.json
+++ b/data/brands/amenity/pub.json
@@ -1,6 +1,15 @@
 {
   "brands/amenity/pub": [
     {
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks @Cj-Malone , this looks fine to me.. Sorry for the slow merge!

PR #4905, opened 2021-02-16
Adding  Brand names

waso99:
Adding  some Brand names, and correct misspelling.


waso99:
Thanks, for support 👍

bhousel:
Thanks @waso99 looks great 👍   
When we fix a misspelling, it's important to also add a `matchNames` property - this so the build script will match the old name also, so that it will not try to add it back to the index. 

https://github.com/osmlab/name-suggestion-index/blob/71f979a93c5fe23a86f08d36b3ca3df7323008a3/data/brands/amenity/fuel.json#L4892-L4898

I also removed the redundant `alt_name:en` tags, because they were the same as `name:en`.

Issue #4868, opened 2021-01-20
Duplication of Japanese mobile_phone


venue-map:
[Softbank](https://nsi.guide/index.html?t=brands&k=shop&v=mobile_phone&tt=softbank)
[docomo](https://nsi.guide/index.html?t=brands&k=shop&v=mobile_phone&tt=docomo)

According overpass all are in Japan,.

nuxper:
Hi @venue-map

Do you know Softbank ?

I understand there are two brands in wikidata for mobile_phone shops:
* SoftBank Telecom (Q7553832)
* SoftBank Shop (Q11315281)
Are both legitimate and distinct retail chains ?

Then there are 258 nodes with just name=SoftBank in openstreetmap. Hence name-suggestion-index automatically creates an entry fot this third brand. Do you think nodes with name=SoftBank are actually part of Q7553832 ? or Q11315281 ? or a third retail network maybe ?

venue-map:
Hi @nuxper -san
Sorry for the late reply.

> I understand there are two brands in wikidata for mobile_phone shops:
> SoftBank Telecom (Q7553832)
> SoftBank Shop (Q11315281)
> Are both legitimate and distinct retail chains ?

The Japanese version of Wikipedia clearly states that "SoftBank Telecom has merged with SoftBank Mobile.".
[ソフトバンク - Wikipedia](https://ja.wikipedia.org/wiki/%E3%82%BD%E3%83%95%E3%83%88%E3%83%90%E3%83%B3%E3%82%AF)
> 2015年4月1日付で、ソフトバンクモバイルがソフトバンクBB、ソフトバンクテレコム（ソフトバンクテレコム販売と逆さ合併した2代目）、ワイモバイルを吸収合併。

By the way, I don't think there was a store for "SoftBank Telecom (Q7553832)" itself.

------
> Then there are 258 nodes with just name=SoftBank in openstreetmap. Hence name-suggestion-index automatically creates an entry fot this third brand. Do you think nodes with name=SoftBank are actually part of Q7553832 ? or Q11315281 ? or a third retail network maybe ?

For "name = SoftBank", the Wikidata for "Q11315281" is correct.
The signboard of "Softbank shop" says "Softbank", so I think that's what it says.

nuxper:
I think I did the required modifications. Is that ok with you @venue-map ?

venue-map:
Hi @nuxper -san.

Thank you for editing.
[...]

Issue #4906, opened 2021-02-16
Store more information (including name handling rules) for each category


bhousel:
Moving this discussion to a new issue from the Greene King PR:

_"What to do with names"_ is really the final unresolved thing about NSI that I need to figure out before making a real v5 release - you can see on https://github.com/openstreetmap/iD/pull/8305 that we're running into this same thing with Amazon Lockers (they all have unique names), and on https://github.com/openstreetmap/id-tagging-schema/issues/119 the same thing with cinema chains.

I see three ways to handle a `name` tag, depending on what kind of a feature we have:
- "strict name" - NSI provides the correct name, and iD should enforce it (e.g. "Burger King")
- "default name" - NSI provides a name that works as a suitable default, but users can change it:  (e.g. "Amazon Locker" / "Amazon Locker - Hemlock") - _I think pubs are like this_
- "no name" - NSI doesn't contain a name tag and takes no opinion on what the user wants to do - bus routes are like this.

I think this information should live in NSI somewhere, so we can remove the iD preset-field hack.  I also think that this should be a per-category decision (maybe items can override it), but this is tricky because NSI has quite a lot of categories.

So I'll be adjusting the file formats a bit to settle on a nicer way to capture more information for each category.   

_Originally posted by @bhousel in https://github.com/osmlab/name-suggestion-index/issues/4902#issuecomment-779351342_

kjonosm:
> "default name" - NSI provides a name that works as a suitable default, but users can change it: (e.g. "Amazon Locker" / "Amazon Locker - Hemlock") - I think pubs are like this

Some more potential candidates for "default names" based on my experience:
- `shop=car` these usually have their own name, often in combination with the name of the brand of the cars they sell.
- `office=insurance` small branch offices are often named after the actual broker (person) who runs it.
- `tourism=hotel` hotels often combine brand name and location to make it easier to find them.
- `shop=travel_agency` small branch offices are often named after the person who runs it.

I think your suggestion to split name handling in 3 categories works well. Is it enough to define a category based only on the feature tag or do we need to categorize individual brands/features?

bhousel:
> Some more potential candidates for "default names" based on my experience:

All good suggestions!   I'd add amenity=cinema to that list too.. https://github.com/openstreetmap/id-tagging-schema/issues/119

> Is it enough to define a category based only on the feature tag or do we need to categorize individual brands/features?

I'm going to try to make it per-category, but allow override per-brand.

[...]

PR #4916, opened 2021-02-19
merge Softbank

venue-map:
#4868 has been closed because SoftBank Telecom (Q7553832) and SoftBank Shop (Q11315281) are not integrated, so I edited it.

> I understand there are two brands in wikidata for mobile_phone shops:
> SoftBank Telecom (Q7553832)
> SoftBank Shop (Q11315281)
> Are both legitimate and distinct retail chains ?
> 
> The Japanese version of Wikipedia clearly states that "SoftBank Telecom has merged with SoftBank Mobile.".

### DIFF
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -1118,7 +1118,7 @@
       "displayName": "ソフトバンクショップ",
       "id": "softbankshop-0a8be9",
       "locationSet": {"include": ["jp"]},
-      "matchNames": ["softbank","ソフトバンク"],
+      "matchNames": ["softbank","ソフトバンク","ソフトバンクショップ"],
       "tags": {
         "brand": "ソフトバンクショップ",
         "brand:en": "SoftBankShop",


nuxper:
@bhousel : A little help would be appreciated ;) : I'm happy with the proposed pull request : Softbank Telecom was merged in Softbank in 2015 and doesn't seem to exist anymore. However there are still 208 nodes with this brand:wikidata in Japan ( https://overpass-turbo.eu/s/13RP ). Softbank Shop, which I understand is the correct preset now has 391 ( https://overpass-turbo.eu/s/13RQ )

Are there specific steps in this case ? At least remove the name-suggest-index id from wikidata I suppose ? Anything else ? I couldn't find specific instructions in the doc but may have missed them.

venue-map:
@nuxper -san.

> Are there specific steps in this case ? At least remove the name-suggest-index id from wikidata I suppose ? Anything else ? I couldn't find specific instructions in the doc but may have missed them.

I think it's okay to remove the name-suggest-index id from Wikidata.
This is because SoftBank Telecom (Q7553832) does not operate a store.
(Actually, SoftBank Telecom (Q7553832) is not actually involved in the operation of the mobile phone shop.)

Adamant36:
So, not to over turn the apple cart, but their store signs say "SoftBank."  Their social media account says "SoftBank" also. As well as their website. So, where is it officially "SoftBank Shop" and why does it matter in this case what they are "officially" called if they don't use the word "shop" literally anywhere? (I assume there's a space between "SoftBank" and "Shop", but that's a different issue). Or are you just saying that's their "official" name because they are a shop?

Adamant36:
Minor update, I guess the build errors are caused by  the version of Moment being used in the package.json file or something. At least that's the error I got when I did npm run build. It "Can't find module local-compare." I'm not super on my build errors though and it looks like other PRs are having the same problem of showing errors on Githubs end @bhousel.

bhousel:
from @nuxper 
> Are there specific steps in this case ? At least remove the name-suggest-index id from wikidata I suppose ? Anything else ? I couldn't find specific instructions in the doc but may have missed them.

The PR looks ok to me, thanks @venue-map 👍 .  When we sync up with wikidata our script will remove the NSI identifier from their end automatically, but there is no harm in removing it manually.

from @Adamant36 
> So, where is it officially "SoftBank Shop" and why does it matter in this case what they are "officially" called if they don't use the word "shop" literally anywhere? 

I'm not going to worry about this much, since it's primarily a Japanese brand.  The `*:en` tags are really there for our convenience and might just be simple translations of the Japanese names.

> (BTW, it looks like there's a ton of build errors with the build scripts in the Files Changed section also. I don't know if it effects the PR any though. Since it still says it passed the checks, but I thought I'd mention it).

It's ok! these are just warnings.  I'll get rid of them soon, but I have been changing a bunch of stuff around in all the files to make it possible to store per-category config values for #4906.

Adamant36:
> It's ok! these are just warnings. I'll get rid of them soon, but I have been changing a bunch of stuff around in all the files to make it possible to store per-category config values for #4906.

Alright. Good to know 👍

PR #4920, opened 2021-02-21
Clarify Habitat split

peternewman:
I was going to just add GB to the list, but then I did a bit more looking. I think the UK Wikidata item has the wrong URL (should be https://www.habitat.co.uk/) and French Wikipedia doesn't seem to link to the right Wikidata/fully acknowledge the split.

### DIFF
--- a/data/brands/shop/furniture.json
+++ b/data/brands/shop/furniture.json
@@ -470,7 +470,7 @@
       "tags": {
         "brand": "Habitat",
         "brand:wikidata": "Q28232260",
-        "brand:wikipedia": "fr:Habitat (entreprise)",
+        "brand:wikipedia": "en:Groupe Habitat",
         "name": "Habitat",
         "shop": "furniture"
       }
--- a/data/brands/shop/furniture.json
+++ b/data/brands/shop/furniture.json
@@ -436,7 +436,22 @@
       }
     },
     {
-      "displayName": "Habitat",
+      "displayName": "Habitat (UK)",
+      "locationSet": {
+        "include": [
+          "gb"
+        ]
+      },
+      "tags": {
+        "brand": "Habitat",
+        "brand:wikidata": "Q689059",
+        "brand:wikipedia": "en:Habitat_(retailer)",
+        "name": "Habitat",
+        "shop": "furniture"
+      }
+    },
+    {
+      "displayName": "Groupe Habitat",
       "id": "habitat-437cef",
       "locationSet": {
         "include": [
@@ -454,7 +469,7 @@
       },
       "tags": {
         "brand": "Habitat",
-        "brand:wikidata": "Q689059",
+        "brand:wikidata": "Q28232260",
         "brand:wikipedia": "fr:Habitat (entreprise)",
         "name": "Habitat",
         "shop": "furniture"


data/brands/shop/furniture.json
```diff
@@ -454,7 +469,7 @@
       },
       "tags": {
         "brand": "Habitat",
-        "brand:wikidata": "Q689059",
+        "brand:wikidata": "Q28232260",
         "brand:wikipedia": "fr:Habitat (entreprise)",
```
bhousel, line 473: I guess this could link to the en version of the page (which is a sitelink for Q28232260).  
The en version explains that the UK and International organizations have separated..

```suggestion
        "brand:wikipedia": "en:Groupe Habitat",
```

data/brands/shop/furniture.json
```diff
@@ -436,7 +436,22 @@
...
+      "tags": {
+        "brand": "Habitat",
+        "brand:wikidata": "Q689059",
+        "brand:wikipedia": "en:Habitat_(retailer)",
+        "name": "Habitat",
+        "shop": "furniture"
+      }
+    },
+    {
+      "displayName": "Groupe Habitat",
```
nuxper, line 454: I agree with you and with the modification suggested by @bhousel .

Just be careful that interior_decoration.json aslo has a Habitat (UK).
In France at least, the Groupe Habitat sells both furnitures and interior decorations but I believe they focus a bit more on furnitures + Almost all the Habitat shops in France are mapped with shop=furniture. I'm not sure about the british branch though. What do you think ?

nuxper (approved):

data/brands/shop/furniture.json
peternewman: Done thanks. I wasn't sure of the policy regards which language should be linked to for non-English websites.

data/brands/shop/furniture.json
peternewman: Good spot @nuxper . That's got yet another wikidata ID!:
https://github.com/osmlab/name-suggestion-index/blob/ea9d441ffb3e1594fca160910219929b4393eb12/data/brands/shop/interior_decoration.json#L76-L89

The UK seems to have a mix of homewares/interior decoration/furniture currently!

The UK site lists stuff that's not furniture too, so maybe interior decoration is better.

@bhousel should it fail if there are two instances of a place with overlapping match tags (to catch issues like this)?

bhousel: @peternewman currently it will only fail if a duplicate id is generated.  The ids are generated by name + a hashcode that is unique across tree, key, value, locationset

I don't think the build script will fail if there is an overlapping matchtag, so that's a good suggestion..

peternewman: > @peternewman currently it will only fail if a duplicate id is generated. The ids are generated by name + a hashcode that is unique across tree, key, value, locationset

Is tree the file path?

> I don't think the build script will fail if there is an overlapping matchtag, so that's a good suggestion..

I've created #4921 to track this. It certainly hasn't complained currently.

bhousel: I haven't looked at this lately, but is is ok to merge? 
The decision is really whether we want to nudge all of the Habitat stores to use a consistent tag or whether we allow mappers to choose between furniture and interior_decoration.

UKChris-osm:
> I haven't looked at this lately, but is is ok to merge?
> The decision is really whether we want to nudge all of the Habitat stores to use a consistent tag or whether we allow mappers to choose between furniture and interior_decoration.

Habitat branded furniture goods (beds, etc) are available via Argos throughout the rest of the Country. Smaller interior goods are available from Argos as well, but will also be stocked on shelves within Sainsbury's, along with other interior goods.

Individual Habitat stores are so [few in number (4)](https://www.habitat.co.uk/stores) and are unlikely have any more stores added to the brand, that it may not be worth having in the NSI.

peternewman:
> > I haven't looked at this lately, but is is ok to merge?
> > The decision is really whether we want to nudge all of the Habitat stores to use a consistent tag or whether we allow mappers to choose between furniture and interior_decoration.

Sorry @bhousel I got sidetracked with other stuff.

> Individual Habitat stores are so [few in number (4)](https://www.habitat.co.uk/stores) and are unlikely have any more stores added to the brand, that it may not be worth having in the NSI.

Yeah, although I think we've currently got two or three different sets of tagging across those stores!

I don't mind either way, but at a minimum we should add a note to the other Habitat to stop people adding GB to the location list, I wonder if just adding it to NSI would simplify all of that?

bhousel:
> > Individual Habitat stores are so [few in number (4)](https://www.habitat.co.uk/stores) and are unlikely have any more stores added to the brand, that it may not be worth having in the NSI.
> 
> Yeah, although I think we've currently got two or three different sets of tagging across those stores!

I agree, if it's just 4 stores in the UK, we can probably leave it out of NSI, and someone can just manually tag those 4 stores with whatever tags are appropriate, and `not:brand:wikidata=Q28232260`.
 
> I don't mind either way, but at a minimum we should add a note to the other Habitat to stop people adding GB to the location list, I wonder if just adding it to NSI would simplify all of that?

Sounds good.. I'll merge and make the updates and add a note 👍

peternewman:
Were we going to delete the interior design one too @bhousel ?

Also can a Wikidata wizard fix the apparent duplication between https://www.wikidata.org/wiki/Q689059 and https://www.wikidata.org/wiki/Q87145455 .

I've fixed the tagging on Leeds and Finchley Road. I couldn't find/they don't seem to exist for Brighton and Westfield.

PR #4923, opened 2021-02-24
Change alfamart from shop=supermarket to shop=convenience

Jinkiesz:
Its not a supermarket, you can verify it on wikipedia or google


bhousel:
Thanks @Jinkiesz & welcome 👋

PR #4932, opened 2021-03-02
adding 2 brands

waso99:
adding 2 brands (Meed and NAFT)

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -34,6 +34,36 @@
       }
     },
     {
+      "displayName": "ميد",
+      "locationSet": {"include": ["sa"]},
+      "tags": {
+        "amenity": "fuel",
+        "brand": "ميد",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/fuel.json
```diff
@@ -33,6 +33,36 @@
...
+    {
+      "displayName": "ميد",
+      "locationSet": {"include": ["sa"]},
+      "tags": {
+        "amenity": "fuel",
+        "brand": "ميد",
+        "brand:ar": "ميد",
+        "brand:en": "MEED",
+        "brand:wikidata": "Q105492269",
+        "brand:wikipedia": "ar:شركة ميد التجارية",
```
Identitaet, line 45: here we only need شركة ميد insted of  شركة ميد التجارية
don't know why but wikipedia say so

waso99: Sorry, this is mistake.
I will fix it

PR #4933, opened 2021-03-03
Correct "MEED Trading Company"

waso99:
Correct "MEED Trading Company" Wikipedia Arabic name

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -43,7 +43,7 @@
         "brand:en": "MEED Trading Company",
         "brand:wikidata": "Q105492269",
         "brand:wikipedia": "ar:شركة ميد",
-        "name": "شركة ميد",
+        "name": "ميد",
         "name:ar": "ميد",
         "name:en": "MEED"
       }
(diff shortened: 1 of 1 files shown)


data/brands/amenity/fuel.json
```diff
@@ -33,6 +33,36 @@
...
+      "displayName": "ميد",
+      "locationSet": {"include": ["sa"]},
+      "tags": {
+        "amenity": "fuel",
+        "brand": "ميد",
+        "brand:ar": "شركة ميد التجارية",
+        "brand:en": "MEED Trading Company",
+        "brand:wikidata": "Q105492269",
+        "brand:wikipedia": "ar:شركة ميد",
+        "name": "شركة ميد",
```
bhousel, line 46: Normally the `name` tag should match the `name:ar` tag in this case..
I think some validators (KeepRight and JOSM?) may give a warning if there are localized names but none of them matches `name`.

data/brands/amenity/fuel.json
```diff
@@ -33,6 +33,36 @@
...
+      "locationSet": {"include": ["sa"]},
+      "tags": {
+        "amenity": "fuel",
+        "brand": "نفط",
+        "brand:ar": "نفط",
+        "brand:en": "NAFT",
+        "brand:wikidata": "Q105572509",
+        "brand:wikipedia": "ar:شركة نفط (السعودية)",
+        "name": "شركة نفط",
+        "name:ar": "نفط",
```
bhousel, line 62: another `name` / `name:ar` mismatch
(I don't really know if it's right or wrong, just that some validators may complain)

PR #4943, opened 2021-03-06
Australian EV network wikidata

lkernan:
Add operator:wikidata tags for Chargefox and Evie Networks.


UKChris-osm:
Thank you @lkernan 👍

PR #4952, opened 2021-03-10
Further Tas, NZ, UK charging station operators

chuqtas:
Add Bolt, and two city councils which operate chargers


bhousel:
Looks great @chuqtas, thanks for these additions 👍

PR #4955, opened 2021-03-11
Australian Stores and Parcel Lockers

lkernan:
Adds some new Australian stores and Parcel Pickup lockers.

### DIFF
--- a/data/brands/amenity/vending_machine.json
+++ b/data/brands/amenity/vending_machine.json
@@ -88,6 +88,19 @@
       }
     },
     {
+      "displayName": "Australia Post Parcel Locker",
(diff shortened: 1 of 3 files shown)


bhousel:
Looks good @lkernan thanks 👍 

We sometimes find different Wikipedia/Wikipedia for different store formats" (e.g. K-Mart vs K-Hub).  In this situation, assigning K-Hub the `"brand": "Kmart"` seems fine, and it's probably ok to display the Kmart logo for the KHub stores.  I'm just mentioning this in case someday someone creates a separate wikipedia/wikidata page for KHub, we can switch the tag later.

PR #4959, opened 2021-03-12
Modify 2 fast_food

LaoshuBaby:
## Modify:
 + 华莱士
 + 德克士


LaoshuBaby:
Try to describe those 2 brand exactly in cuisine.

The reason for wishing to add the cuisine label is because these chain restaurant brands directly appear "have both fried chicken and burger" as part of the restaurant name

Next are a few restaurant photos that I grabbed by entering the brand name from the Internet. It can be seen that the newly added matchnames all reflect the real aliases for these restaurants, and reflect that the cuisine uses these two common foods as the main dishes at the same time.

华莱士炸鸡汉堡

华莱士炸鸡汉堡

LaoshuBaby:
For this reason, I modify ```cuisine``` to this:

```
"cuisine": "burger;chicken",
```

LaoshuBaby:
Based on my observations on the PR records of osmlab/name-suggestion-index, especially those PRs that have not been agreed to merge into the master for a long time, I think my addition today may cause arguments like #4855 

Therefore, I will let this commit that may cause controversy as this PR separately, so as not to affect the merge of other more basic and no-controversial commits.

望各位领导批评指正！

bhousel:
> Therefore, I will let this commit that may cause controversy as this PR separately, so as not to affect the merge of other more basic and no-controversial commits.

Thanks!  This change is fine .. I think updating `cuisine` tags is not controversial.

PR #4974, opened 2021-03-16
add oberweis dairy/icecream shop

m-hue:
Oberweis: local chain of ice cream/dairy shops, 40-ish locations, mostly concentrated in northeast Illinois.
CD One Price Cleaners: Dry cleaning chain with 40-ish locations in the Chicagoland area

### DIFF
--- a/data/brands/shop/dairy.json
+++ b/data/brands/shop/dairy.json
@@ -22,6 +22,19 @@
       }
     },
     {
+      "displayName": "Oberweis",
+      "id": "oberweis-95235d",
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks @m-hue & welcome 👋 

The "CD One Price Cleaners" entry won't actually show up in editors until we have a Wikidata QID for it.  If there are 40+ locations, it should be notable enough to have one, so maybe someone else can add a sufficiently researched page to Wikidata.

PR #4980, opened 2021-03-18
Add some more Australian retail brands

lkernan:
Adds several well known retailers in Australia.


bhousel:
Thanks @lkernan this looks perfect! 👍

PR #4984, opened 2021-03-20
Update car_sharing.json

kjonosm:
- Delete `name` tag (it is meant to name the individual carsharing station)
- Update French location sets to metropolitan France ("fx")

### DIFF
--- a/data/operators/amenity/car_sharing.json
+++ b/data/operators/amenity/car_sharing.json
@@ -5,22 +5,11 @@
   },
   "items": [
     {
-      "displayName": "AutoCool",
-      "id": "autocool-95fe74",
-      "locationSet": {"include": ["fr"]},
-      "tags": {
-        "amenity": "car_sharing",
-        "name": "AutoCool",
-        "operator": "AutoCool"
-      }
-    },
-    {
       "displayName": "Autolib'",
       "id": "autolib-95fe74",
-      "locationSet": {"include": ["fr"]},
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
So why delete ```name``` tag ?

Usually chain-brand share the same name, unless we think each shop should have different name.

LaoshuBaby:
Besides, thanks @kjonosm for modify ```fr``` to ```fx```, it makes data in nsi more exact

kjonosm:
> Usually chain-brand share the same name

This is what `operator` or `brand` is for in that context.

> unless we think each shop should have different name

Shops are rather meant to be tagged with [`amenity=car_rental`](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcar_rental). 

According to [osm wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcar_sharing) `amenity=car_sharing` however is used to tag unattended car sharing stations usually located outside on reserved parking spaces.

If you check current tagging in OSM you will see that these stations are most often named after their location to better identify them.

LaoshuBaby:
If it tends to be not commercial purpose, but is used as a basic facility and focuses on convenience in daily life, it is indeed some site in the public facility network, and it should be named after the site.（In this reason, I guess, ÖPNVKarte have render it in a special icon） 

You are right. I understand you. This network-like car_sharing should have it individual site name as ```name```

But can you explain by the way the pr next door, which is in #4985

PR #4985, opened 2021-03-20
Update charging_station.json

kjonosm:
Delete `name` tag for charging stations.
According to OSM [wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station) it should not be used in that context.
There seems no information gain as `name` just duplicates the `operator` value.

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -225,7 +225,6 @@
         "amenity": "charging_station",
         "brand": "bike-energy",
         "brand:wikidata": "Q67770877",
-        "name": "bike-energy",
         "operator": "bike-energy",
         "operator:wikidata": "Q67770877"
       }
@@ -253,7 +252,6 @@
         "brand": "BKK",
         "brand:wikidata": "Q4891601",
         "brand:wikipedia": "no:BKK (selskap)",
-        "name": "BKK",
         "operator": "BKK",
         "operator:wikidata": "Q4891601",
         "operator:wikipedia": "no:BKK (selskap)"
@@ -276,7 +274,6 @@
         "amenity": "charging_station",
         "brand": "Blink",
         "brand:wikidata": "Q62065645",
-        "name": "Blink",
         "operator": "Blink",
         "operator:wikidata": "Q62065645"
       }
@@ -329,7 +326,6 @@
       "tags": {
         "amenity": "charging_station",
         "brand": "Bosch eBike Systems",
-        "name": "Bosch eBike Systems",
         "operator": "Bosch eBike Systems",
         "operator:wikidata": "Q234021",
         "operator:wikipedia": "en:Robert Bosch GmbH"
@@ -443,7 +439,6 @@
         "brand": "ChargePoint",
         "brand:wikidata": "Q5176149",
         "brand:wikipedia": "en:ChargePoint",
-        "name": "ChargePoint",
         "operator": "ChargePoint",
         "operator:wikidata": "Q5176149",
         "operator:wikipedia": "en:ChargePoint"
@@ -458,7 +453,6 @@
         "brand": "Chargy",
         "brand:wikidata": "Q62702950",
         "brand:wikipedia": "lb:Chargy",
-        "name": "Chargy",
         "operator": "Chargy",
         "operator:wikidata": "Q62702950",
         "operator:wikipedia": "lb:Chargy"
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
According to the OSM Wiki https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station#Examples


Yes, name is sometimes not needed, but if you don't offer a default name, mapper will have to write it manually. And when render, it will also looking for ```name``` key.


And the data now existing is have name, for example https://github.com/osmlab/name-suggestion-index/blob/main/data/operators/amenity/charging_station.json#L2749


If a charging_station have a special name, not the name heritaged from the brand, I guess edit it when mapping is a better way, and let more mapper can auto input the default name of this brand




------------------&nbsp;Original&nbsp;------------------
From: ***@***.***&gt;; 
Date: 2021年3月20日(星期六) 晚上9:53
To: ***@***.***&gt;; 
Cc: ***@***.***&gt;; 
Subject: [osmlab/name-suggestion-index] Update charging_station.json (#4985)




Delete name tag for charging stations.
 According to OSM wiki it should not be used in that context.
 There seems no information gain as name just duplicates the operator value.
 
 
You can view, comment on, or merge this pull request online at:
 
&nbsp; https://github.com/osmlab/name-suggestion-index/pull/4985
 
Commit Summary
    
Update charging_station.json
  
File Changes
    
     M     data/operators/amenity/charging_station.json     (32)   
  
Patch Links:
    
https://github.com/osmlab/name-suggestion-index/pull/4985.patch
   
https://github.com/osmlab/name-suggestion-index/pull/4985.diff

bhousel:
Yeah I agree with @LaoshuBaby , I think it's ok to leave `name` here as an initial seed value, and I'm not comfortable just removing them all.  

I also don't think the wiki really says not to use names, and if it did, I would maybe check the page history and question who wrote that.  I think it's reasonable to expect renderers to use `name` tag for displaying labels on map pins.

- If a user adds a charging station named  "ChargePoint" that is ok.  As far as I can tell, they don't have unique names.
- Some brands like "Tesla Supercharger" do seem to have unique names for the charging stations?
- We can add a `"preserveTags": ["^name"]` property either to individual items, or to the category as a whole, so that the name won't be overridden if a user wants to rename it to something else.

(I need to add some better guidance around `name` to our project documentation.  How we work with the `name` tag has evolved since #4906 and that's only within the past month).

LaoshuBaby:
ahh……
here are some experience about the ration between individual point and chain brand
some times we mapper at china ,when we add a new point, especially it is a bank or telecommunication office, there will be a ref name called “'XXX支行”/“'XXX分理处” like
if we know the ref/alias of this individual point, we will manualy change the key name to “'XX银行 - XXX支行”，but if we don't know we will use the default suggested name instead because at least we should render a name to let people can see this point belongs to which brand.


for another PR i just merged is because I think at that scenario, single sharing station will be in a network and have special ref name.
but those commcial charging station, the different between brand may large, so i think the name should show this difference.



---Original---
From: "Bryan ***@***.***&gt;
Date: Sun, Mar 21, 2021 01:25 AM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [osmlab/name-suggestion-index] Update charging_station.json (#4985)




 
Yeah I agree with @LaoshuBaby , I think it's ok to leave name here as an initial seed value, and I'm not comfortable just removing them all.
 
I also don't think the wiki really says not to use names, and if it did, I would maybe check the page history and question who wrote that.  I think it's reasonable to expect renderers to use name tag for displaying labels on map pins.
  
If a user adds a charging station named  "ChargePoint" that is ok.  As far as I can tell, they don't have unique names. 

Some brands like "Tesla Supercharger" do seem to have unique names for the charging stations? 

We can add a "preserveTags": ["^name"] property either to individual items, or to the category as a whole, so that the name won't be overridden if a user wants to rename it to something else. 
 
(I need to add some better guidance around name to our project documentation.  How we work with the name tag has evolved since #4906 and that's only within the past month).

kjonosm:
> If a user adds a charging station named "ChargePoint" that is ok. As far as I can tell, they don't have unique names.
> Some brands like "Tesla Supercharger" do seem to have unique names for the charging stations?

Tesla Superchargers seem to have a unique name, see their [website](https://www.tesla.com/de_DE/findus?v=2&bounds=52.56830920355766%2C13.603119382812503%2C52.393298750265174%2C13.108734617187503&zoom=12&filters=store%2Cservice%2Csupercharger%2Cdestination%20charger&search=Berlin%2C%20Germany&location=berlineurefcampussupercharger)

image

kjonosm:
> I also don't think the wiki really says not to use names, and if it did, I would maybe check the page history and question who wrote that.

I wasn't precise on this. The wiki doesn't recommend not to use `name`. It rather doesn't mention the `name` tag at all.  context

kjonosm:
Fastned, another European charging station operator also uses unique names for their stations, see their [website](https://fastnedcharging.com/en/locations)

image

bhousel:
I added `"preserveTags": ["^name"]` to the `operators/amenity/charging_station` category in 57aba8bc5 ..

This will allow the `name` tag as an initial value, but users are encouraged to change it, and NSI will not overwrite the custom names that people add.

PR #4988, opened 2021-03-22
Updated some Aus entries

chuqtas:
Add Zone Bowling, fix up countries codes for a few others


LaoshuBaby:
Good！ Thanks @chuqtas for those edits!

Issue #4784, opened 2020-12-07
GeoJSON for states


michaelblyons:
This is a poor-man's extension to https://github.com/ideditor/country-coder/issues/26. Is it worthwhile to make simple GeoJSON shapes that "loose fit" states?

They would be useful for `locationSet.include`, but not really for `exclude`. I'm imagining something like

```json
    {
      "displayName": "Some Regional Bus System",
      "locationSet": {
        "include": [
          "us-wa.lenient.geojson",
          "us-or.lenient.geojson",
          "us-id.lenient.geojson"
         ]
      },
      "tags": {
        "network": "Some Regional Bus System",
        "route": "bus"
      }
    },
```

It'd be really helpful, I imagine, for doing [state police](https://en.wikipedia.org/wiki/State_police_(United_States)#List_of_state_police_agencies). Also, the country-code (text) search in NSI.guide would start working for locations tagged that way.

UKChris-osm:
If we were going to go down the "_*.geojson_" road for States (and UK Counties 👍), could this data not be pulled from OSM's current boundaries, rather than loosely drawn?

michaelblyons:
> If we were going to go down the "_*.geojson_" road for States (and UK Counties 👍), could this data not be pulled from OSM's current boundaries, rather than loosely drawn?

If there's a way to do this already, I'm all for it (👍). But the library used for NSI (if I understand correctly) does not have ISO 3166-2 codes since it would mean 10x-20x more shapes.

I presume the library would want to be strictly accurate. But here, I don't care if there's some overlap in adjoining shapes. Limiting things loosely, with simple 4-to-16-point shapes is "good enough."

UKChris-osm:
Sorry for being vague - I meant couldn't we pull the boundary data (such as [Florida](https://www.openstreetmap.org/relation/162050)) and save them as our own "_*.geojson_" features for use within the NSI?

bhousel:
Just catching up on this.. 

> It'd be really helpful, I imagine, for doing state police. 

Yes for state police it would be useful.. Unioning the geometries together in the `include` is not very performant, but at least it is cached.  For the case of a regional bus line, it would be much more performant to represent it as a rough polygon or box than to union 3 state polygons together.

> Also, the country-code (text) search in NSI.guide would start working for locations tagged that way.

Ah yeah I need to fix that (#4077) - we shouldn't design the index around that country code search box. 

[...]

PR #4990, opened 2021-03-24
Hanmibank

m-hue:
Bank with locations in CA, IL, NJ, NY, TX and VA.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -5051,6 +5051,18 @@
       }
     },
     {
+      "displayName": "Hanmi Bank",
+      "id": "hanmibank-ea2e2d",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "amenity": "bank",
+        "brand": "Hanmi Bank",
+        "brand:wikidata": "Q5648633",
+        "brand:wikipedia": "en:Hanmi Bank",
+        "name": "Hanmi Bank"
+      }
+    },
+    {
       "displayName": "HBL Bank (ایچ بی ایل پاکستان)",
       "id": "hblbank-da5806",
       "locationSet": {"include": ["pk"]},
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
I trying to find more exact states boundry in country coder but failed.

So I think this is exact enough.

LaoshuBaby:
Thanks @m-hue for improving NSI continuosly!

LaoshuBaby:
Can this locationSet suitable for using geojson to describe?

(Because just received the comment notice for #4784

bhousel:
> Can this locationSet suitable for using geojson to describe?

We don't have state-level boundaries yet, but scoping it to [`us`](https://ideditor.codes/?code=us) (or even [`conus`](https://ideditor.codes/?code=conus)) is ok for now. 

Hanmibank might expand to more states, and it would be extra work for us to maintain a geojson of all the states where its branches are.

LaoshuBaby:
> 
> 
> > Can this locationSet suitable for using geojson to describe?
> 
> We don't have state-level boundaries yet, but scoping it to [`us`](https://ideditor.codes/?code=us) (or even [`conus`](https://ideditor.codes/?code=conus)) is ok for now.
> 
> Hanmibank might expand to more states, and it would be extra work for us to maintain a geojson of all the states where its branches are.

So since there is currently no geojson file for continents, it seems that I have been worrying too much. Thank @bhousel  for your answer! 👍

bhousel:
> So since there is currently no geojson file for continents, it seems that I have been worrying too much. Thank @bhousel for your answer! 👍

yes, for `locationSet` we can support:
- codes recognized by country-coder - https://ideditor.codes
- circle region defined as a point and radius
- custom .geojson file

(more info here: https://github.com/ideditor/location-conflation/blob/main/README.md )

It's somewhat expensive to download or bundle a lot of geojsons, and also extra work for us to maintain them, so I think of it as a fallback option.. They are very useful for sub-country regions like Crimea, or French-speaking Canada, or regions where transit companies operate, etc.  We also have a few situations in the US where there are separate regional brands using the same name ("Price Chopper", "Citizens Bank").

PR #4993, opened 2021-03-25
Merge YAMADA Denki, YAMADA

venue-map:
Corrected the item of YAMADA Denki in Japan.


LaoshuBaby:
Thanks @venue-map  for make NSI more exact！

PR #4995, opened 2021-03-26
Australian IGA stores

lkernan:
Separate the Australian IGA from the USA definition.   Add Ritchies which while an IGA franchisee is large enough to have its own branding.


LaoshuBaby:
Thanks @lkernan  for making location based data in NSI more exactly! 👍

PR #4997, opened 2021-03-27
Add more Australian chains

lkernan:
Mycar - formerly Kmart tyre and auto.
Domayne 
Best & Less 
Rivers 
Lincraft


UKChris-osm:
Thank you @lkernan 👍

PR #5000, opened 2021-03-28
Update estate_agent.json

lkernan:
Add major estate agents in Melbourne Australia

### DIFF
--- a/data/brands/office/estate_agent.json
+++ b/data/brands/office/estate_agent.json
@@ -130,6 +130,17 @@
       }
     },
     {
+      "displayName": "Buxton",
+      "id": "buxton-d66f4e",
+      "locationSet": {"include": ["au"]},
+      "tags": {
+        "brand": "Buxton",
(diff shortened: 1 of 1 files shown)


data/brands/office/estate_agent.json
```diff
@@ -129,6 +129,17 @@
         "office": "estate_agent"
       }
     },
+    {
+      "displayName": "Buxton",
+      "id": "buxton-d66f4e",
+      "locationSet": {"include": ["au"]},
```
UKChris-osm, line 135: Is Buxton only in Victoria? If so, you could use "_au-vic.geojson_"

UKChris-osm:
Thank you @lkernan 👍

data/brands/office/estate_agent.json
lkernan: Thanks, I'll keep that in mind from now on.

PR #5005, opened 2021-04-02
Add EV charging and hydrogen network

chuqtas:
Note: True Zero is a Californian hydrogen network operator.  Should "fuel:h70=yes" be added to the tags to ensure this is always included with locations associated with the brand?

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -4128,6 +4128,16 @@
       }
     },
     {
+      "displayName": "True Zero",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "amenity": "fuel",
+        "brand": "True Zero",
+        "brand:wikidata": "Q106320436",
+        "name": "True Zero"
+      }
+    },
+    {
       "displayName": "Turkey Hill",
       "id": "turkeyhill-c5cf43",
       "locationSet": {"include": ["us"]},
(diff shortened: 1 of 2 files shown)


bhousel:
> Should "fuel:h70=yes" be added to the tags to ensure this is always included with locations associated with the brand?

Yes I think that's a good idea..

I think it might also be helpful to tag that these fuel stations don't sell gasoline, but not sure how to do it.  I checked the OSM Wiki [key:fuel](https://wiki.openstreetmap.org/wiki/Key:fuel) page but didn't see any tag like `fuel:octane=no`.  I'll ask on Slack.

bhousel:
> I think it might also be helpful to tag that these fuel stations don't sell gasoline, but not sure how to do it. I checked the OSM Wiki [key:fuel](https://wiki.openstreetmap.org/wiki/Key:fuel) page but didn't see any tag like `fuel:octane=no`. I'll ask on Slack.

per @matkoniecz,  the `fuel:=*` tags are assumed to be exhaustive, so adding `fuel:h35=yes` and `fuel:h70=yes` tags to these stations should be enough to imply that other types are not available.

matkoniecz:
Disclaimer: I only repeated what I learned in https://github.com/streetcomplete/StreetComplete/issues/2708 and made no own research into that tagging.

I posted https://wiki.openstreetmap.org/wiki/Talk:Key:fuel#Is_tagging_of_fuel%3A_assumed_to_be_exhaustive.3F right now

chuqtas:
Thanks @bhousel and @matkoniecz .  I've added "fuel:h70=yes" to the PR.  From what I can see some of the stations might not support h35 so I'll leave that out for the time being (can be added later if needed).

bhousel:
Thanks @chuqtas  👍

PR #5009, opened 2021-04-06
Australian malls

lkernan:
Fix copy paste error on Vicinity Centres brand
Remove name from Vicinity and GPT, every mall has an individual name.

Change Scentre Group brand to Westfield (Australia) and make Scentre Group the operator.
Most people on the street will never have heard of Scentre group and Westfield is what appears on the buildings. 
Trouble is there isn't a current Westfield Australia wikidata I can link brand to. The one available is the "old" Westfield before it was split between Australia and International operations.
Open to advice on how to improve this.

### DIFF
--- a/data/brands/shop/mall.json
+++ b/data/brands/shop/mall.json
@@ -28,6 +28,7 @@
         "brand": "GPT Group",
         "brand:wikidata": "Q3100507",
         "brand:wikipedia": "en:GPT Group",
+        "name": "GPT Group",
         "shop": "mall"
       }
     },
@@ -55,6 +56,7 @@
         "brand": "Vicinity Centres",
         "brand:wikidata": "Q1054327",
         "brand:wikipedia": "en:Vicinity Centres",
+        "name": "Vicinity Centres",
         "shop": "mall"
       }
     },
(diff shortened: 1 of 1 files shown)


data/brands/shop/mall.json
```diff
@@ -28,19 +28,6 @@
         "brand": "GPT Group",
         "brand:wikidata": "Q3100507",
         "brand:wikipedia": "en:GPT Group",
-        "name": "GPT Group",
```
LaoshuBaby, line 31: I don't think this is a good idea to remove ```"name"```

According to the discussion by @bhousel in #4985 , at the head of this json have a ```"preserveTags": ["^name"],```, so ```"name"``` are encouraged to changed by user. (You can see [this comment](https://github.com/osmlab/name-suggestion-index/pull/4985#issuecomment-803594165))

so we just offer a name to be rendered and let user know what is it —— a shopping mall, and if they know, they can use a special and specific name to rewrite the ```“name”```

LaoshuBaby (changes requested):
So in my opinion, the ```name``` should be added to pass and merged to main branch.

data/brands/shop/mall.json
lkernan: Ok, i've added the name tags back.   
The maill.josn file already has a preserveTags at the top so should be ok.

PR #5010, opened 2021-04-07
Updates on bank brands in Poland

kubahaha:
Hello,
I reviewed some banks avaible in presets for Poland. I made following changes:
1. Since 30 Mar 2019 BGŻ BNP Paribas Bank in Poland changed brand name. Apart of that I'm excluding global BNP Paribas bank from Poland.
2. Raiffeisen is gone from Poland. They were present under name od Raiffeisen Polbank.
3. PEKAO SA is duplicate of Bank Pekao which is better suited name.
4. Eurobank is gone.

Those changes are part of my [automated edit proposal](https://wiki.openstreetmap.org/wiki/Automated_edits/kubahahaha_-_bot_account/Banks_in_Poland#Edit_NSI_templates) and [has been discussed](https://forum.openstreetmap.org/viewtopic.php?pid=824085) on polish OSM forum.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -4317,18 +4317,6 @@
       }
     },
     {
-      "displayName": "Eurobank (Poland)",
-      "id": "eurobank-68054b",
-      "locationSet": {"include": ["pl"]},
-      "tags": {
-        "amenity": "bank",
-        "brand": "Eurobank",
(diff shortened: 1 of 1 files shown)


bhousel:
> 3\. PEKAO SA is duplicate of Bank Pekao which is better suited name.

for things like this you will want to add a `matchName` to match the old name.

PR #5011, opened 2021-04-07
Revert "Updates on bank brands in Poland" 还是太草率了！！

LaoshuBaby:
Reverts osmlab/name-suggestion-index #5010 

我还是太草率了！！

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -2601,15 +2601,15 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


bhousel:
no worries, this is really ok and I dont think you need to revert the whole thing.

We can just add `matchName` or discard filters if we run the build and notice that "Pekao SA" or "BGŻ BNP Paribas" or "Eurobank" comes back.

PR #5013, opened 2021-04-09
Add ETSAN

Robot8A:


data/brands/shop/supermarket.json
```diff
@@ -8327,6 +8327,18 @@
...
       }
+    },
+    {
+      "displayName": "ETSAN",
+      "locationSet": {"include": ["at"]},
+      "tags": {
+        "brand": "Etsan",
+        "brand:wikidata": "Q25916921",
+        "brand:wikipedia": "de:ETSAN",
+        "diet:halal": "only",
```
LaoshuBaby, line 8338: https://wiki.openstreetmap.org/wiki/Key:diet

At first  I was confused with this tag because this is a supermarket not a restaurant.

But After search the osmwiki, This is ok to tag a supermarket with ```diet```

LaoshuBaby (approved):

PR #5015, opened 2021-04-10
Added more readable date stamp to footer

UKChris-osm:
Added more readable date stamp to footer:
nsi-datestamp


bhousel:
I dunno, I kind of like the minimal one better - we don't really need seconds and timezones.  But I'll merge it anyway.

PR #5025, opened 2021-04-18
Update restaurant.json

kjonosm:
Treat "China Express" as a non brand name.


LaoshuBaby:
https://www.topekachinaexpress.com/
http://www.chinaexpressnc.com/
http://www.chinaexpressut.com/
👆
Here, just searched with Baidu

Although I even haven't heard those name of a brand naming China, and I guess it likes Panda express, is just a western-Chinese-rustaurant, and looking for those pages I think it is singal shop, not chain brand.

But the remove for the ```China Express``` should wait for more detailed search.

------
By the way, there is a joke, when I just see the abstract of email notice, I just think "China Express" is a post office like China Post
:(

PR #5028, opened 2021-04-19
Cooper's Hawk

m-hue:
american restaurant & wine store chain, locations in the midwest, florida, arizona, and some on the east coast.

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -874,6 +874,20 @@
       }
     },
     {
+      "displayName": "Cooper's Hawk Winery & Restaurants",
+      "id": "coopershawkwineryandrestaurants-96af40",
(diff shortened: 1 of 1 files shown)


m-hue:
good catch. Copy/Paste fail. Sorry about that. fixed.

kjonosm:
Sorry for nitpicking but do they really serve American cuisine? Their menu online looks quite international to me...

LaoshuBaby (commented):
https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/shop/wine.json

For this brand, treat is as a restaurant or just a wine shop?

bhousel:
> For this brand, treat is as a restaurant or just a wine shop?

It is weird, but it looks like it is actually both..

thanks @m-hue , I'll merge 👍

PR #5031, opened 2021-04-22
Added several Dutch stores

RobinLinde:
Not 100% sure about capitalization, they're now matched to the official websites.

### DIFF
--- a/data/brands/shop/storage_rental.json
+++ b/data/brands/shop/storage_rental.json
@@ -18,6 +18,17 @@
       }
     },
     {
+      "displayName": "ALLSAFE Mini Opslag",
+      "id": "allsafeminiopslag-43269c",
+      "locationSet": {"include": ["nl"]},
+      "tags": {
+        "brand": "ALLSAFE",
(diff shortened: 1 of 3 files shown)


data/brands/shop/storage_rental.json
```diff
@@ -17,6 +17,17 @@
         "shop": "storage_rental"
       }
     },
+    {
+      "displayName": "ALLSAFE Mini Opslag",
```
Identitaet, line 21: isn't mini Opslag more of a slogan and not part of the name ?

RobinLinde: To be honest they do use it as title for their website and their listings on Google Maps, but most people would refer to it simply as Allsafe. As a slogan they tend to use `Geeft ruimte`/`Provides space`. Based on that I'd think it's an appropriate display name.

bhousel:
Thanks @RobinLinde 👍 
If you get a chance, it might be good to add some supporting references to the Wikidata pages that you made, to make sure they aren't deleted.

PR #5033, opened 2021-04-22
Excluding several brands in Poland

kubahaha:
Hello, I excluded some brands not present in reality in Poland, but available via templates.


LaoshuBaby:
Hey! Thanks for making the NSI more exactly, but if somedays it have a brand in Poland, hope that there will have a contributor add entiries for Poland.

kubahaha:
Changes happen :smiley:
I just wanted templates to reflect reality. Is such exclusion something unwanted?

LaoshuBaby:
> 
> 
> Changes happen 😃
> I just wanted templates to reflect reality. Is such exclusion something unwanted?

没，我不是这个意思，我是觉得挺好的

只要未来能持续得到维护就行

kubahaha:
Thanks for merging. I will try to maintain in the future, but it's cannot guarantee what will happen after couple of years.

LaoshuBaby:
Ah, don’t worry too much, we can do what we can do.


 I currently live in China, but I still failed to maintain all the items with location as "cn". 



So, if a few years later, walking on the street, found a new brand, just come to NSI repo in time and write it down according to the facts. I think so.




------------------&nbsp;Original&nbsp;------------------
From: ***@***.***&gt;; 
Date: 2021年4月23日(星期五) 凌晨0:07
To: ***@***.***&gt;; 
Cc: ***@***.***&gt;; "State ***@***.***&gt;; 
Subject: Re: [osmlab/name-suggestion-index] Excluding several brands in Poland (#5033)





 
Thanks for merging. I will try to maintain in the future, but it's cannot guarantee what will happen after couple of years.

PR #5036, opened 2021-04-25
Add Fanta to Beverages

dianacpg:
Just added "Fanta" brand to beverages.

### DIFF
--- a/data/brands/shop/beverages.json
+++ b/data/brands/shop/beverages.json
@@ -50,6 +50,17 @@
       }
     },
     {
+      "displayName": "Fanta",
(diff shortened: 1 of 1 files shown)


kjonosm:
The items under `shop=beverages` are rather about actual shops that sell beverages than about individual brands of drinks. Therefore `Fanta` wouldn't really fit in here. Though I might be unaware of `Fanta` running their own shops or of shops under the `Fanta` brand. Is that the case and can you provide some examples?

dianacpg:
Thank you for your response and you are are right. I misunderstood this section, i'm sorry! I will be more careful next time.

PR #5039, opened 2021-04-25
German computer stores

Contrib1043:
Hey guys, added three German computer stores..

### DIFF
--- a/data/brands/shop/computer.json
+++ b/data/brands/shop/computer.json
@@ -5,6 +5,17 @@
   },
   "items": [
     {
+      "displayName": "ARLT Computer",
+      "id": "arltcomputer-8969c7",
+      "locationSet": {"include": ["de"]},
+      "tags": {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Hi @Contrib1043 and thank you - could you add the "_brand:wikipedia_" tag for each entry as well?

LaoshuBaby:
> 
> 
> Hi @Contrib1043 and thank you - could you add the "_brand:wikipedia_" tag for each entry as well?

Not must need, we can use ```npm run wikidata``` to auto spider it.

LaoshuBaby:
Have a ```wikipedia``` tag is better, but it is not necessary. We can use script to fill it.(But the language is en first)

But remove already exist ```wikipedia``` is not advocated, unless the old one is wrong. I think.

(If there are something inappropriate I've made, please point it! Thanks you!)

We can wait @Contrib1043 for more detailed futher information.

Issue #4958, opened 2021-03-11
Shopping malls


lkernan:
Wondering if it's possible to add shop=mall as an option?   
https://wiki.openstreetmap.org/wiki/Tag:shop%3Dmall

Here in Australia there are two major mall operators I can think of, plus several smaller ones.   
Scentre group (wikidata Q18378048) trading as Westfield
Vicinity Centres (wikidata Q1054327)

I noticed that Westfield in the US/UK/etc (Unibail-Rodamco-Westfield) is currently listed under department_stores which doesn't seem right to me.  It's a mall operator, not a store such as Macy's.

1ec5:
Westfield was originally added as a `shop=mall` in #2735, but 671b85f564906f0eb3e79ff6b693dea84afc4a0c removed it. 82fda4ece49824fa610a1f35274d8603979f1d91 added a different Westfield entry as a `shop=department_store` with possibly problematic wiki tags. I think the department store entry should be removed, but I’m not sure if the circumstances that led to the `shop=mall` entry’s removal are still relevant.

/cc @bhousel

lkernan:
As some more examples:
Direct Factory Outlets (Q5280270) also known as DFO
GPT Group (Q3100507)

bhousel:
Thanks - I added these in 5ce3c226f 
That closed the issue, but I did want to put out there - should I maybe move these to the `operators` tree?
The name-suggestion-index contains:
- things that are definitely brands
- things that are definitely operators
- things that are both, but more brandlike
- things that are both, but more operatorlike
- amenity/bicycle_rental 🤷

Issue #2743, opened 2019-06-02
Westfield is still a mall


1ec5:
#2735 added a `shop=mall` entry for Westfield, which is still a mall brand, even though it’s owned by two different companies that are no longer called Westfield. See [this Wikipedia list](https://en.wikipedia.org/wiki/List_of_Westfield_shopping_centres) and its sublists. The current entry is tagged with a Wikidata item and Wikipedia article for a defunct company whose successor is also defunct.

A Wikidata item needs to be created for the Westfield brand, linked to the current parent companies, and the NSI entry needs to be switched over to the new item. Alternatively, separate Wikidata items can be created for the Scentre Group properties and Unibail-Rodamco-Westfield properties, limited to different sets of countries.

bhousel:
In the mean time, I've removed the wiki tags in ea284122

andrewharvey:
See also my comment at https://github.com/osmlab/name-suggestion-index/commit/ea284122267ef828070d0271cb9377ee22bc7ec8 regarding omitting the name tag.

bhousel:
We can't omit the name tag.

bhousel:
I removed the entry in 671b85f5
We are currently [filtering out](https://github.com/osmlab/name-suggestion-index/blob/671b85f564906f0eb3e79ff6b693dea84afc4a0c/config/filters.json#L32) `shop/mall` from the names we extract from the planet, and this single entry didn't have suitable wiki tags anyway.    

We can revisit this later, but I just want to remove it for right now.

1ec5:
For future reference, [this Wikipedia search](https://en.wikipedia.org/w/index.php?title=Special:Search&search=%22chain+of+malls%22) turns up several more chains of malls, though Westfield is the largest and most strongly branded of them.

PR #5041, opened 2021-04-26
Update mall.json

lkernan:
Add Westfield brand.

### DIFF
--- a/data/brands/shop/mall.json
+++ b/data/brands/shop/mall.json
@@ -81,6 +81,8 @@
       "locationSet": {"include": ["au", "nz"]},
       "tags": {
         "brand": "Westfield",
+        "brand:wikidata": "Q18378979",
+        "brand:wikipedia": "en:Westfield Group",
         "name": "Westfield",
         "operator": "Scentre Group",
         "operator:wikidata": "Q18378048",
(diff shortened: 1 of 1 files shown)


data/brands/shop/mall.json
```diff
@@ -81,6 +81,8 @@
       "locationSet": {"include": ["au", "nz"]},
       "tags": {
         "brand": "Westfield",
+        "brand:wikidata": "Q18378979",
+        "brand:wikipedia": "en:Westfield Group",
```
1ec5, line 85: This Wikipedia article and Wikidata item are about a former company. The two sites are littered with the remnants of former iterations of the company. To confuse things further, I created [this item](https://www.wikidata.org/wiki/Q106623253) to represent the brand per se, which can be added to both this NSI entry and the Euro-American one above.

lkernan: Thanks.  I wasn't sure if that was an appropriate solution or not from the Wikidata perspective. I'll update the PR to use that instead.

bhousel:
Thanks @lkernan & @1ec5 👍

bhousel:
related #4958 and #2743

PR #5042, opened 2021-04-26
Update parking.json

lkernan:
Add Care Park.  Major car park operator in Australia, NZ and apparently also operates in Malaysia.


bhousel:
Thanks @lkernan 👍

PR #5050, opened 2021-04-27
Update train.json

lkernan:
Add wikidata and wikipedia for Sydney Trains and Translink


bhousel:
Thanks @lkernan , looks great 👍 
Only thing I would ask is please try to write a commit message that is more descriptive than "update train.json".
(The pr comment "Add wikidata and wikipedia for Sydney Trains and Translink" is perfect!)

PR #5053, opened 2021-05-02
Remove Liberty fuel from Australia

lkernan:
We have a petrol station called Liberty Oil in Australia, but it's not the one described by this entry.  
The Australian one doesn't have a serious online presence to be able to tag them correctly at the moment.  This will avoid confusing editors with the wrong wikidata.


LaoshuBaby:
Great! I will merge this.

(But if possible, will you add the Australian's Liberty Oil to NSI? )

PR #5054, opened 2021-05-03
Remove wikidata entry from Nelson Alexander.

lkernan:
Someone on Wikidata decided it was "nonsensical" and not notable enough.  
I didn't see the alert in time to respond so it has been deleted.

### DIFF
--- a/data/brands/office/estate_agent.json
+++ b/data/brands/office/estate_agent.json
@@ -459,7 +459,6 @@
       "locationSet": {"include": ["au"]},
       "tags": {
         "brand": "Nelson Alexander",
-        "brand:wikidata": "Q106236053",
         "name": "Nelson Alexander",
         "office": "estate_agent"
       }


bhousel:
Oh in situations like this we can first raise it on the administrator noticeboard and ask to have the Q106236053 page restored.
https://www.wikidata.org/wiki/Wikidata:Administrators%27_noticeboard

Nelson Alexander is an Australia based real estate agent that has been around since 1971.. They have 16 offices, which is on the low side for NSI, but I still think it would meet Wikidata's notability requirements.

This page can serve as a reference for the corporate details:
https://www.zoominfo.com/c/nelson-alexander/74646654

bhousel:
Just following up..
I [opened a request](https://www.wikidata.org/wiki/Wikidata:Administrators%27_noticeboard#Request_for_undeletion_of_Q106236053) on the Wikidata Admin noticeboard, and they restored the [Nelson Alexander page](https://www.wikidata.org/wiki/Q106236053)  👍

Regular reminder to make sure that any pages that our community creates should contain substantial references to avoid being deleted.  References to third parties like news sources are best (i.e. not just a reference to the company's own website).

PR #5066, opened 2021-05-10
Add BrewDog to amenity/pub.json

rskedgell:


kjonosm:
They are also present in Germany

bhousel:
I pulled the list of countries where they are located from their website and updated the item...

PR #5067, opened 2021-05-10
Added Penti

evrifaessa:
added Penti, a Turkish brand for women. has stores in most of the malls in the country.


bhousel:
Thanks @evrifaessa and welcome! 👋

PR #5071, opened 2021-05-11
Added some more Turkish stuff

evrifaessa:


LaoshuBaby:
> I may add some more, I'm tired though.

OK, so this PR will open for 24 hours? Is that long enough

If you really have a lot of items that need to be added, I guess you can contact @bhousel to add permissions of this repo for you？

evrifaessa:
> OK, so this PR will open for 24 hours? Is that long enough

that's OK.

PR #5073, opened 2021-05-12
Removed duplicate of Garanti

evrifaessa:
Wanted to create a PR instead of directly commiting because I'm not sure if it's OK to change display names.


LaoshuBaby (approved):
大胆一点，没问题的👍

bhousel:
Looks good to me too, It's always ok to ask if you're not sure 👍

PR #5074, opened 2021-05-15
Tesco Lotus to Lotus's

MASangys:
Lotus's rebranding in Malaysia and Thailand by end of the year.
https://asia.nikkei.com/Business/Retail/Tesco-disappears-from-Southeast-Asia-as-CP-renames-outlets

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -5615,14 +5615,14 @@
       }
     },
     {
-      "displayName": "Tesco Lotus",
+      "displayName": "Lotus's",
       "id": "tescolotus-fd4d8d",
-      "locationSet": {"include": ["th"]},
+      "locationSet": {"include": ["my", "th"]},
       "tags": {
-        "brand": "Tesco Lotus",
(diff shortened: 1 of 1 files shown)


bhousel:
If this is a brand in transition, I think we should preserve the current/old with a tag `alt_name=Tesco Lotus`.  This will let NSI continue to match on current name, and also will let users find the features in geocoders like Nomanitum that index `alt_name`.

MASangys:
Can, therefore you may add Tesco Lotus for alt name
but this alt name only applicable to Thailand branch, Malaysia's branches are priginally named Tesco before rebranding tho.

LaoshuBaby (changes requested):
If one name only apply to a special conutry/region, you'd better seperate it into 2 different item.

MASangys:
> If one name only apply to a special conutry/region, you'd better seperate it into 2 different item.

But after rebranding, only 1 item left for both countries.

MASangys:
Hello, any respond?

bhousel:
Thanks @MASangys, this is ok 👍

PR #5078, opened 2021-05-18
University of Maryland, College Park

emanbau:
Added UMD to the university amenities! Hopefully this small change is meaningful!

### DIFF
--- a/data/brands/amenity/university.json
+++ b/data/brands/amenity/university.json
@@ -43,6 +43,18 @@
         "name": "University of Phoenix",
         "short_name": "UOPX"
       }
+    },
+    {
+      "displayName": "University of Maryland, College Park",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "amenity": "university",
+        "brand": "University of Maryland, College Park",
+        "brand:wikidata": "Q503415",
+        "brand:wikipedia": "en:University of Maryland, College Park",
+        "name": "University of Maryland, College Park",
+        "short_name": "UMD"
+      }
     }
   ]
 }


data/brands/amenity/university.json
```diff
@@ -43,6 +43,18 @@
...
         "short_name": "UOPX"
       }
+    },
+    {
+      "displayName": "University of Maryland, College Park",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "amenity": "university",
+        "brand": "University of Maryland, College Park",
+        "brand:wikidata": "Q503415",
```
LaoshuBaby, line 53: A university is a chain brand is very special, and I just looked for the wikidata ```Q503415```, it refer to a single university of Maryland, but ```Q1480532``` can refer to the system of a series of Univ. Maryland.

>It is a pity that there is no such kind of chain university in my area, so I am afraid I can't understand why there is this type of data.
>And the University of California, Berkeley, Santa Barbara, these are obviously a series of appearances, and they are not included in the university. Therefore, I am very cautious about adding data to this file, amenity/university.json.

LaoshuBaby (changes requested):

bhousel:
Sorry @emanbau , but we don't add individual universities to this index - we only add the kind of for-profit "universities” that operate like a branded business in multiple locations, such as DeVry or University of Phoenix. I know this is an unusual concept outside of the USA - the “universities” in this index are really more like training centers than true universities.

This index is really for tracking the types of things where there are a lot of them, not just single things. We are considering adding some university operated assets in #4779

emanbau:
@bhousel thank you for the clarification! I'll continue to look for things to add to help out!

PR #5082, opened 2021-05-24
Dänisches Bettenlager changed branding to JYSK in Austria

Contrib1043:
Dänisches Bettenlager changed branding to JYSK in Austria (only in Germany remains with old brand)


data/brands/shop/clothes.json
```diff
@@ -319,6 +319,18 @@
         "name": "Ann Taylor",
         "shop": "clothes"
       }
+    },
```
LaoshuBaby, line 322: Those part have been added in #5083 
So I will just remove it in #5082

data/brands/shop/furniture.json
```diff
@@ -321,7 +321,7 @@
     {
       "displayName": "Dänisches Bettenlager",
       "id": "danischesbettenlager-bc2533",
-      "locationSet": {"include": ["at", "de"]},
+      "locationSet": {"include": ["de"]},
```
LaoshuBaby, line 324: So why not modify the locationSet of JYSK with a ```exclude```

PR #5084, opened 2021-05-24
Add Benny & Co

bbigras:
I'm not sure if the name should have a "dot" at the end or not.﻿

I think "co" is an abbreviation.

Note that I created the wikidata page today.

Also note that "poutine" is not the main theme there. It's a chicken restaurant that also do poutine.

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -253,6 +253,18 @@
       }
     },
     {
+      "displayName": "Benny & Co.",
+      "locationSet": {"include": ["ca"]},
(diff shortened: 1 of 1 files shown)


bhousel:
> I'm not sure if the name should have a "dot" at the end or not.

Yes, I think it should have the dot - this is how the brand describes themselves in their website copy:

Screen Shot 2021-05-24 at 7 15 08 PM

bbigras:
Thanks @bhousel. I fixed it. and the wikidata page.

bbigras:
Thanks!

PR #5085, opened 2021-05-25
Tweak tag

MASangys:
Donki, fix miniso and add daiso for malaysia region

### DIFF
--- a/data/brands/shop/variety_store.json
+++ b/data/brands/shop/variety_store.json
@@ -42,6 +42,50 @@
         "shop": "variety_store"
       }
     },
+    {{
+  "properties": {
+    "path": "brands/shop/variety_store",
+    "exclude": {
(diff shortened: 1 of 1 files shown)


data/brands/shop/variety_store.json
```diff
@@ -42,7 +42,876 @@
         "shop": "variety_store"
       }
     },
+    {{
+  "properties": {
```
LaoshuBaby, line 46: Sorry, but you put too much wrong information

I guess maybe you just paste the whole document in that place

LaoshuBaby (changes requested):

data/brands/shop/variety_store.json
MASangys: Ikr, something went wrong

PR #5087, opened 2021-05-25
Update Miniso, Daiso and Don Don Donki, remove mydin

MASangys:
see

### DIFF
--- a/data/brands/shop/variety_store.json
+++ b/data/brands/shop/variety_store.json
@@ -127,6 +127,18 @@
       }
     },
     {
+      "displayName": "Daiso",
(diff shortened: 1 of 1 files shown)


bhousel:
Why did you remove mydin?

MASangys:
> Why did you remove mydin?

Originally, my research is there is no MyMydin and change to Mydin. But after further investigation, Mydin should be only supermarket although some of them might only have a small place not like in a complex.

bhousel:
> But after further investigation, Mydin should be only supermarket although some of them might only have a small place not like in a complex.

oh ok, thanks for clarification - I do see it also in the supermarket.json file.

PR #5091, opened 2021-05-26
Add rapid kl for mrt lines

MASangys:


bhousel:
lets discuss on #5092 (we can do all the Rapid KL in the same pull request)

PR #5092, opened 2021-05-26
add rapid kl

MASangys:


bhousel:
I would think either Rapid KL ([Q3271384](https://www.wikidata.org/wiki/Q3271384)) or Rapid KL Light Rail ([Q987598](https://www.wikidata.org/wiki/Q987598)) would be the network?  
Under the transit tree the `network` key is the defining key, not the `operator`.

MASangys:
> I would think either Rapid KL ([Q3271384](https://www.wikidata.org/wiki/Q3271384)) or Rapid KL Light Rail ([Q987598](https://www.wikidata.org/wiki/Q987598)) would be the network?
> Under the transit tree the `network` key is the defining key, not the `operator`.

Nope, RapidKL is operator only, but network is still the lines that in the official.metro maps

bhousel:
> Nope, RapidKL is operator only, but network is still the lines that in the official.metro maps

I think there is a mismatch though - we are considering "network" as the transit company identity and source of the logo for the transportation network.  RapidKL may be both the operator and the network.  But "Klang Valley Integrated Transit System" is just a page to tie all of the various lines on a single map.  It's not the network in the OSM sense.

kjonosm:
> But "Klang Valley Integrated Transit System" is just a page to tie all of the various lines on a single map. It's not the network in the OSM sense.

According to [wikipedia](https://en.wikipedia.org/wiki/Klang_Valley_Integrated_Transit_System), "Klang Valley Integrated Transit System" seems more than just a map. It looks like a proper transit system with integrated ticketing run by multiple operators. In that sense I think it should fit well into the `network` category. 

As the train companies of the network operate within the same area, it currently seems not possible to list each of them differentiated by their respective locationSet.

Hence I would suggest to only include one entry for "Klang Valley Integrated Transit System" without operator tags.

bhousel:
This the same issue we had on #4496 where somebody had created a "Philadelphia Subway" wiki page and wikidata to tie together the disparate transit systems in the city. Even if they use common fare collection or ticketing system, that's not what a network is.

"Klang Valley Integrated Transit System" does not have a website.  It doesn't have a logo.  It doesn't have a GTFS feed.  It could maybe be a superrelation that ties together the actual networks underneath it.

kjonosm:
hmm, #4496 looks like a mismatch between the network's name and its wikidata. For all entries in nsi, the display name should of course be named after the network and not after the operator, as this pr wrongly suggests.

I see your point though @bhousel. Over here in Germany if you buy a public transport ticket you don't really care about which operator runs the train or bus - You just buy a network ticket for your area. For these circumstances osm tagging and handling in nsi works very well together. I think that there are similar schemes across Europe.

How is the situation in the US? A lot of entries suggest that there are no such integrated networks because the network tag seems to be used for just adding the operator's short name or brand name, which is understandable if you follow the idea of the `network` tag being the tag that matters for mappers/users (and for the nsi).

@MASangys: What is the situation over in KL?  If it is a network with an operator-shared ticketing, maybe (idk) shared tracks and shared public transport stations and probably a coordinated timetable, I would assume it should be mapped as such by using the `network` tag even if it's not advertised with a dedicated website or logo. Operators within the network can easily be tagged with `operator`. I don't think there is a need for an additional super-relation.

MASangys:
> hmm, #4496 looks like a mismatch between the network's name and its wikidata. For all entries in nsi, the display name should of course be named after the network and not after the operator, as this pr wrongly suggests.
> 
> I see your point though @bhousel. Over here in Germany if you buy a public transport ticket you don't really care about which operator runs the train or bus - You just buy a network ticket for your area. For these circumstances osm tagging and handling in nsi works very well together. I think that there are similar schemes across Europe.
> 
> How is the situation in the US? A lot of entries suggest that there are no such integrated networks because the network tag seems to be used for just adding the operator's short name or brand name, which is understandable if you follow the idea of the `network` tag being the tag that matters for mappers/users (and for the nsi).
> 
> @MASangys: What is the situation over in KL? If it is a network with an operator-shared ticketing, maybe (idk) shared tracks and shared public transport stations and probably a coordinated timetable, I would assume it should be mapped as such by using the `network` tag even if it's not advertised with a dedicated website or logo. Operators within the network can easily be tagged with `operator`. I don't think there is a need for an additional super-relation.

In KL, there are 3 operators
-KTM for suburban rail
-LRT for light metro (Own and operated by RapidKL)
and -MRT for mass metro (Own by another company called MRT Corporation, but the MRT Corp dont operated it, they give it to RapidKL)

So one can interchange seamlessly from MRT to LRT (e.g. Masjid Jamek station), but they cannot interchange to suburban rail station

(forgot to mention, there is another private operator call ERL that operate train from KL to its airport.)

In this NSI, I only include metro subway (LRT and MRT)  since they exist in subway validator as one entity (and also one operator), I dont think there is any issue regarding this PR.

However I am quite not sure about other region, but I will follow the final discussion by yours.

bhousel:
closing as stale

PR #5093, opened 2021-05-26
add pantai timor mart

MASangys:

PR #5094, opened 2021-05-26
add billion and pantai timor

MASangys:


bhousel:
Let's discuss on #5093 (we can do all of whatever this is on a single Pull Request)

bhousel:
I did the research and it makes sense now.

PR #5095, opened 2021-05-27
Add Veritas in Belgium

jduchateau:
Adds Veritas which is a clothing and sewing store in Belgium.

It seems that `npm run build` removes the _sewing_ tag, 
I don't know if it is possible to have multiple values or is there another method to represent multiple values?

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -6402,6 +6402,17 @@
       }
     },
     {
+      "displayName": "Veritas",
+      "locationSet": {"include": ["be", "lu"]},
+      "tags": {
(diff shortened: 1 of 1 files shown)


bhousel:
> I don't know if it is possible to have multiple values or is there another method to represent multiple values?

Semicolon-delimited values are not supported - instead in this project we just pick the best match for a brand.

jduchateau:
Ok, thanks for the prompt response.
Clothes seem more general and more suitable in this case.

bhousel:
> Clothes seem more general and more suitable in this case.

I agree - thanks, and welcome to the project 👋

PR #5097, opened 2021-05-28
added Villeroy & Boch

Contrib1043:


bhousel:
Thanks @Contrib1043 - I wasn't sure at first whether they actually operate retail shops, but it looks like they do.

PR #5101, opened 2021-05-30
added German houseware stores from Fissler

Contrib1043:


data/brands/shop/clothes.json
```diff
@@ -4606,6 +4618,7 @@
       },
       "tags": {
         "brand": "Okaïdi",
+		"brand:de": "Okaidi",
```
bhousel, line 4621: Let's not add a `brand:de` tag.  We don't add these for any other item.

PR #5109, opened 2021-06-02
Distinction between ALDI Nord and ALDI Süd

Contrib1043:
There is no legal company or common brand named **ALDI**, it's either ALDI **Nord** or ALDI **Süd**

This are two completely independent companies with own branding. In each country stores are either operated by ALDI Nord **or** ALDI Süd, but never mixed. In Germany there's the so called "ALDI equator": all stores in Northern Germany belong to ALDI Nord, all stores in Southern Germany belong to ALDI Süd.

Thus removed the (misleading) general ALDI entity and assigned each country to either Aldi Süd or ALDI Nord

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -195,60 +195,58 @@
       }
     },
     {
-      "displayName": "Aldi Nord",
-      "id": "aldinord-0a839e",
-      "locationSet": {"include": ["de","fr","es","pt","pl","dk","be","nl","lu"]},
+      "displayName": "ALDI",
+      "locationSet": {
+        "include": ["au", "cn", "ch", "gb", "hu", "ie", "it", "us"],
+		"exclude": ["de"]
+      },
+	  "note": "countries in which ALDI Süd group is active",
       "tags": {
-        "brand": "Aldi Nord",
-		"brand:fr": "Aldi",
-		"brand:es": "Aldi",
-		"brand:pt": "Aldi",
-		"brand:pl": "Aldi",
-		"brand:dk": "Aldi",
-		"brand:be": "Aldi",
-		"brand:nl": "Aldi",
-		"brand:lu": "Aldi",
+        "brand": "ALDI",
+        "brand:wikidata": "Q41171672",
+        "brand:wikipedia": "en:Aldi",
+        "name": "ALDI",
+        "shop": "supermarket"
+      }
+    },
+	{
+      "displayName": "ALDI",
+      "locationSet": {
+        "include": ["be", "dk", "fr", "lu", "nl", "pl", "es", "pt"],
+		"exclude": ["de"]
+      },
+	  "note": "countries in which ALDI Nord group is active",
+      "tags": {
+        "brand": "ALDI",
         "brand:wikidata": "Q41171373",
         "brand:wikipedia": "en:Aldi",
-        "name": "Aldi",
-		"name:de": "Aldi Nord",
-		"name:fr": "Aldi",
-		"name:es": "Aldi",
-		"name:pt": "Aldi",
-		"name:pl": "Aldi",
-		"name:dk": "Aldi",
-		"name:be": "Aldi",
-		"name:nl": "Aldi",
-		"name:lu": "Aldi",
+        "name": "ALDI",
         "shop": "supermarket"
       }
     },
     {
-      "displayName": "Aldi Süd",
-      "id": "aldisud-0a839e",
-      "locationSet": {"include": ["de","us","gb","ie","au","ch","cn","hu","it"]},
+      "displayName": "ALDI Nord",
+      "id": "aldinord-0a839e",
+      "locationSet": {"include": ["de"]},
+	  "note": "German ALDI Nord branding",
       "tags": {
-        "brand": "Aldi Süd",
-		"brand:us": "Aldi",
-		"brand:gb": "Aldi",
-		"brand:ie": "Aldi",
-		"brand:au": "Aldi",
-		"brand:ch": "Aldi",
-		"brand:cn": "Aldi",
-		"brand:hu": "Aldi",
-		"brand:it": "Aldi",
+        "brand": "ALDI Nord",
+        "brand:wikidata": "Q41171373",
+        "brand:wikipedia": "en:Aldi",
+        "name": "ALDI Nord",
+        "shop": "supermarket"
+      }
+    },
+    {
+      "displayName": "ALDI Süd",
+      "id": "aldisud-0a839e",
+      "locationSet": {"include": ["de"]},
+	  "note": "German ALDI Süd branding",
+      "tags": {
+        "brand": "ALDI Süd",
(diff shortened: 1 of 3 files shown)


Contrib1043:
for reference: 

ALDI Nord countries
https://www.aldi-nord.de/unternehmen/ueber-aldi-nord/die-unternehmensgruppe-aldi-nord.html

ALDI Süd countries
https://www.aldi-sued.de/de/unternehmen/ueber-uns/aldi-sued-weltweit.html
(to make things more complicated ALDI Süd operates under Hofer brand in Austria and Slovenia though)

data/brands/shop/supermarket.json
```diff
@@ -194,41 +194,61 @@
...
-      "locationSet": {"include": ["de"]},
+      "locationSet": {"include": ["de","fr","es","pt","pl","dk","be","nl","lu"]},
       "tags": {
         "brand": "Aldi Nord",
+		"brand:fr": "Aldi",
+		"brand:es": "Aldi",
+		"brand:pt": "Aldi",
+		"brand:pl": "Aldi",
+		"brand:dk": "Aldi",
+		"brand:be": "Aldi",
```
LaoshuBaby, line 208: Do you really need to add so many language versions? Especially because most of them are the same? Just keep the ```brand```？
（only my advice）

Contrib1043: There might be other options, but I thought this would be the most easy one:

- 1 record for ALDI Süd
- 1 record for ALDI Nord 

Another option is using

- 1 record for ALDI Süd in Germany (brand="ALDI Süd")
- 1 record for ALDI Nord in Germany (brand="ALDI Nord)
- 1 record for ALDI Süd for all countries except Germany (brand="ALDI")
- 1 record for ALDI Nord for all countries except for Germany (brand="ALDI")

But then it's 4 records instead of 2 records.

Since it's really two completly independent companies, we need a solution as the current tagging scheme is error prone in Germany where both companies are active.

bhousel (changes requested):
Let's not change Aldi

Contrib1043:
I didn't want to offensive, so when misleading was the wrong word please execuse my bad English..

What I meant are examples like this:
https://www.openstreetmap.org/way/118726901

This is clearly an ALDI Nord store, but it's wrongly tagged as Q125054, probably due to the "150" tag in 

https://github.com/osmlab/name-suggestion-index/blob/09b4faa7618b0663714f5fcf09f600050eeb8181/data/brands/shop/supermarket.json#L201

In https://overpass-turbo.eu/s/1853 you'll see that the Q125054 is widely used in Germany, where instead Q41171373 *or* Q41171672 should have been used

So a minimal fix would add {"exclude":["de"]} on the generic ALDI.

But even then we mix up two companies on aldi-883bfe, legally independent, with different corporate identities and assortments on the ground. I read https://github.com/osmlab/name-suggestion-index/wiki/Contributing#identical-names-multiple-brands and understood that in such situations, two distinct records are proposed. That's why I suggested to remove the generic tag and introduce two different tags. 

How shall we proceed?

bhousel:
> But even then we mix up two companies on aldi-883bfe, legally independent, with different corporate identities and assortments on the ground.

ok, I did some reading on https://en.wikipedia.org/wiki/Aldi and I understand now.  I thought that outside of Germany they were all the same thing, but I was wrong.

What you proposed here looks mostly to me - it's ok to have multiple records for how things look in different countries. 

> Another option is using
> 
> 1 record for ALDI Süd in Germany (brand="ALDI Süd")
> 1 record for ALDI Nord in Germany (brand="ALDI Nord)
> 1 record for ALDI Süd for all countries except Germany (brand="ALDI")
> 1 record for ALDI Nord for all countries except for Germany (brand="ALDI")
> But then it's 4 records instead of 2 records.

Contrib1043:
Hope this looks better now :-)

bhousel (approved):

bhousel:
> Hope this looks better now :-)

thanks, yes this looks good 👍

Contrib1043:
> Given there is no overlap in Nord v Sud outside of Germany, would it not make sense just to have the preset name "Aldi" for the rest of the world?

Unfortunately the preset name must be unique, i.e. we can't have two companies under the same name (even if the countries don't overlap). We could rename the preset to something like _ALDI (countries AU, CH, CN, HU, IE, IT, SI, US, UK)_ , but that maybe looks strange too.. Another option would be local presets for each country.

m-hue:
I'd disagree with your conclusion that "Aldi" is the dominant spelling.
https://www.aldi.com.au/ - Welcome to ALDI (heading on the page)
https://www.aldi.be/nl/ - De beste aanbiedingen vind je bij ALDI (heading on the page)
https://www.aldi-sued.de/de/homepage.html - ALDI Filialen (store finder link, on the top of the page)
https://www.aldi.hu/ - ALDI International (link on the top of the page)
https://www.aldi.us/ - About ALDI (link on the bottom of the page)
etc. etc.

The UK website appears to be the exception rather than the norm (and it's inconsistent within itself, too - see the "ALDI Hub" link in the navigation pane on the top of the page).

m-hue:
Also, at least here in the US, the weekly ad flyer thingy as well as the receipts use the uppercase spelling consistently (not just within the logo, but everywhere).

bhousel:
It looks like everyone else is labeling it "ALDI" on their maps too.  

I really don't see this "ALDI" vs "Aldi" as significant - we generally try to pick a `name` tag that matches whatever is on the sign, styled how the brand wants to call itself.

Screen Shot 2021-09-07 at 10 28 32 PM
Screen Shot 2021-09-07 at 10 28 50 PM

rjw62:
Thanks for looking at the ALDI vs Aldi change. You may well be right about it, but I think such a change should have been discussed carefully first, including with local editors where tagging would be affected.

Perhaps more importantly, I think there needs to be a plan to deal with stores currently tagged with the old brand:wikidata=Q125054. There are currently over 6,000 of these worldwide, the majority of which are likely to have been as a result of the previous (now deemed incorrect) NSI entry. I therefore think the NSI maintainers have a responsibility to propose a bulk edit to correct those entries, rather than just allowing a slow natural conversion via the iD suggestions. For data users, it's a real pain to have to deal with inconsistent transitional data like this.

bhousel:
>  I therefore think the NSI maintainers have a responsibility to propose a bulk edit to correct those entries, rather than just allowing a slow natural conversion via the iD suggestions. 

Sure, I'm happy to bulk update the existing features `name` from "Aldi" -> "ALDI" and adjust the `brand:wikidata` "Q125054" -> whichever is better in that country, if that's what people want.  This is really easy to do with JOSM/Overpass.

>  the majority of which are likely to have been as a result of the previous (now deemed incorrect) NSI entry.

To be clear, I don't see this as an "incorrect" -> "correct" thing, but more like "good" -> "better".   
[Q125054](https://www.wikidata.org/wiki/Q125054), still is a valid QID that points to "ALDI" the conceptual parent of ALDI Nord / ALDI Süd.

PR #5110, opened 2021-06-03
Update Hog's Breath Cafe

lkernan:
Wikipedia has caught up with the brand change, now to update NSI.


LaoshuBaby:
Just looking at the name, I thought it was a coffee shop at first, and then I realized that this is actually a restaurant
Thank you @lkernan

PR #5115, opened 2021-06-06
Sinn changed its branding, Schuh not active in Germany

Contrib1043:


LaoshuBaby:
All changes seem to be very normal modifications and no problem, so I merged it. Thanks for @Contrib1043

PR #5120, opened 2021-06-12
Remove dubious entries for Polish fire station operators

gdabski:
KM PSP <place name> is merely the name of the local command of _Państwowa Straż Pożarna_, the state-run professional firefighting organisation, which is the most popular operator name in this category on Poland.

ZOSP RP is not an operator of fire stations but a union of non-public associations, which themselves operate volunteer fire brigades.

Judging by spatial distribution each of the two entries was likely introduced by edits of a single person so the goal is not to make them spread further.


kjonosm:
Hey @gdabski, it would be fantastic, if you added the correct operator tags and included the mentioned values as matchNames!

LaoshuBaby:
> 
> 
> Hey @gdabski, it would be fantastic, if you added the correct operator tags and included the mentioned values as matchNames!

Maybe there isn't  any chain fire-station ?

gdabski:
Hi, thanks for the feedback, I'll look into this. I've only recently learned about NSI and am just learning about the configuration features available :)

gdabski:
@kjonosm , I need some advice here as I'm not sure how matchNames work for operators. I understand they're not regular expressions. But are they matched only against values from the `operator` tag or also against various `name*` tags? Is an exact match required or are substring matches also allowed?

It would make sense to suggest setting/standardising the `operator` to _Państwowa Straż Pożarna_ for `amenity=fire_station`s where the `name*` contains any of `psp`, `jrg`, `państwow.+ straż.? pożarn.+` or `jednostka ratowniczo[- ]gaśnicza`, but it seems for now to me this isn't possible. Otherwise I can only add the acronym `psp` as matchName (as it's done for several similar organisations), although I think few would use that form for `operator`.

As for all other (non-professional or volunteer) fire stations in Poland, it doesn't make sense for the operators to be listed as items in NSI. Basically each of these stations is operated by a different entity, whose name is normally equal to the name of the station, e.g. _Ochotnicza Straż Pożarna w \<place name\>_ (_\<place name\> Volunteer Fire Brigade_). The generic part could be standaridsed (_Ochotnicza Straż Pożarna_ vs _OSP_) but this seems again out of scope.

bhousel:
> @kjonosm , I need some advice here as I'm not sure how matchNames work for operators. I understand they're not regular expressions. But are they matched only against values from the `operator` tag or also against various `name*` tags? Is an exact match required or are substring matches also allowed?

Hey @gdabski there is some guidance here:
https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#matchnames-matchtags

`matchNames` will match against any of the namelike tags, but not substring match. 

If you do want to standardize the `operator` tags of the Fire Stations in Poland, I think the best way is to coordinate with the community to get consensus, then use tools like Overpass -> JOSM or level0  to make a mass change.  Having the "correct" tags in NSI (and the editors that use it) can help keep the fire stations correctly tagged.

gdabski:
OK, then I updated the pull request with the above mentioned matchName and added the Polish generic name to `exclude.generic`. There's nothing more that can be done here (referring to @kjonosm 's request to add operators and matchNames).

bhousel:
Thanks @gdabski 👍

PR #5125, opened 2021-06-19
Change brand for Woolworths Petrol

lkernan:
Ampol is the new name for Caltex in Australia.  
They remain the fuel supplier to Euro Garages Australia (who are taking forever to remove the Woolworths Petrol name anywhere....)

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -4680,7 +4680,7 @@
       "locationSet": {"include": ["au"]},
       "tags": {
         "amenity": "fuel",
-        "brand": "Caltex",
+        "brand": "Ampol",
         "brand:wikidata": "Q5023980",
         "brand:wikipedia": "en:EG Australia",
         "name": "Woolworths Petrol"


bhousel:
I'm really confused by these tags.

`brand` should match the related `brand:wikipedia` and `brand:wikidata` tags, and should be the brand of the station, not the brand of the fuel sold.  

I'll accept the PR anyway, but I think it looks wrong, and I think you should consider revising this later.  `brand` here should really have something to do with Woolworths or EG (I realize the rebranding situation there is really weird and confusing - NSI sometimes uses the `alt_name` tag for a name being phased out 🤷 ).

lkernan:
It’s a tough one since they’ve done such a poor job of replacing the branding on stations.  Most station convenience stores still say Woolworths  (but not all) and they use Ampol (ex Caltex) fuel which is the branding on the canopy!

EG has a very poor social presence here which doesn’t help.  I’m hoping they get on with the rebranding soon which will make it all easier.

PR #5131, opened 2021-06-22
changes on German brands

Contrib1043:
added 2 EDEKA sub-brands, which are used for in various regions / various shop formats
also added common matchTags for Jacques Weindepot

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -1845,6 +1845,17 @@
       }
     },
     {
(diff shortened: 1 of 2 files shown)


bhousel:
thanks @Contrib1043 - the subbrands will still display with the Edeka logo unless they get their own unique wikidata qids, but I think this is ok 👍

Contrib1043:
image

At least Elli makes it very transparent, that they belong to the EDEKA group, so I think, it's fine ;-)

PR #5145, opened 2021-07-15
Corrected wikidata ID of Swiss Post

dafadllyn:
Corrected wikidata ID of Swiss Post, which is [Q107231758](https://www.wikidata.org/wiki/Q107231758).

Note that [Q614803](https://www.wikidata.org/wiki/Q614803) is the Wikidata ID for "Die Schweizerische Post AG", the parent organisation of "Swiss Post" (official company name: "Post CH AG"), to which "PostFinance AG" (financial institution) and "PostAuto AG" (bus company) also belong.

By the way, `operator=*` is currently set to "Die Post", the German brand name, but it should either be set to "Die Post" in German-speaking municipalities, "La Poste" in French-speaking municipalities or "La Posta" in Italian- and Romansh-speaking municipalities. How can this be solved? Is it possible to omit `operator=*` or to add two additional entries for `operator=La Poste` and `operator=La Posta`?

### DIFF
--- a/data/operators/amenity/post_office.json
+++ b/data/operators/amenity/post_office.json
@@ -363,8 +363,7 @@
         "operator:en": "Swiss Post",
         "operator:fr": "La Poste",
         "operator:it": "La Posta",
-        "operator:wikidata": "Q614803",
-        "operator:wikipedia": "de:Die Schweizerische Post"
+        "operator:wikidata": "Q107231758",
       }
     },
     {


LaoshuBaby:
Thanks you for give us more information. If we are not lived in Swiss, we will not know those information.

But here are some points that I confused:

1. ```operator=*``` already have multi language tag such as ```operator:en=*``` or ```operator:fr=*```. I don't know how the original editor use ```Die Post``` for ```operator=*``` ,but it only can use one name.
One posiible answer is that [your given LOGO for this brand](https://commons.wikimedia.org/wiki/File:Die_Post_(ch)_Logo_2019.svg) is using ```Die Post```, another posiible answer is because this brand have ```"locationSet": {"include": ["ch"]},```, so you can refer to the [tagging guideline](https://wiki.openstreetmap.org/wiki/Multilingual_names) for this country/region.

2. [The wikidata Q107231758 you given us](https://www.wikidata.org/wiki/Q107231758) is a almost empty item, and [the "parent item"Q614803 ](https://www.wikidata.org/wiki/Q614803) have detailed information. so maybe most people think the company in Q614803 wikidata item can refer to this brand. (I'm confused by the so much name of company you said, ```Note that Q614803 is the Wikidata ID for "Die Schweizerische Post AG", the parent organisation of "Swiss Post" (official company name: "Post CH AG"), to which "PostFinance AG" (financial institution) and "PostAuto AG" (bus company) also belong.```, so I can't judge which is better)

LaoshuBaby (changes requested):
Maybe you need to answer for [the 2 question](https://github.com/osmlab/name-suggestion-index/pull/5145#issuecomment-881812321) before merge this changes.

dafadllyn:
你好！

>     1. `operator=*` already have multi language tag such as `operator:en=*` or `operator:fr=*`. I don't know how the original editor use `Die Post` for `operator=*` ,but it only can use one name.
>        One posiible answer is that [your given LOGO for this brand](https://commons.wikimedia.org/wiki/File:Die_Post_(ch)_Logo_2019.svg) is using `Die Post`, another posiible answer is because this brand have `"locationSet": {"include": ["ch"]},`, so you can refer to the [tagging guideline](https://wiki.openstreetmap.org/wiki/Multilingual_names) for this country/region.

There are four national languages in Switzerland (German, French, Italian and Romansh). Thus, the Swiss Post has a German name ("Die Post"), a French name ("La Poste"), an Italian and Romansh name ("La Posta") and additionally an English name ("Swiss Post"). It has also four logos.

On OpenStreetMap, the name in the local official language is used for `operator=*`. It would be very unfortunate and likely not accepted by the community if the German name were used for `operator=*` throughout Switzerland. 

Isn't it technically possible to add an additional entry for `operator=La Poste` (with the same Wikidata ID) and one for `operator=La Posta`?

>     2. [The wikidata Q107231758 you given us](https://www.wikidata.org/wiki/Q107231758) is a almost empty item, and [the "parent item"Q614803 ](https://www.wikidata.org/wiki/Q614803) have detailed information. so maybe most people think the company in Q614803 wikidata item can refer to this brand. (I'm confused by the so much name of company you said, `Note that Q614803 is the Wikidata ID for "Die Schweizerische Post AG", the parent organisation of "Swiss Post" (official company name: "Post CH AG"), to which "PostFinance AG" (financial institution) and "PostAuto AG" (bus company) also belong.`, so I can't judge which is better)

The situation is indeed very confusing – and, as i just found out, it has become even more confusing. Let me try to explain (for simplicity i'll only mention the German name of the companies):

"Die Schweizerische Post AG" ([Q614803](https://www.wikidata.org/wiki/Q614803)) is a holding company, which, through its subsidiaries, operates Switzerland's postal system, a national bus network and a financial institution. Until recently, "Post CH AG" ([Q107231758](https://www.wikidata.org/wiki/Q107231758)) was the subsidiary responsible for the postal services. However, now, "Post CH AG" is only responsible for the logistics part of the postal system while the new company "Post CH Netz AG" ([Q107567701](https://www.wikidata.org/wiki/Q107567701)), which was founded in 2020, operates the post offices.

Note that the brand name of "Die Schweizerische Post AG" and its subsidiaries (except from the bus network and the financial institution) is "Die Post". This is the name by which the company and its subsidiaries is best known and which we therefore use for the `operator=*` tag.

Theoretically, this means that post offices (`amenity=post_office`) and postboxes (`amenity=post_box`) now need different `operator:wikidata=*` tags, which i think is too complicated. Besides, if the company structure will change again, all post offices and postboxes would have to be edited. Therefore, it seems to me that using `operator:wikidata=Q614803` probably makes most sense, even if this isn't 100% accurate. But i'll ask on the Swiss OpenStreetMap mailing list what the other contributors think of it. For now, i'll close this pull request.

此致敬礼

PR #5148, opened 2021-07-19
Create drinking_water.json

UKChris-osm:
It's been a while since I have contributed, so I'm not sure if this file has been created correctly.

Wessex Water in the South West of England are installing many water refill points, and I think this would be a good addition to the NSI.


data/operators/amenity/drinking_water.json
```diff
@@ -0,0 +1,22 @@
+{
+  "properties": {
+    "path": "operators/amenity/drinking_water",
+    "skipCollection": true,
```
UKChris-osm, line 4: I am guessing this line refers to the planet collection? If so, I expect it should be removed?

bhousel: Yes I think so.. If you remove this line and run the build again, it will bring in all the "amenity/drinking_water" items found in the planet.  There are a lot of them and they look mostly useful 👍 

https://github.com/osmlab/name-suggestion-index/blob/main/dist/collected/operators_all.json#L3643-L3737

bhousel:
@UKChris-osm this looks good to me - I don't really have anything to add, but you can merge it if you like it 👍

Issue #2588, opened 2019-04-25
Local experts list


matkoniecz:
There is a very long list of brands and typical person will be familiar with only part of them. For example I can relatively easily to spot issues in brands that are active in Poland (or ask family), but I lack any local knowledge for example for Japanese brands.

I made a small script that can automatically detect issues in name-suggestion-index that are present for brands in a specific country.

At this moment I am looking through ones in Poland and post to this issue tracker about Russia with ping to person familiar with local situation.

See for example https://github.com/osmlab/name-suggestion-index/issues/2584 where I summoned local expert with ( using @ syntax - see https://github.blog/2011-03-23-mention-somebody-they-re-notified/ )

If there are other people who would be happy to answer questions about local situation in their country (or countries) - please comment in this issue and list relevant countries.

matkoniecz:
In my case it would be Poland.

Adamant36:
I'll take western America if you want. Although, its probably not to hard to find someone who can do it and stuff here is easy to search for anyway.

ENT8R:
I am always open to answer questions about the situation in Germany 😄

bmillemathias:
I can be ping for france related question.

SilentSpike:
Not an expert, but have local knowledge of Scotland/UK

1ec5:
I’m a local expert on the U.S., which isn’t saying much. 😉 Aside from that, I’m not from Vietnam but I do speak Vietnamese, so I can help w/ research there as well.

matkoniecz:
Thanks everybody! Some issues were already created (and fixed!) but there are still many regions where there some obvious local check waiting. 

-------

For example Switzerland, Ukraine, Indonesia, South Korea and Croatia have known issues requiring local knowledge/research{1} and I may create rickets once there is someone likely to answer them.

If anyone has local knowledge (in some cases local language may be sufficient) then you are welcomed here!

Just answering questions like at https://github.com/osmlab/name-suggestion-index/issues/2648 is enough. PR itself may be done by someone else/ Though there is obviously nothing wrong with answering by making a PR! There is no shortage of things that may be improved.

{1} this list applies to problems reported by `npm run build` and localized within a single country.

jeisenbe:
I can help with Indonesia and Singapore. I also have a little knowledge of Malaysia and Thailand, and I can speak Malay (a.k.a Malaysian, a.k.a Indonesian).

[...]

PR #5150, opened 2021-07-20
GB Coop improvments

Cj-Malone:
In GB there is "the coop", but it's actually a whole bunch of things (and arguably not a coop). According to [Wikipedia](https://en.wikipedia.org/wiki/Co-op_Food#List_of_UK_Co-operatives_with_food_retail_operations), there are 16 companies using it in food shops. The biggest (The Co-operative Group) rebranded to "Co-op Food" a while ago, they tend to just have The Co-operative Group logo, and a sub name on the fascia, but NSI, OSM and the company call them "Co-op Food".

However the other stores didn't go through this brand, and either still called "The Co-operative Food" or have a unique brand. NSI is pushing people from "The Co-operative Food" to "Co-op Food" which is wrong a lot of the time. This is an attempt to fix that, it should be reviewed properly and possibly needs some local experts. https://github.com/osmlab/name-suggestion-index/issues/2588 @rjw62 

This doesn't take Tamworth Co-operative into account as it doesn't have Wikipedia or Wikidata pages. It also ignores Allendale Co-op, Coniston Co-op and Grosmont Co-operative Society as they each only have 1 store.

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -690,6 +690,7 @@
     },
     {
       "displayName": "Central England Co-operative",
+      "id": "centralenglandcoop-4fbfd3",
       "locationSet": {"include": ["gb"]},
       "tags": {
         "brand": "The Co-operative Food",
@@ -703,6 +704,7 @@
     },
     {
       "displayName": "Midcounties Co-operative",
+      "id": "midcountiescoop-fe00f8",
       "locationSet": {"include": ["gb"]},
       "tags": {
         "brand": "The Co-operative Food",
@@ -716,6 +718,7 @@
     },
     {
       "displayName": "Southern Co-operative",
+      "id": "southerncoop-81d078",
       "locationSet": {"include": ["gb"]},
       "tags": {
         "brand": "The Co-operative Food",
@@ -742,6 +745,7 @@
     },
     {
       "displayName": "Chelmsford Star Co-operative Society",
+      "id": "chelmsfordcoop-a7f467",
       "locationSet": {"include": ["gb"]},
       "tags": {
         "brand": "The Co-operative Food",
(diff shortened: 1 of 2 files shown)


rjw62:
I'm not sure exactly what the consensus is (or if there is one) on the use of the brand=* and brand:*=* tags, but I've have thought that if you have, e.g., brand=CO-OP Daily, then the brand:wikidata and brand:wikipedia tags should point to CO-OP Daily entries. At the moment, the pull request has them pointing to entries about the operator (e.g. East of England Co-op Society). This doesn't seem right to me, as the operator is distinct from the brand.

Cj-Malone:
Yeah, good point. I've just done [Co-op Daily](https://www.wikidata.org/wiki/Q107589681) modelled from [Morrisons Daily](https://www.wikidata.org/wiki/Q99752411). I'll do the rest later and update this PR then.

Cj-Malone:
Don't really know what to do with Central England Co-operative, Midcounties Co-operative, Southern Co-operative, Chelmsford Star Co-operative Society, Heart of England Co-operative Society and Radstock Co-operative Society as they all use the same "The Co-operative Food" brand.

rjw62:
If brand=* is genuinely for brands, I guess any shops using "The Co-operative Food" brand should have the same values in those tags. The operator=* tags can distinguish between the different Co-op organisations if necessary. I don't know if/how the NSI mechanisms can cope with that though.

Cj-Malone:
I've created [The Co-operative Food](https://www.wikidata.org/wiki/Q107617274), [Co-op Food](https://www.wikidata.org/wiki/Q3277439) probably needs some adjustments.

I think the failing tests are due to NSI not supporting multiple entries for one brand, removing operators and dropping down to one entry for "The Co-operative Food" will presumably fix this. From my view, it's a bit of a shame, but I think it could get out of hand, imagine if each McDonald's operator was listed in NSI.

bhousel:
Thanks @Cj-Malone the latest looks ok to me 👍

PR #5151, opened 2021-07-20
common branding: TKK instead of Techniker Krankenkasse 

Contrib1043:
added short_name


kjonosm:
@Contrib1043
Wouldn't TK fit better as `short_name`? It's what appears on their logo, see also their website https://www.tk.de
TKK could still be mentioned as `matchName`.

PR #5154, opened 2021-07-23
Fix Pick 'n Pay listing for South Africa + Add Metrorail Eastern Cape

Firefishy:
- Fix Pick n Pay listing
- add Metrorail Eastern Cape


bhousel:
Thanks @Firefishy 👍

PR #5155, opened 2021-07-24
couple of more German brands

Contrib1043:


bhousel:
thanks @Contrib1043 👍

PR #5158, opened 2021-07-26
Update C-TRAN entry in bus.json

arch0345:
Added wikidata, wikipedia, and operator tags to C-TRAN


tas50:
Nice catch. This will be handy to have around town.

PR #5159, opened 2021-07-27
Add RVTD entry to bus.json

arch0345:
I'm not sure if the 'id' I put in this entry is right; feel free to change it if it's incorrect.

Rogue Valley Transportation District (RVTD) is a transit network in Jackson County, OR


tas50:
Another handy one. Thanks

Issue #5138, opened 2021-07-06
iD incorreclty suggesting MTA of Harris County when in LA


tas50:
We need to sprinkle some locations on these. Metro is way too common a name so the suggestions coming from iD aren't the best right now.

![image](https://user-images.githubusercontent.com/1015200/124638595-fdf61180-de3f-11eb-924e-7628475209d6.png)

UKChris-osm:
Perhaps common terms like "_Metro_" should have a filter that won't suggest it unless it has a non-country-code locationSet - such as only when there is a .geojson or a radius locationSet?

I am also wondering, is there any possibility of adding to the yellow box area in iD, and including a link to "_report incorrect suggestion_" so users who see an NSI suggestion they know is incorrect can flag it, or be linked to this repository?

@bhousel

bhousel:
> We need to sprinkle some locations on these. Metro is way too common a name so the suggestions coming from iD aren't the best right now.

I agree!  I think all of the transit and many of the operator items could benefit from a smaller `locationSet` than country-wide.

My workflow for adding these is still kind of cumbersome, so I might look for ways to improve it.

### Step 0
Figure out where the thing is actually used.  I check with Overpass or just look at the official transit map.  We have `network=Metro` all over the place:
<img width="750" alt="Screen Shot 2021-07-28 at 9 39 17 AM" src="https://user-images.githubusercontent.com/38784/127332220-49e00579-0093-4146-b1c9-c3a0e809b599.png">


### Step 1
Go to geojson.io and make a point where you want the center to be:
<img width="750" alt="Screen Shot 2021-07-28 at 9 33 11 AM" src="https://user-images.githubusercontent.com/38784/127331420-7fbb91fe-44bc-4187-bb56-fdc6aadff1d5.png">


### Step 2  
Go to the [location-conflation viewer page](https://ideditor.github.io/location-conflation/) and paste that in, trim the excess precision, and adjust the center and radius manually until it looks right:
<img width="750" alt="Screen Shot 2021-07-28 at 9 33 41 AM" src="https://user-images.githubusercontent.com/38784/127331451-9e58fc9e-bb1d-4662-968e-aba769d2b555.png">

Then you have a `locationSet` that can be used in NSI.

So, I'm thinking it would be great to add some simple drawing tools to that location-conflation page, so I dont need to manually adjust the size and position.

bhousel:
[...]

Issue #4921, opened 2021-02-22
Fail if there are two instances of a place with overlapping match tags


peternewman:
Fail if there are two instances of a place with overlapping match tags (to catch issues like this)?

_Originally posted by @peternewman in https://github.com/osmlab/name-suggestion-index/pull/4920#discussion_r580367185_

peternewman:
These two effectively have the same tagging via the match tags:
https://github.com/osmlab/name-suggestion-index/blob/ea9d441ffb3e1594fca160910219929b4393eb12/data/brands/shop/interior_decoration.json#L76-L89
https://github.com/osmlab/name-suggestion-index/blob/e63037f4d9a4435eddb9b4784b55c464d47d358a/data/brands/shop/furniture.json#L438-L452

So it should complain.

See discussion in https://github.com/osmlab/name-suggestion-index/pull/4920/files#r580084376 for more detail.

PR #5160, opened 2021-07-27
Create separate entries for SMART bus routes

arch0345:
Add 'smart-add5eb' for Suburban Mobility Authority for Regional Transportation and added wikidata/wikipedia and operator tags to each

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -10868,6 +10868,7 @@
       "locationSet": {"include": ["us"]},
       "tags": {
         "network": "South Metro Area Regional Transit",
+        "network:short": "SMART",
         "network:wikidata": "Q7567940",
         "network:wikipedia": "en:South Metro Area Regional Transit",
         "operator": "South Metro Area Regional Transit",
@@ -10880,6 +10881,7 @@
       "locationSet": {"include": ["us"]},
       "tags": {
         "network": "Suburban Mobility Authority for Regional Transportation",
+        "network:short": "SMART",
         "network:wikidata": "Q7632360",
         "network:wikipedia": "en:Suburban Mobility Authority for Regional Transportation",
         "operator": "Suburban Mobility Authority for Regional Transportation",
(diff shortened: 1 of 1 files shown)


tas50:
If you run `npm run build` you should get a unique id here

arch0345:
Sorry I'm a bit new to contributing on GitHub. I'm assuming the `npm run build` you're talking about is under 'checks'. When I looked at the error it said there were duplicate ids for the SMART network in Detroit, listing the id for the SMART network in Wilsonville (smart-a21cc9). Is it talking about the literal `id` tag in this entry? I added a different id for the SMART network in Detroit (smart-add5eb).

tas50:
You run that locally in the git repo you checked out. There's some information in the readme on the commands you can use I believe. It generates unique IDs.

arch0345:
Thanks for the clarification. I found the article you were talking about on the [Info for Maintainers](https://github.com/osmlab/name-suggestion-index/wiki/Info-for-Maintainers) wiki page.

After running `npm run build` a couple of times, it seemed to only work if I expanded the names to make them unique, which was fine by me. I'm still a little confused on why that error kept popping up even though there are multiple 'Metro' entries on the same file.

LaoshuBaby:
> ).

I guess it is because the conflict of ```displayName``` tag, next time you can try to screenshot or copy the wrong hint.

if 2 item have same displayName, nodejs script will throw out error.

(2 item share same id will also)

kjonosm:
@bhousel: Is it ok to use the same `network:short=SMART` for two separate items within the same location set? How does the validator work in such instances?

bhousel:
> @bhousel: Is it ok to use the same `network:short=SMART` for two separate items within the same location set? How does the validator work in such instances?

I wouldn't do it - the matcher code will try to match on "smart" and might pick the wrong one.  

It would be better to have more narrowly scoped locationSets (like we're discussing in #5138)
And we should warn or fail if there are ambiguities like "smart" within a locationSet (which is #4921)

bhousel:
I ended up redoing this in 72d137d0 after seeing how much trouble it can be to rename a `network` value that's already in use (see https://github.com/openstreetmap/iD/issues/8606) .

So I decided it's better to leave `network=SMART` for the different networks that use it (and I found another light rail that uses this value too), but gave them all unique locationSets scoped to where they are.

PR #5162, opened 2021-07-28
Updates Washington, DC transit authority name in subway.json

mxxcon:
This corrects operator name from its abbreviated name to the full name according to OSM best practices in https://wiki.openstreetmap.org/wiki/Names#Abbreviation_.28don.27t_do_it.29

### DIFF
--- a/data/transit/route/subway.json
+++ b/data/transit/route/subway.json
@@ -520,7 +520,8 @@
         "network": "Washington Metro",
         "network:wikidata": "Q171221",
         "network:wikipedia": "en:Washington Metro",
-        "operator": "WMATA",
+        "operator": "Washington Metropolitan Area Transit Authority",
+        "operator:short": "WMATA",
         "operator:wikidata": "Q7972051",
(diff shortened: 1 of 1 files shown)


bhousel:
I think isn't the "no abbreviations" guidance more for addressing?  

I don't have an opinion - but if Washington DC transit mappers have previously preferred "WMATA" over "Washington Metropolitan Area Transit Authority", we usually just leave the tags how they've already been mapped.  

A change in NSI like this would prompt users to start upgrading tags in editors, so we should tread lightly.  It might be ok, but I'd feel better if you reached out to whomever is actively mapping WMATA stuff and ask if they are ok with a mass retagging of their stuff.

mxxcon:
It's already happening and this prompted me to create this PR https://www.openstreetmap.org/changeset/108710508

bhousel:
> It's already happening and this prompted me to create this PR https://osmcha.org/changesets/108710508

ok thanks - I did not realize that someone had renamed all of these already, and that NSI was trying to rename them back.

PR #5164, opened 2021-07-28
LA Metro wiki tags

wmisener:
Added operator wiki tags for Metro Local, Metro Rapid, Metro Limited, and Metro Express networks, and network wiki tags for the former two. Also set the display name to include disambiguation of which Metro system the tags correspond to.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -6731,20 +6731,39 @@
       }
     },
     {
-      "displayName": "Metro Limited",
-      "id": "metrolimited-a21cc9",
+      "displayName": "Metro Express (Los Angeles)",
       "locationSet": {"include": ["us"]},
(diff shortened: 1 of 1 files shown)


data/transit/route/bus.json
```diff
@@ -6731,20 +6731,39 @@
...
     },
     {
-      "displayName": "Metro Local",
+      "displayName": "Metro Local (Los Angeles)",
       "id": "metrolocal-a21cc9",
       "locationSet": {"include": ["us"]},
       "tags": {
         "network": "Metro Local",
+        "network:wikidata": "Q77610",
+        "network:wikipedia": "en:Metro Local"
```
tas50, line 6763:
```suggestion
        "network:wikipedia": "en:Metro Local",
```

PR #5165, opened 2021-07-28
I have added Lotte hotel for us, ru, kr

YounessBird:
I have added Lotte Hotel, which is under 3 different display names in 3 different countries USA, Russia and South Korea.


bhousel:
Thanks @YounessBird  👍 
We don't really want items for individual hotels (so I'll remove these), but I think it's ok to add the Lotte brand in the countries where it exists.  The hotel category is set up to allow flexible naming so it's ok for each individual hotel to have a custom name.

Issue #4925, opened 2021-02-24
operator tags on post_box and atms (at least in France)


nuxper:
Note : related to https://github.com/osm-fr/osmose-backend/issues/1174

I implemented a tag suggestion analysis in osmose: http://osmose.openstreetmap.fr/fr/map/#item=3130&class=31302&country=france_rhone_alpes_rhone&zoom=15&lat=45.75165&lon=4.83582&level=1%2C2%2C3&tags=&fixable=

I find the nsi quite verbose for atms and post_box. I'd like your thoughts before I change anything to the nsi:

I observe the following in France at least:
Post boxes are currently all tagged with the following tags:
amenity = post_box
collection_times = Mo-Fr 14:30, Sa 09:00
operator = La Poste
ref = A1N4G1
source = data.gouv.fr:LaPoste - 05/2016 

In France, all post_boxes are operated by La Poste.
If I follow the name suggestion index, I would add:
**brand = La Poste
brand:wikidata = Q373724
brand:wikipedia = fr:La Poste (entreprise française)
operator:wikidata = Q373724
operator:wikipedia = fr:La Poste (entreprise française)**
Even though that's not incorrect, I believe that's a lot of redundant information to add to all 16000 post_boxes in the country. If we wanted to add those tags, a mechanical edit would be more appropriate.

I suggest to exclude "La Poste" from the post_box template.

In a similar way, atms now suggest adding:
**brand = Crédit Mutuel
brand:wikidata = Q642627
brand:wikipedia = en:Crédit Mutuel
operator:wikidata = Q642627
operator:wikipedia = en:Crédit Mutuel**
amenity = atm
name = Crédit Mutuel
operator = Crédit Mutuel
source = Credit-Mutuel-de-Bretagne_20090205 

Maybe operator:wikidata could be useful but but wikipedia for atms ? brand for atm ?
less than 1% of the nodes have a brand tag: https://taginfo.openstreetmap.org/tags/amenity=atm#combinations

UKChris-osm:
Using [Royal Mail](https://nsi.guide/index.html?t=operators&k=amenity&v=post_box#royalmail-f65bfd) Post Boxes as an example, I don't think having both "_brand_" and "_operator_" is necessary if both sets of data are the same, so I'd probably suggest that with entries where "_operator_" is the primary focus (such as Post Box's) that "_brand_" data should be removed.

bhousel:
[...]

Issue #4736, opened 2020-11-26
use network tag with bicycle-sharing


kjonosm:
Continue discussion from #4734

[OSM wiki](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dbicycle_rental) recommends/documents using `network=*` instead of `brand=*` when tagging `amenity=bicycle_rental` stations.

Is there a way to check which tag is used more often with regard to bicycle-sharing?

kymckay:
According to https://taginfo.openstreetmap.org/tags/amenity=bicycle_rental#combinations

`network=*` has significantly more uses in combination

bhousel:
Hmm that's kind of frustrating.  We moved it from `brand` to `operator` back to `brand`.  It's really `network`?  I sort of get why people would say that, but it's very unintuitive to me.

I can switch it over to be in the `transit` tree the next time I reprocess the planet, but we currently are only collecting `network` in combintion with `power` and `route`, not `amenity`.

kjonosm:
`amenity=car_sharing` might have a similar issue. for some reason taginfo doesn't show the numbers...

bhousel:
Yeah I will try to process another planet soon.. If I knew and had it set up already, the data would be in [`dist/collected/networks_all.json`](https://github.com/osmlab/name-suggestion-index/blob/main/dist/collected/networks_all.json)

peternewman:
The main London hire bike scheme has the whole set, network, brand and operator!:
https://github.com/osmlab/name-suggestion-index/blob/fd5e3e886d90390b00ec44e6b51945dd2e305172/data/brands/amenity/bicycle_rental.json#L1526-L1549

kjonosm:
[`dist/collected/networks_all.json`](https://github.com/osmlab/name-suggestion-index/blob/main/dist/collected/networks_all.json) now contains around 260 `bicycle_rental` networks.
Is there a way to automatically merge these into [`data/brands/amenity/bicycle_rental.json`](https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/amenity/bicycle_rental.json)?

bhousel:
> [`dist/collected/networks_all.json`](https://github.com/osmlab/name-suggestion-index/blob/main/dist/collected/networks_all.json) now contains around 260 `bicycle_rental` networks.
[...]

Issue #4672, opened 2020-11-13
Remove redundant data


pyrog:
Hi all,

At least for France, there are a lot of **redundant** tags.

`*:wikipedia` could be removed.

By example, for ATM or post box, (at least french) community could choose to keep `brand` or `operator` tags, **not both** 😃 

Related to #2883

```
amenity=atm
brand=Banque Populaire
brand:wikidata=Q846647
brand:wikipedia=fr:Banque populaire
name=Banque Populaire
operator=Banque Populaire
operator:wikidata=Q846647
operator:wikipedia=fr:Banque populaire
```

```
amenity=post_box
brand=La Poste
brand:wikidata=Q373724
brand:wikipedia=fr:La Poste (entreprise française)
name=La Poste
operator=La Poste
operator:wikidata=Q373724
operator:wikipedia=fr:La Poste (entreprise française)
```

bhousel:
Yes, I'm pretty ok with this redundancy.  I don't really see it as a problem.

Adamant36:
@pyrog do they just not have brands in France? Or is it that they don't think an entity can be an operator and a brand at the same time?

pyrog:
We have brands and operators in France,
And some could be both 😃 

But:
1. `brand:wikipedia=fr:La Poste (entreprise française)` is redondant
1. And especially not welcome when it link to a english page (cf. `brand:wikipedia=en:Bank of France`)

Apps should display the appropriate wikipedia page according to the user language.
So that's why I think that `brand:wikipedia` should be removed from NSI (at least for France 😉)

In the first exemple this should be enough to display useful information in iD or any other application:

```
amenity=atm
name=Banque Populaire
operator=Banque Populaire
operator:wikidata=Q846647

```

Or even 😉 
```
amenity=atm
operator:wikidata=Q846647
```

It could be `brand:wikidata=Q846647` (this is a detail for me), just avoid ~"Banque Populaire" 7 times for an ATM of this brand/operator.

Adamant36:
> brand:wikipedia=fr:La Poste (entreprise française) is redondant

Why when it's both the operator and a brand?

> And especially not welcome when it link to a English page (cf. brand:wikipedia=en:Bank of France)

The fix for that would be to change it to the French Wikipedia article. 

> Apps should display the appropriate wikipedia page according to the user language.

[...]

Issue #4136, opened 2020-09-12
FR: Add Emmaüs (charity)


gendy54:
Hello,

Country code: FR (only)

"shop":"charity"
"second_hand":"yes"
"operator":"Mouvement Emmaüs"
"operator:wikidata":"Q989437"
"operator:wikipedia":"fr:Mouvement Emmaüs"

I deliberately changed "brand" to "operator".

UKChris-osm:
> I deliberately changed "brand" to "operator".

Why so?

Adamant36:
> I deliberately changed "brand" to "operator".

We're only doing brand:* at this point. The operator tag was intentionally removed awhile back because it wasn't really clear.

gendy54:
Because Emmaus is an association. So the brand tag doesn't seem at all appropriate to me.

bhousel:
Hey @gendy54 - OpenStreetMap convention is to use the `operator` tag for the company, corporation, person or any other entity who is directly in charge of the current operation of a single location.. People have been confused by this, so we don't add them right now.  See https://wiki.openstreetmap.org/wiki/Key:operator for more info, also these old issues.

#2471   
#2736
#2816

If these are storefronts with common signage or logo, `brand` is ok.

pyrog:
Hi Brian,

Just an explanation from this side of the Atlantic 😃 

Brand mean (profit) company, GAFAM…
Emmaüs is a non profit, a charity association.

So at least french couldn't use brand 😅

bhousel:
@pyrog You can see from here that there are quite a lot of charity shops, blood banks, social facilities, (on both sides of the Atlantic) using the `brand` tag in this way:  
https://nsi.guide/index.html?t=brands&k=shop&v=charity
https://nsi.guide/index.html?t=brands&k=healthcare&v=blood_donation
https://nsi.guide/index.html?t=brands&k=amenity&v=social_centre

Brand really doesn't mean "for profit company". 

I'd strongly encourage the French community to try to do what everyone else in OSM does, but if the `brand` tag really bothers you, we can use `operator` now.  

(For a long time we didn't include `operator` tags here because many contributors were misusing it, but it is allowed now).

gendy54:
If the tag `operator `is accepted, that's fine. Honestly, no French person will put a tag `brand `on structures carried by non-commercial entities: associations, foundations, public services,... It's counter-intuitive.

Adamant36:
[...]

PR #5167, opened 2021-07-28
Moved PubliBike from brand to operator

dafadllyn:
PubliBike is an operator (a company), not a brand. Besides, the networks operator by PubliBike are called "Lausanne-Morges", "Agglo Fribourg-Freiburg", "Lugano-Malcantone", "Sion", "Sierre", "Région de Nyon", "Velohauptstadt" and "Züri Velo". There isn't a network called "PubliBike".

### DIFF
--- a/data/brands/amenity/bicycle_rental.json
+++ b/data/brands/amenity/bicycle_rental.json
@@ -2504,21 +2504,6 @@
       }
     },
     {
-      "displayName": "PubliBike",
-      "id": "publibike-61df48",
-      "locationSet": {"include": ["ch"]},
-      "tags": {
-        "amenity": "bicycle_rental",
-        "brand": "PubliBike",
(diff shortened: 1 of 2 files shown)


bhousel:
You can't just move a few items in the bike sharing file from `brand` to `operator`.  Normally I like to let people do whatever they want in the name-suggestion-index, but this is a situation where I'm concerned that this PR might break things.  

`bicycle_sharing` is a particularly weird category - there is some history on #4736 and #4925 and I think we haven't completely settled on what these things are.  They could be `brands` or `operators` or `networks` really.

Also, I'm aware from #4672 and #4136 that the French community has a hangup with using `brand` in some situations - maybe bicycle_sharing stations is one of these. 

Can you open a new issue first and we'll discuss?  By closing this, I'm not saying "no" , I'm saying "not without really understanding the issue completely".

PR #5168, opened 2021-07-28
Update old transport service names

dhmartin:
May I try a few changes in order to have the new version of ID more comfortable with our bus routes? I have fixed some references in Wikipedia and Wikidata and I hope I made this pull request in a proper way…

I'm not sure if the "id" change in line 922 is appropriate. Even though it shouldn't be visible, it inherits a tiny old misspelling,.

David

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -918,19 +918,18 @@
       }
     },
     {
-      "displayName": "Autobuses Urbano de Burgos",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Welcome to NSI community! Thanks @dhmartin for fix this item with wikipedia.
Changes looks good, but ```id``` will be generated by program automatedly, you don't need to modify it manually.

Issue #3557, opened 2020-02-23
Add Goodwill Donation Express


tas50:
This one is tricky because there doesn't seem to be a great tag combo to use, but Goodwill operates locations called Goodwill Donation Express which are drop off only locations.

Example location in Portland metro: https://www.yelp.com/biz/goodwill-donation-express-gresham?osq=goodwill+donation+express

1ec5:
Some regional Goodwill affiliates also have other drop-off-only formats, like this [drive-through car donation center](https://www.openstreetmap.org/way/42425057) in Northern Kentucky. I tagged it and [a similar facility](https://www.openstreetmap.org/way/193832589) as recycling centers for cars, by analogy with clothes donation bins that NSI tags as recycling bins. It is admittedly a bit of a stretch to tag “recycling” instead of “donation”.

mds08011:
These seem pretty similar to what are branded as "Goodwill Donations" in San Diego, California (USA).  Usually located in shopping centers but only intake donated items: nothing is sold.

I have been tagging them as `amenity=recycling`+`recycling_type=centre` along with `recycling:clothes=yes` to clarify that they aren't a "cans and bottles" type of facility.

<a title="Mds08011 / CC BY (https://creativecommons.org/licenses/by/4.0)" href="https://commons.wikimedia.org/wiki/File:Goodwill_Donations_storefront_in_San_Diego,_CA.jpg"><img width="512" alt="Goodwill Donations storefront in San Diego, CA" src="https://upload.wikimedia.org/wikipedia/commons/thumb/b/b7/Goodwill_Donations_storefront_in_San_Diego%2C_CA.jpg/512px-Goodwill_Donations_storefront_in_San_Diego%2C_CA.jpg"></a>

<a href="https://commons.wikimedia.org/wiki/File:Goodwill_Donations_storefront_in_San_Diego,_CA.jpg" title="via Wikimedia Commons">Mds08011</a> / <a href="https://creativecommons.org/licenses/by/4.0">CC BY</a>

bhousel:
I think #5171  closed this..

PR #5171, opened 2021-07-29
Add Goodwill to recycling.json (#3557)

arch0345:
Closes issue #3557

Used generic name Goodwill as different branches use different names for their donation centers (donation express, donations, etc.) and left out recycling key in tags as some are donation bins that only accept clothes while others accept a variety of items


tas50:
I do think we need to use the Donation Express naming when we tag for recycling. Those locations are drop off only vs. the existing Charity Store tagging that sell and take drop offs.

PR #5174, opened 2021-07-30
Refine Cincinnati-area public transit networks

1ec5:
Various fixes to public transportation networks in the Cincinnati area:

* Distinguish between Metro and the Cincinnati Bell Connector, which are both operated by SORTA as separate systems. (Cincinnati Bell Connector features [have already been retagged](https://www.openstreetmap.org/changeset/108872326) in OSM.)
* Clermont Transportation Connection operates buses, not trams.
* Added a feature roughly corresponding to the Metro service area, which CTC and BCRTA also operate within.


bhousel:
Thanks @1ec5 👍

PR #5177, opened 2021-08-01
Add operator to Cherriots

arch0345:
This operator value is already in use by most of Cherriots' bus routes

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -2260,6 +2260,7 @@
         "network": "Cherriots",
         "network:wikidata": "Q7404020",
         "network:wikipedia": "en:Cherriots",
+        "operator": "Salem Area Mass Transit District",
         "route": "bus"
       }
     },


UKChris-osm:
I think personally I would add this as a matchName rather than an operator, as it seems the "_Cherriots_" is the general name used day to day, with "_Salem Area Mass Transit District_" being the behind the scenes legal name?

@bhousel what would your take on this be?

bhousel:
> I think personally I would add this as a matchName rather than an operator, as it seems the "_Cherriots_" is the general name used day to day, with "_Salem Area Mass Transit District_" being the behind the scenes legal name?
> 
> @bhousel what would your take on this be?

I think how he submitted it is ok.  `operator` is _complicated_ when it comes to transit. 

I've been adding locations to lots of regional transit things the past few days, mostly in the US, and I see the `operator` field used in different ways.  Sometimes it's the same as the `network` , sometimes it's the name spelled out, sometimes it's the name of the local transit department or authority, sometimes a transit agency uses a single (or even multiple) private companies to operate their bus lines.

So I don't really assume anything with `operator` but I do try to match the existing tagging.  If someone decides later that they want to change the `operator` of `Cherriots`, it's easy to do.

PR #5181, opened 2021-08-01
Complete version with Renfe network entries.

dhmartin:
I have added some Spanish railway networks, as the proposals we're getting in the ID editor are somehow weird...

I understood that the "id" field gets automatically generated ("id will be generated by program automatedly, you don't need to modify it manually.") so I completely avoided them this time.

Regards,
David

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -1236,6 +1236,97 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @dhmartin , looks good 👍

PR #5182, opened 2021-08-01
Add missing bus networks in Oregon

arch0345:
It's recommended to modify the locationSet of Canby Area Transit and Columbia Area Transit as they both have the same network value. I'm not sure how add a specific area for the locationSet, so I just left it as 'us' for now.

These missing networks were found from this category page on Wikipedia:
https://en.wikipedia.org/wiki/Category:Transit_agencies_in_Oregon

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -1895,6 +1895,18 @@
       }
     },
     {
+      "displayName": "Canby Area Transit",
+      "id": "canbyareatransit-783b59",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "network": "CAT",
+        "network:wikidata": "Q5031301",
+        "network:wikipedia": "en:Canby Area Transit",
+        "operator": "City of Canby",
+        "route": "bus"
+      }
+    },
+    {
       "displayName": "Cape Town IRT",
       "id": "capetownirt-4d0252",
       "locationSet": {"include": ["za"]},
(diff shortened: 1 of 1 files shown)


arch0345:
The only error is because of the duplicate network values for Canby Area Transit and Columbia Area Transit

LaoshuBaby:
Thanks for detail, but ```id``` is auto generated by ```"network":*``` so if you use ```"network":"CAT"``` for both 2 item, and because they are in same ```locationSet```, program will think they are same ```"cat-a21cc9"``` and throw error, without edit ```bus.json```, so you will hardly found that.
It is better to use ```"network:short":"CAT"``` and leave ```"network":*``` as their full name.
 (If have better way, I hope that senior in NSI can point for me!)

data/transit/route/bus.json
```diff
@@ -1899,7 +1899,8 @@
       "id": "canbyareatransit-783b59",
       "locationSet": {"include": ["us"]},
       "tags": {
-        "network": "CAT",
+        "network": "Canby Area Transit",
+        "network:short": "CAT",
```
LaoshuBaby, line 1903: Use ```"network:short":*``` will let program not use ```CAT``` to generate ```id```

arch0345:
Thanks for fixing these issues.

In #5160, it looks like bhousel used coordinates as the locationSet to fix having 2 networks of the same name. However, in this case, it's fine to leave the network and network:short tags because I've found routes of these networks already use the unabbreviated names.

LaoshuBaby:
Oops! I just done the unsuggestive behavier that maintainer pointed。
Sorry to make bad commit……I apologise……
But I don't familiar with geojson in US， so **I will just revert my modification and wait for another editor**

LaoshuBaby:
Great!

Issue #3409, opened 2019-11-28
Use French Wikipedia for French brands


bikepunk:
The request is to switch brand:wikipedia tag from "en:article" to "fr:article" since these are French companies operating mainly in france

-Leader Price
-Groupe Caisse d'Épargne
-Renault
-Géant Casino
-Brico Dépôt

Adamant36:
Hi there. There isn't really a specific policy or anything that dictates what languages the Wikipedia articles in. Most of them were just picked randomly at the time they were added. At this point it's mostly about having people submit PRs for them to be changed. So, feel free to do a PR for these if you want. If the decision to change them was made by others besides you, maybe you could encourage them to contribute PR's also. It would be really helpful.

doktorpixel14:
I would not change Renault because it operates in serveral countries. 

Otherwise, I don't really get the point of adding 'brand:wikipedia'. Wouldn't it be enough to only add 'brand:wikidata'? The Wikidata object is connected to all Wikipedia languages so isn't 'brand:wikipedia' obsolete?

bhousel:
> Otherwise, I don't really get the point of adding 'brand:wikipedia'. Wouldn't it be enough to only add 'brand:wikidata'?

Yes technically all of the Wikipedia tags are unnecessary in OSM.  Wikidata QIDs offer a more stable link to a concept.  But it really doesn't hurt anything to add the corresponding Wikipedia tags, and I think for now people expect them to be there.

Someone could ask the broader OSM community in other channels whether it might make sense to just remove all the Wikipedia tags from OSM and standardize on Wikidata for everything, but I feel like there would likely be pushback.

Adamant36:
> I would not change Renault because it operates in several countries.

Same goes for Leader Price and Brico Dépôt.

camelCaseNick:
> Yes technically all of the Wikipedia tags are unnecessary in OSM. Wikidata QIDs offer a more stable link to a concept.

In most situations...

Though with de:Aldi or en:Aldi, we have an article for Q41171373 (Aldi Nord) and Q41171672 (Aldi Süd), since the company split into two, while Aldi is the common brand, the Wikidata tags are more precise to which brand and group one store belongs.

Why do I tell? The problem then is, it is not simply a bijection between Wikidata and Wikipedia and it isn't just simply selecting the corresponding Wikipedia article for the Wikidata entry, since Aldi Nord and Süd don't have one.

Adamant36:
[...]

PR #5192, opened 2021-08-03
update brand "Hofer" in fuel.json

Luzandro:
actual brand is "Diskont", "Hofer" is just the cooperating supermarket

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -1977,15 +1977,16 @@
       }
     },
     {
-      "displayName": "Hofer",
+      "displayName": "Hofer Diskont",
       "id": "hofer-3baa0f",
       "locationSet": {"include": ["at", "si"]},
       "tags": {
         "amenity": "fuel",
-        "brand": "Hofer",
-        "brand:wikidata": "Q15815751",
-        "brand:wikipedia": "de:Hofer KG",
-        "name": "Hofer"
+        "brand": "Diskont",
+        "brand:wikidata": "Q107803455",
+        "name": "Hofer Diskont",
+        "operator": "Avanti",
+        "website": "https://www.dikonttanken.at/"
       }
     },
     {


bhousel:
thanks @Luzandro and welcome 👋 
There are some issues with this, but I'm probably just going to merge and clean up, since I do want to get releases out quickly this week.

- I'm concerned that you just created Q107803455 this morning and it doesn't have a lot of supporting information. This isn't bad necessarily, but an entry like this is at risk for deletion from the Wikidata admins. It would be great if you could add some more references about the "Hofer Diskont" brand (like if it was mentioned in a newspaper or trade publication).
- The `brand` "Diskont" should probably be "Hofer Diskont" to match the `name`, and to match the label on Q107803455
- We don't normally add `operator` tags to brands.  I don't know whether it's universally true that "Avanti" operates all of these fuel stations or whether they operate like franchises with individual operators.
- We don't add `website` tags because people can look up this information from the wikidata. Also it looks misspelled.
- other than all that, it looks fine 👍

Luzandro:
Oh, sorry - I've started a discussion on the austrian osm forum a few months ago and apparently nobody noticed the typo until now. There are also a few more references, all in german: https://forum.openstreetmap.org/viewtopic.php?id=71945

I had hoped the official website would be reference enough to start with. As far as I understand it, NSI needs to link to a wikidata object, but the previously linked wikidata/wikipedia entries represent the supermarket and don't mention the petrol stations at all.

Luzandro:
As for people can look up the website from wikidata - isn't that true for wikipedia as well, which is added most of the time? The first time I came across NSI, was when I was wondering why nearly all of the Eurospar supermarkets in austria referenced a danish wikipedia article
https://nsi.guide/index.html?t=brands&k=shop&v=supermarket&tt=eurospar

Another problem, that I came across in this context was that I couldn't even add the german article to the wikidata object, as https://de.wikipedia.org/wiki/Eurospar is just a redirect to https://de.wikipedia.org/wiki/Spar#Eurospar and Spar is already linked to another wikidata object.

bhousel:
> As for people can look up the website from wikidata - isn't that true for wikipedia as well, which is added most of the time?

Yes people ask about this a lot - the `*:wikipedia` tag doesn't really matter much and it is kind of unnecessary.  We just add it because some people like to see it and it doesn't cause any problems. (https://github.com/openstreetmap/iD/issues/8620 , #4672, #3409)

> Another problem, that I came across in this context was that I couldn't even add the german article to the wikidata object, as https://de.wikipedia.org/wiki/Eurospar is just a redirect to https://de.wikipedia.org/wiki/Spar#Eurospar and Spar is already linked to another wikidata object.

Probably the right way to do this is to replace that redirect with a new German wikipedia article about Eurospar brand.  Then you can have it linked to the Eurospar [Q12309283](https://www.wikidata.org/wiki/Q12309283)

PR #5199, opened 2021-08-04
add meps

MASangys:


data/brands/amenity/atm.json
```diff
@@ -60,6 +60,21 @@
...
     },
+    {
+      "displayName": "MEPS",
+      "id": "meps-396707",
+      "locationSet": {"include": ["my"]},
+      "tags": {
+        "amenity": "atm",
+        "brand": "MEPS",
+        "brand:wikidata": "Q6741957",
+        "brand:wikipedia": "en:Malaysian Electronic Payment System"
```
LaoshuBaby, line 71: 忘加逗号了，小问题

LaoshuBaby:
We seldom add item to atm.jsom because most of atm are belonging to a bank, and atm.json will herit from bank.json

But from [wikipedia](https://en.wikipedia.org/wiki/Malaysian_Electronic_Payment_System), This is OK.

Thanks @MASangys  for adding local brands.

PR #5204, opened 2021-08-04
Update Wikipedia language for De Lijn in bus.json (#5201)

### LINKED ISSUE #5201: De Lijn (Flemish, Belgian transport agency) links to the french wikipedia page somehow
Hi,

While the actual source file seems correct (linking to the NL, dutch wikipedia language. Since De Lijn is only active in the Flemish/dutch speaking part of belgium):
https://github.com/osmlab/name-suggestion-index/blob/d2531645f52c6056d0e853fd1e5d7f061f2ebc8b/data/transit/route/tram.json#L93-L102
(there's multiple De Lijn entries)

But on the nsi guide and iD it shows the FR wikipedia tag:
https://nsi.guide/index.html?t=transit&k=highway&v=bus_stop&tt=lijn

arch0345:
Closes #5201 

Changed Wikipedia tag prefix from 'fr' to 'nl'

Also added operator tags and changed prefix of wikidata/wikipedia from network to operator.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -3312,9 +3312,9 @@
       "locationSet": {"include": ["be", "nl"]},
       "tags": {
         "network": "DLAn",
-        "operator": "De Lijn", 
-        "operator:wikidata": "Q614819",
-        "operator:wikipedia": "nl:De Lijn",
+        "network:wikidata": "Q614819",
+        "network:wikipedia": "nl:De Lijn",
+        "operator": "De Lijn",
         "route": "bus"
       }
     },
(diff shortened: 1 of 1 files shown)


bhousel:
For these we definitely still need the `network:*` tags to be there.  Those are the tags that define the transit item.

bhousel (changes requested):
looks like `network:*` was changed to `operator:*` but should stay as `network:*`

bhousel (approved):

bhousel:
Thanks for the quick fix @arch0345 👍

PR #5205, opened 2021-08-04
Add missing bus networks in the Pacific Northwest

arch0345:
Modified displayName and locationSet of Brockton Area Transit

Removed blank entry 'CAT' as two other networks I added in previous PRs have the same name. Also shortened displayNames of these networks (location instead of unabbreviated name)

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -1134,14 +1134,25 @@
       }
     },
     {
-      "displayName": "BAT (Brockton Area Transit)",
+      "displayName": "BAT (Massachusetts)",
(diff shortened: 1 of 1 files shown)


data/transit/route/bus.json
```diff
@@ -2084,13 +2095,7 @@
       }
     },
     {
-      "displayName": "CAT",
```
tas50, line 2087: If you just remove something without match it's going to come back when the data gets generated again.

bhousel:
Looks good to me!
Thanks for disambiguating the BATs 🦇 and the CATs 🐱

PR #5166, opened 2021-07-28
Add BottleDrop (#3556)

arch0345:
BottleDrop is a recycling center in Oregon that provides refunds to deposits on cans/bottles. In addition to self-service centers, they also have bag drop centers (bags of cans/bottles rather than individual) and a kiosk to withdrawal funds via voucher at select locations.

Added BottleDrop (individual self-service), BottleDrop Express (bag drop) as amenity=recycling, and BottleDrop Plus (fund withdrawal kiosk) as amenity=payment_terminal

PR #5206, opened 2021-08-06
Add BottleDrop bottle return vending machines

arch0345:
BottleDrop is a bottle return service in Oregon that refunds bottle deposits. I already added entries for recycling centers and payment terminals relating to this brand in #5166. Would using bottle return vending machines for partner retailers make more sense than recycling centers?

### DIFF
--- a/data/brands/amenity/vending_machine.json
+++ b/data/brands/amenity/vending_machine.json
@@ -168,6 +168,23 @@
       }
     },
     {
+      "displayName": "BottleDrop",
+      "id": "bottledrop-f8d7ba",
+      "locationSet": {"include": ["us"]},
+      "matchNames": ["bottle drop"],
+      "note": "To be used for self-service stations at partner retailers",
+      "tags": {
+        "amenity": "vending_machine",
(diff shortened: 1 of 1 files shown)


bhousel:
> Would using bottle return vending machines for partner retailers make more sense than recycling centers?

I could see it either as `amenity=vending_machine` or `amenity=recycling`.  I think it might be good to get some more opinions on Slack or the tagging mailing list.

bhousel:
oh I just realized https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dvending_machine
lists an option for `vending=bottle_return`
https://wiki.openstreetmap.org/wiki/Tag:vending%3Dbottle_return

So yeah, I think it would be `amenity=vending machine` + `vending=bottle_return` + whatever `recycling:*` tags are appropriate.

bhousel:
> So yeah, I think it would be amenity=vending machine + vending=bottle_return + whatever recycling:* tags are appropriate.

I'm going to just merge it and do this  ☝️

PR #5209, opened 2021-08-08
Update pub.json

kjonosm:
List "Railway Hotel" as generic pub name


kjonosm:
Change Railway Hotel from "generic" to "named"

PR #5211, opened 2021-08-08
Update restaurant.json

kjonosm:
Treat Al Dente, Amadeus, Chinatown, L'Ardoise, Marina, Pizza Roma, Kyoto, Royal as not part of any chain


bhousel:
Looks good, thanks for filtering these out 👍

PR #5213, opened 2021-08-08
Update sports_centre.json

kjonosm:
Treat "Piscinas Municipales" as not part of any chain

### DIFF
--- a/data/brands/leisure/sports_centre.json
+++ b/data/brands/leisure/sports_centre.json
@@ -31,6 +31,7 @@
       ],
       "named": [
         "^orlik$",
+        "^piscinas municipales$",
         "^sokolovna$",
         "^дюсш$"
       ]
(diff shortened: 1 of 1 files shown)


bhousel:
I feel like this one could probably go in generic - Google translate says it just means "Municipal Swimming Pools"

kjonosm:
> I feel like this one could probably go in generic

I thought about that. But iirc when listed in "generic", the iD validator suggests to remove the operator tag altogether. Because this tag carries a bit more meaning in it (it's a municipal operator) I thought it would be better placed in the "named" category. I don't mind if you want to file it under "generic" though...

kjonosm:
I've just read your comment in #5217 which kind of is the same reasoning to move "Piscinas Municipales" to generic and makes sense to me...

PR #5215, opened 2021-08-08
Update tailor.json

kjonosm:
Treat "Ladies Tailoring" as generic name


bhousel:
seems good - the test failures are unrelated..

PR #5217, opened 2021-08-08
Update drinking_water.json

kjonosm:
Move some entries to either "generic" or "named"


bhousel:
I'd put more of them in generic - "Public" or "Private".  These are strings where if the user adds them, we could show them a warning that it's not really the value.

PR #5220, opened 2021-08-09
Add bakery brand "Kolls" and wlan operator "MobyKlick"

Unkn0wnCat:
Hi all,

This pull request adds the bakery brand Kolls, mainly used in northern Germany as in [this OSM-Feature](https://www.openstreetmap.org/node/6058479489), and the WiFi-Hotspot-Operator "MobyKlick".

If you have any questions, feel free to contact me! 🙂 

Have a nice day,
Kevin

### DIFF
--- a/data/brands/shop/bakery.json
+++ b/data/brands/shop/bakery.json
@@ -952,7 +952,6 @@
       "tags": {
         "brand": "Kolls",
         "brand:wikidata": "Q107151170",
-        "operator": "Bäckerei Konditorei Kolls GmbH",
         "shop": "bakery"
       }
     },
--- a/data/brands/shop/bakery.json
+++ b/data/brands/shop/bakery.json
@@ -941,6 +941,22 @@
       }
     },
     {
+      "displayName": "Kolls",
+      "id": "kolls-3f448e",
+      "locationSet": {"include": ["de"]},
+      "matchNames": ["kolls familienbäckerei"],
+      "matchTags": [
+        "shop/bakery",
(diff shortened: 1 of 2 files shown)


kjonosm:
For bakeries, we usually don't include `operator=` tags. And could you also add the primary `name=` tag to the feature?

Unkn0wnCat:
> For bakeries, we usually don't include `operator=` tags. And could you also add the primary `name=` tag to the feature?

Thanks for the feedback! 

This particular group of bakeries does not use a common name, as each bakery also has an own name. How should I go about tagging that? (For example the one I've linked above is named "Beluga-Café", but is still of the brand "Kolls"

I'll remove the operator tag as soon as I'm at my PC later!

Unkn0wnCat:
@kjonosm I've removed the `operator`-tag from the feature.

kjonosm:
> This particular group of bakeries does not use a common name, as each bakery also has an own name. How should I go about tagging that?

If that's the case you can leave out the name tag. I thought you just forgot about it..

Unkn0wnCat:
Then this should be ready, or is there anything I missed?

bhousel:
> Then this should be ready, or is there anything I missed?

looks good, thanks @Unkn0wnCat and welcome 👋

Unkn0wnCat:
Thanks @bhousel and @kjonosm for your guidance through this first pull-request on this repo. I hope to be able to contribute more in the future! 🙂

PR #5221, opened 2021-08-09
New -> North for North Carolina and North Dakota

arch0345:
iD suggested changing the subject of the North Carolina flag in an area I was mapping from 'North Carolina' to 'New Carolina'. Was there a reason why the subject tags for North Carolina and North Dakota are 'New Carolina' and 'New Dakota'?


1ec5 (approved):
Thanks, this appears to have been copypasta in d641cb284ce4915273037c7769356d9e3133e899.

PR #5222, opened 2021-08-09
Update substation.json

arrival-spring:
Add Northern Powergrid operator:wikidata and wikipedia tags


bhousel:
Looks pretty good - can you also update the `locationSet` to include where they operate?  I guess for this item it could be `"include": ["gb-eng"]`

arrival-spring:
Makes sense, done

bhousel:
Thanks @arrival-spring 👍

PR #5224, opened 2021-08-10
Wikidata/Wikipedia tags for substation operators

arch0345:
Also added locationSet and matchNames for duplicate entries.

### DIFF
--- a/data/operators/power/substation.json
+++ b/data/operators/power/substation.json
@@ -6262,7 +6262,7 @@
         "Tauron Dystrybucja S.A.",
         "Tauron Dystrybucja S.A. Oddział w Częstochowie",
         "Tauron Dystrybucja S.A. Oddział w Krakowie"
-      ]
+      ],
       "tags": {
         "operator": "Tauron",
         "operator:wikidata": "Q2397198",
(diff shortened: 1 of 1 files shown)


bhousel:
Wow thanks @arch0345 !  I appreciate all the work tracking down the duplicates 👍

data/operators/power/substation.json
```diff
@@ -4536,18 +4562,23 @@
     {
       "displayName": "Pacific Gas & Electric",
       "id": "pacificgasandelectric-fc3a87",
-      "locationSet": {"include": ["001"]},
+      "locationSet": {"include": ["us"]},
+      "matchNames": ["PG&E"],
       "tags": {
         "operator": "Pacific Gas & Electric",
+        "operator:wikidata": "Q1815011",
+        "operator:wikipedia": "en:Pacifi Gas and Electric Company",
```
1ec5, line 4570: This should be a link to “[Pacifi**c** Gas and Electric Company](https://en.wikipedia.org/wiki/Pacific_Gas_and_Electric_Company)”, not “Pacifi Gas and Electric Company”.

By the way, the company consistently uses an ampersand in its logo but spells out “and” in its name. This differs from some other companies, such as San Diego Gas & Electric.

arch0345: Thanks for catching this, it looks like this has already been fixed. I'll replace the ampersand with 'and'.

bhousel: For a bunch of these, we collect both versions from the OSM planet, but I tried to keep the more popular version.  We can check the counts in the `dist/collected/operators_all.json` file.  I don’t have a preference between “&” vs “and” but someone might.

PR #5227, opened 2021-08-10
Update substation.json

kjonosm:
update location sets
add wikidata
merge duplicates


arch0345:
I've resolved most of the conflicts, the only ones remaining are operators I merged in a previous commit. I'm fine either way if these are separated or merged.

bhousel:
> I'm fine either way if these are separated or merged.

Thanks! I resolved the conflicts, I think generally it's probably better to merge items unless there's a very good reason to keep them separate.

PR #5234, opened 2021-08-16
Update Canadian power operators info

wolfy1339:


arch0345:
Do Hydro Ottawa and Toronto Hydro only operate minor lines? If not, it would be better to add these operators to line.json instead.

wolfy1339:
They don't high voltage transmission or generation. They do distribution to customers.

Issue #4399, opened 2020-10-09
Add Egon


Adamant36:
Restaurant chain located in Norway and Sweden. 

https://en.wikipedia.org/wiki/Egon_(restaurants)
https://www.wikidata.org/wiki/Q1298438

xhika:
#4407

Adamant36:
Closed in #4407

PR #5237, opened 2021-08-16
Move Egon from fast food to restaurant

mathiash98:
Norwegian restaurant chain (not a fast_food chain) which serves lots of different type of foods:
https://egon.no/menyen

Relevant issue for when Egon was added https://github.com/osmlab/name-suggestion-index/issues/4399

Don't know if I should create an issue first of it this is good enough?

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -1409,20 +1409,6 @@
       }
     },
     {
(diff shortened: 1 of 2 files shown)


LaoshuBaby:
I think it is ok, but maybe the cuisine is too much, could you select some main cuisine?

PR #5240, opened 2021-08-16
Peppes Pizza Move to restaurant

mathiash98:
Sorry for accidently deleting PRs then reopening a new right away 😅 

Peppes Pizza is a restaurant chain where you can order pizza and have to wait for a while before getting it. It does also provide home delivery and takeaway


bhousel:
> Sorry for accidently deleting PRs then reopening a new right away 😅

Thats ok - thanks for the fix 👍

PR #5241, opened 2021-08-16
Big Bite Norwegian fast food sandwich chain

mathiash98:
71 stores in Norway, going over all stores now and adding to OpenStreetMap
https://www.bigbite.no/om-oss

https://www.wikidata.org/wiki/Q11960977
https://no.wikipedia.org/wiki/Big_Bite


arch0345:
Thanks @mathiash98 !

PR #4987, opened 2021-03-21
Fix 美而美 series brand

LaoshuBaby:
According to #4956 , just fixed a headache brand series problem.

Below are note about this.

 (written in Chinese, to feedback to that DWG menber supaplex, sure that he can understand)

Issue #4956, opened 2021-03-11
美而美 in Taiwan


bhousel:
@Supaplextw left some feedback on [my wikidata talk page](https://www.wikidata.org/wiki/User_talk:Bhousel#%E7%BE%8E%E8%80%8C%E7%BE%8E_in_Taiwan):

> 美而美 is a collective terms for brands previous use the name 美而美. Now 美而美 is four separated band that was operated by different companies. You could use Google Translate to figure it out on zh Wikipedia.

Is there anything actionable for us to do on the NSI side?
The brand they are referring to is "Mei and Mei"
https://nsi.guide/index.html?t=brands&k=amenity&v=restaurant#meiandmei-aafd63
https://www.wikidata.org/wiki/Q56197246

It sounds like we might be linked to the wrong thing, or this is actually a generic name to exclude?  Anyone have more context?

LaoshuBaby:
I will try to review this, thanks!

PR #5247, opened 2021-08-19
Taiwan Chain update

Supaplextw:
Mei & Mei related brand fix, there are actually several companies and several brans, not a single one.
Add the localized Burger King in Taiwan


LaoshuBaby:
Related #4956 #4987

感谢您对NSI完善的贡献

Issue #5149, opened 2021-07-19
MAPCO


maxerickson:
https://www.mapcorewards.com/store-locator

I guess there may be multiple related brands. Currently owned by https://en.wikipedia.org/wiki/Empresas_Copec and doesn't have a separate Wikidata/Wikipedia entry.

UKChris-osm:
Is Mapco just a gas station brand?

I've started a Wikidata for it: [Q107589462](https://www.wikidata.org/wiki/Q107589462)

maxerickson:
Their Twitter mentions several types of location: https://twitter.com/MAPCO

MuskanTripathi:
> Is Mapco just a gas station brand?
> 
> I've started a Wikidata for it: [Q107589462](https://www.wikidata.org/wiki/Q107589462)

Mapco is owned by Chile's COPEC, Mapco's competitive advantage is that it operates under seven brands: Mapco Express, Mapco Mart, East Coast, Fast Food and Fuel, Favorite Markets, Delta Express and Discount Food Mart. The chain's former owner, Delek US Holdings, sold the Mapco retail assets to the Santiago, Chile-based energy conglomerate in 2016. Mapco's c-stores offer a standard range of convenience items, including cigarettes, snacks, drinks and beer and, increasingly, proprietary made-to-order meals. Some units also feature a branded quick-service restaurant partner, with brands that include Subway, Quiznos and Blimpie. Most of the chain's stores are freestanding structures with attached gas stations that are generally located on heavily trafficked intersections.

aayush920:
Hi ! I would like to work on this  issue but I would like some guidance.

bhousel:
Hey @aayush920 welcome 👋 
We have some contributing information on the project wiki: https://github.com/osmlab/name-suggestion-index/wiki
You can also look on https://nsi.guide for items in our index which need a link to a wikidata item or which need a better location.

aayush920:
Hey @bhousel 
I have one query. It seems that mapco doesn't have a wikipedia page. Should I just leave it or add the wikipedia page of its parent company "Delek US".
Thanks in advance.

bhousel:
> I have one query. It seems that mapco doesn't have a wikipedia page. Should I just leave it or add the wikipedia page of its parent company "Delek US".

It's ok to skip it.  The `brand:wikidata` tag is the important tag, and the `brand:wikipedia` is optional.

aayush920:
[...]

PR #5255, opened 2021-08-20
added Mapco 

aayush920:


arch0345:
Closes #5149

data/brands/amenity/fuel.json
```diff
@@ -49,6 +49,21 @@
...
+      "locationSet": {
+        "include": ["001"],
+        "exclude": ["pl"]
+      },
+      "matchNames": ["mapco", "MAPCO"],
+      "tags": {
+        "amenity": "fuel",
+        "brand": "Mapco",
+        "brand:wikidata": "Q107589462",
+        "brand:wikipedia": "en:Mapco",
```
arch0345, line 63: There doesn't seem to be an article for Mapco in the English Wikipedia, only Wikidata tags are required. You could make a Wikipedia article, but note that it's more difficult than creating a Wikidata page.

data/brands/amenity/fuel.json
```diff
@@ -49,6 +49,21 @@
         "name": "7-Eleven"
       }
     },
+    {
+      "displayName": "Mapco",
+      "locationSet": {
+        "include": ["001"],
```
arch0345, line 55: From my understanding, Mapco only has locations in the US. You should change the locationSet to include 'us' and remove the exclude argument.

arch0345 (changes requested):
Thanks for your first addition to the name suggestion index! You should fix the locationSet to include 'us' rather than '001' (worldwide) and remove the exclude argument and also remove the brand:wikipedia tag. There doesn't seem to be a Wikipedia article for this brand yet and this is not a requirement for a brand to be added to the NSI.

aayush920:
Thank you for telling. I have done the needfull.

arch0345 (approved):

arch0345:
Thanks @aayush920!

PR #5256, opened 2021-08-21
add 7 for all mankind

m-hue:
chain of clothing stores in the US, selling denim.

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -39,6 +39,18 @@
       }
     },
     {
+      "displayName": "7 For All Mankind",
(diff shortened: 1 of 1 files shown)


arch0345:
Thanks @m-hue!

UKChris-osm:
Thank you for this @m-hue - I have changed the locationSet to world (_001_) for now as this brand seems available in many area's, and I've added a matchName for "_7fam_" which Wikipedia lists as being a common term [f1ed1c6](https://github.com/osmlab/name-suggestion-index/commit/f1ed1c68bb680f83b6a6c2faffbe5246c10a8cd8)

PR #5261, opened 2021-08-22
Create substation.json

kjonosm:
Add wikidata & match names

### DIFF
--- a/data/operators/power/substation.json
+++ b/data/operators/power/substation.json
@@ -4775,7 +4775,7 @@
     },
     {
       "displayName": "PSE",
-      "id": "pse-83e94a",
+      "id": "pseoperator-83e94a",
       "locationSet": {"include": ["pl"]},
       "matchNames": [
         "pse operator",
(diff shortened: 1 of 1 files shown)


data/operators/power/substation.json
```diff
@@ -8657,7 +8672,7 @@
     },
     {
       "displayName": "沖縄電力",
-      "id": "okinawaelectricpowercompany-85251b",
+      "id": "417df7-85251b",
```
LaoshuBaby, line 8675: 图片
Why so much Japanese substation lost their ```id``` ?
Another nodejs error?

bhousel: That is pretty weird - I think I'm going to just merge this and run build locally to see whether the ascii ids return?

bhousel: The Japanese ids were regenerated in 471365fa4 -  I dont know why they went away.

This code in `idgen.js` is supposed to generate the name part of the id by looking through several tag values:
https://github.com/osmlab/name-suggestion-index/blob/471365fa4b8da58aee949fc215d26d9acbc0bb94/lib/idgen.js#L29-L39

PR #5276, opened 2021-08-26
Amend cafe.json for brands in JP & TW

assanges:
Add/modifying brands that operate in Japan & Taiwan
±hoshinocoffee-0cf217 (Hoshino Coffee/ 星乃珈琲店)
+[new] Kebuke/ 可不可熟成紅茶
+[new] Hechalou/ 鶴茶樓
+[new] Truedan/珍煮丹 (トゥルーダン)

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -2637,6 +2637,24 @@
       }
     },
     {
+      "displayName": "可不可熟成紅茶",
+      "id": "kebuke-7d270d",
+      "locationSet": {"include": ["tw"]},
+      "tags": {
+        "amenity": "cafe",
+        "brand": "可不可熟成紅茶",
+        "brand:en": "Kebuke",
+        "brand:wikidata": "Q45572389",
+        "brand:zh": "可不可熟成紅茶",
+        "cuisine": "bubble_tea",
+        "name": "可不可熟成紅茶",
+        "name:en": "Kebuke",
+        "name:zh": "可不可熟成紅茶",
+        "short_name": "可不可",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/cafe.json
```diff
@@ -2789,9 +2807,10 @@
     },
     {
       "displayName": "星乃珈琲店",
-      "id": "hoshinocoffee-0cf217",
-      "locationSet": {"include": ["jp"]},
+      "id": "hoshinocoffee-7f2f5a",
+      "locationSet": {"include": ["jp", "tw"]},
       "tags": {
+        "alt_name:zh": "星乃咖啡店",
```
LaoshuBaby, line 2813: Why not use ```name:zh-Hans``` because this is written in Simplified Chinese?

LaoshuBaby: And write a ```name:zh-Hant=星乃珈琲店```  at the same time

assanges: Yes and no, actually `星乃珈琲店` is a Japanese Kanji word, 
it would be same Chinese characters `星乃咖啡店` as in both Traditional and Simplified Chinese. 
However, I kept the `name=星乃珈琲店` w/ `alt_name:zh=星乃咖啡店` since they use the same branding in TW.

`星乃珈琲店` 其實是日本漢字, 簡中/䌓中的字符實際上一樣是 `星乃咖啡店`, 
但它們上陸時候沿用所以才會選擇用 `name=星乃珈琲店` + `alt_name:zh=星乃咖啡店` 的組合

PR #5283, opened 2021-08-29
Change cuisine=kebap to cuisine=kebab

starsep:
Change cuisine=kebap to cuisine=kebab
According to taginfo 18122 vs 4 uses

https://taginfo.openstreetmap.org/tags/cuisine=kebab
https://taginfo.openstreetmap.org/tags/cuisine=kebap

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -2567,7 +2567,7 @@
       "tags": {
         "amenity": "fast_food",
         "brand": "Kebab King",
-        "cuisine": "kebap",
+        "cuisine": "kebab",
         "name": "Kebab King",
         "takeaway": "yes"
       }


LaoshuBaby:
Welcome to NSI community to contributor your local knowledge.
Looks like it is a typo problem, like ```brand:ja``` with ```brand:jp```

LaoshuBaby:
This must be a typo problem, because this brand named with ```Kebab King```! 😀

but we must say that those 4 point in the world are not modified by NSI, because this item haven't got a wikidata, so it will not be suggested by iDeditor.

图片

You can fix those 4 points manually.

LaoshuBaby:
图片
Oops……maybe we‘ve made big wrong……
Is ```kebap``` in some language (such as germany?) means ```kebab``` in English?

bhousel:
> Oops……maybe we‘ve made big wrong……
> Is `kebap` in some language (such as germany?) means `kebab` in English?

I think there are probably a bunch of ways to spell Kebab around the world, so the names can have a lot of different values, but they should all use the more standard OSM tag [`cuisine=kebab`](https://wiki.openstreetmap.org/wiki/Tag:cuisine=kebab)

PR #5300, opened 2021-08-31
Smyk: fix countries include

starsep:
Should be Poland, Romania, and Ukraine
NOT Poland, Russia, and United Kingdom


arch0345:
Thanks for fixing this @starsep

PR #5301, opened 2021-08-31
Selgros: also available in Poland, Romania, and Russia

starsep:
Source: https://en.wikipedia.org/wiki/Selgros


arch0345:
Thanks for these fixes!

PR #5310, opened 2021-09-02
Added Chinese to T&T Supermarket: Almost all stores post the Chinese

andrepoiy:
Pretty much all stores post the Chinese name before the English name, therefore adding the Chinese name before the English name on OSM would be more accurate.

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -5813,14 +5813,17 @@
       }
     },
     {
-      "displayName": "T&T Supermarket",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
图片
Yes, you are right, according to wikimedia, almost everyone have this 3 Chinese characters.

At Multilingual names pages haven't found Canada(https://wiki.openstreetmap.org/wiki/Multilingual_names#Issues), so I look for this https://wiki.openstreetmap.org/wiki/Canadian_tagging_guidelines#Naming and haven't some multilingual restrict

So I think this changes is OK.

PR #5313, opened 2021-09-03
Update beauty.json

kjonosm:
Move some nail studios to the "named" section. These don't seem to be chains or franchises but rather small independent often owner-operated shops.


LaoshuBaby:
{{support}} I agree, because those name, in every-language's google can't find a certain brand chain, those name are just common beautiful words, let people feel good when talking with those brand.

LaoshuBaby (approved):

PR #5315, opened 2021-09-05
Discard 'anganwadi' and similarly spelled brands

arch0345:
[Anganwadi](https://en.wikipedia.org/wiki/Anganwadi) seems to be more of a generic name for childcare centers than a brand name. I'm assuming that 'anganavadi' and 'anganawadi' are just variations of Wikipedia's spelling.


kjonosm:
I merged them all into `brand=Anganwadi`, see https://en.wikipedia.org/wiki/Anganwadi

arch0345:
Ok, I guess this makes sense as they're all affiliated with the Indian government.

PR #5316, opened 2021-09-05
California Welcome Centers

1ec5:
Added [California Welcome Centers](https://www.visitcalifornia.com/welcome-centers/), a network of independently operated welcome centers across the state that has coherent branding (including standardized freeway signage).

This is the first `tourism=information` entry in the repository. `npm run build` wants to bring in a lot of things that don’t look like chains, so I’ve omitted those entries from this PR.

### DIFF
--- a/data/brands/tourism/information.json
+++ b/data/brands/tourism/information.json
@@ -1,0 +1,24 @@
+{
+  "properties": {
+    "path": "brands/tourism/information",
+    "preserveTags": ["^name"],
+    "exclude": {"generic": ["^information$"]}
+  },
+  "items": [
+    {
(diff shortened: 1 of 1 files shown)


data/brands/tourism/information.json
```diff
@@ -0,0 +1,24 @@
+{
+  "properties": {
+    "path": "brands/tourism/information",
+    "preserveTags": ["^name"],
+    "exclude": {"generic": ["^information$"]}
```
bhousel, line 5: This looks good to me 👍 
If you want to tell the build script not to bring in all the other `tourism=information` things that we've collected, you can add this line up here:
```js
    "skipCollection": true,
```

PR #5334, opened 2021-09-12
Added university of A Coruña to charging_station.json

lucumon:
Also, I want to add it as an operator for "building=university", but I don't know if it's possible.

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -3419,6 +3419,16 @@
       }
     },
     {
+      "displayName": "Universidade da Coruña",
+      "locationSet": {"include": ["es"]},
+      "tags": {
+        "amenity": "charging_station",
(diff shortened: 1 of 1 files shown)


arch0345:
After running an [overpass query](https://overpass-turbo.eu/?w=%22operator%22%3D%22Universidade+da+Coru%C3%B1a%22+global&R), there doesn't seem to be any EV charging stations with this operator. Would any of the existing operators in charging_station.json be applicable to the EV charging stations at Universidade da Coruña?

The NSI currently doesn't have a building/university category, but it looks like other people in the past have been interested in adding this and amenity/university, such as in #4779.

lucumon:
The university has some charging stations operated by themselves around their campuses, but they aren't still mapped, I think. I may start adding them today.

bhousel:
thanks @lucumon & welcome 👋

PR #5336, opened 2021-09-13
Discard 'ABC Store'

arch0345:
ABC Store is a generic name used in [Alcohol Beverage Control (ABC) states](https://en.wikipedia.org/wiki/Alcoholic_beverage_control_state).

> Most of these states have an "Alcoholic Beverage Control" (ABC) board and run liquor stores called ABC stores or state stores.

### DIFF
--- a/data/brands/shop/alcohol.json
+++ b/data/brands/shop/alcohol.json
@@ -6,7 +6,8 @@
         "^((разливное|живое) )?пиво$",
         "^(alcohol|beer|liquor)( store)?$",
         "^(botiller|licorer)[ií]a$"
-      ]
+      ],
+      "named": ["^abc store$"]
     }
   },
   "items": [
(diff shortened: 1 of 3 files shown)


bhousel:
Sounds good to me 👍 

I did check this on overpass:
Screen Shot 2021-09-13 at 10 06 01 AM

They are mostly clustered in North Carolina and Alabama.
These are branded inconsistently, so discarding the "ABC Store" name sounds ok.

If these states ever do collect their state run liquor stores under a common brand, we could add them to the index like how we have [Virginia](https://nsi.guide/index.html?t=brands&k=shop&v=alcohol#virginiaabc-2740a5), [Pennsylvania](https://nsi.guide/index.html?t=brands&k=shop&v=alcohol#finewineandgoodspirits-2740a5), or [New Hampshire](https://nsi.guide/index.html?t=brands&k=shop&v=alcohol#newhampshireliquorandwineoutlet-2740a5) liquor stores.

PR #5348, opened 2021-09-17
Update interior_decoration.json

alessandrocarmeli:
coin casa added (IT)

### DIFF
--- a/data/brands/shop/interior_decoration.json
+++ b/data/brands/shop/interior_decoration.json
@@ -7,14 +7,15 @@
   },
   "items": [
       {
-      "displayName": "Coin Casa",
+      "displayName": "Coincasa",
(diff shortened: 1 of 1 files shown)


arch0345:
Thanks for adding this brand! However, based on the wikipedia page linked, it looks like the name of this brand is stylized as 'Coincasa'. Also, an [overpass query](https://overpass-turbo.eu/?w=%22name%22%3D%22Coincasa%22+global&R) shows that most of these are tagged as shop/furniture and shop/houseware. If shop/interior_decoration is a better tag to use than these, you should add matchTags so that shops tagged with these values will also get updated. To do this, add a line under locationSet with:
`matchTags: ["shop/furniture", "shop/houseware"],`

carmeliale:
Ok, i've got added machtags an correction

Issue #5275, opened 2021-08-25
Adding `brand/operator:website` when running `npm run wikidata`


arch0345:
I noticed that [```brand:website```](https://taginfo.openstreetmap.org/keys/brand%3Awebsite) is a somewhat prominent tag for brand websites. I was wondering if this tag could be automatically added from the 'official website' statement of the respective wikidata page similarly to how brand:wikipedia tags are added.

peternewman:
Hopefully it goes without saying, but we shouldn't set `website` as that's often used for a link to a particular store.

What would be really cool with all these things is if something could automatically enrich the data, so you could use Overpass Turbo for example, or the OSM API, either directly or in an enriched mode where these fields would be set if missing based on the impliedTags values. So people get the best of all worlds.

Adamant36:
> Hopefully it goes without saying, but we shouldn't set website as that's often used for a link to a particular store.

Well I don't support this change, I don't think that would be an issue because the `website` field would still be empty if `brand:website` was auto filled.

arch0345:
I'm going to close this because, as @bhousel mentioned, you could just get this information from the wikidata page, and a more relevant website pertaining to that specific location might already be in the `website` tag

PR #5351, opened 2021-09-18
Add Pijalnia Wódki i Piwa (bar chain)

starsep:


data/brands/amenity/bar.json
```diff
@@ -42,6 +42,19 @@
...
       }
     },
+    {
+      "displayName": "Pijalnia Wódki i Piwa",
+      "id": "pijalniawodkiipiwa-ca2c56",
+      "locationSet": {"include": ["pl"]},
+      "tags": {
+        "amenity": "bar",
+        "brand": "Pijalnia Wódki i Piwa",
+        "brand:website": "http://pwip.com.pl",
```
LaoshuBaby, line 52: this can be fetched from ```brand:wikidata``` so I don't think this is needed

LaoshuBaby (changes requested):

LaoshuBaby:
> I feel like the OSM community preference might be trending in the other direction... We often get feedback that brand:wikidata is an important tag but brand:wikipedia is redundant and we could probably stop adding it. (openstreetmap/iD#8620, #4672 #5192 (comment) ).
> 
> So I don't think we should add brand:website or other things that can just be fetched from the wikidata page (unless there was a wider discussion on OSM's tagging mailing list and someone specifically asked us to start adding this tag for a reason).

You can see #5275 for more detail.

data/brands/amenity/bar.json
starsep: I removed `brand:wikipedia` and `brand:website`
I added website to wikidata item

LaoshuBaby (approved):

PR #5356, opened 2021-09-18
Update pub.json

alessandrocarmeli:
[BEFeD Brew Pub] added

### DIFF
--- a/data/brands/amenity/pub.json
+++ b/data/brands/amenity/pub.json
@@ -23,6 +23,15 @@
     }
   },
   "items": [
+     {
+      "displayName": "BEFeD Brew Pub",
+      "locationSet": {"include": ["it"]},
+      "tags": {
+        "amenity": "pub",
+        "brand": "BEFeD Brew Pub",
+        "matchTags" : ["name/Befed"],
+      }
+    },
     {
       "displayName": "Belhaven Pubs",
       "id": "belhavenpubs-eaed91",
--- a/data/brands/amenity/pub.json
+++ b/data/brands/amenity/pub.json
@@ -26,10 +26,10 @@
      {
       "displayName": "BEFeD Brew Pub",
       "locationSet": {"include": ["it"]},
+      "matchTags" : ["name/Befed"],
       "tags": {
         "amenity": "pub",
-        "brand": "BEFeD Brew Pub",
-        "matchTags" : ["name/Befed"],
+        "brand": "BEFeD Brew Pub"
       }
     },
     {


LaoshuBaby:
This PR have 2 question

1. ```matchTags``` or ```matchNames``` should outer the ```tags```, this is for NSI program's identify, and content in ```tags``` is used for mapping a point/area/relation
2. In JSON, the latest item in a array don't need to add comma at last.

I've commit a patch to it.

LaoshuBaby:
And anothetr thing I want to pointed is if this brand haven't got a wikidata, it won't be suggested in iDeditor.

alessandrocarmeli:
strange, another contributor had strongly advised me to use one of these tags!

LaoshuBaby:
> 
> 
> strange, another contributor had strongly advised me to use one of these tags!

Yes! matchNames is strongly advised because some brand may have different call name in different area.

matchTags should be added if necessary.

Your problem is like this:

**CORRECT**

```
     {
      "displayName": "BEFeD Brew Pub",
      "matchTags" : ["name/Befed"],
      "tags": {
        "amenity": "pub"
      }
    },
```

** WRONG **

```json
     {
      "displayName": "BEFeD Brew Pub",
      "tags": {
        "amenity": "pub",
        "matchTags" : ["name/Befed"]
      }
    },
```

The ```matchTags``` should outer ```tags```

LaoshuBaby:
And another mistake, I haven't found at first

https://github.com/osmlab/name-suggestion-index/commit/2720aaa67c67fe3cdddf8b81eecacbd771ec11b9

is fixed in this commit.

Because ```Befed``` is the brand name, and you only list this one name, you don't need to use ```matchNames``` in this situation.

you only need to put this

```
"name":"Befed",
```

alessandrocarmeli:
but danger: `"name":Befed` is incrorrect, correct values is `name:BEFeD`

LaoshuBaby:
> 
> 
> but danger: `"name":Befed` is incrorrect, correct values is `name:BEFeD`

Oh, this is from your commit, you can edit it in there

https://github.com/osmlab/name-suggestion-index/blob/3463454e5ecc79ee1cb3771e476742ee34d5a4e7/data/brands/amenity/pub.json#L33

This PR had already been merged so you may need to re create a PR.

bhousel:
I changed this a little bit.. Since the brand calls themselves "BEFED" I set the name to that in ec49edc17

Screen Shot 2021-09-18 at 12 24 04 PM

PR #5358, opened 2021-09-18
Update shoes.json

alessandrocarmeli:
[Bata] added

### DIFF
--- a/data/brands/shop/shoes.json
+++ b/data/brands/shop/shoes.json
@@ -9,6 +9,16 @@
     }
   },
   "items": [
+      {
+      "displayName": "Bata",
+      "tags": {
+        "brand": "Bata Corporation",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
```"locationSet"``` is needed to restrict the effected country/region for this brand.

If don't know which country to input or this is a global brand, you can use ```"locationSet": {"include": ["001"]},```

LaoshuBaby:
But I still need to draft this PR because this brand have already in NSI

https://github.com/osmlab/name-suggestion-index/blob/d0be4b7029eaec1484527244b58d6df3c046742a/data/brands/shop/shoes.json#L134-L142

You can do something to make those 2 entry different in some prospect

bhousel:
Ah yeah this should probably be closed if it's already in the index.

PR #5363, opened 2021-09-19
Update atm.json

alessandrocarmeli:
postamat (poste italiane) added


bhousel:
Thanks @alessandrocarmeli , this looks good but I will fix the missing commas ',' causing the error.

PR #5364, opened 2021-09-19
Update cosmetics.json

alessandrocarmeli:
bottega verde italy added (MatchTags correct???)

### DIFF
--- a/data/brands/shop/cosmetics.json
+++ b/data/brands/shop/cosmetics.json
@@ -10,13 +10,13 @@
     {
       "displayName": "Bottega Verde",
       "locationSet": {"include": ["it"]},
+      "matchTags": ["shop/perfumery","shop/chemist"],
       "tags": {
         "brand": "Bottega Verde",
         "brand:wikidata": "Q894870",
         "brand:wikipedia": "it:Bottega Verde",
         "name": "Bottega Verde",
         "shop": "cosmetics"
-        "matchTags": ["shop/perfumery", ["shop/gift", "shop/chemist", "shop/herbalist", "shop"],
       }
     },
     {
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
matchTags and matchNames can't be write in tags, they should been put out of tags

```json
    {
      "displayName": "Bottega Verde",
      "matchTags": ["shop/perfumery","shop/chemist"],
      "tags": {
        "brand": "Bottega Verde",
        "brand:wikidata": "Q894870",
        "brand:wikipedia": "it:Bottega Verde",
        "name": "Bottega Verde",
        "shop": "cosmetics"
      }
    },
```

bhousel:
> (MatchTags correct???)

not quite! check out the docs here:  
https://github.com/osmlab/name-suggestion-index/wiki/Contributing#matchnamesmatchtags-optional

First thing, the `matchTags` property needs to go outside of the `tags` block. (everything _inside_ `tags` are the tags this feature should have in OSM)

Second thing, `shop/cosmetics` is already in a match group with all of these other categories, so you don't need to include them:
```js
    "convenience": [
      "shop/beauty",
      "shop/chemist",
      "shop/convenience",
      "shop/cosmetics",
      "shop/grocery",
      "shop/newsagent"
    ],
```

(we can add `shop/perfumery` to this list)

LaoshuBaby:
> (we can add `shop/perfumery` to this list)

I noticed this gruop before, this time the ```shop/chemist``` is my careless,

but for ```shop/perfumery``` we can wait for another entry need this matchTags ?

bhousel:
> but for `shop/perfumery` we can wait for another entry need this matchTags ?

I didn't realize this until now but we do have a category for [shop/perfumery](https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/shop/perfumery.json) and almost all the entries there have matchTags for `shop/beauty`, `shop/cosmetics`, `shop/chemist` - so it makes sense that this tagpair should be in the match group too.

I can correct it now.

PR #5365, opened 2021-09-19
Update bank.json

alessandrocarmeli:
Banca Popolare di Cividale

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -20,6 +20,19 @@
     }
   },
   "items": [
+      {
+      "displayName": "Banca Popolare di Cividale",
+      "locationSet": {"include": ["it"]},
+      "tags": {
+        "amenity": "bank",
+        "brand": "Banca Popolare di Cividale",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/bank.json
```diff
@@ -20,6 +20,19 @@
...
+      {
+      "displayName": "Banca Popolare di Cividale",
+      "locationSet": {"include": ["it"]},
+      "tags": {
+        "amenity": "bank",
+        "brand": "Banca Popolare di Cividale",
+        "brand:wikidata": "Q15639942",
+        "brand:wikipedia": "it:Banca Popolare di Cividale",
+        "name": "Banca Popolare di Cividale",
+        "matchNames": ["CiviBank", "Civi Bank"],
```
m-hue, line 32: Hello @alessandrocarmeli,
thanks for your contribution. Could you please move the matchNames out of the tags array? They should sit o the same level as displayName, locationSet, tags etc.

Thank you!

m-hue (changes requested):

PR #5369, opened 2021-09-19
Update pub.json

alessandrocarmeli:
befed edited

### DIFF
--- a/data/brands/amenity/pub.json
+++ b/data/brands/amenity/pub.json
@@ -24,13 +24,15 @@
   },
   "items": [
     {
-      "displayName": "BEFED",
+      "displayName": "BEFeD",
       "id": "befed-1d1bef",
       "locationSet": {"include": ["fx", "it"]},
+      "matchNames": ["Befed"]
       "tags": {
         "amenity": "pub",
-        "brand": "BEFED",
-        "name": "BEFED"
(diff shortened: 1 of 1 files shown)


arch0345:
My bad; I mixed up Wikimedia Commons with the English Wikipedia. In the future, you should upload fair use logos to the [English Wikipedia](https://en.m.wikipedia.org/wiki/Wikipedia:Non-free_content) instead as [Wikimedia Commons doesn't allow fair use images](https://commons.m.wikimedia.org/wiki/Commons:Fair_use).

data/brands/amenity/pub.json
```diff
@@ -24,13 +24,15 @@
   },
   "items": [
     {
-      "displayName": "BEFED",
+      "displayName": "BEFeD",
       "id": "befed-1d1bef",
       "locationSet": {"include": ["fx", "it"]},
+      "matchNames": ["Befed"]
```
LaoshuBaby, line 30: don't need this matchNames because it can match both uppercase and lowercase

bhousel:
As I said here https://github.com/osmlab/name-suggestion-index/pull/5356#issuecomment-922335761 the brand calls themselves "BEFED" not "BEFeD", but whatever. 🤷

PR #5370, opened 2021-09-19
Update chemist.json

alessandrocarmeli:
caddy's added

### DIFF
--- a/data/brands/shop/chemist.json
+++ b/data/brands/shop/chemist.json
@@ -10,6 +10,18 @@
     }
   },
   "items": [
+     {
+      "displayName": "Caddy's",
+      "locationSet": {"include": ["it"]},
+      "matchTags": ["shop/conveneience", "shop/houseware", "shop/cosmetics"],
+      "tags": {
+        "brand": "Caddy's",
+        "brand:wikidata": "Q108604630",
+        "name": "Caddy's",
+        "old_name": "Cad",
+        "shop": "chemist"
+      }
+    },
     {
       "displayName": "101 Drogerie",
       "id": "101drogerie-a99570",


data/brands/shop/chemist.json
```diff
@@ -10,6 +10,18 @@
     }
   },
   "items": [
+     {
+      "displayName": "Caddy's",
+      "locationSet": {"include": ["it"]},
+      "matchTags": ["shop/conveneience", "shop/houseware", "shop/cosmetics"],
```
m-hue, line 16: Hello @alessandrocarmeli,
thanks for your submission. Could you please correct the spelling on the matchTags (shop/conve**nie**nce).
Also, are you sure that the shop/houseware is appropriate? I'm not saying it isn't, I just wouldn't have expected it for a place that's primarily tagged as a chemist.
Thanks,

m-hue (changes requested):

data/brands/shop/chemist.json
```diff
@@ -10,6 +10,18 @@
...
+      "tags": {
+        "brand": "Caddy's",
+        "brand:wikidata": "Q108604630",
+        "name": "Caddy's",
+        "old_name": "Cad",
```
tas50, line 21: Is this just to tag for a rebranding?

data/brands/shop/chemist.json
alessandrocarmeli: I'm sure

data/brands/shop/chemist.json
```diff
@@ -10,6 +10,18 @@
     }
   },
   "items": [
+     {
+      "displayName": "Caddy's",
+      "locationSet": {"include": ["it"]},
+      "matchTags": ["shop/convenience", "shop/houseware", "shop/cosmetics"],
```
LaoshuBaby, line 16: too much， don't need to include those matchTags,
you can refer to the matchgroups and only add a matchTag that don't in the matchgroup

LaoshuBaby (changes requested):

data/brands/shop/chemist.json
bhousel: yah looks like a 2017 rebrand: https://ovostudio.it/portfolio/cad-caddys
I'll move this to a matchName, since that was a while ago.

PR #5373, opened 2021-09-20
adding fitness centres

waso99:
adding fitness centres


bhousel:
Looks perfect, thank you! 👍

Issue #4769, opened 2020-12-04
Park operators


michaelblyons:
What's the chance of getting an `operators/leisure/park.json`? Is it as simple as adding that file?

bhousel:
That's a good suggestion.  It's not hard to do, but we aren't yet collecting operators for leisure/park.

You can see in [operators_all.json](https://github.com/osmlab/name-suggestion-index/blob/main/dist/collected/operators_all.json) the sorts of things we do collect already.  Like police and fire services, many larger cities park and school systems might be worth adding items for.

michaelblyons:
USA has 62 national parks (my main target), as well as a spread of state parks. This in additional to larger counties and/or municipalities.

UKChris-osm:
Would this sort of thing be "_operator/boundary/national_park.json_"?

michaelblyons:
> Would this sort of thing be "_operator/boundary/national_park.json_"?

That is actually a good point. I did not know about those. However... I now would like _both._ (I'm difficult like that. 😉) The US National Park Service controls some huge swathes of land that belong in `boundary=national_park`, but also some [smaller bits](https://www.openstreetmap.org/relation/11158003#map=14/38.8867/-77.0344) that are tagged as `leisure=park`.

Also, the more I dig into this, the weirder it gets. There are 62 "National Parks," but there are at least 400 places in the "National Park System" that seem to be operated by NPS.

bhousel:
> That is actually a good point. I did not know about those. However... I now would like _both._ (I'm difficult like that. 😉) The US National Park Service controls some huge swathes of land that belong in `boundary=national_park`, but also some [smaller bits](https://www.openstreetmap.org/relation/11158003#map=14/38.8867/-77.0344) that are tagged as `leisure=park`.

This is ok, we can support both here.  Just like "Tesco" can be a brand in both `amenity/fuel` or a `shop/supermarket`, "National Park Service" can be an operator in `leisure/park` or a `boundary/national_park`.

UKChris-osm:
Maybe a "_matchGroup_" for both would be the best thing?

TheAdventurer64:
This would also be good for parks owned by the government. I have a tagging schema for Honolulu parks which I think would be good in this situation: 

https://www.openstreetmap.org/way/218680183

https://www.openstreetmap.org/way/161726419

zekefarwell:
[...]

Issue #4765, opened 2020-12-03
Transit stations, stops


bhousel:
tl;dr Adding items for transit stations is going to be tricky or maybe impossible.

---
#### Background

A lot of this project's code works on the assumption that we can identify a thing by a `key, value, name` tuple.
* key = amenity
* value = cafe
* name = star-bucks  (people type all kinds of stuff into OSM)

When we say "name", we don't anymore mean the literal `name` tag, but it can be a name-like value like we find in other tags like `brand`, `operator`, `flag:name`, or whatever.  But the gist of it is that we can somehow match 1 key, 1 value, and 1 "name" to determine what a thing _is_.

---
#### The problem

Transit stations unfortunately don't all work like other tags due to OSM's [public_transport tagging](https://wiki.openstreetmap.org/wiki/Key:public_transport) convention.

Historically you could do this:
* key = amenity
* value = bus_station
* network = SEPTA (we'll use this as the "name")

And tagging like that is still common in OSM, but it has been deprecated in favor of the [Public Transport v2](https://wiki.openstreetmap.org/w/index.php?oldid=625726) tag schema.

* public_transport = platform
* bus = yes

That's 2 k's and 2 v's.  

So my problem is - I don't know how to support this, when a feature type can't be inferred from a single key,value pair.

We could support features right now that use legacy tagging, but I know that people aren't always leaving the legacy `amenity=bus_station` tags around, and some kinds of stations (like light rail and subway) require multiple tags anyway.

---
#### Steps forward
It sounds like we will need move away from using OSM key/value pairs to identify things.
I'm just thinking out loud here at what I might need to change:

* Rewrite parts of the `build_index.js` code so that `k/v` doesn't literally mean k/v, but instead it's just a thing that we make up.  For example, there is code to add an `amenity=cafe` tag if you happen to add an item in that file missing that tag.
* Rewrite the matcher to handle not just a single k/v but I guess an array k/v tag pairs.
* Rewrite the id generator, hopefully preserving all the existing generated ids that hashed a single k/v
* Rewrite the parts of nsi.guide?  It could be less "OSM-tag-centric" and just say "Cafe" instead of "amenity/cafe"
* So we might need a file somewhere of all the mappings between OSM Tags and whatever we consider categories to be.
* Maybe we just go all-in with wikidata ?  Q2175765 is a better identifier than `public_transport=station+tram=yes`

bhousel:
[...]

PR #5374, opened 2021-09-21
Create leisure/park and other public land operator categories

arch0345:
Re: #4769 

Not sure how I would be able to create separate entries to national monuments and other types of protected areas operated by the same department. For example, when adding National Wildlife Refuges (operated by FWS), running build gives an error as the two have the same operator. If this isn't possible I could just make these entries less specific (National Park Service, Bureau of Land Management, etc.).

### DIFF
--- a/data/operators/boundary/aboriginal_lands.json
+++ b/data/operators/boundary/aboriginal_lands.json
@@ -1,0 +1,22 @@
+{
+  "properties": {
+    "path": "operators/boundary/aboriginal_lands",
+    "skipCollection": true,
+    "exclude": {"generic": ["^aboriginal lands$"]}
+  },
+  "items": [
+    {
+      "displayName": "Bureau of Indian Affairs",
+      "id": "bureauofindianaffairs-9a1c88",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "boundary": "aboriginal_lands",
+        "operator": "Bureau of Indian Affairs",
+        "operator:short": "BIA",
+        "operator:type": "public",
+        "operator:wikidata": "Q1010563",
+        "operator:wikipedia": "en:Bureau of Indian Affairs"
+      }
+    }
+  ]
+}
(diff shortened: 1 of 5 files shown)


bhousel:
> For example, when adding National Wildlife Refuges (operated by FWS), running build gives an error as the two have the same operator. If this isn't possible I could just make these entries less specific (National Park Service, Bureau of Land Management, etc.).

I think for right now, making them less specific is the best we can do..  This relates a bit to #4765, because I have a similar problem with rail stations.

Currently, each item in NSI must be uniquely identified by a `key`,`value`,`name`, `location` tuple:
`key = boundary`
`value = protected_area`
`name = United States Fish and Wildlife Service`
`locationSet = { include: { ['us'] } }`

This is why you can't add an additional FWS item in that same `boundary/protected_area` file.. (If I'm reading correctly, you'd want one item with `protect_title=National Monument` and another item with `protect_title=National Refuge`)

bhousel:
Looks good to me, thanks @arch0345 !  
It will be great to have national parks and other similar features matched to `operator:wikidata` tags 👍

bhousel:
I merged this and did a release today.. It is pretty cool to immediately see National Parks in RapiD  🎉 

Screen Shot 2021-10-01 at 1 13 22 PM


Unfortunately not all of of these features will show up as iD presets yet:
```
Warning - no iD source preset found for operators/boundary/aboriginal_lands
Warning - no iD source preset found for operators/boundary/national_park
Warning - no iD source preset found for operators/boundary/protected_area
```

The Name-suggestion-index project is dependent on the presets in https://github.com/openstreetmap/id-tagging-schema, which is essentially a dead project at this point.  

[I have offered to maintain it](https://github.com/openstreetmap/id-tagging-schema/pull/134#issuecomment-850419046), but it sounds like the OSMF still wants to tightly control who is allowed to maintain things.  If Quincy isn't around to do it, the iD related projects should really be returned to me.

arch0345:
That really sucks. At least parks and nature reserves work for now; looking forward to adding more park operators!

PR #5390, opened 2021-09-27
Add Albertsons Pharmacy

tas50:
They run their own branded in store pharmacy chain all over the US.

### DIFF
--- a/data/brands/amenity/pharmacy.json
+++ b/data/brands/amenity/pharmacy.json
@@ -82,6 +82,19 @@
       }
     },
     {
+      "displayName": "Albertsons Pharmacy",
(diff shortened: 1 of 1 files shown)


arch0345:
I noticed that some of these in the Portland area are signed as Sav-on pharmacies, do you know if these should be added as a separate entry or as a matchName?

tas50:
I noticed that there's a bunch that say Sav-On on the signs around here. I wasn't sure if that was the branding inside too. The website lists those locations as just Albertsons. It's probably worth taking a drive to see what they brand them as inside.

tas50:
I'm going to merge this since I know there are Albertsons branded pharms. The real question is do we need a Sav-On entry as well.

PR #5401, opened 2021-09-28
Adding some telecom companies

waso99:
Adding some telecom companies


arch0345:
Thanks @waso99!

waso99:
Thanks

PR #5407, opened 2021-09-29
Update bakery.json

alessandrocarmeli:
Il pane quotidiano


arch0345:
Note that this won't be made into a preset until it gets a brand:wikidata tag

PR #5408, opened 2021-09-30
Move North Face to shop/outdoor

peternewman:


arch0345:
At least in the US, The North Face mainly sells clothes. Other similar outdoor clothing brands such as Columbia Sportswear are also in shop/clothes when they also sell other items. I think shop/outdoor is meant more for brands like REI, Bass Pro Shops, and Cabela's (looks like these aren't in the UK?) that sell more items such as kayaks, fishing/hunting gear, etc.

tas50:
I agree with @arch0345 on this one. It's primarily a clothing store with a small amount of extra camping gear depending on location. An outdoor store needs to be focused on outdoor equipment. 

https://wiki.openstreetmap.org/wiki/Tag:shop%3Doutdoor

peternewman:
> At least in the US, The North Face mainly sells clothes.

I did start by just adding match tags and left it in shop/clothes, then changed my mind.

> Other similar outdoor clothing brands such as Columbia Sportswear are also in shop/clothes when they also sell other items.

Do you feel that's correct, or are they mis-tagged too?

> I think shop/outdoor is meant more for brands like REI, Bass Pro Shops, and Cabela's (looks like these aren't in the UK?) that sell more items such as kayaks, fishing/hunting gear, etc.

Yeah those places aren't in the UK. So apart from occasionally Costco, nowhere on the high street here really sells kayaks. Kathmandu and Regatta are in a similar place, Regatta may have a bit more camping stuff than the others.

So do we feel an outdoor shop needs to have things like tent pegs on their own, stoves, cooking pots? i.e. could we come up with some examples that actually match an outdoor shop in the UK?

Maybe I should revert this and just add match tags to this and Columbia and whatever else is in clothing but does more?

arch0345:
> Do you feel that's correct, or are they mis-tagged too?
I feel that this is correct as just like North Face, Columbia is mainly focused on their apparel.

> So do we feel an outdoor shop needs to have things like tent pegs on their own, stoves, cooking pots? i.e. could we come up with some examples that actually match an outdoor shop in the UK?
Looking at shop/outdoor on [nsi.guide](https://nsi.guide), there are some British counterparts to the brands I mentioned, such as Black's, [Cotsworld Outdoor](https://www.wikidata.org/wiki/Q5175488), [Go Outdoor](https://www.wikidata.org/wiki/Q75293941), and [Mountain Warehouse](https://www.wikidata.org/wiki/Q6925414).

Adamant36:
I agree with leaving this as a clothing store. One thing that might help is adding the clothes=outdoor tag though. If people know it's an outdoor clothing store then they will also probably think they sell some outdoor accessories. Most outdoor clothing stores usually do.

UKChris-osm:
I'd leave this as "_shop=clothes_", and add "_clothes=outdoor_" as suggested by @Adamant36. I'd probably remove the "_alt_name_" as well as I don't think dropping "_the_" is really an alternative name.

bhousel:
Sounds like the consensus on this PR is to leave as-is.. 

@UKChris-osm said:
> I'd leave this as "shop=clothes", but add "clothes=outdoor" as suggested by @Adamant36. I'd probably remove the "alt_name" as well as I don't think dropping "the" is really an alternative name.

I agree - in 7494967c9 I did this ☝️

peternewman:
Hi all,

Sorry for going silent on this PR, I wanted to try and get inside my local North Face which took longer than planned for various reasons (I've actually still failed but had a good look through the window at least).

I'd agree with the consensus here, I've opened #5671 along similar lines to resolve the outstanding issue with it not being offered when tagged in an alternative way.

PR #5409, opened 2021-09-30
Add more locations for Jack Wolfskin

peternewman:
Based on some guesswork, some searching and checking their site.


peternewman:
I can't find this one on their site, and there may be other countries I've missed too:
https://www.openstreetmap.org/node/3987606699

arch0345:
Looks good @peternewman!

PR #5411, opened 2021-09-30
Add Yeme

andrewshadura:
Yeme is a Slovak supermarket chain (currently only operates 5 shops in Bratislava).


andrewshadura:
Sorry, I used a wrong wikidata item; the correct one is https://www.wikidata.org/wiki/Q108757006.

PR #5412, opened 2021-09-30
Add Yeme

andrewshadura:
Yeme is a Slovak supermarket chain (currently only operates 5 shops in Bratislava; a shop in Nitra is about to open).

A fixed iteration of PR #5411 which was closed without giving enough time to actually fix the error; GitHub doesn’t seem to allow updating a PR in such cases.

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -6521,6 +6521,17 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @andrewshadura and welcome 👋 
Sorry for closing the original issue so quickly..

Issue #4717, opened 2020-11-20
Have build_wikidata.js script fix broken *:wikipedia tags?


bhousel:
Following up from #4716 - if we can add missing `*:wikipedia` tags, we can also check all of them and fix any that are broken.

UKChris-osm:
One other thing I did wonder about with "_build_wikidata.js_", with the web being more of an "_SSL_" platform these days, and with so many web sites listed as "_non-https_" on Wikidata which have since moved to "_https_", could the build script highlight any "_non-https_" web sites, or even auto-check if a redirect to "_https_" exists?

bhousel:
> One other thing I did wonder about with "_build_wikidata.js_", with the web being more of an "_SSL_" platform these days, and with so many web sites listed as "_non-https_" on Wikidata which have since moved to "_https_", could the build script highlight any "_non-https_" web sites, or even auto-check if a redirect to "_https_" exists?

I think it's kind of out of scope for this project to clean up links in Wikidata, but do I agree that it's worth doing.  Someone could write a script to go through `dist/wikipedia.json` and check the links.  Or even a SPARQL query to check all the official_website links in Wikidata.  A process like that might be a thing that exists already - I'm less familiar with how things work on the Wikidata side.

PR #5413, opened 2021-10-02
Added secure SSL check to nsi.guide

UKChris-osm:
This PR adds a small check to each NSI entry to see if the URL listed on the Wikidata is secure or not. If it isn't, it's highlighted in red. This can be seen using a test site [https://nsi.haignet.co.uk/](https://nsi.haignet.co.uk/)

I asked about this a little while ago [here](https://github.com/osmlab/name-suggestion-index/issues/4717#issuecomment-731568841), and thought I would give it a go myself. It's only a basic check, testing to see if the homepage link starts with "_https://_"

Not all web sites are SSL these days, but this little addition invites people to check the Wikidata to see if an SSL version of the web site is available, and to update the Wikidata if it is. #4717


bhousel:
Screen Shot 2021-10-04 at 10 20 29 AM

I'm ok with the idea of displaying the secure vs insecure websites differently, but I don't really consider this a "social link" and I think it looks kind of out of place where it is. 

Can you move the icon up to be right before where the URL is displayed?  I think proximity to the URL might better signal what the icon is for.

UKChris-osm:
I'll give it a go.

I placed the icon with the other social links as it seemed like the simplest way to code it in to the existing application.

docs/style.css
```diff
@@ -339,6 +339,13 @@ tbody td.wikidata div.viewlink {
   color: #2a2a2a;
   background: #88c;
 }
+
+/* SSL lock / unlock icons */
+.viewlink svg.fa-lock   { margin-left: 0.3em; color: green; background: white; }
+.viewlink svg.fa-unlock { margin-left: 0.3em; color: red;   background: white; }
+.dark .viewlink svg.fa-lock   { margin-left: 0.3em; color: green; background: #2a2a2a; }
+.dark .viewlink svg.fa-unlock { margin-left: 0.3em; color: red;   background: #2a2a2a; }
```
peternewman, line 347: I'm very late to the party @UKChris-osm , but I think these look a lot better without the white/dark background, so they're naturally transparent and match the underlying page colour; maybe that has changed since this went in?

PR #5415, opened 2021-10-03
Add amenity/school

arch0345:
This could canonicalize the operators of schools, mainly public, around the world. Exclude list needs some refining before merging.


arch0345:
A geojson file for Los Angeles Unified School District should be created to disambiguate with Los Alamitos Unified School District due to their identical abbreviated name and how close they are to one another.

Adamant36:
> This could canonicalize the operators of schools, mainly public, around the world. Exclude list needs some refining before merging.

As far as I know we aren't adding schools operators at this point unless they operate multiple campuses. I thought there was a discussion at one point to add the operator to individual objects on a campus like buildings, but I don't think it went anywhere. So my guess is that this won't be workable under the current way the NSI does things. Unless there are specific entries that have satellite campuses, then you could add those individually, but just adding every possible school operator regardless of if they do or not would not be good. Nor would it be useful IMO.

bhousel:
I'll try to review it this week.. School Districts are definitely like other publicly operated things, like the fire, police, ambulance stuff that we now include in the index.  The collection script sure has included a lot of them (I took a quick look at `operators_keep.json` diff!)

bhousel:
Hey @arch0345 this looks pretty good to me - go ahead and merge when you are happy with it 👍 
After you merge it, I can help simplify the exclude regexes a bit.

It's a lot of school districts, and will be a big project for people to dive into!

PR #5416, opened 2021-10-03
Added First Choice Liquor to brands/shop/alcohol in au region

orangegrove1955:
First Choice Liquor is an Australian liquor retailer. As such, I have added it to the list of shop/alcohol brands, including its Wikipedia and Wikidata entries, and run `npm run build` to generate an id for it


arch0345:
Thanks @orangegrove1955!

PR #5417, opened 2021-10-03
Added Duncans to brand/shop/alcohol in au

orangegrove1955:
Duncans is a liquor retailer in Australia, as such I have added it to brands/shop/alcohol then run `npm run build` to generate an id for it. There does not appear to be a wikipedia page or a wikidata entry for it, but they do have a website as proof of its existence:
https://www.duncans.com.au/

### DIFF
--- a/data/brands/shop/alcohol.json
+++ b/data/brands/shop/alcohol.json
@@ -141,6 +141,16 @@
       }
     },
     {
+      "displayName": "Duncan's",
+      "id": "duncans-cb3c62",
+      "locationSet": {"include": ["au"]},
+      "tags": {
+        "brand": "Duncan's",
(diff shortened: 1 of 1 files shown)


arch0345:
Looks good @orangegrove1955! Note that this won't be generated as an iD preset as it doesn't have a linked wikidata page. This can easily be created; make sure you add some statements (instance of, country, official website, etc.) so that the page won't get deleted. NSI gets logos from either wikimedia or the linked facebook/twitter profile on the wikidata page, so if you want the logo to be shown, you should add at least one of these to the wikidata page.

orangegrove1955:
Thanks for that @arch0345 that's very useful to know. I'll try to get a wikidata page created for this and update further if possible!

PR #5419, opened 2021-10-03
Update bus.json

rkdclgh:
Modified the bus network of Gyeonggi Province, South Korea.
G-Bus, 광주시 공영버스 and 하남시 마을버스/시내버스/직행좌석버스 is merged into 경기도 시내버스.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -16903,7 +16903,8 @@
       "matchNames": [
         "하남시 마을버스",
         "하남시 일반시내버스",
-        "하남시 직행좌석버스"
+        "하남시 직행좌석버스",
+        "광주시 공영버스"
       ],
       "tags": {
         "network": "경기도 시내버스",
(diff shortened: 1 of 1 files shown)


data/transit/route/bus.json
```diff
@@ -16909,18 +16903,11 @@
       "tags": {
         "network": "경기도 시내버스",
         "network:ko": "경기도 시내버스",
+        "network:en": "G-BUS"
```
tas50, line 16906:
```suggestion
        "network:en": "G-BUS",
```

data/transit/route/bus.json
```diff
@@ -16909,18 +16903,11 @@
...
         "route": "bus"
```
tas50, line 16907:
```suggestion
        "route": "bus",
```

arch0345:
Thanks @rkdclgh and welcome! In the future, you could add matchNames of brands you merged as shown in my commits so that they won't get added again in the future.

PR #5421, opened 2021-10-03
Add missing Indonesians university

irhamdz:
This PR will add missing Indonesian university:
- Bandung Institute of Technology,
- Sunan Kalijaga Islamic University,
- University of Indonesia


arch0345:
At this time, we're only collecting for-profit universities that operate multiple campuses, hence the brand tag instead of operator. After some research, it seems that these are public universities that only consist of one campus.

PR #5422, opened 2021-10-03
Add new mosque amenity and add missing indonesian mosque

irhamdz:
This PR will add new amenity/mosque and add missing Indonesian mosque:
- Istiqlal Mosque
- Grand Mosque of Bandung

### DIFF
--- a/data/brands/amenity/mosque.json
+++ b/data/brands/amenity/mosque.json
@@ -1,0 +1,55 @@
+{
+    "properties": {
+        "path": "brands/amenity/mosque",
+        "skipCollection": true,
+        "exclude": {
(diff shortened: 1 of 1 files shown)


irhamdz:
may I know, any step to adding new amenity like this? Thank you!

Adamant36:
> may I know, any step to adding new amenity like this? Thank you!

Sorry, but the index isn't really meant for adding places that are single location. As they can easily be added by an individual mapper on their own. Plus, I don't think we support religions yet either. So I'm going to close this. Your free to add whatever tags you want to the actual map POIs on your own though. 

Thanks

PR #5424, opened 2021-10-03
add Wikidata tag to existing NSI entries for Dutch public transport

pdejan:
In reference to issue #5414 (first part)
All Wikidata tags added to existing NSI entries
npm build didn't fail with any errors


pdejan:
All data in reference to issue #5414 has now been added, ran npm build, didn't fail.

arch0345:
Thanks so much for taking the time to resolve #5414 @pdejan!

PR #5434, opened 2021-10-05
Update supermarket.json

alessandrocarmeli:
zazzaron added

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -21,6 +21,16 @@
     }
   },
   "items": [
+     {
+      "displayName": "Supermercati Zazzeron",
+      "locationSet": {"include": ["it"]},
+      "tags": {
+        "brand": "Supermercati Zazzeron",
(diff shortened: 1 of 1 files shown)


bhousel:
This seems okay but https://www.wikidata.org/wiki/Q108805425 really needs a bunch more details added to it, otherwise it runs the risk of being deleted by Wikidata admins.  I'm going to add some more details to the page now.. 

Please check this guide for editing Wikidata if you decide to add more of these in the future.  It's a very useful task and worth doing! - but we want to make sure not to upset the Wikidata project.  
https://github.com/osmlab/name-suggestion-index/wiki/Editing-Wikidata

bhousel:
From a web search, I can't really tell how many of these supermarkets there are, but it doesn't look like many.  We usually only add brands for retail chains that operate in a lot of locations, like 50 locations or so (we are flexible on this, but it's a rough guideline).  

I'll leave it in the index just because I really don't want to do any more research on this right now, but going forward please don't add a brand if it's just like 5 stores in one town.

PR #5436, opened 2021-10-05
Update college

pratikpatil9:
Added new college Dbatu.

### DIFF
--- a/data/brands/amenity/college.json
+++ b/data/brands/amenity/college.json
@@ -2,13 +2,13 @@
   "properties": {
     "path": "brands/amenity/college",
     "skipCollection": true,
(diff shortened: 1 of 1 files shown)


bhousel:
Sorry @pratikpatil9 , but we don't currently add individual colleges or universities to this index - we only add the kind of for-profit "universities” that operate like a branded business in multiple locations, such as DeVry or University of Phoenix. I know this is an unusual concept outside of the USA - the “universities” in the name-suggestion-index are really more like training centers than true universities.

This index is really for tracking the types of things where there are a lot of them, not just single things. We are considering adding some university operated assets in #4779

PR #5442, opened 2021-10-06
Added Chadstone Shopping Centre to Australian malls

orangegrove1955:


bhousel:
If it's just a single shopping center, we wouldn't add it to this index.. 
This project is more for tracking commonly mapped things, so all the items in the `mall.json` are for groups that operate a chain of malls.

PR #5445, opened 2021-10-06
merge two banks

waso99:
merge (SAMBA bank) with (National Commercial Bank) to become Saudi National Bank.


bhousel:
This looks ok to me - does it make sense to add a `matchNames` property that contains the old names, so that software like iD can offer tag upgrades?

bhousel:
Thanks @waso99 👍

PR #5447, opened 2021-10-06
Update JOSM presets link in README

Undearius:
Old link redirects to the Map Projection page
Corrected to Tagging Presets page


bhousel:
Thanks @Undearius  and welcome! 👋

Issue #5440, opened 2021-10-06
Add VASA Fitness


tas50:
US gym chain https://vasafitness.com

Issue #5432, opened 2021-10-05
Add Outdoor Supply Hardware


tas50:
Orchard Supply Hardware was shut down by Lowe's and was bought up to become Outdoor Supply Hardware. 99.9% same store down to the logo. There's a bunch in California.

PR #5448, opened 2021-10-06
Add brands Vasa Fitness and Outdoor Supply Hardware

viveknshah:
Adding Vasa Fitness to leisure/fitness_centre.json. Solves #5440
Adding Outdoor Supply Hardware to shop/hardware.json. Solves #5432


viveknshah:
Taken @UKChris-osm feedback and removed brand:website

PR #5457, opened 2021-10-07
Add missing "cuisine" tags to some brands

Sukkoria:
Hello! 
I added some missing tags, based on warnings raised while running "npm run build" and fetching new commits.


arch0345:
Thanks @Sukkoria!

Issue #5455, opened 2021-10-07
Add Graybar


tas50:
These are tagged all over the place. It's a building supply center specifically for electrical supplies.

https://en.wikipedia.org/wiki/Graybar

PR #5464, opened 2021-10-09
Add Graybar

Brainshaker95:
added graybar to electronics shops. #5455


UKChris-osm:
Would including the location for Canada be suitable with this PR, or should Graybar Canada be listed on its own?

PR #5465, opened 2021-10-09
Add missing Bulgarian brands

Dimitar5555:


arch0345:
> I'm not sure how I can fix the error. Any help would be appreciated.

The PR check says that there's a duplicate displayName for `Медея`. I recommend checking [nsi.guide](https://nsi.guide) to see if this is already in the index. There are also some minor syntax errors, I'll point out these in my review

arch0345:
Thanks @Dimitar5555 for adding these brands and fixing errors

arch0345:
> Everything should be fixed now. Thanks for the help.

No problem!

PR #5466, opened 2021-10-09
add Synevo

matkoniecz:


matkoniecz:
> Do you know if a wikidata page for all of these countries could be added or would it be easier to add these separately?

No idea. I used this specific one because some PR person was kind enough to create soon-to-be-deleted Wikipedia article on Polish Wikipedia and Wikidata entry that should survive.

If anyone feels that ignoring Q108474045 and creating more generic wikidata entry for entire Synevo brand would be better - feel free to create it.

arch0345:
Ok, I'll merge this for now and I'll create an issue for a more generic page to be created

PR #5469, opened 2021-10-10
Added Jack Stack Barbecue to restaurant.json

watkins656:


watkins656:
Oh shoot, I may have missed that in the contributing guideline.  Is there a
section that outlines the minimum requirements?

UKChris-osm:
@watkins656 to be honest it's a slightly grey area, as there isn't a specific minimum number. The automatic planet scan will look for anything that has 50 or more entries when it comes to restaurant "_brands_", but only 10 when looking for "_operators_" such as fire stations, so I might leave this PR open to see what others say for the time being.

Adamant36:
I was thinking it might be worth merging as a way to disambiguate the brand from other places with similar names, but all I could find is a kitchen supply store and a driving store. At least in America. It's possible there are other brands called "Jack Stack" or something along those lines in other countries. but I don't know🤷‍♂️

Adamant36:
> I can try to update the contributing guidelines to make it clear that our index is really for things that are "notable" in the Wikidata sense, and "tagged consistently" in the OSM sense.

Kind of an off-topic side thought, but from what I've seen Wikidata is pretty inconsistent with what they consider notable or not when it comes to stores. A lot of the times it's based on if someone feels like doing a deletion request for an entry or not, and it's rarely (if ever) based on the number of locations a store has. So I think the NSI's threshold (whatever it might be) is the only reliable standard we can go by. With brands in particular there's lots of benefit, and not much downside, to going below 50 instances. 10 instances is a good good bottom end for bus routes though. I doubt there's many routes out there with only like 4 stops. Shuttles maybe, but that's all I can think of and they are generally rare.

PR #5471, opened 2021-10-10
add burger to braums' list of cuisines

Tavlor:
Braums serves burgers as well as ice cream (gotta use the whole cow, y'know)

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -462,7 +462,7 @@
         "brand": "Braum's",
         "brand:wikidata": "Q4958263",
         "brand:wikipedia": "en:Braum's",
-        "cuisine": "ice_cream",
+        "cuisine": "ice_cream;burger",
         "name": "Braum's",
         "shop": "dairy",
         "takeaway": "yes"


bhousel:
It sounds a bit like the United Dairy Farmers item:
https://github.com/osmlab/name-suggestion-index/blob/2b56162b4c1df2349b43c8f6e188a830a2ea1c73/data/brands/shop/convenience.json#L3087-L3103

So I think it's ok to combine `amenity=ice_cream` and `shop=convenience` as long as you:
- have the item in one category (shop/convenience) and
- matchTags for the other category (amenity/ice_cream)

1ec5:
I’m familiar with both Braum’s and UDF (#2234). I think the dining and market areas of a Braum’s are more distinct than at a typical UDF. UDF is like a convenience store with tables added for convenience (not unlike the tables at a Domino’s). But there’s variation in both chains and we’re really splitting hairs here.

If we provide separate presets for the dining and market areas, then there’s a high chance that only one or the other will be mapped at a given store location. We’ll also see inconsistency with some mappers tagging the whole standalone building as a Braum’s market, others tagging it as a Braum’s parlor, and others as two separate points within the building. There’s precedent for such distinctions (in drugstores/pharmacies), but we should be judicious about them. Otherwise, we might end up with rendered maps that, to an end user, look like clutter due to a lack of POI conflation.

bhousel:
Anyway this has gone off topic a bit but I'm going to merge the PR.  
Thanks @Tavlor for the PR & welcome 👋

PR #5481, opened 2021-10-11
Update fast_food.json

andrewpmk:
Remove "Harveh's" alternate name from Harvey's. Looks like this was a name used in a short-lived marketing campaign https://twitter.com/harveyscanada/status/1397284046359642121 - do we really want the iD validator automatically suggesting this as a correction? When I use the iD editor I always ignore this correction.


arch0345:
Thanks @andrewpmk and welcome!

PR #5485, opened 2021-10-12
Add E-Mart

prizef:


bhousel:
> The store is similar to Target or Walmart, but groceries are a larger part of their business.

Sounds good - the term of art for these kinds of stores (Target or Walmart but with groceries) is a ["hypermarket"](https://en.wikipedia.org/wiki/Hypermarket).  Current OSM convention is to use the [`shop=supermarket`](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dsupermarket) tag for hypermarkets, so this PR is good to merge 👍

PR #5502, opened 2021-10-13
find and replace long name states w/ ISO names

jdcarls2:
Well, I'd meant the "deleting files" to be its own commit, but w/e.

Now that the ISO-named states are in, I did a quick find-and-replace for all the existing long-name state geojson files. That is, `california.geojson` -> `US-CA.geojson`

There are loads of entries that can still be refined with the new shapes, but they can't all be done by state name alone. "Texas Roadhouse", for example, has locations all over. But for the changes made here, these were (mostly) already state-specific, and safe targets for a find and replace.

There were a few things that I changed from just ["us"] to being a state that also seemed safer, like utility companies.


bhousel:
Looks good to me, thanks again @jdcarls2 👍

Issue #5501, opened 2021-10-13
Cellaris should be Cellairis


Adamant36:
I fixed it on Wikidata's end, but the name in the NSI entry still needs to be changed. Along with the identifier after it's updated.

PR #5506, opened 2021-10-14
updated cellaris to cellairis #5501

lexoskeleton:


arch0345:
Thanks @lexoskeleton! Just letting you know that you only need to update the file in the `data` directory, the others can be updated by maintainers with a script before the next release

PR #5521, opened 2021-10-17
added Dutch Supermarkets

wvanderp:
added Dutch supermarkets based on the most used names from tag info and local knowledge


arch0345:
Thanks @wvanderp and welcome!

PR #5526, opened 2021-10-18
Add support for amenity/library in operator tree

arch0345:
This is similar to other categories I've added to the index in the past: amenity/school and leisure/park. 

There are many public operators of libraries that operate multiple locations (at least in the US), so adding this category in the index could be beneficial. So far the only entry I've completed is the New York Public Library, but it looks like a lot more have been collected from the latest planet file.

### DIFF
--- a/data/operators/amenity/library.json
+++ b/data/operators/amenity/library.json
@@ -196,6 +196,23 @@
       }
     },
     {
+      "displayName": "Cincinnati and Hamilton County Public Library",
+      "id": "cincinnatiandhamiltoncountypubliclibrary-f8d891",
+      "locationSet": {
+        "include": ["us-oh.geojson"]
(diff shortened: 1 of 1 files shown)


1ec5:
> There are many public operators of libraries that operate multiple locations (at least in the US), so adding this category in the index could be beneficial.

Agreed, Wikidata even [imported](https://www.wikidata.org/wiki/Special:WhatLinksHere/Q69494266) [a database](https://www.wikidata.org/wiki/Q69494266) of over 17,000 U.S. library systems and branch locations, so there’s a lot of room for this tree to expand (and for individual branches to be tagged). Alas, it’s probably too late for @bhousel to spin up another massive Hacktoberfest issue series around these library systems.

PR #5528, opened 2021-10-19
Add SITP (Bogotá)

desiderantes:
Since no one calls this bus network by the full name, and the abbreviation is used officially everywhere, I decided to leave it like that


arch0345:
Looks good @desiderantes!

PR #5531, opened 2021-10-19
Fix country code on UK flag

1ec5:
The `country` key is set to an ISO&nbsp;3166-1&nbsp;alpha-2 code. The canonical code for the United Kingdom is `GB`, even though `UK` is reserved for the same purpose. `GB` also predominates in OSM usage. The build script insists that the name of the preset include the country code, not an alias, so the preset is now named “GB – United Kingdom”, not “UK – United Kingdom”. (Otherwise, the build script would’ve turned it into “GB – UK – United Kingdom”.)

### DIFF
--- a/data/flags/man_made/flagpole.json
+++ b/data/flags/man_made/flagpole.json
@@ -1772,6 +1772,21 @@
       }
     },
     {
+      "displayName": "GB - United Kingdom",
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks, I agree - I don't know why I did it that way originally..

PR #5536, opened 2021-10-19
Stadtmobil is a brand not an operator

sb12:
The Stadtmobil brand is used by several independant operators per city using the same brand and booking systems, but e.g. different pricing. 
Stadtmobil Südbaden is a different brand.

### DIFF
--- a/data/operators/amenity/car_sharing.json
+++ b/data/operators/amenity/car_sharing.json
@@ -538,24 +538,79 @@
       }
     },
     {
-      "displayName": "stadtmobil",
-      "id": "stadtmobil-088af7",
+      "displayName": "Stadtmobil Karlsruhe",
+      "id": "stadtmobilKarlsruhe-088af7",
       "locationSet": {"include": ["de"]},
-      "matchNames": [
-        "stadtmobil carsharing gmbh & co. kg",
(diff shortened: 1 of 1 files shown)


sb12:
But is it possible to have different operators for the same brand in the brand branch? 
Looking through the list in car_sharing.json, there are actually several cases where brand and operators are mixed up, e.g. Cambio is using a similar operator scheme as Stadtmobil, Flinkster is a brand by Deutsche Bahn Connect GmbH.
I am happy to fix these, as soon as I know how to do that without breaking the index.

bhousel:
Hey @sb12 , since doing #5625 I'm more confident that it's ok for a category like `amenity/car_sharing` to have some items under 'brands' and some items under 'operators'.

in 928764e2e I created `brands/amenity/car_sharing.json` and moved "stadtmobil" into it.
I think this addresses your concern that the brand has different operators depending on what region it is.

If you want to go further and create separate "stadtmobil" items for each city/operator (like this PR), you can do this but you should give them each a regional locationSet - they shouldn't all just be `{"include": ["de"]}`.

Let me know if you have questions!  It would be great to have your help in cleaning up these car sharing brands 👍

PR #5538, opened 2021-10-19
Add wikidata items for bike sharing operators

Sukkoria:
Hi, here are some more wikidata items, a good part being the addition of BicinCittà, a major Italian operator.


arch0345:
Thanks @Sukkoria!

PR #5542, opened 2021-10-21
More state implementations

jdcarls2:
Continuing the progress of refining locationSets where possible.

This PR focuses mostly on utility companies and similar entities with clear, state-based boundaries of operation.

There were also a couple of locationSets that had an ISO state name (`us-hi`) without the `.geojson`, so I added that when I caught it.

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -137,10 +137,8 @@
     },
     {
       "displayName": "ABC (Hawaii)",
-      "id": "abc-7a9151",
-      "locationSet": {
-        "include": ["us-hi.geojson"]
-      },
+      "id": "abc-540f13",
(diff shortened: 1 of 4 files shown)


arch0345:
Based on [location-conflation.com](https://location-conflation.com/?locationSet=%7B%20include%3A%20%5B%27us-hi%27%5D%20%7D%0A), it looks like Hawaii, Alaska, and other US territories are in the [ideditor/country-coder](https://github.com/ideditor/country-coder) repo to disambiguate with the Contiguous United States. IIRC it takes longer for iD and other editors that rely on NSI to reference geojsons from this project than using country codes.

jdcarls2:
Oh! That makes perfect sense. I'll amend those particular edits.

PR #5547, opened 2021-10-22
Even more states

jdcarls2:
Working through more of the US states, swapping in state-specific geojsons in the locationSet when possible.
Also fixed a few situations where a Missouri brand/operator had `us-mi.geojson` as the location, which would be Michigan.


bhousel:
Thanks @jdcarls2 👍

PR #5555, opened 2021-10-24
Add Sparkasse Worms-Alezy-Ried

coolultra1:
I think this one should be in here. A German bank.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -12814,6 +12814,19 @@
         "name:nan-TL": "Ko-hiông Gîn-hâng",
         "name:zh": "高雄銀行"
       }
+    },
+    {
+      "displayName": "Sparkasse Worms-Alzey-Ried",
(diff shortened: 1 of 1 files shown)


kjonosm:
@coolultra1 , @bhousel, @arch0345 
For ATMs I keep seeing wrong suggestions, eg. [here](https://www.openstreetmap.org/node/1552072162)
Bildschirmfoto 2021-10-30 um 11 43 51
If this entry is kept, its location set should be narrowed down to the operating area of Sparkasse Worms-Alzey-Ried.

arch0345:
I narrowed the locationSet to [this area](https://location-conflation.com/?locationSet=%7B%20include%3A%20%5B%5B8.2%2C%2049.7%2C%2025%5D%5D%20%7D%0A&referrer=nsi), which should cover their operating area.

coolultra1:
@arch0345 That does cover their operating area, thank you for doing so, I wasn't really sure how to do circle radius thingy correctly.

PR #5576, opened 2021-10-27
Display NSI Tags alongside OSM tags on nsi.guide

UKChris-osm:
Currently the [nsi.guide](https://nsi.guide) web site displays the OSM tags that each entry will add - this PR also displays the NSI only tags, such as **matchNames**, **matchTags**, and **note**. [example here](https://nsi.haignet.co.uk)

nsi-tags-example


UKChris-osm:
As it turns out I forgot to add the text "_NSI Tags_" to this branch anyway, so no need to worry about the label @bhousel :)

PR #5594, opened 2021-10-29
Made Edits To Filter Section For Mobile Experience

Justin-Bento:
Here are some changes I made to the Filter Section; 

- [x] Changed filter from an inline-block to flex, for making responsive code easier.
- [x] Removed max-width to separate the introduction and filter section.
- [x] Updated the spacing for the text and components.
- [x] Checked different viewports to verify the spacing.


bhousel:
Thanks @Justin-Bento 👍

PR #5595, opened 2021-10-29
Added dutch providential flags and added wiki tags to french flags

wvanderp:
I added dutch Province flags. Provinces are equivalent to states in the USA or Bundesländer  (or states) in Germany.

also added flag wiki tags to France Regions (also similar to Provinces)


bhousel:
Thanks @wvanderp 👍

PR #5596, opened 2021-10-29
Add A&O hotels and hostels

Robot8A:


UKChris-osm:
Hi @Robot8A just to let you know I fixed a build error with the new hostel.json file you're proposing, as a } was in the wrong place.

PR #5601, opened 2021-10-31
Add LifeLabs and Hamilton Bike Share; add operator and matchNames to Hamilton Street Railway

kyy0602:


arch0345:
Thanks @kyy0602 and welcome! Note that Hamilton Bike Share won't get generated as an iD preset since it doesn't have a wikidata page linked to it.

kyy0602:
> 
> 
> Thanks @kyy0602 and welcome! Note that Hamilton Bike Share won't get generated as an iD preset since it doesn't have a wikidata page linked to it.

Thanks @arch0345 ! I would probably add a wikidata page and link to it in a future PR since there isn't one yet.

arch0345:
>I would probably add a wikidata page and link to it in a future PR since there isn't one yet.
Great! Just make sure you add some statements (instance of, country, official website, etc) so it doesn't get deleted. Since we collect logos from Facebook/Twitter profiles on their wikidata page, I also recommend adding at least one of these if they have one.

kyy0602:
> 
> 
> > I would probably add a wikidata page and link to it in a future PR since there isn't one yet.
> 
> Great! Just make sure you add some statements (instance of, country, official website, etc) so it doesn't get deleted. Since we collect logos from Facebook/Twitter profiles on their wikidata page, I also recommend adding at least one of these if they have one.

Thanks for the tips!

Issue #5512, opened 2021-10-17
Add SEVA Beauty hairdressers


tas50:
Hairdresser chain with location within Walmarts

https://sevabeauty.com/

pdejan:
Should this be categorized under hairdresser or beauty? Their website indicates services oriented towards beauty treatments.

PR #5603, opened 2021-10-31
Add Seva Beauty brand + npm run build

pdejan:
In reference to #5512 issue


bhousel:
Thanks @pdejan, looks good to me 👍

Issue #5589, opened 2021-10-29
Add Karma Automotive


arnowrld:
Karma Automotive is a luxury vehicle manufacturer & brand with a dealer network found in North America, Europe, South America, and the Middle East.

Wikidata: https://www.wikidata.org/wiki/Q22671741

PR #5604, opened 2021-10-31
Adding Karma & Lucid Motors (Showroom/Dealer)

arnowrld:
#5589 with additions of Lucid Motors.


bhousel:
Thanks @arnowrld , looks good.. 
I think usually the display name should try to match the name, so I'm going to change it from "Lucid Studio" to "Lucid Motors" in a followup commit.

PR #5613, opened 2021-10-31
Post offices and related items need to be in the brands tree, not the operators tree

### LINKED ISSUE #5446: Incorrect operator tags for UK Post Offices
I've just had an incorrect "tag upgrade" suggestion from iD that I think is due to an NSI entry in data/operators/amenity/post_office.json . It involved suggesting the addition of the following tags to an amenity=post_office object:

operator=Post Office
operator:wikidata=Q1783168
operator:wikipedia=en:Post Office Ltd

In the UK, only a small number of larger branches (known as "Crown Post Offices") are directly operated Post Office Ltd. The rest (the majority) are run on a franchise model either by other businesses or individual sub-postmasters. So while it would be correct to offer to add Post Office Ltd brand and brand:* tags to individual post offices (as long as the amenity=post_office isn't on another shop=* node that might need a different brand), the operator tags above will normally be incorrect.

### LINKED ISSUE #5561: Posten Norge
For "Posten (Norge)", category amenity=post_office:
* operator is not Posten, but the local, private operator
* operator:wikidata=Posten is therefore perhaps not strictly correct - could use brand:wikidata=Posten instead (but please note that the brand is not part of the post office name)

bhousel:
This PR moves the post offices from the `operators` tree back to the `brands` tree.

This fixes the situation for Poste Norge and Post Office (and probably others) where they are really brands that have independently operated locations.  This also more closely matches the current guidance on the OSM wiki for tagging post offices.

closes #5564 
closes #5561 
closes #5446


data/brands/amenity/post_box.json
```diff
@@ -1,6 +1,6 @@
 {
   "properties": {
-    "path": "operators/amenity/post_box",
+    "path": "brands/amenity/post_box",
```
1ec5, line 3: To be clear, https://github.com/osmlab/name-suggestion-index/issues/5561#issuecomment-954028769 described [an edit to the `amenity=post_office` documentation](https://wiki.openstreetmap.org/wiki/Special:Diff/2213092) that should’ve had little bearing on `amenity=post_box`. Conversely, these features were originally moved from the `brands` tree to the `operators` tree in response to feedback in #4925, but that feedback was about `amenity=post_box`, not `amenity=post_office`.

I think the back and forth is because post offices and postboxes are being treated identically. But it seems like the expectation – most of the time – is that post offices need the distinction between brands and operators more than postboxes do. Contract post offices are a widespread phenomenon in many countries, but contract postboxes seem would be much more of an edge case.

(I did argue in gravitystorm/openstreetmap-carto#1023 that postboxes should be labeled based on their `operator`s, but that was before I learned about `brand` and before it had become common on postal facilities in the U.S.)

bhousel: oh ok - If I'm reading your comment correctly, it sounds like `amenity/post_box` should remain in `operators` tree?  That seems ok to me.

bhousel:
I ended up reverting this PR, there is some investigation on #5613 and I'm now convinced that we can have a category like `amenity/post_office` that has some items in the operators tree and some in the brands tree.

PR #5614, opened 2021-10-31
Add MIKA Hundetoiletten

coolultra1:
These things are bought by city councils, park/zoo owners etc. They all have a clear branding. MIKA themselves state, currently over 1000 citys and councils are using these.


arch0345:
Looks good 👍

PR #5615, opened 2021-10-31
Update sports.json

alessandrocarmeli:
sportler added

### DIFF
--- a/data/brands/shop/sports.json
+++ b/data/brands/shop/sports.json
@@ -6,6 +6,18 @@
     }
   },
   "items": [
+     {
+      "displayName": "Sportler",
+      "locationSet": {"include": ["it", "at"]},
+      "tags": {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Hi @alessandrocarmeli, and thanks for the PR - we tend not to include a web site within the Name Suggestion Index however, as this data is available via the Wikidata. I don't think "_strapline_" is something that is currently considered either, so I would suggest removing both of these values.

bhousel:
I agree with @UKChris-osm , I merged and removed the `strapline` and `website` tags in 1cb5f6193 (we don't add these tags in the index).

kjonosm:
@alessandrocarmeli, @bhousel  Could you double check the name value (Sportoler)? It looks very much like a typo...

UKChris-osm:
Good spot @kjonosm

Issue #3184, opened 2019-10-28
Research Conoco


wireguy:
Conoco (amenity/fuel) is currently appearing in the dissolved.json file, getting kicked out because Conoco was marked dissolved in wikidata: https://www.wikidata.org/wiki/Q1126518

Conoco has become ConocoPhillips, which has a wikidata entry of:
https://www.wikidata.org/wiki/Q214763

should Q1126518 be changed to Q214763 in nsi?
should exisiting osm pois get updated?

bhousel:
Yeah this is #3154 - I'm not really sure what the best solution is.
I like @1ec5's suggestion that we maintain `dissolved.json` manually rather than using a script to look for dissolution dates in Wikidata.

PR #5620, opened 2021-11-01
Updating Conoco (Fuel)

arnowrld:


bhousel:
Oh you found an interesting edge case. Conoco relates to issues #3184 and #3154 .. 

I think when we researched this last time, [Conoco Q1126518](https://www.wikidata.org/wiki/Q1126518) was considered on Wikidata/Wikipedia to be "the business that got dissolved in 2002". 

When I look at the https://en.wikipedia.org/wiki/Conoco page today, it now acknowledges that it is an active brand owned by Phillips 66.  This is good!

When I look at https://www.wikidata.org/wiki/Q1126518 wikidata today, it still does claim a dissolved date of 2002.  I think for us to use this QID, we'd need to remove this dissolved date, or maybe add a qualifier to it somehow.  Maybe @1ec5 has some ideas on what to do?

I'm marking this as a Draft PR until we can research it a bit more.

data/brands/amenity/fuel.json
```diff
@@ -1096,8 +1096,8 @@
       "tags": {
         "amenity": "fuel",
         "brand": "Conoco",
-        "brand:wikidata": "Q214763",
-        "brand:wikipedia": "en:ConocoPhillips",
+        "brand:wikidata": "Q1126518",
+        "brand:wikipedia": "en:Conoco",
```
1ec5, line 1100: I split out [Q109341187](https://www.wikidata.org/wiki/Q109341187) for the Conoco brand as it has existed under multiple owners over the years and refocused [Q1126518](https://www.wikidata.org/wiki/Q1126518) to be about Conoco Inc. specifically. [The English Wikipedia article](https://en.wikipedia.org/wiki/Conoco) is an odd mixture of the two Conocos, so I linked it to the new item instead of the old one. The other Wikipedias only have articles that focus solely on Conoco Inc., so I left those sitelinks alone.

```suggestion
        "brand:wikidata": "Q109341187",
        "brand:wikipedia": "en:Conoco",
```

1ec5: Phillips 66 should probably get the same treatment at some point.

bhousel:
Thanks @arnowrld and @1ec5 👍

PR #5623, opened 2021-11-01
Adding Tesla Destination Chargers

arnowrld:
I felt like this aspect of Tesla's network was missing.

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -3476,6 +3476,25 @@
       }
     },
     {
+      "displayName": "Tesla Destination Charger",
+      "id": "tesladestinationcharger-c22d3a",
+      "locationSet": {"include": ["001"]},
+      "matchNames": [
+        "tesla motors",
+        "tesla motors inc"
+      ],
+      "tags": {
+        "amenity": "charging_station",
+        "brand": "Tesla Destination Charger",
+        "brand:wikidata": "Q478214",
+        "name": "Tesla Destination Charger",
+        "operator": "Tesla, Inc."
+        "operator:wikidata": "Q478214",
+        "operator:wikipedia": "en:Tesla, Inc.",
+        "short_name": "Tesla"
+      }
+    },
+    {
       "displayName": "Tesla Supercharger",
       "id": "teslasupercharger-c22d3a",
       "locationSet": {"include": ["001"]},
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -3488,7 +3488,7 @@
         "brand": "Tesla Destination Charger",
         "brand:wikidata": "Q478214",
         "name": "Tesla Destination Charger",
-        "operator": "Tesla, Inc."
+        "operator": "Tesla, Inc.",
         "operator:wikidata": "Q478214",
         "operator:wikipedia": "en:Tesla, Inc.",
         "short_name": "Tesla"


arch0345:
I think these weren't added because of how charging stations are in the operator tag. While they're branded as Tesla chargers, according to these [Quora answers](https://www.quora.com/What-is-the-difference-between-a-Tesla-Super-Charger-and-a-destination-charger-Are-the-costs-different) and [this blog](https://www.solarreviews.com/blog/tesla-charging-stations-everything-you-need-to-know), it seems like they're operated by "Tesla Partners" and not Tesla themselves. I couldn't find anything on the [official website](https://www.tesla.com/destination-charging) that directly says this.

arnowrld:
That... makes sense, though I still think that the concept of Tesla Destination Chargers has to be added in some way or form.

arch0345:
Maybe a file with just this could be added to the brands tree just like what I did with `amenity/library`

bhousel:
> Maybe a file with just this could be added to the brands tree just like what I did with `amenity/library`

oh I didn't realize you did that - is it working?
Now I see libraries in both 'brands' and 'operators' and this is likely to cause issues.

arch0345:
> > Maybe a file with just this could be added to the brands tree just like what I did with `amenity/library`
> 
> oh I didn't realize you did that - is it working?
> Now I see libraries in both 'brands' and 'operators' and this is likely to cause issues.
> 

It seems to be working - I made sure to add `skipCollection` to the one in the brands tree and the ones in the brands tree won't also get added to the operators one as a city, county, or NGO operates the individual libraries

bhousel:
> It seems to be working - I made sure to add `skipCollection` to the one in the brands tree and the ones in the brands tree won't also get added to the operators one as a city, county, or NGO operates the individual libraries

I'm just thinking out loud about the things that might be an issue:
- The NSI matcher code probably will have subtle issues, because its match index is built on `k/v`.  This might mean the items and the exclude lists from one file may overwrite those things from other file.  I expect that this code will just never match stuff in whatever file it parsed first.
- In iD there is other code to try to "guess" what tree a `k/v` pair belongs to.  (This code might guess wrong, but it might be ok anyway because brands and operators are similar - I dont know).  The reason for this "tree guessing" is because iD is just looking at raw OSM tags and it needs to decide which tags are considered "namelike".
- For example, the iD validator sees a `shop=yes`+`name:en=USPS - Morristown`+`height=3.5`+`roof:colour=orange`, it should be able to run the `name:en` tag through the matcher and guess that that is probably a US Post Office, even without actually seeing a defining tag like `amenity=post_office`, and also not matching the "Orange" mobile phone brand because of that `roof:colour` tag..

This probably just needs to be its own issue now, because the 'brand'-'operator' split is causing a whole lot of grief for me lately, and I _just_ moved the post offices in #5613 (mostly for Norway, UK, and things like UPS store) and I'm still not sure I did the right thing.

data/operators/amenity/charging_station.json
```diff
@@ -3475,6 +3475,25 @@
...
+      "matchNames": [
+        "tesla motors",
+        "tesla motors inc"
+      ],
+      "tags": {
+        "amenity": "charging_station",
+        "brand": "Tesla Destination Charger",
+        "brand:wikidata": "Q478214",
+        "name": "Tesla Destination Charger",
+        "operator": "Tesla, Inc."
```
UKChris-osm, line 3491: Just an FYI - a missing comma here is causing a build failure.

bhousel:
> Maybe a file with just this could be added to the brands tree just like what I did with `amenity/library`

☝️ I ended up doing this in dd4c7369b
now that I'm feeling more ok about having 2 files post-#5625

The `brands/amenity/charging_station.json` file just has the Tesla chargers for now.. This is ok because the supercharger network has a unique `brand:wikidata` value and the destination chargers can all have different operators.

We can move other charging stations from `operator`->`brand` if they work in a similar way (have consistent brand but different operators).

PR #5629, opened 2021-11-01
Update bus.json

alessandrocarmeli:
trieste trasporti edited


UKChris-osm:
Thanks @alessandrocarmeli - but I don't think "_old_operator_" is needed in this PR.

data/transit/route/bus.json
```diff
@@ -14275,6 +14275,8 @@
         "network": "Trieste Trasporti",
         "network:wikidata": "Q3539115",
         "network:wikipedia": "it:Trieste Trasporti",
+        "old_operator": "Trieste Trasporti"
```
UKChris-osm, line 14278: I would suggest removing this line.

UKChris-osm: Sorry, ignore the matchNames comment as I see old operator is the same name as display name.

PR #5630, opened 2021-11-01
Update bus.json

alessandrocarmeli:
tpl fvg added

PR #5631, opened 2021-11-01
Update bus.json

alessandrocarmeli:
tpl fvg added


UKChris-osm:
This seems to be a repeat of #5630

PR #5652, opened 2021-11-05
fix Rossmann name

### LINKED ISSUE #5510: brand=ROSSMANN is incompatible with the naming convention
According to the accepted convention of using names, it should be "Rossman" in "brand" and "name"

matkoniecz:
stylistic capitalization is uniformly used in marketing materials (typically also with red color and embedded horseman symbol).

It is ignored by people outside Rossmann, and therefore should not be put into `name` tag.

Fixes #5510

### DIFF
--- a/data/brands/shop/chemist.json
+++ b/data/brands/shop/chemist.json
@@ -329,16 +329,16 @@
       }
     },
     {
-      "displayName": "ROSSMANN",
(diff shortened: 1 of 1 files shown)


bhousel:
I don't really understand your comment about marketing, but ok - this seems like a pretty minor change.

PR #5654, opened 2021-11-06
Divide Stadtmobil into single operators 

sb12:
As discussed in #5536:
- Added locationSet for each one
- Added the local operators
- Give Stadtmobil Südbaden its own brand


bhousel:
This looks great, thanks @sb12! 👍

PR #5655, opened 2021-11-06
Carsharing: Combine Flinkster with DB Connect GmbH

sb12:
Flinkster is a brand and car sharing network operated by Deutsche Bahn Connect GmbH.


bhousel:
Thanks @sb12 👍 

Another tip: you don't need to add `matchNames` for
https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#matchnames-matchtags

PR #5656, opened 2021-11-06
Update Deutsche Bahn ticket machines

sb12:
The ticket vending machines branded as Deutsche Bahn are operated by DB Vertrieb GmbH, a subsidiary of Deutsche Bahn.
They don't have any names usually.


arch0345:
Looks good 👍

PR #5663, opened 2021-11-06
Updating Sheetz (Fuel)

arnowrld:


data/brands/amenity/fuel.json
```diff
@@ -3883,7 +3883,7 @@
     {
       "displayName": "Sheetz",
       "id": "sheetz-c5cf43",
-      "locationSet": {"include": ["us"]},
+      "locationSet": {"include": ["us-md.geojson", "us-nc.geojson", "us-oh.geojson", "us-pa.geojson", "us-va.geojson", "us-wv.geojson"]},
```
tas50, line 3886: Is there a particular reason to limit this to a set of states? Is it conflicting with other entries?

PR #5665, opened 2021-11-07
Update bank.json

rkdclgh:
Update the list of Korean Bank

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -7646,6 +7646,10 @@
       "displayName": "SC제일은행",
       "id": "2a04be-10d607",
       "locationSet": {"include": ["kr"]},
+      "matchNames": [
+        "스탠다드차타드은행",
+        "제일은행"
+      ],
       "tags": {
         "amenity": "bank",
         "brand": "SC제일은행",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/bank.json
```diff
@@ -11196,15 +11213,16 @@
       "displayName": "하나은행",
       "id": "koreaexchangebank-10d607",
       "locationSet": {"include": ["kr"]},
+      "matchNames": ["KEB하나은행"],
       "tags": {
         "amenity": "bank",
         "brand": "하나은행",
-        "brand:en": "Korea Exchange Bank",
+        "brand:en": "KEB Hana Bank",
```
arch0345, line 11220: According to the [English Wikipedia article](https://en.wikipedia.org/wiki/Hana_Bank), it looks like they're branded in South Korea as just "Hana Bank" in English. Let me know if this is not the case

data/brands/amenity/bank.json
```diff
@@ -11080,62 +11084,73 @@
...
+        "name:en": "NH Bank",
+        "name:ko": "NH농협은행"
       }
     },
     {
       "displayName": "새마을금고",
       "id": "ca214a-10d607",
       "locationSet": {"include": ["kr"]},
+      "matchNames": [
+        "MG새마을금고",
```
arch0345, line 11151: Based on [their website](https://www.kfcc.co.kr/) it looks like they brand themselves as "MG새마을금고"

arch0345 (commented):
Thanks @rkdclgh for these updates! I've given you some feedback

data/brands/amenity/bank.json
rkdclgh: In general, it is called "Hana Bank", but they have designated the official English name, "KEB Hana Bank".

data/brands/amenity/bank.json
rkdclgh: I updated it. Thanks!

PR #5672, opened 2021-11-08
IntercityHotel: no space in brand and name + include ch

sb12:
This is how the name is spelt on the hotel's chain website and their brochures etc., as well as in their logo. Wikidata also uses this spelling.
There is an older logo where it was spelt InterCityHotel. 
Taginfo finds all 3 variants for the name tag in use, but it looks like the values where heavily influenced by NSI,so it does not really represent the mapper's choice.
I included Switzerland, because there seems to be one in Zürich.

### DIFF
--- a/data/brands/tourism/hotel.json
+++ b/data/brands/tourism/hotel.json
@@ -1273,11 +1273,12 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


arch0345:
Thanks for catching this @sb12!

PR #5673, opened 2021-11-09
Updating Locations in Hotels

arnowrld:
I thought this would help in terms of updating the nsi.


data/brands/tourism/hotel.json
```diff
@@ -61,7 +61,41 @@
     {
       "displayName": "AC Hotel",
       "id": "achotel-779ccb",
-      "locationSet": {"include": ["001"]},
+      "locationSet": {
+        "include": [
+          "at",
+          "au",
+          "ca-qc.geojson",
```
arch0345, line 68: I know it might be a little confusing that some of these .geojsons are their full name while others are ISO 3166-2. A list of .geojson features can be found in the [features directory](https://github.com/osmlab/name-suggestion-index/tree/main/features)
```suggestion
          "quebec.geojson", 
```

arch0345 (changes requested):
It looks like the build error was due to an invalid locationSet. I've added a suggestion

tas50:
I'm a bit concerned that we're making the index less useful with changes to large international brands like this. Many of these brands that operate in 20 countries will easily support their 21st or 22nd country within the next few years. When that change happens there's very little chance we'll notice and update the index. IMO we should only be using country codes with companies that operate in a single country or where there's a conflicting name to avoid. As soon as a brand goes multi-national we should just leave it at 001.

arch0345:
> I'm a bit concerned that we're making the index less useful with changes to large international brands like this. Many of these brands that operate in 20 countries will easily support their 21st or 22nd country within the next few years. When that change happens there's very little chance we'll notice and update the index. IMO we should only be using country codes with companies that operate in a single country or where there's a conflicting name to avoid. As soon as a brand goes multi-national we should just leave it at 001.

After reading this I agree that it isn't useful to list a whole bunch of countries for these international brands. A couple months ago I remember needing to add Scotland to DoubleTree after noticing iD matching locations to Hilton instead.

arnowrld:
Maybe add stuff like the region they're a part of (or at least more), for instance, Europe ["150"], the Middle East, Greater China, Oceania, North America, etc.

peternewman:
> > I'm a bit concerned that we're making the index less useful with changes to large international brands like this. Many of these brands that operate in 20 countries will easily support their 21st or 22nd country within the next few years. When that change happens there's very little chance we'll notice and update the index. IMO we should only be using country codes with companies that operate in a single country or where there's a conflicting name to avoid. As soon as a brand goes multi-national we should just leave it at 001.
> 
> After reading this I agree that it isn't useful to list a whole bunch of countries for these international brands. A couple months ago I remember needing to add Scotland to DoubleTree after noticing iD matching locations to Hilton instead.

Do we need a CI report (or something on the website) that lists conflicting names so they can be investigated to see if they can be narrowed down (and just leave everything else as 001)?

LaoshuBaby (changes requested):
Too long, 001 is better
Maybe tomorrow it will open a new hotel in a new country

bhousel:
I agree with what others have said here, that I think once a brand expands beyond a few countries, and especially if it's actively expanding, it doesn't make sense for this index to try to track every place in detail.  

There is also a performance penalty for unioning together a lot of things in a locationSet, but I'm not too concerned about this to actually measure it.  It might just mean iD users need to wait a few seconds extra for the NSI presets to be available.

I'd switch to `001` if not sure, or some combination of other M49 codes like listed here:
https://en.wikipedia.org/wiki/UN_M49
(btw that Wikipedia page got nicer recently!)

Also you can test these codes on https://location-conflation.com/ or https://ideditor.codes if you want to see on a map what they cover.

peternewman:
As a thought for an alternative, could we not use both, possibly with a new locationSet category.

Use it as a weighting method:
1. - if it's got a locationSet include and not an exclude, show it
2. - if it's in an exclude, don't show it (e.g. the special Burger King)
3. - if it's got neither, offer it as a lower priority

Then we could default to blank locationSet rather than 001 and they'd all fall into the third category.

We'd get a hotel which matched out specified area if we'd specified one, but we'd still be able to see the others if we hadn't populated it.

bhousel:
Closing, it sounds like people are in general agreement that this level of detail in the locationSet is too much.

PR #5676, opened 2021-11-09
Add Costa Drive Thru

peternewman:
See https://github.com/openstreetmap/iD/issues/8794

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -992,6 +992,24 @@
       }
     },
     {
+      "displayName": "Costa Drive Thru",
+      "locationSet": {"include": ["001"]},
+      "matchNames": [
+        "costa drive-thru",
+        "costa drive-through"
+      ],
+      "tags": {
+        "amenity": "cafe",
+        "brand": "Costa",
+        "brand:wikidata": "Q608845",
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @peternewman 👍 
It's missing a comma, but I'll just merge it and fix that..

peternewman:
Was just in the process of fixing that... You should probably protect branches so they can only merge then they're passing CI!

I'm also going to add the unhyphenated "drive through" variant.

bhousel:
> Was just in the process of fixing that... You should probably protect branches so they can only merge then they're passing CI!

Haha, I usually prefer to just merge small mistakes and fix them right away rather than holding up a PR until it passes CI. 

> I'm also going to add the unhyphenated "drive through" variant.

The matcher actually ignores variation in punctuation, so these extra match names are not needed 😄

peternewman:
> Haha, I usually prefer to just merge small mistakes and fix them right away rather than holding up a PR until it passes CI.

Ah fair enough, I thought you'd always want to keep your main branch valid and clean, even if it meant you editing the PR yourself.

> The matcher actually ignores variation in punctuation, so these extra match names are not needed smile

Ah damn, I was wondering...

PR #5684, opened 2021-11-11
Update mobile_phone.json

giorgiomec:
TIM's company name is TIM Group (officially TIM S.p.A.); Wind is now Wind Tre


data/brands/shop/mobile_phone.json
```diff
@@ -10,7 +10,7 @@
   },
   "items": [
     {
-      "displayName": "3 Store",
+      "displayName": "3",
       "id": "3store-e135a8",
       "locationSet": {"include": ["gb", "it"]},
       "matchNames": ["3", "three", "three store"],
```
arch0345, line 16: To match existing 3 Stores to this modified entry
```suggestion
       "matchNames": ["3 store", "three", "three store"],
```

arch0345 (changes requested):

UKChris-osm:
I think three stores are now branded as "_three._" (as in the word, not the number) aren't they?

arch0345:
> I think three stores are now branded as "_three._" (as in the word, not the number) aren't they?

Yeah this seems to be the case - most of if not all the locations I saw on https://locator.three.co.uk/ show these stores being signed as "Three."

arch0345:
> TIM's company name is TIM Group (officially TIM S.p.A.)

TIM Group seems to just be the English translation to Gruppo TIM on Wikidata: maybe this could be changed to Gruppo TIM?

data/brands/shop/mobile_phone.json
```diff
@@ -10,15 +10,15 @@
   },
   "items": [
     {
-      "displayName": "3 Store",
+      "displayName": "",
```
bhousel, line 13: It needs a `displayName` here.  It should probably be the same as the name "Three."

data/brands/shop/mobile_phone.json
```diff
@@ -843,14 +843,14 @@
...
+      "displayName": "TIM Group",
       "id": "tim-a04a47",
-      "locationSet": {"include": ["br", "it"]},
+      "locationSet": {"include": ["br", "it", "sm"]},
       "tags": {
         "brand": "TIM",
         "brand:wikidata": "Q144617",
         "brand:wikipedia": "en:Gruppo TIM",
-        "name": "TIM",
+        "name": "TIM Group",
```
bhousel, line 853: Are these stores actually called "TIM Group", like on signs and stuff?  
The name is more what people commonly refer to the store as, not the official company name.

data/brands/shop/mobile_phone.json
```diff
@@ -981,14 +981,14 @@
...
-      "displayName": "Wind",
+      "displayName": "Wind Tre",
       "id": "wind-c4d7ba",
       "locationSet": {"include": ["it"]},
       "tags": {
-        "brand": "Wind",
-        "brand:wikidata": "Q650467",
-        "brand:wikipedia": "en:Wind (Italy)",
-        "name": "Wind",
+        "brand": "Wind Tre",
```
bhousel, line 988: When stores re-brand, we try to match the old name in an `alt_name` tag or `matchNames` property.  

I'm not sure how completely this rebrand has happened in Italy?  Are all stores now signed as "Wind Tre" or is it still a mix?

bhousel (changes requested):
I have some questions about the changes being proposed..

data/brands/shop/mobile_phone.json
arch0345: Yeah, I forgot to replace the displayName

UKChris-osm:
Currently 3 is set to gb and it, but has the Wind Tre merger taken over from 3 in Italy?

I've added 3 Ireland as I think they still refer to their stores as "3 Stores", but should the Italy location be removed from the current 3 store entry?

bhousel (approved):

bhousel:
Merging as-is..  It's probably fine.

boothym:
If all of the UK's Three stores have changed their name, shouldn't there be a mass edit of the shops in OSM?

PR #5697, opened 2021-11-18
Updating & Adding Hotel Chains

arnowrld:
Can somebody help me with the Wikipedia thing? I never did that in my editing, since this took days to edit.

### DIFF
--- a/data/brands/tourism/hotel.json
+++ b/data/brands/tourism/hotel.json
@@ -85,7 +85,7 @@
     {
       "displayName": "Adagio",
       "id": "adagio-612c69",
-      "locationSet": {"include": ["150"]},
+      "locationSet": {"include": ["001"]},
       "tags": {
         "brand": "Adagio",
         "brand:wikidata": "Q2823880",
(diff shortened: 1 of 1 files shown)


arch0345:
I wouldn't worry about the `brand:wikipedia` deletions - these will get readded the next time the `npm run wikidata` script is used (usually right before the next release).

data/brands/tourism/hotel.json
```diff
@@ -82,6 +82,17 @@
         "tourism": "hotel"
       }
     },
+    {
+      "displayName": "Adagio",
+      "id": "adagio-612c69",
+      "locationSet": {"include": ["150"]},
```
arch0345, line 88: Their [website](https://www.adagio-city.com/gb/destinations.shtml) shows they also have locations in the Middle East and Brazil
```suggestion
      "locationSet": {"include": ["001"]},
```

arch0345 (changes requested):

arch0345 (approved):

PR #5701, opened 2021-11-18
Add "Norges speiderforbund" and "Norges KFUK-KFUM-speidere"

noenandre:
Created new category for club=scout, and added to scout associations.


arch0345:
Thanks for your patience, this PR looks good!

PR #5704, opened 2021-11-18
Add hairdresser Cutters.

noenandre:


data/brands/shop/hairdresser.json
```diff
@@ -109,6 +109,19 @@
         "shop": "hairdresser"
       }
     },
+    {
+      "displayName": "Cutters",
+      "id": "cutters-a2b284",
+      "locationSet": {"include": ["no"]},
```
arch0345, line 115: Based on their [website](https://www.cutters.no/) it looks like they also have locations in Finland and Sweden
```suggestion
      "locationSet": {"include": ["fi", "no", "se"]},
```

arch0345 (changes requested):

data/brands/shop/hairdresser.json
noenandre: Yes, you are right. I read through the Norwegian Wikipedia article only, which start of with "Cutters are a Norwegian hairdresser chain..."

arch0345 (approved):

PR #5708, opened 2021-11-19
Update bus.json

alessandrocarmeli:
updats tpl-fvg


arch0345:
`matchTags` is only used for entries that commonly get mistagged (ie McDonalds as `amenity=restaurant` instead of `amenity=fast_food`). IIRC the index already matches operators with network tags on the NSI

PR #5712, opened 2021-11-21
Exclude Bulgaria from Unicredit Bank

Dimitar5555:
Unicredit operates under the Unicredit Bulbank brand in Bulgaria.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -8630,7 +8630,7 @@
     {
       "displayName": "UniCredit Bank",
       "id": "unicreditbank-b7a026",
-      "locationSet": {"include": ["001"], "exclude: ["bg"]},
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Thanks @Dimitar5555 - are you able to add "_Unicredit Bulbank_" for Bulgaria to this PR as well, otherwise it leaves Bulgaria with no NSI entry for the Unicredit brand.

Dimitar5555:
There is an entry for Unicredit Bulbank at https://github.com/osmlab/name-suggestion-index/blob/98a7a1efcb5ee069e1899ea0f889b34c957cbbfa/data/brands/amenity/bank.json#L10409

UKChris-osm:
Ah I see, thanks @Dimitar5555

PR #5715, opened 2021-11-21
Fix Sigo Bicycle rental

coolultra1:
This was kind of ~~lovelessly~~ automatically added, so I fixed it. Not really sure if removing the unique id was right, but I think so, since I'm also changing the name.

### DIFF
--- a/data/brands/amenity/bicycle_rental.json
+++ b/data/brands/amenity/bicycle_rental.json
@@ -2914,13 +2914,13 @@
       }
     },
     {
-      "displayName": "sigo E-Lastenrad Standort",
-      "id": "sigoelastenradstandort-06d22c",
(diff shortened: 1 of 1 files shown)


coolultra1:
Just took a look at the `network` key's wiki page again, and now think it should be added. The thing is: With this rental service, you must return your bike at the station you got it from. But since
> this key identifies an organization or brand that operates a similar amenity in various other locations

, I think having `network` is alright.

UKChris-osm:
It's okay to remove an ID @coolultra1 as this will be changed automatically if the name and / or locationSet is changed.

coolultra1:
@UKChris-osm That's nice, then I did it right, thanks, so this is mergable now.

LaoshuBaby (approved):

PR #5716, opened 2021-11-21
Set bus operator to TPL FVG on 3 networks

alessandrocarmeli:
new TPL FVG added

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -636,7 +636,7 @@
         "network": "APT",
         "network:wikidata": "Q3601116",
         "network:wikipedia": "it:APT Gorizia",
-        "operator": "TPL FVG"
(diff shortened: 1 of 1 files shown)


data/transit/route/bus.json
```diff
@@ -636,6 +636,7 @@
         "network": "APT",
         "network:wikidata": "Q3601116",
         "network:wikipedia": "it:APT Gorizia",
+        "operator": "TPL FVG"
```
UKChris-osm, line 639: Just an FYI, there is a missing comma at the end of this line, causing a build error.

alessandrocarmeli: see https://github.com/osmlab/name-suggestion-index/pull/5716/commits/480f13dac5543b57afd4839f4b17261ec141bfdb

PR #5723, opened 2021-11-24
Added Ymanci as a french financial office

le-jun:


data/brands/office/financial.json
```diff
@@ -66,6 +66,16 @@
...
+      },
+      {
+      "displayName": "Ymanci",
+      "locationSet": {"include": ["fr"]},
+      "tags": {
+        "brand": "Ymanci",
+        "name": "Ymanci",
+        "operator": "Premista",
+        "office": "financial",
+        "note": "Three branches: Assurances, Crédit immobilier and Regroupement de crédits"
```
LaoshuBaby, line 78:
```suggestion
```

data/brands/office/financial.json
```diff
@@ -66,6 +66,16 @@
...
         "office": "financial"
+      },
+      {
+      "displayName": "Ymanci",
+      "locationSet": {"include": ["fr"]},
+      "tags": {
+        "brand": "Ymanci",
+        "name": "Ymanci",
+        "operator": "Premista",
+        "office": "financial",
```
LaoshuBaby, line 77:
```suggestion
        "office": "financial"
        }
```

LaoshuBaby (changes requested):
1. Please be careful for JSON's format, especially the `{}` and `,`
2. About note, if needed, add it outer the tags. See: https://github.com/osmlab/name-suggestion-index/wiki/Contributing#note-optional

data/brands/office/financial.json
le-jun: In this case I wanted the note to be adressed to NSI users in case of further development of the operator, not towards OSM users. I thought it fitted the "brand's status" note category.

LaoshuBaby: so you don‘t need to put it here
if it don’t for OSM mapper but for NSI dev，it should outer `tags`

LaoshuBaby:
Note: If a entry haven't got a wikidata, it won't be suggested by NSI in iD/RapiD

So I've created one and add it in [this commit](https://github.com/osmlab/name-suggestion-index/commit/287a453d64fbbe52a9ab2e81bb4f1a91e7a33a38).

https://github.com/osmlab/name-suggestion-index/blob/287a453d64fbbe52a9ab2e81bb4f1a91e7a33a38/data/brands/office/financial.json#L72-L81

le-jun:
Thanks a lot for your help!

PR #5724, opened 2021-11-24
add matchNames for Сушишоп and Суши wok

afanasovich:
I wrote all possible spellings of the names of two sushi bar chains.
This is my first experience of editing a file on github, I could make mistakes

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -6412,17 +6412,8 @@
       "displayName": "Суши Wok",
       "id": "sushiwok-af0c49",
       "locationSet": {"include": ["ru", "ua"]},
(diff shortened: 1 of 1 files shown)


LaoshuBaby (changes requested):
Maybe we don't need so much `matchNames` in one entry.

You can look for this document https://github.com/osmlab/name-suggestion-index/wiki/Contributing#matchnamesmatchtags-optional

LaoshuBaby:
图片

As the same question I said in #5726 

> If you don't know whether a matchName is needed, you can run `npm run build` to fix it, if a name can be matched by current `brand`, `name` and other tag, it will be removed.
> 
> Besides, you only need to use lowercase char in matchNames, uppercase char will be replaced by former one.

PR #5726, opened 2021-11-24
Add matchNames for "Dodo Pizza"

afanasovich:


LaoshuBaby:
Hey! those matchNames are too much!

If you don't know whether a matchName is needed, you can run `npm run build` to fix it, if a name can be matched by current `brand`, `name` and other tag, it will be removed.

图片

PR #5728, opened 2021-11-24
Add brand for Better (Greenwich Leisure Limited) sports centres

codeinabox:


bhousel:
This might be ok, but I think it might need a separate wikidata for the "Better" brand to actually be used.

codeinabox:
Thank you. I don't believe Better has its own Wikidata entry as as their website states:
> Better is a registered trademark and trading name of GLL (Greenwich Leisure Limited)

LaoshuBaby:
> This might be ok, but I think it might need a separate wikidata for the "Better" brand to actually be used.

But the brand haven't found……This situation had meet not only once
Only the operator have a wikidata, the brand sometimes was redirect to company's page……

Or if needed, maybe we can add a script automatic create wikidata for those kind of brand? 

(#ﾟДﾟ)

bhousel:
I wouldn't use a script and I think automatically creating these pages with a script would probably upset the Wikidata community.

Ideally someone just needs to create a good Wikidata page for Better, add supporting documentation for it, and add a claim that it's owned by Greenwich Leisure Limited.

https://www.better.org.uk/

LaoshuBaby:
> Ideally someone just needs to create a good Wikidata page for Better, add supporting documentation for it, and add a claim that it's owned by Greenwich Leisure Limited.
> 
> https://www.better.org.uk/

Yes, so I try to created a wikidata, and add to this entry in https://github.com/osmlab/name-suggestion-index/commit/0a9d7ec67f3345c8f6269c6e37402c3878f8e4c0

https://github.com/osmlab/name-suggestion-index/blob/0a9d7ec67f3345c8f6269c6e37402c3878f8e4c0/data/brands/leisure/sports_centre.json#L57-L67

Issue #5647, opened 2021-11-04
How to handle Caterpillar dealership brands


tas50:
We have a Caterpillar brand tag, but Caterpillar doesn't have their own stores. Here in the PNW they are either PAPE Caterpillar or Peterson Caterpillar. PAPE is actually now owned by Caterpillar, but operated under that brand name still. I imagine worldwide there are additional dealership brands that we should include since it does matter to those buying the equipment.

bhousel:
I guess these are sort of like Polaris #5233 right?  I probably wouldn't put Caterpillar in the brand tag unless the store were branded just as that.

But it does look like both Peterson Caterpillar and Papé have enough locations that they'd be considered notable, so I think it makes sense to add those: 
https://www.petersoncat.com/about/locations
https://www.pape.com/locations

arch0345:
I just removed CAT and started to add some chains of dealers such as Peterson CAT and Papé

Issue #5648, opened 2021-11-04
Add GoHealth Urgent Care


tas50:
Urgent Care clinics in the US:

https://www.gohealthuc.com/

arch0345:
I was going to add this when I added Legacy to the index a couple months ago, but I wasn't sure if or GoHealth or the partners listed on the website operate these facilities since I noticed the Legacy partnered clinics are branded with both logos.

PR #5734, opened 2021-11-25
Add local Goodwill affiliates as operators

1ec5:
Added a number of local [Goodwill](https://www.wikidata.org/wiki/Q5583655) affiliates to the `operator` tree. These are the Goodwill affiliates that have been tagged (correctly or incorrectly) in OpenStreetMap so far, plus any other affiliates that Wikidata happened to have an item for. This list is likely missing some of the largest affiliates in the United States, but I couldn't readily find a list of them to guide this PR.

## Rationale

The Goodwill brand is shared among a network of independent nonprofits, many of which retain some degree of branding. For example, Goodwill of Silicon Valley posts its full name prominently at each of its stores and donation centers, as well as on its trucks and donation bins. On the other hand, some other Goodwill affiliates use the parent organization’s branding as is.

[kiely](https://www.mapillary.com/app/?pKey=278303837272609)

People tend to care which organization is enabling them to get a tax deduction on their donated items. The Wikidata items are tagged with IRS EINs and Charity Navigator profile IDs, so there could be a real use case for cross-referencing the operator.

I think it makes sense to keep all the stores tagged with the same `brand` for the parent organization but with different `operator`s. This differs from how we’re dealing with Caterpillar in #5647, but anyways there’s a significant difference between a dealer network and how Goodwill operates.

## Caveats

* I’m not sure if the new entries should specify `shop=charity`, since that would be covered by Goodwill’s existing `brand` entry. Can I omit it, in case someone tags a donation bin with `amenity=recycling` and `operator=Goodwill of …`?
* I set the `locationSet` of each affiliate to the state or states they operate in. For Wikidata, I did a bit more research and came up with the specific counties that some of them operate in, but I didn’t bother to add any GeoJSONs for those service areas to NSI.
* Some of the Wikidata items are tagged with TikTok and Snapchat user names, in case anyone feels like adding [social links](https://github.com/osmlab/name-suggestion-index/blob/fce09fc75976493749ef2162b666211c3bc3ff3d/app/src/CategoryRowSocialLinks.jsx#L17-L64) for them. I didn’t, because I still don’t get it.


arch0345:
This situation is similar to GoHealth Urgent Care (#5648): local companies seem to operate the same brand

bhousel:
> * I’m not sure if the new entries should specify `shop=charity`, since that would be covered by Goodwill’s existing `brand` entry. Can I omit it, in case someone tags a donation bin with `amenity=recycling` and `operator=Goodwill of …`?

I think all of these should just be in the brand file?  

e.g. an item should look like:
```js
    {
      "displayName": "Goodwill (Florida)",
      "locationSet": {
        "include": ["us-fl.geojson"]
      },
      "tags": {
        "brand": "Goodwill",
        "brand:wikidata": "Q5583655",
        "brand:wikipedia": "en:Goodwill Industries",
        "name": "Goodwill",
        "operator": "Goodwill Industries of South Florida",
        "operator:wikidata": "Q54624094",
        "shop": "charity"
      }
    },
```

That's like saying, Goodwill (Florida) will override the Goodwill (USA), but only in Florida.

It's better to think of these as presets that are exclusively applied in a particular region - not as rulesets that can combine together depending on what tags are present.

1ec5:
Interesting, I didn’t realize brand entries could specify `operator`s like that. I think the display names would still have to be more specific, because I didn’t create features for each affiliate. Some affiliates divide up counties, while others have overlapping service areas.

data/operators/shop/charity.json
```diff
@@ -4,6 +4,18 @@
     "exclude": {"generic": ["^charity$"]}
   },
   "items": [
+    {
+      "displayName": "Goodwill Industries of the Columbia Willamette",
+      "id": "goodwillindustriesofnorthlouisiana-239b23",
```
1ec5, line 9: Needs a different ID.

arch0345:
Merging this as is, will continue to add more affiliates in the future

PR #5737, opened 2021-11-25
add Tinkoff Bank

afanasovich:


LaoshuBaby:
ATM data can be inherit from bank.json
You can see that atm.json is short But bank.json is very long. Only If this brand only have atm machine service and haven't a traditional bank bussiness, then you only Add It in atm.json
If not, please Add It to bank.json

LaoshuBaby:
> Only If this brand only have atm machine service and haven't a traditional bank bussiness, then you only Add It in atm.json

Sorry, It is my fault, I misread the "direct bank" in wikipedia. Sure, you are right, this is a atm based bank.

PR #5738, opened 2021-11-25
Update police.json

Crashillo:
Since _Mossos d'Esquadra_ is the autonomous police force in catalonia, is commonly known by its catalan name. Its spanish translation is just... weird.


arch0345:
Thanks @Crashillo!

PR #5741, opened 2021-11-25
Fix typo "netwrok -> network

JesseWeinstein:
And put it under "network:wikipedia:zh" since there is already a "network:wikipedia" tag.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -18151,7 +18151,7 @@
         "network:wikidata": "Q3462238",
         "network:wikipedia": "en:Taichung City Bus",
         "network:zh": "臺中市市區公車",
-        "network:wikipedia:zh": "zh:臺中市市區公車",
+        "network:wikipedia": "zh:臺中市市區公車",
         "route": "bus"
       }
     },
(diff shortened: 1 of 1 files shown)


data/transit/route/bus.json
```diff
@@ -18151,7 +18151,7 @@
         "network:wikidata": "Q3462238",
         "network:wikipedia": "en:Taichung City Bus",
         "network:zh": "臺中市市區公車",
-        "netwrok:wikipedia": "zh:臺中市市區公車",
+        "network:wikipedia:zh": "zh:臺中市市區公車",
```
arch0345, line 18154: I would just replace the English Wikipedia article with the Chinese one since this is the local language of this network
```suggestion
        "network:wikipedia": "zh:臺中市市區公車",
```

data/transit/route/bus.json
```diff
@@ -18151,7 +18151,7 @@
         "network:wikidata": "Q3462238",
         "network:wikipedia": "en:Taichung City Bus",
```
arch0345, line 18152:
```suggestion
```

arch0345 (changes requested):
Thanks and welcome to the NSI community! I made some suggestions, you can add these to your PR by clicking on 'commit suggestion'

LaoshuBaby:
I think the same with @arch0345 
you don't need to add multi wikipedia link in oneentry，if you need you can search this wikidata。The practise in NSI is wikipedia follow the location's local lang

JesseWeinstein:
Sounds good! I just didn't want to remove information, and I wasn't aware of the prefer-local-language guideline. Thanks for merging it!

PR #5746, opened 2021-11-27
TRAID operates the recycling containers

codeinabox:
TRAID operates the recycling containers as the phone number on the containers is theirs.


bhousel:
Thanks @codeinabox 👍

PR #5754, opened 2021-11-28
Add Marugame Udon

peternewman:
I think this is the UK branding for "Marugame Seimen".

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -4953,6 +4953,18 @@
       }
     },
     {
+      "displayName": "Marugame Udon GB",
+      "locationSet": {"include": ["gb"]},
+      "tags": {
+        "amenity": "restaurant",
+        "brand": "Marugame Udon",
+        "brand:wikidata": "Q10877253",
+        "brand:wikipedia": "en:Marugame Seimen",
+        "cuisine": "noodle",
+        "name": "Marugame Udon",
+      }
+    },
+    {
       "displayName": "便宜坊",
       "id": "bianyifang-0a8e9b",
       "locationSet": {"include": ["cn"]},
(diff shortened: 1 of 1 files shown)


data/brands/amenity/restaurant.json
```diff
@@ -4952,6 +4952,18 @@
...
+    {
+      "displayName": "Marugame Udon GB",
+      "locationSet": {"include": ["gb"]},
+      "tags": {
+        "amenity": "restaurant",
+        "brand": "Marugame Udon",
+        "brand:wikidata": "Q10877253",
+        "brand:wikipedia": "en:Marugame Seimen",
+        "cuisine": "noodle",
+        "name": "Marugame Udon",
```
LaoshuBaby, line 4964:
```suggestion
        "name": "Marugame Udon"
```

LaoshuBaby (approved):

peternewman:
Also the other one is in more places now (according to the UK site if they're the same):
https://marugame.co.uk/faqs

bhousel:
> Also the other one is in more places now (according to the UK site if they're the same):
https://marugame.co.uk/faqs

I added these other countries to the item in 5ce0def46, also added a `"alt_name": "Marugame Seimen"` tag, since it seems to sometimes be called that..

peternewman:
> I added these other countries to the item in [5ce0def](https://github.com/osmlab/name-suggestion-index/commit/5ce0def469fbbb7bd4dba91b649b64de70af7a09), also added a `"alt_name": "Marugame Seimen"` tag, since it seems to sometimes be called that..

Thanks @bhousel , I can only guess that they changed the second part of the name due to unfortunately sounding similar to bodily fluids that wouldn't be very appetising... 🤮

PR #5757, opened 2021-11-29
Add wikidata to pharmacy «Вита» (Vita).

afanasovich:
The pharmacy Вита (Vita) has two formats of pharmacy: Вита Центральная (Vita Central —  big shop format) and Вита Экспресс (Vita Express — small format).
I create https://www.wikidata.org/wiki/Q109810737 for this brand.

### DIFF
--- a/data/brands/amenity/pharmacy.json
+++ b/data/brands/amenity/pharmacy.json
@@ -2548,6 +2548,7 @@
       "tags": {
         "amenity": "pharmacy",
         "brand": "Вита Центральная",
+        "brand:wikidata": "Q109810737",
         "healthcare": "pharmacy",
         "name": "Вита Центральная"
       }
@@ -2559,6 +2560,7 @@
       "tags": {
         "amenity": "pharmacy",
         "brand": "Вита Экспресс",
+        "brand:wikidata": "Q109810737",
         "healthcare": "pharmacy",
         "name": "Вита Экспресс"
       }


LaoshuBaby:
Good! Thank you for this local knowledge.

So what about this entry? 

```
"displayName": "Вита",
"id": "30d803-288c27",
```

afanasovich:
> Good! Thank you for this local knowledge.
> 
> So what about this entry? Need we delete it?
> 
> ```
> "displayName": "Вита",
> "id": "30d803-288c27",
> ```

I don’t know.

These are three different pharmacy chains.

https://vitaexpress.ru/ — I add
https://www.aptekavita.ru/ —  id 30d803-288c27
https://vita-plus.ru/ — id 338794-288c27

I live in a region where only Vita Express and Vita Central operate.  How big are the other two networks - I don't know.  I learned about their existence today from Google)

afanasovich:
@LaoshuBaby 👋

Could you please explain to me, how long it takes for these changes to be reflected in the iD-interface? Or what does it depend on? Should iD contributors release a new version or some other event?

bhousel:
> Could you please explain to me, how long it takes for these changes to be reflected in the iD-interface? Or what does it depend on? Should iD contributors release a new version or some other event?

Whenever the NSI project makes a release, iD and RapiD will pick up the new data files.  I do this about once a week, but I can do it anytime. 

It’s not completely automatic because the script to sync up wikidata takes about 15 minutes to run, and sometimes main branch is broken, so we want to make sure everything is ok before publishing.

PR #5767, opened 2021-12-01
Adding Various Crypto ATM networks

arnowrld:


bhousel:
Thanks @arnowrld  - I'm a bit concerned that these look like Wikidata pages that you just added this morning, but I clicked through to the atm sites and they do seem like real things.

bhousel:
We should probably add whatever OSM tags are appropriate for whatever currency you can withdraw from these things.

PR #5775, opened 2021-12-01
add wikidata tags for Cellarbrations and Bottlemart

28688:


arch0345:
Thanks @28688 and welcome to the NSI community! I highly recommend adding Facebook/Twitter usernames for the wikidata page you created if they have one as that's how logos in the index are collected.

PR #5780, opened 2021-12-02
Remove redundant information

Dimitar5555:


bhousel:
It looks like these tags are the ones related to stuff we learned on https://github.com/osmlab/name-suggestion-index/issues/2245

> I've updated the CONTRIBUTING doc and also added name:xx and brand:xx tags for situations where the language used is unambiguous. Rule going forward is as stated above:
>
> "If we set a name:xx tag, make sure there exists a name:xx tag that matches the name tag."
> "If we set a brand:xx tag, make sure there exists a brand:xx tag that matches the brand tag."

Basically, some validator software (KeepRight, JOSM, maybe others?) expects for at least one of the multilingual tags to match the name tag.  So yes they seem redundant but they are there to avoid warnings.

Dimitar5555:
I've added them. There are also some changes to names and new short names. It can be merged.

bhousel:
thanks @Dimitar5555 👍

PR #5788, opened 2021-12-04
e.dis -> E.DIS Netz GmbH

sb12:
E.DIS Netz GmbH is the subsidiary of E.DIS that operates the power grid. 
Spelling according to the *Impressum* on their website. This is also the preferred tagging of the majority of mappers according to [taginfo](https://taginfo.openstreetmap.org/search?q=dis#values). I didn't touch the wikidata and wikipedia tags as I could not find a wikidata page for E.DIS Netz GmbH.

### DIFF
--- a/data/operators/power/line.json
+++ b/data/operators/power/line.json
@@ -1207,16 +1207,15 @@
       }
     },
     {
-      "displayName": "e.dis",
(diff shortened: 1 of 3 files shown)


arch0345:
Thanks for fixing this @sb12!

PR #5800, opened 2021-12-06
the Marks & Spencer is represented in Russia

afanasovich:


data/brands/shop/department_store.json
```diff
@@ -885,9 +885,10 @@
       "displayName": "Marks & Spencer",
       "id": "marksandspencer-75c234",
       "locationSet": {
-        "include": ["gb", "gr", "ie"]
+        "include": ["gb", "gr", "ie", "ru"]
       },
-      "matchNames": ["m and s"],
+      "matchNames": ["m and s", "Маркс и Спенсер"],
+      "matchTags": ["shop/clothes"],
```
bhousel, line 891: This is a good thought, however `shop/clothes` is actually redundant and will be removed automatically the next time the build script runs. 

The matching code in this project has a concept of "match groups" - it means that the tag pairs in a match group are considered similar enough that we will also automatically match the other tag pairs in the group.

https://github.com/osmlab/name-suggestion-index/blob/9be7d79183abfb741ee95888e47e6489e30e1b09/config/matchGroups.json#L77-L87

bhousel:
Thanks @afanasovich , this can merge as-is, I just wanted to mention the match group behavior so you know why the build will remove your matchTag.

PR #5727, opened 2021-11-24
LEON has vegan and vegetarian options, no longer operating in the US

codeinabox:
LEON has vegan and vegetarian options on its menu https://leon.co/menu/

It has also pulled out of the US market https://www.thecaterer.com/news/leon-pulls-america-states-market-pandemic-toll-2021

PR #5801, opened 2021-12-06
Crosstown doughnuts

codeinabox:


data/brands/amenity/fast_food.json
```diff
@@ -1210,6 +1210,20 @@
...
+    {
+      "displayName": "Crosstown",
+      "locationSet": {"include": ["gb"]},
+      "matchNames": ["crosstown doughnuts", "crosstown donuts"],
+      "tags": {
+        "amenity": "fast_food",
+        "brand": "Crosstown",
+        "brand:wikidata": "Q109943342",
+        "cuisine": "donut",
+        "diet:vegan": "yes",
```
LaoshuBaby, line 1222: In #5727 , `diet:vegan=yes` should be consider whether can be merged into main branch.

You can have a look at that PR

LaoshuBaby (changes requested):

data/brands/amenity/fast_food.json
codeinabox: I added this tag as they state on their website that they produce a selection of vegan doughnuts, which is a key part of their brand https://www.crosstown.co.uk/vegan-doughnuts/

bhousel: I looked at the website and I agree, the vegan offerings are a key part of their brand and so the tag is ok 👍

bhousel (approved):

PR #5802, opened 2021-12-06
add wikidata for Евро Кебаб

afanasovich:


data/brands/amenity/fast_food.json
```diff
@@ -6380,6 +6380,7 @@
         "brand": "Евро Кебаб",
         "brand:en": "Еuro Кеbаb",
         "brand:ru": "Евро Кебаб",
+        "brand:wikidata": "109943635",
```
bhousel, line 6383: Don't forget the `Q`  😄 
 
```suggestion
        "brand:wikidata": "Q109943635",
```

bhousel (changes requested):

data/brands/amenity/fast_food.json
afanasovich: > 

I'm not a real developer, I'm just pretending)

bhousel: no worries, computers are all pretend anyway

PR #5803, opened 2021-12-06
Added Everyone Active to sports_centre

rskedgell:
Everyone Active (Sports and Leisure Management Ltd) manage sports and leisure facilities on behalf of UK local authorities.


UKChris-osm:
If there isn't a brand:wikidata, would this be better of under the operator tree?

Or does "_Everything Active_" branding appear on the leisure centres? If it does, I think adding brand:wikidata would be the best option

PR #5806, opened 2021-12-06
Add matchTags for Хеликс

afanasovich:
Helix's often includes both a laboratory and a doctor's office.


arch0345:
Is Хеликс is currently mapped separately as `healthcare/laboratory` and `amenity/doctors`, similarly to how Costco's services are mapped separately (food court, pharmacy, tire center, etc.)? If this is the case, I would add this to `amenity/doctors` instead.

bhousel:
Merging as-is.. can change later if needed.

PR #5807, opened 2021-12-06
add «Petshop.ru»

afanasovich:


UKChris-osm:
Thank you @afanasovich - please take a look at #5802 so we can merge that as well :)

PR #5814, opened 2021-12-08
Улыбка радуги — its shop=chemist, not shop=cosmetic

afanasovich:
«Улыбка радуги» — it's shop=chemist, not shop=cosmetic.
I move «Улыбка радуги» from cosmetics.json to chemist.json


arch0345:
Thanks for fixing this!

PR #5816, opened 2021-12-08
Match against Energie Fitness

codeinabox:
I noticed many of the existing entries for énergie Fitness use the name Energie Fitness eg https://www.openstreetmap.org/node/6484557587

### DIFF
--- a/data/brands/leisure/fitness_centre.json
+++ b/data/brands/leisure/fitness_centre.json
@@ -201,6 +201,7 @@
       "locationSet": {
         "include": ["bh", "gb", "ie"]
       },
+      "matchNames": ["energie fitness"],
       "tags": {
         "brand": "énergie Fitness",
         "brand:wikidata": "Q109855553",


UKChris-osm:
Thanks @codeinabox, but I don't think this matchName is needed, as I have opened up the iD editor and the Name Suggestion Index is already suggesting the brand data update.

Looking at the énergie fitness website, it looks like that particular branch no longer exists.

UKChris-osm:
Just to confirm that "_npm run build_" has removed this matchName via [ecb7069](https://github.com/osmlab/name-suggestion-index/commit/ecb70698cce4860754db55c2ff592ec9cd4c174a) as the matcher seems to factor in variations with accents in names.

codeinabox:
@UKChris-osm thank you for spotting this.

@arch0345 if this change is superfluous I am happy for it to be reverted

arch0345:
No need to revert this - it's already been automatically removed, just keep in mind that the matcher accounts for accents in the future.

UKChris-osm:
> @UKChris-osm thank you for spotting this.

It's an easy thing to miss and thank you for the original submission, no revert is needed as the matcher automatically removes tags it feels aren't needed anymore.

PR #5817, opened 2021-12-08
Change operator value for Česká pošta

jmacura:
Change operator="Česká pošta" to operator="Česká pošta, s.p."
per discussions at talk-cz list (https://openstreetmap.cz/talkcz/c46).
Also discussed at OSM US Slack (https://osmus.slack.com/archives/CBK3JLUJU/p1628112745023200).

### DIFF
--- a/data/operators/amenity/post_office.json
+++ b/data/operators/amenity/post_office.json
@@ -181,10 +181,10 @@
       "displayName": "Česká pošta",
       "id": "ceskaposta-aa9270",
       "locationSet": {"include": ["cz"]},
-      "matchNames": ["česká pošta sp"],
+      "matchNames": ["česká pošta"],
       "tags": {
         "amenity": "post_office",
-        "operator": "Česká pošta",
+        "operator": "Česká pošta, s.p.",
         "operator:wikidata": "Q341090",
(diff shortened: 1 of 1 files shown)


bhousel (changes requested):
Thanks, you'll also want to modify the "matchNames" property to match the unwanted "česká pošta" variant.

(to elaborate: "matchNames" is where we stick alternate spellings that we want to match but not use as tags).

jmacura:
OK, I can't understand the underlaying behaviour now, but I have made the change you request @bhousel .

bhousel (approved):

bhousel:
Thanks @jmacura and welcome to the NSI project! 👋

jmacura:
Thank you very much for your help and swift responses @bhousel !
Are the changes immediately reflected in the iD editor or is the NSI updated in a batch through some release cycle?

bhousel:
> Are the changes immediately reflected in the iD editor or is the NSI updated in a batch through some release cycle?

We need to run a script to sync our project up with wikidata, then another script to generate the new data files, but the whole process doesn't take very long.  We typically publish releases at least once or twice a week, then they are available in iD and RapiD immediately thereafter.

PR #5819, opened 2021-12-09
Simmons bars

codeinabox:


arch0345:
Thanks @codeinabox! Building this removed `amenity/pub` from matchTags as it's already apart of a matchGroup with `amenity/bar`. You can see all the matchGroups here:
https://github.com/osmlab/name-suggestion-index/blob/main/dist/matchGroups.json

Issue #5718, opened 2021-11-22
Fix "Цех 85" to "ЦЕХ85" 


afanasovich:
now:
```
amenity=cafe
brand=Цех 85
cuisine=coffee_shop
name=Цех 85
takeaway=yes

```
properly:
```
amenity=cafe
shop=bakery
brand=ЦЕХ85
cuisine=coffee_shop
name=ЦЕХ85
takeaway=yes
contact:email=info@tseh85.ru
contact:instagram=https://instagram.com/tseh85
contact:phone=+7 812 7757785
contact:vk=https://vk.com/tseh85
contact:website=https://tseh85.ru
name=ЦЕХ85
payment:cash=yes
payment:credit_cards=yes
payment:debit_cards=yes
```
thx

UKChris-osm:
Hi @afanasovich - just to let you know we tend to stick with just the "_name_" tag, as well as "_brand_" or "_operator_", depending on the category. We don't include "_contact_" or "_payment_" tags.

I think "_amenity=cafe_" and "_shop=bakery_" would also clash, so it would have to be either one or the other.

afanasovich:
Hi @UKChris-osm! 

thanks for the explanation, i get it. 

I create Item https://www.wikidata.org/wiki/Q109682756. Сould you add the wikidata to this  Item?

afanasovich:
👋!  Thanks for the clarification.  I don’t understand very well how github works and don’t know English very well)) But I understood something.

I have created the item  https://www.wikidata.org/wiki/Q109682756. Сould you add the wikidata to this Item?

amenity=cafe
brand=ЦЕХ85
brand:wikidata = [Q109682756](https://www.wikidata.org/wiki/Q109682756) 
cuisine=coffee_shop
name=Цех 85
takeaway=yes

arch0345:
Thanks for adding the wikidata page! I'll add this as soon as I can.

If you create a PR I could merge it right away: just go to `data/brands/amenity/cafe.json` and edit (pencil icon). You should be able to find the entry by using Ctrl+f.

PR #5822, opened 2021-12-10
Added french details on the Habitat retail chain

le-jun:
Created a french specific version for the Habitat retail chain, including french exclusive tags. Removed France from the locationSet of "Habitat (non-Uk)" and corrected the wikidata now aiming to the retail rather than the group.


arch0345:
Thanks for adding this! Note we dont use `contact`, `ref`, or `website` tags in the index as this can be added to the linked Wikidata: see #5718

le-jun:
That makes sense. Thanks for reviewing it!

PR #5828, opened 2021-12-10
add Север-Метрополь

afanasovich:


arch0345:
Just as in #5718, I would stick with either `shop/confectionary` or `amenity/cafe` for this

PR #5829, opened 2021-12-10
Add Imperial Porcelain Factory, Saint Petersburg

afanasovich:


afanasovich:
Anybody, add please MatchName «Императорский фарфор»

bhousel:
I really don't think you need to add all the localized `name` and `brand` tags for a brand that will only exist in Russia (unless your community really wants it tagged this way for some reason).

bhousel:
Going to just merge this as-is.  You can change it later if you want to.

PR #5833, opened 2021-12-11
Colicci cafe brand

codeinabox:


UKChris-osm:
If these are just in London, then "_gb-lon.geojson_" might be worth using instead of gb.

codeinabox:
> If these are just in London, then "_gb-lon.geojson_" might be worth using instead of gb.

Thank you for the feedback, all the locations are in London so have updated accordingly

UKChris-osm:
Thank you @codeinabox 👍

PR #5837, opened 2021-12-12
Yard Sale Pizza

codeinabox:
Add preset for Yard Sale Pizza that has expanded across London

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -6230,7 +6230,6 @@
         "brand": "Yard Sale Pizza",
         "brand:wikidata": "Q110066789",
         "cuisine": "pizza",
-        "delivery": "yes",
         "diet:vegan": "yes",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Thanks @codeinabox  - I don't think the matchName would be needed, and I would probably remove the "_diet_" tag as I don't think #4855 has been resolved yet.

codeinabox:
> Thanks @codeinabox  - I don't think the matchName would be needed, and I would probably remove the "_diet_" tag as I don't think #4855 has been resolved yet, and I would probably remove the delivery tag as well, as I don't think it could be guaranteed that delivery would be available if any new outlets are opened in future.

I added the diet tag they have a vegan section of the menu that features quite prominently. All their current stores have delivery but I see your point and am happy to remove that tag.

UKChris-osm:
Thanks @codeinabox 👍

PR #5838, opened 2021-12-12
Added Bricopro: home improvement stores’ frenich chain

le-jun:


data/brands/shop/doityourself.json
```diff
@@ -246,6 +246,16 @@
         "shop": "doityourself"
       }
     },
+    {
+      "displayName": "Bricopro",
+      "locationSet": {"include": ["fr"]},
```
arch0345, line 251:
```suggestion
      "locationSet": {"include": ["fx"]},
```
Their website only shows locations in Metropolitan France (`fr` includes all French territories)

arch0345 (changes requested):

le-jun (commented):
There seems to be some locations outside of metropolian France too. The schematic map on their website shows some markers in Guadeloupe.

arch0345 (approved):
You're right, it looks like they do have locations in Guadeloupe

PR #5841, opened 2021-12-13
Correct network name Mistral

Patchi-Osm:
The name for the network Mistral is "Mistral" and not "Network Mistral".

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -7883,7 +7883,7 @@
       "displayName": "Mistral",
       "id": "reseaumistral-add5eb",
       "locationSet": {"include": ["fx"]},
-      "matchNames": ["mistral"],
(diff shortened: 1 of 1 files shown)


data/transit/route/bus.json
```diff
@@ -7879,6 +7879,18 @@
         "route": "bus"
       }
     },
+    {
+      "displayName": "Mistral",
+      "id": "reseaumistral-add5eb",
+      "locationSet": {"include": ["fx"]},
+      "matchNames": ["mistral"],
```
LaoshuBaby, line 7886: If “Réseau Mistral” is a mistypo, you can add it to matchNames so that other can find it and rename.

```suggestion
      "matchNames": ["mistral","Réseau Mistral"],
```

LaoshuBaby (approved):

PR #5852, opened 2021-12-13
Rename "Фиалка Аптека" → "Фиалка"

afanasovich:
We conferred in the Russian-language chat https://t.me/ruosm, despite the fact that sometimes the word "Аптека" ("Pharmacy") appears in the logo, it would be correct not to use it.


bhousel:
Sounds good - you can also add `"matchNames": ["Фиалка Аптека"]` so that iD will steer users towards the preferred name.

PR #5854, opened 2021-12-13
Green Flag Award

codeinabox:
Many parks through out the world have been given a Green Flag Award, which is installed in the park eg https://www.openstreetmap.org/node/8909783817

https://www.greenflagaward.org/about-the-award/


bhousel:
Sounds good to me 👍 
I guess we just need a `flag:wikidata` to point it to?

codeinabox:
> Sounds good to me 👍 I guess we just need a `flag:wikidata` to point it to?

I think so, not sure if I should create a new Wikidata entry or link to the existing one as the award is a flag

bhousel:
> I think so, not sure if I should create a new Wikidata entry or link to the existing one as the award is a flag

Probably both things - Are these actual flags being flown on flagpoles?

What we usually do is have a `flag:wikidata` pointing to a Wikidata page about the flag design, and a `subject:wikidata` pointing to a Wikidata page about what the flag is for.

So for example:
A flagpole for the US Space Force:
https://nsi.guide/index.html?t=flags&k=man_made&v=flagpole#unitedstatesspaceforce-85bb3f

```
flag:name=United States Space Force
flag:type=military
flag:wikidata=Q94652289
man_made=flagpole
subject=United States Space Force
subject:wikidata=Q55088961
```

The `flag:wikidata` [Q94652289](https://www.wikidata.org/wiki/Q94652289) points to a page about the flag.  It has an `image` claim with an .svg of the current design.
The `subject:wikidata` [Q55088961](https://www.wikidata.org/wiki/Q55088961) points to a page about the US Space Force organization.

LaoshuBaby (approved):

PR #5857, opened 2021-12-14
Patty & Bun burger restaurants

codeinabox:
Patty & Bun burger restaurants, which have a dedicated plant based menu https://www.pattyandbun.co.uk/plant-based-vegan-gluten-free-vegetarian


arch0345:
Thanks @codeinabox!

PR #5859, opened 2021-12-15
Updating hotel.json

arnowrld:
I hope the changes meet the standards of the nsi.

### DIFF
--- a/data/brands/tourism/hotel.json
+++ b/data/brands/tourism/hotel.json
@@ -453,6 +453,17 @@
       }
     },
     {
+      "displayName": "Club Med",
+      "id": "clubmed-j94nk8",
+      "locationSet": {["001"]},
+      "tags": {
+        "brand": "Club Med",
+        "brand:wikidata": "Q1076645",
+        "name": "Club Med",
+        "tourism": "hotel"
+      }
+    },
+    {
       "displayName": "Comfort Inn",
       "id": "comfortinn-134d5e",
       "locationSet": {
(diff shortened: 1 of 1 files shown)


data/brands/tourism/hotel.json
```diff
@@ -979,19 +998,6 @@
         "tourism": "hotel"
       }
     },
-    {
```
tas50, line 982: This is a different brand that needs to stay

data/brands/tourism/hotel.json
```diff
@@ -1078,18 +1085,6 @@
...
-      "displayName": "Holiday Inn Express & Suites",
```
tas50, line 1082: Same with this one. These are different brands and we want to be able to tag them in OSM

arnowrld: Based on the research that I've done, it's proven that they're the same brand.

tas50: They use different signage and they mean different things. If someone wants a suite at their hotel they want to make sure they find the hotel that has suites.

arnowrld: I made sure to keep the different signage with my changes, especially since even if there is different signage, they are under the same chain & brand.

arnowrld:
> I hope the changes meet the standards of the nsi.

Is there a way to implement the changes so it can adequate for the nsi?

data/brands/tourism/hotel.json
```diff
@@ -452,6 +452,17 @@
         "tourism": "hotel"
       }
     },
+    {
+      "displayName": "Club Med",
+      "id": "clubmed-j94nk8",
+      "locationSet": {["001"]},
```
UKChris-osm, line 458: The locationSet seems to be missing "_include_" here.

PR #5865, opened 2021-12-15
French home appliance chain PRO&Cie

le-jun:
One store location in Martinique, so used "fr" locationSet.

### DIFF
--- a/data/brands/shop/electronics.json
+++ b/data/brands/shop/electronics.json
@@ -769,6 +769,19 @@
       }
     },
     {
+      "displayName": "Power",
+      "locationSet": {
+        "include": ["fr"]
+      },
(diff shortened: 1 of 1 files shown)


data/brands/shop/electronics.json
```diff
@@ -757,9 +757,7 @@
     {
       "displayName": "Power",
       "id": "power-9765c4",
-      "locationSet": {
-        "include": ["dk", "fi", "no", "se"]
-      },
+      "locationSet": {"include": ["dk", "fi", "no", "se"]},
```
LaoshuBaby, line 760: Notice:  You don't need to let this in 1 line, program will auto split if it is too long.

Issue #5858, opened 2021-12-14
Add Christmas Tree Shops


arnowrld:
Christmas Tree Shops is a chain of specialty shops.

[Q5111452](https://www.wikidata.org/wiki/Q5111452)

PR #5866, opened 2021-12-15
Added Christmas Tree Shops

le-jun:
https://github.com/osmlab/name-suggestion-index/issues/5858

### DIFF
--- a/data/brands/shop/variety_store.json
+++ b/data/brands/shop/variety_store.json
@@ -139,6 +139,17 @@
       }
     },
     {
+      "displayName": "Christmas Tree Shops",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "brand": "Christmas Tree Shops",
+        "brand:wikidata": "Q5111452",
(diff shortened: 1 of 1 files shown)


data/brands/shop/variety_store.json
```diff
@@ -138,6 +138,17 @@
         "shop": "variety_store"
       }
     },
+    {
+      "displayName": "Christmas Tree Shops",
+      "locationSet": {"include": ["us"]},
```
arch0345, line 143: [Overpass turbo](https://overpass-turbo.eu/?w=%22name%22%3D%22Christmas+Tree+Shops%22+global&R) shows that these also are tagged differently, so we also want to match locations tagged as these
```suggestion
      "locationSet": {"include": ["us"]},
      "matchTags": ["shop/department_store", "shop/gift", "shop/houseware", "shop/interior_decoration"],
```

arch0345 (changes requested):

Issue #5839, opened 2021-12-13
Add Willamette Dental Group


arch0345:
Dental office chain with around 50 locations in the PNW (Oregon, Washington, Idaho)
https://willamettedental.com/

le-jun:
Exactly 47 according to their [locations map](https://locations.willamettedental.com/) (50 on their [media sheet](https://willamettedental.com/help/general-information/media-sheet/)).

PR #5867, opened 2021-12-15
Add Willamette Dental Group

le-jun:
Issue https://github.com/osmlab/name-suggestion-index/issues/5839.

Not sure about the properties, especially for the "exclude"/"generic" part.


arch0345 (changes requested):
Thanks for adding this! This should really go in the existing `amenity/dentist` category

le-jun:
Wasn’t aware of it, change done!

arch0345 (approved):
Looks good 👍

PR #5869, opened 2021-12-15
Adding Concentra, MedExpress & Patient First

arnowrld:
Multiple urgent care chains.

### DIFF
--- a/data/operators/amenity/clinic.json
+++ b/data/operators/amenity/clinic.json
@@ -182,6 +182,17 @@
       }
     },
     {
+      "displayName": "Concentra",
+      "id": "concentra-02op54",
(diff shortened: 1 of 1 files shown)


arch0345:
I suggest adding urgent_care=yes to these: ZoomCare, another urgent care chain that's in the index has this tag

bhousel:
I'm just hearing about the `urgent_care=yes/no` tag for the first time..
https://wiki.openstreetmap.org/wiki/Proposed_features/Urgent_care
https://taginfo.openstreetmap.org/keys/urgent_care

I am ok with this, but it seems like the tag never really got accepted, so maybe we should ask around some other channels before we start adding it.

Issue #5758, opened 2021-11-29
Forever 21 preset missing


bhousel:
We have an entry for it
https://nsi.guide/index.html?t=brands&k=shop&v=clothes&tt=forever#forever21-3937bd

But I don't see it `dist/presets/nsi-id-presets.json`, so there's no preset for it in iD:
<img width="682" alt="Screen Shot 2021-11-29 at 5 34 01 PM" src="https://user-images.githubusercontent.com/38784/143954463-e858d5ca-ff03-4adf-a738-8ebcfe91ff3c.png">

I don't see any obvious reason why this would be missed, so I'm going to look into why it's missing.

maxerickson:
Marked as defunct in Wikidata (parallel edit to the one that marked Mobil defunct, as discussed on OSM US Slack).

bhousel:
Oh good call @maxerickson 👍 

It makes sense now, but I find the behavior kind of surprising.  Maybe we should generate these as non-addable presets?

bhousel:
> Maybe we should generate these as non-addable presets?

☝️  I did this in a0a3505ba

PR #5874, opened 2021-12-15
Update clothes.json

arnowrld:
Please note that the brands I've gotten rid of have become e-commerce & closed their stores. Also, I'm currently moving Burlington to department stores.


arch0345:
It looks like some of the defunct brands removed are still in OSM, so to prevent these from being added again with the build script I would add this to the exclude named list as "^(brand name)$"

bhousel:
> It looks like some of the defunct brands removed are still in OSM, so to prevent these from being added again with the build script I would add this to the exclude named list as "^(brand name)$" found at the top of the file

I think after #5758 that we should leave defunct brands in the index for a little while, as it can be kind of surprising if you are expecting a preset for one and it doesn't show up in iD.

tas50:
I'd also recommend creating a Maproullete challenge to cleanup any dead brands. Otherwise, they'll stick around for the next decade.

UKChris-osm:
> I think after #5758 that we should leave defunct brands in the index for a little while, as it can be kind of surprising if you are expecting a preset for one and it doesn't show up in iD.

Could we add something on this end that highlights a brand as no longer having physical locations within our core data, so we could also highlight it on the web site - and in turn on the iD side which highlights any matches as "_this brand no longer has outlets / branches_"?

no-longer-trade

bhousel:
I ended up restoring  Avenue, Catherines, Dress Barn, and Justice because `npm run build` put them back, because they still exist frequently in OSM.

PR #5875, opened 2021-12-15
Update department_store.json

arnowrld:


data/brands/shop/department_store.json
```diff
@@ -218,6 +218,18 @@
...
+    {
+      "displayName": "Burlington Coat Factory",
+      "id": "burlingtoncoatfactory-0a21d9",
+      "locationSet": {"include": ["pr", "us"]},
+      "tags": {
+        "brand": "Burlington Coat Factory",
+        "brand:wikidata": "Q4999220",
+        "brand:wikipedia": "en:Burlington (department store)",
+        "name": "Burlington Coat Factory",
+        "shop": "clothes"
```
arch0345, line 230: If you want to tag this as `shop=clothes` this should be moved to the existing `shop/clothes` category
```suggestion
        "shop": "department_store"
```

data/brands/shop/department_store.json
```diff
@@ -1318,6 +1318,18 @@
...
+    {
+      "displayName": "Winners",
+      "id": "winners-2047be",
+      "locationSet": {"include": ["ca", "es"]},
+      "tags": {
+        "brand": "Winners",
+        "brand:wikidata": "Q845257",
+        "brand:wikipedia": "en:Winners",
+        "name": "Winners",
+        "shop": "clothes"
```
arch0345, line 1330:
```suggestion
        "shop": "department_store"
```

arch0345 (changes requested):

data/brands/shop/department_store.json
tas50: Another note on Burlington. They're just using the name "Burlington" now. We should set "Burlington Coat Factory" as an alt_name.

tas50:
I updated all the Lord & Taylor stores in the US to disused except the ones in NYC. Some NYC folks have strong opinions in OSM and I want to let them survey and update those locations on their own.

PR #5856, opened 2021-12-14
Aboriginal & Torres Strait Islander flags

codeinabox:
Australian Aboriginal & Torres Strait Islander flags

https://wiki.openstreetmap.org/wiki/Australian_Tagging_Guidelines#Australian_Flags

PR #5885, opened 2021-12-16
Remove redundant word flag from Australian indigenous flags

codeinabox:
A small fix for the PR #5856 removing redundant word flag


bhousel:
Thanks @codeinabox 👍

PR #5887, opened 2021-12-16
Adding APlus?

arnowrld:
This is the name of Sunoco's convenience store chain & franchise.

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -336,6 +336,17 @@
       }
     },
     {
+      "displayName": "APlus (Sunoco)",
+      "id": "aplus-02op54",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "brand": "APlus (Sunoco)",
+        "brand:wikidata": "Q1423218",
+        "name": "APlus (Sunoco",
+        "shop": "convenience"
+      }
+    },
+    {
       "displayName": "Aral",
       "id": "aral-1fe310",
       "locationSet": {"include": ["de", "lu"]},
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Hi @arnowrld - it looks like APlus has it's own [Wikidata](https://www.wikidata.org/wiki/Q4646934) / [Wikipedia](https://en.wikipedia.org/wiki/A-Plus), so I think using these would be more suitable for this PR.

data/brands/shop/convenience.json
```diff
@@ -335,6 +335,17 @@
         "shop": "convenience"
       }
     },
+    {
+      "displayName": "APlus (Sunoco)",
+      "id": "aplus-02op54",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "brand": "APlus (Sunoco)",
+        "brand:wikidata": "Q1423218",
```
LaoshuBaby, line 344:
```suggestion
        "brand:wikidata": "Q4646934",
        "brand:wikipedia":"en:A-Plus",
```

data/brands/shop/convenience.json
```diff
@@ -335,6 +335,17 @@
...
+        "name": "APlus (Sunoco",
```
LaoshuBaby, line 345:
```suggestion
        "name": "APlus (Sunoco)",
```

bhousel:
Can we just use `name=APlus` and `brand=APlus` for these?
https://www.sunocolp.com/retail-solutions/aplus-convenience-stores

PR #5888, opened 2021-12-16
Clapton Craft

codeinabox:
A London chain specialising in craft beer

### DIFF
--- a/data/brands/shop/alcohol.json
+++ b/data/brands/shop/alcohol.json
@@ -145,6 +145,19 @@
       }
     },
     {
+      "displayName": "Clapton Craft",
+      "locationSet": {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
@bhousel should _"drink:beer": "yes"_ be removed from this entry? Do Clapton Craft only sell beer?

bhousel:
> @bhousel should _"drink:beer": "yes"_ be removed from this entry? Do Clapton Craft only sell beer?

First time I've seen the `drink:` prefix.. Here's what the wiki says about it:
https://wiki.openstreetmap.org/wiki/Key:drink

It sounds like a tag that hasn't really caught on, and its use for retail shops is debatable?  It might only be for places that serve drinks?  I'd leave it off for now.

PR #5892, opened 2021-12-17
Australian state and territory flags

codeinabox:


arch0345:
I'm not sure if country tags are appropriate for regional flags. The [wiki page](https://wiki.openstreetmap.org/wiki/Key%3Acountry) for the country key says that it "indicates the country whose flag is flown from the flagpole" and the US state flags in the NSI don't include a country tag

codeinabox:
> I'm not sure if country tags are appropriate for regional flags. The [wiki page](https://wiki.openstreetmap.org/wiki/Key%3Acountry) for the country key says that it "indicates the country whose flag is flown from the flagpole" and the US state flags in the NSI don't include a country tag

I see what you mean, will fix accordingly

Issue #5883, opened 2021-12-16
Event Cinemas Australia


aaronstar:
https://nsi.guide/index.html?t=brands&k=amenity&v=cinema&tt=event

Correct from Event to Event Cinemas, delete redundant official_name

amenity=cinema
brand=Event Cinemas
brand:wikidata=Q5416698
brand:wikipedia=en:Event Cinemas
name=Event Cinemas
X<official_name=Event Cinemas>X -- DELETE

Artumira96:
Hello, I created a PR for this issue, can you review it please? :)

PR #5894, opened 2021-12-17
Correct to Event Cinemas

Artumira96:
Correct from "Event" to "Event Cinemas" in the 'brand' and 'name' atributes, and removed the 'official_name' atribute 
Solves issue #5883


arch0345:
Thanks and welcome to the NSI community!

PR #5895, opened 2021-12-17
remove Louis from clothes (doubled entry of shop=motorcycle)

SafetyIng:
shop=clothes don't fit to Louis, Louis sells also parts and tools for motorcycles. 
It doubles a entry for shop=motorcycle


arch0345:
Thanks for catching this! Could you add matchTags for the entry in shop/motorcycle so that existing Louis locations tagged as shop=clothes are matched? I also recommend adding some of the `motorcycle:` tags found on the [wiki](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dmotorcycle) to show that they sell clothes and parts

UKChris-osm:
> Could you add matchTags for the entry in shop/motorcycle so that existing Louis locations tagged as shop=clothes are matched?

Good call!

> I also recommend adding some of the `motorcycle:` tags found on the [wiki](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dmotorcycle) to show that they sell clothes and parts

Looking at the link above, it seems to indicate that the "_shop=motorcycle_" tag already implies clothing may be sold there, and further down the page it reads "_there is no clear consensus on which group of service tags to use_" - so I'm wondering if it would be best not to include this?

SafetyIng:
Thanks @arch0345 and @UKChris-osm  - I updated the pull request

SafetyIng:
I also added a change for "Polo Motorrad", wich had the same error

arch0345:
Thanks @SafetyIng and welcome to the NSI community!

Issue #5893, opened 2021-12-17
Is the Name_Suggestion_Index (NSI) branding recommended?


sebhan2:
Hello,
I hope this is the right place for my question otherwise please let me know about alternative places to ask this. I also posted my question here but haven't gotten an answer yet.
 https://help.openstreetmap.org/questions/82860/is-the-name_suggestion_index-nsi-branding-recommended

So far I assumed the Name_Suggestion_Index (NSI) feature used in the Id-editor, that allows me to use the branding for particular locations, is a recommended way to give canonical features to particular objects in OSM. I changed some package vending machines of the company DHL to use the NSI brand "DHL Packstation" in germany. This includes besides adding wikidata, also a "name" and an "alt_name".

Now I hear from a user that the "name" and "alt_name" in the branding are key-value-pairs that are not on the OSM wiki page and therefore should not be there. Here is a link: https://wiki.openstreetmap.org/wiki/Tag:vending%3Dparcel_pickup

As far as I understand the concern, the vending machine would be rendered differently on a map depending on whether the key "name" is given. I have not seen OSM rendering vending machines explicitly but due to my lack of experience I assume he may have a valid point.

Here are my questions: 
Is the NSI branding a recommended way to choose key-value-pairs?
Which platform should be updated in my case, the NSI branding suggestion or the OSM wiki?

bhousel:
It sounds like @kjonosm took care of this in #5899, thanks!

Copying [my response](https://help.openstreetmap.org/questions/82860/is-the-name_suggestion_index-nsi-branding-recommended/82869) here:

---

Hello, NSI maintainer here 👋

>Is the NSI branding a recommended way to choose key-value-pairs?

Yes, we (contributors to the NSI project) believe that it represents the best way to tag things today. We know that tagging can change, and people will change the wiki, and brands will change their names, and so we tried to build it in a way that any part of it can be easily changed if needed.

> Which platform should be updated in my case, the NSI branding suggestion or the osm wiki?

[...]

PR #5899, opened 2021-12-18
Update vending_machine.json

kjonosm:
Folks in Germany tend to favor removing the `name` tag from vending machines only keeping `brand` and `operator` tags, see discussion in [German forum](https://forum.openstreetmap.org/viewtopic.php?pid=849708).


SafetyIng:
> But maybe I am wrong - is Packstation a brand?

Jeay - I would mention "Packstation" as a brand. Real it is a service with the service-name "Packstation" so more a product under this branding.

> Can you use a Packstation service in a form that's an alternative to a locker (i.e. a collection, a service desk, etc)?

You can pick up parcels at a Packstation or send them directly. 

> Is Packstation a translation of "_locker_"?

I would say: Yes. In the german dictionary "Packstation" is described as: "A locker-style facility where postal packages can be picked up or dropped off for mailing." So Packstation is also used for "lockers" of ofther operators, e.g. amazon: https://www.come-on.de/volmetal/meinerzhagen/packstation-bft-tankstelle-meinerzhagen-kierspe-90901307.html

bhousel:
Thanks @kjonosm for fixing, and @sebhan2 for raising on #5893 and checking with the community 👍

PR #5900, opened 2021-12-18
change teilAuto

SafetyIng:
teilAuto is not a operator/company it is a branch, operated by the Mobility Center GmbH


sb12:
Please note that there are several operators that use "teilAuto" as a brand name. The wikipedia articles seems to be about the bigger one by Mobility Center GmbH which operates in Central Germany. There are also some smaller operators in Baden-Württemberg in Southern Germany like e.g. http://www.teilauto-hall.de, https://www.teilauto-biberach.de or https://www.teilauto-neckar-alb.de .
teilAuto Neckar-Alb is already in NSI, the others seem to be missing. (They seem to be very local with only a few stations)

Looking at http://overpass-turbo.eu/s/1ehw , iD seems to encourage people to tag everything as the bigger teilAuto brand (You can clearly see that in the history, e.g. https://www.openstreetmap.org/node/8837036998 or https://www.openstreetmap.org/node/1835256298). Best solution is probably to limit the Mobility Center GmbH teilAuto to Central Germany and manually fix all the auto-fixed ones in Baden Württemberg afterwards.

SafetyIng:
Hey @sb12 - I'Ve added the shapes of the states, where teilAuto operates and then updated the entry of teilAuto.

bhousel:
Nice!  I notice that the German state geojsons are a bit more detailed than what we usually include here..

@michaelblyons  - do you have time to create state geojsons for Germany similar to what you did on #4784 / #5449 ? 🙏

michaelblyons:
Haha. That wasn't me.

bhousel:
> Haha. That wasn't me.

Oh sorry!!! It was @jdcarls2 who made the state geojsons!

UKChris-osm:
@SafetyIng those state geojson files are very detailed, so well done 👍 

However, because detailed geojson files can cause a heavy amount of processing when the locationSets are being built, we tend to go for a looser area when it comes to states or regions, with fewer points. Are you able to redraw them in a more loose way?

I would recommend keeping the one's you have drawn on file at home though, as I expect there will be a demand for them somewhere.

SafetyIng:
> those state geojson files are very detailed, so well done 👍
> 
> However, because detailed geojson files can cause a heavy amount of processing when the locationSets are being built, we tend to go for a looser area when it comes to states or regions, with fewer points. Are you able to redraw them in a more loose way?

Thanks, thanks ;) I just use the boundaries as geojson and simplified them to around 5% with mapshaper.org

> Are you able to redraw them in a more loose way?

I would handle it with #5915 for all german states in the next days and then update the teilAuto Element ;)

jdcarls2:
@SafetyIng If you're interested, I think the procedure I followed ended up in the wiki. You'd have to use QGIS, but it allowed me to make topologically correct generalized states. I don't have quite as much time on my hands these days, but I'd be willing to take a crack at doing it myself, too, if you would rather not tackle it in QGIS.

Mapshaper is pretty good as a starting point, but it doesn't always maintain the topology of your shapes, leading to possible gaps or overlaps.

SafetyIng:
Hey @jdcarls2, thanks for the offer, but I'll try myself first. Even though I still have to learn the ropes a bit, it's worked out pretty well so far.

PR #5901, opened 2021-12-18
Exclude Stadtmobil from car sharing operators

sb12:
Stadtmobil is already in car sharing brands, but is apparently still added by some bot.
I hope it works like that.


bhousel:
I think you probably do want the matchNames, but not the generic exclude.. 
Generic excludes are used to warn users not to enter that value in the name tag because it's not actually the name of the business.  (for example `name=Cafe`)

Issue #5882, opened 2021-12-16
Rebel Sport Australia


aaronstar:
https://nsi.guide/index.html?t=brands&k=shop&v=sports&tt=rebel

Correct from Rebel to Rebel Sport

brand=Rebel Sport
brand:wikidata=Q18636397
brand:wikipedia=en:Rebel (company)
name=Rebel Sport
shop=sports

UKChris-osm:
Rebel seems correct to me? It looks like they dropped the "_sport_" from their name?

Mortein:
Please ensure this is for the correct one, as there are both Australian and New Zealand chains.

Australia is "Rebel" and is Q18636397
New Zealand is "Rebel Sport" and is Q110190372

aaronstar:
Yep sorry, you guys are right they did drop the "Sport" from the Australian chain

PR #5902, opened 2021-12-19
Add Rebel Sport in New Zealand

Mortein:
#5882 was to change the name of Rebel to Rebel Sport.
There's an Australian chain (Rebel) and New Zealand chain (Rebel Sport).

Added New Zealand Rebel Sport.
Australian chain should only be offered in Australia.


arch0345:
Thanks @Mortein and welcome to the NSI community!

PR #5903, opened 2021-12-19
Mitre 10 and Mitre 10 MEGA, New Zealand

Mortein:
"Mitre 10" is a chain of DIY stores in New Zealand, and "Mitre 10 MEGA" is the big-box version.
There are around 95 in total, of which 46 are MEGA.


arch0345:
Thanks @Mortein!

PR #5911, opened 2021-12-20
Fix Czech wiki prefix, unify South Moravian networks and operators

kudlav:
- `cz:` wikipedia prefix is invalid, it should be `cs:`
- `cz:IDS JMK` network in `bus.json`, but `trolleybus.json` and `tram.json` were not using this prefix -> prefix removed
- `Dopravní podnik města Brna` in `trolleybus.json` and `tram.json` changed to `DPMB`
  - we have [706 objects](https://overpass-turbo.eu/s/1ejm) and only 36 objects of them was with fullname `Dopravní podnik města Brna`
  - those 36 object were edited by iD editor, and as those rules are not applied to the bus routes, renaming to the fullname would causing inconsistend name convention

### DIFF
--- a/data/operators/amenity/post_office.json
+++ b/data/operators/amenity/post_office.json
@@ -186,7 +186,7 @@
         "amenity": "post_office",
         "operator": "Česká pošta, s.p.",
         "operator:wikidata": "Q341090",
-        "operator:wikipedia": "cz:Česká pošta"
+        "operator:wikipedia": "cs:Česká pošta"
       }
     },
     {
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -3253,11 +3253,11 @@
       "tags": {"network": "CWL", "route": "bus"}
     },
     {
-      "displayName": "cz:IDS JMK",
+      "displayName": "IDS JMK",
       "id": "czidsjmk-1bafda",
       "locationSet": {"include": ["cz"]},
       "tags": {
-        "network": "cz:IDS JMK",
+        "network": "IDS JMK",
         "route": "bus"
       }
     },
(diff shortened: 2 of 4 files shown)


LaoshuBaby:
Yes, Czech Wikipedia use `cs` prefix ( https://en.wikipedia.org/wiki/Czech_Wikipedia )

But NSI already have a `IDS JMK` in bus.json

This is yours:

```json
    {
      "displayName": "IDS JMK",
      "id": "czidsjmk-1bafda",
      "locationSet": {"include": ["cz"]},
      "tags": {
        "network": "IDS JMK",
        "route": "bus"
      }
    },
```

This is anothe `IDS JMK` that already in NSI:

```json
    {
      "displayName": "IDS JMK",
      "id": "idsjmk-151bbe",
      "locationSet": {
        "include": ["at", "cz", "sk"]
      },
      "tags": {
        "network": "IDS JMK",
        "network:wikidata": "Q12020731",
        "network:wikipedia": "cs:Integrovaný dopravní systém Jihomoravského kraje",
        "route": "bus"
      }
    },
```
This system don't allow 2 entry have same displayName, so after you remove `cz:` prefix, this problem raised. Is this more complete network you want to use ?

arch0345:
Yeah as @LaoshuBaby said, we already have an entry for `IDS JMK` without the prefix already. It looks like one with the prefix was also added from the planet due to inconsistent tagging. If you want to match the existing routes tagged with a prefix so that these are tagged consistently, you should add `matchNames` to it

kudlav:
Hi, I've just search for network="cz:IDS JMK" using Overpass turbo, and there are currently no objects with the prefix variant. So it will be ok, to just remove the prefix variant.

Issue #4983, opened 2021-03-20
Everytime after npm run build it change LR to CRLF


LaoshuBaby:
I'm running nodejs on Windows, maybe different from developer's

So every time I run this command, it will change all files to CRLF

![微信图片_20210320115131](https://user-images.githubusercontent.com/42690037/111858287-b8860500-8972-11eb-8743-a44a4c5b1a53.jpg)

I guess this is because I'm a windows user.

So will this be like #4850  and #4871 ,is a environment based feature?

>
>每次交PR之前，都会感觉自己被淹没在changed files的海洋里，这太难受了！

![图片](https://user-images.githubusercontent.com/42690037/111858172-c4bd9280-8971-11eb-9d43-183e6bfdb0a1.png)

bhousel:
Oh yeah this is pretty common.. Can you try the suggestion here: https://stackoverflow.com/a/11535140

```bash
git config --global core.autocrlf true
```

I _think_ what that is supposed to do is tell git to not consider a line ending change as significant.  
So your `npm run build` may still change the line endings but git will not consider that a change worth checking in?
Let me know if that works?

LaoshuBaby:
Thanks @bhousel 

But sorry, it don't works……

![图片](https://user-images.githubusercontent.com/42690037/111861229-9d71c000-8987-11eb-92dc-f4782d7ae50b.png)

LaoshuBaby:
Besides, ```npm run test``` will failed because the crlf problem

It is so———— long that I only show the tail of Powershell Terminal

```
  90:2    error  Expected linebreaks to be 'LF' but found 'CRLF'  linebreak-style

✖ 3892 problems (3887 errors, 5 warnings)
  3887 errors and 0 warnings potentially fixable with the `--fix` option.

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! name-suggestion-index@5.0.0-pre lint: `eslint scripts/*.js lib/*.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the name-suggestion-index@5.0.0-pre lint script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\快乐的老鼠宝宝\AppData\Roaming\npm-cache\_logs\2021-03-20T06_22_48_706Z-debug.log
ERROR: "lint" exited with 1.
npm ERR! code ELIFECYCLE
[...]

PR #5912, opened 2021-12-20
Force LF for end-of-line

Mortein:
I was running in to the same issue as #4983 when using Windows.

Git stores files in the repository as either LF or CRLF, and by default converts them all to CRLF on Windows. When committing back, they're returned to whichever line ending they originally had (or set if the file is new).

When checking out the repo on Windows, the line endings were switching from LF to CRLF, `npm run build` would convert them back to LF, and then hundreds of files would be marked as modified.

Users can optionally set the `core.autocrlf` option from `true` to `false` or `input`, but this isn't really a great solution as it requires every Windows user to both know about the setting and to change it.

This PR sets all text files to use the LF line-ending on all systems (Windows, Linux, MacOS, etc.) using the [gitattributes file](https://www.git-scm.com/docs/gitattributes). Images have to be explicitly specified as binary, as for some reason they're detected as text. 

Users may need to force a reset of the line endings locally:
```
git rm --cached -r .
git reset --hard
```
or they can just delete their local repo and clone again.

This will unfortunately mean that Windows users cannot use Notepad to view files, but anything even slightly smarter (like Notepad++, or much more likely Visual Studio Code) won't have any issues.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,6 @@
+# Force LF line endings for all text files
+* text eol=lf
+
+# These filetypes are binary
+*.png binary
+*.gif binary


bhousel:
Thanks @Mortein this is a great fix 👍 
I thought this project had a .gitattributes file already but I guess not!

PR #5918, opened 2021-12-21
Citizens Advice

### LINKED ISSUE #5868: Citizens Advice in UK
I've noticed there is no established standard or preset for Citizens Advice previously known as Citizens Advice Bureau. Given they are an independent charity organisation with 316 branches specialising in giving advice to assist people with legal, debt, consumer, housing and other problems I'd suggest the tag `office=charity` though I am very open to suggestions if there is a more appropriate tag. However I can't see where I'd put this - would I need to create `charity.json` under `brands/amenity`?

There is also the question of the match tags as to date people have been tagging it using:

- amenity=advice
- office=citizens_advice
- social_facility=outreach
- office=lawyer
- office=ngo
- amenity=information
- office=government

codeinabox:
This resolves #5868 by adding a preset for Citizens Advice in the UK, previously known as Citizens Advice Bureau

### DIFF
--- a/data/brands/amenity/social_facility.json
+++ b/data/brands/amenity/social_facility.json
@@ -129,6 +129,30 @@
       }
     },
     {
+      "displayName": "Citizens Advice",
+      "locationSet": {"include": ["gb"]},
+      "matchNames": [
+        "citizens advice bureau",
+        "citizens' advice bureau"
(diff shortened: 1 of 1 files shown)


arch0345:
Thanks @codeinabox

PR #5921, opened 2021-12-22
Improve fuel stations and bicycle rental in China

Jyunhou:
Fuel stations:
+ CNOOC: `name=` `中国海洋石油`->`中国海油`
  中国海油
+ PipeChina: remove`fuel:lng=yes`, because it may be other fuel
+ 广西交投: new added

Bicycle rental:
+ K-Bike: more data

### DIFF
--- a/data/brands/amenity/bicycle_rental.json
+++ b/data/brands/amenity/bicycle_rental.json
@@ -1798,7 +1798,7 @@
       "displayName": "K Bike",
       "id": "kbike-1bb859",
       "locationSet": {"include": ["tw"]},
-      "matchNames": ["k bike"],
+      "matchNames": ["k bike", "金門縣政府"],
       "tags": {
         "amenity": "bicycle_rental",
         "network": "K-Bike",
--- a/data/brands/amenity/bicycle_rental.json
+++ b/data/brands/amenity/bicycle_rental.json
@@ -1798,11 +1798,20 @@
       "displayName": "K Bike",
       "id": "kbike-1bb859",
       "locationSet": {"include": ["tw"]},
+      "matchNames": ["k bike"],
       "tags": {
         "amenity": "bicycle_rental",
-        "brand": "K Bike",
-        "network": "K Bike",
-        "operator": "金門縣政府"
+        "network": "K-Bike",
+        "brand": "金門公共自行車",
+        "brand:en": "K-Bike",
+        "brand:zh": "金門公共自行車",
+        "brand:wikidata": "Q30942379",
+        "brand:wikipedia": "zh:金門縣公共自行車租賃系統",
+        "name": "金門公共自行車",
+        "name:en": "K-Bike",
+        "name:zh": "金門公共自行車",
+        "operator": "金門縣政府",
+        "operator:type": "government"
       }
     },
     {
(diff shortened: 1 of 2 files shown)


data/brands/amenity/bicycle_rental.json
```diff
@@ -4277,14 +4277,23 @@
...
+        "brand:en": "K-Bike",
+        "brand:zh": "金門公共自行車",
+        "brand:wikidata": "Q30942379",
+        "brand:wikipedia": "zh:金門縣公共自行車租賃系統",
+        "name": "金門公共自行車",
+        "name:en": "K-Bike",
+        "name:zh": "金門公共自行車",
+        "operator": "金門縣政府",
+        "operator:type": "government",
+        "website": "https://www.k-bike.com.tw/"
```
arch0345, line 4296: We don't support website tags in the NSI - this data can be accessed through the linked Wikidata page
```suggestion
```

data/brands/amenity/bicycle_rental.json
```diff
@@ -4277,14 +4277,23 @@
...
     },
     {
-      "displayName": "金門縣政府",
+      "displayName": "金門公共自行車",
       "id": "af5935-1bb859",
       "locationSet": {"include": ["tw"]},
-      "tags": {
-        "amenity": "bicycle_rental",
-        "brand": "金門縣政府",
-        "name": "金門縣政府",
```
arch0345, line 4286: If this name is still used in OSM, I would add this to matchNames so that existing POIs with this name will be matched with the new name

arch0345 (changes requested):
Thanks for these changes! I've left a couple of suggestions

data/brands/amenity/bicycle_rental.json
LaoshuBaby, line 4296: website不需要在这里写，你写进wikidata就行

LaoshuBaby (changes requested):

data/brands/amenity/bicycle_rental.json
Jyunhou: Hi. `金門縣政府` means `Kimen County Government`, which is the operator of K-Bike instead of name, I believe it's a mistake.

LaoshuBaby: 但是您需要matchNames里面添加原来的名称，才能让原来的名称匹配到现在的条目
but you should add it to matchname to let program detect it

data/brands/amenity/bicycle_rental.json
```diff
@@ -1798,11 +1798,20 @@
       "displayName": "K Bike",
       "id": "kbike-1bb859",
       "locationSet": {"include": ["tw"]},
+      "matchNames": ["k bike"],
```
LaoshuBaby, line 1801: 建议在这里添加“金门公共自行车”与“金门县政府”

suggested to add 2 old name in matchname

data/brands/amenity/bicycle_rental.json
Jyunhou: > 但是您需要matchNames里面添加原来的名称，才能让原来的名称匹配到现在的条目 but you should add it to matchname to let program detect it

There is no name=`金門縣政府` yet, there is only `name=*`, `network=K Bike`, `operator=金門縣政府`. See https://overpass-turbo.eu/s/1epm

LaoshuBaby: 不过，您需要确定未来有人这么写的时候，您依然可以正确匹配到，所以依然推荐在这里保留“金门县政府”
一开始就是这么写了才会被nsi-collector收录进来
However, you need to make sure that when someone writes this way in the future, you can still match it correctly, so it is still recommended to keep "Ginmen County Government" here
 That’s how it was written at the beginning to be included in nsi-collector

LaoshuBaby:
All matchNames had been fixed in https://github.com/osmlab/name-suggestion-index/commit/28333101c6dc5b14abf4727af735c516fa47a735

and website had been added in wikidata

PR #5922, opened 2021-12-22
Add common express in China, and matchNames for China Post

Jyunhou:


data/brands/amenity/post_office.json
```diff
@@ -138,6 +138,141 @@
         "name": "The UPS Store",
         "short_name": "UPS"
       }
+    },
```
LaoshuBaby, line 141: I don't agree move China Post to a brand from operator because it is a kind of service

LaoshuBaby (changes requested):

PR #5925, opened 2021-12-22
add Jolt

28688:


arch0345:
Thanks @28688! I've linked their Facebook on their Wikidata page so that their logo will be collected.

PR #5926, opened 2021-12-22
Add Cazoo car shops

codeinabox:
Cazoo customer centres across the UK https://www.cazoo.co.uk/customer-centres/


arch0345:
Thanks @codeinabox!

PR #5927, opened 2021-12-23
Adding BG&E and SMECO to power operators

Eric-Sparks:
Adding Baltimore Gas and Electric and Southern Maryland Electric Cooperative to both minor_line and pole operators.

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -2207,17 +2207,17 @@
     },
     {
       "displayName": "Британские пекарни",
+      "id": "britishbakery-1c12ef",
       "locationSet": {"include": ["ru"]},
-      "matchTags": ["shop/bakery"],
       "tags": {
         "amenity": "cafe",
         "brand": "Британские пекарни",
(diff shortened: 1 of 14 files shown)


Eric-Sparks:
SMECO needs an ID but I'm not sure how to create one (and couldn't find documentation around creating one).

arch0345:
Don't worry about adding ids, we run a script that automatically adds these.

arch0345:
I noticed you added `man_made=utility_pole` and the appropriate `utility=*` tags to the new entries in power/pole. However, the wiki page for this new tag suggests to use only `power=pole` for power poles (https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dutility_pole). It does show that `utility=power` can still be used for `power=pole` so I would just remove `man_made=utility_pole` from these

Eric-Sparks:
> It does show that `utility=power` can still be used for `power=pole` so I would just remove `man_made=utility_pole` from these

Okay, I've removed that tag from the patch.

PR #5930, opened 2021-12-24
Add UK Salvation Army Recycling Boxes

peternewman:
They list lots of other countries:
https://salvationarmy.org/

But I don't know which others have recycling boxes.

Also the one near me does shoes and clothes, I don't know if that's true for all, or in the US etc?

### DIFF
--- a/data/brands/amenity/recycling.json
+++ b/data/brands/amenity/recycling.json
@@ -224,7 +224,7 @@
     {
       "displayName": "The Salvation Army",
       "id": "thesalvationarmy-f8d7ba",
-      "locationSet": {"include": ["us"]},
+      "locationSet": {"include": ["gb", "us"]},
       "matchNames": ["salvation army"],
       "tags": {
         "amenity": "recycling",


data/brands/amenity/recycling.json
```diff
@@ -224,7 +224,7 @@
     {
       "displayName": "The Salvation Army",
       "id": "thesalvationarmy-f8d7ba",
-      "locationSet": {"include": ["us"]},
+      "locationSet": {"include": ["gb", "us"]},
```
peternewman, line 227: Should this mirror the charity shops:
https://github.com/osmlab/name-suggestion-index/blob/74b8e4e/data/brands/shop/charity.json#L516

peternewman:
I've now seen another (well technically four in one place) which did shoes and clothes, again in the UK.

If you're happy @arch0345 could we merge this as is (as the current change is technically accurate and an improvement on the current situation), and I'll open an issue to track the bits which need research?

bhousel:
> If you're happy @arch0345 could we merge this as is (as the current change is technically accurate and an improvement on the current situation), and I'll open an issue to track the bits which need research?

sounds good, thanks @peternewman 👍

data/brands/amenity/recycling.json
peternewman: Raised in #5981 5981

PR #5933, opened 2021-12-26
Add EP: (ElectronicPartner)

1-Byte:
Website: https://www.ep.de

I'm not sure if `brand` should be set to "EP:" (with colon) which would be more accurate but might cause issues (?).

### DIFF
--- a/data/brands/shop/electronics.json
+++ b/data/brands/shop/electronics.json
@@ -331,6 +331,20 @@
       }
     },
     {
+      "displayName": "EP:",
+      "id": "ep-a53b66",
+      "locationSet": {
+        "include": ["at", "ch", "de", "nl"]
+      },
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Thanks @1-Byte - I would probably add the brand with the colon "_EP:_" and add a matchName for without "_ep_".

1-Byte:
Thank you for the review. I added the colon to the brand name. Adding "electronic partner" to `matchNames` does not seem to work (because of normalization?).

bhousel:
> Thank you for the review. I added the colon to the brand name. Adding "electronic partner" to `matchNames` does not seem to work (because of normalization?).

Yep, looks good 👍 
You are correct the variations in punctation and spaces are ignored so it doesn't need to go in matchNames (and the build script will remove it).

PR #5942, opened 2021-12-27
Create union

### LINKED ISSUE #5940: Please add German union "IG BCE" in brands/office/union or brands/office/association
brand=IG BCE
brand:wikidata=Q871045
brand:wikipedia=de:IG BCE
office=union
official_name=IG Bergbau, Chemie, Energie

matchNames:
ig bce

kjonosm:
Create union.json and add major German unions
Closes #5940, #5941


91560:
Thank you.

bhousel:
This looks pretty good to me!  
We are [currently collecting `office=*`](https://github.com/ideditor/nsi-collector/blob/main/collect_osm.js#L29) brand and name values, but there aren't enough in the OSM planet for them to automatically appear here yet.  So I think this is great that we are being proactive in establishing common tagging for union offices.

Did you add `"preserveTags": ["^official_name"],` because it is likely that mappers will customize this value?   We can add this instead to the top so it applies to the whole file, if you want?

kjonosm:
People might want to use `name=`, `official_name=` and also `short_name=` to specify local or regional branches of a union or association see e.g. https://igbau.de/Bezirksverbaende.html.

So adding a single`"preserveTags": ["^official_name"]` in the header instead of each item makes sense.

It might also make sense to maybe treat other `office=` tags the same way...

bhousel:
ok in 9b0838299 I changed it to `"preserveTags": ["^(official_|short_)?name"]` so the validator should leave all these name tags alone.

PR #5945, opened 2021-12-28
Restrict parks to local region

Mortein:
Restrict operators of parks to to their local region.


Mortein:
Updated a bunch of park operators using location 001 with smallest available location (state from features, or country).

arch0345:
Thanks @Mortein for refining the locationSets for all of these park operators!

PR #5954, opened 2021-12-29
Rapha Clubhouses

codeinabox:
Rapha has stores across the globe that primarily sell their cycling apparel and also have cafes inside them

https://www.rapha.cc/gb/en/clubhouses

### DIFF
--- a/data/brands/shop/sports.json
+++ b/data/brands/shop/sports.json
@@ -499,6 +499,8 @@
         "brand:ja": "ラファ",
         "brand:wikidata": "Q7293931",
         "brand:wikipedia": "en:Rapha (sportswear)",
+        "cuisine": "coffee_shop",
+        "food": "yes",
         "name": "Rapha",
         "name:en": "Rapha",
         "name:ja": "ラファ",
--- a/data/brands/shop/sports.json
+++ b/data/brands/shop/sports.json
@@ -473,6 +473,40 @@
       }
     },
     {
+      "displayName": "Rapha",
+      "locationSet": {
+        "include": [
+          "au",
+          "de",
+          "dk",
+          "es",
+          "gb",
+          "hk",
+          "jp",
+          "kr",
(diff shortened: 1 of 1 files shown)


data/brands/shop/sports.json
```diff
@@ -472,6 +472,42 @@
...
+        "shop/bicycle",
+        "shop/clothes"
+      ],
+      "tags": {
+        "brand": "Rapha",
+        "brand:en": "Rapha",
+        "brand:ja": "ラファ",
+        "brand:wikidata": "Q7293931",
+        "brand:wikipedia": "en:Rapha (sportswear)",
+        "cuisine": "coffee_shop",
```
LaoshuBaby, line 502: Add a cuisine in a sport clothes shop maybe a little strange?

My personal opinion is separate them as **IKEA** and **IKEA restaurant**.

IKEA:
https://github.com/osmlab/name-suggestion-index/blob/1854a16c88c758434bb969ddbb037b178865be62/data/brands/shop/furniture.json#L601-L612

IKEA cafe:
https://github.com/osmlab/name-suggestion-index/blob/ed5ecbc1db752eb174b0e72d3ef44c287c62c54d/data/brands/amenity/cafe.json#L1330-L1345

arch0345: Yeah I agree with @LaoshuBaby - these should really be separate entries. Another good example is how drug stores (at least in the States) are mapped as a `shop=chemist` building with a `amenity=pharmacy` node at the pharmacy counter. There's even a wiki page about this: [One feature, one OSM element](https://wiki.openstreetmap.org/wiki/One_feature,_one_OSM_element)

codeinabox: Understood. The shop cafes don't warrant their own entry in the NSI as they are a small part of the shop but I will remove the food and drink tags.

PR #5959, opened 2021-12-29
Shorten name of State Savings Bank of Ukraine

SmartFinn:
During discussion the Ukrainian community in [Telegram](https://t.me/osmUA) decide replace "State Savings Bank of Ukraine" with shorter name "Oshadbank" and add the full name to official_name:en. ATMs have already been updated [CS 115248505](https://www.openstreetmap.org/changeset/115248505)

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -10136,6 +10136,7 @@
       "tags": {
         "amenity": "bank",
         "brand": "Ощадбанк",
+        "brand:uk": "Ощадбанк",
         "brand:en": "Oschadbank",
         "brand:wikidata": "Q4340839",
         "brand:wikipedia": "uk:Ощадбанк",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/bank.json
```diff
@@ -10140,7 +10140,11 @@
         "brand:wikidata": "Q4340839",
         "brand:wikipedia": "uk:Ощадбанк",
         "name": "Ощадбанк",
-        "name:en": "State Savings Bank of Ukraine"
+        "name:en": "Oschadbank",
```
LaoshuBaby, line 10143: 1. why not change `brand:en` too?
2. If only have "*name"+"*name:en" but without "*name:local", many program will prefer the English version, see: https://wiki.openstreetmap.org/wiki/Multilingual_names#Repeating_name_with_language_specific_tag

SmartFinn: Thanks for the remarks it's fixed now.

bhousel:
Thanks @SmartFinn & welcome 👋

PR #5960, opened 2021-12-30
Add denominations tree

### LINKED ISSUE #5232: Adding amenity=place_of_worship to operator tree
This would be helpful for places of worship operated by a common party, such as The Church of Jesus Christ of Latter-day Saints and Kingdom Hall of Jehovah's Witnesses. Recommend to also add landuse=religious to the operator tree using the same operators. iD presets already include Kingdom Hall of Jehovah's Witnesses without any operator tags, so that should be removed if this were to be added to the index.

### LINKED ISSUE #6874: Will NSI handle with religion/denomination standardrization preset?
## Background

In OSMTW's Telegram group Supaplex and @littlebtc [raise a proposal](https://t.me/OSMTaiwan/6886) that add Taiwan local denomination to NSI, like "Tzu Chi(Q717899)", "Fo Guang Shan(Q548694)" or "True Jesus Church(Q1859)"

## How to do

There are mainly four categories at present, `brand/flag/operator/transit`, and for example, `operators/amenity/place_of_worship` means a group of chain churches that are all operated by "XX Church"...

If it is religion, the path in NSI is may like `(religion/denomination)/amenity/place_of_worship`, and this will need to create a new catagory.

**Should we introduce those kind of data to NSI?** I need some suggestion from other maintainer. 😥

arch0345:
So far I've just added denominations found in [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema/tree/main/data/presets/amenity/place_of_worship). If this gets merged we'll probably have to remove these presets from that repo.

### Issues
The build script can't generate an `id` without a `name` for each entry in the `denominations` tree. How are the transit/operator trees set up so that it can generate an `id` without having a `name` tag?

Closes #5232 
Closes #6874


data/denominations/amenity/place_of_worship.json
```diff
@@ -0,0 +1,52 @@
+{
+  "properties": {
+    "path": "denominations/amenity/place_of_worship",
+    "skipCollection": true,
+    "exclude": {"generic": ["^place of worship$"]}
+  },
+  "items": [
+    {
+      "displayName": "Kingdom Hall of Jehovah's Witnesses",
```
1ec5, line 9: Many denominations translate their names into as many local languages as possible. The appropriate name usually varies by language more than by region. As it stands, id-tagging-schema is better suited to localization than NSI. This is already an issue for the flags tree, but the issue would be more acute for any denomination tree.

Or would we limit this tree to just the denominations that have a specialized name for all their places of worship?

data/denominations/amenity/place_of_worship.json
```diff
@@ -0,0 +1,52 @@
+{
+  "properties": {
+    "path": "denominations/amenity/place_of_worship",
```
1ec5, line 3: Besides places of worship, schools, community centers, and social facilities are often identified by their denomination as well. The editor would have to make sure not to “upgrade” something to a place of worship just because it has a matching denomination.

1ec5 (commented):
With a key like `denomination` that’s designed to be an enumeration of well-known values, rather than freeform text, id-tagging-schema still looks more promising to me as a mechanism for standardizing and localizing tags: openstreetmap/id-tagging-schema#254.

data/denominations/amenity/place_of_worship.json
arch0345: Yeah I was also thinking about how the displayNames for these denominations could be translated to a region's local language. I agree that id-tagging-schema would be a better place to have these presets because of these current limitations - I hadn't really thought about how the default presets are localized

> Or would we limit this tree to just the denominations that have a specialized name and icon for all their places of worship?

This could work, though I can't think of any off the top of my head. I was also looking into how the [`denomination=mormon`](https://wiki.openstreetmap.org/wiki/Tag:denomination%3Dmormon) wiki page suggests to use operator tags to distinguish The Church of Jesus Christ of Latter-day Saints from other splinter groups. This particular case could work in NSI under the existing operator tree as localized operator values would be possible.

> These three presets only exist because it was considered inappropriate to use the ✝️ icon for these denominations.

This makes sense now. I was wondering why there were only these three denomination presets

data/denominations/amenity/place_of_worship.json
arch0345: That's a good point - I was thinking about adding schools to the denominations tree but left it out for now. IIRC iD only matches items of a different category if it falls into one of these [matchGroups](https://github.com/osmlab/name-suggestion-index/blob/main/dist/matchGroups.json)

bhousel: Yeah I think in this situation I think it should only offer to upgrade `amenity/place_of_worship`, or maybe also some generically tagged features like `building=yes` or `amenity=yes` when paired with a matching `name` or `operator`.

data/denominations/amenity/place_of_worship.json
bhousel: I really don't want to introduce localization into the Name Suggestion Index display names.  Display name is just the label that the preset shows in RapiD and JOSM.

But I think keeping this tree similar to how the `flags` tree works would be ok (`denomination` is an identifying value, and `name` can be whatever people want).

bhousel:
> The build script can't generate an id without a name for each entry in the denominations tree. How are the transit/operator trees set up so that it can generate an id without having a name tag?

You'll need to change idgen.js too
https://github.com/osmlab/name-suggestion-index/blob/main/lib/idgen.js

data/denominations/amenity/place_of_worship.json
1ec5: > But I think keeping this tree similar to how the `flags` tree works would be ok (`denomination` is an identifying value, and `name` can be whatever people want).

#5982 tracks making flags localizable. Maybe the solution there could inform how we’d make denomination entries localizable too. But in the meantime, id-tagging-schema at least offers a way to localize the display names, even if it doesn’t have a straightforward way to filter denominations by religion (something we’d have to solve as part of this project too).

PR #5963, opened 2021-12-30
Adding BG&E and SMECO to power lines

Eric-Sparks:
The last update to add Baltimore Gas and Electric and Southern Maryland Electric Cooperative to minor_line and poles worked quite well.  I've found places where these operators should be added to power lines as well.

### DIFF
--- a/data/operators/power/line.json
+++ b/data/operators/power/line.json
@@ -459,6 +459,7 @@
       "locationSet": {"include": ["us"]},
       "matchNames": ["bg&e", "bge"],
       "tags": {
+        "frequency": "60",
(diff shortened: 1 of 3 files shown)


arch0345 (changes requested):
All operators in the line category are transferred over to the minor_line and pole categories, so the entries in those two categories should be removed

Eric-Sparks:
> All operators in the line category are transferred over to the minor_line and pole categories, so the entries in those two categories should be removed

That makes sense.  I didn't even see that pull before.  Should be fixed now.

arch0345 (approved):

PR #5977, opened 2022-01-01
Restrict parks to state/country

Mortein:


bhousel:
Thanks @Mortein, I spot checked a few of these and they seem fine..
They still won't appear in RapiD without a `operator:wikidata` value, but it's good to get the locationSet correct as a first pass.

PR #5986, opened 2022-01-03
Added Lyft Rentals and edited Amazon Hub Lockers

PierceWHoward:


bhousel:
For Lyft Rentals:
It feels a little weird that you just added https://www.wikidata.org/wiki/Q110392571 for the Lyft Rentals.  
We probably could have just used Lyft's QID https://www.wikidata.org/wiki/Q17077936 for this.  I worry that we're going to get this QID removed by Wikidata admins as spammy.

For Amazon Hub Lockers:
I feel like we're probably doing this one wrong.  We should probably just call them "Amazon Locker" and matchname the "Amazon Hub" and "Amazon Hub Locker" names.
https://www.lifewire.com/amazon-lockers-and-hubs-4154922

ttomasz:
FYI my pull request touches Amazon Hub Locker in #6020

Lukas458:
> For Amazon Hub Lockers:
I feel like we're probably doing this one wrong. We should probably just call them "Amazon Locker" and matchname the "Amazon Hub" and "Amazon Hub Locker" names.

That might be a good idea.

the tag `brand=Amazon Hub`, which we use in the preset `amazonhublocker-f85b1f` together with `brand:wikidata=Q16974764` is also called **"Amazon Locker"** at wikidata and not **"Amazon Hub"**.

According to a blog, Amazon itself says that "Amazon Hub" is the umbrella brand to which "Amazon Locker" belongs. Under the "Amazon Hub" brand, however, other places such as e.g. post offices operated (see also https://github.com/osmlab/name-suggestion-index/issues/6161#issuecomment-1029907694)

PR #5987, opened 2022-01-03
Streamer Coffee Company

codeinabox:


data/brands/amenity/cafe.json
```diff
@@ -1870,6 +1870,18 @@
         "takeaway": "yes"
       }
     },
+{
+      "displayName": "Streamer Coffee Company",
+      "locationSet": {"include": ["jp"]},
```
LaoshuBaby, line 1875: Although this a English name but after taginfo search this brand are truly in Japan

图片

Issue #5934, opened 2021-12-27
Missing tags for entries


arch0345:
Copy-pasted these warnings from the build script:

- [x]   "Naturals (naturals-2dc292)" -> missing tag? -> "beauty"
- [x]   "Regiomat (regiomat-a242c5)" -> missing tag? -> "vending"
- [x]   "Selecta (selecta-ed4d31)" -> missing tag? -> "vending"
- [x]   "Tokeny.Store (tokenystore-eb9a2c)" -> missing tag? -> "vending"
- [x]   "ALEX (alex-051971)" -> missing tag? -> "cuisine"
- [x]   "Cactus Club Cafe (cactusclubcafe-e68b16)" -> missing tag? -> "cuisine"
- [x]   "Cafe & Bar Celona (cafeandbarcelona-3fc85c)" -> missing tag? -> "cuisine"
- [x]   "Cafe Del Sol (cafedelsol-b8cc79)" -> missing tag? -> "cuisine"
- [x]   "Caravan (caravan-e997f8)" -> missing tag? -> "cuisine"
- [x]   "De Beren (deberen-2b67bf)" -> missing tag? -> "cuisine"
- [x]   "Doppio Malto (doppiomalto-94430a)" -> missing tag? -> "cuisine"
- [x]   "Earls (earls-85c92c)" -> missing tag? -> "cuisine"
- [x]   "Fournos Bakery (fournosbakery-69fbf5)" -> missing tag? -> "cuisine"
- [x]   "Honest Greens (honestgreens-1d6b00)" -> missing tag? -> "cuisine"
- [x]   "Interspar Restaurant (intersparrestaurant-c7016b)" -> missing tag? -> "cuisine"
- [x]   "Joey (joey-85c92c)" -> missing tag? -> "cuisine"
- [x]   "Karas (karas-86098b)" -> missing tag? -> "cuisine"
- [x]   "Kauai (kauai-14572f)" -> missing tag? -> "cuisine"
- [x]   "La Place (laplace-179420)" -> missing tag? -> "cuisine"
- [x]   "Lunch Garden (lunchgarden-c1dacf)" -> missing tag? -> "cuisine"
- [x]   "Madero (madero-ad49e6)" -> missing tag? -> "cuisine"
- [x]   "Mary Grace (marygrace-5259a4)" -> missing tag? -> "cuisine"
- [x]   "Mildreds (mildreds-e997f8)" -> missing tag? -> "cuisine"
- [x]   #8911
[...]

PR #5991, opened 2022-01-03
Add missing cuisine tags

Mortein:
From #5934

Valentino is not fixed, it's just yet another variation.


UKChris-osm:
Thanks @Mortein 👍

PR #5994, opened 2022-01-04
Update wholesale.json

dan13l9:
remove capital letters and other words associated with the brand name


arch0345 (commented):
I'm not really familiar with the other brands but I know that CHEF'STORE is prominently branded with that capitalization (https://www.usfoods.com/why-us-foods/chef-store.html).

bhousel:
I agree with @arch0345 , these should probably stay how they are.

PR #6000, opened 2022-01-04
Update fast_food.json

dan13l9:


bhousel:
I don't really believe that you've researched all of these.. 
I'm pretty sure another contributor put all of those `"exclude": ["pl"]` there for a reason, so they should probably stay.

PR #6002, opened 2022-01-04
Planet Organic

codeinabox:
Planet Organic an organic only supermarket with shops spread over London

https://www.planetorganic.com/pages/our-stores


UKChris-osm:
Thank you @codeinabox 👍

PR #6003, opened 2022-01-04
Updated bicycle_rental.json

kyy0602:
Hamilton Bike Share didn't show up in iD because I didn't add the "brand" value.

### DIFF
--- a/data/brands/amenity/bicycle_rental.json
+++ b/data/brands/amenity/bicycle_rental.json
@@ -1644,7 +1644,7 @@
     },
     {
       "displayName": "Hamilton Bike Share",
-      "id": "hamiltonbikeshareinc-3eadaa",
+      "id": "hamiltonbikeshare-3eadaa",
       "locationSet": {"include": ["ca"]},
       "matchNames": [
         "hamilton social bicycles",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/bicycle_rental.json
```diff
@@ -1644,14 +1644,15 @@
     },
     {
       "displayName": "Hamilton Bike Share",
-      "id": "hamiltonbikeshareinc-3eadaa",
+      "id": "hamiltonbikeshare-3eadaa",
       "locationSet": {"include": ["ca"]},
```
LaoshuBaby, line 1648: notice：
you don't need to change id because operator have the word Inc so every time program will use operator generate it

LaoshuBaby:
Besides, Is this brand has a wikidata?

kyy0602:
> Besides, Is this brand has a wikidata?

Oh yes. I made one (Q109332552) a while ago but forgot to include it. Let me create a new PR along with some other changes.

data/brands/amenity/bicycle_rental.json
kyy0602: Didn't notice that. I ran npm before committing so probably it was done automatically :)

PR #6004, opened 2022-01-04
Updated GO Transit bus and train

kyy0602:
Changed names and added operators to better reflect their operations.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -4653,20 +4653,26 @@
       "tags": {"network": "GMPTE", "route": "bus"}
     },
     {
-      "displayName": "GO Bus",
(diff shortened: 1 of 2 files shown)


data/transit/route/bus.json
```diff
@@ -4670,6 +4661,21 @@
         "route": "bus"
       }
     },
+    {
+      "displayName": "GO Transit",
```
LaoshuBaby, line 4665:
```suggestion
      "displayName": "GO Transit(Bus)",
```

LaoshuBaby (approved):

arch0345:
These could probably be restricted to Ontario (`ca-on.geojson`)

PR #6009, opened 2022-01-05
Update mobile_phone.json

giorgiomec:
TIM (Telecom Italia Mobile) is the name of the brand. Gruppo TIM is only the company, so I updated the Wikidata item


arch0345:
Thanks @giorgiomec!

PR #6012, opened 2022-01-05
Updated Mobil and GO Transit

kyy0602:
Updated the Wikidata and Wikipedia tags used for Mobil petrol stations outside of Japan; restricted the use of GO Transit train and bus routes to Ontario, Canada (`ca-on.geojson`).

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -2849,8 +2849,8 @@
       "tags": {
         "amenity": "fuel",
         "brand": "Mobil",
-        "brand:wikidata": "Q109676002",
-        "brand:wikipedia": "ja:モービル (ブランド)",
+        "brand:wikidata": "Q3088656",
+        "brand:wikipedia": "en:Mobil",
         "name": "Mobil"
       }
     },
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -4664,7 +4664,7 @@
     {
       "displayName": "GO Transit",
       "id": "gotransit-50b470",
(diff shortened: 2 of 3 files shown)


arch0345 (commented):
I wouldn't change the Wikidata QID for Mobil: the one you linked is for the defunct company, so another was recently created for the brand under ExxonMobil:
https://github.com/osmlab/name-suggestion-index/commit/ce2472e483d723c7e267e7df7331bff69e5961f1

I agree though that the Japanese Wikipedia article might be confusing for people, but unfortunately there's no equivalent English article for the brand yet. We run a script before a new release that automatically adds Wikipedia articles linked in the Wikidata page, which is why this article was linked. Because of scenarios like this, there have been talks of getting rid of `:wikipedia` tags altogether in the index. I'll remember to remove this Wikipedia tag manually before the next release

LaoshuBaby:
If you want to change those 3 transport network first priority you can separate this PR to 2, so that we can merge locationSet change first

kyy0602:
> I wouldn't change the Wikidata QID for Mobil: the one you linked is for the defunct company, so another was recently created for the brand under ExxonMobil: [ce2472e](https://github.com/osmlab/name-suggestion-index/commit/ce2472e483d723c7e267e7df7331bff69e5961f1)

Oops. Indeed I was very confused while updating a Mobil in iD but thanks for the info.

> If you want to change those 3 transport network first priority you can separate this PR to 2, so that we can merge locationSet change first

Sure. I will close this PR and create a new one then.

PR #6015, opened 2022-01-06
Updated GO Transit bus and train, Hamilton Street Railway

kyy0602:
Restricted them to Ontario, Canada (`ca-on.geojson`).


arch0345:
Thanks @kyy0602!

PR #6020, opened 2022-01-07
Parcel locker category change

ttomasz:
Implements proposal changing parcel lockers tagging which was recently approved: https://wiki.openstreetmap.org/wiki/Proposed_features/amenity%3Dparcel_locker

Essentially just moves brands that had vending_machine=parcel_pickup to amenity=parcel_locker

I modified brand display name and name tag from "GLS" to "GLS Parcel Locker" since GLS may be too general and was inconsistent with e.g. Amazon and DHL.


arch0345 (commented):
Thanks @ttomasz and welcome! I'm going to wait for another maintainer to review this as there currently isn't a preset for this new tag in [`id-tagging-schema`](https://github.com/openstreetmap/id-tagging-schema), meaning that these brands will no longer have their own preset in iD. Additionally, existing parcel lockers tagged as `amenity=vending_machine` won't get matched without `matchTags` on each entry or an entry in [`matchGroups`](https://github.com/osmlab/name-suggestion-index/blob/main/config/matchGroups.json).

Edit: Looks like you made a PR in id-tagging-schema so we should probably wait until this gets merged.

ttomasz:
Thanks. I created pull request adjusting iD presets: https://github.com/openstreetmap/id-tagging-schema/pull/339
It was merged today.

I also added entry to matchGroups.

Let me know if it's okay now or are there any fixes needed :)

arch0345:
Looks good :+1:

PR #6022, opened 2022-01-07
add 9 New Zealand brands

k-yle:
cc @Mortein, I see you've added some nz brands recently

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -8599,7 +8599,19 @@
       }
     },
     {
-      "displayName": "TSB",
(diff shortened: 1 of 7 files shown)


arch0345:
Thanks @k-yle and welcome! Note that the entries you added without a `:wikidata` tag won't get generated as an iD preset. If these brands don't have one already, it's pretty easy to create one: just make sure to include statements like instance of, country, official website, etc. to prevent it from getting deleted. I also recommend linking their Facebook/Twitter usernames on Wikidata as that's how we collect logos.

PR #6027, opened 2022-01-08
Pizza Union

codeinabox:
Pizza Union has locations around London. I've added a restaurant match tag as often it's been incorrectly tagged as such, when actually it is fast food as you have to order at the counter, tables are first come first served and they even advertise themselves as "superfast pizza".

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -4391,6 +4391,20 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
> when actually it is fast food as you have to order at the counter, tables are first come first served

Yes so we prefer to think it is a kind of fastfood, talked in #4987

PR #6033, opened 2022-01-08
McDonald's in France have unique names

gileri:
In France (at least), McDonald's restaurants have unique/quasi-unique, public-facing names. They can be visible [here](https://www.restaurants.mcdonalds.fr/).

This PR creates a France-specific item that uses the french wikipedia article, does not suggest an incomplete name, while keeping  the other NSI goodies.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -3528,6 +3528,19 @@
       }
     },
     {
+      "displayName": "McDonald's (France)",
+      "id": "mcdonalds-e4dee6",
+      "locationSet": {"include": ["fr"]},
+      "tags": {
+        "amenity": "fast_food",
+        "brand": "McDonald's",
+        "brand:wikidata": "Q38076",
+        "brand:wikipedia": "fr:McDonald's",
+        "cuisine": "burger",
+        "takeaway": "yes"
+      }
+    },
+    {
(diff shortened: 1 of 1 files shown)


arch0345:
Thanks @gileri and welcome! Would you be able to screenshot an example? I'm not able to access mcdonalds.fr for some reason.

LaoshuBaby:
Welcome to NSI, but I have some questions about this.

1. I was also blocked by the website, is this website only can be browsed by a french IP? If so, a screenshot is needed.
图片

2. You can use `preserveTags` to aviod the changing of `name`, and If you want to create a new entry only apply to France, you'd better let the global one (Which include 001) exclude `fr`

gileri:
Thank you both for the warm welcome !

> Would you be able to screenshot an example? I'm not able to access mcdonalds.fr for some reason.

You can have two even :
Restaurant name on website 1
Restaurant name on website 2

>You can use preserveTags to aviod the changing of name
I was going to do that, but to my understanding that would still suggest incomplete names on restaurants that do not currently have a name in OSM.

>If you want to create a new entry only apply to France, you'd better let the global one (Which include 001) exclude fr

You're totally right, thanks ! I assumed for some reason that this entry would override the original in France, but of course it won't... Pushed in d94da900e.

arch0345:
Looks good 👍

PR #6036, opened 2022-01-09
Add brand 'Sonnentor'

haansn08:
Also extends the location set for the 'Pearle' brand to include Austria:
https://www.openstreetmap.org/node/809527980


arch0345:
Thanks @haansn08!

PR #6038, opened 2022-01-10
Add Danish auto repair shops

mikini:


arch0345 (changes requested):
Thanks @mikini and welcome! Note that we don't support website tags as this information can be linked on a Wikidata page. I would search for these brands on https://wikidata.org to see if a page already exists. If they don't have one, it's pretty easy to create one: I recommend adding statements such as instance of, country, official website, and Facebook/Twitter ID to prevent the page from being deleted

mikini:
Hi @arch0345, and thanks. I've added wikidata entries, swapped website out for wikidata and rebased.

Also, I've elaborated about wikidata and detailed tags not being accepted [on the Contributing wiki page](https://github.com/osmlab/name-suggestion-index/wiki/Contributing/7f57e5fa1a4dab785e4a012c926bc6eedb51fd7f). Hopefully, you can avoid some repetition of that message in the future ;).

arch0345 (approved):
I've been meaning to add this to the wiki but never got around to adding it, thanks @mikini!

PR #6044, opened 2022-01-11
Update post_office.json

Mannivu:
Fix operator's name with the correct capitalisation according to official site, Wikidata and Wikipedia


bhousel:
Thanks @Mannivu and welcome 👋

PR #6045, opened 2022-01-11
New Zealand brands

Mortein:
_Hell Pizza_ is more `fast_food` than `restaurant`, as they don't really offer a sit-down service. _Domino's_ in probably a good comparison. This sound right, @k-yle?

Add _Bin Inn_, a bulk food store.
Add _SuperValue_, a supermarket chain (generally in small towns).

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -2479,6 +2479,29 @@
       }
     },
     {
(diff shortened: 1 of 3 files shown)


k-yle (approved):
The change to _Hell Pizza_ makes sense 👍

arch0345:
Thanks @Mortein

PR #6048, opened 2022-01-11
Add Econt and Speedy parcel machines

Dimitar5555:
Econt has a special name for their parcel machines (Econtomat). I've added it to the `name` fields, but I'm not sure if it's correct.

### DIFF
--- a/data/brands/amenity/parcel_locker.json
+++ b/data/brands/amenity/parcel_locker.json
@@ -254,8 +254,10 @@
         "brand": "Еконт",
         "brand:wikidata": "Q12279603",
         "name": "Еконт",
+        "name:bg": "Еконт",
         "name:en": "Econt",
         "alt_name": "Еконтомат",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/parcel_locker.json
```diff
@@ -245,6 +245,33 @@
...
+    {
+      "displayName": "Speedy",
+      "locationSet": {"include": ["bg"]},
+      "tags": {
+        "amenity": "parcel_locker",
+        "brand": "Speedy",
+        "brand:wikidata": "Q541030",
+        "brand:wikipedia": "en:DPDgroup",
+        "name": "Speedy",
+        "name:en": "Speedy",
```
arch0345, line 272: Multilingual name not needed as they're the same
```suggestion
```

arch0345 (approved):
Thanks @Dimitar5555!

PR #6060, opened 2022-01-13
The Barn cafes

codeinabox:
The Barn has cafes across Berlin and in Dubai https://thebarn.de/pages/locations


arch0345:
Berlin has its own geojson so I replaced `de` with it

PR #6069, opened 2022-01-15
Update pub.json

### LINKED ISSUE #6068: All pubs containing name "Irish Pub" are flagged as generic
All pubs in iD editor containing the name Irish Pub are flagged as generic.

image

Looks like there's an issue with the regular expression matching "irish pub" - should be /^irish pub$/ not /irish pub/

andrewpmk:
Fix generic regular expression so all pubs with name containing "Irish Pub" or "Bistrot" are not matched and flagged as "Pub has suspicious name XYZ Irish Pub" - regular expression should start with ^ and end with $

Fixes #6068

### DIFF
--- a/data/brands/amenity/pub.json
+++ b/data/brands/amenity/pub.json
@@ -6,8 +6,8 @@
       "generic": [
         "^(bar|bistro|pub)$",
         "^(commercial )?hotel$",
-        "bistrot",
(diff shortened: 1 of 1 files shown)


arch0345:
Thanks for catching this @andrewpmk

Issue #5500, opened 2021-10-13
Store names which include the branch name


NKAmapper:
Store names in Norway includes the branch name, for example "Spar Arendal Sentrum", not just "Spar", as [discussed](https://lists.nuug.no/pipermail/kart/2012-April/003391.html) by the local community. This is also the case for other categories such as fuel stations, hotels, restaurants etc .

How could this be specified in the index?

Currently, iD is encouraging users to strip the name tag for everything but the brand name, causing a lot of unwanted edits.

arch0345:
By adding `"preserveTags": ["^name"],` to the entry, iD will keep the name of the locations and not suggest splitting into `name=*` and `branch=*`. The `name=*` will only be used when a new POI is added. Here is an example of this:
```
    {
      "displayName": "Verizon",
      "id": "verizon-ce3e5c",
      "locationSet": {"include": ["us"]},
      "preserveTags": ["^name"],
      "tags": {
        "alt_name": "Verizon Wireless",
        "brand": "Verizon",
        "brand:wikidata": "Q919641",
        "brand:wikipedia": "en:Verizon (mobile network)",
        "name": "Verizon",
        "shop": "mobile_phone"
      }
    },
```
All of the entries can be easily viewed under the `data` folder on the main page. Let me know if you need any help with this.

NKAmapper:
I think the index is a great idea, I just did not realize that it would enforce one way of using the name=* tag. I just discovered that a large number of store names had been edited in Norway and tried to understand how that happened.

I guess both methods (name=* with and without the branch name) are being used in other countries as well, but I have not checked. A problem with branch is that search is in general not supported, at least not in Nominatim and in all the apps I have tried. A search will then usually just produce a list of identical brand names within the same city, which is not very helpful.

The practice of name=brand+branch has in fact been used since 2012, and is being actively maintained that way. We have put a lot of effort into getting a complete and well structured set of the most common retail chains in the country into OSM. It has been discussed several times on IRC/#osm-no, including today.

Adamant36:
> No, but not everything in OSM can be seen on the ground. It doesn't mean it isn't the correct information. Many roads have names, but only the ref has a sign, many houses have a house number without having an actual number on the wall.

[...]

PR #6071, opened 2022-01-15
Unique fast-food names

gileri:
Following my [last PR](https://github.com/osmlab/name-suggestion-index/pull/6033), I suggest the same adjustement for other franchises : Do not suggest incomplete names for franchises with unique, public-facing names. That will allow users to fill out the correct name using various editors such as iD, or StreetComplete.

# Fast Foods
## Bagel Corner
bagelcorner-e4dee6
https://restaurants.bagelcorner.fr/

## bigfernand-8ddd61
https://bigfernand.com/restaurants/

## exki-a3472e
https://www.exki.com/restaurants

## fiveguys-200282
https://restaurants.fiveguys.com/pa/allentown
https://restaurants.fiveguys.fr/
https://www.societe.com/etablissement/five-guys-81751824400287.html
https://restaurants.fiveguys.be/en_be
https://restaurants.fiveguys.ae/en_ae
https://restaurants.fiveguys.ca/ab/calgary
https://restaurants.fiveguys.de/
https://restaurantes.fiveguys.es/index.html
[...]

## flowerburger-777b0e
https://www.flowerburger.it/negozi/flower-burger-milano-garibaldi/
https://www.flowerburger.fr/stores/flower-burger-marseille/

## laboiteapizza-e4dee6
https://www.laboiteapizza.com/landing/

## lacroissanterie-7eae30
https://www.lacroissanterie.fr/trouver-restaurant/store/12/ANGOULEME%201%20(LA%20COURONNE)

## malvon-d3d66f
https://www.malvon.es/tiendas

## pommedepain-508db0
https://pommedepain.fr/restaurant/croixrault/
https://pommedepain.fr/restaurant/grombalia/

## pretamanger-027ef2
Only US have purely address based "names". Other places have either unique names, or just 1 locations at the moment.
https://locations.pretamanger.fr/paris/place-louis-armand
https://www.pret.hk/en-HK/find-a-pret
https://www.pret.co.uk/en-GB/shop-finder

## quick-9fdafe
https://www.quick.fr/aulnay-cc
https://www.quick.lu/fr/restaurant/kirchberg
https://www.quick.be/fr/restaurant/gare-du-midi

## sushishop-ac545e
https://www.sushishop.fr/fr/restaurants
https://www.sushiart.ae/en/restaurants
https://www.sushishop.be/fr/restaurants
https://www.mysushishop.ch/fr/restaurants
https://www.sushishop.eu/

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -369,7 +369,7 @@
       "displayName": "Big Fernand",
       "id": "bigfernand-8ddd61",
       "locationSet": {
-        "include": ["ae", "fr", "lu"]
+        "include": ["fr", "lu"]
       },
       "tags": {
         "amenity": "fast_food",
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -263,7 +263,6 @@
         "brand": "Bagel Corner",
         "brand:wikidata": "Q104649272",
         "cuisine": "bagel",
-        "name": "Bagel Corner",
         "takeaway": "yes"
       }
     },
@@ -367,17 +366,14 @@
     },
     {
       "displayName": "Big Fernand",
-      "id": "bigfernand-8ddd61",
-      "locationSet": {
-        "include": ["fr", "lu"]
-      },
+      "id": "bigfernand-fd163a",
+      "locationSet": {"include": ["fr", "lu"]},
       "tags": {
         "amenity": "fast_food",
         "brand": "Big Fernand",
         "brand:wikidata": "Q19521346",
         "brand:wikipedia": "fr:Big Fernand",
         "cuisine": "burger",
-        "name": "Big Fernand",
         "takeaway": "yes"
       }
     },
@@ -1792,7 +1788,6 @@
         "brand:wikipedia": "fr:Exki",
         "cuisine": "healthy;sandwich;salad;pie",
         "diet:vegetarian": "yes",
-        "name": "Exki",
         "takeaway": "yes"
       }
     },
@@ -2002,8 +1997,6 @@
         "brand:wikidata": "Q1131810",
         "brand:wikipedia": "en:Five Guys",
         "cuisine": "burger",
-        "name": "Five Guys",
-        "official_name": "Five Guys Burgers and Fries",
         "takeaway": "yes"
       }
     },
@@ -2017,7 +2010,6 @@
         "brand:wikidata": "Q67168925",
         "cuisine": "burger",
         "diet:vegan": "only",
-        "name": "Flower Burger",
         "takeaway": "yes"
       }
     },
@@ -3194,16 +3186,13 @@
         "brand:wikipedia": "fr:La Boîte à Pizza",
         "cuisine": "pizza",
         "delivery": "yes",
-        "name": "La Boîte à Pizza",
         "takeaway": "yes"
       }
     },
     {
       "displayName": "La Croissanterie",
-      "id": "lacroissanterie-affad8",
-      "locationSet": {
-        "include": ["es", "fr", "gb"]
-      },
+      "id": "lacroissanterie-7eae30",
+      "locationSet": {"include": ["es", "gb"]},
       "tags": {
         "amenity": "fast_food",
         "brand": "La Croissanterie",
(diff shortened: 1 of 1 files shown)


kjonosm:
You can add `"preserveTags": ["^name"]` to these entries, so iD won't suggest changing the name tag, see e.g. [here](https://github.com/osmlab/name-suggestion-index/blob/171446b1f51495cdf64214405499e70d8f30e711/data/brands/shop/car_parts.json#L83)

gileri:
> You can add `"preserveTags": ["^name"]` to these entries, so iD won't suggest changing the name tag, see e.g. [here](https://github.com/osmlab/name-suggestion-index/blob/171446b1f51495cdf64214405499e70d8f30e711/data/brands/shop/car_parts.json#L83)

Thank you for your answer @kjonosm.
The changes I proposed are two folds :

- Do not suggest to replace a more complete name by just the brand (that would be covered by `preserveTags` as you point out)
- Avoid suggesting an incomplete name on OSM objects without a name already (that is not covered by `preserveTags`)

Just like for example McDoanld's are usually at least open `Mo-Sa`, we don't suggest `opening_hours=Mo-Sa` on OSM entities. Even though it's somewhat better than not including any `opening_hours`, that's incomplete and will hinder the precise filling of this value.

The same could be said about the website. Having the brand website on `website=*` is better than no website at all, but what's best is having the website of the branch. Suggesting putting the brand website when we know a dedicated branch website/page exists seems counterproductive.

arch0345:
I agree with @kjonosm on this one: I believe it's better to have a name than no name at all for these fast food brands. For Five Guys, although the website shows they have the branch in their name, they're usually signed as just "Five Guys". If the French community prefers local names like the Norwegian/Swedish community, these brands should probably get their own entry restricted to just France.

gileri:
Thank you for your answer too @arch0345 !

>I believe it's better to have a name than no name at all for these fast food brands

Isn't that why brand=* exist ? Any consumer (e.g. like OSMAnd, JOSM do) can use it in lieu of name=* if it's not available. But suggesting/forcing to put an incomplete/incorrect name just to put a name seems counter-productive for the OSM database.
It's up to the consumers to choose what to do with the data. And it's a lot easier if the data is correct.

>For Five Guys, although the website shows they have the branch in their name, they're usually signed as just "Five Guys"

If the shop chooses to use a specific name in its communication with patrons, why should we (OSM contributors) choose otherwise ?
I skipped some entries on that file because certain shops do not seem to use specific names for each branch. In that case, that's totally correct to put the brand as the name, because that's the ground (and online) reality.

tas50:
Five guys really should not be mapped with the location name. That's now how they're signed and now how either Google Maps or Apple Maps shows it. That's just branch naming on their website. 

image
image

bhousel:
Oh yeah, I'm just seeing this now, @gileri please don't go deleting the `name` tags from all the items - I can't accept this PR as is.

We should publish some better guidance about how this project handles names, but the tl;dr today is:

- **If the brand really does use completely different names for each location**, then it's ok to not have a `name` tag.  This is kind of unusual, but we've seen it for pub-like brands like Wetherspoon (#3009) and Harverster (#4034)
- **If the brand uses a unique name for each branch location and that name includes the brand, and that name is significant to how people refer to the thing**, we add the `name` tag so that a user adding the POI initially will have it filled in, and we include `preserveTags` property so that iD will not try to change the `name` tag later.  These are also exceptions, but somewhat common.  Apple stores, Amazon lockers, cinemas, car dealerships, hotels, and many other categories are set up this way (#4906 has more on this).  Scandinavian community wants their POIs mapped this way (#5500).
- **Otherwise the brand doesn't really talk about each branch like it's a unique thing and people don't really think of it this way**, and so we just set the `name` tag, and people can specify the branch name in the `branch` tag, and iD will try to nudge people towards this style of mapping.  The vast majority of brands are like this.  (Again see #4906 and #5500 for where we landed with this - this was the recommended style of tagging going back a while).

gileri:
I thought that bringing accurate data was a goal of name-suggestion-index, that quality was more important than quantity.

Thank you for clarifying the position of the project. I'll guess I'll have to continue fixing incorrect names instead of making more meaningful contributions to OSM.

gileri:
>Five guys really should not be mapped with the location name. That's now how they're signed and now how either Google Maps or Apple Maps shows it. That's just branch naming on their website.

Thank you for sharing this data @tas50. As you can see by the 7 links I shared, the restaurants have a unique, public-facing name. I think the information on the franchise-owned websites is more authoritative on the subject than Google Maps or Apple Maps.

tas50:
I'm not going strictly off Apple Maps or Google maps although both of those are very useful since the brands themselves generally choose how to be represented there. I'm going off personal experience going to 3 different Five Guys near me and 1 in another state. There's no mention of which store it is anywhere. They're just Five Guys.

PR #6067, opened 2022-01-14
Update furniture.json

dan13l9:
update the list of countries in which the XXXLutz / Momax group has stores

PR #6072, opened 2022-01-15
Update furniture.json

dan13l9:
Updated country list where XXXLutz / Momax operating.
Mömax is Austrian. The wikipedia article in hungarian is limited and refers only to local subsidiary.


arch0345:
Duplicate of #6067

arch0345:
We run a script before each release that adds the Wikipedia article linked on the Wikidata page, and unfortunately there's only a Hungarian Wikipedia article for Mömax currently as the German one redirects to XXXLutz.

PR #6087, opened 2022-01-16
Update pharmacy.json

dan13l9:
update country list where Dr. Max pharmacy are according to https://www.drmax.eu/


UKChris-osm:
Thank you @dan13l9 👍

PR #6091, opened 2022-01-17
Agent Provocateur

codeinabox:
Agent Provocateur has stores across the world https://www.agentprovocateur.com/store-finder

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -229,10 +229,11 @@
     },
     {
       "displayName": "Agent Provocateur",
-      "locationSet": {"include": []},
+      "locationSet": {"include": ["001"]},
       "tags": {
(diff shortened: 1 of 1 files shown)


codeinabox:
Keen for some feedback on what to set the `locationSet` to given it's a lot of countries (see https://www.agentprovocateur.com/store-finder) and also which Wikipedia page(s) to include in the tags.

UKChris-osm:
I would probably just set it to world "_001_" if you're not sure exactly which Countries they trade in, and select the English Wikipedia. It's not a bad thing if places like this are suggested in Countries in which they don't trade (as this is just a suggestion index) unless there are similar named brands that may clash or be confused with one another.

arch0345:
Thanks @codeinabox

PR #6102, opened 2022-01-20
Hackney Council operates recycling, markets and parks

codeinabox:


UKChris-osm:
For a borough within London, I'd be include to suggest using a GPS pin-point with a small radius rather than gb-lon.geojson, for example, Hackney could be [{"include":[[-0.06,51.55,4]]}](https://location-conflation.com/?locationSet=%7B%22include%22%3A%5B%5B-0.06%2C51.55%2C4%5D%5D%7D&referrer=nsi) which sets a 4km radius at the two lat/long GPS points.

I may eventually get around to creating .geojson's for London, but only when there is a higher use for them

PR #6104, opened 2022-01-21
Flip gas station brand to their EVC brand

tohaklim:
Hi all, not sure if this is the best approach, so a PR instead of direct edit.
Most of these EV charging stations are currently mistagged with the incorrect brand (Belorusneft/Белоруснефть), which is the gas station brand of the same operator.
I am keeping them as the operator, and changing the brand in the entry - would this be sufficient to match (on the operator field), and discard Белоруснефть as charging_station brands in the future?
Thanks


arch0345:
I think this should be moved to the brands tree (`data/brands/amenity/charging_station.json`) to show the brand logo for the preset instead of the operator logo

PR #6105, opened 2022-01-21
The Breakfast Club

codeinabox:
The Breakfast Club has locations in London, Oxford, Brighton & Chelmsford

https://thebreakfastclubcafes.com/


codeinabox:
@UKChris-osm I wasn't certain if I should go with restaurant or cafe. It has been mostly mapped as a restaurant and has a more extensive menu than a cafe as you can get full meals.

arch0345:
Thanks @codeinabox, yeah other breakfast diners like IHOP are under `amenity/restaurant` so I think this is fine

PR #6108, opened 2022-01-22
upgrade to id-tagging-schema version 3.2.*

tyrasd:
This affects for example the new `amenity=parcel_locker` brands which currently fall back to use the generic "amenity" preset. See https://github.com/osmlab/name-suggestion-index/pull/6020#pullrequestreview-846237942


arch0345:
Thanks @tyrasd

PR #6109, opened 2022-01-22
Update post_office.json

Mannivu:
After discussing it with the italian community, we agreed that the most common name of the italian post operator has the Capital letter, so "Poste Italiane" instead of "Poste italiane"


arch0345:
Thanks @Mannivu

PR #6110, opened 2022-01-22
Don't change Econt parcel locker's name

Dimitar5555:
After some investigation, I've found out that all of them have specific names which should be preserved.


arch0345:
If this is something similar to how Amazon's parcel lockers are named I would add this instead of removing the names:
`"preserveTags": ["^name"],`
image

Dimitar5555:
@arch0345 It can be merged now.

arch0345:
Thanks @Dimitar5555

PR #6113, opened 2022-01-22
Rename BKK charging station to Eviny

mathiash98:
BKK AS was split into multiple companies where the company for responsible charging stations was renamed to Eviny in autumn 2021. So all physical chargers have now been rebranded.

https://www.eviny.no/ladestasjoner

<strike>When I run `npm run build` BKK is added again in the charging stations file 🤔 </strike>
- Fixed by adding `matchNames: ["bkk"]`
- And lots of other unrelated changes occur

So I have not run `build` in my PR

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -319,21 +319,6 @@
       }
     },
     {
-      "displayName": "BKK",
-      "id": "bkk-476d58",
-      "locationSet": {"include": ["no"]},
-      "tags": {
-        "amenity": "charging_station",
-        "brand": "BKK",
-        "brand:wikidata": "Q4891601",
-        "brand:wikipedia": "no:Eviny",
-        "name": "BKK",
-        "operator": "BKK",
-        "operator:wikidata": "Q4891601",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Hi @mathiash98 

[BKK](https://github.com/osmlab/name-suggestion-index/blob/6e3eeb50e800f222f72b8773a2ba04f238bfb752/dist/filtered/operators_keep.json#L127) will continue to be added back to the index as it appears over 50 times in OpenStreetMap.

To counter this, you can add a "_matchNames_" to Eviny, which will tell our scripts that "BKK is the same as Eviny". You can do this by adding:

"locationSet": {"include": ["no"]},
**"matchNames": ["bkk"],**
"tags": {

... under the "_locationSet_" but above the "_tags_".

mathiash98:
> Hi @mathiash98
>
> To counter this, you can add a "_matchNames_" to Eviny, which will tell our scripts that "BKK is the same as Eviny". You can do this by adding:


Ah of course! Thank you, updated the commit to contain `matchNames`

arch0345:
Thanks @mathiash98

PR #6116, opened 2022-01-23
add Amscot

RayBB:
Amscot provides financial services in Florida and has hundreds of locations.

The Wikipedia says the company is "Amscot Financial" but I've never seen it advertised that way it usually says "Amscot - The Money Superstore". So I think just writing it as Amscot is fine.

This is my first addition to this repo and one thing that wasn't clear from the guide is how the "id" field was set so I just tried to follow the other patterns.

Cheers

### DIFF
--- a/data/brands/shop/money_lender.json
+++ b/data/brands/shop/money_lender.json
@@ -29,6 +29,18 @@
       }
     },
     {
+      "displayName": "Amscot",
+      "id": "amscot-f9c954",
+      "locationSet": {"include": ["us"]},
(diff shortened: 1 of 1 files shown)


arch0345:
Thanks @RayBB and welcome! Yeah I was a bit confused about the `id` too when I first started contributing to this project. We have a script that adds the `id` automatically, so you don't have to add an `id` when creating a new entry. I've updated the Contributing wiki page to (hopefully) make it more clear.

PR #6123, opened 2022-01-26
Longo's

andrewpmk:
Add Longo's grocery store https://www.longos.com/

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -3836,13 +3836,14 @@
     },
     {
       "displayName": "Longo's",
-      "locationSet": {"include: ["ca"]},
+      "id": "longos-76454b",
+      "locationSet": {"include": ["ca"]},
       "tags": {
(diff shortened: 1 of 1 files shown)


data/brands/shop/supermarket.json
```diff
@@ -3834,6 +3834,18 @@
         "shop": "supermarket"
       }
     },
+    {
+      "displayName": "Longo's",
+      "id": "longos-76454b",
+      "locationSet": {"include": ["ca"]},
```
arch0345, line 3840: It [looks like](https://www.longos.com/store-finder)  they only have locations in Ontario so I would refine the locationSet to only include that province
```suggestion
      "locationSet": {"include": ["ca-on.geojson"]},
```

arch0345 (changes requested):
Thanks @andrewpmk! I've added a suggestion to only include Ontario since it looks like they currently only have locations there

arch0345 (approved):

PR #6130, opened 2022-01-27
Add Rindchen's Weinkontor

AlertSubject:
This PR adds a wine store chain in Germany.
Names change depending on location. Therefore, they are not included in this preset.


m-hue:
Thanks @AlertSubject!

PR #6135, opened 2022-01-29
Rabba

andrewpmk:
Add Rabba https://rabba.com/. Not sure if we should use "Rabba" or "Rabba Fine Foods" for the name.
Note there isn't a Wikipedia page for this company right now. They have 34 locations in Greater Toronto Area.


UKChris-osm:
I think "_Rabba_" would be okay a the name as that seems to be the focus on their web site and some store photos, but you would add a matchNames for "_rabba fine foods_".

andrewpmk:
Added matchNames for "rabba fine foods".

arch0345:
Thanks @andrewpmk

PR #6144, opened 2022-02-01
Added University of New Orleans, Rock'n'Bowl, and City Bark to Data.

GrandmaCoding:
Added University of New Orleans to amenity/university.json
Added Rock'n'Bowl to leisure/bowling_alley.json
Added City Bark to leisure/dog_park.json

### DIFF
--- a/data/brands/amenity/university.json
+++ b/data/brands/amenity/university.json
@@ -33,6 +33,18 @@
       }
     },
     {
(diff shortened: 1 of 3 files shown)


bhousel:
Sorry @GrandmaCoding , this index is really for tracking the sorts of features where there are a lot of similarly tagged items (like chain restaurants), we don't add items for single things.

PR #6145, opened 2022-02-01
Added Izzo's Illegal Burrito to data/brands/amenity/fast_food.json

GrandmaCoding:
Izzo's Illegal Burrito is a fast food chain in southeastern Louisiana.
https://www.izzos.com/
I didn't understand the assignment before, but I *think* I do now!

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -2785,6 +2785,18 @@
       }
     },
     {
+      "displayName": "Izzo's Illegal Burrito",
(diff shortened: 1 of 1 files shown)


arch0345:
Thanks @GrandmaCoding and welcome! Note that without a Wikidata page link, this won't get generated as a preset. You can easily [create a Wikidata page](https://github.com/osmlab/name-suggestion-index/wiki/Editing-Wikidata) for this brand if you're interested.

PR #6156, opened 2022-02-03
PreserveTags name for some more brands in Norway

mathiash98:
As per issue [#5500](https://github.com/osmlab/name-suggestion-index/issues/5500) we would like to preserve name for brands in Scandinavian countries.

Pizzabakeren:
- Also changed wikipedia to the norwegian version (danish, swedish and norwegian is readable for all countries). As it contains tiny bit more info and the brand is originally Norwegian.

Intersport:
- Created a new for entry for Norway

McDonald's:
- Created a new entry for Norway similar to france's entry, but with english wikipedia

Handelsbanken:
- Checked against handelsbanken's websites and cross referenced to how handelsbanken is displayed outside of norway as well. And the largest majority of bank locations have name: "Handelsbanken [Location]" -> Therefor included preserveTags name for all countries.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -5484,6 +5484,7 @@
       "locationSet": {
         "include": ["dk", "fi", "gb", "no", "se"]
       },
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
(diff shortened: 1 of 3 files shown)


arch0345:
Thanks @mathiash98

PR #6165, opened 2022-02-04
Add El Rinconsito, a Washington chain

OmnificOne:


arch0345 (changes requested):
Thanks for adding this! Note we [don't support](https://github.com/osmlab/name-suggestion-index/wiki/Contributing#tags-required) `website` tags in NSI.

OmnificOne:
Thanks, hope this does it!

arch0345 (approved):
Looks good :+1:

Note that without a linked Wikidata item this won't get generated as a preset in iD

PR #6176, opened 2022-02-08
Update parcel_locker.json (update InPost parcel_locker tagging)

mutipg:
According to the wiki: InPost name attribute value should be tagged as brand and name should be removed.


arch0345:
Thanks @mutipg and welcome! The example in the wiki article should probably get updated as it shows the generic 'parcel locker' Wikidata item under `brand:wikidata`/`brand:wikipedia`

PR #6177, opened 2022-02-08
Merge all the BTs

peternewman:


data/operators/amenity/telephone.json
```diff
@@ -180,18 +160,14 @@
...
-    },
-    {
-      "displayName": "Deutsche Telekom AG",
-      "id": "deutschetelekomag-2be02b",
-      "locationSet": {"include": ["001"]},
-      "tags": {
-        "amenity": "telephone",
-        "operator": "Deutsche Telekom AG"
+        "operator": "Deutsche Telekom",
+        "operator:wikidata": "Q9396",
```
peternewman, line 169: Assuming this shouldn't be Q2401704 / Telekom Deutschland?

peternewman: Edit, I've found them further down the file, so presumably not...

config/genericWords.json
```diff
@@ -3,7 +3,7 @@
     "^(barn|bazaa?r|bench|bou?tique|building|casa|church)$",
     "^(baseball|basketball|football|soccer|softball|tennis(halle)?)\\s?(field|court)?$",
     "^(club|green|out|ware)\\s?house$",
-    "^(driveway|el árbol|fountain|golf|government|graveyard)$",
+    "^(driveway|el árbol|fountain|generic|golf|government|graveyard)$",
```
peternewman, line 6: Will this be enough to exclude these "operators"?:
https://www.openstreetmap.org/node/1892277713
https://www.openstreetmap.org/node/878266316

peternewman: I was more unclear if that also covered the operator tag or just names.

arch0345:
Looks good to me

Lukas458:
Hi,
may I mention something here? 
The value "Telekom Deutschland" and "Telekom Deutschland (GmbH)" has sometimes been set for some amenity=telephone's in Germany. But that's not correct, as Telekom Deutschland is only a special subdivision of "Deutsche Telekom". _Only_ Deutsche Telekom operates some public phones in Germany, but "Telekom Deutschland" not. So may I delete this section in the nsi?

Also, "Deutsche Telekom" does not operate any _emergency=phone_ in Germany. So we would not need a preset fo emergency=phone with "Deutsche Telekom", too, because this only has the risk of confusion with amenity=telephone.
So, would it be ok to remove this entry in the corresponding preset?

Lukas458:
Also, the operator values "T-Com", "Telekom" and "Telekom AG" are often used as (not-standard-conform) synonymes for the "Deutsche Telekom" operator. "T-Com" was a former name of Deutsche Telekom.

Lukas458:
See #6199

PR #6186, opened 2022-02-13
Choose a more specific iD preset for govt offices

arch0345:
Also added warning if `office=government` is missing `government=*` or the value contains odd punction/spacing

Edit: removed warning for missing `government=*` tag since I noticed some operate more than one type of govt office


bhousel:
Thanks @arch0345 , this is a good idea!

PR #6200, opened 2022-02-17
Update bank.json

Mannivu:
Removed matchname from BCC Roma: in Italy there are quite a lot of BCC and only a few are part of BBC Roma brand. Since I don't think there's the ability to limit for just a region but the NSI only works on national scale I think it's better to delete the matchname

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -2544,7 +2544,6 @@
       "displayName": "BCC Roma",
       "id": "bccroma-7b36b5",
       "locationSet": {"include": ["it"]},
-      "matchNames": ["bcc"],
       "tags": {
         "amenity": "bank",
         "brand": "BCC Roma",


UKChris-osm:
Hi @Mannivu, you can restrict an entry to a geozone if you like, using a [feature](https://github.com/osmlab/name-suggestion-index/tree/main/features), which is a "_.geojson_" file that can be referenced in place of (or as well as) a two letter country code.

You can draw out an area using [geojson.io](https://geojson.io/#map=6/42.220/11.865) and save the content as a "_.geojson_" file within the [feature](https://github.com/osmlab/name-suggestion-index/tree/main/features) folder. In the example of Italy, you might want to create the folder **it** and create the file "_it-rm.geojson_" to represent Rome as ISO_3166-2:IT.

Mannivu:
Ah, this is great! I didn't know about this possibility. I'll close this PR and try to create the _.geojson_. Thank you again!

PR #6204, opened 2022-02-19
add "Tobaccoland" cigarette vending machine

Lukas458:
Tobaccoland is a very common operator of cigarette vending machines in Germany. The wikidata are here: https://www.wikidata.org/wiki/Q1439872


arch0345:
Thanks @Lukas458

PR #6207, opened 2022-02-20
"Deutsche Telekom" doesn't operate any emergency=phone.

Lukas458:
The German company "Deutsche Telekom" does not operate any emergency=phone. Those emergency=phone's which are tagged with this operator (according to taginfo there exist 62 ones with this tagging) are **misinterpreted** as emergency-phones, but they are normal public telephones (amenity=telephone). So there shouldn't be an operator-ing in the nsi for this. See also my comment in https://github.com/osmlab/name-suggestion-index/pull/6177#issuecomment-1042733139

### DIFF
--- a/data/operators/emergency/phone.json
+++ b/data/operators/emergency/phone.json
@@ -240,18 +240,6 @@
       }
     },
     {
-      "displayName": "Deutsche Telekom AG",
-      "id": "deutschetelekomag-3876cc",
-      "locationSet": {"include": ["de"]},
-      "matchNames": ["deutsche telekom"],
-      "tags": {
-        "emergency": "phone",
-        "operator": "Deutsche Telekom AG",
-        "operator:wikidata": "Q9396",
-        "operator:wikipedia": "de:Deutsche Telekom"
-      }
-    },
-    {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Thank you @Lukas458 

However, because this entry comes directly from OSM data (i.e. there are more than 50 objects tagged as operator=Deutsche Telekom) then simply deleting it won't work, as it will be re-added back automatically.

To get around this, if you scroll to the top of the emergency/phone.json file, you should see a section called "_exclude_" and within this section, you should see "_named_". If you could, could you add "_Deutsche Telekom_" to this exclusion list in your PR, such as change:

"exclude": {
  "generic": ["^phone$"],
  "named": ["^straßenmeisterei$"]
}

.. to ..

"exclude": {
  "generic": ["^phone$"],
  "named": [**"^deutsche telekom$",**"^straßenmeisterei$"]
}

... and this will tell our scripts that even though "_Deutsche Telekom_" exists a lot in OSM data, that we are to ignore it.

Lukas458:
Oh, I see. Thanks for your quiick answer, I did that now.

UKChris-osm:
Thanks for doing that @Lukas458 👍

PR #6215, opened 2022-02-22
Add Nero Express

peternewman:
E.g. https://caffenero.com/uk/store/kings-cross-central-541/

Not entirely sure on the best way to handle these, but like Tesco they are badged differently at least, so we're loosing some info if we just matchNames them.

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -590,6 +590,21 @@
       }
     },
     {
+      "displayName": "Nero Express (Caffè Nero)",
+      "locationSet": {
+        "include": ["gb", "ie", "tr", "us"]
+      },
+      "tags": {
+        "amenity": "cafe",
+        "brand": "Caffè Nero",
+        "brand:wikidata": "Q675808",
+        "brand:wikipedia": "en:Caffè Nero",
+        "cuisine": "coffee_shop",
+        "name": "Nero Express",
+        "takeaway": "yes"
+      }
+    },
+    {
       "displayName": "Caffeine",
       "id": "caffeine-f2627b",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/cafe.json
```diff
@@ -589,6 +589,21 @@
...
+    {
+      "displayName": "Nero Express (Caffè Nero)",
+      "locationSet": {
+        "include": ["gb", "ie", "tr", "us"]
+      },
+      "tags": {
+        "amenity": "cafe",
+        "brand": "Caffè Nero",
+        "brand:wikidata": "Q675808",
+        "brand:wikipedia": "en:Caffè Nero",
```
peternewman, line 601: Does this need a discrete Wikidata/Wikipedia entry from the original/bigger one (if so the CI should fail too)?

peternewman:
OSM user harg confirmed my vague memory that at least that one is takeaway=only, I wonder if the others are too and that's the distinction in branding?
https://www.openstreetmap.org/note/2877350

data/brands/amenity/cafe.json
arch0345: The branding for Caffè Nero and Nero Express looks similar so it should be fine to use the same Wikidata/Wikipedia entry, it's not required to have a unique Wikidata item for each entry in NSI

peternewman:
> OSM user harg confirmed my vague memory that at least that one is takeaway=only, I wonder if the others are too and that's the distinction in branding? https://www.openstreetmap.org/note/2877350

This seems to be the exception:
https://www.kensingtonarcade.co.uk/stores/cafe-nero-ex/

But could add `takeaway=yes` although that may be spam as it's probably true of most coffee shops?

PR #6218, opened 2022-02-23
Add Cumberland Building Society

stuartm:
Add Cumberland Building Society (Bank) which operates mostly in the North West of England


arch0345:
Thanks @stuartm and welcome!

PR #6224, opened 2022-02-25
Butchies

codeinabox:
Butchies is a London fried chicken chain and has plant based options on the menu

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -826,7 +826,6 @@
         "brand": "Butchies",
         "brand:wikidata": "Q111015134",
         "cuisine": "chicken",
-        "diet:vegan": "yes",
         "name": "Butchies",
         "takeaway": "yes"
       }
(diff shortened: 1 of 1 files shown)


data/brands/amenity/fast_food.json
```diff
@@ -816,6 +816,21 @@
...
+      "displayName": "Butchies",
+      "locationSet": {
+        "include": ["gb-lon.geojson"]
+      },
+      "tags": {
+        "amenity": "fast_food",
+        "brand": "Butchies",
+        "brand:wikidata": "Q111015134",
+        "cuisine": "chicken",
+        "diet:vegan": "yes",
```
LaoshuBaby, line 829:
```suggestion
```
Sorry, I don't think this time like  #5801 (doughnuts said that vegan is a key point, but this time it is a **chicken** fast food)

So I don't think this tag is needed. Maybe the case is like #5727

LaoshuBaby (changes requested):

data/brands/amenity/fast_food.json
codeinabox: Fair play, I will remove the tag in this case.

PR #6228, opened 2022-02-26
Update pharmacy.json

### LINKED ISSUE #6226: Remove Unichem?
This is regarding _Unichem_ (with location `001`), not _Unichem Pharmacy_ (with location `nz`).
https://github.com/osmlab/name-suggestion-index/blob/2f5213851a595efc684c897c8160c82e8e190e41/data/brands/amenity/pharmacy.json#L2303

From Wikipedia, it looks like it was a former name of [Boots/Alliance Boots](https://en.wikipedia.org/wiki/Alliance_Boots), which merged with [Walgreens](https://en.wikipedia.org/wiki/Walgreens_Boots_Alliance).

There are no [locations named "Unichem"](https://overpass-turbo.eu/s/1grS) apart from those found in New Zealand (which are different).

I'm assuming this is no longer required. Should it be removed?

Ash-KODES:
Fixes #6226

### DIFF
--- a/data/brands/amenity/pharmacy.json
+++ b/data/brands/amenity/pharmacy.json
@@ -2300,17 +2300,6 @@
       }
     },
     {
-      "displayName": "Unichem",
-      "id": "unichem-f5140f",
-      "locationSet": {"include": ["001"]},
-      "tags": {
-        "amenity": "pharmacy",
-        "brand": "Unichem",
-        "healthcare": "pharmacy",
-        "name": "Unichem"
-      }
-    },
-    {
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
> From Wikipedia, it looks like it was a former name of [Boots/Alliance Boots](https://en.wikipedia.org/wiki/Alliance_Boots), which merged with [Walgreens](https://en.wikipedia.org/wiki/Walgreens_Boots_Alliance).

If possible you can add a matchNames to Walgreens?

PR #6229, opened 2022-02-26
Remove period from Drakes name

TownCube:
I wasn't sure on the policy on styling of logos and how that translates to the name fields. In this case the full stop isn't used as an abbreviation and while one part of the logo includes a full stop but another part of the logo and [the website text](https://drakes.com.au/about/) refer to the company as "Drakes Supermarkets" without the period.

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -1861,14 +1861,14 @@
       }
     },
     {
-      "displayName": "Drakes.",
+      "displayName": "Drakes",
       "id": "drakes-c254c9",
       "locationSet": {"include": ["au"]},
       "tags": {
-        "brand": "Drakes.",
+        "brand": "Drakes",
         "brand:wikidata": "Q48988077",
         "brand:wikipedia": "en:Drakes Supermarkets",
-        "name": "Drakes.",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Hello! Welcome to NSI and you've open your first PR! Thank you for your interest.

NSI is suggestion for OSM mapping so in my opinion it should follow the **On The Ground** rule.

https://wiki.openstreetmap.org/wiki/Good_practice#Map_what.27s_on_the_ground

1ec5:
Even so, there are limits to how far we’d go in mimicking stylized signage. There’s a restaurant chain in the U.S. whose signs just say “🌶’s“, but we still tag `name=Chili's`. There’s also an American bank whose logo looks like “usbank”, but we tag it `name=U.S. Bank`.

Essentially, the distinction we’re making is that `name` reflects how the brand name would be written down on paper. This can sometimes differ slightly from what’s on the sign. Consumers of NSI can use NSI’s Wikidata and social media integrations to display the actual logo if that level of fidelity is desired. Hope this helps you decide on an approach for Drakes.

bhousel:
Judging from their website copy, it seems safe to remove the "." from the name.

PR #6232, opened 2022-02-27
New Zealand kindergartens

Mortein:
Add:
- _Kidsfirst_
- _Kindercare_

### DIFF
--- a/data/operators/amenity/kindergarten.json
+++ b/data/operators/amenity/kindergarten.json
@@ -396,6 +396,17 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Thanks! @Mortein

bhousel:
I ended up needing to move these from the `operator` tree (which is more for publicly operated kindergartens)  to the `brand` tree (which is more for privately operated kindergartens) in 9de9d0257
This was because the matcher raised a conflict between the US KinderCare and the NZ Kindercare, so they really needed to go to the same tree.

PR #6233, opened 2022-02-27
Add Germanos (GR)

### LINKED ISSUE #6230: Add brand: Germanos mobile phone stores in Greece
Hi!

I would like to request the addition of Germanos mobile phone chain store in Greece.

shop:mobile_phone
name:Γερμανός
name:el-Latn=Germanos
brand:Γερμανός
brand:wikidata:Q9267542
brand:wikipedia=el:Γερμανός (αλυσίδα καταστημάτων)
website=https://www.germanos.gr
country codes:gr

Thanks!

Mortein:
Closes #6230.
Added additional countries from [Wikipedia (en) article](https://en.wikipedia.org/wiki/Germanos_(chain_of_stores)).

### DIFF
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -1105,18 +1105,8 @@
     },
     {
       "displayName": "Γερμανός",
-      "id": "38b932-e674cf",
-      "locationSet": {
-        "include": [
-          "bg",
-          "cy",
-          "gr",
-          "mk",
-          "pl",
-          "ro",
-          "ua"
-        ]
-      },
+      "id": "38b932-602a3b",
+      "locationSet": {"include": ["cy", "gr"]},
       "tags": {
         "brand": "Γερμανός",
         "brand:wikidata": "Q9267542",
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -1104,6 +1104,29 @@
       }
     },
     {
+      "displayName": "Γερμανός",
+      "id": "38b932-e674cf",
+      "locationSet": {
+        "include": [
+          "bg",
+          "cy",
+          "gr",
+          "mk",
+          "pl",
+          "ro",
+          "ua"
+        ]
+      },
+      "tags": {
+        "brand": "Γερμανός",
+        "brand:wikidata": "Q9267542",
+        "brand:wikipedia": "el:Γερμανός (αλυσίδα καταστημάτων)",
+        "name": "Γερμανός",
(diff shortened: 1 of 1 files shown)


sucrerouge123:
The name=Γερμανός is in Greek. I think this will be a problem when adding this brand to other countries except Greece and Cyprus. Maybe we should add:

brand:bg=Germanos
brand:mk=Germanos
brand:pl=Germanos
brand:ro=Germanos
brand:ua=Germanos

and some other tags of name:bg=... etc

What do you think?

bhousel:
> What do you think?

Adding a whole lot of international name tags is not normally how we create entries in this index.  This would lead to local communities protesting against all the extra tags being added.

A more standard way would be to
- create one item with `name` = "Germanos" scoped to places where the brand is called that, excluding Greece
- create another item for `name` = "Γερμανός" and `name:en` = "Germanos",  just for Greece
- (can create more items if it it is named other things in other places)

sucrerouge123:
Understood. Please add only the country codes "gr" and "cy" here. Don't know if in bg or mk it's called Germanos or it is spelled in the corresponding alphabet too. Thanks!

Mortein:
@sucrerouge123 I've changed the countries.

As per [Names#Localization](https://wiki.openstreetmap.org/wiki/Names#Localization), we should include the Greek name in both `name` and `name:el`, and also avoid transliteration. Is _Germanos_ a valid English name for the store for `name:en`?

If so, it'll be:
- `name`: `Γερμανός`
- `name:el`: `Γερμανός`
- `name:en`: `Germanos`

sucrerouge123:
About the name:el I agree. But, "Germanos" is actually the transliteration and so it is used in all Latin script languages, because it is a brand name. I don't know why we should prefer name:en=Germanos and not name:fr or name:it. That's why I put name:el-Latn.

bhousel:
> Is Germanos a valid English name for the store for name:en?

Yes their website is  https://www.germanos.gr/  and at the bottom of it I see this:

Screen Shot 2022-02-28 at 7 35 57 AM

So I think this name is fine for English use.

sucrerouge123:
Of course it is ok for usage in English, but it is also good for usage in Italian, in Spanish or in German. Why should we only tag name:en? Isn't this a bit arbitrary? My opinion is that either we should tag it with name:el-Latn, which is transliteration, or we shouldn't tag it at all as it is mentioned in the "Avoid transliteration" section. Thanks!

Dimitar5555:
Germanos does not exist in Bulgaria since 2014.

bhousel:
This is fine

PR #6238, opened 2022-03-01
Add bakery Campaillette

PiOverFour:
A common brand of bakeries in France.


arch0345:
Thanks @PiOverFour and welcome!

PR #6241, opened 2022-03-02
Horizon Parking

codeinabox:


data/operators/amenity/parking.json
```diff
@@ -248,6 +248,16 @@
         "operator:wikidata": "Q73072632"
       }
     },
+    {
+      "displayName": "Horizon Parking",
+      "locationSet": {"include": ["gb"]},
```
LaoshuBaby, line 253:
```suggestion
      "locationSet": {"include": ['gb-eng', 'gb-wls', 'gb-sct', 'gb-nir']},
```

`gb` is very big, it contain all oversea region of Great Britain, `gb-eng`, `gb-nir`, `gb-sct`, `gb-wls`, `gb-ukm`(this is not supported in location-confliction so I haven't use) are more accurate.

You can view them in [ideditor.codes](https://ideditor.codes) or test on playground [location-conflation.com](https://location-conflation.com/?locationSet=%7B%20include%3A%20%5B%27gb-eng%27%2C%20%27gb-wls%27%2C%20%27gb-sct%27%2C%20%27gb-nir%27%5D%20%7D%0A%0A)

codeinabox: According to their company website they are "the UK’s fastest growing parking management company", so how would I restrict this just to the United Kingdom?

LaoshuBaby: My suggestion is include `'gb-eng', 'gb-wls', 'gb-sct', 'gb-nir'` those 4 region. （I just tested `gb-ukm` is not supported in NSI software, it only raised in [ideditor.codes](https://ideditor.codes/)）

图片

_Maybe I need to ask local expert @UKChris-osm , sorry for my lacking of GB's knowledge :(_

UKChris-osm:
Does go-ukm really not work? I’m surprised at that.

To be honest using gb alone would have been okay from my perspective because having Horizon Parking available in a few areas it doesn’t operate would only be a problem if another operator shared the name in those areas.

arch0345:
> Does gb-ukm really not work? I’m surprised at that.

It looks like [ideditor/country-coder](https://github.com/ideditor/country-coder/) was last released in June of last year and the commit was from November, so we should be able to use once the next release comes around

PR #6245, opened 2022-03-03
Modify sports shop information for Hong Kong

tommylung:
Modify shop information
- Marathon Sports (Hong Kong)

Add shop information
- GigaSports


LaoshuBaby:
Good! Thank you for commit this, I'm also mixed by those 2 brand named marathon, I just wrong think this is a Ecuador brand utill fully research! 👍

PR #6252, opened 2022-03-03
Nine Hours capsule hotels

codeinabox:


codeinabox:
I am not sure whether to treat this as a hotel or hostel as you do get your own tiny capsule however it's more like an upmarket hostel

bhousel:
> I am not sure whether to treat this as a hotel or hostel as you do get your own tiny capsule however it's more like an upmarket hostel

Interesting!  I also could see it as either `hotel` or `hostel`..  I'll ask on Slack what people there think.

Kovoschiz:
Yes https://wiki.openstreetmap.org/wiki/JA:How_to_map_a#カ made カプセルホテル as `tourism=hostel`. Taginfo shows only `description:*=` being used, nothing specific.  
There are a few `hostel=` out there. If you want, you may do something eg `hostel=capsule_hotel`. (If this isn't too oxymoron or verbose) 
Although be careful in generalizing, some "capsule hotel" have a minority of or exclusively separate non-stacked partitions with doors, basically a bed-only mini-room without toilet and bathroom.

codeinabox:
Thank you, that makes sense and I agree hostel is the most appropriate as arguably what differentiates a hotel and hostel could be if there is a shared bathroom. In the case of Nine Hours the capsule is private however the bathroom is still shared.

PR #6267, opened 2022-03-04
Separate fast food restaurants and restaurants (HK)

tommylung:
Separate
- Din Tai Fung
- KFC
- McDonald's
- Mos Burger
- Saizeriya
- Yoshinoya

Add
- TamJai Yunnan Mixian
- TanJai SamGor

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -3149,6 +3149,7 @@
           "bh",
           "ca-qc.geojson",
           "cn",
+          "hk",
           "kw",
           "mo",
           "om",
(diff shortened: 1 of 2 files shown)


data/brands/amenity/fast_food.json
```diff
@@ -3149,7 +3149,6 @@
           "bh",
           "ca-qc.geojson",
           "cn",
-          "hk",
```
LaoshuBaby, line 3152: it is exclude from global, not include

tommylung: Updated

Issue #6235, opened 2022-02-28
NTT is a brand, but what should we do by region?


blendy-map:
Hello.
Thank you for editing "operators/amenity/telephone.json".

After editing, I suddenly thought, "Isn't NTT (nippontelegraphandtelephone-8c2f5b) a brand managed by NTT East (nippontelegraphandtelephoneeest-8c2f5b) and NTT West (nippontelegraphandtelephonewest-8c2f5b)?"

It is said that in 1999, NTT's business was reorganized into a holding company structure and the business was transferred to wholly owned subsidiaries by region (NTT East Japan, NTT West Japan).

Therefore, it is necessary to rewrite the NTT item according to "brands/amenity/telephone.json" and separate the include of NTT East and NTT West.

I'm wondering if I should use GEOJSON for each prefecture when splitting the include statement, or GEOJSON for the jurisdiction of NTT East and NTT West.

### prefecture-split pattern
```json:prefecture-split.json
    {
      "displayName": "NTT東日本",
      "id": "nippontelegraphandtelephoneeest-8c2f5b",
      "locationSet": {"include": ["jp-tokyo.json","jp-nagano.json","jp-miyagi.json"]},
      "tags": {
        "amenity": "telephone",
        "operator": "東日本電信電話株式会社",
        "operator:en": "Nippon Telegraph and Telephone Eest",
        "operator:ja": "東日本電信電話株式会社",
        "operator:short": "NTT東日本",
        "operator:wikidata": "Q11527212",
        "operator:wikipedia": "ja:東日本電信電話"
      }
    },
```
### jurisdiction-split pattern
```json:jurisdiction-split.json
    {
      "displayName": "NTT東日本",
      "id": "nippontelegraphandtelephoneeest-8c2f5b",
      "locationSet": {"include": ["jp-ntt-east.json"]},
      "tags": {
        "amenity": "telephone",
        "operator": "東日本電信電話株式会社",
        "operator:en": "Nippon Telegraph and Telephone Eest",
        "operator:ja": "東日本電信電話株式会社",
        "operator:short": "NTT東日本",
        "operator:wikidata": "Q11527212",
        "operator:wikipedia": "ja:東日本電信電話"
      }
    },
```
I presented an example like this. (There is a misspelling "s/Eest/East/" in the original, but I will fix it later.)
Please let me know if there is any good way.

arch0345:
[...]

PR #6271, opened 2022-03-05
NTT is a brand

blendy-map:
#6235


arch0345:
Thanks @blendy-map! Yeah, I believe it's best to have their own geojsons since as you mentioned in the issue they span across multiple prefectures.

PR #6281, opened 2022-03-06
Greek and Turkish names for Bank of Cyprus

dmromanov:
Cyprus uses Greek and Turkish languages.

https://en.wikipedia.org/wiki/Bank_of_Cyprus

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -1966,7 +1966,7 @@
         "brand:wikidata": "Q806678",
         "brand:wikipedia": "en:Bank of Cyprus",
         "name": "Bank of Cyprus",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/bank.json
```diff
@@ -1965,7 +1965,9 @@
         "brand": "Bank of Cyprus",
         "brand:wikidata": "Q806678",
         "brand:wikipedia": "en:Bank of Cyprus",
-        "name": "Bank of Cyprus"
+        "name": "Bank of Cyprus",
+        "name:tr": "Kıbrıs Bankası"
```
LaoshuBaby, line 1969:
```suggestion
        "name:tr": "Kıbrıs Bankası",
```

PR #6265, opened 2022-03-04
Update Hong Kong Bus information

tommylung:
Add
- Bus
 NLB (嶼巴)
- Minibus
 Kowloon Green Minibus (九龍專綫小巴)

Update
- Bus
 KMB (九巴)
 LWB (龍運巴士)
 MTR Bus (港鐵巴士)
 Residents' Service (居民巴士)
- Minibus
 Hong Kong Island Green Minibus (香港島專綫小巴)
 New Territories Green Minibus (新界專綫小巴)

Delete
- Bus
 Cross-harbour Bus (過海隧道巴士)
 Hong Kong Island Bus (港島巴士)
 Kowloon & NT Bus (九龍及新界巴士)
 KMB & LW (九巴龍運)
 North Lantau External Bus (北大嶼山對外巴士)

PR #6283, opened 2022-03-06
Remove inappropriate `network=` in Hong Kong worsening bus route tags

Kovoschiz:
They are either official service classifications by the government (routes totally unrelated to each other), or route brandings. The HK community hasn't agreed on a standard tagging on them yet (due to slightly more complex company structures), so I would not complete the info at this moment. Priority is on stopping wrong  re-taggings urgently, as seen from https://osmcha.org/changesets/118155528 et al by the same new user editing basically correct KMB values to meaningless categories (those titles are wrong too --- They aren't the official names used).

I have difficulty understanding  #6265: What is being asked there? @tommylung @LaoshuBaby


tommylung:
Thanks for your research.

May I know where is the document about the standard tagging of Hong Kong's transports? I cannot find the document on OSM wiki. I think the new user just take the reference on other verified routes.

Kovoschiz:
> May I know where is the document about the standard tagging of Hong Kong's transports? I cannot find the document on OSM wiki. I think the new user just take the reference on other verified routes.

The Hong Kong page on HK is not very updated or reviewed. We had some small debates in changesets and Discord in the past, especially concerning how to tag KMB & LWB and Ctb & NWFB (one reason I left it untouched for now).

tommylung:
For example: 
[KMB 48X](https://www.openstreetmap.org/relation/13895715)
- [KMB Bus 48X (Bayview Garden → Wo Che)](https://www.openstreetmap.org/relation/13895700)
- [KMB Bus 48X (Wo Che → Bayview Garden)](https://www.openstreetmap.org/relation/5838583)

May I know this new user followed the standard tagging for Hong Kong?

Kovoschiz:
Can you read your Discord? I message you some info.

arch0345:
As @LaoshuBaby mentioned, these entries will get readded every time someone runs the `npm run build` script. This is because we collect common networks (used 10 or more times) from the planet each week. These should either get added to as one of the `matchNames` for another entry or get added to the `named` `exclude` list at the top of the file. With the exception of Long Win Bus, these entries won't be seen in iD as only entries with a linked Wikidata item get an iD preset.

Kovoschiz:
Thanks for explaining here. I will see what I can do.

Kovoschiz:
Await more general consensus on tagging

PR #6286, opened 2022-03-07
Fixed Chinese problem on Hong Kong banks

tommylung:
Fixed Chinese problem on Hong Kong banks.


LaoshuBaby:
I also have a certain responsibility, because I actually havent found this problem
😱
Thank you!

PR #6287, opened 2022-03-07
Add local zone

Mannivu:
I've added local .geojson to better locate potential BBC Roma banks.


arch0345:
Thanks for adding these geojsons, however, I recommend drawing these by hand instead with geojson.io using the polygon tool as suggested in this [wiki article](https://github.com/osmlab/name-suggestion-index/wiki/Feature-Files) since we try to use as few points as possible, see https://github.com/osmlab/name-suggestion-index/issues/4784#issuecomment-742003434. There's no strict limit, but I would try to use at less than 100 points per geojson.

arch0345:
Thanks for adding these geojsons, however, I recommend drawing these by hand instead with geojson.io using the polygon tool as suggested in this [wiki article](https://github.com/osmlab/name-suggestion-index/wiki/Feature-Files) since we try to use as few points as possible, see https://github.com/osmlab/name-suggestion-index/issues/4784#issuecomment-742003434. There's no strict limit, but I would try to use at less than 100 points per geojson.

Mannivu:
@arch0345 I've updated simplified version of the geojsons. I hope they are fine.

arch0345:
Thanks @Mannivu!

PR #6293, opened 2022-03-09
Update wastewater_plant.json

zymurgic:
United Utilities had a mis-spelt listing. Added the wikidata link to the correct one and removed the mis-spelt listing.


arch0345:
Thanks for catching this @zymurgic!

PR #6299, opened 2022-03-12
Add Eastern Bank

zyphlar:


UKChris-osm:
Hi @zyphlar and thanks for this PR.

Is Eastern Bank only available within Massachusetts? If so, you can use the location “us-ma.geojson” to restrict it to the local area!

zyphlar:
It has at least a few locations in New Hampshire, though I wasn't able to confirm if it has any outside those two states. However, they do have a US Trademark on "Eastern Bank" so any similar bank would have to use other words in full, like "Eastern National Bank" or "Eastern Bank of Florida" to avoid a lawsuit. https://en.wikipedia.org/wiki/Eastern_Bank

PR #6300, opened 2022-03-12
Ministop, Japan locations

dmromanov:
Ministop operates in Japan, Philippines, China, Vietnam and Kazakhstan

https://en.wikipedia.org/wiki/Ministop#International_operations

Not sure what does "001" mean so left as-is.

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -2176,7 +2176,7 @@
     {
       "displayName": "Ministop",
       "id": "ministop-4d6b4b",
-      "locationSet": {"include": ["001", "ja", "ph", "cn", "vn", "kz"]},
+      "locationSet": {"include": ["ph", "cn", "vn", "kz"]},
       "tags": {
         "brand": "Ministop",
         "brand:wikidata": "Q1038929",
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -2176,7 +2176,7 @@
     {
       "displayName": "Ministop",
       "id": "ministop-4d6b4b",
-      "locationSet": {"include": ["001"]},
+      "locationSet": {"include": ["001", "ja", "ph", "cn", "vn", "kz"]},
       "tags": {
         "brand": "Ministop",
         "brand:wikidata": "Q1038929",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
001 means “world” so this would be available everywhere. As you have added specific locations in your PR, you can remove 001 as it isn’t required.

data/brands/shop/convenience.json
```diff
@@ -2176,7 +2176,7 @@
     {
       "displayName": "Ministop",
       "id": "ministop-4d6b4b",
-      "locationSet": {"include": ["001"]},
+      "locationSet": {"include": ["001", "ja", "ph", "cn", "vn", "kz"]},
```
LaoshuBaby, line 2179:
```suggestion
      "locationSet": {"include": ["ja", "ph", "cn", "vn", "kz"]},
```

blendy-map:
I think the "[ja](https://location-conflation.com/?locationSet=%7B%22include%22%3A%5B%22ja%22%5D%7D&referrer=nsi)" in include is a mistake of "[jp](https://location-conflation.com/?locationSet=%7B%22include%22%3A%5B%22jp%22%5D%7D&referrer=nsi)".

It seems that [ministop](https://nsi.guide/index.html?t=brands&k=shop&v=convenience&tt=Ministop) already has Japanese and Korean.
(Personally, I thought it would be good to merge the data of "dc44d4-fab923" and "ministop-652b3e" into "ministop-4d6b4b" so that they would match.)

LaoshuBaby:
> (Personally, I thought it would be good to merge the data of "dc44d4-fab923" and "ministop-652b3e" into "ministop-4d6b4b" so that they would match.)

Thank you for finding this typo mistake!

I've remove `ja`(jp) from "ministop-4d6b4b" so that it will appear in Japan and Korea. Merge to one entry is not needed because in NSI we usually use separated entry for different country/region, and this can suggest different language name for them.

For example, we don't need to suggest `name:ko` for Vietnam.

blendy-map:
@LaoshuBaby -san

> I've remove ja(jp) from "ministop-4d6b4b" so that it will appear in Japan and Korea. Merge to one entry is not needed because in NSI we usually use separated entry for different country/region, and this can suggest different language name for them.

I forgot about it. I'm sorry.

PR #6301, opened 2022-03-12
Remove vending=cigarettes from Tobaccoland's operator

Lukas458:
Hi,
here I removed vending=cigarettes from the Tobaccoland's operator, because it was noticed that Tobaccoland does operate vending machines which sell cigarettes _only_, but this company operates also vending machines which sell cigarettes but also some **other** things like condoms or something else, see https://www.openstreetmap.org/node/4405601460 for example.

So the vending-value should then be vending=cigarettes;condoms, and as there may also exist some other value combinations on Tobaccoland's vending machines which aren't _vending=cigarettes_ only, the NSI shouldn't propose to update the vending=* value to _vending=cigarettes_ at those. I hope it's understandable what I mean. So I removed the vending=cigarettes value for this brand, as the same way was chosen for the "Selecta" brand too, for example, because this brand does also have vending machines selling different/several different things.

### DIFF
--- a/data/brands/amenity/vending_machine.json
+++ b/data/brands/amenity/vending_machine.json
@@ -668,7 +668,6 @@
         "operator": "Tobaccoland Automatengesellschaft",
         "operator:wikidata": "Q1439872",
         "operator:wikipedia": "de:Tobaccoland Automatengesellschaft",
-        "vending": "cigarettes"
       }
     },
     {


Lukas458:
Thanks!

bhousel:
I actually think maybe a better solution than _removing_ the `vending` tag would be to use the `preserveTags` property to just tell the validator to not override the `vending` tag if it is already set to something other than "cigarettes".

Lukas458:
@bhousel 
Oh, well, yes, it seems you're right :-)
Sorry, I did not know this property until now.
So the code would be
`"preserveTags": "vending"`
then?

I would, however, also accept it specifically for this case, since vending=cigarettes is the usual value in most cases. _vending=cigarettes;..._ is seldom, but can appear and would probably even apply more often than currently mapped, but only from afar e. g. you don't necessarily recognize that other things are being sold in addition to cigarettes. But well, that wasn't the question here.

PR #6303, opened 2022-03-13
Bone Daddies ramen

codeinabox:


data/brands/amenity/restaurant.json
```diff
@@ -496,6 +496,21 @@
...
+      "locationSet": {
+        "include": ["gb-lon.geojson"]
+      },
+      "tags": {
+        "amenity": "restaurant",
+        "brand": "Bone Daddies",
+        "brand:wikidata": "Q111204601",
+        "cuisine": "japanese;ramen",
+        "name": "Bone Daddies",
+        "reservation": "no",
```
LaoshuBaby, line 510: This is `reservation` first time appear in NSI, should we record this?

My opinion is that if every branch of this restaurant refuse reservation, and this is a unique feature, we can add it.

codeinabox: On their Twitter account they make a point of saying no bookings https://www.twitter.com/BoneDaddiesLDN

PR #6307, opened 2022-03-13
Updated bike rental scheme

cmap99:


data/brands/amenity/bicycle_rental.json
```diff
@@ -1171,14 +1171,14 @@
       }
     },
     {
-      "displayName": "Coca-Cola Zero Bikes",
-      "id": "cocacolazerobikes-afbce7",
+      "displayName": "TFI Bikes",
+      "id": "",
       "locationSet": {"include": ["ie"]},
-      "matchNames": ["coca-cola zero® bikes"],
+      "matchNames": ["tfi bikes"],
```
LaoshuBaby, line 1177:
```suggestion
      "matchNames": ["Coca-Cola Zero Bikes"],
```

Thank you for complete those data! However, not every time we need a matchNames, the uppercase and lowercase spelling we don't concern. (Program will match both of them, even mixed) 

The `name`, `network`, `operator` can be detected so if matchNames a similar to them, matchNames will be removed when auto sort.

MatchNames is intended for different abbreviation or nickname people usually called. Also, if old brand changed its name, add old brand name to matchNames is needed.

data/brands/amenity/bicycle_rental.json
```diff
@@ -1171,14 +1171,14 @@
...
       "tags": {
         "amenity": "bicycle_rental",
-        "brand": "Coca-Cola Zero Bikes",
-        "network": "Coca-Cola Zero Bikes",
+        "brand": "TFI Bikes",
```
LaoshuBaby, line 1180:
```suggestion
        "brand": "TFI Bikes",
        "network": "TFI Bikes",
```

LaoshuBaby (changes requested):

PR #6310, opened 2022-03-15
add cinemas

cmap99:
Add Omniplex Cinemas & Irish Multiplex Cinemas

### DIFF
--- a/data/brands/amenity/cinema.json
+++ b/data/brands/amenity/cinema.json
@@ -59,6 +59,31 @@
         "brand:wikipedia": "en:AMC Theatres",
         "name": "AMC"
       }
+     },
+     {	
+      "displayName": "Irish Multiplex Cinemas",
+      "id": "",
+      "locationSet": {"include": ["ie"]},
+      "matchNames": ["imc"],
+      "tags": {
+        "amenity": "cinema",
+        "brand": "Irish Multiplex Cinemas",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/cinema.json
```diff
@@ -59,6 +59,31 @@
...
       }
+     },
+     {	
+      "displayName": "Irish Multiplex Cinemas",
+      "id": "",
+      "locationSet": {"include": ["ie"]},
+      "matchNames": ["imc"],
+      "tags": {
+        "amenity": "cinema",
+        "brand": "Irish Multiplex Cinemas",
```
LaoshuBaby, line 70: Should we exchange `brand` and `name`?

cmap99: For matchNames?

LaoshuBaby: I means, will Ireland community prefer the abbreviation as `name` tag and let the full name as `brand` ? This maybe a bit strange. (or you can use `"short_name": "IMC"`)

cmap99: It is a shorter name yes, but they are always be called "IMC *location*". On the buildings they usually also just have the logo which is just IMC and in some instances it would be the logo followed by a location underneath. So I would say the abbreviation is preferred.

PR #6319, opened 2022-03-17
Separate AEON for Hong Kong

tommylung:
Separate 
- AEON

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -9433,7 +9433,6 @@
     },
     {
       "displayName": "永旺 AEON",
-      "id": "aeon-eda947",
       "locationSet": {"include": ["hk"]},
       "matchNames": ["吉之島"],
       "tags": {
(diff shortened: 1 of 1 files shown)


data/brands/shop/supermarket.json
```diff
@@ -9431,6 +9431,27 @@
...
+      "matchNames": ["吉之島"],
+      "tags": {
+        "brand": "AEON",
+        "brand:en": "AEON",
+        "brand:wikidata": "Q271413",
+        "brand:wikipedia": "zh:永旺",
+        "brand:zh": "永旺",
+        "brand:zh-Hans": "永旺",
+        "brand:zh-Hant": "永旺",
+        "name": "AEON",
```
LaoshuBaby, line 9447: For those S/T Chinese exactly same entry, I don't think it need to separate tag

(And looks like HK is using bilingual name)

(This is a only a question, not suggestion)

tommylung: Updated. Could you please review it again? Thanks

PR #6323, opened 2022-03-20
Remove all police station in China

LaoshuBaby:
According to OSM wiki, it said:

> The operator tag is used to name a company, corporation, person or any other entity who is directly in charge of the current operation of a map object.
>
> ——[Key:operator](https://wiki.openstreetmap.org/wiki/Key:operator)

But according to defination, they belonging to county level public security bureau, not the Ministry of Public Security.

> 中华人民共和国公安机关的“派出所”的正式名称为“公安派出所”，为县、县级市、市辖区或旗的公安局或其公安分局的派出机构
>
> Translation: The official name of the "police station" of the public security organs of the People's Republic of China is "public security police station", which is a branch of the public security bureau of a county, county-level city, municipal district or banner or its public security sub-bureau
>
> ——[派出所#中华人民共和国](https://zh.wikipedia.org/wiki/%E6%B4%BE%E5%87%BA%E6%89%80#%E4%B8%AD%E5%8D%8E%E4%BA%BA%E6%B0%91%E5%85%B1%E5%92%8C%E5%9B%BD))

So those entry against OSMwiki, I'm not sure whether I should remove it. The retention or deletion discussion has also been met with disagreement within the Chinese community.

**Thanks to other maintainers for reviewing this PR. I need your advice, I don't know whether this is correct.**

This PR is similar to #6324

PR #6324, opened 2022-03-20
Remove MOE of China

LaoshuBaby:
According to OSM wiki：

> Owner or brand name aren't necessarily the operator
>
> An operator isn't necessarily the owner of the map feature. Many chains (store, restaurants...) use a [franchise](https://en.wikipedia.org/wiki/Franchising) system, where the brand does not operate the point-of-presence.
>
> For example, a lot of transport and communication networks had been built by public administrations and are now operated by private companies. In some places, the national government retains ownership of the infrastructure, as in power grids and railways. Power grids may be operated and maintained by a private entity, while the underlying infrastructure (lines, towers, poles, substations) remain under the ownership of the national government (e.g. in the Philippines). Railways may still be owned and maintained by the national government (e.g. the UK, most of Western Europe, South Korea), but trains using them may be operated by different companies.
>
> ——[Key:operator#How_to_use](https://wiki.openstreetmap.org/wiki/Key:operator#How_to_use)

And the defination of it：

>中央部门（或单位）所属高等学校，简称中央部委直属高校或中央部属高校，是中华人民共和国高等学校中公立学校的一种。

>由国务院组成部门、直属机构、直属事业单位等中央部门（或单位）在全中国范围内直属管理。

> ——[中央部属高校(ministry-owned university)](https://zh.wikipedia.org/wiki/%E4%B8%AD%E5%A4%AE%E9%83%A8%E5%B1%9E%E9%AB%98%E6%A0%A1)

So some mapper in China community think those 2 entry polute OSM data with wrong operator, and think this should be tagged as `owner=MOE`

**Thanks to other maintainers for reviewing this PR. I need your advice, I don't know whether this is correct.**

This PR is similar to #6323

### DIFF
--- a/data/operators/amenity/school.json
+++ b/data/operators/amenity/school.json
@@ -16,7 +16,8 @@
         "^privé(e)?$",
         "^religious$",
         "^school$",
-        "^学校$"
+        "^学校$",
+        "^中华人民共和国教育部"
       ]
     }
   },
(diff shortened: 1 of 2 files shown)


hanchao:
`operator=中华人民共和国教育部` 的用法并不[常见](https://overpass-turbo.eu/s/1hhZ)，无法达成广泛共识，经过讨论暂时移除。

PR #6326, opened 2022-03-20
Update  `amenity=telephone` operators in Canada

wolfy1339:


data/operators/amenity/telephone.json
```diff
@@ -45,20 +45,22 @@
     },
     {
       "displayName": "Bell",
```
wolfy1339, line 47: This entry and the one below are the same.
Everyone calls the company `Bell`, their real name is `Bell Canada`
I'm not sure how you would handle merging the 2 together so they both match

data/operators/amenity/telephone.json
```diff
@@ -45,20 +45,22 @@
...
+        "operator": "Bell",
+        "operator:wikidata": "Q815694"
       }
     },
     {
       "displayName": "Bell Canada",
-      "id": "bellcanada-2be02b",
-      "locationSet": {"include": ["001"]},
+      "id": "bellcanada-25374a",
+      "locationSet": {"include": ["ca"]},
```
LaoshuBaby, line 59:
```suggestion
      "locationSet": {"include": ["ca"]},
      "matchNames": ["bell"],
```

LaoshuBaby (changes requested):

PR #6327, opened 2022-03-20
add Wikidata entry to Der Mann

Findus23:
Quite minor, but I noticed that while there is no Wikipedia article, there is a Wikidata entry for the brand.


arch0345:
Thanks @Findus23 and welcome!

PR #6328, opened 2022-03-20
Update fast_food.json

waso99:
update Arab country codes for (KFC , McDonald's )


data/brands/amenity/fast_food.json
```diff
@@ -7439,7 +7508,13 @@
           "om",
           "qa",
           "sa",
-          "ye"
+          "eg",
```
LaoshuBaby, line 7511: Looks like Mcdonalds' global entry haven't exclude other countries

It's a good time to do that

https://github.com/osmlab/name-suggestion-index/blob/f51f320c3d23441d43452927bd6cc669edf005a5/data/brands/amenity/fast_food.json#L3822-L3827

LaoshuBaby (changes requested):

data/brands/amenity/fast_food.json
```diff
@@ -1718,7 +1718,40 @@
         "name:zh": "唐恩都乐",
         "takeaway": "yes"
       }
-    },
+    },    
+   {
+      "displayName": "دانكن دونتس'",
+      "locationSet": {
+        "include": [
```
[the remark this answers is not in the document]
waso99: OK

data/brands/amenity/fast_food.json
waso99: I think no more (McDonald's) in Russia

waso99:
update Arab country codes for (KFC , McDonald's )

data/brands/amenity/fast_food.json
```diff
@@ -7373,6 +7433,15 @@
           "om",
           "qa",
           "sa",
+          "eg",
```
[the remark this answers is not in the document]
waso99: I think no need to Exclude them in global. because it's already there

data/brands/amenity/fast_food.json
waso99: what about other non Latin countries such as china and japan? it's not excluded global

bhousel: Local brands probably should be excluded from the global item (even though this index has not consistently done it, we should).

The iD validator will still pick the most local one when matching features to their likely brand, but excluding them from the global item will prevent it from showing up in the preset list, which is probably what people want.

waso99: OK!
understood
thanks

PR #6335, opened 2022-03-21
Update pharmacy for HK

tommylung:
Separate
- 屈臣氏 Watsons (CN, HK, MO) (Move pharmacy to chemist)
- 北京同仁堂 Beijing Tong Ren Tang (HK, MO)


data/brands/amenity/pharmacy.json
```diff
@@ -2476,6 +2476,42 @@
         "name": "Watsons"
       }
     },
+    {
+      "displayName": "屈臣氏",
+      "locationSet": {"include": ["cn"], "exclude": ["hk", "mo"]},
+      "tags": {
```
LaoshuBaby, line 2482: Watsons had been in `shop/chemist`

https://github.com/osmlab/name-suggestion-index/blob/de589448d288c4bf880afc780a27c3567cec5619/data/brands/shop/chemist.json#L603-L621

tommylung: Thanks for your finding. Let's use `shop/chemist`.

tommylung: Should we move Watsons into `shop/chemist` too?

LaoshuBaby: > Should we move Watsons into `shop/chemist` too?

Yes, and add a matchTag so that pharmacy one will be redirected.

tommylung: > > Should we move Watsons into `shop/chemist` too?
> 
> Yes, and add a `matchTag` so that pharmacy one will be redirected.

Added.

LaoshuBaby: Besides, this is a Hong Kong brand, so can you decide that whether waston/wastons is correct?

tommylung: > Besides, this is a Hong Kong brand, so can you decide that whether waston/wastons is correct?

Watsons is the correct name. Updated.

PR #6336, opened 2022-03-21
Add book store for HK

tommylung:
Add
- 三聯書店 Joint Publishing
- 中華書局 Chung Hwa Book Company
- 商務印書館 The Commercial Press

Separate
- 誠品書店 Eslite Bookstore

### DIFF
--- a/data/brands/shop/books.json
+++ b/data/brands/shop/books.json
@@ -1377,6 +1377,25 @@
       }
     },
     {
+      "displayName": "誠品書店 Eslite Bookstore",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Because we already have `^owner` tag in NSI so I think this PR can be approved.

https://github.com/osmlab/name-suggestion-index/blob/6c32ce56c00d08cec4beb4a97a1918bdbf897492/data/transit/route/subway.json#L346-L347

PR #6337, opened 2022-03-21
Add mobile phone shops for HK

tommylung:
Add
- 衛訊 Wilson
- 領域 CityLink

Separarte
- 小米 Xiaomi

### DIFF
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -1402,13 +1402,9 @@
         "brand:wikidata": "Q1636958",
         "brand:wikipedia": "zh:小米集團",
         "brand:zh": "小米",
-        "brand:zh-Hans": "小米",
-        "brand:zh-Hant": "小米",
         "name": "小米之家 Mi",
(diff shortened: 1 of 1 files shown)


data/brands/shop/mobile_phone.json
```diff
@@ -1380,16 +1380,35 @@
...
+    },
+    {
+      "displayName": "小米 Xiaomi",
+      "locationSet": {"include": ["hk", "mo"]},
+      "tags": {
+        "brand": "小米 Xiaomi",
+        "brand:en": "Xiaomi",
+        "brand:wikidata": "Q1636958",
+        "brand:wikipedia": "zh:小米集團",
+        "brand:zh": "小米",
```
LaoshuBaby, line 1404: My opinion is that if `brand:zh-Hans` and `brand:zh-Hant` are exactly same, I suggest that we only reserve `brand:zh`

tommylung: Yes, we can do this

Issue #6342, opened 2022-03-21
Primark/ Penneys naming


cmap99:
Outside the Republic of Ireland the store is named `Primark`, in Ireland it's called `Penneys`. Its all one brand just named differently outside Ireland. ([see wikipedia](https://en.wikipedia.org/wiki/Primark)).

Is it possible to add Ireland as an exclusion for the existing `Primark` in the index and create a new entry for Ireland?
This would be the same brand tags only difference would be `name=Penneys`.

As for the existing entry the Primark website lists theses as it's operating countries: 
`{"include":["AT","BE","CZ","FR","DE","IT","NL","PL","PT","SI","ES","UK","US",]}` *(Ireland removed)*
Not sure if its better set like that or how it is currently as `["001"]`

cmap99:
Sorry I might have phrased that poorly. The existing entry *(Primark)* is currently set to the global geo location, I was just wondering if its better left as it is `["001"]` or change it to just include the countries I have listed above. 

As for going about this would I just add `{"exclude":["ie"]}` to the existing entry and create a new one?

cmap99:
> As for going about this would I just add `{"exclude":["ie"]}` to the existing entry and create a new one?

Am I correct in thinking that this wouldn't actually be necessary if I was to change the `Primark` entry from `["001"]` to `["AT","BE","CZ","FR","DE","IT","NL","PL","PT","SI","ES","UK","US"]`.

cmap99:
Created pull request (#6344)

PR #6344, opened 2022-03-21
Update/Split Primark

cmap99:
Updated the locationSet for Primark. Created new entry for Primark in the Republic of Ireland with a different name.

Related #6342


arch0345 (commented):
Thanks @cmap99! I'm going to commit the suggestion I made a couple days ago

PR #6345, opened 2022-03-21
Removed duplicate bus route & added wiki tags

cmap99:


data/transit/route/bus.json
```diff
@@ -3730,6 +3710,9 @@
       },
```
arch0345, line 3710: If these are duplicates, you should add matchNames so that existing bus routes/stops with this network value get matched with this entry:
```suggestion
      },
      "matchNames": ["Bus Éireann Dublin Commuter"],
```

data/transit/route/bus.json
```diff
@@ -4347,7 +4330,10 @@
       "locationSet": {"include": ["ie"]},
```
arch0345, line 4330: If these are duplicates, you should add matchNames so that existing bus routes/stops with this network value get matched with this entry:
```suggestion
      "locationSet": {"include": ["ie"]},
      "matchNames": ["Bus Éireann Expressway"],
```

arch0345 (changes requested):

arch0345:
Thanks for these updates @cmap99! I've added some suggestions that should be reviewed before this is merged

cmap99:
Thank you @arch0345. I added your suggestions.

PR #6347, opened 2022-03-22
Holey Moley miniature golf

### LINKED ISSUE #6213: Holey Moley bars with mini-golf
Holey Moley is a chain of bars with mini-golf https://www.holeymoley.com.au/

What would be the best way to tag these? Looking at the ones already mapped they use both `leisure=miniature_golf` and `amenity=bar`

codeinabox:
Holey Moley miniature golf with locations in Australia, New Zealand and Singapore. Resolves #6213


arch0345:
Thanks @codeinabox

PR #6348, opened 2022-03-22
Update pharmacy.json

kjonosm:
Remove wheelchair tag. This should preferably be confirmed on site.


LaoshuBaby (approved):
I agree with this, because this situation will differ from every shop.

--
BTW, we only have 1 `wheelchair` in whole NSI

图片

arch0345:
It looks like this was added in #2444

kjonosm:
I left it as is and added a note to the item about discussion in #2444

PR #6349, opened 2022-03-22
Update restaurant.json

kjonosm:
Remove pointless diet:vegetarian=limited

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -2363,7 +2363,6 @@
         "cuisine": "steak_house",
         "diet:meat": "yes",
         "diet:vegan": "no",
-        "diet:vegetarian": "limited",
         "name": "Loetje",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
In my opinion, I agree with you. Because those kind of restaurant, steak house, may be far from what people think of as vegetarian food. In this case we don't need to import this tag. Let this tag be used on vegetrian food brands that are more defined and, as a brand identity, very consistent from shop to shop.

图片
(This is a code from #4855, I think it's a abuse of tag)

---

FYI:

We now have 5 `diet:vegetrian` in NSI

+ exki-a3472e  **yes**
+ veggiepret-45095e **only**
+ loetje-2b67bf  **limited**
+ pieminister-9b5453  **yes**
+ tonkotsu-9b5453  **yes**

arch0345:
Looks good

PR #6361, opened 2022-03-25
Updated tags for Maryland State Police and added North Carolina State Highway Patrol

Eric-Sparks:
Apparently, office: government has some additional tags to go along with it.  I'm adding those here.


UKChris-osm:
I've not come across the "_admin_level_" tag before - what does level four mean, and how does it relate to other levels?

Eric-Sparks:
> I've not come across the "admin_level" tag before - what does level four mean, and how does it relate to other levels?

The admin_level tag[0] is used to denote what administrative level of government the thing it's applied to represents.  In the United States, an admin_level=4 represents "State" level[1].  It is a recommended tag to use on government=* tagged objects[2].

[0] https://wiki.openstreetmap.org/wiki/Key:admin_level
[1] https://wiki.openstreetmap.org/wiki/United_States_admin_level
[2] https://wiki.openstreetmap.org/wiki/Tag:office%3Dgovernment

arch0345:
I'm not sure if we should dual tag these with  `office=government`+`government=police` as this is kind of implied by `amenity=police`.

Eric-Sparks:
> I'm not sure if we should dual tag these with `office=government`+`government=police` as this is kind of implied by `amenity=police`.

Doh!  The office=* tag went with the building tag I was using so really we should be adding building=office as well.  This has nothing to do with the amenity=police tag as that describes the amenity while the other describes, much more thoroughly, the building.

bhousel:
I agree with @arch0345  - I'm going to just merge this and remove the extra tags, so it doesn't linger any longer.

`admin_level`, `office`, and `government` aren't totally wrong, but they aren't consistent with the other police entries in this index.   If we decide to add them later, that's ok, but I think it should be discussed separately, and people can make an effort to go through and add these tags to all the police stations that need them.

PR #6364, opened 2022-03-25
ENGIE EV charging in UK has merged to Geniepoint

zymurgic:
Engie EV charging in the UK has merged to their existing GeniePoint brand. All locations in the UK have been adjusted in OSM. Engie still use that brand in other countries for EV charging.


arch0345:
Thanks @zymurgic!

PR #6368, opened 2022-03-26
Add some matchTags

peternewman:
For shop=clothes and clothes=outdoor type cases.

### DIFF
--- a/data/brands/shop/outdoor.json
+++ b/data/brands/shop/outdoor.json
@@ -93,6 +93,7 @@
       "displayName": "Blacks",
       "id": "blacks-f55d12",
       "locationSet": {"include": ["gb"]},
+      "matchTags": ["shop/clothes"],
       "tags": {
         "brand": "Blacks",
         "brand:wikidata": "Q64822933",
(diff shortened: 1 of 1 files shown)


arch0345:
Looks good @peternewman

peternewman:
Thanks @arch0345 . I did vaguely wonder if these should be for the whole category instead rather than just individual brands...

arch0345:
Yeah I also think this could be applied to the whole category, to do this `shop/outdoor` could be added here:
https://github.com/osmlab/name-suggestion-index/blob/main/dist/matchGroups.json#L86-L96

peternewman:
> Yeah I also think this could be applied to the whole category, to do this `shop/outdoor` could be added here: https://github.com/osmlab/name-suggestion-index/blob/main/dist/matchGroups.json#L86-L96

Cool, given that agreement, I've done it, although I think you meant config rather than dist!

I've also done it here instead:
https://github.com/osmlab/name-suggestion-index/blob/fe358961bf175d06732c0a048dcd730602b363e0/config/matchGroups.json#L189-L192

As it was a smaller and more appropriate group.

PTAL at #6385

PR #6390, opened 2022-03-31
Fix Bee-Line bus network operator.

ZLima12:
The error was introduced in 0d83311f23f0b69aeed6961186b9d882af4b8377. See [the wikipedia page](https://en.wikipedia.org/wiki/Bee-Line_Bus_System) for confirmation.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -1374,7 +1374,7 @@
         "network": "Bee-Line",
         "network:wikidata": "Q4879491",
         "network:wikipedia": "en:Bee-Line Bus System",
-        "operator": "Worchester County",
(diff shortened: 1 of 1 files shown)


mxxcon:
If it makes sense maybe also add `operator:short=`?

arch0345:
Oops, sorry about that mistake. Thanks for catching this @ZLima12!

arch0345:
> If it makes sense maybe also add `operator:short=`?

Good point, the wikipedia article mentions `WCDOT`, so I've added that in https://github.com/osmlab/name-suggestion-index/commit/7ccc99fe605dd4d491dda395c01aba42148d5a79

PR #6392, opened 2022-03-31
The Sebel hotels in Australia and New Zealand

codeinabox:


data/brands/tourism/hotel.json
```diff
@@ -2451,6 +2451,17 @@
         "tourism": "hotel"
       }
     },
+    {
+      "displayName": "The Sebel",
+      "locationSet": {"include": ["au", "nz"]},
+      "matchNames": ["sebel"],
+      "tags": {
+        "brand": "The Sebel",
+        "brand:wikidata": "Q110888248",
```
tas50, line 2460: @codeinabox Can you add their twitter or facebook account to the wikidata so we get brand icons while users are searching?

codeinabox: I've added their Facebook account to the Wikidata

PR #6394, opened 2022-04-01
Add Petit Pret

Cj-Malone:
So this is a sub fascia of Pret a Manger, like Veggie Pret but not as common.

It probably could have some miscellaneous tags like `diet:vegan`, `diet:vegetarian` and `takeaway`, but I've never been to one I don't know for sure.

I'm also not sure if this should be fast food vs cafe, but currently [OSM](https://overpass-turbo.eu/s/1hjE) has it 2-1 to cafe.


tas50:
Thanks @Cj-Malone

PR #6395, opened 2022-04-01
Added Uttarakhand Gramin Bank

MG-LSJ:
Added a regional bank with over 286 branches.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -9449,6 +9449,19 @@
         "official_name": "United Services Automobile Association"
       }
     },
+      {
+      "displayName": "Uttarakhand Gramin Bank",
+      "id": "uttarakhandgraminbank-6a1dfb",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/bank.json
```diff
@@ -9449,6 +9449,19 @@
         "official_name": "United Services Automobile Association"
       }
     },
+      {
+      "displayName": "Uttarakhand Gramin Bank",
+      "id": "uttarakhandgraminbank-6a1dfb",
+      "locationSet": {"include": ["in"]},
+      "matchNames": ["Uttarakhand Gramin Bank","Gramin Bank"],
```
LaoshuBaby, line 9456: "Uttarakhand Gramin Bank" is exactly same as "brand" and "name", so you dont need to repeat in matchNames. Duplicated matchNames will be removed when npm run build.

Thank you for share those indian brand with us.

PR #6398, opened 2022-04-01
Update Netz NÖ

bpranz:


sb12:
Shouldn't this be "Netz Niederösterreich GmbH"? That is currently the most used value according to Taginfo. That is also the name used in Wikipedia and on the Imprint of their company website.

"Netz NÖ" seems to be an abbreviations which should generally be avoided in OSM. They also do not use this abbreviation on their own website, you only find it in the logo as "NÖ Netz".

This also needs to be unified for all the power files (substation, lines and maybe others).

data/operators/power/substation.json
```diff
@@ -4764,12 +4764,15 @@
...
+        "nö netz ag", 
+        "netz niederösterreich",
+        "evn netz"
       ],
-      "note": "Might be related to EVN",
       "tags": {
-        "operator": "Netz Niederösterreich",
-        "power": "substation"
+        "operator": "Netz NÖ",
+        "power": "line"
```
tas50, line 4773: This should still be substation

bpranz: True, was a copy/paste mistake

bpranz:
> Shouldn't this be "Netz Niederösterreich GmbH"? That is currently the most used value according to Taginfo. That is also the name used in Wikipedia and on the Imprint of their company website.
> 
> "Netz NÖ" seems to be an abbreviations which should generally be avoided in OSM. They also do not use this abbreviation on their own website, you only find it in the logo as "NÖ Netz".
> 
> This also needs to be unified for all the power files (substation, lines and maybe others).

> Relates to #6398: maybe Netz NÖ could be in `operator:short` and the unabbreviated version in `operator`?

"Netz NÖ" is the common name, used in articles and documents eg. in this examples (one from their own website, the other from a local newspaper):

https://www.netz-noe.at/Netz-Niederosterreich/Presse/Gemeinsame-Sprengubung-von-Netz-NO-und-Feuerwehr.aspx

https://www.noen.at/gmuend/16-millionen-projekt-von-netz-noe-ist-fertig-bezirk-gmuend-redaktionsfeed-stromversorgung-netz-noe-umspannwerk-gross-gerungs-umspannwerk-gmuend-stromleitung-redaktion-304108402

"Netz Niederösterreich" would be the long version of this name; and "Netz Niederösterreich GmbH" would be the legal company name

data/operators/power/substation.json
bhousel: I merged as-is and fixed it in 044b00689

PR #6407, opened 2022-04-02
Longdan supermarkets selling Asian products

codeinabox:


data/brands/shop/supermarket.json
```diff
@@ -3847,6 +3847,17 @@
...
       }
     },
+    {
+      "displayName": "Longdan",
+      "locationSet": {"include": ["gb-eng"]},
+      "tags": {
+        "brand": "Longdan",
+        "brand:wikidata": "Q111462402",
+        "name": "Longdan",
+        "origin": "asian",
```
LaoshuBaby, line 3857: This is first time we introduce a "origin" tag in NSI 

But I think asian goods is the special point for a chain brand, like pure vegan is another kind of special point too. If every branch shop like this and mainly serve asian goods, I agree to introduce it.

LaoshuBaby (approved):

PR #6408, opened 2022-04-02
Caffè Nero countries

dmromanov:
> runs more than 1000 coffee houses in eleven countries: the UK, Ireland, Sweden, Poland, Cyprus, Croatia, Turkey, the UAE, Oman, United States

https://en.wikipedia.org/wiki/Caff%C3%A8_Nero


tas50:
Thanks @dmromanov

PR #6412, opened 2022-04-04
add some brands

k-yle:
This PR adds 4 logistics companies, 3 gas pipeline operators, and 1 pharmacy chain.

I've also tweaked the `displayName` of 2 Australian brands, since the current label is a source of confusion when editing in New Zealand.


arch0345:
Thanks @k-yle

PR #6416, opened 2022-04-04
Unify EnBW

sb12:
The same name should be used for all amenities operated by EnBW.

There are also some entries for `EnBW`/`EnBW AG` in https://github.com/osmlab/name-suggestion-index/blob/main/data/operators/power/substation.json which should not be there. EnBW does not operate any electricity transmission infrastructure anymore, these are now operated either by TransnetBW GmbH or Netze BW GmbH. `enbw` is listed as `exclude` in this file, not sure why it still gets added.

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -1133,16 +1133,17 @@
       }
     },
     {
-      "displayName": "EnBW",
+      "displayName": "EnBW Energie Baden-Württemberg AG",
       "id": "enbw-1299a8",
       "locationSet": {"include": ["de"]},
       "matchNames": [
-        "enbw energie baden-württemberg ag",
+        "enbw",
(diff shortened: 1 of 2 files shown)


LaoshuBaby:
> There are also some entries for `EnBW`/`EnBW AG` in https://github.com/osmlab/name-suggestion-index/blob/main/data/operators/power/substation.json which should not be there. EnBW does not operate any electricity transmission infrastructure anymore, these are now operated either by TransnetBW GmbH or Netze BW GmbH. `enbw` is listed as `exclude` in this file, not sure why it still gets added.

https://github.com/osmlab/name-suggestion-index/wiki/Category-Property-Reference

Maybe you can try write "^enbw$" ?

arch0345:
> Maybe you can try write "^enbw$" ?

Thanks, that seemed to fix it.

PR #6423, opened 2022-04-05
Update line.json

michael-reuter:
Adding match names and tags to Evergy, removing Westar


bhousel:
Thanks @michael-reuter 👍

PR #6424, opened 2022-04-05
Initial capital of the name decapitalized

paunofu:
Initial capital of the name of the supermarket chain masymas decapitalized because it's in lowercase everywhere (website, ticket, signs, ... ) also i have added the convenience shop of the same brand but they are only available in Murcia and València


data/brands/shop/supermarket.json
```diff
@@ -4093,14 +4093,14 @@
...
+      "displayName": "masymas",
       "id": "masymas-0513f1",
       "locationSet": {"include": ["es"]},
       "tags": {
-        "brand": "Masymas",
+        "brand": "masymas",
         "brand:wikidata": "Q6135987",
         "brand:wikipedia": "es:Supermercados Masymas",
-        "name": "Masymas",
+        "name": "masymas",
```
1ec5, line 4103: > Initial capital of the name of the supermarket chain masymas decapitalized because it's in lowercase everywhere (website, ticket, signs, ... )

Capitalization seems to be inconsistent on [some of their websites](https://super-masymas.com/nuestras-tiendas/). Often, brands lowercase or uppercase their logo but continue to use a conventional mixed-case capitalization in plain text contexts, including in prose.

paunofu: Masymas is a brand operated by three companies independently, each operator has his own website and it appears as  masymas, Masymas or Más y Más but seems that is more frequently use masymas than the others.


[https://www.masymas.com/ca/](masymas in Valencia and Murcia)

[https://www.supermasymas.com](masymas in Asturias and León)

[https://super-masymas.com](masymas in Córdoba and Jaén)

1ec5: Ah, thanks for the explanation. It looks like some operators even have different logos for the chain. If necessary, this entry can be split into three separate entries with the same `displayName` but different `id`s, `locationSet`s, `name`s, and even `brand:wikidata`s.

bhousel: Yeah looking over this PR, I'm worried that it's maybe the wrong approach. I do see that the brand is split up into smaller sub-brands with different operators - this seems fine.  

But, the operator wikidata ids were all just added recently and look like duplicates, and I worry that they'll be deleted or merged by wikidata admins.

The display names are just too long too, and will be cut off in RapiD, so would be better to shorten them to just the length needed for a mapper to feel ok that they are using the right preset.

(I'm ok with merging, but this PR just looks sloppy to me).

paunofu: The wikidata pages of the operators were created by me, I'm asking my wikidata local community for help improving the pages.

And about the DisplayName can be deleted because it's only a clarification of the differences, locationSets are defined so is not necessary.

data/brands/shop/supermarket.json
```diff
@@ -4093,14 +4093,62 @@
       }
     },
     {
-      "displayName": "Masymas",
-      "id": "masymas-0513f1",
-      "locationSet": {"include": ["es"]},
+      "displayName": "masymas (Juan Fornés Fornés S.A.)",
+      "locationSet": { "include": [[-0.7197, 39.4717, 60],[-0.3, 38.6983, 60],[-1.041389, 38.207222, 60],[-1.6023, 37.7126, 50],[-0.8792, 37.6237, 20],[0.0096, 40.3, 60]] },
```
LaoshuBaby, line 4097: I'm confused that why so much of location are defined by a series of circle……It's different for other people's understand or modification. Maybe you can add a `note` key?

paunofu: I know that is a bit weird. This is from the [website](http://[www.masymas.es/layout/portada_geolocalizada.php?Idioma=en](http://www.masymas.es/layout/portada_geolocalizada.php?Idioma=en)) BRAND SHARED BY 4 DIFFERENT COMPANIES, WITH TOTALLY INDEPENDENT MANAGEMENT, DISTRIBUTED IN DIFFERENT STATES.

They only share the name and the own brand of supermarket products (in Spanish white label)

data/brands/shop/supermarket.json
bhousel: > And about the DisplayName can be deleted because it's only a clarification of the differences, locationSets are defined so is not necessary.

It is important to use unique display names, at least for when the JOSM presets get generated they will all just show up on one list.

bhousel: The build actually enforces unique Display Names:
Screen Shot 2022-05-06 at 10 51 54 AM

PR #6428, opened 2022-04-07
Loft is a Japanese department store brand

codeinabox:
Loft was incorrectly tagged as a hardware shop when it actually is a department store selling household goods

https://www.loft.co.jp/en/about_us.html
https://en.japantravel.com/tokyo/shopping-at-loft/26841
https://jw-webmagazine.com/loft-the-coolest-store-to-shop-cutting-edge-gifts-5d776944268b/

### DIFF
--- a/data/brands/shop/department_store.json
+++ b/data/brands/shop/department_store.json
@@ -1444,6 +1444,23 @@
       }
     },
     {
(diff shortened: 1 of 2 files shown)


tas50:
Thanks for linking the explaination

PR #6430, opened 2022-04-08
Adding Transurbano for Guatemala

Lewatoto:
After adding some details to the wikidata of https://es.wikipedia.org/wiki/Transurbano I've think is good to add here as well.


tas50:
Thanks @Lewatoto

PR #6454, opened 2022-04-13
Update bank.json

luisluigi639:
Removal of TCF Bank as it has been acquired by an other bank.


LaoshuBaby:
> it has been acquired by an other bank.

So is the new bank in NSI? If so, you can add TCF Bank as a matchName of new bank.

luisluigi639:
Should be set now.

PR #6460, opened 2022-04-14
 Split entries for EKO by country 

Dimitar5555:


bhousel:
I'm ok if people really want to do this (if wikidata folks are ok with it too), but I wonder - why do it?
Why not just use Q903198 everywhere for EKO?  
Is EKO in Bulgaria so special that it needs its own QID ?

Dimitar5555:
It's name in Greece is Hellenic Petroleum. It also has a diffrent logo. Having separate Wikidata entries shouldn't be a problem since they are linked using 'parent company'.

bhousel:
oh sorry - I didn't realize it was differently branded.
But then why does the other one still say "EKO"?

Dimitar5555:
Looks like the preset name was incorrect. Hellenic Petroleum is the parent company which owns EKO in Serbia, Montenegro, Greece, Bulgaria and Cyprus. In North Macedonia it operates under the brand OKTA. I'm going to fix it later today.

bhousel:
I still don't really understand what you're doing.
We wouldn't normally make different QIDs for each country.
Aren't all these petrol stations just branded as "EKO"?

https://www.google.com/search?q=eko+petroleum&rlz=1C5CHFA_enUS903US903&source=lnms&tbm=isch&sa=X&ved=2ahUKEwiL3LPc0Jb3AhWwj4kEHRW3CjEQ_AUoA3oECAEQBQ&cshid=1650045064924299&biw=1680&bih=914&dpr=2

Screen Shot 2022-04-15 at 1 51 54 PM


Hellenic Petroleum might be the owning company, but as far as I can tell, they don't write that on the fuel stations.

Dimitar5555:
All of them have diffrent founding dates, HQs, websites and so on. The only common thing they share is their name and logo. I'm not sure if it's possible to include all of this information in a single Wikidata entry.

Dimitar5555:
Is there something that I'm missing @LaoshuBaby? I can't read your thoughts from your reactions.

LaoshuBaby:
> I can't read your thoughts from your reactions.

That's because I think those changes will be strange, you said "The only common thing they share is their name and logo.", if so why not think this is a brand? Like what we treat McDonald's (I often use it as an example). 
But I'm not sure my opinion is the repository needed, and I think I should still limited to routine maintenance and don't involved in decision making, soI haven't comment, just add a reaction emoji. 😂

Dimitar5555:
They are linked as a brand by having `brand=EKO`. The Macedonian branch has a different name and that (imo) makes it a different brand. The separation by country is on purpose because each country uses a different alphabet (Latin, Cyrillic or both). That makes it easier to have `name:XX` for each country.

Dimitar5555:
I've just noticed that I haven't answered all of @bhousel's questions.

>  Aren't all these petrol stations just branded as "EKO"?

All of them are branded as EKO except the second one on the second row from right to left.

> Hellenic Petroleum might be the owning company, but as far as I can tell, they don't write that on the fuel stations.

That's true, but all of EKO's gas stations are currently getting Hellenic Petroleum's logo, their wikidata entry points to Hellenic Petroleum and part of the `displayName` is written in Greek which is barely readable to most people.
image

bhousel:
Thanks @Dimitar5555 , I just merged it
Also, for these I set the property `"matchNames": ["q903198", "εκο", "еко"],`
So the matcher will also catch anything that happened to be tagged with the parent "Hellenic Petroleum" QID, as well as some of the common Cyrillic/Latin combinations that have been finding their way into the index.

PR #6461, opened 2022-04-14
Change Petrol English name

Dimitar5555:
I'm not sure if it will cause any conflicts because `Petrol` is considered a generic name.


bhousel:
> I'm not sure if it will cause any conflicts because `Petrol` is considered a generic name.

It should be ok - the string "petrol" will only match with precedence in Bulgaria..

PR #6477, opened 2022-04-19
Added brand website for 36.6 pharmacies

a-ney:


arch0345:
Please note that we don't support `:website` tags in NSI: see https://github.com/osmlab/name-suggestion-index/wiki/Contributing#tags-required

PR #6482, opened 2022-04-21
Update ice_cream.json

dan13l9:
updated

### DIFF
--- a/data/brands/amenity/ice_cream.json
+++ b/data/brands/amenity/ice_cream.json
@@ -205,7 +205,7 @@
     {
       "displayName": "Cold Stone Creamery",
       "id": "coldstonecreamery-2155b4",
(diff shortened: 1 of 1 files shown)


bhousel:
Looks like you accidentally removed the display name for "Marble Slab Creamery" and added a few other odd characters..
Can you please fix, 

Also as a maintainer, I appreciate when people leave better comments about what they are changing and why.
Just saying "updated" is not very nice as it puts more work on us to guess what you're doing..

data/brands/amenity/ice_cream.json
```diff
@@ -417,8 +417,9 @@
     },
     {
       "displayName": "Marble Slab Creamery",
+      ",
```
LaoshuBaby, line 420:
```suggestion
```

PR #6484, opened 2022-04-22
Updates to Alliant, CIPCO, Corn Belt, MidAmerican, - ITC Midwest

michael-reuter:
…st (objects updated to ITC)


data/operators/power/line.json
```diff
@@ -2878,16 +2882,6 @@
         "power": "line"
       }
     },
-    {
-      "displayName": "ITC Midwest",
```
LaoshuBaby, line 2882: According to nsi-collector's data

https://github.com/osmlab/name-suggestion-index/blob/919f4d0bbdca40ca4a2ab495d2d78383d757683a/dist/filtered/operators_keep.json#L8113

https://github.com/osmlab/name-suggestion-index/blob/919f4d0bbdca40ca4a2ab495d2d78383d757683a/dist/filtered/operators_keep.json#L9115

There are still 384 line and 82 substation named of ITC Midwest, so if you only delete it without add it to exclude name or add it to other entry's matchName, it will be generated again when formater run.

LaoshuBaby (changes requested):

data/operators/power/line.json
michael-reuter: Thanks for the feedback. Those files have since been updated and ITC Midwest fell off.

arch0345 (approved):

PR #6486, opened 2022-04-22
Update pipeline.json

sb12:
Fix duplicates for some German pipeline operators. Change `operator` from energy suppliers to the actual infrastructure operators.


sb12:
> However the collected name is Netze-Gesellschaft Südwest mbH, so maybe you should whether current map have a mbH instead of GmbH. I add the missing G name to matchNames.

Sorry my mistake. `Netze-Gesellschaft Südwest mbH` is correct, see https://www.netze-suedwest.de/impressum
(The "G" of GmbH is already present in Netze-**G**esellschaft)

PR #6489, opened 2022-04-25
Add Pfister furniture store

brawer:


LaoshuBaby:
Hey @brawer ! Welcome to open your first PR in NSI !

Thank you for share those brand information and make this preset more useful for OSM mapper.

---

By the way, those name only differ in uppercase and lowercase don't need to be a matchName, formater will auto-remove it.

图片

You can see this wiki page for more usage about matchNames key :
https://github.com/osmlab/name-suggestion-index/wiki/Contributing#matchnamesmatchtags-optional

brawer:
> By the way, those name only differ in uppercase and lowercase don't need to be a matchName, formater will auto-remove it.

Good to know, thanks for telling!

PR #6501, opened 2022-04-26
Add Kuhn, a Swiss bakery chain

brawer:
Kuhn is a common name in Switzerland, so we’re restricting matches
to features that are already tagged as `shop=bakery`.

### DIFF
--- a/data/brands/shop/bakery.json
+++ b/data/brands/shop/bakery.json
@@ -1043,6 +1043,17 @@
       }
     },
     {
+      "displayName": "Kuhn",
+      "id": "kuhn-6950bd",
+      "locationSet": {"include": ["ch"]},
+      "tags": {
+        "brand": "Kuhn",
+        "brand:wikidata": "Q111728345",
+        "name": "Kuhn",
+        "shop": "bakery"
+      }
+    },
+    {
       "displayName": "La Mie Câline",
       "id": "lamiecaline-731a34",
       "locationSet": {"include": ["fr"]},


bhousel:
> Kuhn is a common name in Switzerland, so we’re restricting matches
> to features that are already tagged as `shop=bakery`.

Because this item is in the `shop/bakery` file, it will already match `shop=bakery`.  
The `matchTags` property is for _other_ tag pairs that the items should match on.

data/brands/shop/bakery.json
```diff
@@ -1042,6 +1042,19 @@
         "shop": "bakery"
       }
     },
+    {
+      "displayName": "Kuhn",
+      "id": "kuhn-6950bd",
+      "locationSet": {"include": ["ch"]},
+      "tags": {
+        "bakehouse": "no",
```
LaoshuBaby, line 1050: https://wiki.openstreetmap.org/wiki/Key:bakehouse

According to OSM Wiki……If it is not a bakery but it still be tagged as bakery, this will be strange.

Could you tell us more about why add this tag?

data/brands/shop/bakery.json
```diff
@@ -1042,6 +1042,19 @@
...
+        "brand": "Kuhn",
+        "brand:wikidata": "Q111728345",
+        "cafe": "yes",
+        "drink:espresso": "served;takeaway",
```
LaoshuBaby, line 1054: This is the only “drink:*” tag in this file, is every branch shop of Kuhn have espresso?

And, I don't think serve sepresso in a bakery is strange, many of them can serve this, and milk? coffee? We don't need to list all menu.

See https://github.com/osmlab/name-suggestion-index/pull/5888

LaoshuBaby (changes requested):

brawer:
Thanks for your helpful explanation! I’ve removed the fine-grained tags.

data/brands/shop/bakery.json
brawer: Removed it.

data/brands/shop/bakery.json
brawer: Removed it.

arch0345 (approved):
LGTM, thanks @brawer

PR #6502, opened 2022-04-26
Add Brezelkönig pretzel chain

brawer:


data/brands/amenity/fast_food.json
```diff
@@ -617,6 +617,21 @@
...
+      "displayName": "Brezelkönig",
+      "id": "brezelkonig-096f6f",
+      "locationSet": {"include": ["at", "ch", "de"]},
+      "tags": {
+        "amenity": "fast_food",
+        "brand": "Brezelkönig",
+        "brand:wikidata": "Q111728604",
+        "cuisine": "pretzel",
+        "diet:vegetarian": "yes",
+        "drink:espresso": "takeaway",
```
LaoshuBaby, line 630: I have same question with https://github.com/osmlab/name-suggestion-index/pull/6501#discussion_r859334678

Is this necessary?

data/brands/amenity/fast_food.json
```diff
@@ -617,6 +617,21 @@
...
+    {
+      "displayName": "Brezelkönig",
+      "id": "brezelkonig-096f6f",
+      "locationSet": {"include": ["at", "ch", "de"]},
+      "tags": {
+        "amenity": "fast_food",
+        "brand": "Brezelkönig",
+        "brand:wikidata": "Q111728604",
+        "cuisine": "pretzel",
+        "diet:vegetarian": "yes",
```
LaoshuBaby, line 629: subtag of diet have been a very offen issue in this repository, it is the first time you use this tag, so here are some discussion for your information: https://github.com/osmlab/name-suggestion-index/pull/6224 , I want to confirm that which case Brezelkonig fit.

If this is a very pure vegetarian/vegan restaurant, obveriously a "only" value is accepted. Or if vegetarian/vegan is a feature of this chain brand's feature, even they advertise it, you can add a "yes" value.

But according to its website, I think a vegetarian=yes don't need been added to NSI preset.

图片

LaoshuBaby (changes requested):

data/brands/amenity/fast_food.json
brawer: Removed.

data/brands/amenity/fast_food.json
brawer: Thanks for the explanation! Removed the tag.

brawer:
Thanks for the feedback! Please take another look.

PR #6506, opened 2022-04-26
Add Schwyzer Kantonalbank, a bank in Switzerland

brawer:


LaoshuBaby:
If possible those addition in same one catagory can merge in 1 PR, "Add 7 bank in Switzerland"

图片

PR #6516, opened 2022-04-28
Add 12 Swiss cantonal banks

brawer:


LaoshuBaby:
I use `Banque Cantonale` as keyword searched on taginfo and it give me a series of results. So this is a valuable pull request. Thank you for add those information, and hope we can see more data in OSM using them ♥

图片

PR #6527, opened 2022-05-01
Added Shoppers (Baltimore/DC area supermarket chain)

bgo-eiu:
Added Shoppers (Baltimore/DC area supermarket chain). Logo can be derived from Facebook/Twitter pages on the Wikidata. "id" field is blank in my pull request


arch0345:
Thanks @bgo-eiu and welcome!

PR #6532, opened 2022-05-01
Update public_bookcase.json

danieldegroot2:
Added Official BookCrossing Zone (OBCZ), KinderzwerfboekStation, Minibieb

Noteably, KinderzwerfboekStations don't have to be exclusively children books. However, a similar thing applies to the name tag, where the name is not required to be "KinderzwerfboekStation".


arch0345:
Thanks @danieldegroot2 and welcome! Note that entries without a `:wikidata` tag won't get their own iD preset. If they don't have a Wikidata item, you can read about how to create one [here](https://github.com/osmlab/name-suggestion-index/wiki/Editing-Wikidata)

PR #6537, opened 2022-05-02
Restrict 6 Swiss banks to boundary of canton

brawer:


bhousel:
Thanks @brawer  - can you double check the GeoJSON polygons? I think when a Polygon contains multiple rings, the first is treated as an outer, and all other are treated as inners (holes).  You might need to convert these to MultiPolygons if they are boundaries with exclaves.

brawer:
> You might need to convert these to MultiPolygons if they are boundaries with exclaves.

Thanks for noticing! Done. The GeoJSON files also pass the checks on geojsonlint.com.

arch0345:
LGTM, thanks @brawer!

PR #6540, opened 2022-05-02
Added Baltimore Police Department-operated light tower entry

bgo-eiu:
There are no existing entries for Tag:man_made=mast, so it is possible I missed a step here in how to add this. It did look like there is at least some precedent for adding an item like this though, based on the presence of other man_made entries, and some entries for similar infrastructure.

In case it would be helpful to know what these look like, here is an image: https://commons.wikimedia.org/wiki/File:Penn-North_Light_Tower_by_Woodbrook_Ave.jpg

And the corresponding object on OSM: https://www.openstreetmap.org/node/9702020055

These are common to see permanently stationed at various intersections throughout the city. It would be helpful to have these tags grouped together with an iD preset through NSI.

### DIFF
--- a/data/operators/man_made/mast.json
+++ b/data/operators/man_made/mast.json
@@ -1,0 +1,22 @@
+{
+  "properties": {
+    "path": "operators/man_made/mast",
+    "skipCollection": true,
+    "exclude": {"generic": ["^mast"]}
+  },
+  "items": [
+    {
+      "displayName": "BPD Light Tower",
+      "id": "",
+      "locationSet": {"include": ["baltimore_and_dc.geojson"]},
+      "tags": {
+        "man_made": "mast",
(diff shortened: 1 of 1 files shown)


bhousel:
This seems ok, thanks @bgo-eiu 👍 
I'm going to change the locationSet a bit so it is scoped just for Baltimore, rather than extending down to DC.

bgo-eiu:
Alright - I think it's likely that all or most BPD light towers specifically are in Baltimore City, but in case it comes up in the future I'll mention that both Baltimore City and District of Columbia have operations that extend elsewhere in Maryland.

The District of Columbia city government operates a prison in Anne Arundel County which is closer to Baltimore for example. City owned+operated utilities like pipelines and cables are also not confined to the city, as well as some other things like reservoirs and natural resources.

PR #6541, opened 2022-05-03
Added a few UK desert shops

bgo-eiu:


bgo-eiu:
Sorry, dessert shops. Not sure if desert shops are common

PR #6542, opened 2022-05-03
Robidogs always have a bin, too

habi:
See the banner image on https://www.robidog.ch/ :hankey:


LaoshuBaby (approved):
图片

I agree, the vending machine itself is a bin.

PR #6550, opened 2022-05-04
Update public_bookcase.json

danieldegroot2:
Adds Croque-livres. (ignore second capital letter in commit dialog)

### DIFF
--- a/data/brands/amenity/public_bookcase.json
+++ b/data/brands/amenity/public_bookcase.json
@@ -29,6 +29,17 @@
       }
     },
     {
+      "displayName": "Croque-livres",
+      "id": "croquelivres",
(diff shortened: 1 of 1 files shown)


LaoshuBaby (changes requested):
Hey friend, by canonical we mean that this index contains the most correct tags for each kind of feature. Each item in the index is also linked to a [Wikidata](https://www.wikidata.org/) QID identifier, or it will not be suggested by editor.

(And you don't need add `"id"` manually, it will be generated by formatter program)

bhousel:
This PR looks ok - but I agree with what @LaoshuBaby said - the item won't show up as a preset until it has a `brand:wikidata` tag.  That's ok, it just means that somebody needs to make a Wikidata page for this brand.

PR #6551, opened 2022-05-05
Add diet options to Zambrero's

Georift:
I checked the menus in, AU, NZ, IE and US all regions have vegetarian and vegan options available.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -6713,7 +6713,9 @@
         "brand:wikidata": "Q18636431",
         "cuisine": "mexican",
         "name": "Zambrero",
-        "takeaway": "yes"
+        "takeaway": "yes",
+        "diet:vegetarian": "yes",
+        "diet:vegan": "yes"
       }
     },
     {


data/brands/amenity/fast_food.json
```diff
@@ -6713,7 +6713,9 @@
         "brand:wikidata": "Q18636431",
         "cuisine": "mexican",
         "name": "Zambrero",
-        "takeaway": "yes"
+        "takeaway": "yes",
+        "diet:vegetarian": "yes",
```
LaoshuBaby, line 6717: Hello, @Georift ! Before add vegeratian and vegan tag, maybe you can view #4855 before continue, because those tag is difficult to define what is available.

Georift: Thanks for the link I'll have a dive. Apologies if this isn't relevant I noticed at least one other PR adding a vegan tag so I thought it was far game. Let me dig into that post and get back to this.

Georift: Alright I guess there isn't really an appetite for having these types of tags added for restaurants / brands who aren't `vegetarian=only` or `vegan=only`.

Appreciate the quick link. I'll move over to these for some discussion.

PR #6557, opened 2022-05-06
Switch to ‘out center’ in overpass queries

jleedev:
For brands, flags, and operators, but not for transit

Advantages:

- Smaller response size
- Easier render for Leaflet's little brain
- More readable statistics for the main use case of this query:
  before, overpass turbo shows e.g. ‘pois: 64, polygons 58’.
  after, shows ‘pois: 122’ and shows a breakdown of how many of your
  pois were nodes, ways, or relations; nodes comprising the geometry of
  features are excluded.


bhousel:
Thanks @jleedev 👍

PR #6568, opened 2022-05-11
Giraffe Stop

codeinabox:
Giraffe has a sub-brand called Giraffe Stop that is more like a fast food restaurant where you need to order at the counter https://boparanrestaurantgroup.co.uk/giraffe-stop/

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -2340,6 +2340,20 @@
       }
     },
     {
+      "displayName": "Giraffe Stop",
+      "locationSet": {
+        "include": ["ae", "es", "gb"]
+      },
+      "tags": {
+        "amenity": "fast_food",
+        "brand": "Giraffe",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
If it has separate branding then usually it would be a good idea to create a separate Wikidata for it, but if there are no social media accounts available for the specific brand, you could still use the parent brand social media, in the same way Tesco Express has its own Wikidata, but still uses the standard Tesco social media accounts.

Alternatively, the Giraffe Stop logo could be added to wikimedia commons, and this file referenced to within the new Giraffe Stop Wikidata.

arch0345:
image
<a title="Giraffe restaurants, Public domain, via Wikimedia Commons" href="https://commons.wikimedia.org/wiki/File:Giraffe-logo.svg">Giraffe-logo</a>
It looks like the branding is very similar so I'm going to merge this as-is. If a separate Wikidata item is created this can always be changed in the future.

PR #6570, opened 2022-05-13
Add Ike's Love and Sandwiches

wmisener:
Popular sandwich chain on the west coast of the US, >70 locations

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -2799,6 +2799,18 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


arch0345:
Thanks @wmisener, note that entries without a `:wikidata` tag won't get generated as an iD preset. It's relatively easy to create Wikidata item for this if it doesn't have one already, you can [read this wiki article](https://github.com/osmlab/name-suggestion-index/wiki/Editing-Wikidata) for more info on this.

wmisener:
Thanks for letting me know! I'll make one and add it to the listing

PR #6574, opened 2022-05-13
Restrict Northwest Bank to states (IN, NY, OH, PA)

jleedev:
Fairly generic name, distinct from the following (at least):

- one in IA and NE
- one in Rockford, IL
- one in ID, OR, UT, and WA


arch0345:
Thanks @jleedev

PR #6587, opened 2022-05-15
Add "timpsons" to matchNames for Timpson

HandyHat:
e.g. https://www.openstreetmap.org/node/6786665930


UKChris-osm:
Thanks @HandyHat 👍

PR #6594, opened 2022-05-17
Remove name from Toby Carvery

Cj-Malone:
They can have unique names

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -4156,8 +4156,7 @@
         "amenity": "restaurant",
         "brand": "Toby Carvery",
         "brand:wikidata": "Q7811777",
-        "cuisine": "british",
-        "name": "Toby Carvery"
+        "cuisine": "british"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


LaoshuBaby (commented):
What about add a `"preserveTags": ["^name"],` after locationSet?

Cj-Malone:
Urm, I didn't know about `preserveTags` until right now, so some questions.

Is that necessary when NSI doesn't have a `name` tag? Presumably the name tag isn't changing so it wouldn't need to be preserved.

Does this actually make it more confusing when a POI changes? Say something `name=x` is replaced by Toby Carvery with a name `y`. Once you've selected Toby Carvery in iD would that keep the `name=x`?

bhousel:
Sorry @Cj-Malone , our docs on this are not great and I'd like to improve them.
The most comprehensive guidance for how this project treats the `name` tag is here: 
https://github.com/osmlab/name-suggestion-index/pull/6071#issuecomment-1015756763

It does sound like this brand falls into the middle category, where it's reasonable to set the `name` to "Toby Carvery" if it doesn't already have a value but we should leave the `name` tag alone if it does already have a value... 
This is what the `"preserveTags": ["^name"]` property does.

Cj-Malone:
OK, yeah. Sounds like the second. The vast majority are "Toby Carvery", with the website referring to them as "Toby Carvery {location}". There are a few like "Halfway House", "The Artichoke" and "White Hart" where they have unique names, but they'll still be covered in "Toby Carvery" signs. Potentially they should just be `loc_name` anyway, in which case this PR can be closed, but I've updated it in case.

PR #6598, opened 2022-05-18
Updates to fuel.json

luisluigi639:


data/brands/amenity/fuel.json
```diff
@@ -822,18 +808,6 @@
         "name": "Caltex"
       }
     },
-    {
-      "displayName": "CAMPSA",
```
LaoshuBaby, line 826: From taginfo, I noticed that there are a lot of `brand:1992=Campsa` and almost every that fuel station have `brand:wikidata=Q174747`, Is "Repsol" replaced all "CAMPSA"? If so, add "CAMPSA" as a matchName for "Repsol"

luisluigi639: Good catch, I have added matchName under "Repsol".

arch0345:
Thanks @luisluigi639

Issue #6595, opened 2022-05-17
Update Telekom stores in Germany


AlertSubject:
In Germany we have shops owned and operated by the "Deutsche Telekom AG"
These shops are only for phones and related services. 
(The company also operates infrastructure, office buildings, and more. But I'm not talking about those)

The NSI-object has the following ID:                                       [telekom-82963b ](https://nsi.guide/index.html?t=brands&k=shop&v=mobile_phone&tt=telekom)
and can be found and used In the following countries:          "include":["de","hu","ro","si","sk"]

**My issues are as follows:**

- The official name is "Deutsche Telekom" and not just Telekom, as NSI suggests (Also visible in [Wikidata](https://www.wikidata.org/wiki/Q9396))
- Updating NSI to reflect that might cause issues in other countries since im not sure if DT operates in those other countries
- I'm not sure if the generic Telekom.com address is a good fit because it increases the chance that a mapper wont add the proper address of the specific store to a POI ([Example](https://shopseite.telekom.de/nord-ost/hamburg/wandsbeker-marktstr-79))
- And lastly, would it be fine to add an alt_name to the NSI object? The shops itself are commonly called T-Shops.

**I suggest the following changes:**

- Change the name tag to "Telekom Shop" to avoid confusion with office buildings or service buildings operated by the same company
- Add an alt_name tag to enable people to find "T-Shop" when searching for them (name applies to all shops so it makes sense to add the alt_name for every shop)
- Remove the default website tag to motivate mappers to add a specific website instead. A generic website for a company as big and with a thousand different services they provide isn't all too helpful
- Maybe remove "de" from the "include" part of the current object and create a new one, specific to Germany. Mostly because I'm less familiar with how DT operates in other countries. My expertise is based on my work in Germany

I'd love to get some input and feedback on my thoughts before I start opening a PR
Thank you for your time and help :)

kjonosm:
@AlertSubject 

- Changing the name tag to "Telekom Shop" looks like a good refinement, as this name is used on their website: https://www.telekom.de/start/telekom-shops
- I personally wouldn't bother adding `alt_name=T-Shop`. It might make sense though to match shops named "T-Shop" to "Telekom Shop" with `matchNames`. You could also match shops named just "Telekom" or "Deutsche Telekom" accordingly.
- Removing website tags is, as far as I know, out of scope for NSI.
- You want to create a new item for German stores to include all your suggestions.

AlertSubject:
[...]

PR #6599, opened 2022-05-18
Update Telekom

AlertSubject:
Hi,
This PR implements the changes I have proposed [here](https://github.com/osmlab/name-suggestion-index/issues/6595)

Please let me know if this is satisfactory :)

### DIFF
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -735,15 +735,27 @@
     },
     {
       "displayName": "Telekom",
-      "id": "telekom-82963b",
+      "id": "telekom-22825c",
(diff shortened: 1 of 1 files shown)


kjonosm:
Looks great! What do you think about using `brand=Deutsche Telekom` instead? In Germany this seems to be the preferred way as it's used more often than `brand=Telekom` according to a quick check with overpass turbo.

AlertSubject:
Good catch, I completely missed that!
Deutsche Telekom is the official name :)

Alright I think my second commit got added to the PR (?)
Please check if its correct now

arch0345:
LGTM, thanks @AlertSubject!

PR #6604, opened 2022-05-19
Add Paradores de Turismo and Pousadas hotels

Crashillo:


data/brands/tourism/hotel.json
```diff
@@ -1734,6 +1734,16 @@
         "tourism": "hotel"
       }
     },
+    {
+      "displayName": "Paradores",
+      "locationSet": {"include": ["es", "pt"]},
```
1ec5, line 1739: [This Wikidata item](https://www.wikidata.org/wiki/Q1476845) only covers the Spanish chain. Should Portugal should get a separate entry linked to [Pousadas de Portugal](https://www.wikidata.org/wiki/Q2186274)?

Crashillo: No. That's a different brand. I've included portugal since [their website](https://www.parador.es/) include a hotel in Portugal. Perhaps, it doesn't worth it :thinking:, what do you think?

1ec5: OK, if they’re in Portugal too then this is correct.

Crashillo: In fact, let me add the Pousadas item as well, killing two birds with one stone :v:

1ec5 (approved):
Thanks!

data/brands/tourism/hotel.json
```diff
@@ -1734,6 +1734,16 @@
...
+      "note": "No name for Paradores's brand as each hotel is named differently.",
+      "tags": {
+        "brand": "Paradores de Turismo de España",
```
1ec5, line 1742: Based on the `displayName`, I take it that “Paradores” is the most widely recognized name and “Paradores de Turismo de España” is an official name. If that’s the case, then `brand` can be just “Paradores”. We can add a `matchName` of `paradores de truism de españa` in case someone searches for the full name.

Crashillo: Ok, I'll update it accordingly

data/brands/tourism/hotel.json
Crashillo: Done

data/brands/tourism/hotel.json
Crashillo: Done

1ec5 (approved):

PR #6623, opened 2022-05-31
Add `matchNames` to some brands

Dimitar5555:


bhousel:
Thanks @Dimitar5555 this looks ok..

FYI with [matchNames](https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#matchnames-matchtags) - any variations in case and spelling are automatically matched already, so the next time we run build some of these may go away as the matcher would consider them redundant.  

(For example we don't need to add a matchName for "easy credit" if the brand is already named "EasyCredit").

Dimitar5555:
I added it because when searching for a preset, iD doesn't find anything for "Easy Credit".
image
Also it doesn't suggest anything for banks named "Уникредит" or "Булбанк" which should be the expected behaviour?
image

bhousel:
iD's preset matcher could probably be improved to discard space and punctuation.

I do actually see these results when I type "easy" in Bulgaria:
Screen Shot 2022-05-31 at 9 43 09 AM

But unfortunately adding a space makes these results go away.

> "Уникредит" or "Булбанк"

I agree that adding these is good - they should get converted to "terms" and included in the search, after the next release of NSI.

PR #6626, opened 2022-06-01
[nz] add bus operators and other brands

k-yle:


data/operators/leisure/park.json
```diff
@@ -473,7 +473,6 @@
       "tags": {
         "leisure": "park",
         "operator": "Christchurch City Council",
-        "operator:short": "CCC",
```
Mortein, line 476: Just wondering why this was removed. `operator:short` is in use throughout NSI, and triggers a match in iD where `operator` is using the value.

CCC is the abbreviation for the council, including for [their website](https://www.ccc.govt.nz). Although not yet in NSI, would `operator:abbr` be preferred?

k-yle: oh you can add it back, at one point someone complained that I was adding a spam of useless tags like `operator:short=*`, `operator:wikipedia=*`, and `operator:type=government` (which were coming from NSI)... to a certain degree i agree, these tags are a bit useless if `operator:wikidata` exists 

would it make more sense to use `"matchNames": ["CCC"]`? then NSI will suggest changing `operator=CCC` to the correct tags without forcing everyone to add `operator:short=CCC`

arch0345: FYI there's been some recent discussion in #7954 about whether we should use `matchNames` instead of `operator:short` and its variants

PR #6637, opened 2022-06-02
Pan-African flag

### LINKED ISSUE #6629: Please Add Pan African Flag
https://en.wikipedia.org/wiki/Pan-African_flag

1ec5:
Added an entry for the Pan-African flag.

Fixes #6629.

### DIFF
--- a/data/flags/man_made/flagpole.json
+++ b/data/flags/man_made/flagpole.json
@@ -3433,6 +3433,30 @@
       }
     },
     {
+      "displayName": "Pan-African",
+      "id": "panafrican-e5dc93",
+      "locationSet": {"include": ["001"]},
+      "matchNames": [
+        "afro-american",
+        "bendera ya taifa",
+        "black liberation",
+        "international african",
+        "marcus garvey",
+        "rbg",
+        "red black green",
+        "unia",
+        "universal african"
+      ],
+      "tags": {
+        "flag:name": "Pan-African",
+        "flag:type": "cultural",
+        "flag:wikidata": "Q2029729",
+        "man_made": "flagpole",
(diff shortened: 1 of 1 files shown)


data/flags/man_made/flagpole.json
```diff
@@ -3432,6 +3432,30 @@
...
+        "unia",
+        "universal african"
+      ],
+      "tags": {
+        "flag:name": "Pan-African",
+        "flag:type": "cultural",
+        "flag:wikidata": "Q2029729",
+        "man_made": "flagpole",
+        "subject": "Pan-Africanism",
+        "subject:wikidata": "Q282739"
```
1ec5, line 3456: Based on the `flag:type=cultural`, Pan-Africanism seemed like an appropriate `subject`, whereas [Garveyism](https://en.wikipedia.org/wiki/Garveyism) would seem like a more appropriate `subject` for a `political` or `organisation` flag. That said, I’m open to changing it or removing it if there’s variation in what the flag represents when flown.

data/flags/man_made/flagpole.json
```diff
@@ -3432,6 +3432,30 @@
...
+      "matchNames": [
+        "afro-american",
+        "bendera ya taifa",
+        "black liberation",
+        "international african",
+        "marcus garvey",
+        "rbg",
+        "red black green",
+        "unia",
+        "universal african"
```
1ec5, line 3448: These names came from [this Wikipedia article](https://en.wikipedia.org/wiki/Special:PermanentLink/1091075171#Alternative_names).

bhousel:
Thanks, looks good, and thanks also for documenting the reasons behind the tags you chose - we should do more of this 👍

PR #6639, opened 2022-06-03
Additions of Multiple Hotel Chains

arnowrld:


LaoshuBaby (approved):
There are 2 question I want to ask:

1. TRIBE、Sonder、AVANI、Anantara don't match with a wikidata QID, so they won't be suggested in editor.
2. almost all of hotel you listed are using 001 as locationSet, this means they are global brand in so much countries. (But some of them are newly created, Maybe you can refine those brand. (Even some brand I haven't heard in East Asia)(But this isn't mandatory))

PR #6648, opened 2022-06-04
Add London Boroughs for Libraries

HandyHat:
Adds the London Boroughs of Hounslow, Merton, Wandsworth and Richmond Upon Thames to Library Operators, along with their Wikidata and their ISIL refs.

### DIFF
--- a/data/operators/amenity/library.json
+++ b/data/operators/amenity/library.json
@@ -874,6 +874,62 @@
       }
     },
     {
+      "displayName": "London Borough of Hounslow",
+      "id": "londonboroughofhounslow-d8f64f",
+      "locationSet": {
+        "include": ["gb-lon.geojson"]
+      },
+      "tags": {
+        "amenity": "library",
+        "operator": "London Borough of Hounslow",
+        "operator:wikidata": "Q214162",
+        "ref:isil": "GB-UkHnHL"
+      }
+    },
+    {
+      "displayName": "London Borough of Merton",
+      "id": "londonboroughofmerton-d8f64f",
+      "locationSet": {
+        "include": ["gb-lon.geojson"]
+      },
+      "tags": {
+        "amenity": "library",
+        "operator": "London Borough of Merton",
+        "operator:wikidata": "Q32504",
+        "ref:isil": "GB-UkLoML"
+      }
+    },
+    {
+      "displayName": "London Borough of Richmond Upon Thames",
+      "id": "londonboroughofrichmonduponthames-d8f64f",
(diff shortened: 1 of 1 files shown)


data/operators/amenity/library.json
```diff
@@ -873,6 +873,62 @@
         "operator:wikidata": "Q6458744"
       }
     },
+    {
+      "displayName": "London Borough of Hounslow",
+      "id": "londonboroughofhounslow-d8f64f",
```
codeinabox, line 878: Could you remove the `id` for the entries as this is done when a build is created

codeinabox (changes requested):

data/operators/amenity/library.json
```diff
@@ -873,6 +873,62 @@
...
+      "locationSet": {
+        "include": ["gb-lon.geojson"]
+      },
+      "tags": {
+        "amenity": "library",
+        "operator": "London Borough of Hounslow",
+        "operator:wikidata": "Q214162",
+        "ref:isil": "GB-UkHnHL"
```
LaoshuBaby, line 886: I'm curious about `ref:isil` and `ref:ISIL` ? Both of them have many element using according to taginfo:

图片

[OSM Wiki](https://wiki.openstreetmap.org/wiki/Key%3Aref%3Aisil) said this This key references a library by its [ International Standard Identifier for Libraries and Related Organisations](https://en.wikipedia.org/wiki/en:International_Standard_Identifier_for_Libraries_and_Related_Organisations), I want to know is every branch amenity of this library network share the same ISIL?

---

By the way, `"id"` is optional, not mandatory when you add a new entry because it can be auto generated by formatter program. You can remove it or wait for maintaner run a `npm run build` script to format it.

HandyHat: > I want to know is every branch amenity of this library network share the same ISIL?

At least in the UK, the British Library assigns an ISIL to organisations of libraries, rather than each individual branch - see: https://www.bl.uk/help/get-an-isil-for-your-library. Therefore, all the libraries run by the same borough will have the same ISIL as they are part of the same organisation.

> By the way, "id" is optional, not mandatory when you add a new entry because it can be auto generated by formatter program.

Yes, I ran `npm run build` because the wiki says I could do so if possible, but I will remove it if it is simpler.

data/operators/amenity/library.json
HandyHat: I ran `npm run build` myself because the wiki says I could do so if possible, but I will remove the ids if it is simpler.

PR #6652, opened 2022-06-09
Also match shop=kiosk for lots of stuff

peternewman:
As some things might just be smaller versions of other shops and use it

### DIFF
--- a/config/matchGroups.json
+++ b/config/matchGroups.json
@@ -17,6 +17,7 @@
       "shop/alcohol",
       "shop/beer",
       "shop/beverages",
+      "shop/kiosk",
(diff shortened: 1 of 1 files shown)


LaoshuBaby (approved):

peternewman:
I did vaguely wonder if we should differentiate between them, given a kiosk is going to be smaller so probably a smaller range, no seating etc.

Although I'd be inclined to argue it's not a great tagging choice, given you can have a kiosk form of most shops and amenities. It should almost be a `kiosk=yes` type tagging (e.g. ice cream kiosk versus larger ice cream shop).

PR #6653, opened 2022-06-09
Corrected operators for СДЭК and Почта России

Miroff:
Motivation:

1. In Russian community we're tend to use full official name as an operator tag value. We already have [JOSM preset](https://github.com/ruosm-presets/literan-moscow/blob/b875aa4ba62be96f2ec69db251275a9ce6cbfada/russian_shops.xml#L2760) and discrepancy between ID and JOSM presets causes unintentional edit wars. 

2. СДЭК is a private delivery company that uses franchise business model. Each delivery point of СДЭК is operated by local private company rather then СДЭК himself. Some mappers find useful to map real operating companies in such cases. This practice conflicts with predefined operator tag in ID templates.

@literannn is an author of JOSM preset, who asked to make this PR.


LaoshuBaby:
Hey friends, welcome to NSI and glad to see your first contribution!

> 1. In Russian community we're tend to use full official name as an operator tag value. We already have [JOSM preset](https://github.com/ruosm-presets/literan-moscow/blob/b875aa4ba62be96f2ec69db251275a9ce6cbfada/russian_shops.xml#L2760) and discrepancy between ID and JOSM presets causes unintentional edit wars.

No problem, It's good.

> 2\. СДЭК is a private delivery company that uses franchise business model. Each delivery point of СДЭК is operated by local private company rather then СДЭК himself. Some mappers find useful to map real operating companies in such cases. This practice conflicts with predefined operator tag in ID templates.

I don't think remove `operator` can solve this problem, it will only cause this problem because the entry without a brand/operator/network/flag:name/flag:type will cause program can't understand it. So the CI failed.

图片

If it is not operated directly, it may obey osmwiki's guideline: https://wiki.openstreetmap.org/wiki/Key:operator#How_to_use, so I think we can do this like #6323 or #6324

----

BTW, if possible, you can ensure that whether this `УК "СДЭК"` also the same thing:

图片

bhousel:
> Remove СДЭК according to osmlab#6653 and add to exclude list

Instead of excluding this, it sounds like we could move it to the `data/brands/amenty/post_office.json` file, since it sounds like a private company that uses the franchise model -  consistent brand, various operators (like DHL and UK Post Office).

Miroff:
Thank you!

PR #6131, opened 2022-01-27
Add Yettel

Dimitar5555:
Yettel is going to launch on 01.03.2022 in Hungary, Serbia and Bulgaria. So far it's been branded as Telenor.

PR #6656, opened 2022-06-09
remove mobile phone Yettel after #6131 merged

LaoshuBaby:
After #6131 merged, they are 2 part of Yettel in NSI:

**mobile_phone.json**
 
**telecommunication.json**

According to research from Wikipedia:

* https://en.wikipedia.org/wiki/Yettel_Bulgaria
* https://en.wikipedia.org/wiki/Yettel_Hungary
* https://en.wikipedia.org/wiki/Yettel_Serbia

I guess them are telecommunication office more than mobile phone shops  because they mainly offer mobile/fixed network, even IPTV. So I remove them in mobile_phone.json.

This may need local editor (@Dimitar5555 )'s review to confirm whether this is necessary.

### DIFF
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -1030,39 +1030,6 @@
       }
     },
     {
-      "displayName": "Yettel (Bulgaria)",
-      "id": "yettel-2a48ca",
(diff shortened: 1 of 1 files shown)


Dimitar5555:
Your observation is correct, they are telecommunication office.

PR #6659, opened 2022-06-12
Uppercase name is just an stylization

Crashillo:
Format bank name

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -46,14 +46,14 @@
       }
     },
     {
-      "displayName": "ABANCA",
+      "displayName": "Abanca",
       "id": "abanca-ce59ab",
       "locationSet": {"include": ["es"]},
       "tags": {
         "amenity": "bank",
-        "brand": "ABANCA",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/bank.json
```diff
@@ -46,14 +46,14 @@
...
       "id": "abanca-ce59ab",
       "locationSet": {"include": ["es"]},
       "tags": {
         "amenity": "bank",
-        "brand": "ABANCA",
+        "brand": "Abanca",
         "brand:wikidata": "Q9598744",
-        "name": "ABANCA",
+        "name": "Abanca",
         "official_name": "ABANCA Corporación Bancaria"
```
LaoshuBaby, line 57: Should this official_name also be changed?

LaoshuBaby (approved):

data/brands/amenity/bank.json
Crashillo: According to [its wikipedia](https://en.wikipedia.org/wiki/Abanca), don't. Official name usually comes in uppercase

PR #6668, opened 2022-06-14
Added 10 new schools to data/brands/amenity/school.json

NeroSals:
Added 10 schools from [Oak Park Elementary School District](https://en.wikipedia.org/wiki/Oak_Park_Elementary_School_District)


LaoshuBaby (changes requested):
Hello friend, welcome for your contribution, but I should point that name-suggestion-index don't intent to contain single school, it focus on canonical brand/operator things, it should be scaled, common and sharing some basic information. So I don't think those 10 certain school can be merged.

Wikipedia 

According to wikipedia, "[Oak Park Elementary School District](https://en.wikipedia.org/wiki/Oak_Park_Elementary_School_District)" is a operator, maybe you can add 1 single entry in "operator/amenity/school" catagory, and I haven't found "Oak Park" related entry in https://github.com/osmlab/name-suggestion-index/blob/main/data/operators/amenity/school.json , so you can add it there.

Also don't forget to add a `operator=Oak Park Elementary School District` to those schools in OSM.  o(*￣▽￣*)ブ！

---

BTW, because I'm not live in US so I only ask for changes, not close this PR directly, if you think I've made something wrong, you can let other local maintainer judge φ(゜▽゜*)♪

NeroSals:
Thank you for the feedback, I'll make those changes and resend the pull request.

arch0345 (approved):

PR #6672, opened 2022-06-15
Added 4 new operators to data/operators/amenity/school.json.

NeroSals:
Added [Wilmette Public Schools](https://en.wikipedia.org/wiki/Wilmette_Public_Schools_District_39), [Chicago Heights School](https://en.wikipedia.org/wiki/Chicago_Heights_School_District_170), [Glenview School](https://en.wikipedia.org/wiki/Glenview_School_District_34) and [Waukegan Community Unit School](https://en.wikipedia.org/wiki/Waukegan_Community_Unit_School_District_60).

### DIFF
--- a/data/operators/amenity/school.json
+++ b/data/operators/amenity/school.json
@@ -27955,23 +27955,77 @@
         "operator:short": "ZPS",
         "operator:type": "public"
       }
-    },
-    
+    },    
     {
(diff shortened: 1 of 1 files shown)


arch0345 (approved):
Thanks @NeroSals! We usually only specify the state/area when disambiguating between brands/operators of the same name, for example the Portland Public Schools in Maine and Oregon.

PR #6673, opened 2022-06-15
Update Font Awesome dependencies to v6.1.1

arch0345:
It looks like we're only using these dependencies for the nsi.guide website. Tested locally and everything seems to be fine.

Edit: also updated `react-fontawesome` to `v0.2.0`

Closes #6353, closes #6354, closes #6355, closes #6719


arch0345:
Yeah it looks like the social icons aren't showing up:
image

However, as I mentioned above, they aren't showing up on nsi.guide either:
image

LaoshuBaby:
I can see those share icon too:

图片

arch0345:
Oh wait, I still had my adblocker on. I can see them now:
image

image

PR #6676, opened 2022-06-16
adding Bus Stop bar

noahbeito:


arch0345:
It looks like Bus Stop only has a single location:
https://busstopbarsf.com/

Please note that this index is for brands that operate a lot of retail chain locations (things like fast food places like McDonalds).

PR #6703, opened 2022-06-25
Electronic toll collection operators in Texas

### LINKED ISSUE #6677: North Texas Tollway Authority
Add the [North Texas Tollway Authority](https://www.wikidata.org/wiki/Q13416600) to the operators tree as a ~~`man_made=toll_gantry`~~ `highway=toll_gantry` operator. NTTA is a common name for the operator that’s used on route shields and in speech; it could be `operator:short` for consistency with other entries in this tree.

### LINKED ISSUE #6679: TxTag
Add [TxTag](https://www.wikidata.org/wiki/Q7859549) to the operator tree for highway=toll_gantry

1ec5:
Added entries for each of the authorities that operate electronic toll collection (ETC) facilities in Texas. These are the companies or agencies that operate the infrastructure, as opposed to the back-office and collections operations.

I’ve included payment options that, to the best of my knowledge, are always signposted as being accepted at each of the ETC points in these systems. These tags aren’t meant to be exhaustive. I didn’t tag ETC systems that are interoperable but not signposted, such as Pikepass or TollTag outside of NTTA’s service area. These interoperability agreements can sometimes be too nuanced for this index to include. In some cases, I explicitly tagged that pay-by-mail is not accepted.

As a first pass, all these operators are currently scoped to the entire state of Texas, but they can be refined afterwards.

Fixes #6677 and fixes #6679.

### DIFF
--- a/data/operators/highway/toll_gantry.json
+++ b/data/operators/highway/toll_gantry.json
@@ -1,0 +1,132 @@
+{
+  "properties": {
+    "path": "operators/highway/toll_gantry",
+    "preserveTags": ["^name"],
+    "exclude": {"generic": ["^toll gantry$"]}
+  },
+  "items": [
+    {
+      "displayName": "Central Texas Regional Mobility Authority",
+      "id": "centraltexasregionalmobilityauthority-2ee94d",
+      "locationSet": {
+        "include": ["us-tx.geojson"]
+      },
+      "tags": {
+        "highway": "toll_gantry",
+        "operator": "Central Texas Regional Mobility Authority",
+        "operator:short": "Mobility Authority",
+        "operator:wikidata": "Q2504323",
+        "payment:license_plate": "yes",
+        "payment:txtag": "yes"
+      }
+    },
+    {
+      "displayName": "Fort Bend County Toll Road Authority",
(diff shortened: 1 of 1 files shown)


data/operators/highway/toll_gantry.json
```diff
@@ -0,0 +1,132 @@
+{
+  "properties": {
+    "path": "operators/highway/toll_gantry",
+    "preserveTags": ["^name"],
```
1ec5, line 4: Is `preserveTags` needed if none of the entries specifies a `name`? In most if not all of these systems, each individual gantry has a unique name that shows up on maps and bills.

1ec5:
> These interoperability agreements can sometimes be too nuanced for this index to include.

For example, TxTag and EZ Tag are interoperable with TollTag on NTTA-owned roads, but not in NTTA-owned airport parking lots, which also use toll gantries of a similar nature to the toll roads.

data/operators/highway/toll_gantry.json
bhousel: > Is `preserveTags` needed if none of the entries specifies a name?
It is not needed, but also no harm in having it there.

PR #6708, opened 2022-06-28
Added multilingual names to Anganwadi

bgo-eiu:
I am currently in the process of bringing up Punjabi language tags up to date so that every feature with a Punjabi name is tagged in both the Gurmukhi (`name:pa`) and Shahmukhi (`name:pnb`) scripts for the language. I set up a MapRoulette project to help with this (https://maproulette.org/browse/projects/48643) and have found that there are hundreds of Anganwadi features which have `name:pa` without `name:pnb`, as well as the `name:ur` without `name:hi`. (Urdu and Hindi are mutually intelligible but have different scripts, so every feature with an Urdu tag should ideally have a Hindi tag as well.) These multilingual names weren't in the NSI preset for Anganwadi preset but seeing as this feature is very common in India and has these languages tagged so often on the map, it would save a lot of repetitive work to make sure that's consistent for all of them.


bgo-eiu:
Updated as I forgot to add `name:en` as well to match `name`

LaoshuBaby (approved):
South Asian languages ​​are more diverse, which I think makes sense. Also you can see different kind of hokkian dialect in Taiwan's preset.

PR #6711, opened 2022-06-28
Added KFC for India + Pakistan

bgo-eiu:
Same intent as previous issues; since this is common worldwide I created a separate entry scoped to just India and Pakistan for these multilingual labels (like how there are different versions for Japan, Arabic-speaking countries etc. the default name is still the "English" name since that's the convention for a lot of American brand names in India and Pakistan but the additional languages are commonly tagged)


data/brands/amenity/fast_food.json
```diff
@@ -3208,6 +3210,35 @@
         "takeaway": "yes"
       }
     },
+    {
+      "displayName": "KFC",
```
LaoshuBaby, line 3214:
```suggestion
      "displayName": "KFC (India & Pakistan)",
```
different displayName may cause conflict, so this should be unique

PR #6720, opened 2022-06-30
Add Mochinut amenity/fast_food

hplewis:
Mochinut is a donut and boba place in the United States. It is planning
in the future to expand to South Korea and Thailand, but is not there at
the moment.


arch0345:
Thanks @hplewis and welcome!

PR #6725, opened 2022-07-03
Update Chargefox locationSet and matchNames

Samuel12321:
Update Chargefox locationSet to correctly match network operating regions 
add matchNames 'chargefox'

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -538,7 +538,8 @@
     {
       "displayName": "Chargefox",
       "id": "chargefox-943805",
-      "locationSet": {"include": ["au"]},
+      "locationSet": {"include": ["au", "nz"]},
+      "matchNames": ["chargefox"],
       "tags": {
         "amenity": "charging_station",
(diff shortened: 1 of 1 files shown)


data/operators/amenity/charging_station.json
```diff
@@ -538,7 +538,8 @@
     {
       "displayName": "Chargefox",
       "id": "chargefox-943805",
-      "locationSet": {"include": ["au"]},
+      "locationSet": {"include": ["au", "nz"]},
+      "matchNames": ["chargefox"],
```
LaoshuBaby, line 542: Hello friends, welcome to name-suggestion-index, thank you for refine the locationSet.

However, the uppercase and lowercase letter's difference don't need to add as a matchNames, see wiki: https://github.com/osmlab/name-suggestion-index/wiki/Contributing#matchnamesmatchtags-optional

(It will be removed when auto-formatter program run so don't worry)

Samuel12321: thanks for clarifying that

PR #6727, opened 2022-07-03
Update charging_station.json

Samuel12321:
Added the Zero charging network operated by meridian
https://www.meridianenergy.co.nz/for-home/zero

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -3997,6 +3997,16 @@
       }
     },
     {
+      "displayName": "Zero",
+      "id": "meridianenergy-9850e1",
+      "locationSet": {"include": ["nz"]},
+      "tags": {
+        "amenity": "charging_station",
+        "operator": "Meridian Energy",
+        "operator:wikidata": "Q6819305"
+      }
+    },
+    {
       "displayName": "ZES",
       "id": "zes-f274f9",
       "locationSet": {"include": ["tr"]},


data/operators/amenity/charging_station.json
```diff
@@ -3996,6 +3996,16 @@
...
       }
     },
+    {
+      "displayName": "Zero",
+      "id": "meridianenergy-9850e1",
+      "locationSet": {"include": ["nz"]},
+      "tags": {
+        "amenity": "charging_station",
+        "operator": "Meridian Energy",
+        "operator:wikidata": "Q6819305"
```
LaoshuBaby, line 4006: Sorry, this "Meridian Energy" had already been added to NSI

图片

Even the wikidata is same.

I'll add Zero as a matchNames if needed

LaoshuBaby (changes requested):

data/operators/amenity/charging_station.json
Samuel12321: Wasn't quite sure how the best way to do this one was. Meridian has opened up a charging network called Zero - owned and operated by meridian.

kjonosm: In this case (where the operator is different from the brand) you can instead add the new item to the respective brands folder here:

https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/amenity/charging_station.json

There you can also add both brand and operator wikidata separately.

PR #6739, opened 2022-07-05
Added Rüedu

habi:
[Rüedu](https://rüedu.ch) is a chain of small self-service stores in Switzerland, usually 'available' in containers a bit our of the city-center. Currently, there are ~20 locations, more are planned.


habi:
Thanks!

Issue #6712, opened 2022-06-28
Add John's Incredible Pizza


Adamant36:
[Wikidata](https://www.wikidata.org/wiki/Q6217292)
[Wikipedia](https://en.wikipedia.org/wiki/John%27s_Incredible_Pizza_Company)

cusine=pizza

PR #6741, opened 2022-07-05
added Johns Incredible Pizza

ricci2511:
Pull request for issue #6712

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -3034,6 +3034,19 @@
       }
     },
     {
+      "displayName": "John's Incredible Pizza",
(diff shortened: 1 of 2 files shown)


arch0345:
Thanks @ricci2511 and welcome! I moved this to amenity/restaurant since John's is similar to Chuck E. Cheese (which is also in amenity/restaurant):
https://github.com/osmlab/name-suggestion-index/blob/a60390122317f2f7cdd0616163c4f3ed2d21d8e0/data/brands/amenity/restaurant.json#L892-L906

ricci2511:
Thanks for your welcome @arch0345 . I wasn't really sure between restaurant and fast food, so thanks for the clarification and correction.

Issue #6710, opened 2022-06-28
Add Groserov


Adamant36:
Russian supermarket chain. 

Wikidata is Q112764775

PR #6742, opened 2022-07-05
added Groserov in english and russian (#6710)

ricci2511:
Pull request for issue #6710


arch0345:
Thanks @ricci2511, for multilingual entries like these, we usually only have one entry with the `displayName` being in the local language. For `brand` and `name` in `tags`, we follow https://wiki.openstreetmap.org/wiki/Multilingual_names

PR #6787, opened 2022-07-16
Localize catering brands in Hong Kong and Taiwan

Kovoschiz:
Correct some self-service restaurants in Hong Kong and Taiwan.

Changes:
1. It looks misleading for cha chaan teng food style fast food (Fairwood, Cafe de Corel, Maxim)  to be `=chinese` when  at least  half the items are Western, other Asian (depending on how to define Hong Kong curry) and even Japanese (deep-fried pork cutlet). That's better reserved for Super Super Congee & Noodles. From Taginfo, there is 1 `=hong_kong;cha_chaan_teng;fast_food`, and 1 typo `cha_chan_teng` instances, but that might be too specific. There are 42 `=*hong_kong*` and `=*Hong_Kong*` being used now. This seems acceptable. https://en.wikipedia.org/wiki/Cha_chaan_teng refers similarly to https://en.wikipedia.org/wiki/Hong_Kong_cuisine as "Hong Kong style". Hong Kong style fast food has its separate article https://zh.wikipedia.org/zh-hk/港式快餐 and was recorded by the Tourism Board before https://web.archive.org/web/20170118150702/https://www.discoverhongkong.com/tc/dine-drink/what-to-eat/local-flavours/hk-style-fast-food.jsp (dead page). A secondary option could be `cuisine=cantonese;western` (`cuisine=cantonese` is documented), considering the dishes, and prominence of siu mei. 
2. Sukiya uses only English and Japanese name in Hong Kong. Can't find “食其家" here.  https://www.sukiya.com.hk/menu/ (other minor typos corrected)
3. Bafang Dumpling is a counter-served and significantly take-out `=fast_food`, not waiter-served `=restaurant`. For `cuisine=`, the singular `=dumpling` (657 instances) is actually more widespread in Taiwan than the documented `=dumplings` (187 instances), but the word is usually used plural (cf `=fish_and_chips`) and have multiple dumplings served (cf `=wings`).

Questions:
1. Should I split them up, or will you review them together? (sorry for noob question)
2. Is there a requirement for tags to be documented?

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -8472,7 +8472,7 @@
         "brand:zh": "大家樂",
         "brand:zh-Hans": "大家乐",
         "brand:zh-Hant": "大家樂",
-        "cuisine": "hong_kong",
+        "cuisine": "cantonese;western",
         "name": "大家樂 Café de Coral",
         "name:en": "Café de Coral",
         "name:zh": "大家樂",
@@ -8493,7 +8493,7 @@
         "brand:zh": "大快活",
         "brand:zh-Hans": "大快活",
         "brand:zh-Hant": "大快活",
-        "cuisine": "hong_kong",
+        "cuisine": "cantonese;western",
         "name": "大快活 Fairwood",
         "name:en": "Fairwood",
         "name:zh": "大快活",
@@ -8965,6 +8965,22 @@
       }
     },
     {
+      "displayName": "築地銀だこ Gindaco",
+      "locationSet": {"include": ["hk"]},
+      "tags": {
+        "amenity": "fast_food",
+        "brand": "築地銀だこ",
+        "brand:en": "Gindaco",
+        "brand:ja": "築地銀だこ",
+        "brand:wikidata": "Q11603490",
+        "cuisine": "takoyaki",
+        "name": "築地銀だこ Gindaco",
+        "name:en": "Gindaco",
+        "name:ja": "築地銀だこ",
+        "takeaway": "yes"
+      }
+    },
+    {
       "displayName": "繼光香香雞",
       "id": "jandgfriedchicken-60a4d5",
       "locationSet": {"include": ["tw"]},
@@ -8992,7 +9008,7 @@
         "brand:en": "Maxim's MX",
         "brand:wikidata": "Q7213063",
         "brand:zh": "美心MX",
-        "cuisine": "hong_kong",
+        "cuisine": "cantonese;western",
         "name": "美心MX Maxim's MX",
         "name:en": "Maxim's MX",
         "name:zh": "美心MX",
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -8472,7 +8472,7 @@
         "brand:zh": "大家樂",
         "brand:zh-Hans": "大家乐",
         "brand:zh-Hant": "大家樂",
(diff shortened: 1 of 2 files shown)


LaoshuBaby:
Hey kovoschiz, thank you for this complex PR. Here are answer of your question:



> 1. Should I split them up, or will you review them together?

Both is OK, but if some part of changes in this PR raise a long discussion, you can split it to 2 parts so that some changes will pass quickly.



> 2\. Is there a requirement for tags to be documented?

About dumpling/dumplings ? Both of them have a lot of number of element using.

图片

图片

Like "hot dog" and "hot_dog" all used in OSM. If I choose, I will prefer the countable one. Besides the "dumplings" is documented in [OSMWiki](https://wiki.openstreetmap.org/wiki/Key:cuisine)

LaoshuBaby:
About the "hong_kong" as cuisine name, maybe the use of "chinese" is just because it is in the Chinese cultural circle in a broad sense

If want to specify, I think "cantonese" maybe better than "hong_kong" because it emphasizes its origin in the Cantonese cultural circle rather than geographical division, and "cantonese" is currently de facto used as a value of "cuisine". And maybe this can be used in more Hong Kong cuisine, more widely.

(This is only my personal advice about this, as a Asian maintainer in this project)

Kovoschiz:
Mainly I want to distinguish from Super Super Congee & Noodles. Looks odd to put them together. 
I change accordingly.

Kovoschiz:
Oops forgot to mention: 
4. Excludee Matsuya beef rice from HK and Macau.
5. Add Gindanco takoyaki in HK (noticed it above)

To be sure, I don't have to do something like include an empty `"id": ""` for these auto-generated items, right?

LaoshuBaby:
> To be sure, I don't have to do something like include an empty `"id": ""` for these auto-generated items, right?

Yes, I'll do this, don't worry.

data/brands/amenity/fast_food.json
```diff
@@ -9240,11 +9258,28 @@
...
+        "amenity": "fast_food",
+        "brand": "すき家 Sukiya",
+        "brand:en": "Sukiya",
+        "brand:ja": "すき家",
+        "brand:wikidata": "Q6137375",
+        "cuisine": "beef_bowl",
+        "name": "すき家 Sukiya",
+        "name:en": "Sukiya",
+        "name:ja": "すき家",
+        "int_name:zh": "食其家",
```
LaoshuBaby, line 9279: A bit curious about this, why not use "name:zh" but add a "int_" prefix 😀

Kovoschiz: They use "食其家" in Taiwan. I considered not using it as "Sukiya" is the most common name used here too, but at least some people in HK know and uses this word. A few OSM users have been adding "name:zh=食其家" which I have to point out every time. This is annoying. For other solutions, I don't think you want to write `not:name:zh=食其家` here. 
However, `int_name:zh=` creates a problem on what to do about `brand:zh=` as well, so I'm fine with dropping it.

LaoshuBaby: Another resaon is `int_name` may cause problem in other data user, because they use `name:*` to display different language's name.

But you won't feel annoy from today, because once it added to NSI and publish a new version, every time you open iD//RapiD it will check the element's tag with documented in this repository. 😀

So `name=食其家 Sukiya` will be auto suggested to `name=Sukiya` because later one is documented in this repository.

PR #6814, opened 2022-07-21
Updates to SEE power operator

cmap99:
Removed SSE Airtricity from power/generator as it is a retailer that sells electricity and does not operate any generators.
Removed SSE plc as it operates plants / generators under the SSE Renewables and SSE Thermal subsidiaries

### DIFF
--- a/data/operators/power/generator.json
+++ b/data/operators/power/generator.json
@@ -5191,16 +5191,6 @@
       }
     },
     {
-      "displayName": "SSE (Ireland and Great Britain)",
-      "id": "sse-8d3bcf",
-      "locationSet": {"include": ["gb", "ie"]},
-      "tags": {
-        "operator": "SSE",
-        "operator:wikidata": "Q493854",
-        "power": "generator"
-      }
-    },
-    {
       "displayName": "SSE (Slovakia)",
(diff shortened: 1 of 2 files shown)


LaoshuBaby:
I suggest add matchNames so that old operator will be detacted. Such as "SSE Airtricity" to renamed "SSE Thermal"

LaoshuBaby:
> For instance the example of the single SSE Airtricity power plant is on a wind farm way and it should be changed to SSE Renewables.

I see, so this isn't a simple rename, maybe you need to check current map data and modify them after this changes. 😂

cmap99:
Yea unfortunately not 😂. Yea I've been going through a few already, much more to go.

LaoshuBaby:
> Yea unfortunately not 😂. Yea I've been going through a few already, much more to go.

Besides, when I run `npm run build` as auto-formatter, this programme always add `Scottish & Southern Energy` as a entry, this is from `dist/flitered/operator_keep.json` because there are more than 10 element still using this name, and it will be collected to NSI. (If this is not precise enough, the way to solve it is modify data in OSM, and it will be re-counted in next release).

cmap99:
Okay I understand thanks!

Issue #2656, opened 2019-05-22
VinMart / VinMart+ is it OK or a real conflict?


matkoniecz:
`npm run build` reports

> "shop/supermarket|VinMart" -> duplicates? -> "shop/convenience|VinMart+"

VinMart+ is mentioned at https://vi.wikipedia.org/wiki/VinMart and according to Google Translate

> along with more than 1,400 convenience stores VinMart +. [1]

and

> VinMart is a supermarket system belonging to VinGroup, Vietnam. 

so it seems that tagging is OK, but it would be nice to have confirmation from someone who actually understands language, so summoning @1ec5 as Vietnamese speaker for confirmation

1ec5:
Previous discussion in #1306 #1403 #2231. VinMart should be `shop=supermarket`, while VinMart+ should be `shop=convenience_store`.

PR #6821, opened 2022-07-23
Rename VinMart to WinMart; merge VinMart+ with WinMart+

1ec5:
Renamed VinMart to WinMart, reflecting a rebranding that took place earlier this year. The rebranding also affected the supermarket chain’s VinMart+ convenience stores, which are now WinMart+. 919f4d0bbdca40ca4a2ab495d2d78383d757683a added a WinMart+ entry but left the VinMart+ entry in place. This PR merges the two entries.

When Grab went to systematically map the entire WinMart+ chain as part of GRABOSM/Grab-Data#117, they clicked the “Not the same VinMart+” button on each one to preserve `name=WinMart+`. Unfortunately, that has resulted in the entire chain being tagged `not:wikidata:brand=Q60245505` instead of `wikidata:brand=Q60245505`. `not:wikidata:brand=Q60245505` is currently the top value of [`not:wikidata:brand`](https://taginfo.openstreetmap.org/keys/not:brand:wikidata#values) globally by far. Each of these locations will need to be retagged manually.

/ref #2656

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -3489,17 +3489,6 @@
       }
     },
     {
-      "displayName": "VinMart+",
-      "id": "vinmart-61b913",
-      "locationSet": {"include": ["vn"]},
-      "tags": {
-        "brand": "VinMart",
-        "brand:wikidata": "Q60245505",
-        "name": "VinMart+",
-        "shop": "convenience"
-      }
-    },
-    {
       "displayName": "Viva",
       "id": "viva-856fe1",
       "locationSet": {"include": ["150"]},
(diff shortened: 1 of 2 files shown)


arch0345:
Thanks for fixing this, maybe a MapRoulette challenge can be created to retag those locations

1ec5:
I wound up creating a more specific Wikidata item for WinMart+ after all in 950885a46edcad9fd4d12d1f89749cc7ad05ebd9: [Q113236478](https://www.wikidata.org/wiki/Q113236478). So I think iD will correctly suggest adding `brand:wikidata=Q113236478` even when `not:brand:wikidata=Q60245505` is present.

Snowysauce:
The Wikidata item for WinMart+ was deleted; see #10303

1ec5:
Thanks for the heads-up. I [asked the administrator](https://www.wikidata.org/wiki/User_talk:%D7%99%D7%95%D7%A0%D7%94_%D7%91%D7%A0%D7%93%D7%9C%D7%90%D7%A7#WinMart+) to reconsider.

PR #6823, opened 2022-07-23
Warsaw Subway

starsep:


data/transit/route/subway.json
```diff
@@ -274,6 +274,18 @@
         "route": "subway"
       }
     },
+    {
+      "displayName": "Metro Warszawskie",
+      "id": "ztmwarszawa-5379d4",
+      "locationSet": {"include": ["pl"]},
+      "tags": {
+        "network": "ZTM Warszawa",
+        "network:metro": "Metro Warszawskie",
```
arch0345, line 283: It looks like this is the first use of the [`network:metro`](https://taginfo.openstreetmap.org/keys/network%3Ametro) key in NSI, which doesn't seem to be documented. Can you explain the use of this key?

starsep: I copied it from current tagging of Warsaw subway lines.
ZTM Warszawa and Metro Warszawskie are public transport companies owned by the city of Warsaw.
Metro Warszawskie operates metro network, ZTM Warszawa is a "parent company" and coordinates public transportation as a whole. I believe the intention of this tagging is to indidate that this line belongs to both city-wide network and metro-only network.

`network:metro` currently has 932 uses: https://taginfo.openstreetmap.org/keys/network%3Ametro#values

PR #6836, opened 2022-07-24
Remove GoBus

cmap99:
GoBus has recently been [purchased ](https://www.galwaydaily.com/news/transport/irish-citylink-takes-over-galway-based-gobus/) by Irish Citylink.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -4993,16 +4993,6 @@
       }
     },
     {
-      "displayName": "GoBus",
-      "id": "gobus-23677d",
-      "locationSet": {"include": ["ie"]},
-      "tags": {
-        "network": "GoBus",
-        "network:wikidata": "Q111372529",
-        "route": "bus"
-      }
-    },
-    {
       "displayName": "GoCary",
       "id": "gocary-5fe206",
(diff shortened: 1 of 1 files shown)


LaoshuBaby (changes requested):
There already have a `"irishcitylink-23677d"` entry, if this is a Mergers and acquisitions case, I suggest add a matchName to Irish Citylink's entry so that old GoBus will be apply for it.

(And avoid auto recreate a GoBus-001 entry)

cmap99:
There are currently only three bus stops (one of which is actually incorrectly tagged) and one route tagged as GoBus. I'm also not sure what Citylink plan to do with the routes that GoBus served as they both operated similar routes. I don't think a matchNames is necessary. I can removed the tags from the bus stops/ mark them as disused, and also delete the route ( I created it myself not to long ago anyway)

LaoshuBaby:
> There are currently only three bus stops (one of which is actually incorrectly tagged) and one route tagged as GoBus. I'm also not sure what Citylink plan to do with the routes that GoBus served as they both operated similar routes.

Thank you for those extend information, I wrong think this like other merge case in the past 😥

Issue #6808, opened 2022-07-20
Update "One Network Bank" to "BDO Network Bank"


ianlopez1115:
One Network Bank was renamed to BDO Network Bank, change was made sometime in 2019 (see [article](https://www.philstar.com/business/2019/09/18/1952657/bdos-one-network-bank-changes-corporate-name)). Tagging schema can be changed to the following:

amenity=bank
brand=BDO Network Bank
brand:wikidata=Q7093019
name=BDO Network Bank

LaoshuBaby:
Close because fixed in #6852

PR #6840, opened 2022-07-25
update bank and remove duplicate in favor of more detailed entry

AlertSubject:
I've tried to implement the changes requested here (https://github.com/osmlab/name-suggestion-index/issues/6808)
Please verify that I havent missed anything

cheers


AlertSubject:
looks like i somehow closed my own PR without an idea as to how I managed to do that
I'll try again later

PR #6842, opened 2022-07-26
Add MedischeZending

le-jun:
Suriname’s health operator with more than 50 centers.

### DIFF
--- a/data/operators/amenity/clinic.json
+++ b/data/operators/amenity/clinic.json
@@ -652,10 +652,11 @@
       "tags": {
         "amenity": "clinic",
         "healthcare": "clinic",
-        "operator": "Medische Zending Primary Health Care Suriname",
-        "operator:short": "MZPHCS;MZ;Medische Zending",
+        "operator": "Medische Zending",
+        "operator:official": "Medische Zending Primary Health Care Suriname",
+        "operator:short": "MZPHCS;MZ",
         "operator:type": "public",
-        "operator:wikidata": "Q25230853",
+        "operator:wikidata": "Q25230853"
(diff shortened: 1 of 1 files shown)


data/operators/amenity/clinic.json
```diff
@@ -646,6 +646,19 @@
...
       }
     },
+    {
+      "displayName": "Medische Zending",
+      "locationSet": {"include": ["sr"]},
+      "tags": {
+        "amenity": "clinic",
+        "healthcare": "clinic",
+        "operator": "Medische Zending Primary Health Care Suriname",
+        "operator:short": "MZPHCS;MZ;Medische Zending",
```
LaoshuBaby, line 656: Maybe the more common used "Medische Zending" can be operator, not short? 

This is only my personal suggestion, we still respect local mapper's opinion.

le-jun: Doing so, would there be an appropriate key for the operator official full name?

LaoshuBaby: for default name there are `alt_name/official_name` variants , add a `"^.*_"` before `"name"`.

but for `brand`, `operator` and other things, "short" and "official" are surfix to describe the namespace. 

So in OSM's naming habit, I guess "operator:official" maybe the best key. Although there haven't enough element using this key:

图片

le-jun (commented):
Modified operator and operator:official keys

data/operators/amenity/clinic.json
le-jun: Well, that’s fine by me

PR #6847, opened 2022-07-26
Add `operator:wikidata` for four networks of charging stations

brawer:


LaoshuBaby:
Thank you for those contribution! BTW, you don't need to add [P8253](https://www.wikidata.org/wiki/Property:P8253) in wikidata manually, because when maintainer publish a new version, `npm run wikidata` is in routine [checklist](https://github.com/osmlab/name-suggestion-index/wiki/Release-Checklist), and it will add property automatically.

♥

PR #6850, opened 2022-07-26
Split German and Dutch “Cigo” newsagents

brawer:
They are separate companies with different owners and different
brand identities. Not sure if they’ve been related at some point
in the past (possibly before 2010); as of 2022, they’re separate.

### DIFF
--- a/data/brands/shop/newsagent.json
+++ b/data/brands/shop/newsagent.json
@@ -23,7 +23,7 @@
       }
     },
     {
-      "displayName": "Cigo",
+      "displayName": "Cigo (Germany)",
       "id": "cigo-dbf769",
       "locationSet": {"include": ["de"]},
       "tags": {
@@ -34,7 +34,7 @@
       }
     },
     {
-      "displayName": "Cigo",
+      "displayName": "Cigo (Netherlands)",
       "id": "cigo-401e3d",
       "locationSet": {"include": ["nl"]},
       "tags": {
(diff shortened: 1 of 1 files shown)


brawer:
How to fix the build failure? The index build fails because of duplicate `displayName`, but those duplicates are actually in a different `locationSet`.

data/brands/shop/newsagent.json
```diff
@@ -22,10 +22,21 @@
...
     },
+    {
+      "displayName": "Cigo",
+      "id": "cigo-dbf769",
+      "locationSet": {"include": ["de"]},
+      "tags": {
+        "brand": "Cigo",
+        "brand:wikidata": "Q113290782",
+        "name": "Cigo",
+        "shop": "newsagent"
```
LaoshuBaby, line 33: Looks like this brand sells both newspaper and tobacco?

图片

LaoshuBaby:
> but those duplicates are actually in a different `locationSet`.

You can change the `displayName` to things like "Cigo (Germany)" or "Cigo (Netherland)"

data/brands/shop/newsagent.json
brawer: Thanks for noticing! Added tag `tobacco=yes`.

brawer:
Thanks for the review! Made `displayName` unique, and added `tobacco:yes` to the German chain.

PR #6851, opened 2022-07-26
Add Epic Wings N' Things chicken wings chain

hplewis:
I wasn't sure if match names were necessary for substring versions the name, but I put two common variations. I also am not sure if there are any other match names necessary for variations on `N'` (ie `&` and `and`).

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -1918,6 +1918,23 @@
       }
     },
     {
+      "displayName": "Epic Wings N' Things",
+      "id": "epicwingsnthings-4d2ff4",
+      "locationSet": {"include": ["us"]},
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
For matchNames you can see this: https://github.com/osmlab/name-suggestion-index/wiki/Contributing#matchnamesmatchtags-optional

You don't need to add matchNames for `&` and `and`, but this time `'` was consider as a character, not a punctuation.

Thank you for this contribution. 

By the way, US have states geojson, if some brand only open in specific states you can write locationSet like this: `"locationSet":{"include":["us-ca.geojson"]},`

hplewis:
Thanks! And yeah while the chain is based in SoCal, there is one location in AZ I believe, so I did `us`.

PR #6855, opened 2022-07-28
add some more New Zealand brands

k-yle:


data/brands/amenity/bank.json
```diff
@@ -3224,6 +3213,17 @@
         "name": "BNP Paribas Fortis"
       }
     },
+    {
+      "displayName": "BNZ",
+      "id": "bankofnewzealand-279fc5",
+      "locationSet": {"include": ["nz"]},
```
LaoshuBaby, line 3219: I suggest that keep the displayName, and add a `"matchNames":["BNZ"],`

Other changes looks good to me.

PR #6860, opened 2022-08-01
Tag GB recycling containers more appropriately

tomhukins:
This change tags Salvation Army recycling points as containers and
removes names for containers.

People don't generally assign names to recycling containers.  If someone
says "I'm going to Oxfam" you assume they mean a shop, not a container:
https://wiki.openstreetmap.org/wiki/Names#Name_is_the_name_only

### DIFF
--- a/data/brands/amenity/recycling.json
+++ b/data/brands/amenity/recycling.json
@@ -258,13 +258,27 @@
     },
     {
       "displayName": "The Salvation Army",
-      "id": "thesalvationarmy-417047",
-      "locationSet": {"include": ["gb", "us"]},
+      "id": "thesalvationarmy-a9e339",
+      "locationSet": {"include": ["gb"]},
       "matchNames": ["salvation army"],
       "tags": {
         "amenity": "recycling",
         "brand": "The Salvation Army",
         "brand:wikidata": "Q188307",
+        "recycling_type": "container",
+        "recycling:clothes": "yes"
+      }
+    },
+    {
+      "displayName": "The Salvation Army (US)",
+      "id": "thesalvationarmy-f8d7ba",
+      "locationSet": {"include": ["us"]},
+      "matchNames": ["salvation army"],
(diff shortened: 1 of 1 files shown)


data/brands/amenity/recycling.json
```diff
@@ -266,7 +265,7 @@
         "amenity": "recycling",
         "brand": "The Salvation Army",
         "brand:wikidata": "Q188307",
-        "name": "The Salvation Army",
+        "recycling_type": "container",
```
1ec5, line 268: Good catch! It should’ve been a `container` all along.

This entry is currently shared by the UK and U.S. There are a lot of other entries here for donation bins of other U.S. organizations, so there should probably be a discussion in the U.S. community about whether to keep `name` on them. In the meantime, you can copy the entry and have one be for the UK and the other for the U.S. to sidestep the broader issue.

(With as many organizations as we have running donation bins over here, people may well use the brand names as names, in much the same fashion as the gas stations they’re usually placed in front of. However, I’m pretty sure the Salvation Army donation bins go by a distinct name, something like Small Item Donation Center. I’m unsure if it varies by region.)

tomhukins: Thank you for your helpful comments.  I'm happy to update my patch so that the US and GB have separate entries for Salvation Army recycling points following each country's conventions if there's general agreement that's the best approach.  What do others think?

codeinabox:
Thank you for looking into this. I've noticed some other UK recycling providers - Ecolight, Islamic Relief, and TRAID - still have the name tag so should that be removed?

tomhukins:
> Thank you for looking into this. I've noticed some other UK recycling providers - Ecolight, Islamic Relief, and TRAID - still have the name tag so should that be removed?

Perhaps.  I considered changing these in my original pull request, but I don't know enough about these organisations and their recycling points.

PR #6863, opened 2022-08-02
adding Wikidata Tag to Buvette

KarenDouglas:


LaoshuBaby:
Hello, thank you for this contribution! I noticed that Wikidata and its' website all display as "La Buvette", but current brand is "Buvette". Are they not the same brand or should we rename the existing "Buvette" to "La Buvette"?

arch0345:
It looks like the Wikidata item you linked is for a single restaurant in Nebraska. Please note that we only add brands/operators of multiple locations to NSI (think chains like McDonald's or school districts)

KarenDouglas:
I will keep that in mind for my next contribution! Thanks

PR #6876, opened 2022-08-07
Update convenience.json

DianAgesson:
Added Coles Central and Woolworths Metro (AU)

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -880,17 +880,9 @@
         "brand": "Coles Express",
         "brand:wikidata": "Q5144653",
         "name": "Coles Express",
-        "shop": "convenience"
-      }
-    },
-        {
-      "displayName": "Coles Central",
-      "locationSet": {"include": ["au"]},
-      "tags": {
-        "brand": "Coles Central",
-        "brand:wikidata": "Q5144653",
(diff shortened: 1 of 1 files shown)


data/brands/shop/convenience.json
```diff
@@ -3692,6 +3702,16 @@
...
       }
     },
+    {
+      "displayName": "Woolworths Metro",
+      "locationSet": {"include": ["au"]},
+      "tags": {
+        "brand": "Woolworths Metro",
+        "brand:wikidata": "Q111772555",
+        "name": "Woolworths Metro",
+        "shop": "convenience"
```
andrewharvey, line 3712: In my opinion these are fully fledged shop=supermarket, the wiki describes convenience as "A small local shop carrying a small subset of the items you would find in a supermarket", at least the Woolies Metro I go to has pretty much everything compared to the full Wollies, just lacking the deli.

andrewharvey:
We should consider adding `operator=Woolworths Group` so you can see all the supermarket brands part of the Woolworths umbrella. Though not essential as wikidata should also store this information.

DianAgesson:
Thanks @andrewharvey ,

As discussed in the Discord chat, I'll make the changes to Woolworths Metro/Coles Central

PR #6880, opened 2022-08-08
Add Niles Township High School to school.json

e-papanicolas:
Hi ! This is my first PR contribution, so please let me know if I did this incorrectly. Thanks :)

### DIFF
--- a/data/brands/amenity/school.json
+++ b/data/brands/amenity/school.json
@@ -152,18 +152,6 @@
       }
     },
     {
-      "displayName": "Niles Township High School District 219",
-      "id": "",
-      "locationSet": {"include": ["us", "il"]},
-      "tags": {
-        "amenity": "school",
-        "brand": "Niles Township High School District 219",
-        "brand:wikidata": "Q7037048",
-        "operator:type": "public",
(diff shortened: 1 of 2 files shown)


data/brands/amenity/school.json
```diff
@@ -151,6 +151,18 @@
         "operator:type": "private_non_profit"
       }
     },
+    {
+      "displayName": "Niles Township High School District 219",
+      "id": "",
+      "locationSet": {"include": ["us", "il"]},
```
arch0345, line 157:
```suggestion
      "locationSet": {"include": ["us-il.geojson"]},
```
For states, we use this format: `us-[state].geojson`
See https://github.com/osmlab/name-suggestion-index/wiki/Contributing#locationset-required (under 'States within the United States')

In the future, you can use https://location-conflation.com/?referrer=nsi to test out locationSets

arch0345 (commented):
We're currently only using schools in the `brands` directory for schools that all have the same name, which doesn't seem to be the case here. I'll move this entry to the `operators` directory.

arch0345:
Thanks @e-papanicolas and welcome!

PR #6888, opened 2022-08-09
Add Platte Valley Bank

xorgy:


xorgy:
Not to be confused with Platte Valley Bank (Missouri).

arch0345:
Thanks @xorgy and welcome! It looks like they might also have some locations in Colorado: https://pvbank.com/locations/

xorgy:
> Thanks @xorgy and welcome! It looks like they might also have some locations in Colorado: https://pvbank.com/locations/

It's a bit complicated, those are "in network" but they don't belong to the same brand.

jleedev:
Also not to be confused with this single-branch bank in the other half of Nebraska: https://www.pvbonline.com/home/about/hours

jleedev:
Platte Valley Bank of Missouri might no longer exist: https://www.centralbank.net/central-bancompany/recent-news/central-bancompany-platte-valley-bank-merger/

1ec5:
> Also not to be confused with this single-branch bank in the other half of Nebraska: https://www.pvbonline.com/home/about/hours

[Changeset 124,703,646](https://www.openstreetmap.org/changeset/124703646) maps this bank and preemptively tags it with `not:brand:wikidata` to avoid confusion.

PR #6890, opened 2022-08-09
Update flagpole.json

DianAgesson:
Add South Sea Islander Flag

### DIFF
--- a/data/flags/man_made/flagpole.json
+++ b/data/flags/man_made/flagpole.json
@@ -281,7 +281,7 @@
       }
     },
       {
-      "displayName": South Sea Islanders",
+      "displayName": "South Sea Islanders",
       "locationSet": {"include": ["001"]},
       "tags": {
         "flag:name": "South Sea Islanders",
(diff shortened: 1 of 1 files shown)


arch0345 (commented):
Looking at the other flag entries, it appears that 'Flag'/'Flag of' is omitted in `flag:name`.

data/flags/man_made/flagpole.json
```diff
@@ -279,6 +279,19 @@
         "subject": "Tasmania",
         "subject:wikidata": "Q34366"
       }
+    },
+      {
+      "displayName": "AU - South Sea Islanders",
+      "locationSet": {"include": ["001"]},
+      "tags": {
+        "country": "AU",
```
1ec5, line 287: `country` should is for national flags, not cultural flags. It isn’t necessary to include “AU” in the display name either, unless there’s a distinct flag called “South Sea Islanders” that’s used elsewhere.

DianAgesson:
Thanks @1ec5  and @arch0345 

I've adjusted the file now

PR #6904, opened 2022-08-11
Update police.json

monk198230:
Added Merseyside Police
id needs generating


LaoshuBaby:
Thank @monk198230 for making your first contribution to NSI!

PR #6910, opened 2022-08-11
Add Blue Federal Credit Union

xorgy:


LaoshuBaby:
Looks good, thank you for your contribution.

Besides, you don't need to add it manuallly because it will be add automatically when maintainer run validator.

图片

xorgy:
@LaoshuBaby Thanks, wasn't sure whether it was automatic or not so I was going ahead...

PR #6923, opened 2022-08-13
Add Thirsty Camel wikidata reference

Georift:


UKChris-osm:
Thanks @Georift - do Thirsty Camel have any social media accounts (Twitter / Facebook) that can be linked to via the Wikidata page so the NSI can pull in the logo?

Georift:
> Thanks @Georift - do Thirsty Camel have any social media accounts (Twitter / Facebook) that can be linked to via the Wikidata page so the NSI can pull in the logo?

They do but they seem state specific, would that be interesting?

UKChris-osm:
> They do but they seem state specific, would that be interesting?

If they don't have a national Facebook / Twitter account available then a state one would be the next best option, anything with a recognisable logo would be great.

PR #6930, opened 2022-08-14
Adding regions of France (geofencing)

shawn12:
Adding geofencing features for regions of France


UKChris-osm:
Thanks @shawn12 

Do these follow the [ISO_3166-2:FR](https://en.wikipedia.org/wiki/ISO_3166-2:FR) convention?

I won't merge this just yes as others may wish to comment about if the geozones here are a bit too complex, as the NSI tries to use small / basic geofencing so the sizes and computation used to generate the geozone isn't to intense.

arch0345:
As explained in https://github.com/osmlab/name-suggestion-index/issues/4784#issuecomment-933577987, geojsons should be as simple since iD has to download all of these. I would try to simplify these to less than 100 points.

shawn12:
Hi @UKChris-osm and @arch0345. Yes it follows ISO_3166-2:FR convention.

I tried to simplify as much as I could but I think it is too complicated for France as regions are pretty big and are not squares like USA states...
France regions
It is easy to simplify near coasts but not really for inner boundaries.

At first, I started this work in order to fix some issues in France with homonyms on bus companies names.
Currently, iD replaces good data by an homonym network : 
OSM ID wrong suggestion

What advice can you give me to help me solve this at best?

UKChris-osm:
Thank you for added these @shawn12 👍

Issue #4841, opened 2021-01-04
Revolution Laundry (was: add laundry.json)


kjonosm:
Currently we only support `shop=dry_cleaning`.
Can someone add `laundry.json` to collect self service laundry shops [`shop=laundry`](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dlaundry)?

bhousel:
Thanks for opening this - this one is interesting..
It looks like we specifically exclude this kind of shop, but I don't remember the details why.
https://github.com/osmlab/name-suggestion-index/blob/47661e89ea920a9a40a3fcbb6739b44c32765dab/dist/trees.json#L50


This is what we've collected from the OSM planet:
https://github.com/osmlab/name-suggestion-index/blob/47661e89ea920a9a40a3fcbb6739b44c32765dab/dist/collected/names_all.json#L8518-L8533

As far as I can tell, they are all either generic words or there are a few like "Speed Queen" and "Laverap" which I think are stores which sell washing machines, and they should probably instead be tagged as [`shop=appliance`](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dappliance).

kjonosm:
In Germany we have a franchise called Eco-Express ([overpass-link](https://overpass-turbo.eu/s/11Zo), [website](https://www.mein-waschsalon.de)) that lots of (former) students probably know. I don't know about other countries though...

kjonosm:
> "Speed Queen" and "Laverap" which I think are stores that sell washing machines

Actually they are indeed laundry services, see [Speed Queen](https://franchise.speedqueen.com) and [Laverap](http://www.laverap.com) websites.

bhousel:
> Actually they are indeed laundry services, see [Speed Queen](https://franchise.speedqueen.com) and [Laverap](http://www.laverap.com) websites.

cool!, thanks I didn't realize that on first look..
I'll merge in the `launday.json` file now.

UKChris-osm:
How would something like [Revolution Laundry](https://www.revolution-laundry.com/) appear in the index? Still under shop, or maybe as a vending machine?

bhousel:
> How would something like [Revolution Laundry](https://www.revolution-laundry.com/) appear in the index? Still under shop, or maybe as a vending machine?

whoa, I didn't know about those.. 

Here's what I see on [`shop=laundry`](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dlaundry) wiki page today:

[...]

PR #6933, opened 2022-08-14
Add Revolution Laundry

Karthoo:
Add Revolution Laundry (Q113516904)

I was not sure whether to use [shop=laundry](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dlaundry) or [amenity=washing_mashine](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dwashing_machine). The latter seems to be more appropriate, but is barely used.

### DIFF
--- a/data/brands/shop/laundry.json
+++ b/data/brands/shop/laundry.json
@@ -64,6 +64,18 @@
       }
     },
     {
+      "displayName": "Revolution Laundry",
+      "id": "revolutionlaundry-ca17f3",
+      "locationSet": {"include": ["001"]},
+      "tags": {
+        "brand": "Revolution Laundry",
+        "brand:wikidata": "Q113516904",
+        "name": "Revolution Laundry",
+        "self_service": "yes",
+        "shop": "laundry"
+      }
+    },
+    {
       "displayName": "Speed Queen",
       "id": "speedqueen-ca17f3",
       "locationSet": {"include": ["001"]},


UKChris-osm:
Based on [#4841](https://github.com/osmlab/name-suggestion-index/issues/4841) I would suggest [amenity=washing_machine](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dwashing_machine) is the correct usage, despite it's current lack of use.

Karthoo:
@UKChris-osm Thanks for linking that discussion, I didn't notice it. My main concern with [amenity=washing_machine](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dwashing_machine) was that it isn't even used by editor presets, yet. Should we first implement a preset for the tag into iD etc before adding this brand?

UKChris-osm:
I hadn't realised [amenity=washing_machine](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dwashing_machine) wasn't available within iD - I think that's a requirement for the NSI data to be available as a preset in that case.

Actually now that I think about it more, these Revolution machines have a washing machine and dryer as one unit don't they?

That might mean sticking with [shop=laundry](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dlaundry) would perhaps make more sense, rather then limiting them to [amenity=washine_machine](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dwashing_machine), or splitting them with [amenity=dryer](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Ddryer) as well.

UKChris-osm:
Thanks @Karthoo 👍 

I've decided to merge this as there are already a few [Revolution Laundry](https://overpass-turbo.eu/s/1l0A)'s tagged as shops.

PR #6938, opened 2022-08-15
Adding Agglobus (Rodez) and disambiguation

shawn12:
Purpose is to fix iD wrong suggestions due to homonyms.

OSM ID wrong suggestion


bhousel:
LGTM, thanks @shawn12 👍

PR #6939, opened 2022-08-15
Post office

SafetyIng:
Updated german post office brands and service provider based on https://wiki.openstreetmap.org/wiki/Key:post_office/postal_service_provider


bhousel:
I am not familiar with this style of tagging and I can't find any documentation for `service_provider` tag.. 
We should probably hold off on this until we're really sure that it is accepted.

SafetyIng:
> I am not familiar with this style of tagging and I can't find any documentation for `service_provider` tag..

https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dpost_office - here is the documentation. I have made a mistake an correct it to post_office:service_provider.



> We should probably hold off on this until we're really sure that it is accepted.

https://wiki.openstreetmap.org/wiki/Proposed_features/shop_as_post-partner - here is the proposal, that was accepted :)

bhousel:
I can see in the files changed that there are things in here that just won't work at all - the items in the `brands/amenity/post_office.json` file all needs to have an `amenity=post_office` tag.

I've written a bunch in other tickets on this repo why the "shop as post partner" proposal doesn't really make sense to me - our software just can't handle features being mapped this way.  It's better when the post office is operating within a shop to just put a node inside the shop where the post office counter is (same as is already accepted practice for other shop-in-a-shop features like pharmicies or coffee shops).   

For example, if just because a Starbucks appears inside every Target department store, we don't say that Starbucks is their "coffee partner".   If there's a Dunkin Donuts inside a Walmart, that doesn't make them the "donut partner" - you just place nodes for those things where they exist.

bhousel:
Again... our software _can not do this_:

```js
    {
      "displayName": "Deutsche Post - Postfiliale in anderem Shop",
      "matchNames": [
        "deutsche post ag",
        "deutsche post dhl",
        "deutsche post filiale",
        "deutsche postbank ag",
        "postbank"
      ],
      "tags": {
        "post_office": "post_partner",
        "brand": "Deutsche Post",
        "post_office:service_provider": "Deutsche Post",
        "post_office:service_provider:wikidata": "Q157645",
        "post_office:letter_from": "yes",
        "post_office:parcel_from": "yes",
        "post_office:parcel_pickup": "yes",
        "post_office:parcel_to": "yes",
        "post_office:stamps": "yes",
      }
    },
```

The matcher does not work the way you think it does.  It can't just match arbitrary tags.  
It is an in-memory index lookup that requires a [key,value,name] to work, like [`amenity`,`post_office`,`Deutsche Post`].
The lookup returns an item, and that item needs to have a `brand:wikidata`.
This is why the items in the index are organized into files like `brands/amenity/post_office.json`.

If you want to add `post_partner` tags to supermarkets, or PC dealers, kiosks, laundries, that is ok, and you can use any tags you like in OSM, but you have to do it manually.  NSI can't do it for you.

Lukas458:
Hi,

In the course of this, I noticed that we are actually - at least speaking for German - heading for a small problem with the NSI.
The operators "Deutsche Post" and especially Hermes, DPD and GLS have templates for amenity=post_office in the NSI, but they do not operate (at least in Germany) any, really no amenity=post_office.

All their shop services are taken over by commissioned third parties, for example in kiosks.

Conversely, this also means that all existing amenity=post_office in germany with these operators are, strictly speaking, incorrectly - or rather outdated - tagged according to the new tagging.
However, the templates for amenity=post_office that exist in the NSI tempt people to continue tagging the partner post and parcel shops like this, for example:
https://www.openstreetmap.org/node/10047669446

Do you think it would be okay to exclude at least Germany from these corresponding templates for amenity=post_office in the NSI?

Unfortunately, this measure does not give users the opportunity to use the new key post_office with the NSI. But it would at least prevent mistagging, because there are no longer any independent amneity=post_office, even if they are still tagged. The NSI should then at least not promote the new-entry of these.

Lukas458:
Oh, very sorry @bhousel just read your comment now.

The problem is that the NSI now counteracts the accepted (!) post-office tagging.

1ec5:
> The operators "Deutsche Post" and especially Hermes, DPD and GLS have templates for amenity=post_office in the NSI, but they do not operate (at least in Germany) any, really no amenity=post_office.

In Germany, is the closest thing to a post office something like [this Postbank branch](https://commons.wikimedia.org/wiki/File:Hagen,_Postbank_Berliner_Platz_18.jpg) that prominently advertises Deutsche Post services (via the 📯 and DHL signs)?

[](https://commons.wikimedia.org/wiki/File:Hagen,_Postbank_Berliner_Platz_18.jpg)

As @bhousel noted, an NSI entry is represented in editors as a preset, as a way to describe the essence of a feature, whereas a service provider with whom the feature contracts doesn’t rise to the level of a preset. [I raised this as a concern](https://wiki.openstreetmap.org/wiki/Talk:Proposed_features/shop_as_post-partner#Store_within_a_store) last year when the postal partner proposal was just getting started.

An NSI entry results in certain conveniences like consistent tagging and prominent placement in preset search results with a recognizable logo. It sounds like you’d like similar conveniences for describing certain branded details of a feature that don’t rise to the level of a preset. I’d very much appreciate iD displaying likely matches and their logos when I type something into the Operator field (as in #5734), but without cluttering preset search results with every possible operator. It would be something in between a full-fledged preset and a localized combo box string.

Maybe similar functionality for a separate Postal Partner field would be what you’re looking for. But NSI can’t really implement something like that until some downstream projects like iD add support. And before we go there, none of the presets have any fields for `post_office=post_partner` or `post_office:*=*`, so I’d suggest starting a discussion about that in the [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema/) repository first.

SafetyIng:
> In Germany, is the closest thing to a post office something like [this Postbank branch](https://commons.wikimedia.org/wiki/File:Hagen,_Postbank_Berliner_Platz_18.jpg) that prominently advertises Deutsche Post services (via the 📯 and DHL signs)?

The problem here in Germany is: Until about the mid-1990s, there was the "Deutsche Post", which operated both mail, telephone and the "Postbank" as one authority. With privatization, these branches were split into Postbank AG (banking, now part of Deutsche Bank), Deutsche Telekom AG (telecommunications) and Deutsche Post AG (postal services). Almost all Postbank branches also offer Deutsche Post AG services. The remaining branches are not operated by Deutsche Post AG itself, as is common elsewhere, but by third parties. These can, for example, only work for the parcel division of Deutsche Post (under the DHL brand), or only for the letter division... And then there are also different characteristics. Some run a small kiosk, some use it as an additional service in a stationery store, and so on.....

SafetyIng:
@bhousel - okay, then I would hold back this pull Request on only remove the german post companies..... 

I would follow @1ec5 suggestion to contact the [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema/) first. 
Or is there a nother chance, that the nsi would support the post-partner elements?

bhousel:
> I would follow @1ec5 suggestion to contact the [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema/) first.
> Or is there another chance, that the nsi would support the post-partner elements?

I don't see it as a possibility.  
 `amenity=post_office` is what some people would call a "physical" tag - it says what a thing _is_.
 `post_office=post_partner` is what some people would call an "attribute" tag - it says what a thing _has_.

Physical tags can stand alone as features and be rendered directly. Attribute tags need to be attached to something else.

It's the same reason that you can't just draw an area and tag it with a `sport=` tag or a `surface=` tag.  
These tags need another physical tag to say what the thing actually is.  

In iD tagging, the "presets" are for physical features, the "fields" are for attributes attached to those features.  
Yes it's all tags, but some tags can stand alone and some can't.    

For now, NSI can only work with the sorts of things that can be a preset.  We might eventually change that?

Adamant36:
> I don't see it as a possibility.

I'm closing this since it seems to be something that can't be done with how things are currently implemented. Feel free to re-open it if or when that changes though 👍

PR #6943, opened 2022-08-16
Add Sparkassen in Baden-Württemberg

Karthoo:
This PR adds the 49 different Sparkassen banks in the German state of Baden-Württemberg.
All of them with their own bounding polygon.


Karthoo:
Do I have to add the boundaries to another file than the actual geojson or why can't the compiler find it?

data/operators/amenity/atm.json
```diff
@@ -0,0 +1,585 @@
...
+  "properties": {
+    "path": "operators/amenity/atm",
+    "skipCollection": true,
+    "exclude": {"generic": ["^atm$"]}
+  },
+  "items": [
+    {
+      "displayName": "Bezirkssparkasse Reichenau",
+      "id": "bezspkreichenau",
+      "locationSet": {"include": ["bezspk-reichenau"]},
```
bhousel, line 11: Things like this need to be referred to with the .geojson in the name, like `bezspk-reichenau.geojson`..
(This is to differentiate them from recognized locations [from country coder](https://ideditor.codes/)).

bhousel:
Looks like it should mostly work if the locationSets include the `.geojson` in the identifier.

Also - ATMs by default get generated automatically from the banks, so I'm not sure whether these entries should be in both atm.json and bank.json.  My preference would be to get the banks added with a first PR and skip the ATMs for now.

Karthoo:
Thank you for your input. I added the `.geojson` extensions and removed the ATMs. The `network` tags I added to the ATMs would have been nice to have, though. Is there a possibility to add these automatically? They all had the same value `network=Sparkassen-Finanzgruppe` and `network:wikidata= Q2307760`.

coolultra1:
Short reminder we kinda already had the Sparkassen-Drama in #3583

Karthoo:
The main difference of my PR to the previous ones is that I also added the bounding boxes for the areas of service of the banks.

The only question that remains is how to name the banks.
> Closing until there's consensus.

This is the reason why the previous PR has been closed. I think that waiting for the German forum to find a consensus about the naming is pointless as this will never happen from my experience. 
Since NSI supports brands also without the `name` tag for a while now, we could just remove the tag.

kjonosm:
To sidestep the 
> Sparkassen-Drama in https://github.com/osmlab/name-suggestion-index/pull/3583

it might be a good idea to remove the `name` tag from all items or at least add `"preserveTags": ["^name"]` to all items. Besides that I think it's a good addition to the index.

PR #6945, opened 2022-08-16
French train network regions, remove obsolete network

shawn12:
Sorry, I did a technical mishap with Git on my previous pull request.

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -1771,7 +1771,10 @@
       "locationSet": {
         "include": ["fr-occ.geojson"]
       },
-      "matchNames": ["lio ter occitanie"],
+      "matchNames": [
+        "lio ter occitanie",
+        "ter midi-pyrénées"
+      ],
       "tags": {
         "network": "TER Occitanie",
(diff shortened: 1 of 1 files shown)


data/transit/route/train.json
```diff
@@ -1725,22 +1737,12 @@
         "route": "train"
       }
     },
-    {
-      "displayName": "TER Midi-Pyrénées",
-      "id": "termidipyrenees-46b9eb",
-      "locationSet": {"include": ["fx"]},
-      "tags": {
-        "network": "TER Midi-Pyrénées",
```
arch0345, line 1733: It looks like this is still in use (https://taginfo.openstreetmap.org/tags/network=TER%20Midi-Pyr%C3%A9n%C3%A9es). Since this was merged into TER Occitanie, maybe this could be added as one of the [`matchNames`](https://github.com/osmlab/name-suggestion-index/wiki/Contributing#matchnamesmatchtags-optional)

shawn12: Good idea, I will do this when I have some time. Thanks for the suggestion :) !

PR #6948, opened 2022-08-18
Add Irish location to Pret A Manger

cmap99:
Pret A Manger are set to expand into the Republic of Ireland, [source](https://businessplus.ie/news/pret-a-manger-ireland/). With the first store opening on August 26th


UKChris-osm:
Thank you @cmap99 - I see from the article you linked to they are also expanding into India, Spain and Portugal, so I have added these locations as well!

PR #6960, opened 2022-08-22
Add 'Türkiye Petrolleri' brand

wschildbach:
This PR adds the 'Türkiye Petrolleri' brand of fuel stations in Turkey. There are currently 48 nodes with this name occurring in Turkey.

I have not been able to run `npm build` yet; hope to be able to figure this out soon.

Some stations have been labelled as "TP" and wikidata also has "TPPD"; hence, I am adding these two as matchNames. Not sure whether such short matchNames make sense -- I am looking for guidance on this.

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -4678,6 +4678,17 @@
       }
     },
     {
+      "displayName": "Türkiye Petrolleri",
+      "matchNames":["TP","TPPD"],
+      "locationSet": {"include": ["tr"]},
+      "tags": {
+        "amenity": "fuel",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
> I have not been able to run `npm build` yet; hope to be able to figure this out soon.

Are you typing "_npm build_"? If so, the command you need is "_npm run build_".

wschildbach:
> > I have not been able to run `npm build` yet; hope to be able to figure this out soon.
> 
> Are you typing "_npm build_"? If so, the command you need is "_npm run build_".

Just to close the loop, I had troubles installing the right version of npm but have now figured it out, see #6961 . Thanks for the help!

PR #6961, opened 2022-08-22
add Turkish fuel station brands Sunpet and termo

wschildbach:
With this addition, all Turkish fuel station brands that had more than 40 mentions in OSM nodes in Turkey are mentioned in nsi.

I chose 40 as cut-off (rather than the informal 50 threshold) because there is good evidence that there are plenty more stations of each brand; fuzzy searching flags up more than 50 stations for each (and many have either not been mapped or not tagged with a name).

Sunpet is a brand of OPET. I edited this into wikidata as a separate entity, so it goes under a different identifier than OPET.

The PR also includes the output of `npm run build` though only for fuel.json (disregarding any changes to other files). `npm run build` throws a few warnings but nothing connected to these changes, as far as I can see.

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -4302,6 +4302,7 @@
     },
     {
       "displayName": "Sunpet",
+      "id": "sunpet-ec7588",
       "locationSet": {"include": ["tr"]},
       "tags": {
         "amenity": "fuel",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Thank you for maintance Turkish brand! Infact there are threshold in nsi-collector

https://github.com/ideditor/nsi-collector/blob/main/collect_osm.js#L33-L39

but that NOT a mandatory standard because some brands may have a lot in real world and truly notable for local mapper, just haven't been mapped. If local mapper think this is needed, they can add it to NSI directly. Sometimes even a chain brand only have 10 OSM elements is also acceptable for NSI.

Issue #6909, opened 2022-08-11
Malteser Hilfsdienst is brand or operator?


LaoshuBaby:
Now when run `npm run build` it will give a warning:

```
⚠️   Warning - matchIndex errors:
----------------------------------------------------------------------
  `key/value/name` occurs multiple times in the match index.
  To resolve these, make sure the key/value/name does not appear in multiple trees
    (e.g. `amenity/post_office/ups` should not be both a "brand" and an "operator"
----------------------------------------------------------------------
Duplicate cache key "amenity/recycling/q1318397" in trees "operators" and "brands", check items: malteserhilfsdienst-6737d9
total 1
```

This is because `malteserhilfsdienst-6737d9` in [operators/amenity/recycling.json](https://github.com/osmlab/name-suggestion-index/blob/main/data/operators/amenity/recycling.json) and `malteserhilfdienst-f6fab4` in [brands/amenity/recycling.json](https://github.com/osmlab/name-suggestion-index/blame/main/data/brands/amenity/recycling.json) are same entity, they have the same wikidata.

Details are shown below:

https://github.com/osmlab/name-suggestion-index/blob/ec766cfdec7d88dfdf2ea3455fb97eae15ff64c4/data/operators/amenity/recycling.json#L2190-L2199

https://github.com/osmlab/name-suggestion-index/blob/844909e140354d8e7857a71be99401115bd0eb29/data/brands/amenity/recycling.json#L186-L198

Maybe we need a German mapper review this and select which one should be use, and select proper matchNames and tags.

kmpoppe:
Currently discussing on https://t.me/OSM_de whether it's one or the other - or both.

Will come back to you :-)

AlertSubject:
https://github.com/osmlab/name-suggestion-index/issues/2588

Local experts could be pinged

AlertSubject:
Looks like the Telegram group has decided on "brand"
Operator can only be determined through survey

for later reference: https://t.me/OSM_de/81212

kjonosm:
I agree that `brand` should rather be used as the organisation as a whole is split into regional dioceses and further into smaller branches which usually operate services on the local level.

bhousel:
Just linking here to the comment I wrote last year that might help decide whether an item is more brand-like or operator-like:

https://github.com/osmlab/name-suggestion-index/issues/5625#issuecomment-958218478

LaoshuBaby:
[...]

PR #6962, opened 2022-08-23
Update OpenLoop Charging stations 

Samuel12321:
Moved Openloop from operators to brands, to reflect that OpenLoop doesn't actually own or physically operate any EV Chargers, they just provide the app and the software for the stations - so all the chargers appear in the openloop app with status, authentication and billing. 
The Physical stations themselves are owned by other companies, and usually painted with their respective branding.

Please check that this is the correct way to implement this.

### DIFF
--- a/data/brands/amenity/charging_station.json
+++ b/data/brands/amenity/charging_station.json
@@ -95,6 +95,15 @@
       }
     },
     {
+      "displayName": "OpenLoop",
+      "locationSet": { "include": [ "nz" ] },
+      "tags": {
+        "amenity": "charging_station",
+        "brand": "OpenLoop",
+        "brand:wikidata": "Q113289213"
+      }
+    },
+    {
       "displayName": "Recharge",
       "id": "recharge-508422",
       "locationSet": {
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -2652,17 +2652,6 @@
       }
     },
     {
-      "displayName": "OpenLoop",
-      "locationSet": {"include": ["nz"]},
-      "tags": {
-        "amenity": "charging_station",
-        "network": "OpenLoop",
-        "network:wikidata": "Q113289213",
-        "brand": "OpenLoop",
-        "brand:wikidata": "Q113289213"
-      }
-    },
-    {
       "displayName": "WEL Networks",
       "locationSet": {"include": ["nz"]},
       "tags": {
@@ -2671,8 +2660,6 @@
         "operator:wikidata": "Q7948949",
         "network": "OpenLoop",
         "network:wikidata": "Q113289213",
-        "brand": "OpenLoop",
-        "brand:wikidata": "Q113289213"
       }
     },
     {
@@ -2682,10 +2669,8 @@
         "amenity": "charging_station",
         "operator": "Z Energy",
         "operator:wikidata": "Q8063337",
-        "network": "OpenLoop",
-        "network:wikidata": "Q113289213",
-        "brand": "OpenLoop",
-        "brand:wikidata": "Q113289213"
+        "network": "Z Energy",
+        "network:wikidata": "Q8063337",
       }
     },
     {
@@ -2697,8 +2682,6 @@
         "operator:wikidata": "Q108450092",
         "network": "OpenLoop",
         "network:wikidata": "Q113289213",
-        "brand": "OpenLoop",
-        "brand:wikidata": "Q113289213"
       }
     },
     {
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -2659,7 +2659,7 @@
         "operator": "OpenLoop",
         "operator:wikidata": "Q113289213",
         "network": "OpenLoop",
-        "network:wikidata": "Q113289213"
+        "network:wikidata": "Q113289213",
         "brand": "OpenLoop",
         "brand:wikidata": "Q113289213"
       }
@@ -2673,7 +2673,7 @@
         "operator": "WEL Networks",
         "operator:wikidata": "Q7948949",
         "network": "OpenLoop",
-        "network:wikidata": "Q113289213"
+        "network:wikidata": "Q113289213",
         "brand": "OpenLoop",
         "brand:wikidata": "Q113289213"
       }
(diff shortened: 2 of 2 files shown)


Adamant36:
> The Physical stations themselves are owned by other companies, and usually painted with their respective branding

From what I've seen on Google Images there are plenty of charging stations that have "OpenLoop" painted on them and nothing else. So it seems to me that they are a brand themselves, even if they license out some of their charging stations to other companies.

Samuel12321:
They all have a letter sized 'openloop' plaque, some are branded / painted as openloop, but most are pained in their respective owners brands - well networks, Z energy ...
Whether this means they don't own any stations - i'm not sure. But at least the majority are owned by other companies. 
From their website:
> [Does OpenLoop own all EV Chargers on their app?](https://www.openloop.co.nz/faqs/#)
OpenLoop partners with enterprises such as Z Energy, WEL Networks, Counties Energy, and fleet partners such as Custom Fleet to install their chargers and to provide the software used for charging and billing. The chargers are owned by these enterprise partners, and we provide customer support through a call centre for charger users. Please call 0800 OPENLOOP if you need help. If we can’t resolve your issue, we will then contact the relevant enterprise partner before getting back to you.

From a local point of view the majority of the chargers in NZ are ChargeNet or Openloop

Brand seemed the most appropriate tag, so an example station would be labeled as:
brand=OpenLoop
operator= Z Energy
network=OpenLoop

@kjonosm does NSI support Network?

Samuel12321:
Example Z Energy Station: (only Openloop branding is small plaque by screen explaining how to start charge in Openloop app)
image

Example Counties Power Station: Painted in both Openloop and Counties power branding:
image

Adamant36:
In an example like this maybe it's technically the counties "power branding" but it's still an Openloop charging station. People are going to see the big blue "Openloop" when they approach it, not the dinky "Counties Energy" on the bottom and the point in this is to make it easier for people to find things on the map. We go that by tagging things the way they will see the object in person from at least some distance, which in this case clearly isn't "counties power." 

(Outside of that an argument could be that they are either both the brand, Openloop is, or there is no brand in the first place, but such discussions don't usually go anywhere because they are extremely circular. So it's better to just go with what people on the ground are more likely to see. At the end of the day it doesn't matter where the energy for the charging stations comes anyway).

Samuel12321:
another example

Samuel12321:
@Adamant36 @kjonosm keen to get this resolved, what is the best approach?

Adamant36:
I don't know. I'm leaning towards saying the display name should be OpenLoop in all or most cases since like I said it doesn't really matter where the charging station gets its energy from. Especially on the map consumers end. Otherwise, realistically what is the utility in displaying that information?

Say someone uses the OpenLoop app. You can't argue in that case that they won't then be looking for an OpenLoop charging station. Whereas, no one is going to not use an OpenLoop charging station just because they don't know where the power for it is coming from. If anything that information will confuse anyone except for hyper local map consumers, who aren't really NSIs target market anyway. 

That's just my opinion, but is someone from half way across the country really going to care that their car is being charged by the local municipalities energy? Probably not. Where I live we have a city run energy grid and most people think it's PG&E. So I just don't think people really care that much.

Samuel12321:
Totally agree, my concern was more with the Z energy petrol station chargers which are all branded as Z and wel network ones which are usually painted as such. No one cares where they get their power from, it is just a coincidence that a small handful of the chargers are owned by electricity infrastructure companies (counties power and wel networks)

Adamant36:
Cool. I have no problem with it being merged then.

LaoshuBaby (changes requested):
Wait a minute, look at #6909, if we keep both `brand:wikidata=Q113289213` and `operator:wikidata=Q113289213`, we will got the same warning as before.

---
But I don't know how to solve this PR, this situation confused me.

Samuel12321:
I guess the problem lies in the brand / network part. Traditionally a petrol station would be brand "X" operated by "Y" (Which can also be X), however with EV charging stations, the brand is kind of merged with network, and stations can occasionally be painted in the operators colors rather than exclusively painted in their network/brands colors.

I have updated it by removing the operator part from the Openloop entry:
Chargers would therefore be tagged as Openloop if unsure of operator or operated by Openloop. other would be tagged as their respective operators with the brand and network as Openloop. 

does this make some sort of sense?

Adamant36:
> Wait a minute, look at #6909, if we keep both `brand:wikidata=Q113289213` and `operator:wikidata=Q113289213`, we will got the same warning as before.
> 
> But I don't know how to solve this PR, this situation confused me.

@LaoshuBaby would you run into the same problem if the network:wikidata and brand:wikidata are the same or does that not  throw up a warning? (If not, then I say merge this and be done with it 👍 ).

LaoshuBaby:
> > Wait a minute, look at #6909, if we keep both `brand:wikidata=Q113289213` and `operator:wikidata=Q113289213`, we will got the same warning as before.
> > But I don't know how to solve this PR, this situation confused me.
> 
> @LaoshuBaby would you run into the same problem if the network:wikidata and brand:wikidata are the same or does that not throw up a warning? (If not, then I say merge this and be done with it 👍 ).

Sorry for my late reply, current data will raise many warning about both in brand tree and operator tree, this similar to #6909 and we need to choose which one more suitable for OpenLoop

图片

Samuel12321:
Have updated the PR:
- The Z energy discussion is now moot as they have moved to their own brand/ network and no longer use Openloop.
- To fix the warnings i have removed duplicate Openloop entries in operators.
- Openloop is now a 'Brand' only, but have left entries in operators, for counties energy and wel networks as their respective operators with network as openloop. - So stations can now be tagged as either openloop as brand, or counties energy/ wel networks as operator and openloop as network.

What's everyone's thoughts on this approach, does it make sense and resolve the problems?
@LaoshuBaby @Adamant36 @kjonosm

Adamant36:
> What's everyone's thoughts on this approach, does it make sense and resolve the problems?

I don't see why it wouldn't resolve the problem. It's not like we can't update it at some point if any issues with it come up. So I have zero problem with  it being merged. Although I'm not going to be able to merge it since I don't have the time to review the code right now, but if @LaoshuBaby or someone else wants to 👍

bhousel:
This has been going on for a while, so I'm probably just going to merge this and clean it up in a way to avoid the cache key warnings.

PR #6966, opened 2022-08-24
Add Bulgarian operators

Dimitar5555:
In this PR I've added:

- All 9 heating companies in Bulgaria
- All 3 energy distribution companies in Bulgaria
- All water utility companies
- 2 government agencies

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -2885,8 +2885,10 @@
     },
     {
       "displayName": "Маяк",
-      "id": "bc2a95-e1ef51",
-      "locationSet": {"include": ["001"]},
+      "id": "bc2a95-80fe8b",
+      "locationSet": {
+        "include": ["by", "ru", "ua"]
+      },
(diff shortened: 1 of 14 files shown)


Dimitar5555:
I don't see any duplicated entries in that file. What is the problem?
```
Error - Duplicate id '938062-a924a3' in:
  Софийска вода
  ./data/operators/man_made/water_works.json
```

arch0345:
> I don't see any duplicated entries in that file. What is the problem?
> 
> ```
> Error - Duplicate id '938062-a924a3' in:
>   Софийска вода
>   ./data/operators/man_made/water_works.json
> ```

Since you added 'Софийска вода' to `office/water_utility`, you don't need to add it to `man_made=water_works` since we have a template to copy all entries from that file:
https://github.com/osmlab/name-suggestion-index/blob/84ef5219ab21f496fe1df28c970a5cc633fbd644/data/operators/man_made/water_works.json#L8-L15

PR #6975, opened 2022-08-28
added 'Kurumaya Ramen' and 'Olive Hill'

haansn08:
* Kuruma Ramen is a chain restaurant with ~150 stores in Japan: https://www.kurumayaramen.co.jp/shoplist.htm
* Olive Hill is a Japanese chain restaurant similar to Saizeria: https://maps.olivenooka.jp/

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -3141,6 +3141,17 @@
       }
     },
     {
+      "displayName": "オリーブの丘",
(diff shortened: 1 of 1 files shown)


arch0345:
Thanks @haansn08! Note that entries without a `:wikidata` tag won't get generated as an iD preset. If this brand doesn't have a Wikidata item already, [feel free to create one yourself](https://github.com/osmlab/name-suggestion-index/wiki/Editing-Wikidata).

PR #6980, opened 2022-08-28
remove name= on amenity=parcel_locker

### LINKED ISSUE #6979: name= on amenity_parcel_locker
Hi,

It seems to me, some templates for parcel lockers include a `name=` tag, some don't. The tag is generally unneccessary as the name that's rendered on Carto, for instance is the `brand=` tag. How about removing all `name=` from `amenity=parcel_locker`s, then?

coolultra1:
fixes #6979


hanchao:
According to the discussion https://github.com/osmlab/name-suggestion-index/issues/6979#issuecomment-1251796527, 
 I add "preserveTags": ["^name"] to the general properties in the file.

PR #6983, opened 2022-08-30
Update Eric Kayser bakery

Djiril:
sorry i had to merge 2 wikidata set

### DIFF
--- a/data/brands/shop/bakery.json
+++ b/data/brands/shop/bakery.json
@@ -682,7 +682,7 @@
       "locationSet": {"include": ["fr"]},
       "tags": {
         "brand": "Eric Kayser",
-        "brand:wikidata": "Q112961485",
+        "brand:wikidata": "Q99231745",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Yes, Q112961485 was redirected to Q99231745
Thank you for telling us this. (BTW, before publish a new version, maintainer will run a wikidata checker and it will also find this redirect)

bhousel:
Just wanted to add too - for situations where the Wikidata QID is a replacement of another one, we can add it to this file as well:  https://github.com/osmlab/name-suggestion-index/blob/main/config/replacements.json

It can help because this will prevent people from accidentally re-adding the old QID back into the index.

LaoshuBaby:
> It can help because this will prevent people from accidentally re-adding the old QID back into the index.

Done in https://github.com/osmlab/name-suggestion-index/blob/main/config/replacements.json ! Thanks for reminding me! ♥

PR #6989, opened 2022-08-31
Adding Minor Hotels chains

arnowrld:
updating AVANI & adding NH Collection, nhow & Oaks Hotels


data/brands/tourism/hotel.json
```diff
@@ -217,6 +217,7 @@
       "locationSet": {"include": ["001"]},
```
tas50, line 217: There's few enough countries served that we should add them here

arnowrld: Makes sense to me, honestly.

Issue #5853, opened 2021-12-13
Add a UI to resolve warnings from `npm run wikidata`


arch0345:
Every time I run `npm run wikidata` I notice there's a lot of entries where Facebook/Twitter profiles linked on Wikidata are outdated, so unless they have a logo on Wikimedia, they're logo won't show up in iD. Currently these are just kept in [`dist/warnings.json`](https://github.com/osmlab/name-suggestion-index/blob/main/dist/warnings.json). Something similar to MapRoulette would be nice that show if warnings have been resolved by others or not since the last `npm run wikidata` commit - this could be a page on nsi.guide.

UKChris-osm:
I don't know if it's possible to show as being resolved without running `npm run wikidata` again, as our web site doesn't contain any server-side (back end) abilities, so there isn't a way for the `warnings.json` file to be edited when a Wikidata entry is amended.

Could `npm run wikidata` be run via a GitHub Action to check any edits?

bhousel:
> Could `npm run wikidata` be run via a GitHub Action to check any edits?

We probably could automate it, but I don't have a lot of free time to set this up right now.. 

> I hadn't realised we even had a `warnings.json` file, is this a recent addition?

Yes!  Just added this last September.. https://github.com/osmlab/name-suggestion-index/issues/4815#issuecomment-922095101

One idea we had there was to turn it into an RSS feed somehow, that might make it easier for people to follow or monitor.

peternewman:
I'm always a fan of GitHub annotations. Do you know the line number and file name of the source JSON file when the script is run @bhousel or just the warning? If so pulling that out to a PR annotation would be pretty easy.

I don't have a huge amount of time for a while (or more accurately have other bits to do first), but a GitHub Action should be pretty easy.

bhousel:
> I'm always a fan of GitHub annotations. Do you know the line number and file name of the source JSON file when the script is run @bhousel or just the warning? If so pulling that out to a PR annotation would be pretty easy.

Certain conditions we detect in [`build_wikidata.js`](https://github.com/osmlab/name-suggestion-index/blob/main/scripts/build_wikidata.js) will be saved into a `_warnings` array, then we print those to the console and also dump them into the `warnings.json` file.

PR #7002, opened 2022-09-03
Add Wikidata Warnings list to nsi.guide web site

UKChris-osm:
Grabs the warnings.json file and makes it viewable via the nsi.guide web site!

https://nsi.haignet.co.uk/


UKChris-osm:
Ref: #5853

bhousel:
It's a good idea!  I don't really think it belongs in the list of trees though.
Adding it as a special tree kind of increases the complexity of the existing code a bunch.
Can you make it just its own page separate from everything else?

Screen Shot 2022-09-06 at 10 11 36 AM

UKChris-osm:
I don't know React well, but the impression I got was that it had to part of the tree? Is it possible to create an out of tree scoped page?

I thought it fit well with the top tree navigation as these warnings are quite substantial in quantity, and it gave an easy way for people to contribute to some of the problems the NSI finds when trying to connect to Wikidata, especially with Hacktoberfest around the corner.

I could give it a go though, and see if I can make it its own page.

PR #6993, opened 2022-09-02
Brand Name Updated

MG-LSJ:
Brand ESSAR name updated to Nayara.

(Edit: Made a typo in original commit)

PR #7003, opened 2022-09-03
Brand name Updated (2nd pull request)

MG-LSJ:
Updated the Brand Name and fixed an issue with previous PR.
-----
Related： #6993


UKChris-osm:
Thanks for resubmitting the PR @MG-LSJ 

However, looking into it, Essar (https://www.essaroil.co.uk/our-forecourts/) still seems to be trading in the UK, so it might be better to add Nayar as its own entry for now.

bhousel:
Thanks @MG-LSJ , I agree with @UKChris-osm and did a little research, and it looks like the Essar brand is still used in both GB and IN, and the Nayara brand is being phased in in IN, so we can keep both names for now.

I added the wikidata tags and some matchnames in 3ba4180a9

Issue #7013, opened 2022-09-07
Dufry


1ec5:
Dufry, the parent company of Hudson, owns [several chains](https://www.dufry.com/en/company/our-retail-brands) of duty-free stores at airports, including these chains that have Wikidata items:

* [Dufry](https://www.wikidata.org/wiki/Q1264020)
* [World Duty Free](https://www.wikidata.org/wiki/Q15911908)
* [Nuance](https://www.wikidata.org/wiki/Q650593)
* [Hellenic Duty Free Shops](https://www.wikidata.org/wiki/Q5707214)

PR #7018, opened 2022-09-07
Updated for #7013

ANaphade:
Updated for #7013

### DIFF
--- a/data/brands/shop/department_store.json
+++ b/data/brands/shop/department_store.json
@@ -1439,7 +1439,7 @@
     {
       "displayName": "World Duty Free",
       "id": "worlddutyfree-b0f55b",
-      "locationSet": {"include": ["it"]},
+      "locationSet": {"include": ["001"]},
       "tags": {
         "brand": "World Duty Free",
         "brand:wikidata": "Q15911908",
--- a/data/brands/shop/department_store.json
+++ b/data/brands/shop/department_store.json
@@ -1033,7 +1033,7 @@
      {
       "displayName": "Nuance",
       "id": "nuance-b0f55b",
-      "locationSet": {"include": ["eu","uk"]},
+      "locationSet": {"include": ["001"]},
       "tags": {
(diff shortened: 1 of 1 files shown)


data/brands/shop/department_store.json
```diff
@@ -1386,6 +1436,17 @@
         "shop": "department_store"
       }
     },
+    {
+      "displayName": "World Duty Free",
+      "id": "worlddutyfree-b0f55b",
+      "locationSet": {"include": ["it"]},
```
arch0345, line 1442:
```suggestion
      "locationSet": {"include": ["001"]},
```
According to their [Wikipedia article](https://en.wikipedia.org/wiki/World_Duty_Free#World_Duty_Free_Group), they have locations in 20 countries

data/brands/shop/department_store.json
```diff
@@ -364,6 +364,34 @@
...
+        "hk",
+        "co",
+        "ru",
+        "am",
+        "tr",
+        "ch",
+        "ma",
+        "kh",
+        "ae"
+      ]},
```
arch0345, line 388:
```suggestion
      "locationSet": {"include": ["001"]},
```
They have some locations in the US as well, which aren't listed on their Wikipedia article. I think it's best to include the whole world as they could have locations in other countries as well.

data/brands/shop/department_store.json
```diff
@@ -1007,6 +1046,17 @@
         "name": "Nordstrom",
         "shop": "department_store"
       }
+    },
+     {
+      "displayName": "Nuance",
+      "id": "nuance-b0f55b",
+      "locationSet": {"include": ["eu","uk"]},
```
arch0345, line 1053:
```suggestion
      "locationSet": {"include": ["001"]},
```
It appears they also have a location in Canada. (https://toronto.shopdutyfree.com/en/). I think it's best to include the whole world as they could have locations in other countries as well.

PR #7020, opened 2022-09-08
Update residential.json with Greystar info

preeformance:
Added details for Greystar per issue #7010  (did not do npm run build)

### DIFF
--- a/data/brands/landuse/residential.json
+++ b/data/brands/landuse/residential.json
@@ -18,19 +18,6 @@
       }
     },
     {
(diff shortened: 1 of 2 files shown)


arch0345 (commented):
I think Greystar should go in the operators tree rather than the brands tree because unlike Avalon and Moderna, Greystar properties aren't all named 'Greystar' (ie [Oceanaire](https://www.openstreetmap.org/way/1082581428))

PR #7025, opened 2022-09-10
updates related to New Zealand

k-yle:


UKChris-osm:
Thank you for these additions @k-yle 👍

PR #7028, opened 2022-09-11
Change "Deutsche Telekom" charging station operator to the consistent…

Lukas458:
… value "Deutsche Telekom AG"

For the operator "Deutsche Telekom [AG]" in Germany, we should use a consistent value. We do have "operator=Deutsche Telekom AG" (+ operator:wikidata=Q9396) already in the NSI for amenity=telephone and telecom=exchange. So we should use the value "Deutsche Telekom AG" for charging stations, too, to make it consistent. Especially because the operator for the telephones, the telephone exchanges and the charging stations is one adn the same, here.

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -789,12 +789,12 @@
       }
     },
     {
-      "displayName": "Deutsche Telekom",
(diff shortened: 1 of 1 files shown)


Lukas458:
Thank you very much.

PR #7035, opened 2022-09-15
Removed names from wasterwater plants

Dimitar5555:


bhousel:
Thanks! Looks like they were added like this in #6966 - I agree that normally our code doesn't add a `name` tag for these..

PR #7036, opened 2022-09-15
Update clothes.json

xmready:
Cotton On Kids has [it's own wikidata entry](https://www.wikidata.org/wiki/Q113961498) and is a separate brand from Cotton On.


arch0345:
Thanks @xmready and welcome!

PR #7041, opened 2022-09-17
Update cannabis.json

ShinyDog2636:
Added BC Cannabis Stores, a crown corporation selling cannabis products in BC, Canada. This is my first addition to NSI so please let me know if I missed anything.


UKChris-osm:
Thanks @ShinyDog2636 - you can leave out the id tag (rather than set it as unset) as this is something that is automatically generated by the NSI build script.

PR #7042, opened 2022-09-18
Update restaurant.json

acopperlily:
Added Wikidata info and cuisine type to Mission BBQ.


UKChris-osm:
Thank you @acopperlily 👍

PR #7043, opened 2022-09-18
Add MrBeast Burger

Robot8A:


Cj-Malone:
Are these real? There is only 1 in OSM, and 1 dark kitchen.

PR #7053, opened 2022-09-19
Add various Bulgarian brands

Dimitar5555:


data/operators/man_made/pipeline.json
```diff
@@ -5474,6 +5474,86 @@
...
+      "displayName": "Топлофикация Бургас",
+      "locationSet": {"include": ["bg"]},
+      "tags": {
+        "operator": "Топлофикация Бургас",
+        "operator:wikidata": "Q113611230",
+        "man_made": "pipeline"
+      }
+    },
+    {
+      "displayName": "Топлофикация Враца",
```
VladimirAlexiev, line 5488: @Dimitar5555 for all Топлофикации:
- For more precise location, use [Q189502](https://www.wikidata.org/wiki/Q189502) (Vratsa) as per https://github.com/osmlab/name-suggestion-index/wiki/Contributing#locationset-required
- add EN name:
```
       "operator": "Топлофикация Враца",
       "operator:bg": "Топлофикация Враца",
       "operator:en": "Toplofikacia Vratsa",
```

Dimitar5555: Thanks for the idea! I didn't know that it was possible to use wikidata for it. I will create a new PR and fix up the all of the rest such items that I've added in previous PRs.
> add EN name:

I think that the English name is not required since it can be transliterated on the go with a small script (there are plenty of them, for a quick list of what is replaced by what see https://www.lex.bg/laws/ldoc/2135623667).

Dimitar5555: @VladimirAlexiev unfortunately it doesn't seem to work since there isn't a geoshape property for the municipalities nor for the cities in Wikidata. Plan B is to export that data from OpenStreetMap and include it here but I'm not sure if that is possible and if the licenses are compatible.

arch0345: Note that using Wikidata QIDs in locationSets only works for locations in the [ideditor/country-coder](https://github.com/ideditor/country-coder) project (only continents/countries), see https://github.com/osmlab/name-suggestion-index/wiki/Contributing#locationset-required

arch0345 (changes requested):
Some of the locationSets are invalid

data/operators/man_made/pipeline.json
Dimitar5555: @arch0345 I've rolled back these changes.

arch0345 (approved):

data/operators/man_made/pipeline.json
VladimirAlexiev: We could add BG region shapes but if NSI doesn't support them...
Sorry I misled you, @Dimitar5555

arch0345: Yes, you can [add geojsons](https://github.com/osmlab/name-suggestion-index/wiki/Feature-Files) directly to NSI. For example, [we have geojsons for US states](https://github.com/osmlab/name-suggestion-index/tree/main/features/us). Just make sure that you use as few points as possible as larger geojsons take longer to load into iD

PR #7060, opened 2022-09-22
Update restaurant.json

ShinyDog2636:
Add Canadian restaurant chain

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -6956,12 +6956,11 @@
     },
     {
       "displayName": "Cactus Club Cafe",
-      "locationSet": {"include": ["ca-bc.geojson", "ca-ab.geojson"]},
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Thanks @ShinyDog2636 

Just a couple of queries - I think the location would be better set to Canada as a whole, and "_premium_casual_" as a **cuisine** value is one I've not come across before, and doesn't seem to be mentioned on the OSM Wiki.

ShinyDog2636:
You are right, it should be Canada as a whole. My mistake.

Regarding the cuisine, I struggled with this one. Canada has several very similar restaurant chains that all describe themselves as ["premium casual" restaurants](https://en.wikipedia.org/wiki/Types_of_restaurants#Premium_casual). They are semi pub, semi premium restaurant with a globally inspired menu. I have removed the cuisine type for now and I will consider adding it to the OSM Wiki.

Thanks,
Joel

PR #7063, opened 2022-09-22
Update fast_food.json

ShinyDog2636:
Fatburger also operates in Canada


UKChris-osm:
Thank you @ShinyDog2636 👍

Issue #7056, opened 2022-09-20
Issue with the way the Japanese DIY brand "Viva Home" name tag is written


Revtro:
Hello.
I re-created another issue as I wrote in #7055.

```
    {
      "displayName": "VIVA Home",
      "id": "vivahome-dd401d",
      "locationSet": {"include": ["jp"]},
      "tags": {
        "brand": "VIVA Home",
        "brand:en": "VIVA Home",
        "brand:ja": "ビバホーム",
        "brand:wikidata": "Q11229278",
        "name": "VIVA Home",
        "name:en": "VIVA Home",
        "name:ja": "ビバホーム",
        "shop": "doityourself",
        "short_name:en": "VH"
      }
    }
```

```
   {
      "displayName": "SUPER VIVA Home",
      "id": "supervivahome-dd401d",
      "locationSet": {"include": ["jp"]},
      "tags": {
        "brand": "VIVA Home",
        "brand:en": "VIVA Home",
        "brand:ja": "ビバホーム",
        "brand:wikidata": "Q11229278",
        "name": "SUPER VIVA Home",
        "name:en": "SUPER VIVA Home",
        "name:ja": "スーパービバホーム",
        "shop": "doityourself",
        "short_name:en": "SVH"
      }
    }
```

That seems to have already been discussed in #3624, the name tag is written in English.
However, in the [Japanese Naming sample](https://wiki.openstreetmap.org/wiki/JA:Naming_sample) on the OSM wiki, it is written in Japanese.

According to #3624, the name tag seems to be based on "signage". But in Japan, signs are often written in English, even if they are usually used in Japanese.
For example, Aeon which is a very famous supermarket in Japan, Lawson, Family Mart, and 7-Eleven which are also famous convenience stores, and so on.

By the way, in some stores, Viva Home signs may also be written in Japanese.

Furthermore, the Japanese Naming sample has a standard: "**In the name tag, if the store's brand name (as written on the sign) and the name in Japanese are different, the name in Japanese is given priority**".

Therefore, I think that name tags in Japan should be written in Japanese as well as other brands in Japan if it is common to use Japanese for brand names, even if the signage is in English, as is the standard of the OSM Wiki in Japan.

As a supplement, if the official name of the brand is also in English (alphabet), such as "apollostation" in #7055, I think this would not be the case.

I apologize for the length of the post and for the trouble I have caused.

PR #7064, opened 2022-09-22
Japanese second_hand, supermarket, and etc. additions

Revtro:
For Aeon Supercenter, alphabetical notation should be used following the [JA:Naming sample](https://wiki.openstreetmap.org/wiki/JA:Naming_sample#shop=supermarket).
|name|name:en|name:ja|...|brand|...|
|---|---|---|---|---|---|
|イオン|AEON(I think this could also be "Aeon" to match the others)|イオン|...|AEON|...|

As for the modification of the brand:en tag, etc., I try to match the writing style of the JA:Naming sample as much as possible.

Regarding Viva Home, for more information, please refer to #7056.

---

Specifically, if the official name is in Japanese, the name tag is assumed to be in Japanese, and if in English, it is assumed to be in English.
For brand tags, preference is basically given to those commonly used as logos or trademarks, and English translations of name tags are basically lower case except for the initial letter, in line with other examples.

In Japan, brand tags do not necessarily use the local language, but basically use the notation of the logo. There was already a discussion of this on the mailing list.

---

Related: #7055 
Close: #7056


data/brands/shop/supermarket.json
```diff
@@ -9705,63 +9797,6 @@
...
-      "tags": {
-        "brand": "Рода",
-        "brand:en": "Roda",
-        "brand:sr": "Рода",
-        "brand:wikidata": "Q21446192",
-        "name": "Рода",
-        "name:sr": "Рода",
-        "name:sr-Latn": "Roda",
-        "name:en": "Roda",
-        "shop": "supermarket"
```
Revtro, line 9763: Just to be sure: this should be around line 7991.

PR #7071, opened 2022-09-24
Adding Moj Kiosk Serbia

u-spec-png:


data/brands/shop/kiosk.json
```diff
@@ -218,6 +218,19 @@
         "name:ja": "キヨスク",
         "shop": "kiosk"
       }
+    },
+    {
+      "displayName": "Moj Kiosk",
+      "id": "kiosk-15613f",
+      "locationSet": {"include": ["rs"]},
+      "tags": {
+        "brand": "Moj Kiosk",
```
LaoshuBaby, line 227: I'm curious why you don't keep the brand consistent with the name and use `Мој киоск` here

LaoshuBaby:
@u-spec-png 

Thank you for contribute this! I don't means you must change this, I'm just curious why you chose to use English instead of Serbian as brand‘s value. As long as you can provide a explanation for what spelling you choose to use, they are all reasonable.

u-spec-png:
@LaoshuBaby 

Brand name wasn't in English it was in Serbian I just forgot to change it form latin to cyrillic script

PR #7073, opened 2022-09-24
Add AIK Bank

u-spec-png:


data/brands/amenity/atm.json
```diff
@@ -240,6 +240,16 @@
         "name:he": "בנק טינקוף",
         "name:ru": "Тинькофф Банк"
       }
-    }
+    },
+    {
```
LaoshuBaby, line 244: https://github.com/osmlab/name-suggestion-index/blob/0f5370d5d28a099e595a2cb66d67227b567ab308/data/brands/amenity/atm.json#L2-L15

ATM will inherit from bank.json, you don't need to add twice, just add in bank.json will affect both amenity=bank and amenity=atm.

LaoshuBaby (changes requested):

data/brands/amenity/atm.json
u-spec-png: Oh, ok

data/brands/amenity/bank.json
```diff
@@ -14930,6 +14930,28 @@
         "name:en": "Kagoshima Bank",
         "name:ja": "鹿児島銀行"
       }
+    },
+    {
+      "displayName": "АИК банка",
+      "locationSet": {"include": ["rs"]},
+      "matchNames": [
+        "AIK bank",
+        "AIK banka",
```
LaoshuBaby, line 14939: Thank you, but for pure English character, you don't need to add uppercase and lowercase twice in matchname, because when run auto-formatter program they will be transformed to lowercase all.

So I think "AIK BANK" and "AIK BANKA" is not needed. but don't worry and don't need your action, this will be fix automatically.

See https://github.com/osmlab/name-suggestion-index/wiki/Contributing#matchnamesmatchtags-optional

data/brands/amenity/bank.json
```diff
@@ -14930,6 +14930,28 @@
...
+        "AIK BANK",
+        "AIK BANKA",
+        "АИК банк",
+        "АИК БАНК"
+      ],
+      "tags": {
+        "amenity": "bank",
+        "brand": "АИК банка",
+        "brand:wikidata": "Q293650",
+        "website": "https://www.aikbanka.rs/",
```
LaoshuBaby, line 14949:
```suggestion
```

Sorry, NSI don't collect website data. This had documented in https://github.com/osmlab/name-suggestion-index/wiki/Contributing#tags-required
Such detailed data can be sent to Wikidata. So I'll remove this line.

LaoshuBaby (changes requested):

PR #7074, opened 2022-09-24
Fix NIS Pterol

u-spec-png:


data/brands/amenity/fuel.json
```diff
@@ -3050,14 +3050,22 @@
       }
     },
     {
-      "displayName": "NIS",
+      "displayName": "НИС Петрол",
       "id": "nis-a971aa",
       "locationSet": {"include": ["rs"]},
       "tags": {
```
LaoshuBaby, line 3056: I found 300 element using this wikidata while some element have simple `brand=NIS` in taginfo:
simpleNIS
300elements

I suggest that add `"nis"` as a matchNames.

data/brands/amenity/fuel.json
```diff
@@ -3050,14 +3050,22 @@
...
         "amenity": "fuel",
-        "brand": "NIS",
+        "brand": "НИС Петрол",
         "brand:wikidata": "Q1279721",
-        "name": "NIS"
+        "name": "НИС Петрол",
+        "name:sr": "НИС Петрол",
+        "name:sr-Latn": "NIS Petrol",
+        "contact:email": "office@nis.rs",
```
LaoshuBaby, line 3063: Besides, the same reason as https://github.com/osmlab/name-suggestion-index/pull/7073#discussion_r979248251, we don't collect such detailed contact data, and we can't assume that every branch amenity share this. The only thing they should share is the canonical brand features itself.

LaoshuBaby (changes requested):

PR #7075, opened 2022-09-24
National Car Parks

codeinabox:


codeinabox:
Not sure whether I should use NCP as the operator as that is more commonly used though that's an acronym. @UKChris-osm what do you suggest?

UKChris-osm:
I would say this is fine as it is, as you've added a matchName for the acronym 👍

PR #7090, opened 2022-09-27
Add more parcel lockers brands

kubahaha:
Updated templates with 2 more brands from Poland - based on actual data in OSM database (and my own survey).

### DIFF
--- a/data/brands/amenity/parcel_locker.json
+++ b/data/brands/amenity/parcel_locker.json
@@ -162,7 +162,8 @@
         "amenity":"parcel_locker",
         "brand":"DPD Pickup Station",
         "operator:wikidata":"Q541030",
-        "operator":"DPD"
+        "operator":"DPD",
+        "parcel_pickup": "yes"
       }
     },
     {
--- a/data/brands/amenity/parcel_locker.json
+++ b/data/brands/amenity/parcel_locker.json
@@ -161,8 +161,9 @@
       "tags": {
         "amenity":"parcel_locker",
         "brand":"DPD Pickup Station",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/parcel_locker.json
```diff
@@ -143,6 +154,17 @@
...
     },
+    {
+      "displayName": "DPD Pickup Station",
+      "id": "dpdpickupstation-0154ac",
+      "locationSet": {"include": ["pl"]},
+      "tags": {
+        "amenity":"parcel_locker",
+        "brand":"DPD Pickup Station",
+        "operator:wikidata":"Q541030",
+        "operator":"DPD"
```
LaoshuBaby, line 165: looks like this item don't have `brand:wikidata` but a `operator:wikidata`, and is this really don't need `parcel_pickup` key?

kubahaha: As not all of them currently mapped in Poland had = I didn't want to add it, however [official website](https://www.dpd.com/pl/pl/automaty-paczkowe-w-dpd-polska/) says clearly they should be able to meet criteria. I will update my PR with that change.

There is no wikidata page about that brand, however brand name is in use, so I believe proposed tagging is ok. `operator:wikidata` refers to DPDgroup, as it's operating service. What is wrong here?

Here is example photo of such locker:

photo_2022-09-27_22-56-50

kubahaha: Ok, I decided it's best to add Wikidata item.
I updated template with new brand:wikidata reference.

PR #7084, opened 2022-09-26
UA updates

serhii-muchychka:
- Moved Domino's to a common scheme. In Ukraine, the Dominos chain uses only the original name without transliteration
 - Sorted cuisine according to relevance (matches the lexicographic order) for Eurasia
 - Beeline was bought by Kyivstar in 2010. The brand has not survived. There are no entries in the OSM DB
 - Added Eldorado chain of electronics shops

PR #7092, opened 2022-09-27
Removed forgotten exclude statement

serhii-muchychka:
See https://github.com/osmlab/name-suggestion-index/pull/7084

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -1709,8 +1709,7 @@
           "cn",
           "jp",
           "kz",
-          "ru",
-          "ua"
+          "ru"
         ]
       },
       "matchNames": ["dominos pizza"],


LaoshuBaby:
图片

Hey @serhii-muchychka , this may not be a "forgotten" match. You can see this picture.

## In current data

if a user edit a element which named with `Domino's` or `dominos pizza` or `доміно` in **UA**, it will match this:

https://github.com/osmlab/name-suggestion-index/blob/a32f2de0bbb8cba11f2b4be8e7b1f5c27e9168d9/data/brands/amenity/fast_food.json#L1726-L1741

And mapper edit a element in everywhere except **UA, RU, KZ, BY, JP, CN**, if name is `Domino's` or `domino pizza`, it will match this: (But `доміно` won't match any entry)
https://github.com/osmlab/name-suggestion-index/blob/a32f2de0bbb8cba11f2b4be8e7b1f5c27e9168d9/data/brands/amenity/fast_food.json#L1707-L1725

## after delete UA in 001

(then "dominos-50e0ae" should be deleted)

No matter user in UA or other places, only `Domino's` or `domino pizza` can match "dominos-b946c5", that is the blue array in sketch. and only in ru/kz/by when mapper type `Домино` will lead to this brand.

---
I don't know which one is your excepted behavier, those localized entry is intent to allow users to find locally appropriate presets in their local language, I hope this will allow the mapper of the Ukrainian community to better standardize the exist data ♥

serhii-muchychka:
Really. Missed. I just started to understand this project

PR #7097, opened 2022-09-29
MaxValu (supermarket) correction

Revtro:
English names and Wikidata were limited to Tokai region, so they were corrected to common ones.

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -9077,15 +9077,15 @@
     },
     {
       "displayName": "マックスバリュ",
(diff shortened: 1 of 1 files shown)


LaoshuBaby (approved):
Approve. BTW, if you want to keep the old maxvalutokai-fe0970 version, you can follow https://github.com/osmlab/name-suggestion-index/wiki/Contributing#locationset-required and create a new entry using circular locationset.

Revtro:
I see, thank you very much.
But I don't think it is necessary because MaxValu Tokai is only an operating company and the brand itself is MaxValu everywhere.

PR #7110, opened 2022-09-30
Update supermarket.json

ShinyDog2636:
Adding Canadian supermarket chain: PriceSmart Foods

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -9816,6 +9816,18 @@
         "name:ja": "関西スーパー",
         "shop": "supermarket"
       }
+    },
+    {
+      "displayName": "PriceSmart Foods",
+      "locationSet": {"include": ["ca"]},
+      "tags": {
+        "brand": "PriceSmart Foods",
+        "brand:wikidata": "Q17023287",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Hi @ShinyDog2636 - does PriceSmart only have two stores?

ShinyDog2636:
Yes, only two stores but it does have a Wikipedia/Wikidata page so I think it is notable enough? It is a subsidiary of a much larger supermarket chain in Canada.

Ultimately I wanted to capture the Chinese name.

LaoshuBaby:
> Yes, only two stores but it does have a Wikipedia/Wikidata page so I think it is notable enough?

https://github.com/ideditor/nsi-collector/blob/0594a6049c45e3bb91a96d6d29b443b164aff360/collect_osm.js#L33-L39

Sorry, if only have small amount of brands, we can't see it as a canonical common brand. Current a brand occur more than 50 times in OSM data will be collected. In some cases the thershold can be small, but 2 single shop can't make a chain.

Besides, looks like the Chinese name, at least in meaning, fit English name exactly.

ShinyDog2636:
Okay I understand. Thanks for explaining.

PR #7111, opened 2022-09-30
Update cafe.json

ShinyDog2636:
Adding Canadian cafe's: JJ Bean, Bubble Waffle Cafe, and 49th Parallel

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -4744,6 +4744,20 @@
       }
     },
     {
+      "displayName": "Bubble Waffle Cafe",
+      "locationSet": {"include": ["ca"]},
+      "tags": {
+        "amenity": "cafe",
+        "brand": "Bubble Waffle Cafe",
+        "brand:wikidata":"Q114914529",
+        "cuisine": "waffle",
+        "name": "Bubble Waffle Cafe",
+        "name:en": "Bubble Waffle Cafe",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/cafe.json
```diff
@@ -4730,6 +4730,43 @@
...
+        "name": "JJ Bean",
+        "takeaway": "yes"
+      }
+    },
+    {
+      "displayName": "Bubble Waffle Cafe",
+      "locationSet": {"include": ["ca"]},
+      "tags": {
+        "amenity": "cafe",
+        "brand": "Bubble Waffle Cafe",
```
LaoshuBaby, line 4751: Note: this entry haven't a wikidata, it won't be suggested in editor.

https://github.com/osmlab/name-suggestion-index/wiki/Contributing

ShinyDog2636: Okay I removed this entry. Thanks

LaoshuBaby: > Okay I removed this entry. Thanks

Aha, What I means isn't ask you to delete it. I just see its website, looks like it already 24 branch shops. 

Although don't meet current 50 thershold (infact at first it only need 10 branch shops), but I think it still meaningful.

(虽然数量少，也不是不可以添加进来）

If you want it add, you need to offer a wikidata QID, you can search on wikidata or create a new one, then paste it here.

PR #7114, opened 2022-10-01
Add postanska stedionica

u-spec-png:


data/brands/amenity/bank.json
```diff
@@ -15113,6 +15113,20 @@
         "brand:wikidata": "Q869855",
         "name": "Crnogorska Komercijalna Banka"
       }
+    },
+    {
+      "displayName": "Поштанска штедионица",
+      "locationSet": {"include": ["rs"]},
+      "matchNames": ["Poštanska štedionica"],
```
LaoshuBaby, line 15120: Looks like it is the same as `name:sr-Latn` 's value, so you don't need to add this as matchName.

for example, a brand that have `name:zh=邮政储蓄银行`+`name:en=Youzheng Chuxu Yinhang` don't need to add those 2 as matchName, but `Youchu Yinhang` can be.

---
Don't worry, duplicate matchNames will be removed when auto-fomatter run, you don't need to do anything.

PR #7123, opened 2022-10-02
Add Oliver Bonas

peternewman:


data/brands/shop/clothes.json
```diff
@@ -5541,6 +5541,17 @@
         "shop": "clothes"
       }
     },
+    {
+      "displayName": "Oliver Bonas",
+      "locationSet": {"include": ["gb"]},
+      "matchTags": ["shop/boutique", "shop/department_store" "shop/fashion_accessories", "shop/gift", "shop/interior_decoration"],
```
peternewman, line 5547: I think this means it should just be:
https://github.com/osmlab/name-suggestion-index/blob/a976e0577e86f2d17635c532bc693a987968923a/config/matchGroups.json#L77-L87
```suggestion
      "matchTags": ["shop/gift", "shop/interior_decoration"],
```

peternewman: The vast majority are currently tagged as clothes shops, but perhaps they shouldn't be:
https://www.oliverbonas.com/homeware/category/mugs-cups

LaoshuBaby: > The vast majority are currently tagged as clothes shops, but perhaps they shouldn't be: https://www.oliverbonas.com/homeware/category/mugs-cups

Looks like another multiple business brand, like #6850 (a newsagents sells tobacco)

If this brand don't only sell clothes, but mainly sell clothes, I think we should still recognize it as clothes shop, but according to its website:

图片

LaoshuBaby: I run a overpass query: https://overpass-turbo.eu/s/1mn9 , here are result:

+ shop=boutique 1
+ **shop=clothes 28**
+ shop=department_store 2
+ shop=fashion_accessories 1
+ **shop=gift 20**
+ shop=houseware 1
+ shop=interior_decoration 2
+ shop=shop 1
+ shop=yes 3

Looks like UK community prefer think it as clothes shop or gift shop.

peternewman: > Looks like another multiple business brand, like #6850 (a newsagents sells tobacco)

If the shops I've seen aren't so small, I'd have tagged it as a department store, as I see a few people have done.

> If this brand don't only sell clothes, but mainly sell clothes, I think we should still recognize it as clothes shop, but according to its website:
> 
> 图片
> 
> (If in my eye, I'll think this is a gift shop that mainly sell cute small things)

I think it was mostly known for clothes, although Wikipedia mentions homeware too. For me, and the wiki sort of seems to agree, a gift shop would be more trinkets almost or dedicated to a particular theme, this sort of feels a bit too useful...

Thanks for the stats, that's a bit more scientific than my quick search for them and scroll down the list. I'd say with the boutique and fashion accessory ones, clothes definitely seems to win out for now.

I might have to brave going into one to see what they're like inside! :laughing:

peternewman: > I might have to brave going into one to see what they're like inside! laughing

So I looked properly through the window of a nearby one and my earlier assessment seems to fit:

> If the shops I've seen aren't so small, I'd have tagged it as a department store

It's predominantly clothes, but with a chunk of homeware, some trinkets/decorations etc.

So `shop=clothes`? :shrug:

peternewman:
Thanks @LaoshuBaby

PR #7124, opened 2022-10-02
add storage rental firms

k-yle:
3 entries for australia/new zealand


UKChris-osm:
Would [Q21178908](https://www.wikidata.org/wiki/Q21178908) be better for National Storage as it has its own Wikipedia article attached to it?

k-yle:
sorry, not sure how I missed that one. Fixed and merged the wikidata items.

PR #7127, opened 2022-10-03
Update with "Le Chocolat Alain Ducasse"

Djiril:


data/brands/shop/chocolate.json
```diff
@@ -127,6 +127,21 @@
...
+      "id": "",
+      "locationSet": {
+        "include": ["fr","gb"]
+      },
+      "tags": {
+        "brand": "Le Chocolat Alain Ducasse",
+        "brand:wikidata": "Q114374063",
+        "name": "Le Chocolat Alain Ducasse",
+        "shop": "chocolate"
+        "website":"https://www.lechocolat-alainducasse.com"
```
LaoshuBaby, line 142:
```suggestion
```
Hello, @Djiril , thank you for providing us those local information, but NSI don't collect website data. See https://github.com/osmlab/name-suggestion-index/wiki/Contributing#tags-required 

Looks like it already in wikidata, so I suggest remove this line.

LaoshuBaby (changes requested):

PR #7129, opened 2022-10-03
Add Le Chocolat Alain Ducasse

Djiril:


data/brands/shop/chocolate.json
```diff
@@ -127,7 +127,24 @@
...
+        "include": [
+          "fr","gb"
+        ]
+      },
+      "tags": {
+        "brand": "Le Chocolat Alain Ducasse",
+        "brand:wikidata": "Q114374063",
+        "name": "Le Chocolat Alain Ducasse",
+        "shop": "chocolate",
+        "website":"https://www.lechocolat-alainducasse.com"
```
LaoshuBaby, line 144:
```suggestion
```
Hello! I notice you open 2 similar PR in short time and you closed the old one (#7127 ). However the problem still exist, I don't think we need collect website data in NSI.

LaoshuBaby (changes requested):

data/brands/shop/chocolate.json
Djiril: ok i thought the problem was a missing "," a the end of line 143.
i didn't know that NIS does not accept "website", i know now ;)
thanks

PR #7137, opened 2022-10-04
Add reading cinemas

orangegrove1955:
Added Australian and New Zealand cinema chain, Reading Cinemas

https://en.wikipedia.org/wiki/Reading_Cinemas
https://www.wikidata.org/wiki/Q43898773


UKChris-osm:
Thank you @orangegrove1955 👍

Issue #5338, opened 2021-09-14
ATM name field tagging


peternewman:
I don't know if this is tagging for the renderer, but I see iD now offers to upgrade ATMs using the templated tagging.

If you do this with say the bank Foo, then the ATM also gets the tag `name=Foo` which seems a bit confusing, and doesn't match the behaviour of other systems as @bhousel screenshotted before:
https://github.com/osmlab/name-suggestion-index/issues/4975#issuecomment-803634583

It also means that in things like the commit changes bit of iD it's just listed as Foo, rather than Foo ATM which it feels to me like it should be?

Although I guess a different word may want to be used in other countries (perhaps even Cash Machine in the UK).

I see there was further discussion in:
https://github.com/osmlab/name-suggestion-index/issues/2883#issuecomment-726160004
https://github.com/osmlab/name-suggestion-index/commit/46ab8eba06f31571b02039883f59ec5a6204ec00#commitcomment-44125210

So maybe I should be raising this elsewhere, or it's a non issue. It sort of makes more sense on the rendered map then when displayed in ID.

bhousel:
I don't have really strong opinions about what the ATMs are named.. 

**What we do currently** is just use a template item to copy over all the tags from the corresponding banks, and add `operator:*` tags too.

https://github.com/osmlab/name-suggestion-index/blob/6e4902ceb8b46d4588448a1df4c335a7e3c3b2f8/data/brands/amenity/atm.json#L8-L16


For example: https://nsi.guide/index.html?t=brands&k=amenity&v=atm#barclays-2f6feb
```js
amenity=atm
brand=Barclays
brand:wikidata=Q245343
brand:wikipedia=en:Barclays
name=Barclays
operator=Barclays
operator:wikidata=Q245343
operator:wikipedia=en:Barclays
```

There is definitely a bunch of tag redundancy here, but I don't really see this as a big problem.  These tags give renderers a lot of options on how to display ATMs - anything from "no name" to "literal name of bank" to "name of bank + 'ATM'" 

I'm inclined to leave it alone unless it's causing a real problem for someone.

peternewman:
Yeah maybe it's as I was thinking as I wrote it and it should be raised in the renders/iD etc to append the type of object to the name, although I suspect that's not consistent in all cases e.g. https://www.mylondon.news/news/south-london-news/name-new-battersea-power-station-21615110

Issue #3748, opened 2020-04-20
Amazon Locker


UKChris-osm:
I've just noticed two entries under **vending_machine** for Amazon Locker, which differ in language only and use the same wikidata article, and just wanted to ask:

- Can these be merged into a single "_Amazon Locker_" entry with multiple languages and locations?
- Does this entry need its own "_Amazon Locker_" wikidata entry, or could it just use the standard "_Amazon_" wikidata entry, and as such the standard Twitter hosted logo that accompanies it?

tas50:
Amazon is fine as the wikidata entry if there isn't one for amazon locker specifically.

bhousel:
These are the entries:
https://github.com/osmlab/name-suggestion-index/blob/8cae61bbbe18be164e46256593b5517818015414/brands/amenity/vending_machine.json#L2-L29

(Pro tip! You can link to lines of code on GitHub by shift-selecting the lines, clicking the … meatball menu, and selecting "copy permalink".  Then paste it into an issue so we can talk about it easier).

Because Amazon Locker has its own wikidata and wikipedia identies (like many Amazon products and sub-businesses do), what we have in NSI is correct.

* They should not be merged, because they use different languages
* The Amazon Locker wikidata and wikipedia links are correct and should not be changed to the parent Amazon entry.

PR #7140, opened 2022-10-04
Added new parcel lockers

serhii-muchychka:
According to the wiki, the name tag is discouraged, so I haven't added it yet and will wait for advice
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dparcel_locker

### DIFF
--- a/data/brands/amenity/parcel_locker.json
+++ b/data/brands/amenity/parcel_locker.json
@@ -264,6 +264,17 @@
       }
     },
     {
+      "displayName": "Meest",
+      "locationSet": {"include": ["ua"]},
+      "matchNames": ["міст"],
+      "tags": {
+        "amenity": "parcel_locker",
+        "brand": "Meest",
+        "brand:wikidata": "Q25432124",
+        "name": "Meest"
+      }
+    },
+    {
       "displayName": "Mondial Relay",
       "id": "mondialrelay-318f10",
       "locationSet": {"include": ["be", "fr"]},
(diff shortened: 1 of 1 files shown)


bhousel:
Those edits to discourage the `name` tag on the wiki were made just in the past week, by wiki editors who I dont recognize.
https://wiki.openstreetmap.org/w/index.php?title=Tag%3Aamenity%3Dparcel_locker&type=revision&diff=2411122&oldid=2402396

I really wish people wouldn't do this.

1ec5:
> Those edits to discourage the `name` tag on the wiki were made just in the past week, by wiki editors who I dont recognize. https://wiki.openstreetmap.org/w/index.php?title=Tag%3Aamenity%3Dparcel_locker&type=revision&diff=2411122&oldid=2402396

Of these two edits, one is consistent with [the `amenity=parcel_locker` proposal](https://wiki.openstreetmap.org/wiki/Special:PermanentLink/2261304) that was approved, though I didn’t notice that detail until now. Apparently the idea was to make it similar to ATM tagging.

The other edit is talking about putting the locker’s unique identifier in `ref` instead of `name`. This seems consistent with how NSI prefers to put individual store location names in `branch` instead of `name` in most cases.

bhousel:
We've discussed this extensively in the past and the `name` of these Amazon Lockers was always formatted a certain way, at least back in 2020 - 2021 , prior to the introduction of the parcel locker tag.  I even adjusted how NSI code works based on the Amazon Locker issue, and based on how other more mainstream maps display these features.

https://github.com/osmlab/name-suggestion-index/issues/3748
https://github.com/openstreetmap/iD/pull/8305#issuecomment-778710958
https://github.com/openstreetmap/iD/pull/8305#issuecomment-779931193
https://github.com/osmlab/name-suggestion-index/pull/4902#issuecomment-779911733

For ATMs we do the same thing.  Plenty of legacy renderers still use the `name` tag as a display string:
https://github.com/osmlab/name-suggestion-index/commit/46ab8eba06f31571b02039883f59ec5a6204ec00#commitcomment-44125210
https://github.com/osmlab/name-suggestion-index/issues/2883#issuecomment-726160004
https://github.com/osmlab/name-suggestion-index/issues/5338

The most important thing is - people should participate in the discussions we already have rather than just changing the wiki.  If having a `name` tag on a vending machine causes actual problems for anyone making a map, I'd love to hear it.

Issue #6919, opened 2022-08-12
Swiss `post_box`es should be `colour=yellow`


habi:
All Swiss `amenity=post_box`es operated by the Swiss postal service are distinctively yellow (see https://wiki.openstreetmap.org/wiki/Switzerland/amenity%3Dpost_box).
I'd like to add `colour=yellow` to [the relevant entry](https://nsi.guide/index.html?t=operators&k=amenity&v=post_box&tt=Q614803#diepost-c6967d), but could *not* find it https://github.com/osmlab/name-suggestion-index/blob/main/data/operators/amenity/post_box.json or by [searching for the relevant Wikidata ID](https://github.com/osmlab/name-suggestion-index/search?q=Q614803).

If anyone can give me a pointer on where to change this I'm happy to produce a PR.

bhousel:
Hey @habi  - you did have the right file `amenity/post_box.json`, but the post boxes are confusing because they use "templates" to automatically create items based on items from other files.  

I ended up solving this in 53616ff by _excluding_ Swiss post from the default template, and creating a special template that only _includes_ Swiss Post but also adds the `colour=yellow` tag.

habi:
Thanks for the explanation, which makes perfect sense. And the fix!

UKChris-osm:
@bhousel a while ago in the UK, some post boxes were painted a different colour for various reasons (local / national celebrations) and I was just wondering how this might effect adding the colour tag in this way?

If, for example, a post box is repainted, and a local mapper updates the colour, will the NSI then continue to flag this as an error?

LaoshuBaby:
> @bhousel a while ago in the UK, some post boxes were painted a different colour for various reasons (local / national celebrations) and I was just wondering how this might effect adding the colour tag in this way?
> 
> If, for example, a post box is repainted, and a local mapper updates the colour, will the NSI then continue to flag this as an error?

use `preserveTag:["colour"]` to avoid overwrite?

bhousel:
Yea @UKChris-osm and @LaoshuBaby you're both right .. Adding a `colour=yellow` tag here would raise errors for occurrences of these post boxes with other colors, unless we set the `preserveTag` to allow mappers to use other values.

UKChris-osm:
It's been a long while since I've contributed to the NSI, so am a bit behind - so does `preserveTag` prevent any suggestion appearing in iD if the tag in question already exists?

bhousel:
[...]

PR #7154, opened 2022-10-06
Improve names and tags for Correios depots and post offices

### LINKED ISSUE #7100: Move Brazilian post office (Correios) to brands tree
Hey guys!

I saw the discussions in this repository [here](https://github.com/osmlab/name-suggestion-index/issues/5561), [here](https://github.com/osmlab/name-suggestion-index/pull/5613) and [here](https://github.com/osmlab/name-suggestion-index/issues/5625).

However, as [we discussed in the Community Forum](https://community.openstreetmap.org/t/how-to-properly-tag-a-post-office-brand-operator/3056), we got to the conclusion that, at least in case of _[Correios](https://nsi.guide/index.html?t=operators&k=amenity&v=post_office&tt=correio)_, we should use `brand` instead of `operator`.

The reason is that all _Correios_ (which is a public company, much like USPS) branches have this brand, but some of them are private operated. So it is not correct to tag all of them with `operator=Correios`.

Is it possible to move it back to the brands tree?

Also, I would like to ask to create another NSI suggestion: _Correios_ as a amenity=post_depot (not sure if I should open a new issue for that). So we would end with 3 _Correios_ suggestions: `post_office`, `post_depot` and `post_box`

Thanks a lot!

vgeorge:
This should address the issues discussed in #7100. Changes:

- Use extended names to differentiate post offices from depots
- Use `brand:*` instead of `operator:*` in post offices

cc @matheusgomesms @UKChris-osm @willemarcel  

---

Fix #7100

### DIFF
--- a/data/brands/amenity/post_office.json
+++ b/data/brands/amenity/post_office.json
@@ -370,6 +370,22 @@
         "name:zh": "顺丰速运",
         "short_name": "顺丰"
       }
+    },
+    {
+      "displayName": "Agência dos Correios",
+      "id": "correios-e82118",
+      "locationSet": {"include": ["br"]},
+      "matchNames": [
+        "ebct",
+        "ect",
+        "empresa brasileira de correios e telégrafos"
+      ],
+      "tags": {
+        "amenity": "post_office",
+        "brand": "Empresa Brasileira de Correios e Telégrafos",
+        "brand:short": "Correios",
+        "brand:wikidata": "Q3375004"
+      }
     }
   ]
 }
--- a/data/operators/amenity/post_depot.json
+++ b/data/operators/amenity/post_depot.json
@@ -32,7 +32,7 @@
       }
     },
     {
-      "displayName": "Correios",
+      "displayName": "Centro de Distribuição dos Correios",
       "id": "correios-28fd5c",
       "locationSet": {"include": ["br"]},
       "matchNames": [
@@ -42,7 +42,8 @@
       ],
       "tags": {
         "amenity": "post_depot",
-        "operator": "Correios",
+        "operator": "Empresa Brasileira de Correios e Telégrafos",
+        "operator:short": "Correios",
         "operator:type": "public",
         "operator:wikidata": "Q3375004"
       }
--- a/data/operators/amenity/post_office.json
+++ b/data/operators/amenity/post_office.json
@@ -203,7 +203,7 @@
       }
     },
     {
-      "displayName": "Correios",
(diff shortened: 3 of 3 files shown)


matheusgomesms:
Thanks a lot, @vgeorge!! It looks great to me! A couple of details:

1 - I guess we would have to delete the entry from `operators` folder and create another one in `brands` folder, for the `post_office` tag (someone from NSI please confirm that).

2 - It seems that [post boxes are auto populated from post offices](https://github.com/osmlab/name-suggestion-index/issues/6919). While I see why you changed the name, to be clarified when editing, in this case, shouldn't it still be `displayName=Correios`? This is only the display name, and then we can add `name=Correios`, inside `tags`.

3 - If we add the `name=Correios` tag, then we would probably have to create an exception on `amenity/post_box.json` file, removing the `name` tag, since I guess post boxes should not have a name.

I can help you with that, if you want (and agree with my observations).

vgeorge:
@matheusgomesms I moved the post_office entry from `operators` to `brand` folder. About the post boxes, I believe their display name should be different from the post office. I couldn't find a reference to it on Correios's website, but from this [news article](https://www.jcnet.com.br/noticias/geral/2016/12/488548-70--das-caixas-de-correios-terao-fim.html) it seems the terminology used by the company is "Caixa de Coleta dos Correios".

matheusgomesms:
Great! I support _Caixa de Coleta dos Correios_ for display name on post boxes.

Not sure if you agree, but for post offices I would add the tag `name=Correios` as well. What do you think?

vgeorge:
@matheusgomesms I'm inclined use the extended name "Agência dos Correios", "Centro de Distribuição dos Correios" and "Caixa de Coleta dos Correios" in the `name` tag, but I'm not against `name=Correios`. Maybe we leave this change out to discuss it further with the local community elsewhere? 

I'm not sure how to add the post box entry, it would be great if someone could point which changes should I do.

@LaoshuBaby @bhousel I've changed this PR status to draft, this is work in progress.

bhousel:
Thanks @vgeorge and @matheusgomesms , this seems ok to me.

> I'm not sure how to add the post box entry, it would be great if someone could point out which changes should I do.

The `amenity=post_box` items are generated automatically from these rules:
(basically they copy all the "brand" and "operator" post offices, but remove tags like "name" and "shop")
https://github.com/osmlab/name-suggestion-index/blob/3fb8ab7950b7f1b6b5591dbb627cf2297288f80e/data/operators/amenity/post_box.json#L16-L40

PR #7156, opened 2022-10-07
Modify "easybox" parcel lockers

DnaX:
Those "easybox" parcel lockers, are owned and operated by Sameday, a courier company based in Romania and Hungary. Only some of them are just branded by eMAG for sponsoring reasons but other couriers can drop parcels there.

### DIFF
--- a/data/brands/amenity/parcel_locker.json
+++ b/data/brands/amenity/parcel_locker.json
@@ -170,12 +170,12 @@
     {
       "displayName": "easybox",
       "id": "easybox-ba4d23",
-      "locationSet": {"include": ["ro"]},
+      "locationSet": {"include": ["ro", "hu"]},
       "tags": {
         "amenity": "parcel_locker",
         "brand": "easybox",
-        "operator": "eMag",
-        "operator:wikidata": "Q23827008"
+        "operator": "Sameday",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/parcel_locker.json
```diff
@@ -170,12 +170,12 @@
     {
       "displayName": "easybox",
       "id": "easybox-ba4d23",
-      "locationSet": {"include": ["ro"]},
+      "locationSet": {"include": ["ro", "hu"]},
       "tags": {
         "amenity": "parcel_locker",
         "brand": "easybox",
```
arch0345, line 176: Note that since this is in the brands folder, a `brand:wikidata` tag is needed for this to show up as an iD preset.

DnaX: Here: https://www.wikidata.org/wiki/Q114496224

Should I open a new PR?

arch0345: > Here: https://www.wikidata.org/wiki/Q114496224

Added it in https://github.com/osmlab/name-suggestion-index/commit/bbf233215d1c4c6f503b4ec48fd486bbead94696

Issue #7016, opened 2022-09-07
operator=AEP and other questions


Hiausirg:
"American Electric Power" is currently being replaced with "AEP", this is clearly wrong. The name of the company is "American Electric Power Company", not "AEP". Abbreviations can cause confusion and should therefore not be used. Furthermore, the longer version is used more often (and significantly more often before the id editor started replacing it).

And why are
"        
        "aep ohio",
        "aep texas",
        "american electric power",
        "apco",
        "**appalachian power**",
        "i&m",
        "**indiana michigan power**",
        "**indiana michigan power company**",
        "**kentucky power**",
        "pso",
        "**public service company of oklahoma**",
        "**southwestern electric power company**",
        "swepco"
"
also being replaced with AEP, especially the highlighted ones? Only because American Electric is the parent company? Alabama/Georgia/Mississippi Power Companies have the Southern Company as their parent, but no one would even think of generalize everything here and thus worse the accuracy of the data.

And also, where can I find the consensus for wikidata tags on power features, where were those mass changes/"meatbot" edits discussed?

I am absolutely against such tagging as it is a useless duplication of already existing information that just adds more maintenance to do. The point of such wikidata tags is to provide an unique code if the operator tag differs from each other, i.e. when a fast food company operates in multiple countries with different languages/writing systems, but the company is the same, and not to be spammed on every object possible. 
This example dosen't occur on power features, and if, it is extremely rare and dosen't justify the extra tagging/work.
Also, some kind of navigation apps or apps which display brand logos and think they can't use their own OSM-wikidata conversation tables don't display logos of power features anyway, because they're irrelevant there. 

Dont get me wrong, I support unifying operator tags. But don't make it so complicated, just unify the operator tag. Wikidata is not necessary here.

1ec5:
> "American Electric Power" is currently being replaced with "AEP", this is clearly wrong. The name of the company is "American Electric Power Company", not "AEP". Abbreviations can cause confusion and should therefore not be used. 

Yes, `operator=AEP` is incorrect. By convention, `operator` is either an organization’s trade name or its legal name, which in this case would be either “American Electric Power” or “American Electric Power Company, Incorporated”, respectively. However, “AEP” is incorrect not because it’s an abbreviation but rather because it isn’t one of the names that the company primarily goes by. `operator:short=AEP` would be more appropriate in this case.

As it happens, we already have `operator=American Electric Power` `operator:short=AEP` on `route=power` and `power=generator`:

[...]

PR #7167, opened 2022-10-09
Fix Kinder Morgan Texas

Hiausirg:
"Kinder Morgan Tejas Pipeline" is a seperate operating company, NOT a typo. So please remove it.

Also, before any more wikidata tags are added to pipelines, I want this (:wikidata on man_made=pipeline) to be properly discussed - not just silently added.

### DIFF
--- a/data/operators/man_made/pipeline.json
+++ b/data/operators/man_made/pipeline.json
@@ -2426,14 +2426,9 @@
       "displayName": "Kinder Morgan Texas Pipeline LLC",
       "id": "kindermorgantexaspipelinellc-0a386a",
       "locationSet": {"include": ["us"]},
-      "matchNames": [
-        "kinder morgan tejas pipeline",
-        "kinder morgan tejas pipeline llc"
-      ],
       "tags": {
         "man_made": "pipeline",
         "operator": "Kinder Morgan Texas Pipeline LLC",
-        "operator:wikidata": "Q110987354"
       }
     },
     {
--- a/data/operators/man_made/pipeline.json
+++ b/data/operators/man_made/pipeline.json
@@ -2426,6 +2426,9 @@
       "displayName": "Kinder Morgan Texas Pipeline LLC",
       "id": "kindermorgantexaspipelinellc-0a386a",
       "locationSet": {"include": ["us"]},
+      "matchNames": [
+        "kinder morgan texas pipeline",
+                    ],
       "tags": {
         "man_made": "pipeline",
         "operator": "Kinder Morgan Texas Pipeline LLC",
--- a/data/operators/man_made/pipeline.json
+++ b/data/operators/man_made/pipeline.json
@@ -2429,6 +2429,7 @@
       "tags": {
         "man_made": "pipeline",
         "operator": "Kinder Morgan Texas Pipeline LLC",
+        "operator:wikidata": "Q110987354"
       }
     },
     {


1ec5:
> Also, before any more wikidata tags are added to pipelines, I want this (:wikidata on man_made=pipeline) to be properly discussed - not just silently added.

I’m confused, wasn’t this discussed in #7016?

data/operators/man_made/pipeline.json
```diff
@@ -2426,14 +2426,9 @@
       "displayName": "Kinder Morgan Texas Pipeline LLC",
       "id": "kindermorgantexaspipelinellc-0a386a",
       "locationSet": {"include": ["us"]},
-      "matchNames": [
-        "kinder morgan tejas pipeline",
-        "kinder morgan tejas pipeline llc"
```
1ec5, line 2431: These apparently come from 599b859284b539a44cbd22996dde6bcc6b68b668, which removed two separate entries for Kinder Morgan Tejas Pipeline and Kinder Morgan Tejas Pipeline LLC. If that’s a separate company, I think the correct fix would be to restore one of the deleted entries. Based on your PR description, I’ve tagged [Kinder Morgan Tejas Pipeline (Q30973003)](https://www.wikidata.org/wiki/Q30973003) as “different from” Kinder Morgan Texas Pipeline LLC.

data/operators/man_made/pipeline.json
```diff
@@ -2426,14 +2426,9 @@
...
-      ],
       "tags": {
         "man_made": "pipeline",
         "operator": "Kinder Morgan Texas Pipeline LLC",
-        "operator:wikidata": "Q110987354"
```
1ec5, line 2436: Unless I’ve missed something, [Q110987354](https://www.wikidata.org/wiki/Q110987354) is the item for Kinder Morgan Texas Pipeline LLC. This line is consistent with NSI’s practice within the operators/ tree, even if you disagree with the use of `operator:wikidata` specifically on pipelines.

Hiausirg:
"I’m confused, wasn’t this discussed most recently in https://github.com/osmlab/name-suggestion-index/issues/7016?"
- By "properly discussed", I mean on mailing lists, forums or what-do-i-know. Just not in some random Github pull request.
" If you’re drawing a distinction between Wikidata tags on pipelines versus other infrastructure, it’s a bit surprising."
- I think they're just as pointless as on other infrastructure :) But neither do i have the time nor the patience to try changing that now.

"in all the many discussions about Wikidata tags on the mailing lists" - I would like to read some.

My main points againest:
- Consumers should use normal operator tags. There is no point in matching anything with wikidata - and if -, whatever it is, it won't justify maintaining two tags containing the same info. Stuff like "emergency phone number" are funny gimmicks, but just they're just not ... yeah. How should I imagine this, someone sees a burning pipeline, and the first thing they do is look at some (non-existent) pipeline map that just randomly happens to show some emergency numbers? Instead of looking at pipeline markers or just call 911? It's just nonsense. My point is, because it is theoretically somehow possible to do something with an info, it doesn't mean that we have to take care of providing that info.
The example on the other request (find all facilities/assets by parent company): That's actually a good idea, but first, wikidata query service dosen't even read those wikidata operator tags yet, so they have to develop something anyway. And then they should directly process the normal operator tags and match them with QIDs. This can't be too hard. Secondly, the normal operator tag is much more extensively tagged everywhere, so you get better results there right away.
- I keep an eye on pipeline-related changesets. I am one of very few who actively map them and am -without wanting to brag- the one who tags them the most detailed. (talking about the US). And I don't think it's right that people come along who actually have nothing to do with this mapping and want to enforce (or tell the id-editor to enforce) certain tagging variants.

data/operators/man_made/pipeline.json
Hiausirg: re-added it.

data/operators/man_made/pipeline.json
Hiausirg: Should be fixed

1ec5:
> "in all the many discussions about Wikidata tags on the mailing lists" - I would like to read some.

First off, my goal is only to show you that NSI’s adoption of Wikidata tags didn’t come out of a fever dream one night. It wasn’t adopted until there was enough grassroots support for the idea to make it worth the maintainers’ time. It would be impossible to point to a credible, definitive statement that these tags are unconditionally accepted in OSM, because that’s not how OSM works. There is virtually no corner of OSM‘s tag repertoire that is free of controversy, so it should be unsurprising that, no matter how well-established these Wikidata tags have become, there remains hostility to the idea.

With that out of the way, here’s [a partial index of Wikidata-related discussions](https://wiki.openstreetmap.org/wiki/User:Minh_Nguyen/Wikidata_discussions) on the mailing lists, forums, and wiki over the years. I made sure to include some of the longest, most memorable firestorms, as well as later threads showing that some of the participants had apparently warmed up to the idea of Wikidata tagging. There have been a lot of passions for and against it, but most of the controversy is well in the past by this point.

You’ll notice that none of these discussions touched on the issue of adding `operator:wikidata` on a pipeline in the United States for which the `operator` uniquely identifies a company. One can add multiple qualifiers until a condition becomes too difficult to satisfy, but that would require moving the goalposts.

> Stuff like "emergency phone number" are funny gimmicks, but just they're just not ... yeah. How should I imagine this, someone sees a burning pipeline, and the first thing they do is look at some (non-existent) pipeline map that just randomly happens to show some emergency numbers? Instead of looking at pipeline markers or just call 911? It's just nonsense. My point is, because it is theoretically somehow possible to do something with an info, it doesn't mean that we have to take care of providing that info.

Did anyone suggest Wikidata as an alternative to calling 9-1-1? I sure didn’t. In https://github.com/osmlab/name-suggestion-index/issues/7016#issuecomment-1239104324, I mentioned contact information but wasn’t speaking about pipelines or power lines specifically. Besides, there are other reasons to want to contact a utility. Anyways, my point is that including an external identifier relieves us of the need to fuss about information tangential to the feature when mapping the feature. The more tangential the information, the more difficult it is to keep consistent and current on a feature-by-feature level.

>  The example on the other request (find all facilities/assets by parent company): That's actually a good idea, but first, wikidata query service dosen't even read those wikidata operator tags yet, so they have to develop something anyway. And then they should directly process the normal operator tags and match them with QIDs. This can't be too hard.

If you’re looking for a tool that can query OSM and Wikidata at the same time, [Sophox](https://sophox.org/) has been useful for [a lot of things](https://wiki.openstreetmap.org/wiki/SPARQL_examples) that Overpass isn’t good at. (Caveat emptor: Sophox/sophox#27. I think the tool would be in better shape these days if those working on linked data as a hobby weren’t so often vilified.)

When writing federated queries, you can make assumptions about names being unique, but such assumptions severely limit the query’s domain and potentially affect its durability. It’s no different when writing an application that joins OSM with other datasets.

> I keep an eye on pipeline-related changesets. I am one of very few who actively map them and am -without wanting to brag- the one who tags them the most detailed. (talking about the US). And I don't think it's right that people come along who actually have nothing to do with this mapping and want to enforce (or tell the id-editor to enforce) certain tagging variants.

Your time and efforts are appreciated, but nevertheless what you’re contributing to is a community resource. Broadly speaking, I don’t think it’s healthy to feel personal possessiveness over a whole class of features. Some mappers specialize, while others focus on making the map more well-rounded and consistent across themes, but we shouldn’t view ourselves as separate factions waging a turf battle.

For what it’s worth, I used to be pretty much the only one mapping power lines and pipelines across several states back in the day. I’m glad to see subject-matter experts like you run with that aspect of the map. We shouldn’t let this disagreement over Wikidata tags hamper cooperation or derail the needed improvement you’re making in this PR.

Hiausirg:
Thanks for the list - interesting points from both sides!

"Anyways, my point is that including an external identifier relieves us of the need to fuss about information tangential to the feature when mapping the feature. The more tangential the information, the more difficult it is to keep consistent and current on a feature-by-feature level."
This is undoubtedly something important. To make it clear, I have nothing against (operator:)wikidata tags in general. They can  be helpful. If it is suggested to me to add them somewhere, I usually accept without hesitating.
_However_, there is no such information for pipelines - or more precisely, such info exists, but there is no reason why any data consumer would want to use it here. This info provided by Wikidata is simply not needed by anyone - at least not in our context here. If one ever finds something where one can use them usefully, then you can still add it. But until then it is not worth our time to maintain those tags.

PR #7171, opened 2022-10-10
Further cleanup of UK power operators

russss:
Tidy up a few historic operators which I missed in #7153, add wikidata for some more others.


data/operators/power/substation.json
```diff
@@ -1760,26 +1750,6 @@
         "power": "substation"
       }
     },
-    {
```
LaoshuBaby, line 1763: You delete `EDF Energy` and `EDF Energy Networks Ltd.`, but those 2 operator's name wasn't used in any matchNames, so every time run auto-formatter, they will be recreated:

图片

---

Other deleted operator's name was other entry's matchName, so they won't be created again.

russss: Thanks, I think that's fixed now.

Issue #6788, opened 2022-07-16
`npm run build` will destroy correct data and add English name to Japanese name


LaoshuBaby:
After #6787 's merge I use `npm run build` command to auto format the data, but I noticed the Japanese name isn't I just merged.

I manually checked and ensure them, then I run `npm run build` in a clear and up-to-date environment, here is a video:

https://user-images.githubusercontent.com/42690037/179353635-0449b0b1-9d07-4490-8119-776dedb427d0.mp4

(Sorry, old video mistake cut key point, so this is the full screenrecord)

If video can't been play:

![图片](https://user-images.githubusercontent.com/42690037/179353498-0f8be1a3-6caf-4b7b-89d1-dc17dcac1d0f.png)

This may destroy any Japanese name in current data, but I don't know how to do with it. 

I run command on this commit: https://github.com/osmlab/name-suggestion-index/commit/b9a71323af8a3fa157a7b0fceb1d347a7b131f22

@bhousel

bhousel:
Oh I see - yeah it does look like the build script is changing `name:ja` value to be equal to `name` value which contains the mixed ja and en strings (and same with `brand).

I remember I added this behavior a while ago because many validators (JOSM and KeepRight and maybe others) expect that is there is _any_ local tag like `name:XX` one of those local values _must_ match the value in `name`.  We apply this rule to tags like `brand:XX` and `brand` also.

#2245 has a lot more info about it and some screenshot of the warning.

Do you think it would be acceptable for Japanese mapping community if we remove the mixed string values in `name` and `brand` and just keep only the Japanese value?

LaoshuBaby:
Emmm……In fact this is related to HK community. For this issue (Japanese brand in Hong Kong and call themselves only in `ja+en`, the `brand:ja` and `name:ja` is simply Japanese and don't need to obey Hong Kong's `T-Chinese + English` bilingual guideline), those 2 case is very strange.

https://github.com/osmlab/name-suggestion-index/blob/13120890f250f844035df64f13112a5c1b26eedf/data/brands/amenity/fast_food.json#L5845

---

> Do you think it would be acceptable for Japanese mapping community if we remove the mixed string values in `name` and `brand` and just keep only the Japanese value?

I think this suggestion may won't accepted because some place have different bilingual/multilingual rule, they are all documented in https://wiki.openstreetmap.org/wiki/Multilingual_names

[...]

PR #7182, opened 2022-10-12
Correcting Serbian brands to latin script

stalker314314:
There were some Serbian brands that were imported/added with Cyrillic script, but they should be in Latin. We use both Cyrillic and Latin scripts, but some names (like "Maxi" => "Maksi"/"Макси" and other corrected in this PR) are never used in Cyrillic, never written as such, there are no signs on these shops that are written like this (ground truth), website do not use them and are just awkward to be read like that. This PR corrects those found and keeps "name:sr" in Cyrillic for those, but converts "name" and "brand" to Latin versions

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -10307,15 +10307,15 @@
       }
     },
     {
-      "displayName": "АИК банка",
+      "displayName": "AIK banka",
       "id": "aikbank-e9ea45",
       "locationSet": {"include": ["rs"]},
       "matchNames": ["aik banka", "аик банк"],
       "tags": {
         "amenity": "bank",
-        "brand": "АИК банка",
+        "brand": "AIK banka",
         "brand:wikidata": "Q293650",
-        "name": "АИК банка",
+        "name": "AIK banka",
         "name:en": "AIK bank",
         "name:sr": "АИК банка",
         "name:sr-Latn": "AIK banka"
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -5494,7 +5494,7 @@
     {
       "displayName": "ОМВ",
       "id": "637ba4-7131c1",
-      "locationSet": {"include": ["bg", "rs"]},
+      "locationSet": {"include": ["bg"]},
       "tags": {
         "amenity": "fuel",
         "brand": "ОМВ",
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -6917,6 +6917,7 @@
       "displayName": "Univerexport",
       "id": "univerexport-b83687",
       "locationSet": {"include": ["rs"]},
+      "matchNames": ["универекспорт", "univereksport"],
       "tags": {
         "brand": "Univerexport",
         "brand:wikidata": "Q12747294",
@@ -7831,10 +7832,11 @@
       "displayName": "Лидл (Србија)",
       "id": "366fb7-b83687",
       "locationSet": {"include": ["rs"]},
+      "matchNames": ["лидл"],
       "tags": {
-        "brand": "Лидл",
+        "brand": "Lidl",
         "brand:wikidata": "Q114509929",
-        "name": "Лидл",
+        "name": "Lidl",
         "name:sr": "Лидл",
         "name:sr-Latn": "Lidl",
         "shop": "supermarket"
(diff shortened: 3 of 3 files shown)


Adamant36:
I'm not a native speaker, but the whole idea of name:sr-Latn (Serbian Latin I assume?) really seems oxymoronic and pointless since our essentially just re-creating the name tag with another tag. Both are "AIK banka." Plus you still have "AIK bank" for the name:en tag. Which just makes the whole thing all the more convoluted. At the end of the day whatever the "main" language is should be used in the name tag and there should also be name:en if it's not Latin script. Anything other then that is pointless though. Especially if they are only superficial different, "AIK banka", "AIK bank", "AIK banco", "AIK banc", Etc. Etc. The possibilities are essentially endless.

stalker314314:
As per [this](https://wiki.openstreetmap.org/wiki/WikiProject_Serbia#%D0%9D%D0%B0%D0%B7%D0%B8%D0%B2%D0%B8), it is _strongly recommended_ to add "name:sr-Latn". We are usually adding it ([name:sr](https://taginfo.openstreetmap.org/keys/name:sr) vs [name:sr-Latn](https://taginfo.openstreetmap.org/keys/name:sr-Latn) ). I agree it is easily inferred as it is same as "name" and I can remove it from those entries I touched here (but I would not change/remove it everywhere where it is same as "name"), but don't be surprised if in 2 years time someone comes and wants to add them (you can treat it also as a bit of political issue, so we tend to have both). Just let me know if you want to remove them where it is same as "name".

LaoshuBaby:
I prefer to reject this PR. Because I noticed https://wiki.openstreetmap.org/wiki/Multilingual_names#Serbia and it said **The default name=\* tag should be filled in using Serbian Cyrillic**, and if not mentioned, `brand` follow the same guideline as `name`.

Looks like many preset about Serbia is added by @u-spec-png , I want to request a contributor that have local knowledge for help.

stalker314314:
Default name should be written as on the ground truth, especially when name uses letters that are not in cyrillic (as with letter "x" as in "maxi" or with "y" - you just [cannot translate these](https://www.openstreetmap.org/node/5164080633)). I didn't change everything to cyrillic, just the ones that are never in cyrillic (as seen on web site, local signs...). If it helps, here is my OSM username: https://www.openstreetmap.org/user/Branko%20Kokanovic to show I am not some random person, but person with credibility in OSM and with local knowledge.

I can provide couple of example:
Maxi:
* https://www.mapillary.com/app/?lat=44.8031767&lng=20.3863617&z=17&pKey=290832236087293&focus=photo
* https://www.mapillary.com/app/?lat=44.8031767&lng=20.3863617&z=17&pKey=290832236087293&focus=photo

Roda:
* https://www.mapillary.com/app/?lat=44.810347&lng=20.39807&z=17&pKey=1113981082418069&focus=photo
* https://www.mapillary.com/app/?lat=44.831616198644&lng=20.37501999177&z=17&pKey=956617898423632&focus=photo

Tempo:
* https://www.mapillary.com/app/?lat=44.8285476&lng=20.3626441&z=17&pKey=823539585193684&focus=photo

and so on. It would be good if @u-spec-png can provide just one example (photo, web site, mapillary) of those entities I changed (maksi, roda, tempo, lidl, omv, aik banka) that they found somewhere in cyrillic (макси, рода, лидл, омв, аик банка). If there are, I am fine with rejection.

bhousel:
Thanks @stalker314314 for the examples and for educating us on the issues around Serbian naming.  I agree that this PR seems fine and we will try to be aware going forward that the `name` field, often written in Cyrillic, might also often be written in Latin script for well known brands.  

@LaoshuBaby makes a good point - @stalker314314 can you please update the guidance on OSM wiki?  It currently says "The default name=* tag should be filled in using Serbian Cyrillic", and it sounds like that guidance is more for placenames, not necessarily big brands that write their name in Latin script.

@Adamant36, please don't call people's contributions "pointless".  It's normal practice to duplicate the `name` tag value in another local name tag, and there are KeepRight and JOSM validation warnings that encourage this style of mapping - check  https://github.com/osmlab/name-suggestion-index/issues/6788#issuecomment-1187755052 for an example.  Our normal practice on this project is to make sure that if there are _any_ local name tags, one of them should match the `name` tag.  (but there are exceptions like #6788 where Japan, Hong Kong, maybe a few others, prefer their `name` tags written with a multilingual value).

Adamant36:
"please don't call people's contributions "pointless." 

The duplicate name tags were pointless. Not the contribution in general. That should have been obvious since it was literally what my message was about 🤦‍♂️

stalker314314:
Thanks @bhousel for understanding.

@LaoshuBaby - I will update wiki soon!

@Adamant36 I understood your point. History of "name:sr-Latn" might seem weird and historical (sometimes even a bit contentious) and I agree it is usually pointless having it, but this is for some other topic:) Let's just say that I would not go that far to remove it completely from NSI without consensus from local community!

Adamant36:
"Let's just say that I would not go that far to remove it completely from NSI without consensus from local community!" 

That's fine. I was just giving my opinion as a calibrator or whatever. I don't expect you to  do anything that's against the local consensus though.

PR #7183, opened 2022-10-13
Remove Vorarlberger Kraftwerke AG

mcliquid:
[Vorarlberger Kraftwerke AG](https://de.wikipedia.org/wiki/Vorarlberger_Kraftwerke ) was merged into [illwerke vkw AG](https://de.wikipedia.org/wiki/Illwerke_vkw) in 2019 and is no longer a legal entity.

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -4317,16 +4317,6 @@
       }
     },
     {
-      "displayName": "Vorarlberger Kraftwerke",
-      "id": "vorarlbergerkraftwerke-3f60da",
-      "locationSet": {"include": ["at"]},
-      "tags": {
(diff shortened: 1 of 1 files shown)


Adamant36:
Unfortunately there isn't really a good way to deal with things like this since it can take time for rebranding to happen. You usually can't just change the NSI the second the companies merge because they might have changed the actual branding yet. Although it's been a few years in this case. So that probably isn't an issue, but have all the locations been changed over IRL yet?

bhousel:
Thanks @mcliquid - 
In b1b89c590 I also added `matchName` properties for the old names so that users can be prompted to upgrade the name of the existing charging stations, and so that our build script won't pull the old names back in..

PR #7185, opened 2022-10-13
added Keurslager

wvanderp:
In this PR, I added the dutch butcher chain Keurslager.

I've one question about my contribution most of the stores are called: `<name of the butcher>, Keurslager`
does this entry capture that name?

The `Club Alpino Italiano` resulted from building the index and is unrelated to this PR

Let me know if I need to change anything. I'm happy to oblige

### DIFF
--- a/data/brands/shop/butcher.json
+++ b/data/brands/shop/butcher.json
@@ -71,6 +71,18 @@
       }
     },
     {
+      "displayName": "Keurslager",
+      "id": "keurslager-1c7101",
+      "locationSet": {"include": ["nl"]},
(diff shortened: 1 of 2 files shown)


bhousel:
> I've one question about my contribution most of the stores are called: `<name of the butcher>, Keurslager`
> does this entry capture that name?

Thanks @wvanderp 👍   
Yes this looks correct to me.. By adding the `"preserveTags": ["^name"]` property, it means that the preset will set the `name` to "Keurslager" only when the feature gets added, but mappers can customize the store name to be something else and won't be prompted to change it.

Issue #7188, opened 2022-10-13
Bangsamoro and some Native American flags


ianlopez1115:
Consider adding the following to the index:

### Bangsamoro

flag:name = Bangsamoro
flag:type = regional
flag:wikidata = Q74484535
man_made = flagpole
subject = Bangsamoro
subject:wikidata = Q24869612

### Cherokee Nation

flag:colour = orange
flag:name = Cherokee Nation
flag:type = national
flag:wikidata = Q5456858
man_made = flagpole
subject = Cherokee Nation
subject:wikidata = Q14708404

### Chickasaw Nation

flag:colour = indigo
flag:name = Chickasaw Nation
flag:type = national
flag:wikidata = Q114643306
man_made = flagpole
subject = Chickasaw Nation
subject:wikidata = Q3336808

### Choctaw Nation of Oklahoma

flag:colour = purple
flag:name = Choctaw Nation of Oklahoma
flag:type = national
flag:wikidata = Q114642869
man_made = flagpole
subject = Choctaw Nation of Oklahoma
subject:wikidata = Q5103726

### Muscogee (Creek) Nation

flag:colour = white
flag:name = Muscogee Nation
flag:type = national
flag:wikidata = Q114554335
man_made = flagpole
subject = Muscogee Nation
subject:wikidata = Q6940348

### Pawnee Nation of Oklahoma

flag:colour = blue
flag:name = Pawnee Nation of Oklahoma
flag:type = national
flag:wikidata = Q114643616
man_made = flagpole
subject = Pawnee Nation of Oklahoma
subject:wikidata = Q750947

### Seminole Tribe of Florda

flag:name = Seminole Tribe of Florida
flag:type = national
flag:wikidata = Q108837351
man_made = flagpole
subject = Seminole Tribe of Florida
subject:wikidata = Q7449545

### Seminole Nation of Oklahoma

flag:colour = white
flag:name = Seminole Nation of Oklahoma
flag:type = national
flag:wikidata = Q114642332
man_made = flagpole
subject = Seminole Nation of Oklahoma
subject:wikidata = Q7449528

### Notes
* Bangsamoro is an autonomous region in the Philippines established in 2019 replacing the Autonomous Region in Muslim Mindanao
* I've tagged flags representing Native American nations and similar entities as `flag:type = national` following the precedent set in [#5786](https://github.com/osmlab/name-suggestion-index/issues/5786)

Issue #5786, opened 2021-12-04
flagpole preset request: Navajo Nation


ArloJamesBarnes:
discussion context: https://osmus.slack.com/archives/CDBMHU6QY/p1610731707001100?thread_ts=1610586392.001000&cid=CDBMHU6QY

flag WD: https://wikidata.org/wiki/Q20020313
nation WD: https://wikidata.org/wiki/Q1783171

This is quite a large polity by area, so I chose it to kick off what I hope to be a series of flagpole preset requests of indigenous sovereignties in the Americas (preceded by https://github.com/osmlab/name-suggestion-index/issues/4892) due to the likelihood of there being a significant number of mappable flagpoles IRL.

Thanks!

bhousel:
> This is quite a large polity by area, so I chose it to kick off what I hope to be a series of flagpole preset requests of indigenous sovereignties in the Americas

I love this idea!

tas50:
Glad to see this issue come up. I ran into a tribe flag the other day and it got me thinking about how we could get them all added to the index. It's a big project

Issue #4892, opened 2021-02-04
Wiphala / Bolivia


bhousel:
Following a convo on Slack with @1ec5 and @ArloJamesBarnes 
Bolivia is an interesting case because their constitution allows the Wiphala as a _dual flag_.
https://en.wikipedia.org/wiki/Flag_of_Bolivia
https://en.wikipedia.org/wiki/Wiphala

So we want to make sure that
- there is a suitable item in NSI for the Wiphala (possibly https://www.wikidata.org/wiki/Q1193967 as the symbol or a new QID as the flag?)
- test the matcher can handle a reasonable tagging of this, and won't try upgrading a Wiphala flag to Bolivian national flags based on the `country` or `subject` tags.

example flagpole:
https://www.openstreetmap.org/node/8393427376

ArloJamesBarnes:
(research brain dump, not immediately actionable)

Apropos of https://github.com/osmlab/name-suggestion-index/discussions/4759, there are four colour-order variants (from hoist to fly, and using the same looping sequence with different start points) for the four suyus (quadrants of the former Inca holdings as centered on Cusco), which Wiki[pedia/media Commons] refers to by the major diagonal: `yellow;orange;red;purple;blue;green;white;yellow;orange;red;purple;blue;green` (white diagonal, this seems to be the most common in photos and stands for Qulla Suyu), `orange;red;purple;blue;green;white;yellow;orange;red;purple;blue;green;white` (yellow diagonal, stands for [Q/K]unti[n] Suyu), `purple;blue;green;white;yellow;orange;red;purple;blue;green;white;yellow;orange` (red diagonal, stands for Chincha[y/u] Suyu), and `white;yellow;orange;red;purple;blue;green;white;yellow;orange;red;purple;blue` (green diagonal, stands for Anti[n] Suyu which has an alternate too). Sometimes pink is used instead of purple in these. There's also a couple unlikely to be on the map, as they are historical, designed by Túpac Katari.

Other test case: https://osm.org/node/8403476132

bhousel:
I added a Wiphala entry in 426fbd45 :
```js
{
  "displayName": "Wiphala",
  "id": "wiphala-e5dc93",
  "locationSet": {"include": ["001"]},
  "tags": {
    "flag:name": "Wiphala",
    "flag:type": "cultural",
    "flag:wikidata": "Q1193967",
    "man_made": "flagpole"
  }
},
```

This might not be perfect, but here's what I was thinking:
[...]

PR #7195, opened 2022-10-14
Add Bangsamoro and some Native American flags

bhousel:
This adds the flags requested in #7188.  
I did make a small change though..

@ianlopez1115 said:
> I've tagged flags representing Native American nations and similar entities as `flag:type=national` following the precedent set in https://github.com/osmlab/name-suggestion-index/issues/5786

I looked through the current list of flags and - since we've started adding these - a few other people (e9b17fe78, 89661e68, maybe others) have added indigenous flags for Australia and New Zealand and tagged them as `flag:type=indigenous`.  I think this is a better descriptive value for the flag type, and since other people are using it, it seems ok..  I updated the Wiphala flag (#4892) to use this value too.

I notice that `flag:type=cultural` and `flag:type=indigenous` are not documented yet on https://wiki.openstreetmap.org/wiki/Key:flag:type so it would be great if someone would add these values.

Anyway happy (belated) [Indigenous People's Day](https://en.wikipedia.org/wiki/Indigenous_Peoples%27_Day)!

### DIFF
--- a/data/flags/man_made/flagpole.json
+++ b/data/flags/man_made/flagpole.json
@@ -218,10 +218,9 @@
     },
     {
       "displayName": "AU - Australian Aboriginal",
-      "id": "australianaboriginal-e5dc93",
-      "locationSet": {"include": ["001"]},
+      "id": "australianaboriginal-aec753",
+      "locationSet": {"include": ["au"]},
       "tags": {
-        "country": "AU",
         "flag:name": "Australian Aboriginal",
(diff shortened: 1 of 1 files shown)


ianlopez1115:
`flag:type=cultural` and `flag:type=indigenous` are now documented ([see this revision](https://wiki.openstreetmap.org/w/index.php?title=Key:flag:type&oldid=2420167))

PR #7198, opened 2022-10-14
Add Foyles (UK bookshop chain)

tommorris:


data/brands/shop/books.json
```diff
@@ -326,6 +326,17 @@
         "shop": "books"
       }
     },
+    {
+      "displayName": "Foyles",
+      "id": "foyles-4cb062",
+      "locationSet": {"include": ["gb"]},
```
codeinabox, line 332: The shops are only in England so could this be changed to:

```json
"locationSet": {"include": ["gb-eng"]},
```

codeinabox (changes requested):

data/brands/shop/books.json
tommorris: Done, thanks.

UKChris-osm:
Thank you for your first contribution @tommorris 👍

PR #7199, opened 2022-10-14
Add IDEA organic and IDEA Super

u-spec-png:


data/brands/shop/supermarket.json
```diff
@@ -3082,12 +3082,14 @@
     },
     {
       "displayName": "IDEA",
```
LaoshuBaby, line 3084: I notice Q114669650 is IDEA's subsdiare "IDEA Super", could you check this?

u-spec-png: Forgot to change that I'm sorry

PR #7202, opened 2022-10-15
Add Laguna

u-spec-png:


LaoshuBaby:
Looks like you use cyrillic script in displayname and name, don't worry, I don't means you must change that. Because some of brand you added was changed in #7182, and I want to know your opinion about this——should those brand change to latin script. 

There isn't technical error, it pass the CI, and I don't familiar with any slavic language, so whether you choose change or not change, I'll merge this.

u-spec-png:
@LaoshuBaby I don't see a reason to change the brand name. _Laguna_ is the name of the brand in Latin script, _Лагуна_ is in Cyrillic but both names are correct, but Cyrillic is the official script and it's also the script in which map is. I don't see a reason to mix these two.

PR #7209, opened 2022-10-16
Add Železnice Srbije

u-spec-png:


Dimitar5555:
route=railway applies to the railway lines (example: railway line 1 Nis - Beograd), while route=train applies to the train line (Nis - Beograd express train). It seems like Serbian railways operates the trains, not the infrastructure.

u-spec-png:
[Srbija Voz](https://en.wikipedia.org/wiki/Srbija_Voz) operates trains, Železnice Srbije opertates infrastructure.

Dimitar5555:
Isn't Železnice Srbije some sort of a holding company? According to Wikipedia (https://en.wikipedia.org/wiki/Srbija_Voz) it was split in 3 separate companies in 2015.

> In March 2015, the Government of Serbia announced its plan to establish three new railway companies, splitting the Serbian Railways state-owned company in separate businesses – passenger (Srbija Voz), cargo (Srbija Kargo) and infrastructure (Serbian Railways Infrastructure).

PR #7212, opened 2022-10-16
change location of "Nord-Ostsee Sparkasse"

Discostu36:
This bank is only present in the German state Schleswig-Holstein


arch0345:
Thanks @Discostu36 and welcome!

PR #7214, opened 2022-10-17
Update electronics.json

Mannivu:
In Italy we don't have MediaMarkt, but we have the subsidiary MediaWorld.


UKChris-osm:
Thank you for the update @Mannivu 👍

PR #7216, opened 2022-10-17
Update redirecting Shell wikidata

Cj-Malone:
[Q110716465](https://www.wikidata.org/wiki/Q110716465) redirects to Q154950 now.


1ec5:
Thanks for the heads-up, I reverted the changes and added [different from](https://www.wikidata.org/wiki/Property:P1889) statements to hopefully dissuade future editors from reconflating the items.

PR #7217, opened 2022-10-17
[operators] add Westfield, Share Now

Sukkoria:
added Westfield to mall.json, Share Now to car_sharing.json; in car_sharing.json, also merged RTTB into Régie des Transports du Territoire de Belfort.

Have a nice day!

### DIFF
--- a/data/operators/amenity/car_sharing.json
+++ b/data/operators/amenity/car_sharing.json
@@ -463,6 +463,7 @@
       "displayName": "Régie des Transports du Territoire de Belfort",
       "id": "regiedestransportsduterritoiredebelfort-672a50",
       "locationSet": {"include": ["fx"]},
+      "matchNames": ["rttb"],
       "tags": {
         "amenity": "car_sharing",
         "operator": "Régie des Transports du Territoire de Belfort"
@@ -479,21 +480,33 @@
       }
     },
     {
-      "displayName": "RTTB",
-      "id": "rttb-672a50",
-      "locationSet": {"include": ["fx"]},
-      "tags": {
-        "amenity": "car_sharing",
-        "operator": "RTTB"
-      }
-    },
-    {
       "displayName": "Scouter",
       "id": "scouter-088af7",
       "locationSet": {"include": ["de"]},
       "tags": {
         "amenity": "car_sharing",
         "operator": "Scouter"
+      }
+    },
+    {
+      "displayName": "Share Now",
+      "id": "sharenow-6a8c71",
+      "locationSet": {
+        "include": [
+          "at",
+          "ca",
+          "de",
+          "dk",
+          "fx",
+          "it",
+          "nl",
+          "us"
+        ]
+      },
+      "tags": {
+        "amenity": "car_sharing",
+        "operator": "Share Now",
+        "operator:wikidata": "Q62021602"
       }
     },
     {
--- a/data/operators/amenity/car_sharing.json
+++ b/data/operators/amenity/car_sharing.json
@@ -463,10 +463,10 @@
       "displayName": "Régie des Transports du Territoire de Belfort",
       "id": "regiedestransportsduterritoiredebelfort-672a50",
       "locationSet": {"include": ["fx"]},
-      "matchNames": ["rttb"],
       "tags": {
         "amenity": "car_sharing",
-        "operator": "Régie des Transports du Territoire de Belfort"
+        "operator": "Régie des Transports du Territoire de Belfort",
+        "operator:short": "RTTB"
       }
     },
     {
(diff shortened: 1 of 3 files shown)


data/operators/shop/mall.json
```diff
@@ -14,6 +14,32 @@
         "operator:wikidata": "Q2287759",
         "shop": "mall"
       }
+    },
+    {
+      "displayName": "Unibail-Rodamco-Westfield",
```
1ec5, line 19: This conflicts with an existing `brand` entry for Westfield-branded malls in some of these countries:

https://github.com/osmlab/name-suggestion-index/blob/159234c5c34b3900753bd96367f48549160d09a3/data/brands/shop/mall.json#L93-L106

In iD, you can only choose a brand preset or an operator preset, not both, so when both details need to be tagged, there needs to be a single entry in NSI that represents the overlap.

Unfortunately, after 5ce3c226faf42507172e28f83e8206b6c02e4b53, this brand entry ended up conflating the operator with the brand. It should be:

```
brand=Westfield
brand:wikidata=Q106623253
operator=Unibail-Rodamco-Westfield
operator:wikidata=Q608518
```

This would be consistent with the other brand entry for Westfield-branded malls operated by Scentre Group.

Sukkoria: In next commit I moved the info about the operator Unibail-Rodamco-Westfield to the brand item, and the conflict during `npm run build` disappeared (I didn't see it the first time, sorry for the mistake)

data/operators/amenity/car_sharing.json
```diff
@@ -463,6 +463,7 @@
       "displayName": "Régie des Transports du Territoire de Belfort",
       "id": "regiedestransportsduterritoiredebelfort-672a50",
       "locationSet": {"include": ["fx"]},
+      "matchNames": ["rttb"],
```
1ec5, line 466: Should this entry have `operator:short` set to `RTTB`? If so, the `matchNames` will be redundant because NSI will get that from `operator:short`.

data/operators/amenity/car_sharing.json
```diff
@@ -479,21 +480,33 @@
...
-      "locationSet": {"include": ["de"]},
+      "displayName": "Share Now",
+      "id": "sharenow-6a8c71",
+      "locationSet": {
+        "include": [
+          "at",
+          "ca",
+          "de",
+          "dk",
+          "fx",
```
1ec5, line 500: `fx` doesn’t appear to be a valid ISO 3166-2 code. Did you mean `fr` for France?

data/operators/amenity/car_sharing.json
```diff
@@ -479,21 +480,33 @@
...
+          "it",
+          "nl",
+          "us"
```
1ec5, line 503: [The company’s official city page](https://www.share-now.com/de/en/country-list/) only lists eight European countries; it looks like they’ve pulled out of North America.

data/operators/amenity/car_sharing.json
Sukkoria: Yes, it was for France, I found some other exemples (for example in the same file "Alpes Autopartage" and "Autolib") tagged with `fx` instead of `fr`, I assumed it was similarly to the ISO-3 `FXX` code, only for the Metropolitan France (excluding overseas territories)?

bhousel: Oh yeah `fx` is allowed here, maybe even preferred.
https://ideditor.codes/?code=fx

Country Coder supports it, I guess because ISO 3166 includes it as an exception code:  
https://www.iso.org/obp/ui/#iso:code:3166:FX
https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Exceptional_reservations

The complete list of codes supported is 
https://ideditor.codes/
https://github.com/ideditor/country-coder

data/operators/amenity/car_sharing.json
Sukkoria: indeed, I did take the list from the [English Wikipedia page](https://en.wikipedia.org/wiki/Share_Now) but it seems outdated. "ca" and "us" will be removed in next commit.

data/operators/amenity/car_sharing.json
Sukkoria: moved `RTTB` to `operator:short` in next commit

Sukkoria:
hi @1ec5 and/or @bhousel, is there anything still to fix that I missed on this one?

data/operators/amenity/car_sharing.json
1ec5: Thanks, I learned something new!

1ec5 (approved):

Issue #7077, opened 2022-09-25
Duplicated Item


DiogoBaptista98:
Why have you changed the Shell Wikidata item number? Looks like someone have duplicated the item in a wrong way.

This is the original one
Shell (Q154950)
https://www.wikidata.org/wiki/Q154950

This is the duplicated one.
Shell (Q110716465)
https://www.wikidata.org/wiki/Q110716465

Please replace the item with the correct and original number.

bhousel:
This got changed in d91684d by @1ec5 , I assume for a good reason. 

It's become kind of common for fuel brands to live on as their own thing and get transferred to another company even after their original companies get dissolved - or to operate under a different "shell" companies in different parts of the world.  So , I dont know the specifics of how Shell is working around the world, but I would assume that Wikidata has it represented accurately and the QIDs are correct.

I guess relates to: #3154, #4109, #5620, #6292

1ec5:
Like OSM features, Wikidata items sometimes go through iterations of additional detail. (Micro-dataing?) In this case, Shell’s corporate restructuring [in 2005](https://www.wikidata.org/wiki/Q110716465#P127) made it necessary to distinguish the brand from the corporate parent. Because Wikidata is now making this distinction, Q154950 is only correct as an `operator` and `owner` on Shell’s relatively few owned-and-operated stores, while Q110716465 is the correct `brand` for any Shell-branded location.

Ideally, Wikidata would aim to maintain separate items for brands and their corporate owners, just as ideally it would maintain separate items for administrative areas, populated places, and their governmental authorities, but in practice these concepts get conflated until someone gets around to deconflating them. I don’t think there’s an active push to create all these separate items en masse, so the churn should be limited to Shell for now.

Issue #4109, opened 2020-09-03
Shell in Japan needs name:ja and Japanese Wikipedia tags


nicolasmaia:
name:ja=ロイヤル・ダッチ・シェル
https://ja.wikipedia.org/wiki/%E3%83%AD%E3%82%A4%E3%83%A4%E3%83%AB%E3%83%BB%E3%83%80%E3%83%83%E3%83%81%E3%83%BB%E3%82%B7%E3%82%A7%E3%83%AB

Adamant36:
> name:ja=ロイヤル・ダッチ・シェル

From my research Shell gas stations in Japan aren't branded as "Royal Dutch Shell" or "ロイヤル・ダッチ・シェル." The signs just say Shell. So, there isn't a unique Shell sub-brand specific to Japan. is the only point to this to have a link to the Japanese Wikipedia page for Shell stations in Japan or something along those lines?

nicolasmaia:
Just シェル works too!

Adamant36:
> Just シェル works 

Yeah, but why doesn't just Shell? Its what's on the signs and there's assumption of ground varifiability/"on the ground" rule with OSM. So why is the name:ja even needed? (I'm not saying it shouldn't be added. I'm just asking. There's a point where creating a seperate entry for every different language that's spoken where a global brand is located would get a little arduous 
and it could just be wrong. Especially if that langauge isn't what is on the businesses signs or how people there refer to the business).

nicolasmaia:
My main concern is that data users looking for シェル won’t find Shell. But in any case, Latin alphabet letters are foreign here and people don't use them as much as their equivalents in kana in most situations. 

Also it seems to be common practice in this project to add name:ja to brands in Japan.

Adamant36:
> Latin alphabet letters are foreign here and people don't use them as much as their equivalents in kana in most situations. Was there a specific thing that your concern is based on? 

I thought OSM based software deals with language localization if there's a whatever:ja tag present or not. 

> Also it seems to be common practice in this project to add name:ja to brands in Japan

Sure when its a Japanese brand we do. That's a completely different thing then what your requesting in this issue though. Just an FYI, there's a specifc reason I'm asking.

bhousel:
This issue was old so I just took care of it today.

[...]

Issue #7115, opened 2022-10-01
Discrepancy issue between JA:Naming sample and NSI in Japanese brands


Revtro:
Many Japanese brands in NSI are not based on [JA:Naming sample](https://wiki.openstreetmap.org/wiki/JA:Naming_sample) (especially brand tag).
Examples: AEON and its affiliates, BOOK･OFF, SEIYU, ENEOS, COCO'S, etc.

According to the mailing list of the Japanese community, brand tags are basically to use the notation of the logo. In Japan, logos are usually written in alphabetical characters, so brand tags are usually in alphabetical characters.

However, due to the establishment of NSI, most of the brand tags are now written in NSI's notation (local language, i.e. Japanese).
For POIs before the establishment of NSI, or added by mappers who edit based on the naming sample, the notation is based on the naming sample.

Therefore, I do not know whether the NSI should be modified or the Naming sample side should be adapted.

Naming sample is made among the Japanese community, so it is mostly based on a local concept (of course, there are some subtle points, since anyone can edit it).
In this respect, NSI is often seen where it does not match the local way of thinking.

LaoshuBaby:
In fact, seldom have mapper from Japan community contribute to NSI. Last time @blendy-map had done some changes and add 2 geojson, and I haven't remember other people ever raise this issue.

I had ask for help in OSM-Japan's slack channel, but looks like nobody reply to me:
https://osm-japan.slack.com/archives/C010MC9914K/p1664029723796949
![图片](https://user-images.githubusercontent.com/42690037/193400893-9ae17113-6178-424c-a471-8019485743c7.png)

In my opinion entry about local community should follow local guideline, for example, Hong Kong's entry keep bilingual because Chinese and English are both official language in HK.

And there still a lot of Japan entry haven't got a wikidata, in [`bus.json`](https://github.com/osmlab/name-suggestion-index/blob/main/data/transit/route/bus.json), in [`community_centre.json`](https://github.com/osmlab/name-suggestion-index/blob/main/data/operators/amenity/community_centre.json), in [`school.json`](https://github.com/osmlab/name-suggestion-index/blob/main/data/operators/amenity/school.json), etc..

Many current east asian entry are maintained by me, about Japan and South Korea, because NSI seldom have contributor from local community.  I know a little Japanese but isn't communicatable. So I think we need the help of Japan community several times.

And I also want to know other maintainer's opinion.

1ec5:
[...]

PR #7218, opened 2022-10-18
Fix apollostation, Shell in Japan

1ec5:
Showa Shell Sekiyu used to operate Shell-branded gas stations, but it was not itself a brand. It was however a good reason why Wikidata needs a separate item for the Shell brand versus Shell plc: Royal Dutch Shell hasn’t owned or operated any of the Shell stations in Japan for a few years now.

Anyhow, Idemitsu Kosan renamed its Showa Shell Sekiyu subsidiary to RS Energy, then shut it down, and now it’s in the process of [phasing out](https://response.jp/article/2020/11/25/340639.html) the Shell and Idemitsu brands in favor of apollostation by next year, when it loses the Shell franchise. This PR also distinguishes between apollostation and Idemitsu Kosan, which currently owns or operates three different gas station brands and many other things besides.

/ref #4109 #7077 #7216

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -360,7 +360,7 @@
         "brand": "apollostation",
         "brand:en": "apollostation",
         "brand:ja": "アポロステーション",
-        "brand:wikidata": "Q2216770",
+        "brand:wikidata": "Q114731101",
         "name": "apollostation",
         "name:en": "apollostation",
         "name:ja": "アポロステーション"
@@ -6196,6 +6196,26 @@
       }
     },
     {
+      "displayName": "シェル",
+      "id": "shell-fbe2e4",
+      "locationSet": {"include": ["jp"]},
+      "matchNames": [
+        "ロイヤル・ダッチ・シェル",
+        "昭和シェル",
+        "昭和シェル石油"
+      ],
+      "tags": {
+        "amenity": "fuel",
+        "brand": "シェル",
+        "brand:en": "Shell",
+        "brand:ja": "シェル",
+        "brand:wikidata": "Q110716465",
+        "name": "シェル",
+        "name:en": "Shell",
+        "name:ja": "シェル"
+      }
+    },
+    {
       "displayName": "中国国际能源",
       "id": "chinainternationalenergy-e4b6c0",
       "locationSet": {"include": ["cn"]},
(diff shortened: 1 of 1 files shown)


data/brands/amenity/fuel.json
```diff
@@ -6195,6 +6195,26 @@
...
+      "id": "shell-fbe2e4",
+      "locationSet": {"include": ["jp"]},
+      "matchNames": [
+        "ロイヤル・ダッチ・シェル",
+        "昭和シェル",
+        "昭和シェル石油"
+      ],
+      "tags": {
+        "amenity": "fuel",
+        "brand": "シェル",
```
1ec5, line 6209: For now, I left the `brand` and `name` in kana, but we can change it once #7115 comes to a resolution.

Revtro: I believe that the name (brand name) of Shell SS in Japan today is more recognizable as "Showa Shell(昭和シェル)" than "Shell(シェル)".

In fact, the company name at the time of the merger was "Idemitsu Showa Shell" and not "Shell".

Also, according to taginfo, "Showa Shell" is more common than "Shell", and "Showa Shell" is also used in [JA:Naming sample.](https://wiki.openstreetmap.org/wiki/JA:Naming_sample#amenity=fuel)

With the current update to apollostation in progress, I think changing the name of Showa Shell at NSI now would create confusion.

- https://taginfo.openstreetmap.jp/tags/name=昭和シェル
- https://taginfo.openstreetmap.jp/tags/name=昭和シェル石油
- https://taginfo.openstreetmap.jp/tags/name=シェル

1ec5: This was my point of reference when proposing this PR:

[](https://en.wikipedia.org/wiki/File:GasStationHiroshima.jpg)

“Showa Shell Sekiyu” appears in the same gray box as the names of other Shell franchisees and licensees in other parts of the world, so I thought this is a standard way of indicating that Showa Shell Sekiyu is the operating company’s name. However, you’re right that the stations are consistently called “Showa Shell” on maps and other publications. The naming sample’s approach of putting “Showa Shell” in `name` and “Shell” in `brand` makes sense in this case. I’ve updated this PR.

Unfortunately, some retagging will be necessary anyways: `brand:en=Showa Shell Sekiyu` was inconsistent with `brand:ja=昭和シェル` (meaning “Showa Shell”). Also, Q277115 is the item for the defunct company, so it needs to be replaced by Q110716465, which is for the present-day brand.

Revtro: At least, I agree with it. Thank you very much.

PR #7219, opened 2022-10-18
Add Forenom hostels

jduchateau:
9 locations across Finland and Sweden according to https://www.forenom.com/hostels/

They also have [aparthotels](https://www.forenom.com/aparthotels/), but I don't know in which file I should put them if they are generally tagged by `tourism=apartment` ?

### DIFF
--- a/data/brands/tourism/hostel.json
+++ b/data/brands/tourism/hostel.json
@@ -44,6 +44,16 @@
       }
     },
     {
+      "displayName": "Forenom",
+      "locationSet": {"include": ["fi", "se"]},
+      "tags": {
+        "brand": "Forenom",
+        "brand:wikidata": "Q101142662",
+        "name": "Forenom Hostel",
+        "tourism": "hostel"
+      }
+    },
+    {
       "displayName": "Girls Hostel",
       "id": "girlshostel-840e3b",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
> I don't know in which file I should put them if they are generally tagged by `tourism=apartment` ?

Sorry, NSI currently don't have catagory of `tourism=apartment`, you can make a new PR and create it if needed. The inherit from different brand need source code modify so it may need more time to solve.

I counted that there are 12 branch hotels of this Forenom hotels, looks like acceptable. Thank you for your contribution.

jduchateau:
> Sorry, NSI currently don't have catagory of `tourism=apartment`, you can make a new PR and create it if needed. The inherit from different brand need source code modify so it may need more time to solve.

Ok, I will try when I have more time.
Thanks.

LaoshuBaby:
By the way, @jduchateau 

If you want to create a new catagory you can see #5526 as a example, that PR add `amenity=library`

And I only found 3 brand for `tourism=apartment`:

https://github.com/osmlab/name-suggestion-index/blob/159234c5c34b3900753bd96367f48549160d09a3/dist/filtered/brands_discard.json#L5787-L5789

I think you can wait for more those kind of elements collected.

PR #7222, opened 2022-10-19
Move VLOTTE from operator to brand

mcliquid:
VLOTTE is only the brand for their charging stations. The operator is the "illwerke vkw AG". Check their website ([footer](https://www.vlotte.at/)):
image

### DIFF
--- a/data/brands/amenity/charging_station.json
+++ b/data/brands/amenity/charging_station.json
@@ -116,6 +116,15 @@
       }
     },
     {
+      "displayName": "VLOTTE",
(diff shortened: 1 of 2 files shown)


LaoshuBaby:
I don't think so. In [Judge Case](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case) we listed 2 opinion about how to distinguish a entity from brand and operator.

For my peosonal, If VLOTTE in every city's business are all by the "illwerke vkw AG", I'd prefer to see it as operator.

PR #7223, opened 2022-10-19
Workflow improvements

peternewman:


peternewman:
Drafting just so #7224 can be tested first...

bhousel:
> Drafting just so #7224 can be tested first...

cool!  Seeing #7234 and #7235 it looks like this worked 👍

peternewman:
Just so you're aware, Dependabot doesn't always seem to be perfect unfortunately, but probably generally useful:
https://github.com/codespell-project/actions-codespell/pull/50#issue-1386990432

PR #7228, opened 2022-10-19
Add H&M Kids

peternewman:


data/brands/shop/clothes.json
```diff
@@ -3200,6 +3200,17 @@
         "shop": "clothes"
       }
     },
+    {
+      "displayName": "H&M Kids",
+      "locationSet": {"include": ["gb"]},
+      "tags": {
+        "brand": "H&M Kids",
+        "brand:wikidata": "Q188326",
```
peternewman, line 3208: I forget, we don't need a separate QID do we?

data/brands/shop/clothes.json
```diff
@@ -3200,6 +3200,17 @@
         "shop": "clothes"
       }
     },
+    {
+      "displayName": "H&M Kids",
+      "locationSet": {"include": ["gb"]},
```
peternewman, line 3205: This is definitely in the UK. In e.g. Luxembourg, there categories seem to be Youth and Enfant, so I'm guessing they wouldn't be branded as "Kids"

peternewman: Found two more countries based on hits on OSM.

data/brands/shop/clothes.json
LaoshuBaby: Some entries are using country/region's branch company's QID, and on Wikidata, those QID have a same parent brand.
Another and more often time, we use the same QID for every country/region's entries. 
We current don't have guide rule about in which time we should use QID in which hierarchy, global one or regional one. So I think Q188326 for H&M Kids is OK.

data/brands/shop/clothes.json
LaoshuBaby: > This is definitely in the UK. In e.g. Luxembourg, there categories seem to be Youth and Enfant, so I'm guessing they wouldn't be branded as "Kids"

This project's data is collected for OSM's environment, we also follow the "on the ground" rule.

data/brands/shop/clothes.json
bhousel: Yes, this is ok, we don't strictly need a separate QID for each sub-brand.

For comparison, the Gap sub-brands all use `brand=Gap` and `brand:wikidata=Q420822`
https://nsi.guide/index.html?t=brands&k=shop&v=clothes&tt=Gap

So I think here it would be ok to use `brand=H&M` and `brand:wikidata=Q188326`.
(but `brand=H&M Kids` also seems acceptable to me)

data/brands/shop/clothes.json
peternewman: Yeah, hence I just added the ones from countries already present in OSM:
https://www.openstreetmap.org/search?query=h%26m%20kids

data/brands/shop/clothes.json
peternewman: Yeah that makes sense, I hadn't really considered the actual brand tag, now updated.

UKChris-osm:
Are there any H&M Kids stores in the UK? I thought H&M Kids was just a section in each store, which would already be catered to via the "_clothes_" tag in the general H&M store entry?

peternewman:
> Are there any H&M Kids stores in the UK? I thought H&M Kids was just a section in each store, which would already be catered to via the "_clothes_" tag in the general H&M store entry?

https://www.openstreetmap.org/way/386364611

Also 59 versus 21 here:
https://southsidewandsworth.com/sites/southside/files/styles/whats_on_scaled_body/public/images/centre-locations/floor_1.jpg

Harrow is the same as that too.

I don't know if some of them expand out after a while (and split the stores) or couldn't get one big unit initially or what...

UKChris-osm:
> https://www.openstreetmap.org/way/386364611

That location appears to have closed some time ago - [the Heart Shopping Centre](https://heartshopping.co.uk/mall-guide) - Trespass and PureGym seem to be the only occupant in that block at the moment.

I see what you mean about the two locations in London though, they are close together but seems to be separately named / branded stores.

peternewman:
> > https://www.openstreetmap.org/way/386364611
> 
> That location appears to have closed some time ago - [the Heart Shopping Centre](https://heartshopping.co.uk/mall-guide) - Trespass and PureGym seem to be the only occupant in that block at the moment.

Ah yes, looks like it all needs another #surveyme and another store just **named** vacant. 😢 

> I see what you mean about the two locations in London though, they are close together but seem to be separately named / branded stores.

Yeah, seems like quite a clever ploy TBH. I've seen it once personally in the flesh anyway hence this. I'm assuming they don't want people moving between the two stores on either side of the shopping centre without paying for goods first, hence they should be two shops!

UKChris-osm:
Yes I definitely can't imagine they would allow a "_I'll pay for these jeans in H&M Kids_" policy. I wonder if it's just cheaper / easier for them to rent a second smaller store in a retail centre rather than moving an entire store to a bigger unit.

peternewman:
> I wonder if it's just cheaper / easier for them to rent a second smaller store in a retail centre rather than moving an entire store to a bigger unit.

Yeah you wouldn't need to refit the bigger unit, or knock through a wall, or close the shop/dual run while moving. You could also test the water a bit with one store initially and split stuff off if sales are good etc.

PR #7230, opened 2022-10-20
Update bus.json

Abdelkrim2003:
add wikidata busnetwork casablanca and add entrie busnetwork Meknes morocco

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -2460,11 +2460,12 @@
       }
     },
     {
-      "displayName": "Casablanca Bus",
+      "displayName": "Casabus",
       "id": "casablancabus-1ba109",
       "locationSet": {"include": ["ma"]},
       "tags": {
-        "network": "Casablanca Bus",
+        "network": "Casabus",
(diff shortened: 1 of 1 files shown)


data/transit/route/bus.json
```diff
@@ -2460,11 +2460,12 @@
       }
     },
     {
-      "displayName": "Casablanca Bus",
+      "displayName": "Casabus",
       "id": "casablancabus-1ba109",
       "locationSet": {"include": ["ma"]},
```
LaoshuBaby, line 2465:
```suggestion
      "locationSet": {"include": ["ma"]},
      "matchNames": ["Casablanca Bus"],
```
If this is a name change by business merge or reorganization, usually there still have some element in OSM using old name, and formatter program will gen a empty "Casablanca Bus" entry when run it.
So I suggest you add it as matchName.

Abdelkrim2003:
it is the same network only the operator has changed

bhousel:
Thanks @Abdelkrim2003 👍

PR #7231, opened 2022-10-20
Add shops

Mannivu:
Adding forniture, clothes and shoes shop chains

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -9010,19 +9010,6 @@
         "shop": "clothes"
       }
     },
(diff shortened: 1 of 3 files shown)


LaoshuBaby:
Nuna Lie already in https://github.com/osmlab/name-suggestion-index/blob/89b44dab6e19291ebe4281367201f6e4695dbb71/data/brands/shop/clothes.json#L5515-L5523
So I remove duplicated entry. (Also Naheva)

BTW, you don't need use your editor's formatter process this file, and we don't need the exactly same matchName as brand/name

I'll fix other remaining syntax error.

PR #7233, opened 2022-10-21
add missing cuisine tags

Sukkoria:
to four restaurant or fast_food places for now (after checking their menus on their websites)


bhousel:
Thanks @Sukkoria 👍

PR #7251, opened 2022-10-23
Add Japanese computer store, change 100 Man Volt and Kojima to one that seems appropriate

Revtro:
As for 100 Man Volt, "100mv" is an abbreviation for a website or social networking account, and the English translation of the original name "100 Man Volt" would be more appropriate.

As for Kojima, it is now part of Bic Camera, and most "Kojima" brand stores are now branded "Kojima x Bic Camera", with only one "Kojima" brand store remaining.
However, I am not sure if the original "Kojima" should be removed, so I am leaving it as it is.

### DIFF
--- a/data/brands/shop/computer.json
+++ b/data/brands/shop/computer.json
@@ -201,6 +201,7 @@
     },
     {
       "displayName": "ドスパラ",
+      "id": "dospara-8cc09d",
       "locationSet": {"include": ["jp"]},
       "tags": {
         "brand": "ドスパラ",
@@ -215,6 +216,7 @@
     },
     {
       "displayName": "パソコン工房",
+      "id": "6185e9-8cc09d",
       "locationSet": {"include": ["jp"]},
       "tags": {
         "brand": "パソコン工房",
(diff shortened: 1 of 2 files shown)


data/brands/shop/computer.json
```diff
@@ -198,6 +198,34 @@
...
+        "shop": "computer"
+      }
+    },
+    {
+      "displayName": "パソコン工房",
+      "id": "6185e9-8cc09d",
+      "locationSet": {"include": ["jp"]},
+      "tags": {
+        "brand": "パソコン工房",
+        "brand:ja": "パソコン工房",
```
LaoshuBaby, line 223: With brand:en/name:en=UNITCOM?

LaoshuBaby:
> As for Kojima, it is now part of Bic Camera

If "Bic Camera" bought this brand "Kojima", you can add "Kojima" as its matchNames, then delete old one.

Also, "100mv" can be matchNames of "100 Man Volt"

data/brands/shop/computer.json
Revtro: UNITCOM is the name of the company, and since I could not find an official English name for "パソコン工房( In Romaji: Pasokon Kōbō )", I decided to leave the English name out for now.

Revtro:
Since Kojima and its brand have not completely disappeared, but only the brand of most of the stores that were "Kojima" has changed, it is not clear whether they should be removed.

Also, thank you for the advice. I tried adding matchNames.

PR #7253, opened 2022-10-23
Added machi machi

le-jun:


data/brands/shop/beverages.json
```diff
@@ -188,6 +188,16 @@
...
       }
     },
+    {
+      "displayName": "Machi machi",
+      "tags": {
+        "brand": "麥吉 machi machi",
+        "brand:wikidata": "Q114828290",
+        "cuisine": "bubble_tea",
+        "name": "Machi machi",
+        "shop": "beverages"
```
LaoshuBaby, line 198: https://github.com/openstreetmap/id-tagging-schema/blob/7617c45325036f33190895c02529511f61eebdac/data/presets/amenity/cafe/bubble_tea.json#L7-L14

Looks like bubble tea shop usually tagged as cafe

data/brands/shop/beverages.json
```diff
@@ -188,6 +188,16 @@
         "shop": "beverages"
       }
     },
+    {
+      "displayName": "Machi machi",
+      "tags": {
+        "brand": "麥吉 machi machi",
```
LaoshuBaby, line 194:
```suggestion
        "brand": "machi machi",
        "brand:en": "machi machi",
        "brand:zh": "麥吉",
```
I looked their website, only 2 shop in Taiwan, most of their branch shop are all over the world, so I think this is a 001 entry, so use English as brand only.

Besides, according to photo on its website, they only use "machi machi"

LaoshuBaby (changes requested):

data/brands/shop/beverages.json
le-jun: Wasn’t sure about that part, do move it to the category more suitable to you

Issue #6481, opened 2022-04-21
Remove the *:wikipedia tags


bhousel:
We often get feedback that tags like `brand:wikipedia` are redundant and we could probably stop adding it.   This issue is just for me to remove the `*:wikipedia` tags from the index and also from all the `build_wikidata` script that gathers them.

_Followup on comment posted by @bhousel in https://github.com/osmlab/name-suggestion-index/issues/5275#issuecomment-905798587_

bhousel:
This is done! 
As of today's release v6.0.20220421 these tag updates won't be suggested anymore.

Lukas458:
Okay well, thank you, but I only hope that's the consensus here that these data are kind of redundant, so noone will come up with the idea to re-add them.

PR #7254, opened 2022-10-23
Add "brand:wikipedia=en:Burger King" for "Burger King"

TheSnoozer:
Automatically add 'brand:wikipedia=en:Burger King' for 'brand:wikidada=Q177054' (refs https://maproulette.org/challenge/12036)

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -766,6 +766,7 @@
         "amenity": "fast_food",
         "brand": "Burger King",
         "brand:wikidata": "Q177054",
+        "brand:wikipedia": "en:Burger King",
         "cuisine": "burger",
         "name": "Burger King",
         "takeaway": "yes"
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Sorry, I'll reject this PR.

① You can see https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case, according to discussion, add wikipedia tag will bring more quarrel for NSI (for example, which language's wikipedia should be used), also this was reported many times in iD's repository. After https://github.com/osmlab/name-suggestion-index/issues/6481, `*:wikipedia` was removed.

② Even before this removal, we will use localized wikipedia version for different specific region. 
e.g.: "zh:汉堡王" for "cn", "ja:バーガーキング" for "jp"

TheSnoozer:
Ok, thanks for your quick feedback! And no worries, feel free to reject if it doesn't belong here

PR #7259, opened 2022-10-23
Add Warsaw public_transport_tickets vending machine operators

docentYT:
[Source](https://wiki.openstreetmap.org/wiki/Pl:Tag:vending%3Dpublic_transport_tickets) (`ZTM Warszawa` header)
This is how all vending machines in the Polish capital are tagged.


data/operators/amenity/vending_machine.json
```diff
@@ -29,6 +29,67 @@
...
+      "locationSet": {"include": ["pl"]},
+      "tags": {
+        "amenity": "vending_machine",
+        "vending": "public_transport_tickets",
+        "operator": "Mera-Serwis SGL",
+        "network": "network=ZTM Warszawa",
+        "network:wikidata": "Q8066998",
+        "network:wikipedia": "pl:Zarząd Transportu Miejskiego w Warszawie",
+        "currency:PLN": "yes",
+        "payment:coins": "yes",
```
LaoshuBaby, line 44: I'm troubled with payment related tag

See https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case

Some times we allow the addition of payment, but more time we reject it. @bhousel 

---
In my opinion, NSI shouldn't be a treasure chest with every kind of detailed tags, those should be added after on ground survey. But if this entry belongs to a public transport, we have enough reason believe that every single machine support same payment.

（> <）！

arch0345 (changes requested):
Please note that we no longer include `:wikipedia` tags in NSI (see https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#can-we-add-websitephonewikipedia)

data/operators/amenity/vending_machine.json
docentYT: In Warsaw, single-operator vending machines accept the same forms of payment.

data/operators/amenity/vending_machine.json
```diff
@@ -29,6 +29,64 @@
...
       }
+    },
+    {
+      "displayName": "Mera-Serwis SGL (ZTM Warszawa)",
+      "locationSet": {"include": ["pl"]},
+      "tags": {
+        "amenity": "vending_machine",
+        "vending": "public_transport_tickets",
+        "operator": "Mera-Serwis SGL",
+        "network": "network=ZTM Warszawa",
```
bhousel, line 40: This looks like a typo?
```suggestion
        "network": "ZTM Warszawa",
```

docentYT: Yes, thank you!

LaoshuBaby:
Wait, I just found another problem……

@docentYT You add `network:wikidata` for "ZTM Warszawa" in 3 entries, but they are in `data/operator/amenity` tree, so if you want them be suggested, you'd better complete `operator:wikidata` for them.

docentYT:
@LaoshuBaby but `network:wikidata` for ZTM Warszawa is the correct tagging for Warsaw's ticket machines. The company operating the machines is listed as an operator.

LaoshuBaby:
> @LaoshuBaby but `network:wikidata` for ZTM Warszawa is the correct tagging for Warsaw's ticket machines. The company operating the machines is listed as an operator.

I don't means change it or remove it, but add wikidata as operator at the same time. Like this:

```
brand=1
brand:wikidata=Q12345678
network=ZTM Warszawa
network:wikidata=Q87654321
```

Do you think this is accaptable or this way can tag those vending machine correctly?

bhousel:
Yeah these vending machines can be in the `data/operators/amenity/vending_machine.json` file, but they really need the `operator`/`operator:wikidata` tags existing, otherwise nothing will happen.

The `network`/`network:wikidata` tag is fine - when in the operators tree these are ignored (treated like any other random tags).

docentYT:
@LaoshuBaby @bhousel I've just added operator:wikidata.

arch0345 (approved):

PR #7271, opened 2022-10-26
Add french driving_school ECF

bmillemathias:


LaoshuBaby:
You can see #5526, it also a catagory creating PR.
For example, you can move https://github.com/osmlab/name-suggestion-index/blob/b66c59beb7d32aa31a5d137de95f2c4128d360ec/dist/filtered/brands_discard.json#L726-L729
to
https://github.com/osmlab/name-suggestion-index/blob/main/dist/filtered/brands_keep.json

PR #7277, opened 2022-10-28
Add Pharmacy Tsuruha (調剤薬局ツルハドラッグ)

Revtro:
I added the dispensary portion because it seems to be treated as a separate store brand.
<hr>
Not directly related to this PR, but how can I prevent dispensing=yes from being automatically added in the amenity=pharmacy and healthcare=pharmacy presets?<br>
In Japan, drug stores are added with this preset (tags), but not all stores have dispensing available, like the brand I have just added.

I didn't see it in the NSI json, so is it necessary to change the code of the iD preset?

### DIFF
--- a/data/brands/amenity/pharmacy.json
+++ b/data/brands/amenity/pharmacy.json
@@ -4187,6 +4187,22 @@
       }
     },
     {
+      "displayName": "調剤薬局ツルハドラッグ",
+      "id": "pharmacytsuruha-650eee",
+      "locationSet": {"include": ["jp"]},
+      "tags": {
+        "amenity": "pharmacy",
+        "brand": "ツルハドラッグ",
+        "brand:en": "Tsuruha",
+        "brand:ja": "ツルハドラッグ",
+        "brand:wikidata": "Q114891238",
+        "healthcare": "pharmacy",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
> Not directly related to this PR, but how can I prevent dispensing=yes from being automatically added in the amenity=pharmacy and healthcare=pharmacy presets?
> In Japan, drug stores are added with this preset (tags), but not all stores have dispensing available, like the brand I have just added.
> 
> I didn't see it in the NSI json, so is it necessary to change the code of the iD preset?

Looks like this can't be solved in NSI because this was pharmacy's preset, stored in there:
https://github.com/openstreetmap/id-tagging-schema/blob/abe59b253d30cb02ef1d491fe937fbafdcd746fb/data/presets/amenity/pharmacy.json#L3-L11
I suggest that you open a issue in that repository.

Revtro:
> Looks like this can't be solved in NSI because this was pharmacy's preset, stored in there: https://github.com/openstreetmap/id-tagging-schema/blob/abe59b253d30cb02ef1d491fe937fbafdcd746fb/data/presets/amenity/pharmacy.json#L3-L11 I suggest that you open a issue in that repository.

Thank you very much, I will try to open an issue in that repository.

PR #7282, opened 2022-10-29
add more Dutch Brands

wvanderp:
This PR adds more Dutch brands.

Let me know if there are any issues, and I'll fix them

### DIFF
--- a/data/brands/shop/hairdresser.json
+++ b/data/brands/shop/hairdresser.json
@@ -74,8 +74,6 @@
       "tags": {
         "brand": "BrainWash",
         "brand:wikidata": "Q114905133",
-        "female": "yes",
-        "male": "yes",
         "name": "BrainWash",
         "shop": "hairdresser"
       }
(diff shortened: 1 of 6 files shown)


data/brands/shop/hairdresser.json
```diff
@@ -66,6 +66,20 @@
...
     },
+    {
+      "displayName": "BrainWash",
+      "id": "brainwash-7f2667",
+      "locationSet": {"include": ["nl"]},
+      "preserveTags": ["^name"],
+      "tags": {
+        "brand": "BrainWash",
+        "brand:wikidata": "Q114905133",
+        "female": "yes",
```
LaoshuBaby, line 77: Maybe those 2 gender related tag don't need to add, as many hairdresser don't restrict customer's gender.

For example, if this is a male only barbershop, we can add a `male=only` because this tag is difinitely needed, like we treat `vegan=only` and `vegan=yes`.

PR #7283, opened 2022-10-29
Fix GitHub Chalk

peternewman:
See:
https://github.com/osmlab/name-suggestion-index/actions/runs/3351912890/jobs/5553683840

Compared to:
https://github.com/osmlab/name-suggestion-index/actions/runs/3351865802/jobs/5553601255
https://github.com/osmlab/name-suggestion-index/actions/runs/3351865802/jobs/5553601300

`zh`(langcode) rather than `cn`(country/region code)

--
Related: #7226

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,4 +27,8 @@ jobs:
       with:
         node-version: ${{ matrix.node-version }}
     - run: npm install
+      env:
+        FORCE_COLOR: 2
     - run: npm run test
+      env:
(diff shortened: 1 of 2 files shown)


bhousel:
Looks good! This is what was expected right?

Screen Shot 2022-10-29 at 12 38 11 PM

peternewman:
> Looks good! This is what was expected right?

I assume so :shrug: It's definitely more colourful than it used to be without the change! Seems to tie up with what the code is trying to do too...

PR #7287, opened 2022-10-30
Add Sonance brand

bmillemathias:
over-pass link https://overpass-turbo.eu/s/1ndj
I did not put a name tag on purpose because Sonance is a netword of independant shops and I'm not sure all shops are named Sonance

### DIFF
--- a/data/brands/shop/hearing_aids.json
+++ b/data/brands/shop/hearing_aids.json
@@ -181,6 +181,17 @@
       }
     },
     {
+      "displayName": "Sonance Audition",
+      "id": "sonanceaudition-28a4ee",
+      "locationSet": {"include": ["fr"]},
+      "tags": {
+        "brand": "Sonance Audition",
+        "brand:wikidata": "Q114932560",
+        "network": "Sonance Audition",
+        "shop": "hearing_aids"
+      }
+    },
+    {
       "displayName": "Van Boxtel hoorwinkel",
       "id": "vanboxtelhoorwinkel-6f1b2d",
       "locationSet": {"include": ["001"]},
--- a/data/brands/shop/hearing_aids.json
+++ b/data/brands/shop/hearing_aids.json
@@ -183,11 +183,12 @@
     {
       "displayName": "Sonance Audition",
       "id": "sonanceaudition-28a4ee",
+      
       "locationSet": {"include": ["fr"]},
+      "preserveTags": ["^name"],
       "tags": {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
I'm not sure if using the network tag here would be needed as the brand tag would be the main tag here, although from the Overpass link I do see network is used a lot.

If you're unsure about the name, you can add "preserveTags["^name"]; under the locationSet and this will only add a name if there isn't one already assigned (like a default). This means if someone has already added a name, and just wants to add the brand data, the name will not be overwritten.

data/brands/shop/hearing_aids.json
```diff
@@ -180,6 +180,17 @@
...
       }
     },
+    {
+      "displayName": "Sonance Audition",
+      "id": "sonanceaudition-28a4ee",
+      "locationSet": {"include": ["fr"]},
+      "tags": {
+        "brand": "Sonance Audition",
+        "brand:wikidata": "Q114932560",
+        "network": "Sonance Audition",
```
LaoshuBaby, line 190:
```suggestion
        "name": "Sonance Audition",
```

LaoshuBaby (changes requested):

data/brands/shop/hearing_aids.json
bmillemathias: as I said not all shops have the same namethat why I used network key

bmillemathias:
> I'm not sure if using the network tag here would be needed as the brand tag would be the main tag here, although from the Overpass link I do see network is used a lot.

THis is because of me, I updated the shop before submitting the MR

> 
> If you're unsure about the name, you can add "preserveTags["^name"]; under the locationSet and this will only add a name if there isn't one already assigned (like a default). This means if someone has already added a name, and just wants to add the brand data, the name will not be overwritten.

Thanks for the suggestion, I´ll do that

UKChris-osm:
What I would recommend in this PR is adding the name tag as well, as this will act as the default for any new addition to OSM. But, by having `preserveTags["^name"]` you are protecting the name if it's been changed, so iD / NSI won't attempt to overwrite it.

bhousel:
Thanks @bmillemathias  👍 

> What I would recommend in this PR is adding the name tag as well, as this will act as the default for any new addition to OSM.

This is a good suggestion, I added this as a followup commit in 2b55955

bmillemathias:
Thanks !

PR #7289, opened 2022-10-30
Add da emobil (operator for charging_station)

mcliquid:
Currently more than 500 stations.
See https://www.da-emobil.com/

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -790,6 +790,7 @@
     },
     {
       "displayName": "da emobil GmbH & Co KG",
+      "locationSet": {"include": ["de, at, ch, fr, es, hu, dk, it"]},
       "tags": {
         "amenity": "charging_station",
         "operator": "da emobil GmbH & Co KG"
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -790,7 +790,7 @@
     },
     {
       "displayName": "da emobil GmbH & Co KG",
-      "locationSet": {"include": ["at"]},
+      "locationSet": {"include": ["at", "de", "ch", "gb", "ie", "it", "fr", "hu", "dk"]},
       "tags": {
         "amenity": "charging_station",
         "operator": "da emobil GmbH & Co KG"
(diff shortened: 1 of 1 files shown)


data/operators/amenity/charging_station.json
```diff
@@ -788,6 +788,14 @@
         "operator": "Daimler AG"
       }
     },
+    {
+      "displayName": "da emobil GmbH & Co KG",
```
bmillemathias, line 792: There should be a capital letter of firast letter of words so "Da Emobil ..."

data/operators/amenity/charging_station.json
```diff
@@ -788,6 +788,14 @@
...
+      "locationSet": {"include": ["at", "de", "ch", "gb", "ie", "it", "fr", "hu", "dk"]},
+      "tags": {
+        "amenity": "charging_station",
+        "operator": "da emobil GmbH & Co KG"
```
bmillemathias, line 796: same here

bmillemathias (changes requested):

data/operators/amenity/charging_station.json
bhousel: Here the display name should just be "da emobil".  This is the name that you'd see in the JOSM menu, or in the iD preset, not the legal name.

bmillemathias:
if itś done on purpose so I think that's ok

data/operators/amenity/charging_station.json
bhousel: Here you can use whatever `operator` value the local community prefers.  Most local communities don't like to see the "GmBH & Co KG" stuff, but maybe yours does?

The important thing for this to be converted to a preset is to add a `operator:wikidata` tag  (which this PR doesn't have).

LaoshuBaby:
> if itś done on purpose so I think that's ok

Another thing is data in `tags` will be added to OSM element, so they must fit `on the ground` rule, if it was called `da emobil` more offen, you can just it as operator.

For example, in China we only need a `operator=星巴克` for starbucks, not need `operator=星巴克企业管理（中国）有限公司`, that is "Starbucks ~~Enterprise Management (China) Co., Ltd.~~ "

Many country's company will have special surfix, like "Kabushiki gaisha"(株式会社) in Japan, but you haven't see local community add it as operator because its official name, not most common used name. The uppercase problem is also like this, if you seldom see a uppercased name, just leave it as is.

Of course, I don't live in Germany, so use which name should be chosen by local community.

arch0345 (approved):
Thanks @mcliquid!

PR #7300, opened 2022-10-31
some more cleanup for Dutch brands

wvanderp:
Added BESTSELLER stores
add fashion back into Takko
preserved name for Stella
and add extended names of pharmacies

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -5367,7 +5367,7 @@
     {
       "displayName": "NAME IT",
       "id": "nameit-3937bd",
-      "locationSet": {"include": ["001"]},
+      "locationSet": {"include": ["eu"]},
       "tags": {
         "brand": "NAME IT",
         "brand:wikidata": "Q12328259",
@@ -5744,7 +5744,7 @@
     {
       "displayName": "ONLY & SONS",
       "id": "onlyandsons-3937bd",
-      "locationSet": {"include": ["001"]},
+      "locationSet": {"include": ["eu"]},
       "tags": {
         "brand": "ONLY & SONS",
         "brand:wikidata": "Q114965261",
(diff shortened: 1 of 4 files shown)


data/brands/shop/clothes.json
```diff
@@ -5362,6 +5364,18 @@
         "shop": "clothes"
       }
     },
+    {
+      "displayName": "NAME IT",
+      "id": "nameit-3937bd",
+      "locationSet": {"include": ["001"]},
```
LaoshuBaby, line 5370: 001 means this brand is suitable for every countries and region, it will be a global entry. For Japan, Africa, China, and many other places. If this brand only appear in Germany, you can use `de`, or if you can see it in many Europe countries, `eu` is also acceptable.

This comment also apply to other 001 entry added in this PR.

data/brands/shop/clothes.json
```diff
@@ -7276,23 +7315,33 @@
       }
     },
     {
-      "displayName": "Takko",
-      "id": "takko-475eda",
+      "displayName": "Takko Fashion",
```
LaoshuBaby, line 7318: I search taginfo and found there are more than 1k elements named with `Takko`, is this brand changed their name? If so, keep a matchNames is needed. Or many existing elements won't be matched.

wvanderp: I made this change because the logo, storefronts and websites all talk about Takko Fashion with "fashion" part of the brand

https://company.takko.com/en-gb/newsroom/press-releases.html

So I think we should include it in the name. I though the match on qid would fix the name. 
but if there are any further changes needed the please ask

data/brands/shop/clothes.json
wvanderp: applied a fix

PR #7305, opened 2022-11-01
Correct Intermarché wikidata and add Intermarché Hyper

Cj-Malone:
Use specific QIDs since they are available.

Looking at [Wikipedia](https://en.wikipedia.org/wiki/Intermarch%C3%A9) Intermarché Express may be closer to `shop=convenience` than `shop=supermarket`, but then again, maybe it'd have been fixed by now if that was true.

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -1682,7 +1682,7 @@
       "matchTags": ["shop/supermarket"],
       "tags": {
         "brand": "Intermarché Contact",
-        "brand:wikidata": "Q3153200",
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks @Cj-Malone  this seems fine 👍 
I agree the "Express" brands are probably convenience stores, but this probably doesn't matter much.

nuxper:
Hi. 
Thanks for the modifications.
Note that Intermarché express is a sub brand of Intermarché for branches inside cities. But they are still pretty large with 700m2 and you can use a trolley inside. Therefore shop=supermarket seems accurate.

PR #7306, opened 2022-11-01
Update "& Other Stories" location

Crashillo:
This brand has stores all around the world, not just England. https://www.stories.com/en_eur/storelocator.html


bhousel:
Thanks @Crashillo 👍

PR #7312, opened 2022-11-02
Add VVV / Verkehrsverbund Vorarlberg as operator for bicycle_parkings

mcliquid:
@bhousel Is operator:wikipedia a best practice or not welcome here?

And the VVV have also a "brand" for bicycle_parkings with the name "[VMOBIL](https://radbox.vmobil.at/)". But there isn't a existing file for bicycle_parkings in /brands - Should I add one?


data/operators/amenity/bicycle_parking.json
```diff
@@ -34,6 +34,21 @@
...
+        "include": ["at"]
+      },
+      "matchNames": ["cleVVVer", "vmobil"],
+      "tags": {
+        "amenity": "bicycle_parking",
+        "name": "Verkehrsverbund Vorarlberg",
+        "operator": "Verkehrsverbund Vorarlberg GmbH",
+        "operator:short": "VVV",
+        "operator:wikidata": "Q2516495",
+        "operator:wikipedia": "de:Verkehrsverbund_Vorarlberg"
```
LaoshuBaby, line 49: > @bhousel Is operator:wikipedia a best practice or not welcome here?

In #6481 all wikipedia tags removed, see more in https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case

LaoshuBaby (changes requested):

LaoshuBaby:
> And the VVV have also a "brand" for bicycle_parkings with the name "[VMOBIL](https://radbox.vmobil.at/)". But there isn't a existing file for bicycle_parkings in /brands - Should I add one?

I can't search VMOBIL in current OSM data, but I noticed that 6 bicycle parking brand was collected and discarded:

https://github.com/osmlab/name-suggestion-index/blob/e0d5bd30f99796e2aea28b1bb6e2cdbb08e2958e/dist/filtered/brands_discard.json#L318-L323

If you want to add this catagory, you can see #5526 and follow this example.

Issue #7314, opened 2022-11-02
Municipal additions that just add operator?


EvanCarroll:
I live in a rather mediocre city (Houston).

Would it be OK to send it patches for

```
City of Houston - Park
City of Houston - Library
Houston ISD - School Building
```

That automatically fill out the operator field and name wiki data information? We have probably a [couple hundred parks](http://www.houstontx.gov/parks/pdfs/HPARDInventory.pdf), a few hundred schools and probably 20-50 libraries. I'm pretty sure I've seen chains of this size. Moreover, we do have METRO already in the database, which is our municipal bus program.

PR #7315, opened 2022-11-02
add City of Houston park name

EvanCarroll:
First part of #7314


EvanCarroll:
Actually, this should be `access:public` too since all of the parks managed by the city are default `access:public`

PR #7316, opened 2022-11-02
add City of Houston park name

EvanCarroll:
First part of #7314

### DIFF
--- a/data/operators/leisure/park.json
+++ b/data/operators/leisure/park.json
@@ -1396,6 +1396,23 @@
       }
     },
     {
+      "displayName": "City of Houston",
+      "id": "houstonparksandrecreationdepartment-8bfd34",
+      "locationSet": {
+        "include": ["us-tx.geojson"]
+      },
+      "tags": {
+        "addr:state": "TX",
+        "leisure": "park",
+        "operator": "Houston Parks and Recreation Department",
+        "operator:short": "HPARD",
+        "operator:type": "government",
+        "operator:wikidata": "Q115018877",
+        "owner": "City of Houston",
+        "owner:wikidata": "Q16555"
+      }
+    },
+    {
       "displayName": "City of Idaho Falls",
       "id": "cityofidahofalls-e3f3ca",
       "locationSet": {
@@ -2667,7 +2684,7 @@
       "locationSet": {"include": ["in"]},
       "tags": {
         "leisure": "park",
-        "operator": "CCuttack Development Authority",
+        "operator": "Cuttack Development Authority",
         "operator:short": "CDA",
         "operator:wikidata": "Q5196919"
       }


EvanCarroll:
ok I'm done with this. I've attached two patches separate. One adds what I care about one fixes a problem with the pre-existing convention that I followed.

bhousel:
> Previously there was a convention to use operator:short for
> abbreviations. This was fishy, and it turns out not supported per the
> wiki. This should be operator:abbr. This patch fixes this everywhere for
> parks.

Interesting, that's the first I've heard of it.  We use `operator:short` everywhere else in NSI, but this was pretty well established by the rail and transport mappers for a long time.

EvanCarroll:
@bhousel I can role that back if you want to wait on that. But `:abbr` should never be spoken, while `:short` should always be spoken. So the the distinction is pretty substantial (for those with screen readers).

Should I handle that in another issue? I'd like to get this approved in more then that fix. Though it is a lot of auditing, though vim makes it pretty easy.

bhousel:
> @bhousel I can role that back if you want to wait on that. But `:abbr` should never be spoken, while `:short` should always be spoken. So the the distinction is pretty substantial (for those with screen readers).

Is this guidance written down anywhere?
This is the first time I'm seeing an `operator:abbr` tag anywhere, so I'm very hesitant to start changing stuff.. 
 
It was only added to this page last year, as far as I can tell.
https://wiki.openstreetmap.org/wiki/Key:operator#Further_details

EvanCarroll:
That's the same link I would quote, one is documented as

> Abbreviated name of the operator

The other as,

> Short name of the operator

I think it's possible to have a short name that is not an abbreviation, as it pertains to this patch, you could say "Houston" is always short for the legal name of the operator, "City of Houston". And likewise, "Harris County" is typically shortened to "County". However, HPARD would mean almost nothing to 99% of Houstonians not interested in Parks. Even though it's an official acronym and [in widespread use](https://twitter.com/HPARD) it's not a short_name and no one knows it? But maybe this is all a stretch? Not sure, not particularly passionate about it either.

data/operators/leisure/park.json
```diff
@@ -1395,6 +1395,25 @@
...
+      "id": "houstonparksandrecreationdepartment-8bfd34",
+      "locationSet": {
+        "include": ["us-tx.geojson"]
+      },
+      "tags": {
+        "access": "public",
+        "addr:state": "TX",
+        "leisure": "park",
+        "operator": "Houston Parks and Recreation Department",
+        "operator:phone": "+1-832-395-7000",
```
arch0345, line 1409:
```suggestion
```
Please note that we don't include `:phone` or other contact tags in NSI. Instead, we encourage people to add this information to their respective Wikidata item. (see https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#can-we-add-websitephonewikipedia)

data/operators/leisure/park.json
```diff
@@ -1395,6 +1395,25 @@
...
       }
     },
+    {
+      "displayName": "City of Houston",
+      "id": "houstonparksandrecreationdepartment-8bfd34",
+      "locationSet": {
+        "include": ["us-tx.geojson"]
+      },
+      "tags": {
+        "access": "public",
```
arch0345, line 1405: According to the wiki, `access=public` has been deprecated (see https://wiki.openstreetmap.org/wiki/Tag:access=public)

EvanCarroll: So should we use `access=yes` instead to designate that the park is public?

data/operators/leisure/park.json
EvanCarroll: That's an interesting design decision. So we have to add the data to Wikidata, which the search/index/retrieval functionality does not support and that effectively locks the data in Wikidata in perpetuity because of licensing mismatch? I mean it would make sense to have something like Wikidata which OSM could store this data about discrete ways/pois in a centralized location. But it doesn't make much sense when the suggestion is to put the data somewhere none of the tooling currently looks (server or client) and where it can't be imported in the future.

For example, is there a plan to support search by phone-by-number in OSM? That's incredibly useful for linking a phone book to GIS.

arch0345: > that effectively locks the data in Wikidata in perpetuity because of licensing mismatch?

[Wikidata is licensed under CC0 (public domain](https://www.wikidata.org/wiki/Wikidata:Licensing)), so I don't think licensing will be an issue.

> But it doesn't make much sense when the suggestion is to put the data somewhere none of the tooling currently looks (server or client)

That's a good point, that was part of the reason why I opened #5275. This decision was mostly based on https://github.com/osmlab/name-suggestion-index/issues/5275#issuecomment-905798587

1ec5: NSI is designed to help make OSM tagging consistent for the things that mappers would be tagging anyways. It isn’t designed to be a general-purpose vehicle for information tangential to a feature on the ground.

> But it doesn't make much sense when the suggestion is to put the data somewhere none of the tooling currently looks (server or client) and where it can't be imported in the future.

Many OSM data consumers, [including Nominatim](https://nominatim.org/release-docs/3.5.1/data-sources/Wikipedia-Wikidata/), already cross-reference Wikidata for various things, just not phone numbers specifically.

> For example, is there a plan to support search by phone-by-number in OSM? That's incredibly useful for linking a phone book to GIS.

Nominatim has declined to add support for searching by phone number, let alone searching by the operator’s phone number: osm-search/Nominatim#861. I don’t think it has been proposed for Pelias so far. In any event, a query tool would be a more appropriate tool for any use case requiring an exact string match to a unique identifier. Indeed, the Overpass API already supports querying for features by `phone` or any other key.

1ec5:
> Is this guidance written down anywhere?
This is the first time I'm seeing an `operator:abbr` tag anywhere, so I'm very hesitant to start changing stuff.. 

About 80% of the occurrences come from [an import in Latvia](https://wiki.openstreetmap.org/wiki/LV:LVM-import) that took place after this undiscussed addition to the `operator` page. All I could find about the key is a few brief mentions in the old German and Polish forums, including [some criticism](https://forum.openstreetmap.org/viewtopic.php?id=55248). These discussions didn’t present `operator:abbr` and `operator:short` as complementary as described above. This doesn’t seem like a solid enough foundation for NSI to adopt an alternative to a much more established key.

data/operators/leisure/park.json
EvanCarroll: @arch0345 Just to follow up on the licensing question,

> Wikidata is using the CC0 license, which on first sight allows imports into OSM. However, OSM does not, generally, accept imports of data from Wikidata due to the sources often used to contribute to Wikidata. [...] OpenStreetMap tries to be whiter than white, and data should be legally used in any country[[4]](https://wiki.openstreetmap.org/wiki/Wikidata#cite_note-fairhurst-4). Therefore OpenStreetMap acknowledges [sui generis database right](https://en.wikipedia.org/wiki/en:Sui_generis_database_right). As a conclusion, an import of information from Wikidata into OSM **is generally not permitted because Wikidata has lower standards in terms of copyright than OSM**. In addition, Wikidata contains some data from sources we are not allowed to use[[5]](https://wiki.openstreetmap.org/wiki/Wikidata#cite_note-koppenhoefer-5).

bhousel:
> I think I've resolved all the mentioned changes and I dropped the operator:abbr.

Thanks 👍

PR #7322, opened 2022-11-04
Add bus stops operator in Japan

Revtro:
I am not sure this is correct as there were not enough examples.

The additions are as follows:
- ジェイ・アール北海道バス (JR Hokkaido Bus)
- 北海道中央バス (Hokkaido Chuo Bus)
- じょうてつ (Jotetsu)
- 北都交通 (Hokuto Kotsu)


arch0345 (changes requested):
Please note that all entries in the `transit` branch require a `network` tag. Also, bus networks/operators usually get added to `route/bus`, which will automatically create an entry in `highway/bus_stop` because of this [template](https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#templates):
https://github.com/osmlab/name-suggestion-index/blob/0c90fe9a8898d7c43bce87c6a9b85dea30d992cf/data/transit/highway/bus_stop.json#L10-L18

Revtro:
In Japan (at least in the area of the bus stops we added this time), the network tag is rarely used for transportation, and the operator tag is often used instead.
However, due to the terrible errors in the notation, I was wondering if there is any way to add this to the NSI.

Revtro:
If no network tag is not completely allowed in transit, is it possible to create and use operator/highway/bus_stop.json?

arch0345:
> If no network tag is not completely allowed in transit, is it possible to create and use operator/highway/bus_stop.json?

Yes, that would work

Revtro:
> > If no network tag is not completely allowed in transit, is it possible to create and use operator/highway/bus_stop.json?
> 
> Yes, that would work

I understand, I have created it.
I am sorry if my property settings are incorrect as I'm not sure.

bhousel:
> If no network tag is not completely allowed in transit, is it possible to create and use operator/highway/bus_stop.json?

This will probably work, but I really encourage you to work with the local community and try to bring your tagging of bus stops to match what the rest of the world does.   It is kind of weird for us to create a special operator/bus_stop file just for a few items, and I can't guarantee that it won't cause surprises.

https://wiki.openstreetmap.org/wiki/Key:network#Public_transit

Revtro:
Created the PR as #7368.

PR #6856, opened 2022-07-30
Update charging_station.json

Samuel12321:
Add Zero brand operated by meridian.

Related: #6727

PR #7325, opened 2022-11-05
Update charging_station.json

Samuel12321:
Add Zero Charging Network - owned and operated by meridian

### DIFF
--- a/data/brands/amenity/charging_station.json
+++ b/data/brands/amenity/charging_station.json
@@ -81,6 +81,18 @@
         "name": "Mer"
       }
     },
+        {
+      "displayName": "Zero",
+      "locationSet": { "include": [ "nz" ],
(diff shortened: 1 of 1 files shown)


arch0345:
It looks like the entry for Meridian Energy in the operators tree is conflicting with this new entry. This is causing the same matchIndex error shown in https://github.com/osmlab/name-suggestion-index/pull/6856#issuecomment-1200485013

Samuel12321:
> It looks like the entry for Meridian Energy in the operators tree is conflicting with this new entry. This is causing the same matchIndex error shown in [#6856 (comment)](https://github.com/osmlab/name-suggestion-index/pull/6856#issuecomment-1200485013)

Have submitted A PR to remove Meridian as an operator as they have rebranded their chargers to Zero.

PR #7329, opened 2022-11-05
Add shop=gold_buyer + french shop "Or en Cash"

bmillemathias:
ssia

The shops are legally written "Or en Cash" (and on the website too) but they are displayed in streets as "Or en Ca$h" so I hope the definition `"matchNames": ["Or en Ca$h"]` is correct.

https://overpass-turbo.eu/s/1nqI

Thanks


bmillemathias:
Thanks for the review @UKChris-osm

PR #7355, opened 2022-11-06
Change ENEOS(エネオス) notation and merge (remove) EneJet(エネジェット)

Revtro:
For ENEOS, I adapted the notation used on the official website, etc., and originally commonly used in OSM.
I think it is a similar case to apollostation as to whether Japanese or English notation should be used.

As for EneJet, according to the official store locator service, the name of the store itself is "ENEOS" and instead the name of the branch begins with "EneJet" (i.e., attributional, like whether it is self-service or not), so I think there is no need to separate them by brand or name tags.

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -6179,23 +6179,8 @@
       }
     },
     {
-      "displayName": "エネオス",
+      "displayName": "ENEOS",
       "id": "eneos-fbe2e4",
-      "locationSet": {"include": ["jp"]},
-      "tags": {
-        "amenity": "fuel",
-        "brand": "エネオス",
-        "brand:en": "ENEOS",
-        "brand:ja": "エネオス",
-        "brand:wikidata": "Q1640290",
-        "name": "エネオス",
-        "name:en": "ENEOS",
-        "name:ja": "エネオス"
-      }
-    },
-    {
-      "displayName": "エネオス エネジェット",
-      "id": "enejet-fbe2e4",
       "locationSet": {"include": ["jp"]},
       "matchNames": [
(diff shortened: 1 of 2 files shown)


Revtro:
By the way, when I run npm run build, it seems that "エネオス エネジェット" without English name is unexpectedly generated with location as global.

arch0345:
> By the way, when I run npm run build, it seems that "エネオス エネジェット" without English name is unexpectedly generated with location as global.

This is because brands found over >50 times on the [planet file](https://planet.osm.org/) are collected and added to NSI automatically 
(according to Taginfo [there are currently 75 features with `brand=エネオス エネジェット`](https://taginfo.openstreetmap.org/search?q=brand%3D%E3%82%A8%E3%83%8D%E3%82%AA%E3%82%B9+%E3%82%A8%E3%83%8D%E3%82%B8%E3%82%A7%E3%83%83%E3%83%88)). I'll add this to `matchNames`

Revtro:
> > By the way, when I run npm run build, it seems that "エネオス エネジェット" without English name is unexpectedly generated with location as global.
> 
> 
> 
> This is because brands found over >50 times on the [planet file](https://planet.osm.org/) are collected and added to NSI automatically 
> 
> (according to Taginfo [there are currently 75 features with `brand=エネオス エネジェット`](https://taginfo.openstreetmap.org/search?q=brand%3D%E3%82%A8%E3%83%8D%E3%82%AA%E3%82%B9+%E3%82%A8%E3%83%8D%E3%82%B8%E3%82%A7%E3%83%83%E3%83%88)). I'll add this to `matchNames`

Understood, thank you for the correspondence work.

PR #7357, opened 2022-11-06
Add T-ZONE

zyphlar:


tas50:
@zyphlar This will need a wikidata entry to show up in editors

arch0345:
I'm not sure if this brand is notable enough for NSI: [according to their website](https://tzonegifts.myshopify.com/) they only have 3 locations, all of which are in Salem

LaoshuBaby (changes requested):
Maybe a only 3 branch shop's brand can't be called "Canonical common"

bhousel:
Yes please add some supporting references to [Q115111848](https://www.wikidata.org/wiki/Q115111848) or it's likely to be deleted by admins later.

The threshold for notability on Wikidata is pretty low.  If this brand has been written up in a newspaper with an article saying "T-ZONE opens 3 new locations" or "T-ZONE names new CEO" or anything like that, that's what we are looking for.  Just add the reference to the Wikidata page, according to this guide:
https://github.com/osmlab/name-suggestion-index/wiki/Editing-Wikidata

zyphlar:
I'm not going to fight hard for my case, again it just seemed like the easiest way to indicate an exception to the osmose uppercase name checker. Seems to me like we don't need to document every single POI that uses uppercase in its name, but we could sure document the fact that it's indeed a real brand-name with real locations.

LaoshuBaby:
> the easiest way to indicate an exception to the osmose uppercase name checker.

If this shop name is correct, and reflect on the ground things, I think it should be ascribe to Osmose's checker's, and I think open a issue in there: ( https://github.com/osm-fr/osmose-backend/issues ) maybe more helpful. NSI isn't only for osmose, but many osm related project such as iD, JOSM, Every Door, etc. If we collect a new brand, at least it was or will be common in specific region, then it worse a entry. 

https://github.com/ideditor/nsi-collector/blob/0594a6049c45e3bb91a96d6d29b443b164aff360/collect_osm.js#L33-L39
图片

(If code can't load correctly, here is a screenshot about it)

This is threshold of nsi-collector, in some situation the threshold may be looser but a 3 shops brand really unacceptable, sorry.

zyphlar:
Can you add a Wiki description for what's considered notable so we don't waste time? I looked but couldn't find anything besides a general purpose and how-to statement.

I unfortunately doubt that Osmose will be responsive. In the best case it would perhaps check the brand and/or wikidata before throwing an error, but that seems like even more work than adding this entry to NSI, and so the most likely outcome is that there will just be perpetually open Osmose issues on these locations.

bhousel:
> Can you add a Wiki description for what's considered notable so we don't waste time? I looked but couldn't find anything besides a general purpose and how-to statement.

https://www.wikidata.org/wiki/Wikidata:Notability

LaoshuBaby:
> Can you add a Wiki description for what's considered notable so we don't waste time? I looked but couldn't find anything besides a general purpose and how-to statement.

I've add it just now, our document is still writing continually:
https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case/_compare/ecbb13ec9309575476dff86512979d50a74f6d45...9b4d7df06ef5ec359f3f087bb5df58acbf8a9fd5

zyphlar:
@bhousel yes, three physical retail businesses usually count as identifiable, conceptual, material entities. A business website is publicly available and serious, we can also easily call/email/post the place to verify that it's a material entity. If we removed all wikidata items that haven't been mentioned in a newspaper I think that'll be a lot of them.

bhousel:
Sure, we can accept this, but it wouldn't surprise me if the QID just gets deleted anyway.  It's just 3 stores.

PR #7360, opened 2022-11-07
Update charging_station.json

Samuel12321:
Remove Meridian as Operator as the network has been rebranded Zero. Fix error https://github.com/osmlab/name-suggestion-index/pull/7325#issuecomment-1304985808


arch0345:
Thanks @Samuel12321

PR #7363, opened 2022-11-07
Add free.fr as shop to the nsi

TheSnoozer:
Adds https://www.free.fr/boutiques/ to the NSI.

Personally I just tagged it as `shop=telecommunication`, but since It's similar to SFR or Orange (which are tagged as `shop=mobile_phone`) I'd suggest to tag it similar to already existing brands.

### DIFF
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -326,6 +326,17 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Thank you @TheSnoozer

TheSnoozer:
Thank you for your review and of course getting this integrated!

PR #7369, opened 2022-11-08
Add Asurion

Cj-Malone:
uBreakiFix is in the process of rebranding to Asurion Tech Repair & Solutions in the US, I don't know if then end result is for no more uBreakiFix in the US but currently there is a mix of both. 

[OSMUS Slack](https://osmus.slack.com/archives/CDJ4LKA2Y/p1667899800329159)
[ATP](https://github.com/alltheplaces/alltheplaces/pull/4194)

### DIFF
--- a/data/brands/craft/electronics_repair.json
+++ b/data/brands/craft/electronics_repair.json
@@ -119,7 +119,7 @@
         "brand": "uBreakiFix",
         "brand:wikidata": "Q53685733",
         "craft": "electronics_repair",
-        "electronics_repair": "phone",
+        "electronics_repair": "computer;phone",
         "name": "uBreakiFix"
       }
     },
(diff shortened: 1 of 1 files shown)


data/brands/craft/electronics_repair.json
```diff
@@ -123,6 +123,17 @@
...
       }
     },
+    {
+      "displayName": "Asurion Tech Repair & Solutions",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "brand": "Asurion Tech Repair & Solutions",
+        "brand:wikidata": "Q4811938",
+        "craft": "electronics_repair",
+        "electronics_repair": "phone",
```
UKChris-osm, line 133: Could the `electronics_repair` tag be expanded if they offer more than just phone repairs? Do they also offer computer / appliance repairs in store?

Cj-Malone: Possibly `electronics_repair=computer;phone`, but I'm not completely sure, not an American.

This is a copy/paste of the uBreakiFix. So if so that can probably be updated too.

Cj-Malone: So asurion lists [appliances](https://www.asurion.com/repairs/appliance/) on there website too, I can't see the same on uBreakiFix.

PR #7389, opened 2022-11-13
Add a GitHub Action that automatically runs `npm run build`

### LINKED ISSUE #7383: Can't we do a "npm run build" after each merge request
Hello,

just asking if something prevent to run a "npm run build" and commit the result in the repository ?
I don't know if it makes sense and how difficult is it.

arch0345:
Closes #7383

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,39 @@
+name: Build
+
+permissions:
+  contents: write
+
+on:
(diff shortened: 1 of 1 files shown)


bhousel:
I'm against the idea of having github actions perform commits.
I still dont know why people want to have another action that runs builds, when this already happens on the test action, which runs on every push or merge.

PR #7391, opened 2022-11-13
Add ServiceStore DB, a convenience store chain in Germany

brawer:


kjonosm:
Thanks @brawer! Checking [Overpass turbo](https://overpass-turbo.eu/s/1nHh) I added a matchName

brawer:
Thank you!

PR #7393, opened 2022-11-13
add Vanessa Bruno women clothes shops

bmillemathias:
The brand has shops worldwide according https://www.vanessabruno.fr/nos-boutiques.php and came across a couple when walking by big city near home.


kjonosm:
Thanks @bmillemathias !

PR #7417, opened 2022-11-17
Add amenity=photo_booth

Codain:
Add amenity=photo_booth and initialization with some brands


Codain:
On hold for wikidata improvements.

UKChris-osm:
Hi @Codain - just to let you know the file needed a `.json` file extension.

UKChris-osm:
I'm wondering if this would be more suitable as part of the `operator `tree rather than `brand,` as looking at the NSI Collection data, there are no `brands` that appear in the data, but a few under [operator](https://raw.githubusercontent.com/ideditor/nsi-collector/main/dist/osm/operators_all.json).

Adamant36:
> looking at the NSI Collection data, there are no brands that appear in the data, but a few under operator.

I was looking into this earlier and seems like the IRL photo booths are "branded" with the names. So I don't see why it would be a problem to merge it as is whatever the details of how things are currently tagged in OSM are. IMO there isn't enough of these mapped yet to say there's any kind of accepted default that we would be overriding by accepting this either. So unless there's any serious objections I'm going to just merge it. Then we can update it going forward if need be. 

starboothcubebio_UK

starbooth-ID

fotofix-photo-booth-foto-fix-copy-space-to-right-50303080

PR #7427, opened 2022-11-20
Please add the following Belgian telecom operators to the index

mstsbe:
I've added 3 major Belgian telecom operators to the list: Proximus, Telenet and VOO. Proximus is the former state-owned incumbent, and was the historic provider of PSTN services, as well as the first mobile network in Belgium.
Telenet was established as the first private operator in Flanders in the 1990s, and meanwhile offers triple and quad-play services.
VOO is the counterpart of Telenet in Wallonia and some communities in the Brussels Capital Region.

All 3 operate both shops in the categories telecommunications (triple-play: internet, fixed line, IPTV), as well as shops that are already tagged as mobile_phone (all 3 also sell mobile phones and subscriptions, also in triple-play/quad-play packs).

Proximus and Telenet also operate their own communications network as well (both fiber and copper-based), so I've added a new file man_made -> street_cabinet. Proximus, as the former incumbent also operates its own copper telecom exchanges, therefore was also added to telecom -> exchange.

Please add these to the index.


Dimitar5555:
There is no need to duplicate them in both files. If you remove the entries from `shop/mobile_phones.json`, iD will automatically suggest to change `shop=mobile_phones` to `shop=telecommunication`.

mstsbe:
> There is no need to duplicate them in both files. If you remove the entries from `shop/mobile_phones.json`, iD will automatically suggest to change `shop=mobile_phones` to `shop=telecommunication`.

I've removed the entries from mobile_phone.json, so now they're only in telecommunication.json and street_cabinet.json.
Can someone please add the brands to the index now?

PR #7451, opened 2022-11-24
Polish pharmacies - remove duplicates

mutipg:
### id_to_remove > as_duplicate_of_id
aptekagemini-cf61fb > gemini-cf61fb
aptekarodzinna-cf61fb > rodzinna-cf61fb
aptekasloneczna-cf61fb > sloneczna-cf61fb

### DIFF
--- a/data/brands/amenity/pharmacy.json
+++ b/data/brands/amenity/pharmacy.json
@@ -226,39 +226,6 @@
       }
     },
     {
-      "displayName": "Apteka Gemini",
-      "id": "aptekagemini-cf61fb",
-      "locationSet": {"include": ["pl"]},
-      "tags": {
-        "amenity": "pharmacy",
-        "brand": "Apteka Gemini",
-        "healthcare": "pharmacy",
-        "name": "Apteka Gemini"
-      }
-    },
-    {
-      "displayName": "Apteka Rodzinna",
-      "id": "aptekarodzinna-cf61fb",
-      "locationSet": {"include": ["pl"]},
-      "tags": {
-        "amenity": "pharmacy",
-        "brand": "Apteka Rodzinna",
-        "healthcare": "pharmacy",
-        "name": "Apteka Rodzinna"
-      }
-    },
-    {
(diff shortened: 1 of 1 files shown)


data/brands/amenity/pharmacy.json
```diff
@@ -225,39 +225,6 @@
         "name": "Apotheka"
       }
     },
-    {
-      "displayName": "Apteka Gemini",
-      "id": "aptekagemini-cf61fb",
-      "locationSet": {"include": ["pl"]},
```
LaoshuBaby, line 231: There are still a lot elements in OSM named after this according to taginfo, also "Apteka Rodzinna" and "Apteka Słoneczna"

图片

So if you only removed those 3 entry, they will be created once again, our collector will found they are still "widely used" in real world.

If this change is from business merge or rebrand, or bought by another brand, I'd suggest you add a `"matchNames": ["Apteka Gemini"],` in your targeted `gemini-cf61fb`, another 2 also need this.

For example, "頂好超市" was bought by "家樂福超市":

https://github.com/osmlab/name-suggestion-index/blob/f57288c32f5fb2de87fc6c8bc5943ba818bf403b/data/brands/shop/supermarket.json#L9696-L9701

LaoshuBaby (changes requested):

kjonosm:
@matkoniecz You seem to have a good overview of the Polish tagging preferences. Which version do you think we should keep/promote in NSI?

matkoniecz:
here I am not really sure, so created https://community.openstreetmap.org/t/sloneczna-czy-apteka-sloneczna/6296

Dimitar5555:
@matkoniecz Is there any result from the discussion on the forum? If there isn't, it may be beneficial to create a poll and see if there is a majority which prefers one of the options.

matkoniecz:
Skipping "apteka" ("pharmacy") in name seems preferred but it is not unanimous.

bhousel:
stale

PR #7453, opened 2022-11-24
Cyprus telecom brands

dmromanov:
https://en.wikipedia.org/wiki/Telecommunications_in_Cyprus


Dimitar5555:
@dmromanov It would be better to add them in `shop=telecommunication`, not in `office=telecommunication. The first one is for the store fronts, the second one is for the corporate offices.

PR #7457, opened 2022-11-25
Edwin jeans

codeinabox:


codeinabox:
@Revtro for the Japan region should I have separate entry where the name is エドウイン? Based on my trips to Japan and the Japanese website they also use Edwin

Revtro:
I think it is a different industry but similar to ENEOS, so the name tag could be (capitalized) EDWIN.

Also, since the Japanese reading of the name is "エドウィン", unlike the official name "エドウイン", it would be more useful to add it to some name tag, such as the alt_name tag.

codeinabox:
@Revtro Thank you, I've updated accordingly

LaoshuBaby (approved):

PR #7459, opened 2022-11-27
Remove name from WIlson Parking

Kovoschiz:
It is only a contractor for outsourced parking management, not having any dedicated carparks. Users have added useless `name=Wilson Parking`  to `=parking`, when the correct name is the carpark of the building (mall, hotel, etc) it is in.

### DIFF
--- a/data/operators/amenity/parking.json
+++ b/data/operators/amenity/parking.json
@@ -824,24 +824,5 @@
         "operator:wikidata": "Q11415765"
       }
     },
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Thanks @Kovoschiz, but if the `brand` and `operator` data is the same, then I would remove one of these. As this is already part of the `operator` tree, I'd retain that and remove the `brand` data.

PR #7468, opened 2022-11-29
Update vending_machine.json

paunofu:
Added teika, brand of vending machines

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -3837,6 +3837,17 @@
       }
     },
     {
+      "displayName": "Pomodoro",
+      "locationSet": {"include": ["es"]},
+      "tags": {
+        "amenity": "restaurant",
+        "brand": "Pomodoro",
+        "brand:wikidata": "Q7316076",
+        "cuisine": ["pasta","pizza","burger","italian"],
+        "name": "Pomodoro"
+      }
+    },
+    {
       "displayName": "Ponderosa Steakhouse",
       "id": "ponderosasteakhouse-56c3a6",
       "locationSet": {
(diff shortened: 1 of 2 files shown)


paunofu:
Added Pomodoro, franchise of restaurants in Spain

data/brands/amenity/restaurant.json
```diff
@@ -3836,6 +3836,17 @@
...
       }
     },
+    {
+      "displayName": "Pomodoro",
+      "locationSet": {"include": ["es"]},
+      "tags": {
+        "amenity": "restaurant",
+        "brand": "Pomodoro",
+        "brand:wikidata": "Q7316076",
+        "cuisine": ["pasta","pizza","burger","italian"],
```
LaoshuBaby, line 3846:
```suggestion
        "cuisine": "pasta;pizza;burger;italian",
```

https://github.com/osmlab/name-suggestion-index/wiki#about-openstreetmap

I don't know why the test even passed, but looks like current schema haven't change
So, the value of `cuisine` key shouldn't be a **list** of string, a **string** is enough.

(If I missed some technical changes, please point it to me, thanks!)

data/brands/amenity/vending_machine.json
```diff
@@ -762,6 +762,16 @@
         "vending": "newspapers"
       }
     },
+    {
+      "displayName": "Teika",
+      "locationSet": {"include": ["es"]},
+      "matchNames": ["teika","Teica"],
```
LaoshuBaby, line 768: `teika` and `Teika` only have difference in uppercase letter so `teika` don't need to be added. Don't worry, it will be auto removed when formatter run and you don't need to do anything.

PR #7474, opened 2022-11-30
Update police.json

Mannivu:
Fix capitalization: "locale" and "municipale" are simple adjectives, not proper nouns, so they should start with a lowercase letter

### DIFF
--- a/data/operators/amenity/police.json
+++ b/data/operators/amenity/police.json
@@ -2404,27 +2404,6 @@
       }
     },
     {
-      "displayName": "Polizia locale",
-      "id": "polizialocale-f71f16",
-      "locationSet": {"include": ["it"]},
-      "tags": {
-        "amenity": "police",
-        "operator": "Polizia locale",
-        "operator:wikidata": "Q61634147"
-      }
-    },
-    {
-      "displayName": "Polizia municipale",
-      "id": "poliziamunicipale-f71f16",
-      "locationSet": {"include": ["it"]},
-      "matchNames": ["vigili urbani"],
-      "tags": {
-        "amenity": "police",
-        "operator": "Polizia municipale",
-        "operator:wikidata": "Q1431981"
-      }
-    },
-    {
       "displayName": "Prefectura Naval Argentina",
       "id": "prefecturanavalargentina-e73d55",
       "locationSet": {"include": ["ar"]},
--- a/data/operators/amenity/police.json
+++ b/data/operators/amenity/police.json
@@ -2404,23 +2404,23 @@
       }
     },
     {
-      "displayName": "Polizia Locale",
+      "displayName": "Polizia locale",
       "id": "polizialocale-f71f16",
       "locationSet": {"include": ["it"]},
       "tags": {
         "amenity": "police",
-        "operator": "Polizia Locale",
+        "operator": "Polizia locale",
         "operator:wikidata": "Q61634147"
       }
     },
     {
-      "displayName": "Polizia Municipale",
+      "displayName": "Polizia municipale",
       "id": "poliziamunicipale-f71f16",
       "locationSet": {"include": ["it"]},
       "matchNames": ["vigili urbani"],
       "tags": {
         "amenity": "police",
-        "operator": "Polizia Municipale",
+        "operator": "Polizia municipale",
         "operator:wikidata": "Q1431981"
       }
     },


Adamant36:
It looks like they can be either way. For instance Wikipedia has it as capital letters. Same goes for their [official website](https://www.poliziamunicipale.it/).

Mannivu:
> Same goes for their [official website](https://www.poliziamunicipale.it/)

If we follow the official website, we should write "PoliziaMunicipale" without spaces. Anyway, "municipale" and "locale" are simple adjectives, and those police forces are not centralised (so, the official website is not really an official website), and they are not linked with each other. They all respond to the city mayor, and are not part of any Ministery (which is the case for Arma dei Carabinieri, Polizia di Stato, Guardia di Finanza, and Guardia Costiera; these are also capitalised since those are proper noun, not simple adjectives).

kjonosm:
By current numbers, the Italian community clearly favours `Polizia Municipale` [overpass link](https://overpass-turbo.eu/s/1oxv) over `Polizia municipale` [overpass link](https://overpass-turbo.eu/s/1oxu). It also clearly favours `Polizia Locale` [overpass link](https://overpass-turbo.eu/s/1oxA) over `Polizia locale` [overpass link](https://overpass-turbo.eu/s/1oxB).

I don't see a reason for NSI to change accepted tagging in that regard.

bhousel:
Just want to confirm that "Local Police" and "Municipal Police" really commonly tagged things, like a recognizable brand?  

Normally we would exclude names like this from NSI.. but maybe in Italy this is what you want?

Mannivu:
> By current numbers, the Italian community clearly favours Polizia Municipale [overpass link](https://overpass-turbo.eu/s/1oxv) over Polizia municipale [overpass link](https://overpass-turbo.eu/s/1oxu). It also clearly favours Polizia Locale [overpass link](https://overpass-turbo.eu/s/1oxA) over Polizia locale [overpass link](https://overpass-turbo.eu/s/1oxB).

@kjonosm  this might be affected from the NSI itself, since I'm not adding something new, but updating something that has been present on the NSI for awhile (so, stating that the local communty prefers the Capital letter version it's not correct; there should be a more deep research).

@bhousel this is pretty hard to articulate, but I'll do my best. In Italy, almost every Comune (which is the smallest administrative entity in Italy) has a Polizia municipale, which are not linked between them, but all have the same duties (which are regulated by a specific law). So, even if they are not entirely linked, they act as a linked network. The Polizia locale is an hyperonym, since it comprehends Polizia municipale, Polizia provinciale and Polizia regionale. But, explaining it to you, I wonder if you are right and maybe a network tag is more appropriate. I'll ask with the Italian community.

kjonosm:
@Mannivu looking at the different colours in linked overpass turbo query results one can assume that only green coloured  points are probably updated via NSI suggestion. The remaining yellow points still outnumber your suggested spelling by far.

kjonosm:
However, @bhousel's question remains, whether to keep these general entries for those types of police forces at all or just exclude them because they really should be differentiated as e.g. listed here: https://www.poliziamunicipale.it/comuni

Adamant36:
>The official website is not really an official website

By official I really meant "main" since what we usually go with when it comes to how to spell a default is what people are most likely going to see IRL (or in this case when looking at "official" documents). Personally, I could really care if it's "official" or not, but the fact remains that people are usually going to see Polizia Municipale in capital letters when viewing the website. As well as well as other places, for instance their police uniforms. I'm sure you'll have a reason that how it's spelled on the uniforms should be disregarded to though.
italian-polizia

What about how it's spelled on the sides of their cars? 

360_F_468306719_UKA4Vp3HwN6ek9tN6Kaa304NgpzerF69

No? What about the signs in front of their buildings? 

italian-police-station-with-polizia-municipale-car-parked-outside-E62MYK

I'm sure you get my point.

Mannivu:
UPDATE
Based on the ML inquiry, the Italian community has suggested that "Polizia locale" and "Polizia municipale" should not be part of the NSI, since there's no central governance and are therefore all separate entities.

bhousel:
> Based on the ML inquiry, the Italian community has suggested that "Polizia locale" and "Polizia municipale" should not be part of the NSI, since there's no central governance and are therefore all separate entities.

Thanks for checking!
I'm also going to add these as named exclude so they dont return when we process the osm planet.

Mannivu:
> I'm also going to add these as named exclude so they dont return when we process the osm planet.

Thank you!

PR #7484, opened 2022-12-03
lots of New Zealand brands

k-yle:


kjonosm:
At least in Germany, Ticketmaster ticket shops are operated by different franchisees. So maybe just leave the operator info out or create a single entry for NZ.

k-yle:
thanks, i've removed the `operator` tags for that entry

PR #7492, opened 2022-12-04
Add a cafe brand 'Ediya'

lens0021:


arch0345:
Thanks @lens0021 and welcome!

PR #7494, opened 2022-12-05
Fix typo

u-spec-png:
There should be something to prevent this from happening in future, this is second time this typo happened.


LaoshuBaby:
Sorry this is my fault, I haven't found this out in #7491

But currently we haven't a validator about name subtag, if needed you can open a new issue request this.

PR #7497, opened 2022-12-05
Updated locationSet of Santander US

paunofu:
Santander with wikidata Q5835668 is exclusively from USA https://www.santanderbank.com/ and I also wanted to update the ATM (santander-2f6feb) but I didn't find where is it

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -8604,7 +8604,7 @@
     {
       "displayName": "Santander",
       "id": "santander-b7a026",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
> I also wanted to update the ATM (santander-2f6feb) but I didn't find where is it

https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/amenity/atm.json#L2-L18

Because in most cases you don't need to add ATM manually, its tags are inherited from bank.

PR #7504, opened 2022-12-06
Remove dead code for one brand and improve one preset with adding a "…

Lukas458:
…vending" value

Hi,
here I removed dead code for the brand "Tobaccoland", because the code line isn't needed anymore, and I added the key "vending", so that the preset is more complete. I also added "preserveTags", because sometimes the vending key may differ from vending=cigarettes (e. g. vending=cigarettes;condoms). See also https://github.com/osmlab/name-suggestion-index/pull/6301

### DIFF
--- a/data/brands/amenity/vending_machine.json
+++ b/data/brands/amenity/vending_machine.json
@@ -18,7 +18,6 @@
         "^zigarettenautomat$",
         "^自販機$"
       ],
-      "named": ["^tobaccoland$"]
     }
   },
   "items": [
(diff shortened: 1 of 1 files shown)


bhousel:
`preserveTags` needs to be outside of the "tags" section, like up where `matchNames` is.
(The way you have it now, it will try to make a tag with key "preserveTags").

Also it needs to be an Array in square brackets, like `"preserveTags": ["vending"]`

Lukas458:
I see. Thank you very much for your help!

I made the requested changes.

PR #7506, opened 2022-12-07
[nz] Add brand:wikidata to a liquor shop

k-yle:
Looking at [the history of the wikidata item](https://www.wikidata.org/w/index.php?title=Q112038718&action=history), I can't tell why it was removed from NSI days after it was originally added...


LaoshuBaby:
图片

“Super Liquor” had been moved to this file 2 years ago, and at that time it haven't been linked to Wikidata.

Maybe some maintainer conduct a check that removed orphan property for items that have invalid P8253.

PR #7520, opened 2022-12-10
Fix Y!mobile

ffff-23:
Y!mobile (Q5322556) was  absorption merger into Softbank.
Y!mobile as the current Softbank brand is Q18455742.

### DIFF
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -1348,7 +1348,7 @@
         "brand": "ワイモバイル",
         "brand:en": "Y!mobile",
         "brand:ja": "ワイモバイル",
-        "brand:wikidata": "Q5322556",
+        "brand:wikidata": "Q18455742",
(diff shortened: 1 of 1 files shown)


Revtro:
As far as I have reviewed, SoftBank is actually running the brand now, and Wikipedia and Wikidata reflected that long ago, so I think this change is appropriate.

LaoshuBaby:
So had the logo show on the street changed to Softbank's? If not, I think this changes should only be done on wikidata aspect and keep this NSI entry.

Revtro:
No, the logo itself has not been changed.
But, since the corporate status of Ymobile Corporation described in NSI's Wikidata ([Q5322556](https://www.wikidata.org/wiki/Q5322556)) has already ceased to exist and the brand has been transferred to SoftBank Corporation, I think [Q18455742](https://www.wikidata.org/wiki/Q18455742) is more appropriate for the Wikidata of the brand itself.

PR #7525, opened 2022-12-12
Add London Borough of Newham to amenity/parking

rskedgell:


UKChris-osm:
Thanks @rskedgell - but I would suggest that [Q7018141](https://www.wikidata.org/wiki/Q7018141) would be the correct Wikidata, as the Newham London Borough Council would run the car parks?

rskedgell:
On 12/12/2022 16:15, UKChris-osm wrote:
> Thanks @rskedgell <https://github.com/rskedgell> - but I would suggest 
> that Q7018141 <https://www.wikidata.org/wiki/Q7018141> would be the 
> correct Wikidata, as the Newham London Borough Council would run the car 
> parks?

Quite right. I hadn't noticed the separate articles for the 
administrative area vs the council itself.

Adamant36:
I'm merging this since it looks like the Wikidata has been changed. Thanks @rskedgell 👍

PR #7526, opened 2022-12-12
Remove Papyrus

Cj-Malone:
https://en.wikipedia.org/wiki/Papyrus_(company)

Closed in early 2020, the branded products are still sold through other shops like Kroger and RiteAid.

I've never removed anything from NSI before, it this all we do?

### DIFF
--- a/data/brands/shop/gift.json
+++ b/data/brands/shop/gift.json
@@ -268,17 +268,6 @@
       }
     },
     {
-      "displayName": "Papyrus",
-      "id": "papyrus-43d507",
-      "locationSet": {"include": ["us"]},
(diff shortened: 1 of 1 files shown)


UKChris-osm:
You would also need to add the brand to the exclude list at the top of the file if there are more than 50 POI's currently within the OSM database, otherwise the planet scan will keep adding it again.

Cj-Malone:
Only 24 with the [Wikidata](https://taginfo.openstreetmap.org/tags/brand%3Awikidata=Q28222692) but 82 with the [name](https://taginfo.openstreetmap.org/tags/name=Papyrus)

PR #7532, opened 2022-12-13
Added short_name for Japan Post

Lonerat:
Added officially used short name.
https://www.post.japanpost.jp/
<s>Also, since courier service is different from post office, I tried to create office=courier written in JA:How to map A (https://wiki.openstreetmap.org/wiki/JA:How_to_map_a#%E3%82%BF) and move Yamato Transport and Sagawa Express.</s>


Adamant36:
I have reservations about the usage of office=courier here since the numbers for it are much lower then the alternatives. For instance office=logistics and amenity=post_office. Plus, it's not clear how a courier is distinct from those tags either. I actually brought that up on the office=courier articles talk page last year but no one replied. There should really be some discussion somewhere about the tag before it's implemented on a wide scale like this though. My suggestion would be to ask about it on the [community forums](https://community.openstreetmap.org/) (probably [general talk/tagging](https://community.openstreetmap.org/c/general/tagging/), or the tagging mailing list (which I unfortunately don't have a link to right now but is easy to find).

Lonerat:
I see, then I will only add the short_name of Japan Post in this PR for now.
I had mistakenly thought it was accepted in Japan because it was written in JA:How to map A, the tagging rules for POI in Japan, I'm sorry.

Adamant36:
Thanks 👍

PR #7533, opened 2022-12-13
Adding wikidata to power files

michael-reuter:


Adamant36:
I'm not a super big fan of the operator tags being changed to the long form, but I'm going to merge this anyway. Maybe either don't do that next time or at least make it clear your doing more then just adding Wikidata entries. Otherwise it could lead to problems because it's easy to miss things like that when reviewing the code and some people don't like certain fields being arbitrarily changed. Especially if it's being done under the radar. Thanks 👍

PR #7553, opened 2022-12-18
fix Create SD(クリエイトSD)

ffff-23:
The name of the store is "クリエイトSD", but the sign of the store says "薬CREATE", so I think it is appropriate to make the brand "薬CREATE".
IMG_1534
IMG_1551

### DIFF
--- a/data/brands/amenity/pharmacy.json
+++ b/data/brands/amenity/pharmacy.json
@@ -3843,9 +3843,9 @@
       "locationSet": {"include": ["jp"]},
       "tags": {
         "amenity": "pharmacy",
-        "brand": "クリエイトSD",
-        "brand:en": "CreateSD",
-        "brand:ja": "クリエイトSD",
+        "brand": "薬CREATE",
+        "brand:en": "CREATE",
+        "brand:ja": "薬クリエイト",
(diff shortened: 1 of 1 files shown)


Lonerat:
In my opinion, that is not an official name (The intended meaning would rather be "薬 CREATE" rather than "薬CREATE", meaning it is a store called CREATE, that sells drugs), and should be in an alternative tag for searching, such as `alt_name` key.

Lonerat:
I'm not sure if this will affect but I hit the revert button by mistake, I'm sorry.

ffff-23:
For reference, the name "薬クリエイト" is used on the business introduction page* of the holding company.

However, it could also be in the form `alt_name=薬CREATE` , `alt_name:ja=薬クリエイト`


*https://www.createsdhd.co.jp/individual/tabid/63/Default.aspx (3.他社に比べてどんな特徴があるの？)

Lonerat:
I saw it used in few pages on the website, but the number of hits in search results was less than "クリエイトSD", also other brands, especially on signs, may be written as "薬" + "original name", so I think it isn't the actual name rather to say the official name.
Therefore, I personally think it should rather be moved to the `alt_name` key.

PR #7557, opened 2022-12-20
Modified 2 banks and added 2 more

paunofu:
I have added matchNames and name:ca for Caixa Ontinyent and Caixa Popular, also added Caixa Altea and Caixa Callosa


Adamant36:
For the banks with "rural" match names, it looks like they are different unique brands that are separate from the main ones. Not just a spelling error or whatever. So they should probably have their own entries. 

5b46de91-440e-4b3b-b760-60ad902f6b2b_source-aspect-ratio_default_0

paunofu:
I’m not sure I understand your point. Banks called Caja Rural ... means they operate in rural areas. Also, this bank may be members of the Caja Rural Group, of which the bank in the photo is a member or of the Caja Mar Group of which the two I add are members. Although the banks belong to a group, the group does not control the bank, they only "join forces".

https://www.grupocooperativocajamar.es/es/comun/
Grupo Caja Mar


https://www.grupocajarural.es/en
Grupo Caja Rural

paunofu:
My intention is to add all the banks of both groups, but first I have started with the two I know

Adamant36:
> I’m not sure I understand your point.

My point is that there should be individual entries for each of the member banks. Instead of them just being match names. For instance Caja Rural de Asturias has like [113 ATMs](https://www.cajaruraldeasturias.com/en/atm-network) that they operate in the Oviedo area. So they should have a separate entry in the index for those ATMs instead of just being a match name of Grupo Caja Rural. The fact that they are a member of Grupo Caja Rural doesn't really matter. They are a unique brand and should have their entry because of it. Same goes for the other "rural" banks. Hopefully that explains things better.

paunofu:
But, I'm not trying to add the group, I'm trying to add two banks, "Caixa Rural Altea" and "Caixa Rural Callosa d'en Sarrià". This is why I don't understand your point

data/brands/amenity/restaurant.json
```diff
@@ -5004,6 +5004,17 @@
         "name": "Twin Peaks"
       }
     },
+    {
+      "displayName": "UDON",
+      "locationSet": {"include": ["001"]},
```
codeinabox, line 5009: UDON isn't a global franchise, they only operate in the following areas:

Spain
Portugal
República Dominicana
USA
Puerto Rico

codeinabox (changes requested):

Adamant36:
> But, I'm not trying to add the group, I'm trying to add two banks

I'm aware. But you have the other banks as match-names for the banks that you want to add. Which means those banks will be changed to the ones your adding when they shouldn't be. For instance you have Caixa Rural Callosa as a match name for Caixa Callosa when they are different banks. Do you just not get what the point in match naming is or something?

paunofu:
> For instance you have Caixa Rural Callosa as a match name for Caixa Callosa when they are different banks.

My first language is not English, so maybe I mislead some information. Caixa Callosa and Caixa Rural Callosa is the same bank, you will see Caixa Callosa signed, but in the official papers and Register of Cooperatives at Valencian Community you will see written Caixa Rural de Callosa d'en Sarrià. So, Caixa Callosa, Caixa Rural de Callosa and Caixa Rural de Callosa d'en Sarrià are different names for the same bank. If I mislead something my apologize.

https://www.grupocooperativocajamar.es/caixacallosa/es/comun/informacion-corporativa/sobre-caixacallosa/

Adamant36:
> My first language is not English, so maybe I mislead some information.

No worries. This isn't a super simple thing. I'm just trying to figure out the best way forward without screwing up anything. You say Caixa Callosa and Caixa Rural are the same bank, but they clearly look like different banks in real life and that's what matters. We don't do this based on official papers because that's not how people map things. 

To give an example, there's currently 17 uses of [Caixa Rural Galega](https://taginfo.openstreetmap.org/tags/name=Caixa%20Rural%20Galega) in the database. We can sit here and argue all day about what the official paper says or whatever, but someone choose to name those 17 banks Caixa Rural Galega for a reason and I don't feel like overriding their judgement by having the banks upgraded to Caixa Callosa. 

This 
Cambados_Caixa_Rural_Galega

is clearly different from 

Screenshot 2022-12-22 042847

They don't even have the same colors or logo.

paunofu:
Yes, I will delete the matchName, is the easiest thing. In a future PR when I add all the banks of Grupo Cajamar and Grupo Caja Rural I will explain It better.

Adamant36:
> Yes, I will delete the matchName, is the easiest thing. In a future PR when I add all the banks of Grupo Cajamar and Grupo Caja Rural I will explain It better.

Thank you. Just an FYI but I did a lot of work on adding banks to the index and created/edited a lot of their Wikidata entries. Including Spanish banks. I also did a research paper in college that partly had to do with the origins of the Spanish banking system. As well as cooperative banks. So I'm not ignorant to this topic by any means and I get your explanation perfectly fine. I just disagree with you about it based on my experience doing this for a while and researching the topic. It's not a super big deal though. Match names aren't really necessary anyway. It's not like they can't be added back later if need be either 👍

data/brands/amenity/restaurant.json
codeinabox: Or alternatively you could split this out in a separate PR as this one seems to be focussed on banks

paunofu: Corrected. I'm learning how to use github so sometimes I mistake and add all to the same PR 😅

Adamant36:
The tests are failing because the country codes you added with UDON aren't valid. Can you just delete the entry and add it in another PR please? It's usually better to stick to what the PR was originally opened for instead of adding a bunch of unrelated stuff to it after the fact. Thanks.

data/brands/amenity/restaurant.json
Adamant36: +1 for splitting it.

PR #7559, opened 2022-12-21
Fix Create SD(クリエイトSD)

ffff-23:
It was pointed out that it should be added to `alt_name`, so I reverted `brand` and added it to `alt_name`.

fix #7553


data/brands/amenity/pharmacy.json
```diff
@@ -3853,10 +3853,12 @@
...
       "locationSet": {"include": ["jp"]},
       "tags": {
+        "alt_name": "薬CREATE",
+        "alt_name:ja": "薬クリエイト",
         "amenity": "pharmacy",
-        "brand": "薬CREATE",
-        "brand:en": "CREATE",
-        "brand:ja": "薬クリエイト",
+        "brand": "クリエイトSD",
+        "brand:en": "CreateSD",
```
Lonerat, line 3860:
```suggestion
        "brand:en": "Create SD",
```
This has nothing to do with the issue I mentioned, but as far as the official website is concerned, it might be better to open a space in the English name.

PR #7560, opened 2022-12-21
Update of supermarkets.json

rog-wilco:
Added a new supermarket


arch0345:
Thanks @rog-wilco and welcome!

PR #7564, opened 2022-12-23
Add 4 bank in valencian community

paunofu:
Added Caixa Ontinyent, Caixa Callosa, Caixa Altea and Caixa Popular

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -3696,6 +3696,42 @@
       }
     },
     {
+      "displayName": "Caixa Ontinyent",
+      "locationSet": {"include": ["es"]},
+      "matchNames": ["caixa ontinient","caixa ontiniente","caja ontiniente","caja ontinyent","caja ontinient"],
+      "tags": {
+        "amenity": "bank",
+        "brand": "Caixa Ontinyent",
+        "brand:wikidata": "Q8254941",
+        "name": "Caixa Ontinyent",
+        "name:ca": "Caixa Ontinyent"
+      }
+    },    
+    {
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Community bank? It's a pity that I couldn't understand the introduction in Spanish or Catalan, what I concerned is will they have enough branch banks? Because I feel that community banks serve a small region and may does not have sufficient notability.

paunofu:
This PR comes from a previous PR that I mess it up a little #7557.

Valencian Community is the name of the region, is the same as saying Texas, Ohio,...

Caixa Popular has 74 offices and ATMs
Caixa Altea has offices 11 and ATMs
Caixa Callosa has 3 offices and ATMs
Caixa Ontinyent has 46 offices and ATMs

The number of ATMs could be higher because each office has at least 1 office but could have more.

Adamant36:
I'm a little worried about using match-names to turns of banks called "Caja" into "Caixa" since we already have a bunch of Caja banks in the database and I'm not really sure how it will effect them or really what the difference is. From what I can tell, "Caja" is separate from "Caixa." So they shouldn't be treated as matches, but then again, 🤷‍♂️ @LaoshuBaby or anyone else have an opinion on this or know if it might screw things up somehow down the line?

paunofu:
Oh, okay, now I think why are you worried. Caixa is in Catalan or Galician and Caja is in Spanish. The intention of this PR is add new brands to nsi and correct that people that put the name in Spanish despite of the name is not in Spanish. I thought that matchName corrects the name with the whole entry "Caja Callosa", "Caja Popular",... and not only "Caja".

PR #7565, opened 2022-12-23
Add UDON restaurant

### LINKED ISSUE #7498: UDON restaurants
Whilst travelling in Spain I stumbled upon the Spanish brand UDON that has a lot of restaurants: https://www.udon.com/restaurantes/

It should be tagged as a Japanese restaurant ie

```
amenity=restaurant
name=UDON
cuisine=japanese
```

paunofu:
Added UDON, japanese restaurant requested at issue #7498 

Close #7498

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -5006,12 +5006,12 @@
     },
     {
       "displayName": "UDON",
-      "locationSet": {"include": ["it","es","us","pr","pr","do"]},
+      "locationSet": {"include": ["it","es","us","pr","do"]},
       "tags": {
         "amenity": "restaurant",
         "brand": "UDON",
         "brand:wikidata": "Q115679798",
-        "cuisine": "japanese",
+        "cuisine": "udon",
         "name": "UDON"
       }
     },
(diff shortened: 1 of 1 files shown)


codeinabox:
Thank you! I made a slight tweak as the cuisine is Japanese

LaoshuBaby:
> Thank you! I made a slight tweak as the cuisine is Japanese

Yes, but udon is obvious japanese food? At least when we Asian mention udon we don't think of anything else. I use more specific `udon` is because `japanese` may refer to “nippon ryouri”

enwiki-udon

enwiki-nipponryouri

This may also suitable for `soba`, another Japanese noodle, and we have those kind of case before:

https://github.com/osmlab/name-suggestion-index/blob/d883ddddfff32e69c201e9f8c4f40cb44aa3060f/data/brands/amenity/fast_food.json#L9442-L9458

So I'd prefer a more exact value for it.

codeinabox:
Despite the name and having been to their restaurant in Valencia, I wouldn't say they are a restaurant that specialises in udon as it's only a small part of their menu https://order.udon.com/en/food

PR #7568, opened 2022-12-23
Remove Create(クリエイト)

ffff-23:
Wikidata item for Create(クリエイト) (Q17215731) is an ad agency, not a pharmacy. As far as I know/researched, the only pharmacies with the name "Create" are Create SD(クリエイトSD) (Q11299163) and Create Pharmacy(クリエイト薬局) (Q115682176) operated by Create SD Co., Ltd.

Based on the above, I think it is appropriate to remove it.

### DIFF
--- a/data/brands/amenity/pharmacy.json
+++ b/data/brands/amenity/pharmacy.json
@@ -3833,25 +3833,10 @@
       }
     },
     {
-      "displayName": "クリエイト",
-      "id": "create-650eee",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Thank you for point this error, it was migrated 2 years ago in https://github.com/osmlab/name-suggestion-index/commit/a182f59550f4770a12933ab50a82e325716dd6f2, I'll merge this fix soon.

PR #7576, opened 2022-12-24
Add railway=station suggestion

akadouri:
Hello, I would like to add [`railway=station`](https://wiki.openstreetmap.org/wiki/Tag:railway%3Dstation) suggestions. 

I'm working through stations in New York like this one: https://www.openstreetmap.org/node/7092423541

It would be helpful to have network & operator information fill out once one of the tags is set.

### DIFF
--- a/data/transit/railway/station.json
+++ b/data/transit/railway/station.json
@@ -1,0 +1,26 @@
+{
+    "properties": {
+      "path": "transit/railway/station",
+      "skipCollection": true,
+      "exclude": {
+        "generic": ["^railway\\s?(station)?$"]
+      }
+    },
+    "items": [
+        {
+            "displayName": "LIRR",
+            "id": "lirr-94d192",
+            "locationSet": {
+              "include": ["us-ny.geojson"]
+            },
+            "tags": {
+              "network": "LIRR",
+              "network:wikidata": "Q125943",
+              "operator": "Long Island Rail Road",
+              "operator:wikidata": "Q125943",
+              "route": "train"
+            }
+          }
+    ]
+  }
+  
--- a/data/transit/railway/station.json
+++ b/data/transit/railway/station.json
@@ -18,7 +18,7 @@
               "network:wikidata": "Q125943",
               "operator": "Long Island Rail Road",
               "operator:wikidata": "Q125943",
-              "route": "train"
+              "railway": "station"
             }
           }
     ]


data/transit/railway/station.json
```diff
@@ -0,0 +1,26 @@
...
+      "path": "transit/railway/station",
+      "skipCollection": true,
+      "exclude": {
+        "generic": ["^railway\\s?(station)?$"]
+      }
+    },
+    "items": [
+        {
+            "displayName": "LIRR",
+            "id": "lirr-94d192",
```
akadouri, line 12: I copied this template from https://github.com/osmlab/name-suggestion-index/blob/main/data/transit/route/train.json#L831, I'm not sure how you pick IDs.

arch0345:
Relates to #4765

akadouri:
Oh interesting, I didn't know about the "two key value" identifier problem. I assumed in this case `railway=station` is enough to identify these features, they do [sometimes](https://taginfo.openstreetmap.org/tags/railway=station#combinations) have `subway=yes` or `train=yes` but that doesn't really change their network/operator suggestions.

bhousel:
> Oh interesting, I didn't know about the "two key value" identifier problem. I assumed in this case `railway=station` is enough to identify these features, they do [sometimes](https://taginfo.openstreetmap.org/tags/railway=station#combinations) have `subway=yes` or `train=yes` but that doesn't really change their network/operator suggestions

I tend to agree.. #4765 was a while ago and I do think OSM is maybe taking a step back from PTv2 a bit.
A while ago there were people who wanted to deprecate tags like `railway=station` and `highway=bus_stop` - these days not so much.

`railway=station` probably is enough to identify these features.

Rather than accept this PR I'm going to add a template category for the `railway=station` presets.

Robot8A:
Hi, should `railway=station` also be filled with `route=tracks` or `route=railway`?
In the European Union, the law mandates the separate management of tracks and train operations. However, for the station themselves, who gets to operate them differs by country:
 - In the Netherlands, they are operated by the train operator (https://openrailwaymap.app/#view=7.33/52.133/5.025&style=operator)
 - In Italy, they are operated by the railway tracks operator (https://openrailwaymap.app/#view=7.4/45.029/10.331&style=operator)
 - In Spain, they are operated by the tracks operator, except on commuter lines, where the train operator does it (https://openrailwaymap.app/#view=7.73/41.662/1.598&style=operator)

PR #7578, opened 2022-12-25
Removed name tag for IGA (Australia)

Scottie0001:
IGA stores across Australia trade under many different names however they are all part of IGA (Australia).

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -3175,7 +3175,6 @@
       "tags": {
         "brand": "IGA",
         "brand:wikidata": "Q5970945",
-        "name": "IGA",
         "shop": "supermarket"
       }
     },
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -3172,9 +3172,11 @@
       "displayName": "IGA (Australia)",
       "id": "iga-c254c9",
       "locationSet": {"include": ["au"]},
+      "preserveTags": ["^name"],
       "tags": {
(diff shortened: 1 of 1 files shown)


data/brands/shop/supermarket.json
```diff
@@ -3175,7 +3175,6 @@
       "tags": {
         "brand": "IGA",
         "brand:wikidata": "Q5970945",
-        "name": "IGA",
```
1ec5, line 3178: This will result in new POIs with this brand having no name by default. Is it incorrect or only suboptimal to tag it with a name of “IGA”?

To keep “IGA” as the default name but allow the user to easily modify an individual location’s name, keep the `name` but add a `preserveTags` to the entry instead:

https://github.com/osmlab/name-suggestion-index/blob/8dcc09b2265644e230781b8c60b6a713985f09f9/data/brands/shop/supermarket.json#L850

This will keep the Name field unlocked and suppress the warning that currently occurs when you change the name away from “IGA” to something more specific. This change would be equally applicable to IGA in the United States: https://github.com/osmlab/name-suggestion-index/issues/5500#issuecomment-943095957.

Scottie0001: That is the intended result.

1ec5: > This change would be equally applicable to IGA in the United States: https://github.com/osmlab/name-suggestion-index/issues/5500#issuecomment-943095957.

dda2181cd3ab593016f43953f0173fb6067c9379 applies the same to the IGA entry for North America.

PR #7580, opened 2022-12-26
update Shopee Xpress tags

iigmir:
Add Wikidata entry and i18n names for Shopee Xpress.


kjonosm:
Hi @iigmir. Thanks for your contribution!

PR #7586, opened 2022-12-27
Update public transport operator and network in Switzerland

DBrgr:
CH transit networks updated/cleaned up; npm run build

### DIFF
--- a/data/brands/shop/doityourself.json
+++ b/data/brands/shop/doityourself.json
@@ -11,6 +11,16 @@
   },
   "items": [
     {
+      "displayName": "McCoy's Building Supply",
+      "locationSet": {"include": ["us-tx.geojson", "us-nm.geojson", "us-ok.geojson"]},
+      "tags": {
+        "brand": "McCoy's",
+        "brand:wikidata": "Q27877295",
+        "name": "McCoy's",
+        "shop": "doityourself"
+      }
+    },
+    {
       "displayName": "Ace Hardware",
       "id": "acehardware-092aaf",
       "locationSet": {"include": ["001"]},


data/transit/route/train.json
```diff
@@ -2381,11 +2370,19 @@
...
+      "id": "zurcherverkehrsverbund-44bbb3",
+      "locationSet": {"include": ["001"]},
       "tags": {
         "network": "Zürcher Verkehrsverbund",
-        "network:short": "ZVV",
+        "route": "train"
+      }
+    },
+    {
+      "displayName": "ZVV",
```
LaoshuBaby, line 2381: I looked at de.wikipedia.org/wiki/Zürcher Verkehrsverbund and found the "Zürcher Verkehrsverbund" related wikidata is Q201977. And currently more than 1000+ OSM elements have a "network:wikipedia=de:Zürcher Verkehrsverbund", looks like it still in use.

Is there some different with them?

data/transit/route/tram.json
```diff
@@ -36,18 +36,6 @@
         "route": "tram"
       }
     },
-    {
-      "displayName": "Bernmobil",
```
LaoshuBaby, line 40: I search taginfo and found 1227 elements still have a `operator=Bernmobil`, so if you just remove it, in next weekly collection run, program will pick it out from planet.osm file again. 
Is this operator merged or bought by another operator? Or just changed to a new name?

data/transit/route/train.json
DBrgr: The common name is ZVV, we always used it this way, until iD users started changing the tags based the suggestion. I'm already cleaning up the current mess, it will take a bit.

data/transit/route/tram.json
DBrgr: Bernmobil is not an operator, it's a brand name. The operator is still SVB, which unfortunately got changed by iD users based on the suggestion. Also the lines are part of the 'Libero' network, but not all Libero lines are operated by SVB/Bernmobil.

DBrgr:
It would be customary to respond, The current setup is something between bloatware and spamming. Not sure who approved this, but it's NOT in line with longstanding tagging specifics from the OSM wiki (Switzerland:Public Transport), in particular the use of the official abbreviations, that are unique and clear across **all four official languages**.

PR #7593, opened 2022-12-29
Update Translink tagging for Queensland

Scottie0001:
Translink (Queensland) is the only Translink network in Queensland.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -15501,7 +15501,7 @@
       }
     },
     {
-      "displayName": "TransLink (Queensland)",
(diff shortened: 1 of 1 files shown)


LaoshuBaby (changes requested):
There are still 407 elements using Translink Cairns and 166 Translink SEQ

https://taginfo.openstreetmap.org/tags/network=Translink%20Cairns
https://taginfo.openstreetmap.org/tags/network=Translink%20SEQ

bhousel:
I'm going to just merge this and add matchnames for the other names.

PR #7599, opened 2022-12-30
Update school.json

DianAgesson:
The Victorian Department of Education and Training is changing names.

### DIFF
--- a/data/operators/amenity/school.json
+++ b/data/operators/amenity/school.json
@@ -7313,7 +7313,7 @@
       }
     },
     {
-      "displayName": "Department of Education and Training (Victoria)",
+      "displayName": "Department of Education (Victoria)",
       "id": "victoriandepartmentofeducationandtraining-1fc2f5",
       "locationSet": {
         "include": ["au-vic.geojson"]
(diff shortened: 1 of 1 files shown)


data/operators/amenity/school.json
```diff
@@ -7313,7 +7313,7 @@
       }
     },
     {
-      "displayName": "Department of Education and Training (Victoria)",
+      "displayName": "Department of Education (Victoria)",
```
LaoshuBaby, line 7316: 图片

Sorry, but I haven't found this changes in their website, and I think you can add old name as a matchName

DianAgesson: Hey Laoshu,

The Department is changing its name as part of the Machinery of Government changes, and it will take effect on Jan 01 2023. https://vpsc.vic.gov.au/about-public-sector/machinery-of-government/

I figured if I submitted it a day early it would avoid problems when the changeover happens! :)

DianAgesson: @LaoshuBaby The Department has now updated the logo in its website, and the rest of the site will be progressively changed as the transition to the new name continues. (Similar to the other departments in the State)

Would it be possible to progress this change? I’m planning on doing a bulk edit and I don’t want to introduce the change and have iD suggest it needs to be changed back. :)

PR #7606, opened 2023-01-02
Update bus.json

alessandrocarmeli:


data/transit/route/bus.json
```diff
@@ -172,17 +181,6 @@
         "route": "bus"
       }
     },
-    {
-      "displayName": "AIM Mobilità",
```
LaoshuBaby, line 176: 图片

There are still 400+ elements using this nerwork name, it will be collected again if you only delete this entry. Did it change its name, was it bought by other company, or did it have to be removed for some other reason?

alessandrocarmeli: Sorry if I answer only now.
The AIM company managed a lot of public services in Vicenza (Italy) (managed garbage disposal, transport, cemetery services, street lamp management etc...), now the public transport part has dissolved from AIM and has been created the new company SVT

kjonosm:
Hi @alessandrocarmeli,
i fail to understand why the display name differs from the network name and the name of the wikidata item? Could you please double check?

PR #7608, opened 2023-01-02
Add McCoy's

Cj-Malone:
https://www.mccoys.com/stores


Cj-Malone:
I thought this might happen because it didn't work in https://location-conflation.com/, but I know some states are supported. What's easier, change this to all of the US, or support these states?

UKChris-osm:
US States currently use custom `.geojson` files and so each reference to a US State also needs to include the file extention.

Cj-Malone:
Thank you

PR #7612, opened 2023-01-03
Testing code commit, adding Network IDs for PTNA

DBrgr:
Test run of my tools setup. At first no changes, only a dry run using the release checklist, to rebuild the index.

Awaiting approval, addition/cleanup of three transit networks used by PTNA.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -30,6 +30,17 @@
   },
   "items": [
     {
+      "displayName": "Cousins Subs",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "amenity": "fast_food",
+        "brand": "Cousins Subs",
+        "brand:wikidata": "Q5178843",
+        "cuisine": "sandwich",
+        "name": "Cousins Subs"
+      }
+    },
+    {
       "displayName": "&pizza",
       "id": "andpizza-4d2ff4",
       "locationSet": {"include": ["us"]},


data/transit/route/bus.json
```diff
@@ -2750,17 +2750,23 @@
...
-        "operator": "Auto AG Uri",
-        "operator:short": "AAGU",
-        "operator:wikidata": "Q33058015",
-        "route": "bus"
-      }
+      "tags": {"network": "CH-UR", "route": "bus"}
+    },
+    {
+      "displayName": "CH-VS",
+      "id": "chvs-ba16da",
```
LaoshuBaby, line 2767: > At first no changes, only a dry run using the release checklist, to rebuild the index.

Hello, thank you for contributing to NSI, and this is not required when open a PR because this checklist is for maintainer when they publish a new npm version.

---

However, a entry without wikidata won't be suggested in editor, see:

https://github.com/osmlab/name-suggestion-index/wiki/Contributing#tags-required

LaoshuBaby (changes requested):

data/transit/route/bus.json
DBrgr: I see. I don't expect, that there will be a (valid) wikidata entry for those pseudo-networks. Should they be removed then? Because CH-UR has no wikidata entry and it's incorrect, that there's only one operator. Also KtGR (should actually be CH-GR) has no wikidata entry. Suggestions?

PR #7616, opened 2023-01-06
Add TfL's Legible London maps

rskedgell:
Street maps present at hundreds of locations throughout Greater London. 100 of these are already mapped with network=legible_london

### DIFF
--- a/data/brands/tourism/information.json
+++ b/data/brands/tourism/information.json
@@ -54,7 +54,6 @@
         "information": "map",
         "map_size": "local",
         "map_type": "street",
-        "network": "legible_london",
         "operator": "Transport for London",
         "operator:type": "government",
         "operator:wikidata": "Q682520",
--- a/data/brands/tourism/information.json
+++ b/data/brands/tourism/information.json
@@ -44,6 +44,23 @@
         "name": "i-SITE",
         "tourism": "information"
       }
+    },
+    {
+      "displayName": "Legible London map",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Thanks @rskedgell 

I'm not to sure about the `network` tag being part of a map, as I feel like in this instance it would be better to focus on `brand` or `operator`, as `network` (to me) implies each POI would be physically connected in some way, rather than just being static elements with a common feature, but perhaps other would disagree?

Do the maps reference one another at all? Such as "_you are are map 16, go this way to get to map 17 or this way for map 18_"?

If the `network` tag is fine, however, I'd probably write it in a way that's similar to how you have written the `brand` tag, without the underscore and including `network:wikidata`.

rskedgell:
The network tag is a legacy which I believe predates NSI and probably wouldn't be used if we started now. It was used for the maps in a similar way to the TfL/Santander Cycle Hire docking stations, which are tagged with network=tfl_cycle_hire (in NSI as lower case with underscores, along with other UK-based cycle hire stations). I have no particular attachment to the network tag here and am happy to drop it from the pull request, as I doubt any data consumers use it.

The maps don't reference each other in any obvious way. They have a 5 digit ref on the side, but that is for maintenance purposes. In general, the areas covered by the maps will overlap, particularly within central London.

UKChris-osm:
Thanks @rskedgell 👍 This should be available in the next release in a couple of days.

PR #7634, opened 2023-01-09
Update school.json

ffff-23:


arch0345:
Please note that entries without a `:wikidata` tag won't get generated as a preset. If this school operator doesn't have a Wikidata item yet, you can easily add one using [this guide](https://github.com/osmlab/name-suggestion-index/wiki/Editing-Wikidata).

PR #7638, opened 2023-01-10
Add T-Mobile US

Cj-Malone:
[T-Mobile US](https://www.wikidata.org/wiki/Q3511885) has a different QID to [T-Mobile International](https://www.wikidata.org/wiki/Q327634), so this introduces some complexity.

I don't have the drive to push for a OSM bulk edit of `brand:wikidata=Q327634` -> `brand:wikidata=Q3511885` in the US. What do we have to do to encourage iD users to update the tag?

### DIFF
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -703,12 +703,22 @@
       }
     },
     {
-      "displayName": "T-Mobile",
+      "displayName": "T-Mobile International",
       "id": "tmobile-ee5434",
-      "locationSet": {"include": ["001"]},
+      "locationSet": {"include": ["001"], "exclude": ["us"]},
       "tags": {
         "brand": "T-Mobile",
         "brand:wikidata": "Q327634",
+        "name": "T-Mobile",
+        "shop": "mobile_phone"
+      }
+    },
+    {
+      "displayName": "T-Mobile US",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "brand": "T-Mobile",
+        "brand:wikidata": "Q3511885",
         "name": "T-Mobile",
         "shop": "mobile_phone"
       }


UKChris-osm:
I'm slightly hesitant about this purely because from the [Overpass query](https://overpass-turbo.eu/?Q=%5Bout%3Ajson%5D%5Btimeout%3A100%5D%3B%0A(nwr%5B%22name%22%3D%22T-Mobile%22%5D%3B)%3B%0Aout%20center%3B%0A%0A%7B%7Bstyle%3A%0Anode%2C%0Away%2C%0Arelation%0A%7B%20color%3Ared%3B%20fill-color%3Ared%3B%20%7D%0Anode%5Bshop%3Dmobile_phone%5D%2C%0Away%5Bshop%3Dmobile_phone%5D%2C%0Arelation%5Bshop%3Dmobile_phone%5D%0A%7B%20color%3Ayellow%3B%20fill-color%3Ayellow%3B%20%7D%0Anode%5Bshop%3Dmobile_phone%5D%5Bbrand%3DT-Mobile%5D%5Bbrand%3Awikidata%3DQ327634%5D%2C%0Away%5Bshop%3Dmobile_phone%5D%5Bbrand%3DT-Mobile%5D%5Bbrand%3Awikidata%3DQ327634%5D%2C%0Arelation%5Bshop%3Dmobile_phone%5D%5Bbrand%3DT-Mobile%5D%5Bbrand%3Awikidata%3DQ327634%5D%0A%7B%20color%3Agreen%3B%20fill-color%3Agreen%3B%20%7D%0A%7D%7D&R), it seems almost all the mobile shops are tagged in a way that's suitable - i.e. with the `brand` _T-Mobile_. Do they brand themselves differently, such as "_T-Mobile US_"?

If no bulk edit was done, then the NSI would only suggest an edit when an iD (or other supporting editor) user clicks on any given US T-Mobile POI, and will be presented with the "_this looks like a brand_" message. I'm not sure if there are other tools available that use the NSI to auto scan and highlight possible suggestions.

Cj-Malone:
So I think there is a bit of a disconnect the `brand:wikidata` and the `brand` string. `brand:wikidata` usually points to the business entity, maybe they should all point a [brand](https://www.wikidata.org/wiki/Q431289), that is then connected to a company/companies in different regions, but that's not the state of NSI, or the state of micro mapping in Wikidata. This has already happened with Aldi vs Aldi North vs Aldi South, there is no "Aldi Süd" branding in GB, but that's the better legal entity, NSI changed to it, and then OSM did.

Continuing to link to T-Mobile International is clearly wrong, it's wrong technically, and the details are wrong. An OSM consumer would no longer be able to use Wikidata for other details, eg [website](https://www.wikidata.org/wiki/Q327634#P856) is already wrong, theoretically it could be social media accounts or even logos.

Maybe I spoke to quickly on the US bulk edit, I don't think they would have much drama with this kind of edit, compared to other regions. But this will happen more and more as Wikidata gets more detailed, I think we need a better solution. But for now I will do the edit, I've done them [before](https://www.openstreetmap.org/changeset/115908807) and the only feedback I got from that was gratitude.

The driving force between this is All The Places linked to the [US QID](https://github.com/alltheplaces/alltheplaces/blob/master/locations/spiders/tmobile_us.py#L25), this is then blocking the NSI matching and preventing us from getting the categories from NSI or comparing against OSM stats. I could try and convince ATP to use the wrong QID to match, but then we'd lose all the Wikidata attributes.

Cj-Malone:
I think you was being the correct amount of cautious. This decision needed to be documented, and OSM needs a solution. For now, I'll contact Talk US.

PR #5924, opened 2021-12-22
Move China Post from operator to brand

Jyunhou:
China Post should be a brand instead of an operator.

Each sub post offices named `China Post`(中国邮政), like the following photo. In Beijing, it named `China Post`, in Canton, it named `China Post`, everywhere it named `China Post`.

image

Different from other post offices, eg Chunghwa Post, in Taipei there is sub post offices named `Nangang Post Office`(南港郵局) in Nangang district ,and `Xinzhuang Post Office`(新莊郵局) in Xinzhuang district, New Taipei like the following photo, and more. So `Chunghwa Post` is an operator.

image

But `China Post` always named `China Post`, so it should be a brand.

Issue #5625, opened 2021-11-01
Can we allow categories to be in both 'operators' and 'brands' trees


bhousel:
This is affecting kind of a lot of things now, and I need to think if I can just make some minor adjustments - but not a whole rewrite - to make it work ok.

> It seems to be working - I made sure to add `skipCollection` to the one in the brands tree and the ones in the brands tree won't also get added to the operators one as a city, county, or NGO operates the individual libraries

I'm just thinking out loud about the things that might be an issue:
- [x] Revert #5613 if it turns out I can pull this off with minimal effort.
- [x] The NSI matcher code probably will have subtle issues, because its match index is built on `k/v`.  This might mean the items and the exclude lists from one file may overwrite those things from other file.  I expect that this code will just never match stuff in whatever file it parsed first.
- [x] Raise a warning (by `matcher.buildMatchIndex`?) if a same-named item exists in several trees. re [#5625-comment]( https://github.com/osmlab/name-suggestion-index/issues/5625#issuecomment-956513420)
- [x] In iD there is other code to try to "guess" what tree a `k/v` pair belongs to.  (This code might guess wrong, but it might be ok anyway because brands and operators are similar - I dont know).  The reason for this "tree guessing" is because iD is just looking at raw OSM tags and it needs to decide which tags are considered "namelike".  
UPDATE: I checked the [code here](https://github.com/openstreetmap/iD/blob/3e7fd6e0330a14319206c9c9eb47afb3d97ca8d9/modules/services/nsi.js#L288-L299) and I _think_ it's ok.  Any items that appear in both trees will still be checked for `name`, `operator`, and `brand` keys.
- For example, the iD validator sees a `shop=yes`+`name:en=USPS - Morristown`+`height=3.5`+`roof:colour=orange`, it should be able to run the `name:en` tag through the matcher and guess that that is probably a US Post Office, even without actually seeing a defining tag like `amenity=post_office`, and also not matching the "Orange" mobile phone brand because of that `roof:colour` tag..

This probably just needs to be its own issue now, because the 'brand'-'operator' split is causing a whole lot of grief for me lately, and I _just_ moved the post offices in #5613 (mostly for Norway, UK, and things like UPS store) and I'm still not sure I did the right thing.

_Originally posted by @bhousel in https://github.com/osmlab/name-suggestion-index/issues/5623#issuecomment-956279182_

bhousel:
[...]

PR #7643, opened 2023-01-11
Add some Dutch brands, add some missing countries to existing ones

RobinLinde:
Adds the following brands (with Wikidata):
- AnyTyme
- Sympany
- Amslod
- Fietsenwinkel.nl
- Fietswereld
- Mastermate

### DIFF
--- a/data/brands/amenity/recycling.json
+++ b/data/brands/amenity/recycling.json
@@ -260,6 +260,22 @@
       }
     },
     {
+      "displayName": "Sympany",
+      "id": "sympany-dfbbb3",
+      "locationSet": {"include": ["nl"]},
+      "tags": {
+        "amenity": "recycling",
+        "brand": "Sympany",
+        "brand:wikidata": "Q2646891",
+        "name": "Sympany",
+        "operator": "Sympany",
+        "operator:wikidata": "Q2646891",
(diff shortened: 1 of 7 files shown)


data/brands/amenity/recycling.json
```diff
@@ -259,6 +259,22 @@
...
+    {
+      "displayName": "Sympany",
+      "id": "sympany-dfbbb3",
+      "locationSet": {"include": ["nl"]},
+      "tags": {
+        "amenity": "recycling",
+        "brand": "Sympany",
+        "brand:wikidata": "Q2646891",
+        "name": "Sympany",
+        "operator": "Sympany",
```
LaoshuBaby, line 271: I notice you move this entry from operator tree to brand tree, I don't means I request a change, but just don't know which is better because I'm not a local mapper.

You can see those 2 comments that clarify which is brand and which is operator:
+ https://github.com/osmlab/name-suggestion-index/issues/5625#issuecomment-958218478
+ https://github.com/osmlab/name-suggestion-index/pull/5924#issuecomment-999635804

I have no problem with other entry.

RobinLinde: I've moved it back to operators, as that's probably the most import value, but still with some more details

PR #7650, opened 2023-01-13
Add Belgian Toolstation branches

mstsbe:
The DIY-chain Toolstation has also branches in Belgium. I've added "be" to the localeSet. Please merge them into NSI.


Cj-Malone:
Thanks

PR #7651, opened 2023-01-14
Add African Bank

arrival-spring:
I assume it gets automatically sorted into alphabetical order, so just added near the bottom.


UKChris-osm:
Thanks @arrival-spring 👍

PR #7654, opened 2023-01-15
Add SQDC to cannabis.json

LeoColman:


LeoColman:
The SQDC is more known by the short name, so I added the full name as alt text.

<details>
<summary>Image of a SQDC store</summary>

image

</details>


I'm not sure if this should be added as a brand or as an operator. As per https://www.wikidata.org/wiki/Q43255076 I got very confused. The stores are operated by the government, but it's also a brand.

LeoColman:
My idea with this pull request is to make it easier to map the SQDCs, as they're available only on Google Maps.
One of them became available as per my edit https://www.openstreetmap.org/node/10553244014 and I see that this is pretty much the preset for all SQDCs.

arch0345:
Thanks @LeoColman and welcome!

PR #7662, opened 2023-01-17
Update Wikdiata for ESB Networks

cmap99:
Updated Wikidata QID for ESB Networks in route/power which was missed in PR #7645


Cj-Malone:
Thank you.

PR #7663, opened 2023-01-17
Fix ja-Hira and ja-Latn and Key

ffff-23:
According to the OSM Wiki page for [name:ja-Hira](https://wiki.openstreetmap.org/wiki/JA:Key:name:ja-Hira) and [Multilingual names](https://wiki.openstreetmap.org/wiki/Multilingual_names#Japan), it is written that name:ja-Hira describes hiragana ([Wikipedia](https://en.wikipedia.org/wiki/Hiragana)), so I changed katakana ([Wikipedia](https://en.wikipedia.org/wiki/Katakana)) to hiragana.

Also, for name:ja-Latn ([OSM Wiki](https://wiki.openstreetmap.org/wiki/Key:name:ja-Latn)), there were parts that were in English instead of rōmaji([Wikipedia](https://en.wikipedia.org/wiki/Romanization_of_Japanese)), so I changed English to rōmaji.

### DIFF
--- a/data/brands/amenity/language_school.json
+++ b/data/brands/amenity/language_school.json
@@ -145,7 +145,6 @@
       "locationSet": {"include": ["jp"]},
       "tags": {
         "amenity": "language_school",
-        "alt_name:ja": "ガバ",
         "brand": "GABA",
         "brand:en": "Gaba",
         "brand:ja": "GABA",
@@ -165,9 +164,9 @@
       "id": "nova-411869",
       "locationSet": {"include": ["jp"]},
       "tags": {
-        "amenity": "language_school",
         "alt_name:ja": "ノヴァ",
         "alt_name:ja-Latn": "Nobua",
+        "amenity": "language_school",
         "brand": "NOVA",
         "brand:en": "Nova",
         "brand:ja": "NOVA",
--- a/data/brands/amenity/language_school.json
+++ b/data/brands/amenity/language_school.json
@@ -249,14 +249,14 @@
         "brand": "セイハ英語学院",
         "brand:en": "Seiha English Academy",
         "brand:ja": "セイハ英語学院",
-        "brand:ja-Hani": "せいはえいごがくいん",
+        "brand:ja-Hira": "せいはえいごがくいん",
         "brand:ja-Latn": "Seiha Eigo Gakuin",
         "brand:wikidata": "Q7446694",
         "language:en": "main",
         "name": "セイハ英語学院",
         "name:en": "Seiha English Academy",
         "name:ja": "セイハ英語学院",
-        "name:ja-Hani": "せいはえいごがくいん",
+        "name:ja-Hira": "せいはえいごがくいん",
         "name:ja-Latn": "Seiha Eigo Gakuin"
       }
     },
--- a/data/brands/amenity/language_school.json
+++ b/data/brands/amenity/language_school.json
@@ -12,15 +12,15 @@
       "tags": {
         "amenity": "language_school",
         "brand": "AEON",
-        "brand:en": "Aeon",
-        "brand:ja": "AEON",
+        "brand:en": "AEON",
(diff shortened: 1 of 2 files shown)


data/brands/amenity/language_school.json
```diff
@@ -14,15 +14,15 @@
         "brand": "AEON",
         "brand:en": "Aeon",
         "brand:ja": "AEON",
-        "brand:ja-Hira": "イーオン",
```
1ec5, line 17: Would it be appropriate to preserve this name in `brand:ja` or `brand:ja-Kana`?

data/brands/amenity/language_school.json
```diff
@@ -14,15 +14,15 @@
...
-        "brand:ja-Latn": "AEON",
```
1ec5, line 18: Would it be appropriate to preserve this name in `brand:en`?

data/brands/amenity/language_school.json
ffff-23: Yes, I think brand:ja is fine. should I change `name:ja` to `イーオン` as well?

LaoshuBaby (approved):
Although I have a question that `name/brand:en` should be `AEON` or `Aeon` because it looks like [JA:Naming sample](https://wiki.openstreetmap.org/wiki/JA:Naming_sample) use first letter uppercase English words as English name.

Based on my little understanding of Japanese, I think this PR can emphasize that Hira and Latn are different kinds of transliterated scripts in BCP47.

Lonerat (approved):
I agree with this PR, but FYI below, JA:Naming sample seems to emphasize the use of the logo notation for the brand, mainly according to past Talk-ja mailing list discussions.

However, for `name:en`, someone changed most of the brands to capitalize only the first letter, but it is not clear if there was any agreement.

The English spelling of many Japanese brands in NSI is capitalized only at the beginning, regardless of Key:brand or name, and this has almost become the standard in Japan as a result of many editors using it.

For the use of ja-Latn, it seems to be ambiguous in many respects.
For example, an interchange on a Japanese expressway is commonly expressed as xx IC, which reads Intāchenji.
However, in most places, ja-Latn is tagged xx IC instead of xx Intāchenji, which is effectively the same as name:en, except with or without macrons.

Personally, I think it should be Intāchenji as this PR way, but I think it is still ambiguous because many people put it in as in the previous example.
Therefore, there does not seem to be much of a clear distinction between "romanization" and "English notation", and it largely depends on the person.

In addition, there are various types of romanization, such as the Hepburn system and the Kunrei system, but it has not been decided which system should be used.
Recent signboards seem to indicate that the Hepburn system or its derivatives are the most common, but in the end it depends on the person, so it is still difficult to determine.

LaoshuBaby:
> In addition, there are various types of romanization, such as the Hepburn system and the Kunrei system, but it has not been decided which system should be used.

Yes, I've heard of this, but I don't know what is most used in Japan, so I think that's what the local mapper chooses, and I search taginfo and found no tag can indicate which romanization system used. (probably not important?)

Lonerat:
> Yes, I've heard of this, but I don't know what is most used in Japan, so I think that's what the local mapper chooses, and I search taginfo and found no tag can indicate which romanization system used. (probably not important?)

I agree, I think it is not as important as others (since Hepburn system seems to be more of a de facto standard), but it is a problem because it is difficult to read the tags mechanically, and even if judging by localism, there are not always where more than one type is used locally, so it is impossible to decide which notation should take priority.

For example, "群馬" (ぐんま) is "Gunma" in Kunrei-shiki or Nihon-shiki, but "Gumma" in Hepburn.
The sound "ち" is also sparsely written with "ti" and "chi" in many places.

Also, there is no single "Hepburn", "Kunrei-shiki", or "Nihon-shiki" method, and the large number of derivative methods is one of the reasons why this issue has become a stalemate.

PR #7666, opened 2023-01-18
Add UK operators for man_made=mast and man_made=tower

russss:
This adds common operators for man_made=mast in the UK. It also adds a JSON file for man_made=tower which imports the operators from man_made=mast.

### DIFF
--- a/data/operators/man_made/mast.json
+++ b/data/operators/man_made/mast.json
@@ -19,6 +19,122 @@
         "tower:construction": "freestanding",
         "tower:type": "lighting"
       }
+    },
+    {
+      "displayName": "Arqiva",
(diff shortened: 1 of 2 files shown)


UKChris-osm:
I think it would be better to replace "_BT_" with [Openreach](https://en.wikipedia.org/wiki/Openreach) as I believe they deal with the majority (if not all) of the networking side of things as a wholly owned subsidiary.

russss:
On second thoughts I'm going to remove BT for the moment, as it's unclear whether BT means formerly-Cellnet (!), currently-Openreach, or something else.

UKChris-osm:
Ah, I see now, you meant BT as in the mobile tower / mast networks, not the underground cables, my mistake.

PR #7670, opened 2023-01-20
Switzerland, public transit networks modified

DBrgr:
Common network names corrected; removed forced operators from transit networks, because every network consists of multiple operators, so don't force one.


kjonosm:
> removed forced operators from transit networks, because every network consists of multiple operators, so don't force one.

While this is certainly true for e.g. bus operators, could you point me to a case where there are several tram or trolleybus operators in the respective networks you want to adjust?

DBrgr:
In ZVV there is now AVA (tram), and even FB is technically a tram (switching to 'narrow gauge' outside the city), or there is VMCV in Vaud (trolleybus).

PR #7674, opened 2023-01-21
Add English name and Wikidata for Board of Education in Japan

Lonerat:
Added English names of Japanese school governing boards and Wikidata.

The duplicated operators such as "xx city" and "xx city board of education" are merged into "xx city board of education" and added to matchNames, since the number of operators already registered was larger for the "xx city board of education" type.

### DIFF
--- a/data/operators/amenity/school.json
+++ b/data/operators/amenity/school.json
@@ -29391,9 +29391,9 @@
     },
     {
       "displayName": "世田谷区教育委員会",
-      "id": "366cb3-557cea",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Great! I've noticed those city was collected to NSI but don't know choose which wikidata item is best, a Committee's or City's? So thank you for specify them

Lonerat:
Just in case anyone is wondering, the reason for prioritizing "Board of Education" is because they are the organizations which "directly" operate the schools, and as I wrote, they are basically larger than the number already registered, and so on.
Of course if there is a problem, I will change it.

PR #7685, opened 2023-01-24
Add TFG clothing brands

arrival-spring:


LaoshuBaby:
I random check some brand‘s website and found they truly have many branch store that haven’t been tagged on OSM, it a hard job for local mapper. Thank you for contributing your knowledge about this and hope this will help mapper in South Africa.

PR #7690, opened 2023-01-24
Add Foschini

arrival-spring:
I couldn't find a definitive list of countries, the website simply says "With over 270 stores in Southern Africa..." so I hope worldwide is ok?

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -38,6 +38,18 @@
       }
     },
     {
+      "displayName": "Foschini",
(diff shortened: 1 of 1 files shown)


Cj-Malone:
It's hard to say definitely because they have some bad data, but to me it looks like: BW, KE, LS, NA, SZ, ZA, ZM. NA and ZA are the common ones.

arrival-spring:
I've refined it to those for now then

Cj-Malone:
Thanks

PR #7702, opened 2023-01-25
Add Inver fuel and remove Topaz

cmap99:
Add Inver fuel stations in Ireland and Canada. Removed Topaz as has been replaced by Circle K


Cj-Malone:
Thank you

PR #7704, opened 2023-01-26
Add alt_name to Vkusno i tochka

linux2good:
Add some references to McDonald's in Vkusno i tochka brand.


data/brands/amenity/fast_food.json
```diff
@@ -8071,7 +8071,9 @@
         "name": "Вкусно — и точка",
         "name:en": "Vkusno i tochka",
         "name:ru": "Вкусно — и точка",
-        "takeaway": "yes"
+        "takeaway": "yes",
+        "alt_name": "Макдональдс",
```
LaoshuBaby, line 8075: I don't think add "McDonald's" as a alt_name is good idea, because "Vkusno i tochka" is another brand, the only connection between them is "Vkusno i tochka" operating a post-McDonald's divestment restaurant. 

If you want to let entry match current restaurant still named "McDonald's" but operated by "Vkusno i tochka", add it as matchNames.

LaoshuBaby (changes requested):

PR #7714, opened 2023-01-28
Remove matchNames from Inver

cmap99:
Setting the matchNames to Topaz in this situation is incorrect as Inver has not replaced Topaz it is its own separate brand. Topaz had been purchased by Circle K and they have rebranded old Topaz location to Circle K but I haven't added a matchName to Circle K as i'm not certain every location that was a Topaz is now a Circle K

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -2479,6 +2479,7 @@
       "displayName": "Inver",
       "id": "inver-f499fc",
       "locationSet": {"include": ["ca", "ie"]},
+      "note": "don't need to add topaz as matchnames, see #7714",
       "tags": {
         "amenity": "fuel",
         "brand": "Inver",
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -2479,7 +2479,6 @@
       "displayName": "Inver",
       "id": "inver-f499fc",
       "locationSet": {"include": ["ca", "ie"]},
-      "matchNames": ["topaz"],
       "tags": {
         "amenity": "fuel",
         "brand": "Inver",


LaoshuBaby:
I should apologize because I saw in taginfo that there are still a lot of Topaz and mistook this to be consistent with the rebranding scenario we often encounter 
:(

LaoshuBaby:
I search commit history and found it was added in #7702 

--
Currently a lot of elements in OSM data still named Topaz so weekly collection will let program create a empty Topaz entry everytime we run build script. @cmap99 

See: https://github.com/osmlab/name-suggestion-index/commit/bb082c6e367d9bcb2267f96a33daa99ccb0d2d98

图片

cmap99:
> I should apologize because I saw in taginfo that there are still a lot of Topaz and mistook this to be consistent with the rebranding scenario we often encounter :(

That's okay I could've explained the situation better, or made two separate PRs.

Inver is its own brand that's been around for a while but was never added to NSI.

Topaz was its own brand but had been purchased by Circle K, and the majority of Topaz Fuel Stations were rebranded under the Circle K brand (but not all of them).

I have been reviewing the remaining locations still tagged as Topaz and correcting them. 



> I search commit history and found it was added in #7702
> 
> -- Currently a lot of elements in OSM data still named Topaz so weekly collection will let program create a empty Topaz entry everytime we run build script. @cmap99
> 
> See: [bb082c6](https://github.com/osmlab/name-suggestion-index/commit/bb082c6e367d9bcb2267f96a33daa99ccb0d2d98)
> 
> 图片

@LaoshuBaby Is there a need for a new PR to removed this once I've corrected the situation and replaced the Topaz brand?

PR #7719, opened 2023-01-29
Add Clothing Junction

arrival-spring:


arrival-spring:
@UKChris-osm 
Similar to #7720 where did those other countries come from?

UKChris-osm:
When I went to the web site linked in the Wikidata and selected stores, they all showed up there.

arrival-spring:
The map had been blocked for me, so I didn't spot that at first.

Some poor data there, unfortunately. They seem to simply be doing a geocode on their addresses and it's failing a lot.

e.g. their Charlesville store, which has Cape Town in the address has somehow showed up in the state of Georgia in the US

UKChris-osm:
> Some poor data there, unfortunately. They seem to simply be doing a geocode on their addresses and it's failing a lot.

Ah, so all those outside of South Africa could be incorrect?

arrival-spring:
I didn't check every single one, but I'm quite sure they are all incorrect. The website and social media seem to only reference SA and a web search yields nothing for any other country.

PR #7720, opened 2023-01-29
Add Sleepmasters

arrival-spring:


arrival-spring:
@UKChris-osm 
I'm intrigued as to how you know the brand is present in Egpyt?
I can't see it on the website or mapped in OSM at all

UKChris-osm:
When I went to the web site linked in the Wikidata and selected stores, it showed up there.

arrival-spring:
Ah, that map had been blocked for me. That's an error in the data though as well

UKChris-osm:
> Ah, that map had been blocked for me. That's an error in the data though as well

So there really isn't a store in Egypt?

arrival-spring:
No store in Egypt. That one says it is in Pretoria and all of the categories on the left are the provinces in South Africa.

PR #7722, opened 2023-01-30
Fix Phone House wikidata

Cj-Malone:
[Phone House](https://www.wikidata.org/wiki/Q325113) and [Carphone Warehouse](https://www.wikidata.org/wiki/Q118046) have different Wikidata entries. The logos are pretty similar so I assume they are owned by the same people.

### DIFF
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -641,7 +641,7 @@
       "matchNames": ["the phone house"],
       "tags": {
         "brand": "Phone House",
-        "brand:wikidata": "Q118046",
+        "brand:wikidata": "Q325113",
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Depends how mappable a "shop in a shop" is. I tend to say if it has separate opening hours, contact details or tills it should be separate. I don't know if Carphone Warehouse fits that or not, AFAIK they are only in currys now.

UKChris-osm:
I'd agree with those aspects, I haven't been in a Currys for a while so I'm not sure how Carphone Warehouse is utilised in there.

Cj-Malone:
I'll try to remember to check one out next time I'm near

PR #7741, opened 2023-02-01
Moved "La Sirena" and "Superverd" from wrong category

blacx0:


UKChris-osm:
Thank you for this @blacx0 and for your first contribution, all looks good 👍

PR #7745, opened 2023-02-02
Add Irish convenience stores

cmap99:
- Add Maxol 
- Add Applegreen
- Change location set of Centra to `["ie", "gb-nir"]`


kjonosm:
Thanks @cmap99!

PR #7747, opened 2023-02-02
Krefel is a Belgian chain, not active in France or NL

mstsbe:
Krefel is a Belgian electronics chain, and not active in France nor the Netherlands. I guess the creator of this brand somehow confused the languages in which it operates (which are indeed Dutch and French) for the countries.

Correcting this by this patch.

### DIFF
--- a/data/brands/shop/electronics.json
+++ b/data/brands/shop/electronics.json
@@ -682,7 +682,7 @@
     {
       "displayName": "Krëfel",
       "id": "krefel-86b1a0",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Oh, this is a problem lasting for 3 years, its locationSet was generated automatically in https://github.com/osmlab/name-suggestion-index/blob/75c268de639d436dc97feab404d2083e1b47a2a4/brands/shop/electronics.json#L318-L327

Thank you for pointing this out!

PR #7749, opened 2023-02-03
Add more Irish brands

cmap99:
Added the following brands:
- O'Briens (Cafe)
- Carry Out (Alcohol)
- O'Briens Wine (Alcohol)
- DNG (Estate Agent)
- Sherry FitzGerald (Estate Agent)


UKChris-osm:
Thank you for all of these additions @cmap99 👍

PR #7751, opened 2023-02-04
Midwest brand additions

smichae:
Monical's Pizza, Hicksgas, and RP Lumber additions - in and around Illinois.  NPM Build hasn't been run yet to generate IDs


UKChris-osm:
Thank you for these @smichae 👍 

Just as an FYI, Hicksgas won't appear in the index as it currently doesn't have a Wikidata assigned to it.

PR #7752, opened 2023-02-04
Add South African health departments

arrival-spring:
PR so that the checks can go through before committing

### DIFF
--- a/data/operators/amenity/hospital.json
+++ b/data/operators/amenity/hospital.json
@@ -27,6 +27,118 @@
   },
   "items": [
     {
+      "displayName": "KwaZulu-Natal Department of Health",
+      "locationSet": {"include": ["za"]},
+      "matchNames": [
+        "operator/KZN Department of Health",
+        "operator/KZN Health Department",
+        "operator/KwaZulu-Natal Health Department"
+      ],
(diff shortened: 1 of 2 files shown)


data/operators/amenity/hospital.json
```diff
@@ -26,6 +26,118 @@
...
+    {
+      "displayName": "Free State Department of Health",
+      "locationSet": {"include": ["za"]},
+      "matchNames": ["operator/Free State Health Department"],
+      "tags": {
+        "amenity": "hospital",
+        "healthcare": "hospital",
+        "operator": "Q116648781",
+        "operator:type": "government",
+        "operator:wikidata": "Free State Department of Health"
```
UKChris-osm, line 66: Is there a wikidata for "_Free State Department of Health_"?

arrival-spring: Oops, I'll fix that now

arrival-spring: Done, but should I have removed the id, as it now looks a bit odd?

UKChris-osm: The ID's are auto generated so you don't need to worry about those at all 👍

PR #7755, opened 2023-02-06
Update alcohol.json

neviksasha:
Add Ukrainian alcohol store


Cj-Malone:
I'm a bit unsure about the all caps, but I'm sure someone will change it if they feel strongly.

PR #7765, opened 2023-02-11
Update Japanese cafe, pharmacy, restaurant, and other brands

Lonerat:
Updated tagging and category changes for Japanese brands.


data/brands/amenity/pharmacy.json
```diff
@@ -4004,21 +4004,28 @@
...
         "brand:wikidata": "Q11304804",
+        "brand:zh-Hans": "札幌药妆",
+        "brand:zh-Hant": "札幌藥妝",
         "healthcare": "pharmacy",
         "name": "サツドラ",
-        "name:en": "Sapporo Drug Store",
-        "name:ja": "サツドラ"
+        "name:en": "Satudora",
+        "name:ja": "サツドラ",
+        "name:th": "ซัทสึโดระ",
```
LaoshuBaby, line 4026: Curious why you need to add Thai and Chinese here

LaoshuBaby, line 4026: Curious why you need to add Thai and Chinese here

data/brands/amenity/restaurant.json
```diff
@@ -5773,9 +5773,9 @@
       "locationSet": {"include": ["jp"]},
       "tags": {
         "amenity": "restaurant",
-        "brand": "ステーキガスト",
```
LaoshuBaby, line 5776: 图片

There are still a lot of name like this in current OSM data, if simple change brand name record, `ステーキガスト` will be collected to NSI in the future.

---
FYI: history of this entry in NSI: https://github.com/osmlab/name-suggestion-index/blame/02f82aa6da4f08369c85416b0eff7b73de92fa52/data/brands/amenity/restaurant.json#L5727-L5742

data/brands/amenity/restaurant.json
```diff
@@ -6146,22 +6146,6 @@
         "name:zh": "便宜坊"
       }
     },
-    {
-      "displayName": "元気寿司",
```
LaoshuBaby, line 6150: 图片

The same reason as ステーキガスト, if this 元気寿司 collapsed or rebranding?

LaoshuBaby (changes requested):

data/brands/amenity/pharmacy.json
Lonerat: This is because the official website has pages in these languages as tax free stores and I thought it would be useful to make them multilingual.
But I will remove if redundant.
https://satudora.jp/taxFree/zh-cn/

data/brands/amenity/restaurant.json
Lonerat: This is due to the duplication with the fast_food one.
https://github.com/osmlab/name-suggestion-index/blob/2a4a38a8a4423516275fb224be5c523a1b853908/data/brands/amenity/fast_food.json#L9634

data/brands/amenity/restaurant.json
Lonerat: This is because it is another variant of [ガスト](https://github.com/osmlab/name-suggestion-index/blob/02f82aa6da4f08369c85416b0eff7b73de92fa52/data/brands/amenity/restaurant.json#L5505), which is the main brand, and also because I referenced [JA:Naming sample](https://wiki.openstreetmap.org/wiki/JA:Naming_sample#amenity=restaurant).

data/brands/amenity/pharmacy.json
LaoshuBaby: I tend to remove because they are not commonly used languages ​​in the whole of Japan, like we don't need to add dozens of languages ​​​​to KFC at the same time, and I've done this.

Lonerat: I see, thank you for removing.

data/brands/amenity/pharmacy.json
```diff
@@ -4004,20 +4004,21 @@
...
       "displayName": "サツドラ",
-      "id": "sapporodrugstore-650eee",
+      "id": "satudora-650eee",
       "locationSet": {"include": ["jp"]},
       "tags": {
         "alt_name": "サッポロドラッグストアー",
-        "alt_name:en": "Satsudora",
+        "alt_name:en": "Sapporo Drug Store",
         "alt_name:ja": "サッポロドラッグストアー",
+        "alt_name:zh": "札幌薬粧",
```
ffff-23, line 4013: Chinese alt_name has also been added, should this be remove as well?

PR #7768, opened 2023-02-12
Add shizu/十足 store

strongwillow:
It's a common convenience store.

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -5811,7 +5811,6 @@
         "name": "十足",
         "name:en": "Shizu",
         "name:zh": "十足",
-        "website": "https://www.cnshizu.com",
         "shop": "convenience"
       }
     }
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -5808,6 +5808,7 @@
         "brand:en": "Shizu",
         "name": "十足",
         "name:en": "Shizu",
+        "website": "https://www.cnshizu.com",
         "shop": "convenience"
       }
     }
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -5806,8 +5806,10 @@
       "tags": {
         "brand": "十足",
         "brand:en": "Shizu",
+        "brand:zh": "十足",
         "name": "十足",
         "name:en": "Shizu",
+        "name:zh": "十足",
         "website": "https://www.cnshizu.com",
         "shop": "convenience"
       }
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -5798,6 +5798,18 @@
         "name:zh": "超市发",
         "shop": "convenience"
       }
+    },
+    {
+      "displayName": "十足",
+      "id": "shizu-zhejiang-cn",
+      "locationSet": {"include": ["cn"]},
+      "tags": {
+        "brand": "十足",
+        "brand:en": "Shizu",
+        "name": "十足",
+        "name:en": "Shizu",
+        "shop": "convenience"
+      }
     }
   ]
 }
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -5807,6 +5807,7 @@
         "brand": "十足",
         "brand:en": "Shizu",
         "brand:zh": "十足",
+        "brand:wikidata": "Q116763971",
         "name": "十足",
         "name:en": "Shizu",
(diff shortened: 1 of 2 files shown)


data/brands/shop/convenience.json
```diff
@@ -5798,6 +5798,18 @@
...
+    {
+      "displayName": "十足",
+      "id": "shizu-zhejiang-cn",
+      "locationSet": {"include": ["cn"]},
+      "tags": {
+        "brand": "十足",
+        "brand:en": "Shizu",
+        "name": "十足",
+        "name:en": "Shizu",
+        "shop": "convenience"
```
LaoshuBaby, line 5811: You need to have a wikidata item linked to this entry so that it can be suggested in editor

See: https://github.com/osmlab/name-suggestion-index/wiki/Contributing#tags-required

data/brands/shop/convenience.json
```diff
@@ -5798,6 +5798,18 @@
         "name:zh": "超市发",
         "shop": "convenience"
       }
+    },
+    {
+      "displayName": "十足",
+      "id": "shizu-zhejiang-cn",
```
LaoshuBaby, line 5804: It is a pity that we currently do not have the geojson of China's provinces, so we cannot distinguish it. Also you don't need to manually add this id, we'll run build script everytime publish the package.

data/brands/shop/convenience.json
```diff
@@ -5798,6 +5798,18 @@
...
+      "locationSet": {"include": ["cn"]},
+      "tags": {
+        "brand": "十足",
+        "brand:en": "Shizu",
```
LaoshuBaby, line 5808: https://wiki.openstreetmap.org/wiki/Multilingual_names#Repeating_name_with_language_specific_tag

It is recommend to repeat `brand:zh` as same as `brand`

LaoshuBaby (changes requested):

data/brands/shop/convenience.json
strongwillow: Hmm what if there is no wikipedia and Wikidata item for it? It's not on either of those.

data/brands/shop/convenience.json
strongwillow: Good to know. I wondered how to generate this id :)

data/brands/shop/convenience.json
strongwillow: Updated

data/brands/shop/convenience.json
LaoshuBaby: > Hmm what if there is no wikipedia and Wikidata item for it? It's not on either of those.

Create one if needed (This require a wikimedia account)

图片

You can add more information on it: https://www.wikidata.org/wiki/Q116763971

data/brands/shop/convenience.json
LaoshuBaby: > Good to know. I wondered how to generate this id :)

Clone the whole repository and run `npm install` in project root dictionary, and run `npm run build` every time you've done some changes.

You can see maintainer chapter of project wiki: https://github.com/osmlab/name-suggestion-index/wiki/Info-for-Maintainers

strongwillow: When I run `npm run build` many files changed. This pollutes my commit. I guess it's maintainers who should run it right?

LaoshuBaby: > When I run `npm run build` many files changed. This pollutes my commit. I guess it's maintainers who should run it right?

Yes, so I said you don't need to care about this 😃

data/brands/shop/convenience.json
```diff
@@ -5798,6 +5798,22 @@
...
+        "brand:zh": "十足",
+        "brand:wikidata": "Q116763971",
+        "name": "十足",
+        "name:en": "Shizu",
+        "name:zh": "十足",
+        "website": "https://www.cnshizu.com",
```
LaoshuBaby, line 5814: > Note for brands, that common tags with additional details such as "website" or "facebook" is not accepted directly. Instead a tag referencing the brand's [wikidata](https://www.wikidata.org/) entry ("[brand:wikidata](https://wiki.openstreetmap.org/wiki/Key:brand:wikidata)") containing those details can be included. If no such exists, consider [adding it](https://www.wikidata.org/wiki/Special:NewItem).

This was mentioned in https://github.com/osmlab/name-suggestion-index/wiki/Contributing#tags-required, I haven't found this before, sorry, and I think you can add it in wikidata as P856 (I've done)

Issue #7387, opened 2022-11-12
Name Suggestion Index 7 / NSI 7


UKChris-osm:
This issue is just to jot down some ideas for NSI version 7:

- An ability to create a master brand or operator, and the ability to assign it to multiple categories, so any change to the master entity only needs to be changed once (handy for local authority operators who appear in multiple categories).
- An ability to highlight the `name` tag as likely to be unique (such as a pub chain who names each pub uniquely but still shares a common brand) so that iD can highlight that "this name shouldn't be the same as the brand" or something to that effect. A step up from preserveTags.
- An NSI Bot that can connect to outside API's (such as Wikidata) and check for / pull / push data without the need for individuals to have their own account.
- Use of Github Actions / NSI Bot to display check results within a PR so the PR creator can see any potential errors right away, such as "the Wikidata submitted in this PR doesn't have a logo or social media accounts available, so the NSI can't display an icon".
- The planet scan could pull out any wikidata entries already assigned, and apply them if there is only one stand out QID.
- The planet scan could add a locationSet if all of the newly found entries only appear in a small selection of areas.
- The planet scan could create a new file that allows users to see which new entries were added during the last run.
- The NSI web site could display each entry on its own page, with a little more detail (such as last planet count, similar entries, etc), and a map, removing the need for the Overpass Turbo link. Similar entries could mean "This supermarket is also available as a convenience store" (like matchTags).

Dimitar5555:
> An ability to create a master brand or operator, and the ability to assign it to multiple categories, so any change to the master entity only needs to be changed once (handy for local authority operators who appear in multiple categories).

> The NSI web site could display each entry on its own page, with a little more detail (such as last planet count, similar entries, etc), and a map, removing the need for the Overpass Turbo link. Similar entries could mean "This supermarket is also available as a convenience store" (like matchTags).

> Ability to add branding without a specific top level tag.

Mixing up the above suggestions: each brand having it's own automatically generated page with all of it's NSI entries (including `operator` and `transit` of there are any).

Dimitar5555:
Another related issue: https://github.com/osmlab/name-suggestion-index/issues/3990

Cj-Malone:
- More details in `wikidata.json`, payment methods, cuisine/vegan/halal, labels for more languages.
    - Some kind of QA to highlight these to users, and encourage the Wikidata entries to be improved. Keep the labels and `brand:xx` in sync?
- Constant ids
[...]

PR #7772, opened 2023-02-13
Remove Things Remembered

Cj-Malone:
https://www.newsday.com/business/things-remembered-closing-stores-bankruptcy-gift-ad-populum-enesco-yedfu5ey
https://www.thingsremembered.com/

"moving into an exciting new chapter" == Firing all the staff and selling the brand to an online only retailer.

But the real reason for the PR: is adding the name to the excludes enough, or do we also have to flag the QID?

Bonus question, I only use JSOM, so I'm not that familiar with iD, but do dead brands make it to iD and the warn user? In GB we have a QA tool to [highlight](https://osm.mathmos.net/ghosts/) them, but obviously getting that into the editors is better.

### DIFF
--- a/data/brands/shop/gift.json
+++ b/data/brands/shop/gift.json
@@ -6,7 +6,10 @@
         "^(gift|souvenirs?)( shop)?$",
         "^(подарки|сувениры)$"
       ],
-      "named": ["^(papyrus)$"]
+      "named": [
+        "^(papyrus)$",
+        "^(things remembered)$"
+      ]
     }
   },
   "items": [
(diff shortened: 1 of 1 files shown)


UKChris-osm:
> But the real reason for the PR: is adding the name to the excludes enough, or do we also have to flag the QID?

Removing the entry is often enough, but adding the name to the exclusion list is needed if it currently appears [more than 50 times](https://github.com/ideditor/nsi-collector/blob/main/dist/osm/brands_all.json#L3926) in OSM data, otherwise the planet scan will keep re-adding it.

> 
> do dead brands make it to iD and the user warned?

I don't believe so at the moment, but it's something I would like to see occur, maybe in [NSI 7](https://github.com/osmlab/name-suggestion-index/issues/7387) :)

Cj-Malone:
> if it currently appears [more than 50 times](https://github.com/ideditor/nsi-collector/blob/main/dist/osm/brands_all.json#L3926) in OSM data

So just the name? We don't have to worry about `brand:wikidata` causing a new entry?

PR #7775, opened 2023-02-14
Use more specific brand:wikidata value for Martin's newsagents

rjw62:
The current wikidata entry https://www.wikidata.org/wiki/Q16997477 is actually for McColl's the owner of the chain. Martin's newsagents use a distinct fascia brand from other stores in the group. It would be better to use the more specific https://www.wikidata.org/wiki/Q116779207 .


rjw62:
> Makes sense to me, thank you. Will you update OSM? Or shall I?

@Cj-Malone Feel free to go ahead if/when you've got time. :)

PR #7782, opened 2023-02-17
Updates to Turkish shops and amenities

ddpasa:
This pull requests adds additional information (mostly websites) to Turkish shops and amenities.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -295,7 +295,8 @@
         "amenity": "bank",
         "brand": "Akbank",
         "brand:wikidata": "Q416624",
-        "name": "Akbank"
+        "name": "Akbank",
+        "website": "https://www.akbank.com/"
       }
     },
     {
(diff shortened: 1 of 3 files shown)


LaoshuBaby (changes requested):
I may reject this PR because those website may not been shared for different branch shop, although they are same brand, the `website` should reflect the current element's status.

We have a guide about this: https://github.com/osmlab/name-suggestion-index/wiki/Contributing#tags-required

So those `website` won't be merged into NSI. But, thank you for finding them! I suggest you add them in wikidata, using the "official website(P856)" property! If you haven't a wikimedia account you can ask for their community for help.

UKChris-osm:
Thank you for your first contribution @ddpasa, but as @LaoshuBaby has mentioned, we don't include a `brand` website in our core data, as generally the `website` tag refers to each location's own web page (such as shop.com/london or shop.com/paris) which is individualisation that we can't manage.

We also don't include `opening_hours` as these can differ between each location.

We tend to focus just on the name of the  `brand` / `operator` to ensure they are canonical across OpenStreetMap.

PR #7786, opened 2023-02-19
add various businesses

k-yle:
Caltex had disappeared for some reason. Plus some other additions


UKChris-osm:
Thanks for these @k-yle 👍

PR #7787, opened 2023-02-21
Display on nsi.guide if an entry is based on a template

UKChris-osm:
NSI Template


Cj-Malone:
Is it possible to link to the exact entry?

UKChris-osm:
> Is it possible to link to the exact entry?

Not that I could see from the data available in nsi.json, as there is currently only a value of true set for any item that comes from a template, that's why I added the check manually.

It might be possible to link into the search feature, such as linking to a search for "_name_" (such as Abcana) in the template link.

UKChris-osm:
Added a search link as well

NSI Template-2

LaoshuBaby (approved):
Great job, we've had meet users who want to edit directly in `atm.json` or `bus_stop.json` more than once, I used to do that too. This allows viewers to directly realize that these entry tags are inherited

PR #7789, opened 2023-02-22
Latin word "Novus" doesn't match charset for Ukrainian. Nevertheless Cyrillic name is not used on the ground

serhii-muchychka:


LaoshuBaby:
This name was added in  #7764

PR #7790, opened 2023-02-22
Add Büchereien Wien (Vienna public libraries)

ServusWorld:


UKChris-osm:
Thank you for your first contribution @ServusWorld  👍

PR #7791, opened 2023-02-22
Change displayNames in serbian to Latin and add multilanguage brand tags

u-spec-png:


u-spec-png:
I changed `displayName` because most people type in Latin alphabet, so when people type in brand name they can't find it so presets end up getting unused.

PR #7792, opened 2023-02-22
Change brand tag for ステーキガスト (Steak Gusto) to ガスト (Gusto)

Lonerat:
Steak Gusto is a variant brand of Gusto, and has many of the same menu items as Gusto, although there are some differences, such as specializing in griddle cuisine.
FYI:
Steak Gusto https://www.skylark.co.jp/en/steak_gusto/
Gusto https://www.skylark.co.jp/en/gusto/

Gusto is also used in the brand tag of Steak Gusto in JA:Naming sample.
https://wiki.openstreetmap.org/wiki/JA:Naming_sample#amenity=restaurant

But even if it were to be corrected, I don't know how to prevent adding the so called wrong one.

See also: #7765

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -5785,11 +5785,11 @@
       "locationSet": {"include": ["jp"]},
       "tags": {
         "amenity": "restaurant",
-        "brand": "ステーキガスト",
-        "brand:en": "Steak Gusto",
-        "brand:ja": "ステーキガスト",
-        "brand:wikidata": "Q92599119",
-        "cuisine": "steak",
+        "brand": "ガスト",
+        "brand:en": "Gusto",
+        "brand:ja": "ガスト",
+        "brand:wikidata": "Q87724117",
+        "cuisine": "steak;western;japanese",
         "name": "ステーキガスト",
         "name:en": "Steak Gusto",
         "name:ja": "ステーキガスト"


LaoshuBaby:
About those 2 brand I'd prefer to treat them as two brands with similar names but different, so I think the brand of Steak Gusto should be Steak Gusto, not Gusto

You said this PR meant to prevent adding the so called wrong one, can you describe what the editor suggest you to do on which kind of OSM elements? Thank you!

Lonerat:
Sorry for the confusion.
I mean, in Steak Gusto, if brand is "Gusto" as A and brand is "Steak Gusto" as B, I don't know how to avoid what you said in your previous PR, that "even if you correct to A, B will be added".

It may be recognizable as a different brand, but I think it is a case similar to this:
https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/shop/doityourself.json#L1544-L1587

LaoshuBaby:
> I don't know how to avoid what you said in the previous PR, that "even if you correct to A, B will be added".

I think there may be some misunderstanding here, if a restaurant already has brand=ガスト and brand:wikidata=Q87724117, it will not be corrected to ステーキガスト by the second suggestion, here is a video (14s) I just record:


https://user-images.githubusercontent.com/42690037/220708336-1efe1843-5704-4d6c-88a2-62ffc2a7e0cd.mp4




> It may be recognizable as a different brand, but I think it is a case similar to this:
> https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/shop/doityourself.json#L1544-L1587

If so, then we need to have a wider scope to discuss for those kind of brand series.

Lonerat:
> if a restaurant already has brand=ガスト and brand:wikidata=Q87724117, it will not be corrected to ステーキガスト by the second suggestion

Sorry again, I meant what you said in 7765 about:

> if simple change brand name record, `ステーキガスト` will be collected to NSI in the future.

Most people would put Steak Gusto in the name tag, so it would be expected behavior that a restaurant with the name tag Gusto would not be modified to Steak Gusto.


> If so, then we need to have a wider scope to discuss for those kind of brand series.

I agree. In the case of Komeri, for example, there are many instances where the tag is accidentally updated to the relatively minor Komeri PRO if it is updated without checking.

However, I think it is useful to have a unified brand tag when searches, etc., because I think it is possible to identify which type of business it is of the same brand by the name tag.

Also, although it is a little out of the category of brand, I think consistency with JA:Naming sample is also an issue.
JA:Naming sample is often inconsistent with NSI, which basically uses Japanese for Japanese brand, because the brand tag is more focused on the logo description.

Lonerat:
It is difficult to decide, but while Steak Gusto and Gusto may be perceived as different brands, at least for Komeri, there are multiple variants (or brands) within the Komeri brand, so it is natural that the brand tag is unified to Komeri.
FYI: https://www.komeri.bit.or.jp/company/business/

However, the same may be said for Steak Gusto.

LaoshuBaby:
Question 1

> if simple change brand name record, ステーキガスト will be collected to NSI in the future.

That is, if there isn't a `brand=ステーキガスト` in NSI but this tag appear many times in OSM data, in last week's routine maintain, program will found this “un-collected” brand and create a empty entry like this:

```
displayName: ステーキガスト 
locationSet: 001
tags: {amenity=restaurant + brand=ステーキガスト }(and only those 2)
```
Because at that time, NSI data only have a ガスト and another ガスト

That's why I don't suggest change name

Question 2

> Also, although it is a little out of the category of brand, I think consistency with JA:Naming sample is also an issue.
> JA:Naming sample is often inconsistent with NSI, which basically uses Japanese for Japanese brand, because the brand tag is more focused on the logo description.

Looks like this problem has been around since the massive start of maintaining Japanese brand information in NSI, We will not discuss this difference under this PR

Lonerat:
> Question 1

Sorry I'm not familiar with it, but can that happen even if the one with displayName "ステーキガスト" already exists?
When I add it to try and then run `npm run build`, it seems to fail with an error that the displayName is duplicated.

Lonerat:
Doesn't that mean that duplicates are detected and not collected, or they are collected and then deleted?
In other words, if the collector finds "brand=ステーキガスト" and tries to generate one with displayName "ステーキガスト", won't it stop collecting because it already exists, or detected as a duplicate and deleted during build?

bhousel:
> Doesn't that mean that duplicates are detected and not collected, or they are collected and then deleted?
> In other words, if the collector finds "brand=ステーキガスト" and tries to generate one with displayName "ステーキガスト", won't it stop collecting because it already exists, or detected as a duplicate and deleted during build?

DisplayName is something that exists separately from the tags.  It's really used to identify the item when it is displayed in lists like the JOSM menu or the Rapid preset picker.  If 2 items have the same DisplayName, users won't be able to know which one to pick.

Lonerat:
I mean, if the displayName of the "brand=ステーキガスト" will be "ステーキガスト" when collected by the collector, doesn't that mean that it will not be collected for duplication, if there already exists a item with displayName "ステーキガスト"?

Let A be the item already registered with a brand tag of ガスト and a name tag of ステーキガスト, and let B be the item collected by the collector with a brand tag of ステーキガスト and no name tag.

Sorry if I'm wrong, but my current understanding is that the displayName for both items will be ステーキガスト.

If this case is correct, B would have been detected as duplicates and removed/not collected, wouldn't it?

bhousel:
Gusto and Steak Gusto are already exising as 2 different items in the restaurant.json file.  
They are both scoped to match in Japan.
They have 2 different Wikidata , and on these pages I see they have different social media and menus and things.  
- Gusto:  https://www.wikidata.org/wiki/Q87724117
- Steak Gusto:  https://www.wikidata.org/wiki/Q92599119

So even though they may have the same parent company, these are different enough store formats that it's ok for there to be 2 of them.

Screenshot 2023-02-23 at 1 15 43 PM

LaoshuBaby:
I've read the comments above again, @Lonerat thinks that ステーキガスト and ガスト should be considered the same brand ガスト since they appear to be one series

> However, I think it is useful to have a unified brand tag when searches, etc., because I think it is possible to identify which type of business it is of the same brand by the name tag.
>
> In https://github.com/osmlab/name-suggestion-index/pull/7792#issuecomment-1440518471

I think the real commonality between them should be that they are all operated by skylark, The methodology of judging whether it is a brand by @Lonerat is based on whether the names of multiple brands are similar, and they are considered to be a series of brand. I must say that they do seem to have some unusual relationship, but I suggest to consider it based on the duck principle, that is, if at least skylark presents them as two brands, using different LOGO and menu, they seem to be two very similar entities - but two.

> If so, then we need to have a wider scope to discuss for those kind of brand series.
> 
> In https://github.com/osmlab/name-suggestion-index/pull/7792#issuecomment-1440475023

I think that if it is due to differences in the perspectives of what a brand is in different cultures, there may be more cases where multiple brands are regarded as the same brand value in NSI, shall introduce operator key to describe them appropriately? Because it is not recommended to introduce tags like `brand_series` at present, the relationship between ステーキガスト and ガスト is more suitable to be described by wikidata. And thanks for @bhousel pointing current fact.

Here is my idea about why I reasons for disapproving of this brand tagging approach from beginning to end.

Lonerat:
> @Lonerat thinks that ステーキガスト and ガスト should be considered the same brand ガスト since they appear to be one series

Rather than that, I am trying to do based on the original Japanese tagging practices, based on JA:Naming sample and other. Sometimes my opinion is different from them.
To be honest, I originally had some of the same thoughts as you. However I don't think that the original should be ignored for that.

The original Japanese tagging practice was to use the Steak Gusto brand tag as Gusto, but NSI changed the brand tag to Steak Gusto.
But even if each social account and wikidata has its own, that is not a reason to deny that it is a variant of Gusto.

For example, if Komeri has separate social accounts for each business category, or if Wikidata is registered for each business category, that would not necessarily be a reason to have separate brand tags.

Basically, the brand tag should rather indicate that it is a brand in the same family (not an operator).

Therefore, I thought a little more discussion is needed.
So I have some doubts about this being closed in the middle of the process.

It is also still not clear whether or not there is duplication by the collector.

Lonerat:
In your judgment, wouldn't the brand tag for most items only have the same value as name? If so, the brand tag is almost meaningless, and leaving it to Wikidata alone to describe brands in the same family would lose the meaning of the data within OSM.

Of course, I am not suggesting that the brand tag should necessarily be unified in the same family, but I do not think it makes sense for NSI to override the tagging on its own when it should have originally been done like this as the Japanese standard practice tagging.
In fact, I frequently see tagging that was practiced or de facto in Japan being overwritten by NSI Presets.

Lonerat:
As for one of the reasons why brand tags should be as uniform as the Komeri, to use an analogy with other brands, for example, there are other similar cases in Japan, such as "Big Boy" and "Steak Big Boy".
In such a case, I would at least make the brand tag "Big Boy" for both.
This is because this is just a difference in business type on the same brand.

The company may treat it as a separate brand, but the general perception is that this is just a variation of the "Big Boy" brand, so it should be appropriate for the brand tag to be so.

Other examples of those already registered with NSI are "Viva Home" and "Super Viva Home", although they are not restaurants.
These are also just different in size and type of business, but the brand name in the general perception is "Viva Home".
And the same could be said for Steak Gusto.

I apologize for the length, but I hope we can discuss this more.

bhousel:
It's hard for me to understand why this matters so much?  

In the NSI project we have examples of both styles of mapping.. Sometimes different store formats all share the same `brand`/`brand:wikidata` tags, and sometimes there is a different `brand`/`brand:wikidata` for each format.

In this case, there are 2 different Wikidata QIDs - Q87724117 and Q92599119.  I really think we should just follow that for the `brand:wikidata`, and `brand` is just a human readable tag that sits alongside it.

Cj-Malone:
I'll always vote for the most specific Wikidata entry. Any link between Steak Gusto and Gusto should be described there.

Some of my reasoning is mentioned in  https://github.com/osmlab/name-suggestion-index/pull/7638

Lonerat:
Because there are many such Japanese brands and I think this will continue to be a problem.
For example, if it is that way, there should be separate wikidata tags created for Viva Home and Super Viva Home, and the brand tag should be like that as well.

Also, the separate brand for Steak Gusto was originally created by NSI and was not used in Japan from the beginning.

I'm not saying it's wrong to use Wikidata to describe it, but I don't see any major reason to reject this, which should have been accepted in Japan originally?

Lonerat:
> In this case, there are 2 different Wikidata QIDs - Q87724117 and Q92599119. I really think we should just follow that for the brand:wikidata, and brand is just a human readable tag that sits alongside it.

So I disagree with that; Wikidata and OSM are essentially different datasets, and I don't think that's a reason to not necessarily describe them in OSM while ignoring the original tagging in Japan.
If all brands could be represented by wikidata for each variant, and there is such a consensus, then maybe they should, but I don't think that is the case yet.

If it is already that way, something like the cuisine tag, for example, would be registered in Wikidata, not in OSM (since it would not be necessary to register such brand information in OSM if it can all be read by Wikidata).

Lonerat:
Another reason is that it is frequently seen that registered brands of wikidata are in fact those of the operating company, or even incorrect. In such a current state, it would be difficult for data users to use wikidata, and it would be premature to make wikidata a standard at that level.

Initially, the brand:wikidata tag should had only been used to connect to wikidata. So if that tag is to be the standard even if it changes the original tagging, there has to be a clear consensus in each of the communities in the areas covered by NSI.

And with few people creating Wikidata for each brand, and no clear consensus, there are concerns about depending that much on Wikidata in this current situation.
It should not be until the creation of such detailed Wikidata and the use of them is more clearly established.

My compromise would be to leave brand:wikidata unchanged and just change the brand tag. However, this would normally not be recommended, as it is not the intended use of the tag and does not match what is actually stated in the wikidata.
But if the brand tag is just for human reading, then as a machine it should be able to simply read the wikidata and recognize that this is a variant of "gusto".

Sorry if this has been troublesome, but I thought this is a relatively important issue, so I thought it would be worth discussing.

Cj-Malone:
> If it is already that way, something like the cuisine tag, for example, would be registered in Wikidata, not in OSM (since it would not be necessary to register such brand information in OSM if it can all be read by Wikidata).

`cuisine` in OSM is a weird one, it's basically part of the category and so definitely belongs in NSI.

Other "cuisine"/diet details such as vegan/vegetarian/halal aren't really in NSI, but should be in Wikidata. Hopefully OSM data consumers will also use that and contribute back to all Open Datasets: OSM, Wikidata and NSI.

Lonerat:
Yes, that would be the ideal use, but in practice I don't think it's been accepted by either editors or data users (which is probably why the cuisine tag was created and continues to be used).
So the fact that it can't be expressed in OSM is problematic.

PR #7796, opened 2023-02-25
Kwantum is also active in Belgium, adding locationSet

mstsbe:
The Dutch furniture/house decoration chain Kwantum has also shops in Belgium (Flanders). Therefore also adding Belgium to its locationset.


Cj-Malone:
Thank you

PR #7799, opened 2023-02-26
Added feature files for Austria

haansn08:
source: https://osm-boundaries.com/
simplified with: https://mapshaper.org/


UKChris-osm:
Thank you for doing this @haansn08 👍

PR #7809, opened 2023-02-28
Adding chains from Colombia and Peru

whubsch:
Added a couple prominent fast food chains in Colombia and Peru and a Peruvian convenience store chain.

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -3829,6 +3829,17 @@
       }
     },
     {
+      "displayName": "Tambo",
+      "id": "tambo-853840",
+      "locationSet": {"include": ["pe"]},
+      "tags": {
+        "brand": "Tambo",
+        "brand:wikidata": "Q116369947",
+        "name": "Tambo",
+        "shop": "convenience"
+      }
+    },
+    {
       "displayName": "tạp hoá",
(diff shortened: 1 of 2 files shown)


data/brands/shop/convenience.json
```diff
@@ -3828,6 +3828,17 @@
...
       }
     },
+    {
+      "displayName": "Tambo",
+      "id": "tambo-853840",
+      "locationSet": {"include": ["pe"]},
+      "tags": {
+        "brand": "Tambo",
+        "brand:wikidata": "Q116369947",
+        "name": "Tambo",
```
tommylung, line 3838: Hi @whubsch , may I know the convenience's name is `Tambo` or `Tambo+` in Peru? Because there named `Tambo+` in both Wikidata and Wikipedia.

whubsch: Yeah good catch, it should be `Tambo+`, as long as that would still match with existing POIs already named just `Tambo`. Thanks.

LaoshuBaby: If so, you'de better add `Tambo` as a matchName, and I'll do this

tommylung: Thanks for your great help @LaoshuBaby

PR #7812, opened 2023-02-28
Changed wikidata from global to local subdivison

serhii-muchychka:
Local one contains actual contacts

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -8403,7 +8403,7 @@
         "brand:en": "McDonald's",
         "brand:ru": "МакДоналдс",
         "brand:uk": "МакДональдз",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/fast_food.json
```diff
@@ -8403,7 +8403,7 @@
         "brand:en": "McDonald's",
         "brand:ru": "МакДоналдс",
         "brand:uk": "МакДональдз",
-        "brand:wikidata": "Q38076",
+        "brand:wikidata": "Q4043856",
```
1ec5, line 8406: McDonald’s in Ukraine is a business unit, but is it a brand distinct from the global McDonald’s brand? Wikidata doesn’t seem to think so.

PR #7826, opened 2023-03-02
Remove Sガスト (S Gusto)

Lonerat:
All S Gusto are closed by 2020 and have been removed from the official website.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -6925,23 +6925,6 @@
       }
     },
     {
-      "displayName": "Sガスト",
-      "id": "sgusto-3e7699",
-      "locationSet": {"include": ["jp"]},
-      "tags": {
-        "amenity": "fast_food",
-        "brand": "ガスト",
-        "brand:en": "Gusto",
-        "brand:ja": "ガスト",
-        "brand:wikidata": "Q87724117",
-        "cuisine": "western;japanese",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
I've found those news report like: https://j-town.net/2020/07/08307656.html?p=all



---

However, I need to remind you that there are still a certain number of fast food restaurants named "Sガスト" in OSM. If you are an OSM mapper, maybe you can consider cleaning up these stores after this PR is merged. Thank you.

图片

LaoshuBaby:
> If you are an OSM mapper, **maybe you can consider cleaning up these shops after this PR is merged.** Thank you.

I've done this in:

+ https://www.openstreetmap.org/changeset/133214468
+ https://www.openstreetmap.org/changeset/133216871

If you have time can you review them? (I made those edit by streetview armchair mapping) and found that they were replaced by other skylark brand such as [とんから亭](https://www.skylark.co.jp/tonkara_tei/) or [から揚げ専門 から好し](https://www.skylark.co.jp/ja/karayoshi/)

Lonerat:
Thank you very much.

I think it is mostly ok. The only concern I have is that the street view is a Google thing?
If so, I don't think there is that much of a problem with chain restaurants like "Karayoshi", since it can be checked from their official website, but as for smaller stores and buildings, I think it should simply be removed since it cannot be checked from the valid source and could cause problems.

PR #7829, opened 2023-03-02
Update& Add LAWSON's

ffff-23:
Updated LAWSON's tag with reference to [JA:Naming sample](https://wiki.openstreetmap.org/wiki/JA:Naming_sample#shop=convenience).

For LAWSON+toks, '+' is also used for name because '+' is also used on the [website](http://www.e-toks.co.jp/business/convenience).

[LAWSON+Three F (Mapillary)](https://www.mapillary.com/app/?pKey=411364724039138)
[LAWSON+Poplar(Mapillary)](https://www.mapillary.com/app/?pKey=490777225316624)

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -5493,7 +5493,7 @@
       "locationSet": {"include": ["jp"]},
       "matchNames": ["ﾛｰｿﾝ"],
       "tags": {
-        "brand": "ローソン",
+        "brand": "LAWSON",
         "brand:en": "LAWSON",
         "brand:ja": "ローソン",
         "brand:wikidata": "Q1557223",
@@ -5507,17 +5507,14 @@
       "displayName": "ローソン・スリーエフ",
       "id": "lawsonthreef-652b3e",
       "locationSet": {"include": ["jp"]},
-      "matchNames": [
-        "lawson+スリーエフ",
-        "ﾛｰｿﾝ・スリーエフ"
-      ],
+      "matchNames": ["ﾛｰｿﾝ・スリーエフ"],
       "tags": {
-        "brand": "ローソン・スリーエフ",
-        "brand:en": "LAWSON･Three F",
-        "brand:ja": "ローソン・スリーエフ",
+        "brand": "LAWSON+スリーエフ",
+        "brand:en": "LAWSON+Three F",
+        "brand:ja": "ローソン+スリーエフ",
         "brand:wikidata": "Q24866804",
         "name": "ローソン・スリーエフ",
-        "name:en": "Lawson･Three F",
+        "name:en": "Lawson-Three F",
         "name:ja": "ローソン・スリーエフ",
         "shop": "convenience"
       }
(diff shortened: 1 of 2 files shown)


data/brands/shop/convenience.json
```diff
@@ -5537,6 +5549,23 @@
...
+      "id": "lawsontoks-652b3e",
+      "locationSet": {"include": ["jp"]},
+      "matchNames": ["ローソン+トークス"],
+      "note": "LAWSON+toks has two types of stores: convenience store and kiosk.",
+      "tags": {
+        "brand": "LAWSON+toks",
+        "brand:en": "LAWSON+toks",
+        "brand:ja": "ローソンプラストークス",
+        "brand:wikidata": "Q115868119",
+        "name": "ローソンプラストークス",
```
Lonerat, line 5563: Since "ローソン＋トークス" seems to be in common use [like this](https://www.rise.sc/shopguide/detail/?cd=000105), and the + is more like a symbol, so maybe this is a better one for the name tag?
```suggestion
      "note": "LAWSON+toks has two types of stores: convenience store and kiosk.",
      "tags": {
        "brand": "LAWSON+toks",
        "brand:en": "LAWSON+toks",
        "brand:ja": "ローソンプラストークス",
        "brand:wikidata": "Q115868119",
        "name": "ローソン+トークス",
```

data/brands/shop/kiosk.json
```diff
@@ -232,6 +232,23 @@
...
+      "id": "lawsontoks-84165a",
+      "locationSet": {"include": ["jp"]},
+      "matchNames": ["ローソン+トークス"],
+      "note": "LAWSON+toks has two types of stores: convenience store and kiosk.",
+      "tags": {
+        "brand": "LAWSON+toks",
+        "brand:en": "LAWSON+toks",
+        "brand:ja": "ローソンプラストークス",
+        "brand:wikidata": "Q115868119",
+        "name": "ローソンプラストークス",
```
Lonerat, line 247: Same as the convenience store version, I thought this one might be better.
```suggestion
      "note": "LAWSON+toks has two types of stores: convenience store and kiosk.",
      "tags": {
        "brand": "LAWSON+toks",
        "brand:en": "LAWSON+toks",
        "brand:ja": "ローソンプラストークス",
        "brand:wikidata": "Q115868119",
        "name": "ローソン+トークス",
```

data/brands/shop/convenience.json
```diff
@@ -5507,27 +5507,39 @@
...
-        "brand": "ローソン・スリーエフ",
-        "brand:en": "LAWSON･Three F",
-        "brand:ja": "ローソン・スリーエフ",
+        "brand": "LAWSON+スリーエフ",
+        "brand:en": "LAWSON+Three F",
+        "brand:ja": "ローソン+スリーエフ",
         "brand:wikidata": "Q24866804",
         "name": "ローソン・スリーエフ",
-        "name:en": "Lawson･Three F",
+        "name:en": "Lawson-Three F",
```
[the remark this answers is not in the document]
soprano1125: [JA:Naming sample](https://wiki.openstreetmap.org/wiki/JA:Naming_sample#shop=convenience) doesn't seem to be wrong.
If the value is correct, please correct JA:Naming sample.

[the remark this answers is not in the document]
Lonerat: Since it is a matter of notation, JA:Naming sample would not be particularly wrong.
But rather than using the - instead of the ・, I thought it should be matched what is on the official site.

data/brands/shop/convenience.json
ffff-23: I chose "ローソンプラストークス" because the official website used the notation "ローソンプラストークス". This suggestion may be better since there are also examples of the use of "ローソン+トークス".

PR #7836, opened 2023-03-04
Correct brand name of Norli book stores

eisams:
National Book Store is not the correct brand name, unclear where it came from

### DIFF
--- a/data/brands/shop/books.json
+++ b/data/brands/shop/books.json
@@ -656,7 +656,7 @@
       "note": "https://github.com/osmlab/name-suggestion-index/issues/5500",
       "preserveTags": ["^name"],
       "tags": {
-        "brand": "National Book Store",
+        "brand": "Norli",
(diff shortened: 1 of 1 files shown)


data/brands/shop/books.json
```diff
@@ -656,7 +656,7 @@
       "note": "https://github.com/osmlab/name-suggestion-index/issues/5500",
       "preserveTags": ["^name"],
       "tags": {
-        "brand": "National Book Store",
+        "brand": "Norli",
```
LaoshuBaby, line 659: The ”National Book Store“ was add as brand when ”Norli“ was collected to NSI： https://github.com/osmlab/name-suggestion-index/commit/8453824e8e91ce2bd1621292316b2913b29d5b50  (Nov 3, 2020)

But it does seem strange, because after checking the information, it is indeed unrelated to the government

PR #7839, opened 2023-03-04
Adding bicycle_rental=docking_station to Citi Bike

hdevine825:
Adding because all Citi Bike stations are docking stations. Not experienced with Github so please let me know if I did anything wrong.


Cj-Malone:
Looks good to me, thank you.

PR #7841, opened 2023-03-05
Update Japanese mobile stores

Lonerat:
Revised the Japanese mobile store brand tag to what is actually recognized as the brand and changed the English name to the one used in the official description.
The names of some stores were also changed to their official names as stores instead of brands.


Lonerat:
@LaoshuBaby 
If this is OK, could you please merge this?

PR #7842, opened 2023-03-06
Update notation of ヤマダ電機 (Yamada Denki) to ヤマダデンキ

Lonerat:
Around 2020, the "Denki" part of the Japanese trade name was changed from kanji to katakana.
However, it is also true that kanji notation continues to be commonly used, so I have described it in alt_name.

### DIFF
--- a/data/brands/shop/electronics.json
+++ b/data/brands/shop/electronics.json
@@ -1652,18 +1652,19 @@
       }
     },
     {
-      "displayName": "ヤマダ電機",
+      "displayName": "ヤマダデンキ",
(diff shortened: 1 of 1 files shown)


Lonerat:
@LaoshuBaby 
Could you please merge this if no problem?

LaoshuBaby:
Sorry for late, the kanji name was introduced by https://github.com/osmlab/name-suggestion-index/commit/a182f59550f4770a12933ab50a82e325716dd6f2 in 2020, thank you for share this local knowledge to make entry conform to ground truth.

PR #6180, opened 2022-02-09
Update Japanese Operators

blendy-map:
merged the item of Japanese operators.

PR #7843, opened 2023-03-06
Update Japanese Electric Power Companies

Lonerat:
Basically, Japanese electric power companies separate companies between generation and transmission lines, so I changed it accordingly.
FYI:
https://www.enecho.meti.go.jp/category/electricity_and_gas/electric/summary/electric_transmission_list/
https://www.enecho.meti.go.jp/category/electricity_and_gas/electricity_measures/004/list/

In addition, since the Japanese operator tag basically does not include "株式会社" and there are some short names that are actually commonly used, I changed it so that it does not include it.

### DIFF
--- a/data/operators/power/generator.json
+++ b/data/operators/power/generator.json
@@ -7930,12 +7930,12 @@
       "displayName": "中国電力",
       "id": "chugokuelectricpowercompany-11cc64",
       "locationSet": {"include": ["jp"]},
+      "matchNames": ["中国電力株式会社"],
       "note": "This 中国 don't means China",
       "tags": {
-        "operator": "中国電力株式会社",
+        "operator": "中国電力",
         "operator:en": "Chugoku Electric Power Company",
-        "operator:ja": "中国電力株式会社",
-        "operator:short": "中国電力",
+        "operator:ja": "中国電力",
         "operator:wikidata": "Q73171",
         "power": "generator"
       }
@@ -7944,11 +7944,12 @@
       "displayName": "中部電力",
       "id": "chubuelectricpowercompany-11cc64",
       "locationSet": {"include": ["jp"]},
+      "matchNames": ["中部電力株式会社"],
       "tags": {
-        "operator": "中部電力株式会社",
+        "operator": "中部電力",
         "operator:en": "Chubu Electric Power Company",
-        "operator:ja": "中部電力株式会社",
-        "operator:short": "中部電力",
+        "operator:ja": "中部電力",
+        "operator:short": "中電",
         "operator:wikidata": "Q1091203",
         "power": "generator"
       }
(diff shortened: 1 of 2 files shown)


UKChris-osm:
@LaoshuBaby I see you assigned yourself to this, were you planning on making some edits to this PR or does it need additional changes / checks?

LaoshuBaby:
Sorry for my delay, because this involves the question of whether the common name is short one or more legal and official one, I hesitated at the time, and then I gradually forgot about this PR when I dealt with other things. I am very sorry, and I will try my best to get rid of the procrastination problem. I tried my best to restore the points I thought at the time, as follows:

I search the kabushiki gaisha on taginfo and found almost all operator include it, see:

图片

Even if the ones involved in this PR are excluded, there are still 100+ pages of operators containing "株式会社", so I can guess that company name have those words is used more widely?

Another reason that I hasitite is those 2 website are truly official and their register name contain "株式会社"

(Besides, looks like those entry was first introduced by blendy-map in #6180, there are also other active Japan mapper contribute to NSI, perhaps we could consider hearing from other editors? However, if there have been other discussions before, just follow the results of the discussion.）

LaoshuBaby:
> However, there may be some problems with English names, such as "TEPCO" or "Tokyo Electric Power Company", which use different abbreviations or include the word "Company".
> So I would be grateful if someone could provide other opinions.

Coincidentally, there is currently a PR for the short name of the operator, #7954, and I am keep watching it now.

(Although it seems that most of them do not involve "company", after all, the suffixes are also related to different regions, And most of the entries involve regions dominated by Latin letters)

Lonerat:
> > However, there may be some problems with English names, such as "TEPCO" or "Tokyo Electric Power Company", which use different abbreviations or include the word "Company".
> > So I would be grateful if someone could provide other opinions.
> 
> Coincidentally, there is currently a PR for the short name of the operator, #7954, and I am keep watching it now.
> 
> (Although it seems that most of them do not involve "company", after all, the suffixes are also related to different regions, And most of the entries involve regions dominated by Latin letters)

Thank you for the information.

As for the short names, both "東電(Toden)" and "TEPCO" are used by the general public and by the company itself, so they are worth including as information. So I do not agree with that PR, at least if that is done in Japan.

However, the question is which one should be prioritized. For now, my opinion is Japanese, as it is the native language of the country to which the company belongs.
But if possible, I think it would be more convenient to list both.

Lonerat:
> Yes, this is a possible and customary practice, if have to choose one
> 
> BTW, here is a example from bilingual (Chinese + English) region, they don't keep the "company" at all:
> 
> https://github.com/osmlab/name-suggestion-index/blob/e681f59b37ff81c3d40e41a5913dede5f85db9f7/data/operators/power/substation.json#L10946-L10964

I see, thank you again for the information.

Well, I guess companies include "Company" or "Co." in their abbreviations because it sounds good and is easy to say, and I didn't see a problem with including it.

Issue #7850, opened 2023-03-08
Honest Greens in Spain and Portugal 


codeinabox:
Honest Greens has a presence in both Spain and Portugal https://honestgreens.com/

I've already created a Wikidata entry for it
https://www.wikidata.org/wiki/Q116859710

PR #7854, opened 2023-03-10
add honest greens

AlertSubject:
closes: https://github.com/osmlab/name-suggestion-index/issues/7850

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -2180,7 +2180,7 @@
     {
       "displayName": "Honest Greens",
       "id": "honestgreens-acf031",
-      "locationSet": {"include": ["es"]},
+      "locationSet": {"include": ["es", "pt"]},
       "tags": {
         "amenity": "restaurant",
         "brand": "Honest Greens",
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -2178,6 +2178,17 @@
       }
     },
     {
+      "displayName": "Honest Greens",
+      "id": "honestgreens-acf031",
+      "locationSet": {"include": ["es"]},
+      "tags": {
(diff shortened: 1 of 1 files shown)


data/brands/amenity/restaurant.json
```diff
@@ -2177,6 +2177,17 @@
         "name": "Honest Burgers"
       }
     },
+    {
+      "displayName": "Honest Greens",
+      "id": "honestgreens-acf031",
+      "locationSet": {"include": ["es"]},
```
Cj-Malone, line 2183:
```suggestion
      "locationSet": {"include": ["es", "pt"]},
```

@codeinabox mentioned Portugal.

codeinabox:
I'm not sure I'd tag this as a restaurant as you have to order at the counter and there is no table service

AlertSubject:
> I'm not sure I'd tag this as a restaurant as you have to order at the counter and there is no table service

They themselves call it restaurants so I went with that
https://honestgreens.com/en/our-restaurants-2/

It seems to not entirely match up with the description of restaurants that can be found in the osm Wiki
Fastfood seems more fitting maybe?
https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dfast_food

Is there no table service or no tables?

If tables are present then I think restaurant could still be a fitting tag even if there's no service

UKChris-osm:
The POI's currently in OSM seem to all be marked as `amenity=restaurant`, if that makes any difference?

codeinabox:
> The POI's currently in OSM seem to all be marked as `amenity=restaurant`, if that makes any difference?

In that case let's go with restaurant, though I think cafe would be equally as appropriate as they serve many breakfast items and you order at the counter and they bring it to your table

tas50:
Sounds like a fast casual restaurant which would be restaurant in OSM

PR #7858, opened 2023-03-11
Updated lifeguard.json

MrMicheeel:
Removed Brigada Militar (police service) and updated the existing police.json

### DIFF
--- a/data/operators/emergency/lifeguard.json
+++ b/data/operators/emergency/lifeguard.json
@@ -18,16 +18,6 @@
       }
     },
     {
-      "displayName": "Brigada Militar",
-      "id": "brigadamilitar-4f4d5b",
-      "locationSet": {"include": ["br"]},
-      "tags": {
-        "emergency": "lifeguard",
-        "operator": "Brigada Militar",
-        "operator:wikidata": "Q9666780"
-      }
-    },
-    {
       "displayName": "Corpo de Bombeiros Militar de Santa Catarina",
       "id": "corpodebombeirosmilitardesantacatarina-4f4d5b",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Brigada Militar appears a lot in OpenStreetMap as being tagged as lifeguard along the coast, are you sure they don't also operate the lifeguard posts in the area?

As they appear so often, deleting them means you also have to add an additional filter / instruction at the top of the file to stop them being re-added again. Because they appear more than 10 times as an operator of lifeguard posts, our scan thinks they are popular enough to be automatically added, so we have to override that.

MrMicheeel:
> Brigada Militar appears a lot in OpenStreetMap as being tagged as lifeguard along the coast, are you sure they don't also operate the lifeguard posts in the area?
> 
> As they appear so often, deleting them means you also have to add an additional filter / instruction at the top of the file to stop them being re-added again. Because they appear more than 10 times as an operator of lifeguard posts, our scan thinks they are popular enough to be automatically added, so we have to override that.

Years ago Brigada Militar did the service of lifeguard, but now its only the Corpo de Bombeiros Militar (do Rio Grande do Sul) that work in lifeguard posts. Should be added the Corpo de Bombeiros Militar as lifeguard then (wikidata Q10260737)

UKChris-osm:
In that case, to stop `Brigada Militar` being re-added, at the top of the file, you'll need to add it to the exclusion list, like so:


`  "properties": {`
`    "path": "operators/emergency/lifeguard",`
`    "exclude": {`
`      "generic": ["^lifeguard$"],`
**`      "named": ["^brigada militar$"]`**
`    }`
`  },`

PR #7859, opened 2023-03-11
Updated police.json

MrMicheeel:
Updated wikidata ID for Brigada Militar do Rio Grande do Sul


UKChris-osm:
Why did you close this @MrMicheeel?

PR #7868, opened 2023-03-13
Removed brand:name "Common Service Centres"

MG-LSJ:
Common Service Centres is a certification given by the Government of India and is not the name of the shop. Shops have their own names along with the "CSC" branding.


MG-LSJ:
> Hi @MG-LSJ, just to let you know that the Name Suggestion Index no longer includes `brand:wikipedia`

Noted. And also updated the image in the wiki.

PR #7873, opened 2023-03-14
Expand BreadTalk brand listing to Asia and Europe

LuisAPI:
The Singaporean brand operates in certain other Southeast Asian countries (Malaysia and Thailand), and then has franchises "all over Asia and Europe" as per [WP](https://en.wikipedia.org/wiki/BreadTalk#History).

> The Group has a network of owned bakery outlets in Singapore, PRC, Malaysia, Hong Kong and Thailand as well as franchised bakery outlets across Asia, Europe and the Middle East.

### DIFF
--- a/data/brands/shop/bakery.json
+++ b/data/brands/shop/bakery.json
@@ -417,7 +417,7 @@
     {
       "displayName": "BreadTalk",
       "id": "breadtalk-96a6b3",
-      "locationSet": {"include": ["sg"]},
+      "locationSet": {"include": ["142", "150"]},
       "tags": {
         "brand": "BreadTalk",
         "brand:wikidata": "Q1106640",


data/brands/shop/bakery.json
```diff
@@ -417,7 +417,7 @@
     {
       "displayName": "BreadTalk",
       "id": "breadtalk-96a6b3",
-      "locationSet": {"include": ["sg"]},
+      "locationSet": {"include": ["142", "150"]},
```
LaoshuBaby, line 420: Wikipedia usually lists all countries and regions where stores exist, but this does not mean that all of them must be added completely, because there may be only one or two scattered stores in some countries and regions, and the changes over time are great. It is not wise to include all of Europe and Asia, my suggestion is to add only a few regions where chain store networks are clearly present, such as Hong Kong, Mainland of China, Thailand, Malaysia and etc.

[图片](https://location-conflation.com/?locationSet=%7B+include%3A+%5B%27142%27%2C+%27150%27%5D+%7D%0A)

LaoshuBaby (changes requested):

data/brands/shop/bakery.json
matkoniecz, line 420: Note what is actually described there - from https://www.straitstimes.com/business/breadtalk-debuts-in-london-with-din-tai-fung-flagship

> Taiwanese restaurant Din Tai Fung opened in London's Covent Garden yesterday, as the BreadTalk Group marked its first foray into Europe.

BreadTalk Group opened something, that does NOT mean any BreadTalk bakeries are in Europe

Is there any confirmation of even single BreadTalk bakery in Europe? It seems that there are none in Poland, Google Maps also fails to find any.

matkoniecz (changes requested):

bhousel:
This PR is getting old and got some pushback, so I'm just going to close it. 
It's probably ok to just list out the countries where BreadTalk operates rather than just including all of Asia and Europe.

PR #7877, opened 2023-03-16
Add brands recently included in All The Places

davidhicks:
All The Places (ATP) is a project that consumes NSI data. Some spiders were added recently to ATP that did not have corresponding NSI brands. This pull request seeks to add additional brands to NSI to match the brands which have recently been added to ATP.


Cj-Malone:
LGTM, thank you

PR #7855, opened 2023-03-10
add Toby's Estate

AlertSubject:
closes https://github.com/osmlab/name-suggestion-index/issues/7838

PR #7883, opened 2023-03-18
Update cafe.json

AlertSubject:
for reference: https://github.com/osmlab/name-suggestion-index/pull/7855#issuecomment-1464538752


Cj-Malone:
Thank you

PR #7884, opened 2023-03-18
Add America Today as new clothing chain

mstsbe:
Please add the Dutch clothing chain America Today (it had already a Wikidata item, but was not very complete. So I added more information used in NSI).


UKChris-osm:
Thank you @mstsbe 👍

PR #7898, opened 2023-03-22
Add AbouThai (HK)

tommylung:
Add:
- AbouThai (HK)

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -5943,6 +5943,20 @@
         "name:zh": "超市发",
         "shop": "convenience"
       }
+    },
+    {
+      "displayName": "阿布泰 AbouThai",
+      "locationSet": {"include": ["hk"]},
+      "tags": {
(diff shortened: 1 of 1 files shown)


data/brands/shop/convenience.json
```diff
@@ -5943,6 +5943,20 @@
         "name:zh": "超市发",
         "shop": "convenience"
       }
+    },
+    {
+      "displayName": "阿布泰 AbouThai",
+      "locationSet": {"include": ["hk"]},
+      "tags": {
+        "brand": "阿布泰 AbouThai",
```
Cj-Malone, line 5951: Do we usually put both languages in the brand/name tags?

I can see they are both on the top left of the website, but personally I'd have split them up

bhousel: Hong Kong community does, and maybe a few others.
There's some more info on this ticket, and the stuff linked to it
https://github.com/osmlab/name-suggestion-index/issues/6788#issuecomment-1187755052

PR #7899, opened 2023-03-22
These name:* doesn't match charset & redundant

serhii-muchychka:


bhousel:
Offhand I can't tell why they are there either, but we should maybe `git blame` these lines to see why the original contributor might have included them?  

I don't have knowledge of Ukraine brands, and maybe the community there prefers there to always be a `:uk` and `:en` and `:ru` names that matches the signage and situation on the ground.

serhii-muchychka:
@bhousel  yesterday I asked people in https://uk.tgstat.com/chat/@osmua chat, how is better. All members who answered don't support inserting names in English/Latin names into **name:uk**. This tag must contain names in Cyrillic

Nevertheless, many POIs in Ukraine don't have names in Ukrainian on the ground, only English

PR #7903, opened 2023-03-22
shop/laundy: add Aramark Uniform and Cleanroom Services

davidhicks:


Cj-Malone:
NSI needs `brand:wikidata` now, otherwise it won't be published.

1ec5:
This is an industrial laundry service, as opposed to a consumer-facing retail laundry chain. #7903 proposed some similar chains, but we didn’t end up merging it because the tagging was not yet established. I’d be hesitant to stretch the meaning of `shop=laundry` and indeed the `shop=*` key without first checking with the global OSM community about whether there’s a preferred alternative, such as `industrial=laundry`.

davidhicks:
Thanks for the comments. I'll close this PR for the same reason pull request #4899 was closed.

PR #7915, opened 2023-03-25
Remove duplicate, update information

neviksasha:


data/brands/shop/bakery.json
```diff
@@ -2143,16 +2143,6 @@
...
-      "id": "e5a0f8-3664c0",
-      "locationSet": {"include": ["ua"]},
-      "tags": {
-        "brand": "Кулиничи",
-        "name": "Кулиничи",
-        "shop": "bakery"
-      }
-    },
     {
       "displayName": "Кулиничі",
```
Dimitar5555, line 2147: It would be good to add `"matchNames": ["Кулиничи"]` in order to match the ~150 objects with that name.

UKChris-osm:
@Dimitar5555 the name from the deleted entry should already match in this case as it's been added as a `name ` tag to another entry 👍

PR #7918, opened 2023-03-26
Create advertising.json

rskedgell:
Initial entry for Clear Channel UK

### DIFF
--- a/data/operators/man_made/advertising.json
+++ b/data/operators/man_made/advertising.json
@@ -1,0 +1,19 @@
+{
+  "properties": {
+    "path": "operators/man_made/advertising"
+  },
+  "items": [
+    {
+      "displayName": "Clear Channel",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
https://wiki.openstreetmap.org/wiki/Key:advertising reads:

> advertising=* tags were advocated to be accompanied by [man_made](https://wiki.openstreetmap.org/wiki/Key:man_made)=[advertising](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dadvertising), but this tag is not mandatory and is typically omitted.

Would this mean that using `man_made=advertising` is perhaps the wrong tag to be using?

https://overpass-turbo.eu/s/1sTP shows red markers for all advertising, and green for those that include  `man_made=advertising` which shows hardly any usage.

rskedgell:
Good point. I'll withdraw this pull request and create a new one under advertising/

PR #7919, opened 2023-03-26
Create advertising/billboard

rskedgell:
Create advertising/billboard with initial entry for Clear Channel UK


UKChris-osm:
Thanks @rskedgell 👍 

Just an FYI, a `matchNames` won't be needed if the only difference is capitalisation.

rskedgell:
Thanks! Noted re: matchNames.

PR #7920, opened 2023-03-26
Create advertising/poster_box

rskedgell:
Initial entry for JCDecaux street furniture/poster boxes.  I know they are present in more countries than FR and GB, but don't have a list.


UKChris-osm:
I'd suggest setting the location to global `001` if you're unsure about the locations, as it's something we can limit later if there is a clash.

rskedgell:
Would just dropping the locationSet line for now work?

UKChris-osm:
Unfortunately not, as it's a required setting.

PR #7925, opened 2023-03-27
Update Etam lingerie

lmagreault:
Replace "lingerie" by "underwear" for Etam lingerie like Aubade, Darjeeling, Dim.


Cj-Malone (approved):
Makes sense to me https://taginfo.openstreetmap.org/keys/clothes#values

UKChris-osm:
Seems roughly [50 / 50](https://overpass-turbo.eu/s/1sXw) at the moment.

PR #7930, opened 2023-03-30
Set location & add matchNames for Prostor

serhii-muchychka:


Cj-Malone:
Seems OK to me. Thank you.

PR #7937, opened 2023-03-31
Chanced Batopin to Bancontact CASH

JoeyKentin:
Batopin is the operator of Bancontact CASH not the brand


UKChris-osm:
Thank you for the update @JoeyKentin 👍

PR #7945, opened 2023-04-02
Change operator key to network for OSP in Poland

mutipg:
After arrangements with the Polish OSM community, OSP is a generally accepted collective name for independent organizations with the same task. This is not an organization that has central management and therefore instead of the key 'operator' a more appropriate key is 'network'

### DIFF
--- a/data/operators/amenity/fire_station.json
+++ b/data/operators/amenity/fire_station.json
@@ -1315,8 +1315,8 @@
       "matchNames": ["osp"],
       "tags": {
         "amenity": "fire_station",
-        "operator": "Ochotnicza Straż Pożarna",
-        "operator:wikidata": "Q55289175"
+        "network": "Ochotnicza Straż Pożarna",
+        "network:wikidata": "Q55289175"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
The problem here is that `operator` is the key tag as this entry is part of the `operator` tree. Removing it might cause the fire station entry to fail to appear at all.

@bhousel could you confirm if the above would occur?

bhousel:
> The problem here is that `operator` is the key tag as this entry is part of the `operator` tree. Removing it might cause the fire station entry to fail to appear at all.
> 
> @bhousel could you confirm if the above would occur?

Yes, normally fire stations with common tags are identified by their `operator` tag.  If OSP doesn't fit this model, we should remove it from the index.

It sounds like the "Polizia locale" and "Polizia municipale" from #7474 that we ended up removing also.

mutipg:
Thank you for response.  I am removing the entry for the OSP in Poland due to the findings of the community, which has objections to treating 'Volunteer Fire Department' as a unified operator and the network branch for amenity=fire_station is not supported.

PR #7949, opened 2023-04-03
Support Weibo as the social links

tommylung:
Weibo is a famous social platform in China. I would like to support to link up with Weibo in NSI.guide. But I'm not sure is the code edit enough to support or not.


UKChris-osm:
@bhousel would this cause any clashes if merged? It seems okay to me.

1ec5 (approved):

bhousel:
> @bhousel would this cause any clashes if merged? It seems okay to me.

Thanks! looks good to me too 👍 
We should see them show up next time @arch0345 builds a new release..

UKChris-osm:
@tommylung Weibo is now live on https://nsi.guide 👍

PR #7950, opened 2023-04-03
Use Cyrillic alfabet for Serbian Starbucks

u-spec-png:
Starbucks has its logo in Cyrillic in Serbia it also has one on English but Cyrillic script is the "main" one.
6a9ktriab3i71


LaoshuBaby:
LGTM, even if you enter Starbucks instead of Старбакс in `ru` or `rs`, it will still suggest the correction to Старбакс.

PR #7954, opened 2023-04-03
Remove operator:short

Hiausirg:
operator:short tags should be removed. Such short variants are stored at wikidata, and such broad operator information is exactly why the operator's wikidata item is linked.  Additionally, I removed Gulf Power because they merged; their assets are now operated by Florida Power & Light (someone retagged already every OSM object).

Greetings

### DIFF
--- a/data/operators/power/line.json
+++ b/data/operators/power/line.json
@@ -3213,11 +3213,12 @@
       "displayName": "JR東日本",
       "id": "eastjapanrailwaycompany-e95322",
       "locationSet": {"include": ["jp"]},
-      "matchNames": ["JR東日本"],
+      "matchNames": ["東日本旅客鉄道株式会社"],
       "tags": {
-        "operator": "東日本旅客鉄道株式会社",
+        "operator": "東日本旅客鉄道",
         "operator:en": "East Japan Railway Company",
-        "operator:ja": "東日本旅客鉄道株式会社",
+        "operator:ja": "東日本旅客鉄道",
+        "operator:short": "JR東日本",
         "operator:wikidata": "Q499071",
         "power": "line"
       }
@@ -3226,11 +3227,12 @@
       "displayName": "JR東海",
       "id": "centraljapanrailwaycompany-e95322",
       "locationSet": {"include": ["jp"]},
-      "matchNames": ["JR東海"],
+      "matchNames": ["東海旅客鉄道株式会社"],
       "tags": {
-        "operator": "東海旅客鉄道株式会社",
+        "operator": "東海旅客鉄道",
         "operator:en": "Central Japan Railway Company",
-        "operator:ja": "東海旅客鉄道株式会社",
+        "operator:ja": "東海旅客鉄道",
+        "operator:short": "JR東海",
         "operator:wikidata": "Q513679",
         "power": "line"
       }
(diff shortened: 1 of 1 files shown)


kjonosm:
Please note, that NSI uses `short:operator=*` and its variants also for name matching. From the end user perspective these name variants are also helpful when searching for specific objects.

bhousel:
Yeah I really don't have a preference, but I don't see any harm in keeping `operator:short` tags around, as they are used for name matching, and it's kind of an established convention by other infrastructure mappers.  (Rail mappers especially want the tags there).

LaoshuBaby:
> Additionally, I removed Gulf Power because they merged; their assets are now operated by Florida Power & Light (someone retagged already every OSM object).

I manually made this changes and sync main branch, so this PR only related with removal of operator:short now.

For the removal, I don't think it is a good idea, because those short name also play important role as common used name for many users (nobody will spell its full name everytime). And it seems that especially in countries where the Latin alphabet widely used, the difference is very large.

1ec5:
At least the short names should be preserved as `matchNames`, which won’t be added to the OSM database. It would be an interesting project for NSI to automatically generate `matchNames` from Wikidata labels or statements, but that would be kind of overkill for now.

Hiausirg:
Okay, I added the short versions back as "matchNames", so they should be getting replaced with full names. Actually even helpful for me, since I can just type the short version and let it auto-expand. Very good idea indeed, thanks!

In my opinion: The short operator variant dosen't belong to the OSM object, if operator's wikidata is already present. AFAIK, operators wikidata on OSM objects is intended for stuff like this (general operator information) and then it should also be used for this. And not the next workaround "triplication" of operator information added.

Hiausirg:
Re-added the short operators for japanese companies. Diff looks a bit cluttered but that's because it should also reflect the recent changes to the power file. If you're an active power infrastructure mapper in Japan and think those tags are helpful for you - okay fine.

I, however, edit very extensively in the US and for me, such redudant tags which provide no demonstrated value are extremely disruptive during mapping. (not limited to those operator: tags).
The more tags on a feature, the more to look over, the more to take notice of, the more clutter, the more annoying. This all takes time - not worth mentioning with a few objects, but if you're dealing with thousands of objects, it all adds up.

I expect from a data consumer to use their own conversation tables which use existing operator/operator:wikidata tags* - for ALL kinds of general operator information (except normal human-readable name). I do NOT want do deal or just get confronted with de facto useless tags** all the time. This is all in all muuch less work for everyone involved.


_*Such "conversation tables" can be generated by wikidata query using only a few clicks, so that's not the problem._
_** "defacto useless" because I still haven't seen a good use case for this which couldn't be solved much more efficiently otherwise_

Lonerat:
That may be ideal, but there is no such consensus at the moment, and it would be premature to do so.

wikidata tag is just a tag that links OSM features to Wikidata, and after all, OSM and Wikidata are fundamentally different databases, and it should be difficult to achieve such an integration unless there is consensus from both editors and users of OSM.

For example, items that are easily managed in Wikidata, especially for things like `cuisine` tag for brands, they would be described by Wikidata. But even though NSI has been around for a while, they are still widely used.
So this does not mean that there is any consensus with each community to integrate such information into Wikidata even if somewhat redundant, and should not be done at this time.

(Although this may not be a general view at NSI, which is more focused on Wikidata)

As for me, I am not very positive about such integration, at least for now; it might be smart, but it would also almost force both OSM editors and users to use Wikidata.
As mentioned above, there is no such consensus, and since the two databases have different approaches, that much integration is not desirable at this time.

Hiausirg:
I doubt that there was ever any consensus under (US) power line mappers to mass-add operator:short to power lines. Then, I also don't need to gain consensus to remove it.

cusine=* and operator:short=* isn't compareable. cusine describes a physical "on the ground" property - this shouldn't be outsourced to wikidata, because when anyone want to search OSM data for e.g. cheese restaurants, the restaurant needs a cusine=cheese tag - otherwise, the search will be impossible or extremely hard.
operator:short (and all other tags related to such broader information about the operator) on the other hand is merely an alternative version of the operator, no one will search OSM for alternative versions of operators.
(now reading your post again I'm actually unsure if you're pro or contra, but anyway)

bhousel:
Lets keep these `operator:short` for now, there's no good reason to remove a tag that some people might be deriving some value from it.

PR #7959, opened 2023-04-04
Adding wikidata for Translink and Adding Casco Bay Lines

bkissin:
Again, please check to make sure I did this right. Thanks for your help.

### DIFF
--- a/data/transit/route/ferry.json
+++ b/data/transit/route/ferry.json
@@ -108,7 +108,7 @@
     {
       "displayName": "Casco Bay Lines",
       "id": "cascobaylines",
-      "locationSet": {"include": ["us-me"]},
+      "locationSet": {"include": ["us-me.geojson"]},
       "tags": {
         "network": "Casco Bay Lines",
         "network:wikidata": "Q5048264",
--- a/data/transit/route/ferry.json
+++ b/data/transit/route/ferry.json
@@ -456,7 +456,7 @@
       "locationSet": {"include": ["ca"]},
       "tags": {
         "network": "Translink",
-        "route": "ferry"
+        "route": "ferry",
         "operator:wikidata": "Q1142140"
       }
     },
--- a/data/transit/route/ferry.json
+++ b/data/transit/route/ferry.json
@@ -106,6 +106,17 @@
       }
     },
     {
+      "displayName": "Casco Bay Lines",
+      "id": "cascobaylines",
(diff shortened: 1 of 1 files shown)


data/transit/route/ferry.json
```diff
@@ -106,6 +106,17 @@
...
     },
     {
+      "displayName": "Casco Bay Lines",
+      "id": "cascobaylines",
+      "locationSet": {"include": ["us-me"]},
+      "tags": {
+        "network": "Casco Bay Lines",
+        "network:wikidata": "Q5048264",
+        "operator": "Casco Bay Lines",
+        "operator:wikidata": "Q5048264",
```
LaoshuBaby, line 116: It seems that Casco Bay Lines is both the network itself and its operator? But we know that a ferry network and a transport company are two different things.
It is possible to have such a tag, but it is indeed a bit strange, and it can be 2 different items if possible.

LaoshuBaby (approved):

data/transit/route/ferry.json
bkissin: OK, I will try to go into the code and get rid of the network info

data/transit/route/ferry.json
```diff
@@ -448,6 +457,7 @@
       "tags": {
         "network": "Translink",
         "route": "ferry"
```
tommylung, line 459:
```suggestion
        "route": "ferry",
```
Missing comma.

data/transit/route/ferry.json
```diff
@@ -106,6 +106,15 @@
...
     {
+      "displayName": "Casco Bay Lines",
+      "id": "cascobaylines",
+      "locationSet": {"include": ["us-me.geojson"]},
+      "tags": {
+        "operator": "Casco Bay Lines",
+        "operator:wikidata": "Q5048264",
+        "route": "ferry"
+      }
+    },
```
tommylung, line 117:
```suggestion
    },
    {
```
Missing braces.

bhousel:
> @bhousel An entry needs a `network` tag to be a part of the transit tree doesn't it? It can't just be an `operator`, is that correct?

Yes, transit networks should have `network`/`network:wikidata` tags.

bhousel:
I'm going to just merge this and add `network`/`network:wikidata` tags for Casco Bay Lines.

PR #7971, opened 2023-04-08
Add pharmacy chains in Israel

yotamN:


data/brands/amenity/pharmacy.json
```diff
@@ -2515,8 +2515,81 @@
         "amenity": "pharmacy",
         "brand": "Super-Pharm",
         "brand:wikidata": "Q13220217",
+        "brand:wikipedia": "en:Super-Pharm",
```
LaoshuBaby, line 2518:
```suggestion
```

Thank you for editing NSI, but we've deprecated wikipedia tag since last year, you can see https://github.com/osmlab/name-suggestion-index/issues/6481 for more information.

data/brands/amenity/pharmacy.json
```diff
@@ -2515,8 +2515,81 @@
...
+      }
+    },
+    {
+      "displayName": "מאוחדת פארם",
+      "locationSet": {
+        "include": ["il"]
+      },
+      "tags": {
+        "amenity": "pharmacy",
+        "brand": "מאוחדת פארם",
```
LaoshuBaby, line 2531:
```suggestion
        "brand": "מאוחדת פארם",
        "brand:en": "Meuhedet Pharm",
        "brand:he": "מאוחדת פארם",
```

If possible, can you duplicate name:he? because this can avoid wrong fallback to English on Hebrew language computer.

See: https://wiki.openstreetmap.org/wiki/Multilingual_names#Repeating_name_with_language_specific_tag

LaoshuBaby (changes requested):
First, thanks for your contributions to NSI, but there are two classes of problems that might be worth your time, so an example of each is given.

data/brands/amenity/pharmacy.json
yotamN: Sure, fixed

data/brands/amenity/pharmacy.json
yotamN: Fixed

PR #7972, opened 2023-04-08
Add Belgian chain of electronics stores Eldi

mstsbe:
I've revised the Wikidata entry for the Belgian chain of electronics stores named Eldi, so it is usable for NSI (links to social media accounts/logos). Also added an entry in NSI itself.


Cj-Malone:
Thanks.

PR #7977, opened 2023-04-09
Exclude generic names

serhii-muchychka:
In Ukrainian, these words are shop types, not names


tas50 (approved):

tas50:
Thanks @serhii-muchychka

PR #7988, opened 2023-04-11
Support Weixin (Wechat) as the social links

tommylung:
Weixin (WeChat) is a famous social platform in China. I would like to support to link up with Weibo in NSI.guide.
Similar to Weibo PR: https://github.com/osmlab/name-suggestion-index/pull/7949


1ec5 (approved):
Thanks!

PR #7991, opened 2023-04-11
Add Wikidata for Vinnies

codeinabox:


codeinabox:
@UKChris-osm thank for merging these in. With the Wikidata entries, I find you don't need to set the start time for Twitter accounts as [BorkedBot](https://www.wikidata.org/wiki/User:BorkedBot) will do that automatically

UKChris-osm:
Just to add, we're not entirely sure if we'll be able to continue using the Twitter API for logos, as they may soon stop free access, so if a Facebook profile is also available, it would be handy to add these to any Wikidata as well.

PR #8003, opened 2023-04-16
Adding bar: Vino Volo

ShinyDog2636:
Adding Vino Volo, a bar found in several American and Canadian airports. Most Vino Volo locations are mapped as bars, and the Wikipedia description identifies the business as bars, so I have used this tag.


1ec5 (approved):
Yep, these are bars. Thanks!

PR #8006, opened 2023-04-17
Updated Cole Haan from US-only to international

ShinyDog2636:
Cole Haan is an international brand, see https://stores.colehaan.com/index.html


UKChris-osm:
Thank you @ShinyDog2636 👍

PR #8008, opened 2023-04-17
TPL Marche -> Contram Mobilità only for Macerata province

Robot8A:


arch0345:
Hi, I noticed that this PR introduces semicolon delimited values for `network`/`network:short`/`network:wikidata`. Ideally there should only be one value for these per entry to avoid warnings when running the build script:
image

Robot8A:
TPL Marche is separated into subnetworks for the different provinces:
http://orari.trasporti.marche.it/ (see on the bottom of the page)
Bus stops are marked as shown in this picture: https://www.cronachepicene.it/2022/11/22/start-sciopero-il-2-dicembre-possibili-disagi/373859/ TPL, and provincial network
Should I just put the provincial networks and not the TPL of the whole region as the sole network?

PR #8010, opened 2023-04-18
The Grumpy Baker

codeinabox:
The Grumpy Baker, which despite the name, describes its locations as cafés https://www.thegrumpybaker.com.au/


UKChris-osm:
Looks good to me @codeinabox 👍

PR #8016, opened 2023-04-21
remove Entergy local subsidiary replacement

Hiausirg:
Local subsidiaries shouldn't be replaced with the parent company, as this reduces accuracy.

### DIFF
--- a/data/operators/power/line.json
+++ b/data/operators/power/line.json
@@ -151,7 +151,7 @@
       }
     },
     {
-      "displayName": "Alliant f",
+      "displayName": "Alliant Energy",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Would the same be true for `/power/substation.json`?

Hiausirg:
Yes, good catch! 

I created another pull request at https://github.com/osmlab/name-suggestion-index/pull/8020

UKChris-osm:
You're welcome to make the edit within this PR as well, no need for two PRs :)

Also, if the `matchNames` is going to be empty with your edit, you can remove the tag in its entirety.

PR #8027, opened 2023-04-24
Add Belgian fashion chain e5 Mode

mstsbe:
Please add the Belgian fashion chain e5 Mode. The Wikidata page has the name with capital "E", but it seems the company uses and prefers the lowercase version. Therefore also including a matchname handle (the "Mode" part is frequently omitted).


UKChris-osm:
Thank you @mstsbe 👍

PR #8031, opened 2023-04-26
Add Wikidata tags for Arva

AntiCompositeNumber:


bhousel:
Thanks @AntiCompositeNumber 👍

AntiCompositeNumber:
This commit added wikidata tags for `substation` and `line`. `minor_line` inherits from `line`, and does not usually need to be added seperately.

PR #8039, opened 2023-04-27
Add a number of electric operators in west Africa

russss:
There's a huge amount of development going on in this region and it would be good to get these in ahead of that. I suspect I'll have more.


Cj-Malone:
Thank you

PR #8063, opened 2023-04-30
Corrected details for TIC bus network in Compiègne, France

phcarval:
Hello,

Here is a request to correct the properties of the bus network of Compiègne, France. I've separated the entities "TIC" (which is the network) and "Acary" or "Transdev Acary" (which is the operator, Transdev Picardie being the owner of Acary). I've propoed the full name of the network "Transports Intercommunaux du Compiégnois" to be used to refer to the network, while keeping the short name to be displayed.
I've also added the tag "fee":"no", as TIC is one of the very first free bus networks in France, dating back to 1975. I fully expect this to continue in the future. However, if this is not appropriate, the tag can be removed.

Thank you

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -15670,6 +15670,7 @@
         "network": "Transports Intercommunaux du Compiégnois",
         "network:wikidata": "Q60846198",
         "operator": "Acary",
+        "operator:wikidata": "Q118108069",
         "route": "bus",
         "fee": "no"
       }
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -15665,7 +15665,7 @@
     {
       "displayName": "TIC",
       "id": "tic-add5eb",
-      "locationSet": {"include": ["fx"]},
+      "locationSet": {"include": ["fr-hdf.geojson"]},
       "tags": {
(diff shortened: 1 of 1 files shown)


phcarval:
On second thought, I removed the fee tag, since it is entirely possible that a paid bus line is added someday.

UKChris-osm:
Would using [fr-hdf.geojson](https://github.com/osmlab/name-suggestion-index/blob/main/features/fr/fr-hdf.geojson) be a suitable geo restriction as well @phcarval?

phcarval:
> Would using [fr-hdf.geojson](https://github.com/osmlab/name-suggestion-index/blob/main/features/fr/fr-hdf.geojson) be a suitable geo restriction as well @phcarval?

Yes indeed! What is the associated key?

UKChris-osm:
This would be part of the `locationSet` tag, which is currently set to mainland France `fx`, if you change this value to `fr-hdf.geojson` it will tell our index to target only that region for this particular entry.

phcarval:
> This would be part of the `locationSet` tag, which is currently set to mainland France `fx`, if you change this value to `fr-hdf.geojson` it will tell our index to target only that region for this particular entry.

Thank you very much! I have made the change.

UKChris-osm:
Thank you @phcarval 👍

PR #8065, opened 2023-04-30
Merge CLH and GPSS pipeline operators into Exolum

russss:
The CLH-PS (formerly GPSS) pipelines in the UK are now called the Exolum Pipeline System.

Exolum only appears to operate in Spain and the UK.

https://www.linewatch.org.uk/news/2021-03-15/clh-ps-becomes-exolum-pipeline-system

### DIFF
--- a/data/operators/man_made/pipeline.json
+++ b/data/operators/man_made/pipeline.json
@@ -940,15 +940,6 @@
       }
     },
     {
-      "displayName": "CLH",
-      "id": "clh-143312",
-      "locationSet": {"include": ["001"]},
-      "tags": {
-        "man_made": "pipeline",
-        "operator": "CLH"
-      }
-    },
-    {
       "displayName": "CNR",
       "id": "cnr-143312",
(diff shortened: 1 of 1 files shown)


data/operators/man_made/pipeline.json
```diff
@@ -1966,11 +1957,13 @@
     },
     {
       "displayName": "Exolum",
-      "id": "exolum-143312",
-      "locationSet": {"include": ["001"]},
+      "id": "exolum-0a3816",
+      "locationSet": {"include": ["es", "gb"]},
+      "matchNames": ["clh", "gpss"],
```
LaoshuBaby, line 1962:
```suggestion
      "matchNames": ["clh", "clh-ps", "gpss"],
```
Do you need to add `CLH-PS` also as a matchName?

russss: Probably not necessary (it's technically the name of the pipeline system not the operator) but it is present a few times in the UK so I've added it.

Issue #6791, opened 2022-07-17
Mongolian barbecue should be its own cuisine


1ec5:
The index currently treats [Mongolian barbecue](https://en.wikipedia.org/wiki/Mongolian_barbecue) as either a Mongolian kind of barbecue or just barbecue. In reality, it’s a Chinese stir-fry restaurant format that has as much to do with barbecue as a Philly cheesesteak. I think we should tag these restaurants as `cuisine=mongolian_barbecue` or `cuisine=mongolian_bbq`. There are [actual Mongolian restaurants](https://www.kqed.org/arts/13915312/real-mongolian-barbecue-khorkhog-richmond) that serve actual Mongolian-style barbecue; they should be the ones tagged `cuisine=mongolian;barbecue`.

https://github.com/osmlab/name-suggestion-index/blob/989ddbdace1e67d7a0e0cb86bd4cda0ee91209c7/data/brands/amenity/restaurant.json#L277-L278 https://github.com/osmlab/name-suggestion-index/blob/56d53931da306b1611d62d39e764ad1a61abd86c/data/brands/amenity/fast_food.json#L2803-L2805

LaoshuBaby:
Yes, after research, I agree with your idea, this "Mongilian barbecue" infact just a "铁板烧"(Teppanyaki), using minced meat, and not real mongolian cuisine, it was invented in Taiwan.

Real mongolian food ([Khorkhog?](https://en.wikipedia.org/wiki/Khorkhog)) usually use bigger piece of meat, like [this example](https://www.sohu.com/a/227429465_167788)(China's Inner mongolia) or [this one](https://www.popdaily.com.tw/forum/life/946219)(Mongolia). I search for photo of those 2 brand and ensure they aren't traditional mongolian restaurant.

In my opinion, I'd tag those 2 brand with `cuisine=mongolian_grill` and left traditional with `cuisine=khorkhog`(maybe need a Mongolian people's proposal), and I'll just change them.

1ec5:
Thanks – by the way, the cuisine is normally called “Mongolian barbecue”. Some American chains put “Mongolian grill” in their names to give it a more upscale sound, but the independent restaurants always call it Mongolian BBQ. This is how [`cuisine=mongolian_bbq`](https://taginfo.openstreetmap.org/tags/cuisine=mongolian_bbq) entered the database.

PR #8066, opened 2023-04-30
O'Tacos is not mexican food despite the name

phcarval:
Hello,
I propose another pull request. Simple correction: the french fast food brand O'Tacos does not sell mexican food. French tacos are an entirely different kind of food; I propose the value "french_tacos" to represent it.
Thank you


LaoshuBaby:
图片

Oh, there are already many restaurant using this "french_tacos" value, so this looks good to me. We also meet with other "not original from" ploblem like #6791 (?)

PR #8067, opened 2023-04-30
Add Belgian brand Neuhaus to chocolate shops

mstsbe:
Please add the Belgian chocolate brand Neuhaus to shop/chocolate.
As for the locationset: I know Neuhaus is also active in other countries, but it is retailed there as a "shop in shop" or a shop corner (department store), and a few airports here and there.
In Belgium they have the most of their physical stores under their own management, so therefore making this the primary location.

### DIFF
--- a/data/brands/shop/chocolate.json
+++ b/data/brands/shop/chocolate.json
@@ -183,6 +183,16 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Thank you @mstsbe 👍

PR #8072, opened 2023-05-01
s.Oliver is a worldwide brand

u-spec-png:


data/brands/shop/clothes.json
```diff
@@ -7425,7 +7425,7 @@
       "displayName": "s.Oliver",
       "id": "soliver-618b05",
       "locationSet": {
-        "include": ["at", "be", "de", "fr", "hr"]
+        "include": ["001"]
```
LaoshuBaby, line 7428: 图片
Maybe not even as wide as global? If these areas are recorded in wikipedia are correct, you can use `{ include: ['150', 'in', 'cn'] }` to represent

u-spec-png: It seems like it's right. I will correct this.

PR #8078, opened 2023-05-01
Need to fix the duplicate of METRO

serhii-muchychka:
Not sure if this is the best solution. I will be happy for any advice


Dimitar5555:
LGTM, thanks @serhii-muchychka!

PR #8084, opened 2023-05-02
Add Buzz Gym

codeinabox:


codeinabox:
These gyms are marketed as being open 24 hours, should I add `opening_hours=24/7`?

codeinabox:
These gyms are advertised as being open 24 hours, should I add `opening_hours=24/7`?

PR #8100, opened 2023-05-05
Add Iberdoex

juanro49:
Iberdoex is a spanish group of service stations that include shop whith local products and parcel shop services

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -2432,7 +2432,7 @@
         "post_office:service_provider": "GLS",
         "shop": "yes",
         "website": "https://iberdoex.com/",
-        "wikimedia_commons" "Category:Iberdoex"
+        "wikimedia_commons": "Category:Iberdoex"
       }
     },
     {
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -2418,6 +2418,24 @@
       }
     },
     {
+      "displayName": "Iberdoex",
+      "locationSet": {"include": ["es"]},
+      "tags": {
+        "amenity": "fuel",
+        "brand": "Iberdoex",
+        "brand:wikidata": "Q25689041",
+        "brand:wikipedia": "ext:Iberdoex",
+        "name": "Iberdoex",
+        "post_office": "post_partner",
+        "post_office:parcel_from": "yes",
+        "post_office:parcel_pickup": "yes",
+        "post_office:service_provider" "GLS",
+        "shop": "yes",
+        "website": "https://iberdoex.com/",
+        "wikimedia_commons" "Category:Iberdoex"
+      }
+    },
+    {
       "displayName": "IES",
       "id": "ies-e8f712",
       "locationSet": {"include": ["it"]},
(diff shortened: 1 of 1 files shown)


data/brands/amenity/fuel.json
```diff
@@ -2417,6 +2417,24 @@
...
+      "tags": {
+        "amenity": "fuel",
+        "brand": "Iberdoex",
+        "brand:wikidata": "Q25689041",
+        "brand:wikipedia": "ext:Iberdoex",
+        "name": "Iberdoex",
+        "post_office": "post_partner",
+        "post_office:parcel_from": "yes",
+        "post_office:parcel_pickup": "yes",
+        "post_office:service_provider" "GLS",
```
tommylung, line 2432: Missing colon.
```suggestion
        "post_office:service_provider": "GLS",
```

juanro49: > Missing colon.

Thanks! I updated this (and missing colon in wikimedia_commons tag)

Cj-Malone:
I'm not sure about this one, `wikimedia_commons` and `website` should be POI values, not brand ones. And then `shop=yes`, in GB shops and fuel stations are usually separate POIs, and it's usually `shop=convenience `

juanro49:
> I'm not sure about this one, `wikimedia_commons` and `website` should be POI values, not brand ones. And then `shop=yes`, in GB shops and fuel stations are usually separate POIs, and it's usually `shop=convenience `

Do you think it would be more appropriate to put [brand:website](https://wiki.openstreetmap.org/wiki/Key:brand:website) for the brand website instead of just website which would be reserved in case the specific establishment has another website? I think it is a good option to change to `brand:website`.

Regarding `shop=yes`, I think that according to the documentation you are right and it would be more appropriate to indicate it in different nodes, although it is also allowed in the same node https://wiki.openstreetmap.org/wiki/Tag:shop%3Dconvenience.

Cj-Malone:
> Do you think it would be more appropriate to put [brand:website](https://wiki.openstreetmap.org/wiki/Key:brand:website) for the brand website instead of just website which would be reserved in case the specific establishment has another website? I think it is a good option to change to `brand:website`.

Not really, I don't think it's worth mappers time adding or maintaining `brand:website`. As @bhousel said, it's available from [Wikidata](https://www.wikidata.org/wiki/Q25689041#P856), having brand data solely there means it can be instantly updated and fixed, whereas having it on the POI means consumers would have to choose between OSM and Wikidata, and mappers would have to do bulk updates to OSM, which can be, politically challenging.

That being said, there can also be drama if when removing a brand url from `website`, so on POIs (Like Lidl) where there isn't a POI website I'd:
```patch
-website=https://www.lidl.co.uk/
+brand:website=https://www.lidl.co.uk/
```
Just to avoid any drama, I wouldn't add it to begin with. And hopefully this is just a temporary negotiation, at some point it should be stopped.

For POIs where there are POI websites, I'd just remove the brand website and add the specific one:
```patch
-website=https://www.greggs.co.uk/
+website=https://www.greggs.co.uk/shop-finder?shop-code=0185
```

juanro49:
Thanks for the tips, it will help me to improve future contributions!

PR #8110, opened 2023-05-07
Added translations for Ukrainian bank and transport systems

neviksasha:
POI Alfa Bank (now Sens Bank) had translations into Russian and Italian. Added these translations for automatic replacement.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -9725,11 +9725,16 @@
         "amenity": "bank",
         "brand": "Сенс Банк",
         "brand:en": "Sense Bank",
+        "brand:it": "Sense Bank",
         "brand:uk": "Сенс Банк",
+        "brand:ru": "Сенс Банк",
         "brand:wikidata": "Q16691757",
         "name": "Сенс Банк",
         "name:en": "Sense Bank",
-        "name:uk": "Сенс Банк"
+        "name:it": "Sense Bank",
+        "name:uk": "Сенс Банк",
+        "name:ru": "Сенс Банк",
+        "operator": "Сенс Банк"
       }
     },
     {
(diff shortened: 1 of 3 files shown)


neviksasha:
Added information about Kyiv bus and subway

data/brands/amenity/bank.json
```diff
@@ -9725,11 +9725,16 @@
         "amenity": "bank",
         "brand": "Сенс Банк",
         "brand:en": "Sense Bank",
+        "brand:it": "Sense Bank",
```
LaoshuBaby, line 9728: What I am more confused about is why a bank in Ukraine needs an Italian brand name, because we don't need to contain every language's name in preset. But if you insist or think it is necessary, it can be added too.

serhii-muchychka: As far as I know there is no word "Bank" in the Italian and "Сенс" in the Russian. And there is no information about these translations in the bank's resources

neviksasha: I added translations into Russian and Italian so that in the ID editor, when you click on the proposal for correcting data, there are no incorrect names left.
An example of such an object:
https://www.openstreetmap.org/node/6847006410

Perhaps this is not the right decision, then please tell me how to do it better

bhousel: The preference is to only add the localizations that are used by the brand on the ground where the stores are. 

If previous mappers have added alternate transliterations like these, NSI won't be able to correct it  (our rules don't specify any tags to _remove_, and in any case it's really up to the mapper or community to decide that an Italian name was added wrongly to a Ukrainian bank). 

It's easy enough to remove these tags in bulk with tools like Overpass/level0/JOSM if it is really decided they aren't needed.

Issue #7545, opened 2022-12-15
EV Chargers - 'Named chargers'


tastrax:
Example - Adding a Tesla supercharger or destination charger and it gets 'named' (Tesla Supercharger or Tesla Destination Charger). This seems to be wrong according to both

https://wiki.openstreetmap.org/wiki/Key:name

and 

https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station

in reality very few, if any, charging stations have 'names'. This seems to be tagging for the renderer. I suggest name be removed from all EV charging stations as its more likely the 'names' should either be brand, operator or network.

bhousel:
Last time we considered this topic was #4985, and the research turned up that several brands (including Tesla Supercharger) do assign their charging stations unique names.

Where we left things was:  I added `"preserveTags": ["^name"]` to the charging station category.  This will allow the `name` tag as an initial value, but users are encouraged to change it, and NSI will not overwrite the custom names that people add.

tastrax:
Trouble is that people seldom add anything to the default name so you end up with heaps of non unique records which just duplicates the Brand (and sometimes the network and operator as well)

bhousel:
This does come up for discussion sometimes , like #5338 and #7236
On that #7236 one, the person even tried manipulating their screenshot, so I hope you understand that we like to move slowly on issues like this and do some research.

We have found on #4985 that some charging stations have unique names, so I still think that seeding the value with a name and letting users change it is the right thing to do.

> Trouble is that people seldom add anything to the default name so you end up with heaps of non unique records which just duplicates the Brand (and sometimes the network and operator as well)

Why does this matter though?

tastrax:
For me its about the value of the name and the consistency of its application across all charge points. Its not applied to all charging_points and in many, maybe most cases, it adds little value over the brand. Its often just the brand and location  which is generally obvious. It just looks like tagging for the renderer to me so maybe that's also an issue that could be handled by carto in using the brand as label rather than name.

Adamant36:
> Its not applied to all charging_points and in many, maybe most cases, it adds little value over the brand.

I've seen Tesla charging points in California that say "Tesla" in big letters on top of them. The cars are named Tesla, the company is named Tesla, the chargers say Tesla. Drawing an arbitrary line at the chargers just seems a little weird. Do you really think having that in the name field would have no value what-so-ever? Or for that matter that people even get the distinction between the name and the brand in case like this to begin with?

[...]

Issue #7236, opened 2022-10-21
ATMs shouldn't have a name tag


RedAuburn:
<img width="396" alt="Screenshot 2022-10-21 at 16 28 40" src="https://user-images.githubusercontent.com/26939824/197232926-793eff20-0a77-4373-8856-e56fb6c1ff70.png">

putting the brand in the name tag isn't correct

endim8:
>oh yeah this is https://github.com/osmlab/name-suggestion-index/issues/5338

oops sorry, didn't see that one 😅

The main problem is that it's incorrect data, as the ATM isn't called 'NatWest', it's just an ATM.
Data users should ideally display the name as `[brand=NatWest] + " ATM"` resulting in `NatWest ATM`.
Then, if a name tag is present, only display the name tag. 

This would fix strange edge cases, for example If an ATM actually has the name "John's ATM", an additional "ATM" will be appended, resulting in "John's ATM ATM" in software.

generally it's just a good idea not to mix up tags, especially when there are correct tags for that data

endim8:
This also applies for amenity=charging_station

LaoshuBaby:
> This also applies for amenity=charging_station

?!

https://github.com/osmlab/name-suggestion-index/blob/33a38d6b37e68b24e32e631712cdb696665eba6b/data/brands/amenity/charging_station.json#L2-L7
https://github.com/osmlab/name-suggestion-index/blob/b66c59beb7d32aa31a5d137de95f2c4128d360ec/data/operators/amenity/charging_station.json#L2-L4

We already have preserveTags in catagory's properties, so amenity=charging_station's name won't be overwrite, if you meet with this problem, can you give us a example OSM element?

endim8:
> We already have preserveTags in catagory's properties, so amenity=charging_station's name won't be overwrite, if you meet with this problem, can you give us a example OSM element?

it suggests setting name=*` to the brand name with and without an already existing name tag
<img width="763" alt="Screenshot 2022-10-26 at 19 03 36" src="https://user-images.githubusercontent.com/26939824/198102252-95b73e3c-499b-42e1-b3fd-86c4b7e3679a.png">
(https://www.openstreetmap.org/node/8842525755)

bhousel:
@endim8  It looks like you have modified your screenshot, or you are using a customized iD?

Normally users are prompted to enter a `name` field in the "Fields" section:

[...]

PR #8115, opened 2023-05-08
Remove brand names in name=* from charging_station

endim8:
The name tag should be used for the name of the individual feature, not the brand name. It also duplicates the info in brand/operator.

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -26,7 +26,6 @@
         "amenity": "charging_station",
         "brand": "7-Eleven",
         "brand:wikidata": "Q259340",
-        "name": "7-Eleven",
         "operator": "7-Eleven",
         "operator:wikidata": "Q259340"
       }
@@ -460,7 +459,6 @@
         "amenity": "charging_station",
         "brand": "bike-energy",
         "brand:wikidata": "Q67770877",
-        "name": "bike-energy",
         "operator": "bike-energy",
         "operator:wikidata": "Q67770877"
       }
@@ -486,7 +484,6 @@
         "amenity": "charging_station",
         "brand": "Blink",
         "brand:wikidata": "Q62065645",
-        "name": "Blink",
         "operator": "Blink",
         "operator:wikidata": "Q62065645"
       }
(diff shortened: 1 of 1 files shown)


bhousel:
We closed this before on #7545 - not sure whether there has been any more discussion about it, but please read that issue and everything linked from there to get up to speed.

endim8:
This time is different 😄 a new proposal has been approved, redefining `amenity=charging_station` to represent a site instead of individual charger, which are now mapped with `man_made=charge_station`. 
 https://wiki.openstreetmap.org/wiki/Proposal:EV_Charging_Station_Mapping

The name used to display the feature to the end-user should determined by the software, not forced through tagging. If a charge_station doesn't have a name tag, software can easily swap in the brand tag instead. The `name=*` tag implies that its value is the _explicit_ name for that individual feature.

This should imo also be the case for chain stores like McDonald's, the name shouldn't be tagged. Instead, the brand and branch tags should be used & software should display those.

kjonosm:
@endim8 According to the proposal, you're still encouraged to use `name=` for `amenity=charing_station`, see https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dcharging_station#Tags_to_use_in_combination

However, it is discouraged to use `name=` for `amenity=charge_point`, see https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dcharge_point#Tags_to_use_in_combination

bhousel:
Thanks for the research @kjonosm .. 
I'm going to do basically the same thing we did for Banks and ATMs (#7236)

`amenity/charging_station` can be the primary category and these things can have names
`amenity/charge_point` can be a template category that auto expands from the charging stations and these things will not have names.

Cj-Malone:
I think the vote was actually for `man_made=charge_point`

bhousel:
> I think the vote was actually for `man_made=charge_point`

Yea but it sort of changed the meaning of charging_station too.

If I'm reading it correctly, I think the situation is now analogous to fuel stations?

`amenity=charging_station` is the area like a fueling station
`amenity=charge_point` is the point like a gas pump

Cj-Malone:
Yeah, but I'm pretty sure it's `amenity=charging_station` and `man_made=charge_point`, not `amenity=charge_point`

PR #8122, opened 2023-05-09
Add shop/appliance to the electronics group

serhii-muchychka:
Often these types of stores are hard to clearly categorize

Examples:
https://nsi.guide/?id=arcelik-f0338d


bhousel:
Good idea, thanks! 👍

PR #8126, opened 2023-05-10
Update EV charging station operators in the UK

russss:
* Be.EV is a brand of Iduna and now operates outside Greater Manchester
* ecarNI is the same brand as ecars in the Republic of Ireland
* Ecotricity's charging network ("Electric Highway") was acquired by Gridserve and I believe it has been absorbed into their network.
* "POLAR" and "Rolec" are manufacturers of charging stations, not charging networks.

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -12,7 +12,9 @@
         "^europark as$",
         "^jernbaneverket$",
         "kommune",
-        "privat"
+        "privat",
+        "polar",
+        "rolec"
       ]
     }
   },
@@ -384,14 +386,13 @@
     {
       "displayName": "Be.EV",
       "id": "beev-116e8a",
-      "locationSet": {
-        "include": [
-          "gb-greater-manchester.geojson"
-        ]
-      },
+      "locationSet": {"include": ["gb-eng"]},
+      "matchNames": ["iduna"],
       "tags": {
         "amenity": "charging_station",
-        "operator": "Be.EV"
+        "brand": "Be.EV",
+        "operator": "Iduna",
+        "operator:wikidata": "Q118263083"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


data/operators/amenity/charging_station.json
```diff
@@ -1743,22 +1725,23 @@
...
       "id": "esbecars-179e3e",
-      "locationSet": {"include": ["ie"]},
-      "matchNames": ["esb"],
+      "locationSet": {"include": ["ie", "gb-nir"]},
+      "matchNames": ["esb", "ecarni"],
       "tags": {
         "amenity": "charging_station",
-        "operator": "ESB ecars",
-        "operator:wikidata": "Q3050566"
+        "operator": "ESB Group",
```
Cj-Malone, line 1732: I have no knowledge of this brand, but should this be?

```suggestion
        "operator": "ESB",
```

> Charge your electric vehicle on the ESB public charging network

> ESB owns, operates and maintains over 1,350 public charge points across the island of Ireland. Check out our interactive map to find the nearest ESB charge point to you

https://esb.ie/what-we-do/ecars

Cj-Malone: Oh, I've just seen a `brand` as well.

I've got no strong opinion on this.

russss:
> 2\. crossing operator and brand trees.

Yeah, I was wondering about that too. I think in most cases the operator will imply the brand so it does make sense to tag both. There are a lot of operators in the `amenity=charging_station` brand file and a lot of brands in the operator file so I'd be inclined to consider merging them.

And then there's also mention of the network tag on the wiki...

(I'm not even that interested in tagging charging stations - just happened to spot a few easy wins here.)

Cj-Malone:
I'm sure someone will tell us if this is wrong. Thank you.

PR #8127, opened 2023-05-10
Fix groups for shop/beauty

serhii-muchychka:
**shop/beauty** - places where services are provided. Therefore, it shouldn't be in a group with a categories where sell goods

It seems that **shop/hairdresser** and **shop/hairdresser_supply** were confused


bhousel:
The point of these matchgroups is to capture tags that are frequently confused for one another - so if these tags are frequently mixed up, they can both be in the group.

PR #8141, opened 2023-05-14
Removed EB Games; added it as GameStop matchname

ShinyDog2636:
EB Games branding was replaced with GameStop

### DIFF
--- a/data/brands/shop/video_games.json
+++ b/data/brands/shop/video_games.json
@@ -10,7 +10,7 @@
       "locationSet": {"include": ["sk"]},
       "tags": {
         "brand": "Brloh",
-        "brand:wikidata": "Q12763551",eb
+        "brand:wikidata": "Q12763551",
         "name": "Brloh",
         "shop": "video_games"
       }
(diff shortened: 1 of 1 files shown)


TagaSanPedroAko:
@ShinyDog2636 Actually, that only applies to Canada locations. Australia and NZ locations are not affected.

TagaSanPedroAko:
@ShinyDog2636 Are you in the Talk-ca mailing list and saw those related to those chains?

ShinyDog2636:
@TagaSanPedroAko Thanks for catching that, I made a bad assumption.

It looks like both EB Games and GameStop exist in Australia and NZ so this may be a little more difficult to fix. I'll try to delete this for now.

TagaSanPedroAko:
@ShinyDog2636 Will do a better pull request removing EB games from Canada only (and adding it as matchName for GameStop"). Not sure if we should provide two separate presets for GameStop US and GameStop Canada; I don't know if it also has a separate brand for Quebec.

PR #8145, opened 2023-05-14
Remove duplicate matchTags that duplicate the primary

### LINKED ISSUE #8137: Build script doesn't clean "shop/tobacco" from matchTags for "shop/tobacco" entry
See T-ZONE entry in shop/tobacco
https://nsi.guide/?id=tzone-fbb387

peternewman:
Closes #8137

### DIFF
--- a/lib/matcher.js
+++ b/lib/matcher.js
@@ -184,11 +184,12 @@ export class Matcher {
       items.forEach(item => {
         if (!item.id) return;
 
-        // Automatically remove redundant `matchTags` - #3417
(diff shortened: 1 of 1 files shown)


peternewman:
Should we also do something similar for the generic KV stuff, in case someone puts say `shop/yes` in matchTags?

peternewman:
> Should we also do something similar for the generic KV stuff, in case someone puts say `shop/yes` in matchTags?

What do you think about this @arch0345 ?

arch0345:
> > Should we also do something similar for the generic KV stuff, in case someone puts say `shop/yes` in matchTags?
> 
> What do you think about this @arch0345 ?

I think that that would be fine

PR #8148, opened 2023-05-15
Add additional European brands, extending locationset of Engie charging stations

mstsbe:
I've added another number of European brands, mostly in Belgium and the Netherlands.

The locationset of charging stations operated by Engie was extended to Belgium (be) as well.


Cj-Malone:
Thank you

PR #8161, opened 2023-05-17
Extend locationset for some brands with "be", added some other new brands

mstsbe:
Holland & Barrett and JD Sports are also active in Belgium, therefore extending their locationSet.
Also added 2 brands that weren't featured yet, but did already have Wikidata entries: department store Inno (the Belgian version of Galeria Kaufhof/Karstadt) and the parking operator Interparking (mainly multi-storey or underground parking garages).

### DIFF
--- a/data/brands/shop/department_store.json
+++ b/data/brands/shop/department_store.json
@@ -695,6 +695,17 @@
       }
     },
     {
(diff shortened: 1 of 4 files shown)


Cj-Malone:
Thank you

PR #8163, opened 2023-05-17
Update laboratory.json - Added brand 'Diagnostyka' in Poland

mutipg:
Added brand 'Diagnostyka' in Poland


Dimitar5555:
Thanks @mutipg! NSI does not collect information about the websites of brands. It is better to leave that information in Wikidata (as you've done).

PR #8164, opened 2023-05-17
Update sample_collection.json - update Diagnostyka data

mutipg:


Dimitar5555:
Thanks! Same note about `website` applies here as well (https://github.com/osmlab/name-suggestion-index/pull/8163#issuecomment-1551921152).

Issue #8157, opened 2023-05-17
Add region restriction and localizations for METRO


TagaSanPedroAko:
METRO (the German warehouse club) needs some region restriction, in response to the brand getting added to unrelated chains in the Philippines (see #8125). Looks like it's available across the world in the present NSI.

TagaSanPedroAko:
Location restriction to apply for Europe, India, Pakistan, Turkey. Some other examples elsewhere are the addition of the wrong brand (e.g. Indonesia, South Africa, Colombia).

kjonosm:
@TagaSanPedroAko Can this issue be closed?

PR #8184, opened 2023-05-19
Country restrictions for Costco, Costco Business Center, Metro and Sa…

TagaSanPedroAko:
adds country restriction for these warehouse club chains: Costco, Costco Business Center, METRO and Sam's Club.

Restriction for METRO solves #8157

### DIFF
--- a/data/brands/shop/wholesale.json
+++ b/data/brands/shop/wholesale.json
@@ -71,7 +71,7 @@
       "locationSet": {
         "include": ["ca", "es", "fr", "gb", "mx", "us"],
         "exclude": ["jp", "tw"]
-                     },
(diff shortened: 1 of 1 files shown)


UKChris-osm:
When it comes to `locationSet`, the `exclude` property is only useful if the `include` property is set to worldwide (`001`). If the `include` property contains a reduced selection of locations, then any location not listed is automatically ignored, which means you don’t need to add the `exclude` property in the edits above.

TagaSanPedroAko:
@UKChris-osm Removed them already. Good to go.

kjonosm:
Thanks @TagaSanPedroAko

PR #8190, opened 2023-05-20
Removing manually added ID for Shwapno

Syfur007:
The id was added manually and I didn't knew that the id would automatically be generated.


bhousel:
Thanks - build script will replace these automatically anyway, so it's not really necessary, but I merged it.

PR #8193, opened 2023-05-22
Update bus.json

bkissin:
Adding wikidata entries for some US bus services, removing duplicate entry for Addis Ababa

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -92,15 +92,6 @@
       }
     },
     {
-      "displayName": "addis abeba public transport",
-      "id": "addisabebapublictransport-2e2b93",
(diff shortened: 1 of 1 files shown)


data/transit/route/bus.json
```diff
@@ -91,15 +91,6 @@
         "route": "bus"
       }
     },
-    {
-      "displayName": "addis abeba public transport",
```
LaoshuBaby, line 95: 图片

There are still 20 relation contain this network value, are they still valid?

bkissin: @LaoshuBaby Can you double-check the list? I think I got all of them.

bkissin:
Should be good to commit

Issue #8179, opened 2023-05-18
Please Add Butera Market


Zol87:
https://buteramarket.com/

tommylung:
I have created a wikidata item [Q118629448](https://www.wikidata.org/wiki/Q118629448).

PR #8199, opened 2023-05-24
Add Butera Market

tommylung:
#8179


Cj-Malone:
`name:en` and `brand:en` seem unnecessary on this?

tommylung:
> `name:en` and `brand:en` seem unnecessary on this?

Oh yes, they are same as `name` and `brand`. I will remove them then. Thanks for reminding me

PR #8215, opened 2023-05-30
Yolk fast food in London

codeinabox:


Cj-Malone:
@codeinabox I'm not a Wikidata expert, but I think they basically require entries to have a "instance of" property. I've just had to get some entries un deleted because they were lacking the required details.

`instance of=business` or `brand` is the easiest, or we could be more specific like `fast food restaurant chain` or `café chain`.

PR #8232, opened 2023-06-02
add Maynilad

TagaSanPedroAko:


serhii-muchychka:
We can add this to `office=water_utility `and then it will be automatically generated for `man_made=water_works`. Maybe this is a better option?

TagaSanPedroAko:
@serhii-muchychka That's better.

TagaSanPedroAko:
also add it to man_made=wastewater_plant

TagaSanPedroAko:
@serhii-muchychka this can go already.

data/operators/man_made/water_works.json
```diff
@@ -185,6 +185,17 @@
...
+      "displayName": "Maynilad",
+      "locationSet": {
+        "include": ["ph"]
+      },
+      "tags": {
+        "man_made": "water_works",
+        "operator": "Maynilad",
+        "operator:wikidata": "Q6797462"
+      }
+    },
```
arch0345, line 198:
```suggestion
```
Since Maynilad is being added to `office/water_utility`, an entry will automatically be generated from this template:
https://github.com/osmlab/name-suggestion-index/blob/c7a44ef5fe8a36aa92c2cc1c9ca11fea3a53c54d/data/operators/man_made/water_works.json#L7-L15

You can read more about templates [here](https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#templates)

PR #8243, opened 2023-06-05
Remove unnecessary capitalization 'TOPGUN' -> 'TopGun'

AntonKhorev:
Not an abbreviation.

See in another preset:
https://github.com/ruosm-presets/literan-moscow/blob/6c22490b90777fc0dc8b9c72404d5fbb4d561c94/russian_shops.xml#LL2370C32-L2370C38
although they kept brand tag in all caps for some reason.

### DIFF
--- a/data/brands/shop/hairdresser.json
+++ b/data/brands/shop/hairdresser.json
@@ -758,12 +758,12 @@
       }
     },
     {
-      "displayName": "TOPGUN",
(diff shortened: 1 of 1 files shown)


bhousel:
Yeah I agree with @Cj-Malone  - we tend to try to honor the brand's own naming preference, and if they call themselves TOPGUN, we should also.

PR #8258, opened 2023-06-09
Create Japan GeoJSON README.md

ffff-23:


LaoshuBaby:
jp-01

It seems that the geojson defined in NSI is NSI-only and cannot be accessed in other projects, location-conflation can only support region code defined in country-coder:

https://github.com/rapideditor/location-conflation/blob/main/README.md#thanks

bhousel:
> It seems that the geojson defined in NSI is NSI-only and cannot be accessed in other projects, location-conflation can only support region code defined in country-coder:

Actually if you add `&referrer=nsi` to the url, you can use the NSI geojsons:

https://location-conflation.com/?locationSet=%7B%22include%22%3A%5B%22jp-01.geojson%22%5D%7D&referrer=nsi

Screenshot 2023-06-10 at 7 59 30 AM


The urls that we create on the nsi.guide site all have this referrer parameter too.
Sorry I should document this on the location-conflation project somewhere.

LaoshuBaby:
> Actually if you add `&referrer=nsi` to the url, you can use the NSI geojsons:

It's my problem, my browser anti-tracking plug-in automatically removed the referrer, thank you for demonstration 👍

bhousel:
Oh that’s funny! I hadn’t thought that a browser might do that .. maybe I should rename the parameter?

LaoshuBaby:
> Oh that’s funny! I hadn’t thought that a browser might do that .. maybe I should rename the parameter?

I tested and finally found is this plugin's side affect: https://addons.mozilla.org/en-US/firefox/addon/clearurls/

And here is its rule: https://gitlab.com/ClearURLs/rules/-/blob/master/data.min.json , `referrer` appears 9 times, looks like it isn't a good parameter name?

PR #8277, opened 2023-06-15
Update Coop sweden brand

Cj-Malone:
I don't think this one is controversial as I can't see them [using that as term](https://www.coop.se/). But I just wanted to make sure there wasn't a specific reason for it.

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -1527,11 +1527,13 @@
         "coop konsum",
         "konsum",
         "lilla coop",
+        "coop sverige",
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks, this seems fine 👍 
If it was me that named it that before (I don't remember),  there was some time a while back when I was just using whatever label the Wikidata page used.

[Q15229319](https://www.wikidata.org/wiki/Q15229319) does say Coop Sweden, but yeah these labels aren't necessarily how the brand calls themself, and we should prefer the name the brand calls themself.

PR #8281, opened 2023-06-16
Add Xarxa de Casals i Ateneus dels PPCC

jmaspons:


data/brands/amenity/social_centre.json
```diff
@@ -68,6 +68,22 @@
...
+        "ateneu independentista",
+        "ateneu popular",
+        "casal independentista",
+        "casal popular",
+        "xarxa de casals i ateneus dels països catalans"
+      ],
+      "tags": {
+        "network": "Casals i Ateneus dels Països Catalan",
+        "network:wikidata": "Q119411870",
+        "network:website": "https://www.casalsiateneus.cat"
```
LaoshuBaby, line 85: > Each item requires a tags value. This is just an Object containing all the OpenStreetMap tags that should be set on the feature.
>
> Note for brands, that common tags with additional details such as "website" or "facebook" is not accepted directly. Instead a tag referencing the brand's [wikidata](https://www.wikidata.org/) entry ("[brand:wikidata](https://wiki.openstreetmap.org/wiki/Key:brand:wikidata)") containing those details can be included. If no such exists, consider [adding it](https://www.wikidata.org/wiki/Special:NewItem).

https://github.com/osmlab/name-suggestion-index/wiki/Contributing#tags-required

bhousel:
This won't work without the `amenity=social_centre` tag

Robot8A (changes requested):
A trailing "s" is missing at the network name: Catalan**s**

Should the .geojson be inside the subfolder "ppcc"?

bhousel:
closing as there's still issues with this missing tags

PR #8284, opened 2023-06-17
Update bank.json

Davis-Klavins:
In Latvia, Nordea has been merged into Luminor Bank (previously DnB).


LaoshuBaby (commented):
https://github.com/osmlab/name-suggestion-index/blob/6c1ea0b09f72bb14e8223bce282e41da812cd1cd/data/brands/amenity/bank.json#L7564-L7577

Looks good, but what about add "Nordea" as matchName for Luminor Bank?

Davis-Klavins:
If it can be added to specific countries (Latvia and Lithuania), then yes (from https://en.wikipedia.org/wiki/Nordea_Bank_Lietuva seems that Lithuania can also be removed from Nordea). In Estonia, Nordea is still operating. In Latvia, I already manually corrected all tags in OSM where Nordea was still used.

PR #8286, opened 2023-06-17
added local bakery chains Lagler, Taumberger and Zeppitz

haansn08:
Adds a few local bakery chains which have multiple stores in Carinthia:
* Lagler: https://www.lagler.at/filialen/
* Taumberger: http://www.taumberger.co.at/filialen/
* Zeppitz: https://zeppitz.at/200_standorte.html

### DIFF
--- a/data/brands/shop/bakery.json
+++ b/data/brands/shop/bakery.json
@@ -1208,6 +1208,16 @@
       }
     },
     {
+      "displayName": "Lagler",
+      "locationSet": {"include": ["AT-2.geojson"]},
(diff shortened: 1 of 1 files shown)


arch0345:
Thanks for adding these brands, please note however that entries without a `:wikidata` tag won't get generated as a preset in iD. If these brands don't have a Wikidata item yet, you can easily create one using [this guide](https://github.com/osmlab/name-suggestion-index/wiki/Editing-Wikidata).

PR #8287, opened 2023-06-17
Add Ballad Health to hospital operators

wcedmisten:
https://en.wikipedia.org/wiki/Ballad_Health


arch0345:
Thanks @wcedmisten and welcome!

PR #8290, opened 2023-06-19
Move 2 entries from post_office to courier

govvin:
Update the NSI entry for 2 entries mis-tagged as `amenity=post_office` to `office=courier`, following the wiki documentation for [office=courier](https://wiki.openstreetmap.org/wiki/Tag:office%3Dcourier), which states "A courier delivery service, which is not a post office or national post. "

The office=courier tag is relatively newer than the post office tag, but has gained wider use in 2018, so many other establishments that fit this description are likely to be incorrectly tagged as the older `amenity=post_office.`

This is my first time to "move" an entry from an existing category, to a new one.  I'm not sure if I correctly created the courier.json file correctly.

### DIFF
--- a/data/operators/amenity/post_office.json
+++ b/data/operators/amenity/post_office.json
@@ -642,17 +642,6 @@
       }
     },
     {
-      "displayName": "LBC",
-      "id": "lbc-7e1eed",
-      "locationSet": {"include": ["ph"]},
-      "matchNames": ["lbc express"],
-      "tags": {
-        "amenity": "post_office",
-        "operator": "LBC",
-        "operator:wikidata": "Q17075584"
-      }
-    },
-    {
       "displayName": "Lietuvos paštas",
       "id": "lietuvospastas-738531",
       "locationSet": {"include": ["lt"]},
--- a/data/operators/amenity/post_office.json
+++ b/data/operators/amenity/post_office.json
@@ -568,18 +568,6 @@
       }
     },
     {
-      "displayName": "J&T Express",
-      "id": "jandtexpress-b78fda",
-      "locationSet": {
-        "include": ["id", "my", "ph"]
-      },
-      "tags": {
-        "amenity": "post_office",
-        "operator": "J&T Express",
-        "operator:wikidata": "Q95690344"
-      }
-    },
-    {
       "displayName": "Jersey Post",
       "id": "jerseypost-984d36",
       "locationSet": {"include": ["je"]},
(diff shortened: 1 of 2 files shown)


bhousel:
I really the idea of this tag - it's weird that in OSM we tag courier companies as post offices.

But I'm kind of worried switching these items over to a new tag that hasn't really caught on, and from the history on the OSM wiki page, it looks like someone just made it up recently without much discussion.  Is there some more discussion elsewhere about it?  I'd hate to switch over NSI to the new tag and then have random OSMers show up here mad.

govvin:
Fun fact: the tag status of `amenity=post_office` is 'de facto', and never went through the formal voting process.

The lifecycle of our tags in OSM either go through the formal voting process , or grow organically starting from being 'in use', until it gets wider adoption, where at some point after,  it becomes 'de facto'  .

I believe the wiki editor who updated the status of `office=courier` wasn't being devious at all, and is merely reporting the observation that it's already in use (globally), and actually follows the community-defined standard for ["in use"](https://wiki.openstreetmap.org/wiki/Tag_status#.22In_use.22_status) status.  (Side note, tho unrelated to NSI: the iD tagging scheme recognizes [`shop=water`](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dwater)  although its status is also "in use.")

image

The usage growth is interesting, while it was first used in 2014, the uptake didn't happen until 2018, and the sharp adoption happened in 2021.  Perhaps the usage trend is good indicator that the tag is well on its way to wider adoption, and the 315 unique users who has so far used the tag, despite the 2017 [rejection](https://wiki.openstreetmap.org/wiki/Proposal:Courier#Voting) of the original proposal (13 for, 15 against, 1 abstention) ?

Anyway, I suspect that many users continue to use `amenity=post_office` because, (1) it's similar to what they're looking for, but not exactly; (2) and it's rendered on the map. The alternative tag is never suggested by all editing software, so it can only mean that they're being consciously added based on information gleaned from the wiki. Besides, regular map user folks do make a distinction between post offices, and couriers.



Rather than re-classifying existing entries, and cause any unnecessary uproar, would it be more acceptable to suggest new entries that are still not in NSI, but fits the `office=courier` definition? In our region, we definitely have a lot more brands in that can be added.

Cj-Malone:
I think Purolator is one for this too. I'll wait for the id tagging schema PR before adding it though.
```json
    {
      "displayName": "Purolator",
      "tags": {
        "brand": "Purolator",
        "brand:wikidata": "Q7261372",
        "name": "Purolator",
        "office": "courier"
      }
    }
```

PR #8296, opened 2023-06-21
Lantana cafe restaurants in London

codeinabox:


kjonosm:
Judging their website despite the name they rather look like a restaurant place?
https://lantanacafe.co.uk
Can you confirm it's actually a cafe?

PR #8303, opened 2023-06-24
Split Himawari Pharmacy

ffff-23:
@LaoshuBaby continuation of  https://github.com/osmlab/name-suggestion-index/commit/e776465fd19de8a9f176a89018fc244f3f0456a4

These are pharmacies with the same name, but I believe they are unrelated and just happen to use the same name.
It does not seem to be a franchisor-franchisee relationship.

I created each wikidata item and split it up, but the number of stores for each Himawari Pharmacy seems to be less than 10, is this a problem?

### DIFF
--- a/data/brands/amenity/pharmacy.json
+++ b/data/brands/amenity/pharmacy.json
@@ -4531,6 +4531,22 @@
       }
     },
     {
+      "displayName": "スーパードラッグひまわり",
+      "id": "superdrughimawari-650eee",
+      "locationSet": {"include": ["jp"]},
+      "tags": {
+        "amenity": "pharmacy",
+        "brand": "スーパードラッグひまわり",
+        "brand:en": "Super Drug Himawari",
+        "brand:ja": "スーパードラッグひまわり",
+        "brand:wikidata": "Q119871355",
+        "healthcare": "pharmacy",
+        "name": "スーパードラッグひまわり",
+        "name:en": "Super Drug Himawari",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
> If you can be sure that one of them does not belong to this brand but to a local franchiser's, please feel free to move it to the Wikidata created for regional subsidiaries
> —— In e776465fd19de8a9f176a89018fc244f3f0456a4

> These are pharmacies with the same name, but I believe they are unrelated and just happen to use the same name.
> It does not seem to be a franchisor-franchisee relationship.

If they do not belong to the same branch of "Himawari Pharmacy" in each prefecture, but are different, then they really should not belong to the same brand

Well, we are talking about a series of companies like "Himawari Pharmacy in Saitama Prefecture", "Himawari Pharmacy in Shiga Prefecture", and so on.

> but the number of stores for each Himawari Pharmacy seems to be less than 10, is this a problem?

It seems that there are still some prefecture's himawari pharmacie with a large number of stores, such as https://himawarinews.com/service/pharmacy/

If you are convinced that a certain himawari pharmacy is too small to have notability and cannot be collected by https://github.com/ideditor/nsi-collector/blob/0594a6049c45e3bb91a96d6d29b443b164aff360/collect_osm.js#L33-L36, my personal opinion is not recommend to add, but only for this entry

@ffff-23

ffff-23:
Thank you!

I removed Himawari Pharmacy which has small number of stores and, add Pupule Himawari Pharmacy.

Issue #8264, opened 2023-06-12
shop=tool_hire vs shop=plant_hire


mstsbe:
I've encountered some examples of companies that are already in NSI under shop=tool_hire. But, some of the branches also rent out larger equipment (mobile gensets = mobile power generators, larger bulldozers, temporary toilets, temporary container units, aerial work platforms, etc..). This type of equipment rental is tagged with shop=plant_hire.

2 examples of such cases (at least in my region) are Boels Verhuur/Boels Rental and Loxam.
https://nsi.guide/index.html?t=brands&k=shop&v=tool_hire#boels-d0c09f
https://nsi.guide/index.html?t=brands&k=shop&v=tool_hire#loxam-33fef8
Both do individual tool rental, but also large to very large equipment rental.

Can we make a match group so these 2 companies are caught for shop=tool_hire as well as shop=plant_hire ?

peternewman:
So https://github.com/osmlab/name-suggestion-index/blob/main/config/matchGroups.json would make them all equivalent/match either way IIRC, but are you saying they all do both, or there are two sub-brands, one of which does the big stuff.

You can also do it on individual shops if you don't think it should apply to the whole category with:
https://github.com/osmlab/name-suggestion-index/blob/ac47ca45d1343c6e9b2fb244fe19be1c3f4d6b55/data/brands/shop/doityourself.json#L393

mstsbe:
Thanks for the insight. The second option is probably the better way to go, otherwise we may be making too disruptive changes in that category. I will make a PR so both brands are matched by both tags.

PR #8309, opened 2023-06-25
Boels Rental and Loxam: add matchTags entry to resolve issue #8264

mstsbe:
As discussed in issue #8264 : I'm adding a matchTags entry of shop/plant_hire to the existing entries of Boels Rental & Loxam, since they perform both functions at some of their locations.
After merging this, the issue may be closed.

### DIFF
--- a/data/brands/shop/tool_hire.json
+++ b/data/brands/shop/tool_hire.json
@@ -24,7 +24,7 @@
       },
       "matchNames": ["boels rental"],
       "matchTags": ["shop/plant_hire"],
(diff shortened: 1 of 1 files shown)


peternewman:
> After merging this, the issue may be closed.

FWIW @mstsbe if you use any of the magic words before the issue number it will automatically link them and it then gets auto-closed when its merged:
https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #8327, opened 2023-06-30
Add Viva.json

j13m126:
Also add matchTags for recently added convenience stops

https://www.omv.at/de-at/unsere-stationen/viva

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -645,17 +645,6 @@
       }
     },
     {
-      "displayName": "Viva",
-      "locationSet": {"include": ["at"]},
-      "matchTags": ["shop/supermarket"],
-      "tags": {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Hi @j13m126 

Viva already exists within this file, are you adding a different Viva to the one that's already there? If so, you will need to add a unique `display name`, such as `Viva (Austria)`.

You shouldn't need the `matchTags` as I believe this is already covered using the `matchGroup` property that's already a part of the Name Suggestion Index.

j13m126:
Oh, I'm sorry. I thought I checked that. 

The reason why I added matchTags to Billa now, stop&shop and unterwegs is, that I can still manually add a supermarket poi with the name 'Billa Now' and iD would suggest this: 
image

Would matchTags help here?

PR #8194, opened 2023-05-23
Philippine bus networks

TagaSanPedroAko:
add network tag preset for provincial bus companies Victory Liner and Five Star. also merge/deprecate PH-colon-separated values in favor of plain text values (PH:PUB/PUJ:(region) into LTFRB PUB/PUJ or more specific values). also delete PH:PUB:BGC in favor of plain text BGC Bus (using service branding).

PR #8339, opened 2023-07-01
delete Philippine bus network values already merged to single network

### LINKED ISSUE #8338: Deleted PH bus networks restored
Any idea why some Philippine bus networks that have been removed at #8194 was restored?

TagaSanPedroAko:
networks in question has been deleted under #8194, but has been restored yet again for some reason.

Close #8338

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -11212,60 +11212,6 @@
       }
     },
     {
-      "displayName": "PH:CAR:Baguio",
-      "id": "phcarbaguio-a45453",
-      "locationSet": {"include": ["001"]},
-      "tags": {
-        "network": "PH:CAR:Baguio",
-        "route": "bus"
-      }
-    },
-    {
-      "displayName": "PH:CAR:Benguet",
-      "id": "phcarbenguet-a45453",
-      "locationSet": {"include": ["001"]},
-      "tags": {
-        "network": "PH:CAR:Benguet",
-        "route": "bus"
-      }
-    },
-    {
-      "displayName": "PH:PUB",
-      "id": "phpub-a45453",
-      "locationSet": {"include": ["001"]},
-      "tags": {
-        "network": "PH:PUB",
-        "route": "bus"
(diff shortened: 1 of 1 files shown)


TagaSanPedroAko:
@tommylung You restored this?

LaoshuBaby:
> Is this due to npm run build restoring already-deleted entries for some reason?

It's not his fault, no matter who it is, just doing `npm run build` now will cause them to be re-added, and that command is in publish checklist.

That's because of those data record:

https://github.com/osmlab/name-suggestion-index/blob/ff3bb985765eceafccb4d908c47ec4c398046e2b/dist/filtered/transit_keep.json#L912-L917

They all appear more than 10 times in the OSM data and thus belong to the category of "should be included in NSI". This is why I suggested in #8194 that you match them with the changed operator and cleanup OSM data if possible - it doesn't have to be, of course.

If you are sure that such colon-separated network names no longer exist in OSM data, this PR will be kept until the next update of filtered data.

LaoshuBaby:
图片
I'm downloading latest pbf file and I'll generate a latest count file imitate [this PR](https://github.com/ideditor/nsi-collector/pull/13), so take it easy.

TagaSanPedroAko:
Maybe that's worth a MapRoulette challenge.

TagaSanPedroAko:
Just completed a sweep of routes using PH:PUJ:Cebu. Will follow up networks with PH:PUJ:Cavite and PH:P2P.

TagaSanPedroAko:
Maybe we can mark this change as pending for now. Public transport network tag value format for the Philippines is currently being discussed through Telegram.

TagaSanPedroAko:
Closed this. Discussion ended up with giving no preference for plain or prefixed network value, but encouraged using service branding for bus routes save Manila's city bus network. By the way, PH:PUJ: Sub values have been reinstated for use and will be kept so to localize tagging of jeepney routes rather than lumping them together in one nationwide network value.

PR #8343, opened 2023-07-04
Add entry for "Smartfuels" in fuel.json

govvin:
"Smartfuels" is a chain of gas stations in the Philippines.


govvin:
I think it's better to split this PR into two.

Issue #8355, opened 2023-07-06
short_name in the operator tree


serhii-muchychka:
In the NSI DB collected many entries, where `short_name` is an abbreviation for the `operator` rather than `name`. When using presets from NSI without additional tags, validators (JOSM, osmose) indicate an error "alternative name without the main one".

Can/Should they be replaced with other tags like `operator:short`?

There may also be a similar problem with other `*_name` tags (`alt_name`, `official_name`)

**Example:**
https://nsi.guide/index.html?t=operators&k=amenity&v=hospital&tt=Department%20of%20Health%20Services#departmentofhealthservices-585fc8

Cj-Malone:
I feel like I've read people talking about this recently, but I can't find it. It makes sense to me though.

bhousel:
Oh yeah it was #7954 
Some people wanted to get rid of it, but there seem to be valid reasons why people use this tag, so it should stay.

I didn't know about the JOSM warning for "alternative name without main one" - sounds like another good reason to use `operator:short`.

Cj-Malone:
The 2 issues are a little different, that was about removing `operator:short` this is saying we have some `short_name` that should actually be `operator:short`.

@serhii-muchychka as far as I'm considered, this is valid. PRs welcome, or specific examples for us to fix. Thanks

PR #8359, opened 2023-07-06
Replace short_name->operator:short for Operators Tree

serhii-muchychka:
only simple cases so far

ref: #8355


bhousel:
Seems ok - we can try it out and see if anything breaks 🔨

PR #8361, opened 2023-07-07
Brompton Bike Hire

codeinabox:


codeinabox:
I am not sure what to do for `network` or `operator`, though looking at existing entries it's also set to Brompton Bike Hire

https://www.openstreetmap.org/node/9978144492
https://www.openstreetmap.org/node/4424913296
https://www.openstreetmap.org/node/7231604757

PR #8363, opened 2023-07-07
Add UNO Parks

DivyaB1:
Hi,
I am trying to add UNO Parks - https://www.unoparks.eu/

### DIFF
--- a/data/brands/leisure/sports_centre.json
+++ b/data/brands/leisure/sports_centre.json
@@ -41,6 +41,18 @@
   },
   "items": [
     {
+      "displayName": "UNO Parks",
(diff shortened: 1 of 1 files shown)


arch0345:
> https://www.unoparks.eu/

It looks like they are builders of adventure parks: the locations in their portfolio are listed as clients at the bottom of the homepage. The ones in the USA (ie [Let's Play](https://letsplayparty.com/)) also don't mention UNO Parks.

However, it looks like they do operate 4 locations in Lithuania:
https://www.unoparks.lt/

DivyaB1:
I think you are right, Sorry about this. 
Closing this request.

PR #8366, opened 2023-07-08
Add dispensing=yes to PH pharmacies

govvin:
Add dispensing=yes to following pharmacies in PH: 

- Generika
- Grace Pharmacy
- Rose Pharmacy
- Saint Joseph Drug
- Southstar Drug
- TGP
- ThreeSixty


bhousel:
Isn't `dispensing=yes` an implied default?
I kind of thought `dispensing=no` was the only tag that was used by people, to tag the exceptions.

govvin:
According to the [dispensing section](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dpharmacy#Dispensing) of the pharmacy page, it depends on jurisdiction.

We definitely have (smaller) pharmacies here that only sell over-the-counter drugs, and not prescriptions.

bhousel:
sgtm 👍

PR #8370, opened 2023-07-09
Adding Cobblestone Hotels

arnowrld:


arch0345:
Based on their Wikipedia article, it looks like Cobblestone Hotels is a parent company that owns [a few hotel brands](https://en.m.wikipedia.org/wiki/Cobblestone_Hotels#Brands). I would instead create an entry for each brand and either create a Wikidata item for each brand or just point all the entries to the Cobblestone Hotels QID (this is what we did with Choice Hotels brands before most of them got their own QID)

PR #8375, opened 2023-07-09
Correcting shop=* type for Salt and Sunrise and updating Wikidata entry for Sunrise

dafadllyn:
Salt and Sunrise are two telecommunications companies in Switzerland. Their shops should be `shop=telecommunication`, not `shop=mobile_phone`, as they not only sell mobile phones, but primarily mobile and landline contracts. (Swisscom, the third and largest Swiss telecommunications company, is already in `shop/telecommunication`.)

Besides, the Wikidata entry for Sunrise need to be updated as the current company (Sunrise GmbH, [Q112573820](https://www.wikidata.org/wiki/Q112573820)) was created in 2022 when UPC (Schweiz) GmbH ([Q362582](https://www.wikidata.org/wiki/Q362582)) took over Sunrise Communications AG ([Q672648](https://www.wikidata.org/wiki/Q672648)) in 2021.

### DIFF
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -712,17 +712,6 @@
       }
     },
     {
-      "displayName": "Salt",
-      "id": "salt-741c3f",
-      "locationSet": {"include": ["ch"]},
-      "tags": {
-        "brand": "Salt",
-        "brand:wikidata": "Q17325543",
-        "name": "Salt",
-        "shop": "mobile_phone"
-      }
-    },
-    {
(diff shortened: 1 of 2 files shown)


arch0345:
Thanks @dafadllyn for your first contribution! Just want to double check with members of the Switzerland community before this gets merged.

cc @habi @jmarchon1

habi:
No double-checking needed contributions for @dafadllyn ;)

jmarchon1:
I'm not very familiar with either store, but the change looks good to me as well.

PR #8378, opened 2023-07-10
Updated wikipedia screenshot

MG-LSJ:
Wikipedia has switched the position of "Wikidata item" on their website. I have updated the image in the docs.


MG-LSJ:
> It seems that the position where "Wikidata item" appears is not fixed? At least for me, I haven't set anything up on the English Wikipedia, but it appears on the far right with no drop-down menu. I've seen several times Wikipedia page styles change several times in recent years, and I think they change quite frequently.
> 
> However, at least there is still a big visual change from left to right, so I approve this PR
> 
> 图片

The default ui behaviour of wikipedia is the collapsed tools menu. I have checked this on multiple browsers. 

(If you click on tools[hide] it will reset to the default. Wikipedia seems to remember the state of the Tools menu)

PR #8382, opened 2023-07-11
Multiple Indian brands updated.

MG-LSJ:


data/brands/amenity/atm.json
```diff
@@ -17,6 +17,16 @@
...
       }
     },
+    {
+      "displayName": "Hitachi Money Spot ATM",
+      "locationSet": {
+        "include": ["in"]
+      },
+      "tags": {
+        "amenity": "atm",
+        "brand": "Hitachi",
```
LaoshuBaby, line 27: https://github.com/osmlab/name-suggestion-index/wiki/Contributing#tags-required

NSI use wikidata as unified identifier for brand/network, so if it haven't been provided, the item won't be suggested in editor.

PR #8384, opened 2023-07-12
Added MAX&Co. clothes stores

DnaX:
MAX&Co. it's the youth fashion brand of Max Mara, present wordlwide.

https://www.wikidata.org/wiki/Q120570926


Cj-Malone:
Thanks

PR #8386, opened 2023-07-12
Remove name tag from HELLO CYCLING

ffff-23:
[Import](https://wiki.openstreetmap.org/wiki/Import/Catalogue/GBFS_HelloCycling_Import) is in progress. A port name is tagging in the `name` tag, but name:ja and name:en are not subject to import, so name:en=HELLO CYCLING is suggestion. I removed it because I thought it was inappropriate.

https://www.openstreetmap.org/node/11040587211


LaoshuBaby:
Have you considered using `"preserveTags":[]`?

ffff-23:
I'm not worried about tags being overwritten.
`name`=`セブンイレブン　葛飾奥戸２丁目店`
`name:en`=`HELLO CYCLING`
`name:ja`=`ハローサイクリング`

I don't think it's appropriate to be tagging as above, so I think it's better to remove it.
Is it possible to solve it by using `"preserveTags":[]`?

PR #8388, opened 2023-07-13
Add Ubitricity

karen-kerr:


serhii-muchychka:
Hi, @karen-kerr could you please double check this? There is the same problem of duplication in the brands and operators
I don't know which one is better to leave, but if you tell me, I can quickly fix it myself

https://nsi.guide/?id=ubitricity-36827b

PR #8391, opened 2023-07-14
Update pawnbroker.json

govvin:
Remove duplicate entry. 

As seen from this street-level imagery, this pawnbroker brand simply uses "Villarica", and not "Villarica Pawnshop":  https://www.mapillary.com/app/?pKey=408491760912150


data/brands/shop/pawnbroker.json
serhii-muchychka, line 1: Could you add match name for the "Villarica" entry please

```
"matchNames": ["villarica pawnshop"],
```

govvin: Done. That sounds sensible.  Thanks.

PR #8392, opened 2023-07-14
Update fuel.json for Coles Express

Scottie0001:
Coles Express uses the Shell Australia brand of fuel and Discount is aquired from Coles (Supermarkets)

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -1146,6 +1146,9 @@
       "matchNames": ["coles"],
       "tags": {
         "amenity": "fuel",
+        "brand": "Shell Australia",
+        "brand:wikidata": "Q7493635",
+        "fuel:discount": "Coles",
         "brand": "Coles Express",
         "brand:wikidata": "Q5144653",
         "name": "Coles Express"
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -1149,8 +1149,8 @@
         "brand": "Shell Australia",
         "brand:wikidata": "Q7493635",
         "fuel:discount": "Coles",
-        "brand": "Coles Express",
-        "brand:wikidata": "Q5144653",
+        "operator": "Coles Express",
+        "operator:wikidata": "Q5144653",
         "name": "Coles Express"
       }
     },


UKChris-osm:
Hi @Scottie0001 

Each entry here in the Name Suggestion Index can only include one `brand` tag, and your edit adds a second one. Were you trying to replace / update the existing `brand` tag?

in this context, the `brand` tag refers to the brand of the company / retail chain, rather than the brand of fuel sold.

What some mappers do in this case is add the forecourt of a fuel station as two separate edits, one for the shop (such as Cole’s) and one for the petrol pump area, such as Shell.

Scottie0001:
My apologies, I have updated it with the desired change. As stated previously, the brand of fuel used is Shell Australia however the operator of the fuel station is Coles Express. In Australia, ou must go into the store to pay meaning that they do run together. If it is better suited, the brand/operator can be swapped around depending on the requirements. Coles Supermarkets discount for fuel can be used at the fuel station.

bhousel:
I tried to settle this better in 149a7de

- Coles Express is the convenience store brand and we already do have an entry for them under [shop/convenience](https://nsi.guide/index.html?t=brands&k=shop&v=convenience#colesexpress-0b0ae9) so there is nothing to change there.
- Shell Australia is the fuel brand, so I've excluded Australia from the global Shell entry and renamed the "Coles Express" item in fuel.json to use the Shell Australia brand.
- I didn't feel comfortable including the `operator` or `fuel:discount` tags, because on the [Coles Express page](https://en.wikipedia.org/wiki/Coles_Express), it says "As of 2012, approximately 270 independently franchised Shell service stations remain unassociated with Coles Express where the fuel discount offer is not accepted."  It would be wrong to include these tags on all Shell fuel stations in Australia, so mappers will need to add these tags to the stations where appropriate.

lkernan:
I'm not sure about this.  Shell servos in Australia are owned and operated by Viva Energy.  
They have no connection to the global Shell brand except that Viva has a license to use the brand.

Viva also has a license to the Coles Express brand (which they brand the store with) although that's about to change to OTR which Viva has purchased for that exact usage.

The receipt you get for your fuel says Coles Express, not Shell.  

On the Coles Express site (https://www.coles.com.au/coles-express) it even says:
"Coles Express is one of Australia's leading fuel and convenience retailers. With 723 stores across Australia and over 4,800 team members, we're always nearby to help keep you on the move.

Whether you're filling up the car with quality Shell fuel or shopping for a wide range of food and drinks on the go, we’re always there for you when you need us most. Our friendly team members at our service station and convenient stores are happy to help."

No mention of Shell anywhere.

bhousel:
ok - in 7b8c1aec8 I just retained "Coles Express" as the brand of the fueling stations, and it will also matchName ["coles", "shell", "shell australia"] so any of those names will get prompted to upgrade to Coles Express..

After the OTR rebrand, we can change it again.

If that doesn't match the situation on the ground, we can point people back to this issue.

lkernan:
It looks like this is about to be solved either way.
https://www.smh.com.au/business/companies/petrol-station-shakeup-goodbye-coles-express-hello-otr-20230822-p5dyjj.html

PR #8395, opened 2023-07-14
Update dist_files.js

kjonosm:
Add some more keys to exclude listing separate values in `taginfo.json`. Is it possible to also add `ref:isil`, `ref:DK:cvr` and `internet_access:ssid` or will colons within the key string lead to syntax problems?

### DIFF
--- a/scripts/dist_files.js
+++ b/scripts/dist_files.js
@@ -461,7 +461,7 @@ function buildTaginfo() {
 
       // Don't export every value for many tags this project uses..
       // ('tag matches any of these')(?!('not followed by :type'))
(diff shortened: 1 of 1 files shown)


bhousel:
> Add some more keys to exclude listing separate values in `taginfo.json`. Is it possible to also add `ref:isil`, `ref:DK:cvr` and `internet_access:ssid` or will colons within the key string lead to syntax problems?

oh good idea, thanks!
It's a regular expression, so:
* Colons are ok, you don't need to escape them
* I think you can include string fragments, so just adding 'ref' would match anything that includes 'ref' in it.

kjonosm:
I've now also added `ref` and `internet_access:ssid`.

PR #8402, opened 2023-07-16
Update "Tambunting" entry in pawnbroker.json

govvin:
Add money_transfer=* values, based on [official website](https://tambunting.ph/Main?padala) information.


LaoshuBaby (changes requested):
Just like "NSI is not a menu" so you don't need to list all the dish names in the cuisine, I think if there are a lot of recipients that can be transferred, you don't need to list them all

If you insist on listing, I suggest you pick the one that is most representative, otherwise it really isn't necessary

govvin:
Noted. In that case, I'll close this request.

PR #8419, opened 2023-07-20
Add Hollys KR

karen-kerr:
Hollys KR is a cafe selling coffee, rather than a shop selling coffee.


bhousel:
Thanks @karen-kerr 👍

PR #8311, opened 2023-06-26
Add 'REEL Cinemas'

DivyaB1:
Hi, I'm trying to add 'REEL Cinemas' - https://reelcinemas.co.uk/reel-cinemas

Issue #6662, opened 2022-06-13
Change "NORMA" back to "Norma" (shop=supermarket)


fnordiff:
It seems like the name/brand/displayName of "id":"norma-c1eeeb" had been changed to uppercase in [this change](https://github.com/osmlab/name-suggestion-index/commit/33904164faf229d7b4f97af2a4549c3ac94a3f34) but I don't understand why.

The typical tagging ([as of 2016](https://wiki.openstreetmap.org/wiki/Germany/Shops), without recent changes of this index) was "Norma" (1034x) vs. "NORMA" (128x). The [current tagging](https://taginfo.openstreetmap.org/keys/name#values) after 10 month with suggestion "NORMA" is still "Norma" (957x) vs. "NORMA" (632x). Since the sum of both is higher now it seems like most of the old supermarkets are still "Norma" and new ones are tagged by simply accepting any suggestion. Which means that the index has huge impact on the tagging and should be as good as possible. 

* The linked [wikidata](https://www.wikidata.org/wiki/Q450180) entry writes "Norma" (not uppercase).
* The [wikipedia](https://de.wikipedia.org/wiki/Norma_(Supermarkt)) entry writes "Norma Lebensmittelfilialbetrieb Stiftung & Co. KG". If we skip the legal form of the company as usual we end up with "Norma" (not uppercase).
* The [commercial register](https://www.online-handelsregister.de/handelsregisterauszug/by/Nuernberg/HRA/7886/Norma-Lebensmittelfilialbetrieb-Stiftung-Co-KG) writes "Norma Lebensmittelfilialbetrieb Stiftung & Co. KG". If we skip the legal form of the company as usual we end up with "Norma" (not uppercase).
* The [company itself](https://www.norma-online.de/de/impressum) writes in its imprint "Nor­ma Le­bens­mit­tel­fi­li­al­be­trieb Stif­tung & Co. KG". If we skip the legal form of the company as usual we end up with "Norma" (not uppercase).

The `brand=` might be tagged like the company communicates its brand. But the `name=` should be written like it is officially written.

There are many examples that a "brand word mark" is _not_ accepted as `name=`. People tag `name=Woolworth` ([shop=department_store](https://www.wikidata.org/wiki/Q183538)) instead of `name=WOOLWORTH`. They tag `name=Caterpillar` instead of `name=CATERPILLAR`. And so on.

[...]

Issue #5509, opened 2021-10-15
JYSK in lowercase


NKAmapper:
The correct brand name text for JYSK is "Jysk", according to Scandinavian grammar (names which are pronounced  like a normal word should have lowercase letters). Jysk also happens to be a normal word for in the Danish language, for the Jutlandic dialect.

Examples in Wikipedia: https://en.wikipedia.org/wiki/Jysk

shop/furniture

bhousel:
Are you really sure?  I really don't want to change this unless there's a good reason to.

There is plenty of evidence that the brand calls themselves "JYSK":

1. We try to use the name that most mappers in OSM prefer.  Our scripts have collected these counts from the OSM planet file:
https://github.com/ideditor/nsi-collector/blob/80803dadd41281f870ea52111c6b20fcce07ec6c/dist/osm/names_all.json#L8696-L8697
```js
"shop/furniture|Jysk": 181,
"shop/furniture|JYSK": 839,
```

2. The JYSK about us page:
https://jysk.no/om-jysk
<img width="847" alt="Screen Shot 2021-10-15 at 12 48 13 PM" src="https://user-images.githubusercontent.com/38784/137523962-7adcb06b-7f7e-4a82-8a5a-04f219aa14a3.png">


3. They seem to be styled as "JYSK" everywhere, even in Norway:
<img width="1169" alt="Screen Shot 2021-10-15 at 12 45 25 PM" src="https://user-images.githubusercontent.com/38784/137523697-b8a1451c-90a7-4013-8f41-8326e0958b5a.png">

PR #8424, opened 2023-07-21
Add Astron Energy

DivyaB1:
Hi,
I am adding 'Astron Energy' - https://www.astronenergy.co.za/

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -32,7 +32,7 @@
        "amenity": "fuel",
        "brand": "Astron Energy",
        "brand:wikidata": "Q120752181",
-       "name": "ASTRON ENERGY"
+       "name": "Astron Energy"
      }
    },
     {
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -26,6 +26,16 @@
   },
   "items": [
     {
+     "displayName": "Astron Energy",
+     "locationSet": {"include": ["za"]},
+     "tags": {
+       "amenity": "fuel",
+       "brand": "Astron Energy",
+       "brand:wikidata": "Q120752181",
+       "name": "ASTRON ENERGY"
+     }
+   },
+    {
       "displayName": "1-2-3",
       "id": "123-67b972",
       "locationSet": {"include": ["dk", "pl"]},


data/brands/amenity/fuel.json
```diff
@@ -25,6 +25,16 @@
...
   },
   "items": [
+    {
+     "displayName": "Astron Energy",
+     "locationSet": {"include": ["za"]},
+     "tags": {
+       "amenity": "fuel",
+       "brand": "Astron Energy",
+       "brand:wikidata": "Q120752181",
+       "name": "ASTRON ENERGY"
```
bhousel, line 35: From their store locator page, it looks like they call themselves "Astron Energy" in most places and on maps.  


Screenshot 2023-07-24 at 9 52 51 AM

The allcaps "ASTRON ENERGY" name looks kind of weird (this is more normal in Japan but not in South Africa as far as I know).  Would you be ok with just using Title Case here?




```suggestion
       "name": "Astron Energy"
```

bhousel (commented):
Just one question about the name case

data/brands/amenity/fuel.json
DivyaB1: Thanks @bhousel, for sure I'll update the name case.
Quick question about the 'name' tag - Does the name have to be what the logo of brand says or what it says on their map/store locator?
While adding 'Reel Cinemas' I got an impression it should be the logo, so used the same logic https://github.com/osmlab/name-suggestion-index/pull/8311#issuecomment-1607971655

DivyaB1:
Thanks, I've updated the name case. Question about this tag - does it have to be what the logo says or what the store locator/maps say. After https://github.com/osmlab/name-suggestion-index/pull/8311#issuecomment-1607971655 I was under the impression 'name' tag gets what the logo has.

Cj-Malone:
OSM name should be the fascia/sign.

But OSM has a complicated history with all caps. It's considered an error in many QA tools.

bhousel:
> Question about this tag - does it have to be what the logo says or what the store locator/maps say. After [#8311 (comment)](https://github.com/osmlab/name-suggestion-index/pull/8311#issuecomment-1607971655) I was under the impression 'name' tag gets what the logo has.

We try to find a balance between how the brand calls themself and what name would look like in a list or on a map pin.  So we really don't use all caps names unless it's in a place that writes them normally this way (like Japan), or for a brand that is really consistent about writing their name this way (like IKEA or JYSK). We do get some people very particular about the uppercase names that will open issues about it.

See https://github.com/osmlab/name-suggestion-index/issues/6662#issuecomment-1177618297
or https://github.com/osmlab/name-suggestion-index/issues/5509#issuecomment-944450197

PR #8425, opened 2023-07-21
Add German American Bank

DivyaB1:
Hi,
I'm adding 'German American Bank' - https://germanamerican.com/personal/

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -22,7 +22,7 @@
   },
   "items": [
      {
-      "displayName": "German American Bank",
+      "displayName": "German American",
       "locationSet": {"include": ["us"]},
       "tags": {
         "amenity": "bank",
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -26,7 +26,7 @@
       "locationSet": {"include": ["us"]},
       "tags": {
         "amenity": "bank",
-        "brand": "German American Bank",
+        "brand": "German American",
         "brand:wikidata": "Q120753420",
         "name": "German American"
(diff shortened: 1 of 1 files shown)


data/brands/amenity/bank.json
```diff
@@ -21,6 +21,16 @@
     }
   },
   "items": [
+     {
+      "displayName": "German American Bank",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "amenity": "bank",
+        "brand": "German American Bank",
```
bhousel, line 29:
```suggestion
        "brand": "German American",
```
Based on their branding, I would just call it "German American" in both the `brand` and `displayName` props.

Screenshot 2023-07-21 at 12 30 07 PM

data/brands/amenity/bank.json
```diff
@@ -21,6 +21,16 @@
     }
   },
   "items": [
+     {
+      "displayName": "German American Bank",
```
bhousel, line 25:
```suggestion
      "displayName": "German American",
```

bhousel (changes requested):
Thanks!  What do you think of using the simpler name?  Their branding doesn't seem to include the word "Bank"

DivyaB1:
Hi,
Thanks, they referred themselves as 'German American Bank' in their 'About Us' page so I added it. I've corrected it now!
https://germanamerican.com/about/history/#:~:text=The%20history%20of-,German%20American%20Bank,-tells%20quite%20an

bhousel:
Thanks @DivyaB1 👍

bhousel (approved):

PR #8429, opened 2023-07-22
Update bus.json

MASangys:
Rapid KL, might have different type of network, with 3 systems (normal, MRT Feeder Bus and LRT Feeder Bus (also different livery). Therefore my suggestion is to not adding network tagging, leave it to the mapper to add since some bus stop maybe served by 2/3 livery (can be using semicolon to add diff networks).


MASangys:
Yes

bhousel:
Ok I did that in 93849532b

PR #7273, opened 2022-10-27
Limit location of Kołobrzeski Rower Miejski

starsep:
Users of iD apply Kołobrzeski Rower Miejski brand sugestion for bicycle rental station outside of Kołobrzeg.
Many bicycle rentals systems are operated by Nextbike. Not sure why this particular network is suggested, perhaps Polish letter `ł` causes issues with text comparison?
I decided to limit area where Kołobrzeski Rower Miejski is applicable.
According to https://github.com/ideditor/location-conflation/blob/main/README.md#location-conflation default radius is 25km.

PR #8434, opened 2023-07-23
Limit area of Nextbike bicycle rental in Poland

starsep:
Follow-up to https://github.com/osmlab/name-suggestion-index/pull/7273


bhousel:
Thanks! 👍

PR #8435, opened 2023-07-23
Add D-parking

ffff-23:
This parking is operated by two operator. I created each item and also created the item only for brand.

http://www.d-parking.jp

Should I choose only one of brand tree or operator tree?

Thank you.

### DIFF
--- a/data/brands/amenity/parking.json
+++ b/data/brands/amenity/parking.json
@@ -1,0 +1,35 @@
+{
+  "properties": {
+    "path": "brands/amenity/parking",
+    "skipCollection": true,
+    "preserveTags": ["^name"],
+    "exclude": {
+      "generic": [
+        "^parking( lot)?$",
+        "^visitor parking$",
+        "^автостоянка$",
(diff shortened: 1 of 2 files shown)


bhousel:
If the 2 different operators have distinct locations, we could have 2 entries in the `operator` tree for them with each having the locationSet of where they operate..

If the 2 operators both operate in the same location (all Japan?), then we need to have a single entry in the `brand` tree, and the users would need to supply the operator details themselves.  

The details about why:
A good way to think about this question is - "If a user adds parking lot and types D-Parking for name, would the software be able to figure out which NSI item that means?"    
The software matches on `['key', 'value', 'name', 'location']`.
..so if `['amenity', 'parking', 'D-Parking', 'jp']` can match 2 things in NSI, it won't work right.

ffff-23:
I checked the website that both operators are active all over Japan, so I removed them from the operator tree.

Thank you, I will refer to it in the future!

bhousel:
Thanks for the quick fix!

PR #8436, opened 2023-07-23
Added Canadian auto parts retailer Lordco

ShinyDog2636:


data/brands/shop/car_parts.json
```diff
@@ -326,6 +326,20 @@
         "name": "NAPA Auto Parts",
         "shop": "car_parts"
       }
+    },
+        {
+      "displayName": "Lordco Auto Parts",
+      "locationSet": {
+        "include": ["ca"]
```
bhousel, line 333: From their store locator at https://www.lordco.com/stores , it looks like we could scope it to Alberta and British Columbia: 

```suggestion
        "include": ["ca-ab.geojson", "ca-bc.geojson"]
```

PR #8437, opened 2023-07-23
Moved CEFA Earling Learning from social_facility to kindergarten

ShinyDog2636:
Also added matchNames and matchTags


bhousel:
Thanks @ShinyDog2636  LGTM 👍

Issue #5233, opened 2021-08-15
Add Polaris


cliffordsnow:
Polaris is a US based manufacturer of ATV, snowmobiles, and motorcycles with sales offices worldwide except for a small handful of countries.

name=Polaris Inc.
brand:wikipedia=en:Polaris Inc.
brand:wikidata=Q1850587
shop=atv;snowmobile;motorcycle;boats

logo https://en.wikipedia.org/wiki/Polaris_Inc.#/media/File:Polaris_Industries_logo.svg

Notes: 
1. Most Polaris dealers I'm familiar with have different names, much like car dealers. 
2. Not all dealers sell all Polaris products.
3. Products may have different brand names, such as the Indian Motorcycle or the Slingshot 3 wheel motorcycle.

arch0345:
After looking at all the brands Polaris offers, here are the brands that should be added:

ATV:
- [ ] Polaris
- [ ] Hammerhead

Motorcycle:

- [ ] Indian Motorcycle
- [ ] Polaris

Snowmobile:
- [ ] Polaris
- [ ] Timbersled

Boat:
- [ ] Bennington
- [ ] Godfrey
- [ ] Hurricane

Car:
- [ ] Axiam

Other (potentially):
- [ ] Goupil
- [ ] Taylor-Dunn

We don't have categories for shop=atv and shop=snowmobile right now, but those could be added in the future.

arch0345:
I've added car/motorcycle brands for now as they had wikidata pages.

bhousel:
Was looking into this issue today and I'm kind of confused by this.. Polaris sells through dealers, they don't have "Polaris" retail stores .  We don't really use the `brand` tags this way in NSI.  

I think this sort of tagging might be more compatible with the proposed [`sells:` namespace](https://wiki.openstreetmap.org/wiki/Proposed_features/Sells)

arch0345:
> Was looking into this issue today and I'm kind of confused by this.. Polaris sells through dealers, they don't have "Polaris" retail stores . We don't really use the `brand` tags this way in NSI.
> 
> I think this sort of tagging might be more compatible with the proposed [`sells:` namespace](https://wiki.openstreetmap.org/wiki/Proposed_features/Sells)

[...]

PR #8438, opened 2023-07-23
Adding SAS, an American shoemaker and retailer

ShinyDog2636:


bhousel:
I took a look at their store locator  https://www.sasshoes.com/stores/  , and it looks like this brand might operate as a reseller through other stores, not operate their own branded stores.  In this situation we wouldn't add it to NSI because it would be incorrect for an independent shoe store to be branded as "SAS" just because they carry shoes by them.

We have a few issues like this before, see 
https://github.com/osmlab/name-suggestion-index/issues/5233#issuecomment-923184175

There is an old abandoned proposal to tag these reseller brands [using a `sells:` namespace](https://wiki.openstreetmap.org/wiki/Proposal:Sells), but it's not accepted tagging, and our software isn't set up to work with these tags yet.

ShinyDog2636:
Thanks for flagging this, you caused me to review this more closely and I discovered it's not as I originally understood.

However, I think this is different than the situation you linked to. In addition to non-branded resellers there are 1) company branded and operated stores, and 2) company branded but independently operated stores.

Their website lists several types of stores 'SAS Store - Company', 'SAS Store - Independent', and 'Multi-Brand Store'.

Here is an example of a company branded and operated store:
https://www.sasshoes.com/store-details/?StoreID=323

Some examples of company branded but independently operated stores:
https://www.sasshoes.com/store-details/?StoreID=220
https://sananshoes.com/

https://www.sasshoes.com/store-details/?StoreID=419
https://mysasshoes.ca/locations

Does the NSI need to make a distinction between the brand= value of independent vs company run stores? I think the difference would be shown in the operator= value.

Thoughts? Thanks for reviewing.

data/brands/shop/shoes.json
```diff
@@ -22,6 +22,18 @@
         "shop": "shoes"
       }
     },
+    {
+      "displayName": "SAS (San Antonio Shoemakers)",
+      "locationSet": {"include": ["001"]},
+      "matchNames": ["SAS Shoes", "SAS", "San Antonio Shoe","San Antonio Shoemakers"],
```
bhousel, line 28: This is just a comment, and doesn't matter at all:  Because "SAS" is used as a value on the `name` tag, it doesn't need to be included in the `matchTags` array.  It's already matched by being a `name`.  When our build script runs, the redundant names will be removed automatically from this list.  Just letting you know for future contribution, or in case you notice the next commit remove it.

```suggestion
      "matchNames": ["SAS Shoes", "San Antonio Shoe", "San Antonio Shoemakers"],
```

PR #8443, opened 2023-07-25
Update operator name

Scottie0001:
Moreton Bay Regional Council has changed it's name to City of Moreton Bay effective from 24/07/2023 local time


bhousel:
Thanks @Scottie0001 👍 
This project also supports a `matchName` property for capturing older or less preferred names of things, so that mappers may be prompted to upgrade the names of these. 

I added a matchName property in b7ed50469

PR #8445, opened 2023-07-25
iD doesn't suggest NSI presets for Nextbike

### LINKED ISSUE #8444: iD doesn't suggest NSI presets for Nextbike
**Steps to reproduce:**

- Add new node with tags `amenity=bicycle_rental;name=Nextbike` in the area where Nextbike works

I suspect this is due to exclusion pattern `"named": ["^nextbike( gmbh)?$"]`

serhii-muchychka:
closes #8444

### DIFF
--- a/data/brands/amenity/bicycle_rental.json
+++ b/data/brands/amenity/bicycle_rental.json
@@ -6,8 +6,7 @@
       "generic": [
         "^bicycle rental$",
         "^bike sharing$"
-      ],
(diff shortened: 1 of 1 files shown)


bhousel:
Oh thanks! 
I originally added this for #3378, but it looks like the situation has improved in the last 4 years and now most of the Nextbike sub brands have been added with their own locations.  Seems safe to remove the exclusion 👍

PR #8450, opened 2023-07-26
Add 'Sun Tan City'

DivyaB1:
Hi,
I am adding 'Sun Tan City' - https://www.suntancity.com/

### DIFF
--- a/data/brands/leisure/tanning_salon.json
+++ b/data/brands/leisure/tanning_salon.json
@@ -1,0 +1,19 @@
+{
+  "properties": {
+    "path": "brands/leisure/tanning_salon",
+    "exclude": {"generic": ["^tanning salon$"]}
+  },
+  "items": [
+    {
+      "displayName": "Sun Tan City",
+      "locationSet": {"include": ["us"]},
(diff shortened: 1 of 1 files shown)


bhousel:
Oh, we've been adding these tanning salons into the shop/beauty file..

I actually don't know why there are different tags for this in OSM, but my memory is that the `shop=beauty` came first and then people added `leisure=tanning_salon` later and it was controversial. 

Today the wiki for both pages has changed a bunch, so I don't know.  I imagine people have changed the wiki over time to say whatever they want.
https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dtanning_salon
https://wiki.openstreetmap.org/wiki/Tag:shop%3Dbeauty

Anyway, can you put it in shop/beauty for now?  We can move them all later if that's what people want.  The actual tag values don't matter very much.

https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/shop/beauty.json

DivyaB1:
Got it, thanks! Closing this one and adding it to `shop=beauty`

PR #8453, opened 2023-07-27
Fix spelling and expand location set

playerEric1:
Fixed spelling of Lawson and expanded location set for niko and..., Daiso, and Ippudo.


tommylung:
Thanks

PR #8455, opened 2023-07-27
Add 'Sun Tan City'

DivyaB1:
Hi,

Adding 'Sun Tan City' to `shop=beauty` - https://www.suntancity.com/


DivyaB1:
Looks like 'Sun Tan City' is already present so its duplicated. The original entry does not have a wikidata ID, so I'll update it there.

PR #8467, opened 2023-08-01
sample_collection: Added Romanian "Clinica Sante"

DnaX:
This a sample collection chain of Romania. See here: https://www.clinica-sante.ro/filiale/

### DIFF
--- a/data/brands/healthcare/sample_collection.json
+++ b/data/brands/healthcare/sample_collection.json
@@ -23,6 +23,15 @@
       }
     },
     {
+      "displayName": "Clinica Sante",
+      "locationSet": {"include": ["ro"]},
+      "tags": {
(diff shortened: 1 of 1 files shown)


DnaX:
Very few of them are also analyzing laboratory, usually one laboratory per capital city. Should we have as well the brand for them?

```
healthcare=laboratory
sample_collection=yes
name=Clinica Sante
brand=Clinica Sante
brand:wikidata=Q121080834
```

arch0345:
> Very few of them are also analyzing laboratory, usually one laboratory per capital city. Should we have as well the brand for them?

I would definitely say so if they had good number of locations (10+). However, we have allowed brands in the past that had less locations (#7357)

PR #8474, opened 2023-08-03
Add Poštin paketomat 

u-spec-png:


arch0345:
Thanks for adding this brand! Please note however that without a `brand:wikidata` tag, this unfortunately won't get generated as a preset in iD. If this brand doesn't have a Wikidata item already, it's relatively easy to create one. We have a guide on the project's wiki which goes over how to edit Wikidata: https://github.com/osmlab/name-suggestion-index/wiki/Editing-Wikidata

PR #8482, opened 2023-08-05
add emergency phones, refine locationSet

k-yle:
- add 2 emergency phones
- use a more specific `locationSet` for some presets
- update 1 `network` tag, see comment below

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -99,12 +99,17 @@
       }
     },
     {
-      "displayName": "Auckland Transport",
-      "id": "aucklandtransport-68ffa1",
-      "locationSet": {"include": ["nz"]},
+      "displayName": "AT (Auckland Transport)",
+      "id": "at-8a95ba",
+      "locationSet": {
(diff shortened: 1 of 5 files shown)


data/transit/route/train.json
```diff
@@ -99,12 +99,17 @@
...
+        "include": [
+          "nz-auk.geojson",
+          "nz-wko.geojson"
+        ]
+      },
+      "matchNames": ["auckland transport"],
       "tags": {
-        "network": "Auckland Transport",
-        "network:short": "AT",
+        "network": "AT",
```
k-yle, line 112: NSI is trying to change `network=AT` to `network=Auckland Transport` again - see https://github.com/osmlab/name-suggestion-index/pull/8182#discussion_r1197786710

I'm not objecting to this change per-se, but it doesn't make sense to only change 1 out of 8 presets, and I think it would be important to prewarn the community that 3,000+ features are about to be retagged (especially since some software uses this tag value)

PR #8486, opened 2023-08-06
Update supermarket.json

asutekku:
Added Fresta, a japanese supermarket chain found mostly in Hiroshima prefecture


arch0345:
Thanks @asutekku and welcome!

PR #8495, opened 2023-08-10
Update LePerc

ffff-23:
"NTTル・パルク", including "NTT", is a more accurate name. "NTTル・パルク" is used on official sites and on fare signs.

At least the operator is "NTTル・パルク".

IMG_1911

mapillary: https://www.mapillary.com/app/?pKey=375911020697676

official website: https://www.le-perc.co.jp

### DIFF
--- a/data/operators/amenity/parking.json
+++ b/data/operators/amenity/parking.json
@@ -492,6 +492,27 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks for the update! 👍

PR #8498, opened 2023-08-11
Favorite Chicken

codeinabox:


Cj-Malone:
Just looking at the website, I might have called this one "Favorite Chicken & Ribs" or maybe just "Favorite".

codeinabox:
> Just looking at the website, I might have called this one "Favorite Chicken & Ribs" or maybe just "Favorite".

Shall we add some match names? Also what if people spell it as Favourite Chicken?

codeinabox:
> I don't think we actually need a match name, I added [brand:wikidata](https://www.openstreetmap.org/changeset/139870979), osm will fix itself overtime. Or we could fix it if we have consensus.

I am not familiar with this brand, I only stumbled upon it, so I am happy to defer to your expertise on this.

Cj-Malone:
I only know about it from scraping the website: https://github.com/alltheplaces/alltheplaces/pull/5656

So obviously if anyone has real knowledge of the brand, there input is very welcome.

Until then, I have just changed it to "Favorite Chicken & Ribs" in https://github.com/osmlab/name-suggestion-index/commit/6cc6fdd2295f16c576a44f601560ce7b524b745c

PR #8511, opened 2023-08-13
Added fuel company Jio BP

MG-LSJ:
Jio BP is a rebrand of Reliance Petroleum in partnership with BP in India


bhousel:
Thanks for the update! 👍

PR #8512, opened 2023-08-13
3 UK brands

joewestcott:
What should I do with the latter half of the id field? I've followed suit and made up a 6 char hex string.
Thanks.


bhousel:
You can skip the `id` field , our build script will generate them.

joewestcott:
@bhousel Ah, that makes way more sense. Amended, thanks.

PR #8517, opened 2023-08-14
Update Robidog entry

habi:


habi:
Should have been added to https://github.com/osmlab/name-suggestion-index/pull/8516, but I couldn't figure out how to do it...

habi:
Yes, because of https://github.com/osmlab/name-suggestion-index/issues/957#issuecomment-770282691

habi:
Thanks for the clarification, I didn't know that.
Changed accordingly.

habi:
Yes, that's true.
But a while ago I've updated the Robidog preset to include `bin=yes`, which I actually think is semantically more correct.
The are all waste baskets, too, but often their opening is small enough to only accept 🐶 💩  bags.

habi:
The 'issue' arose that EveryDoor is using nsi.guide, and searching for 'Robidog' there doesn't bring up the excrement bag dispenser anymore.

bhousel:
supersceded by #9247

PR #8521, opened 2023-08-14
add I'Velo on bicycle_rental.json

dan13l9:


data/brands/amenity/bicycle_rental.json
```diff
@@ -1726,6 +1726,18 @@
...
+      "id": "",
+      "locationSet": {"include": ["ro"]},
+      "tags": {
+        "amenity": "bicycle_rental",
+        "brand": "I'Velo",
+        "brand:wikidata": "Q119495434",
+        "network": "I'Velo",
+        "operator": "Asociația Green Revolution"
+        "matchNames": "Bucharest bike share"
+      }
```
Cj-Malone, line 1740: `matchNames` shouldn't be in the tags, but on the item.
```suggestion
      },
      "matchNames": ["Bucharest bike share"]
      }
```

PR #8528, opened 2023-08-15
add Bangkok MRT

OumKaZa:


data/transit/route/subway.json
```diff
@@ -37,6 +37,17 @@
         "route": "subway"
       }
     },
+    {
+      "displayName": "Bangkok MRT",
+      "id": "metropolitanrapidtransit-1e4f6f",
+      "locationSet": {"include": ["th"]},
+      "tags": {
+        "network": "Metropolitan Rapid Transit",
```
LaoshuBaby, line 45: I'm curious why you chose to use English instead of the native language of Thailand (since BTS Skytrain was just merged in #8529) รถไฟฟ้ามหานคร
At the same time, I also hope that the two PRs can cross-review, and I hope to see more mappers who are familiar with Thailand contribute local knowledge

PR #8530, opened 2023-08-16
Added Canadian money lender / payday loan shop

ShinyDog2636:


UKChris-osm:
The build is failing because `Cash Money` already exists in this file, so rather than adding it as a new entry, you can simply add the missing information (such as `brand:wikidata`) to that.

PR #8535, opened 2023-08-18
add Salomon on clothes.json

dan13l9:


data/brands/shop/clothes.json
```diff
@@ -7800,6 +7800,20 @@
...
+      "id": "",
+      "locationSet": {"include": ["001"]},
+      "tags": {
+        "alt_name": "Salomon Sports",
+        "brand": "Salomon",
+        "brand:wikidata": "Q2120822",
+        "name": "Salomon",
+        "shop": "clothes",
+        "clothes": "outdoor", "sportswear"
+        "
```
Cj-Malone, line 7815: So syntax wise, this should be:
```json
        "clothes": "outdoor;sportswear"
```
But it seems `sports` is the value, not `sportswear` https://taginfo.openstreetmap.org/keys/clothes#values
```suggestion
        "clothes": "outdoor;sports"
```
Also `outdoor` seems to have low usage. I don't know NSIs position on that.

dan13l9: yes, the brand it's very similar to "Columbia Sportswear" and "The Northface". They have outdoor tag and Patagonia also has it.

Cj-Malone: Prior art!

I feel like a Judge :p

PR #8540, opened 2023-08-20
New Zealand brands

Mortein:
Several New Zealand brands, and brands with stores in New Zealand.


UKChris-osm:
Thanks @Mortein, although I didn't think Lazy Boy stores traded in the UK.

Mortein:
> Thanks @Mortein, although I didn't think Lazy Boy stores traded in the UK.

It turns out they don't. I'll include it as a correction in another PR.

PR #8312, opened 2023-06-26
Cache all wikidata offical websites in wikidata.json

Cj-Malone:
Hey, so may as well say this publicly now.

I've been pushing TomTom to be a consumer of NSI. It's going quite well. I've been putting some of there suppliers through to show them how good it can be. Some suppliers (like [OPIS](https://en.wikipedia.org/wiki/Oil_Price_Information_Service)) already have a brand, so I've expanded some NSI `matchNames` and everything works great.

There are however some suppliers that don't have a brand label, or the brand is just bad. With the current one I'm looking at, it does however have good website urls. So I'm matching based on domain names. But it would be beneficial for me if the `wikidata.json` cache stored all the websites, not just one. Cases like `https://www.burgerking.co.uk/store-locator/store/restaurant_3614` can't match to `"Q177054": {"identities": {"website": "https://www.bk.com/"}}`

I've added a `officialWebsites` array, and not changed the current `website` to avoid breaking anything.

```json
    "Q177054": {
      "description": "global chain of hamburger fast food restaurants headquartered in Florida, United States",
      "identities": {
        "facebook": "burgerking",
        "instagram": "burgerking",
        "linkedin": "burger-king",
        "pinterest": "burgerkingkorea",
        "tiktok": "burgerking",
        "twitter": "burgerking_es",
        "website": "https://www.bk.com/",
        "youtube": "UC23ZqC2LTzl7dfOi6EmwJhg"
      },
      "label": "Burger King",
      "logos": {
        "facebook": "https://graph.facebook.com/burgerking/picture?type=large",
        "wikidata": "https://commons.wikimedia.org/wiki/Special:FilePath/Burger King 2020.svg"
      },
      "officialWebsites": [
        "https://www.bk.com/",
        "https://www.burgerking.co.uk/"
      ]
    },
```

I've also added `urlMatchPatterns`, while this probably wont be beneficial for matching, at least not right now. I have a personal goal of getting iD and JOSM to have a validator warning when there is a bad website tag on a POI.

```json
    "Q127120": {
      "description": "British multinational fitness center chain",
      "identities": {
        "facebook": "FitnessFirstGER",
        "instagram": "fitnessfirstger",
        "twitter": "fitnessfirst_uk",
        "website": "https://www.dwfitnessfirst.com",
        "youtube": "UCQ9asDDs--XEKYJK1G7Ug8A"
      },
      "label": "Fitness First",
      "logos": {
[...]

PR #8545, opened 2023-08-21
Include deprecated websites in wikidata.json

Cj-Malone:
As previously mentioned in #8312, I'm matching some non OSM datasets against NSI.

Unfortunately not all datasets are up to date, I'd like to be able to match deprecated websites such as `https://conradhotels3.hilton.com/` to `Q855525` and `conrad-ee1a43`.

### DIFF
--- a/scripts/build_wikidata.js
+++ b/scripts/build_wikidata.js
@@ -273,7 +273,7 @@ function processEntities(result) {
     }
 
     // P856 - official website
-    const officialWebsites = getClaimValues(entity, 'P856');
+    const officialWebsites = getClaimValues(entity, 'P856', true);
     if (officialWebsites) {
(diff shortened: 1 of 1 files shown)


Cj-Malone:
@bhousel can I poke you for an opinion on this

bhousel:
It seems okay..  I can see how matching deprecated websites would be useful, so I'm ok with merging it..

A thing I've wondered is whether all that work we are doing in `getClaimValues` can just be done by some function of the [`wikibase-sdk`](https://github.com/maxlath/wikibase-sdk) library that we are using.  But I haven't really had time to dig and see whether that's a possibility - might be something to consider for later!

Cj-Malone:
Thank you

PR #8549, opened 2023-08-22
Add Salesforce

DivyaB1:
Hi,

I am adding Salesforce - https://www.salesforce.com/ca/

### DIFF
--- a/data/brands/office/company.json
+++ b/data/brands/office/company.json
@@ -5,6 +5,17 @@
   },
   "items": [
     {
+     "displayName": "Salesforce",
+     "locationSet": {"include": ["001"]},
+     "tags": {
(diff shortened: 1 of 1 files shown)


data/brands/office/company.json
```diff
@@ -4,6 +4,17 @@
     "exclude": {"generic": ["^company$"]}
   },
   "items": [
+    {
+     "displayName": "Salesforce",
+     "locationSet": {"include": ["001"]},
+     "tags": {
+       "brand": "Salesforce",
+       "brand:wikidata": "Q941127",
```
1ec5, line 12: Salesforce isn’t a brand of IT company offices. They operate IT offices, so the `operator` key would be more appropriate than the `brand` key.

PR #8553, opened 2023-08-25
Update locationSet for supermarket in Hokkaido (Japan)

ffff-23:
These supermarkets exist only in Hokkaido, and for LUCKY, the locationSet was changed to that of Hokkaido since other area also exist.

https://www.hokuyu-lucky.co.jp/shops/


UKChris-osm:
Thanks @ffff-23

Issue #8462, opened 2023-07-30
Unable to add a space when using the search feature 


UKChris-osm:
Currently on [nsi.guide](https://nsi.guide) using the search feature doesn’t allow the space bar to add a space when using an iPad or Android phone.

kjonosm:
Same problem with Safari on macOS...

Lokeshsri11:
Please assign me this issue #8462

PR #8554, opened 2023-08-25
Fixed spacebar key issue

vrun1208:
Fixed spacebar key issue on the search field #8462

### DIFF
--- a/app/src/Filters.jsx
+++ b/app/src/Filters.jsx
@@ -3,6 +3,7 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faFilter } from '@fortawesome/free-solid-svg-icons';
 
 import { AppContext, getFilterParams } from './AppContext';
+import { pathExistsSync } from 'fs-extra';
 
 
 export function Filters() {
@@ -10,7 +11,7 @@ export function Filters() {
   const params = context.params;
   const setParams = context.setParams;
   const filters = getFilterParams(params);
-  const klass = 'filters' + (Object.keys(filters).length ? ' active' : '');
+  const klass = 'filters' + (Object.keys(filters).length ? 'active' : '');
 
   return (
     <div className={klass}>
@@ -29,7 +30,7 @@ export function Filters() {
 
       <span className='field'>
         <label htmlFor='inc'>Incomplete:</label>
-        <input type='checkbox' id='inc' name='inc' checked={filters.inc === 'true'} onChange={filtersChanged} />
+        <input type='checkbox' id='inc' name='inc' checked={filters.inc === 'true'} onChange={Keychanged} />
       </span>
 
       <span className='field'>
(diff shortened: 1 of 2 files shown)


UKChris-osm:
Thanks @vrun1208 

Are you able to deploy this to GitHub pages under your repository so I can see it in action before merging?

vrun1208:
hi @UKChris-osm,  sure working on it!

UKChris-osm:
At the moment on your fork at [https://vrun1208.github.io/name-suggestion-index/](https://vrun1208.github.io/name-suggestion-index/) I see that pressing the space bar adds an underscore, rather than honouring the space bar.

vrun1208:
actually the issue is that the brands or categories does not contain spaces as they are seperated with underscores. for example:- car_rental, car_repair.

and if we try to search with spaces in the input field they provide no more suggestions for the given keyword. so I replaced the space with underscore

Lokeshsri11:
> actually the issue is that the brands or categories does not contain spaces as they are seperated with underscores. for example:- car_rental, car_repair.
> 
> and if we try to search with spaces in the input field they provide no more suggestions for the given keyword. so I replaced the space with underscore

Hey @vrun1208 There are many companies that use space in there name like Burger king, So according to me, we should give all the options in the search bar for spacing as well as underscore.

app/src/Filters.jsx
```diff
@@ -3,14 +3,15 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faFilter } from '@fortawesome/free-solid-svg-icons';
 
 import { AppContext, getFilterParams } from './AppContext';
+import { pathExistsSync } from 'fs-extra';
```
bhousel, line 6: I can't think of a reason for this import to be here?

vrun1208: Yes you're right. i'm not sure of it, maybe it was created when i was defining function. i'll remove it.

vrun1208:
Hey! @UKChris-osm, @bhousel  check out now, everything is working fine!

UKChris-osm:
Thanks @vrun1208 this seems to be working as intended now.

UKChris-osm:
Hi again @vrun1208 

When you removed the single blank space before `' active' : '');`` was there a reason for this? 
`const klass = 'filters' + (Object.keys(filters).length ? 'active' : '');`

It's currently causing a small problem on the live web site.

vrun1208:
hi @UKChris-osm, css was misbehaving while running in local, so I removed the space before `' active': ' '` and that works for me . And it doesn't affect when I deployed it as you saw earlier.

UKChris-osm:
It's odd, as looking at your local copy you have deployed, your version isn't removing the space `<div class="filters active">`, but the version on our live site is removing the space, leaving us with `<div class="filtersactive">`.

vrun1208:
Okay let me see, get back to you soon

PR #8573, opened 2023-08-31
Add Aflac brand

Guerme:
Adding Aflac brand - https://www.aflac.com/


arch0345:
Thanks @Guerme and welcome!

PR #8579, opened 2023-09-01
Add 'Rush Truck Center'

DivyaB1:
Hi,

I am adding Rush Truck Centers - https://www.rushtruckcenters.com/

### DIFF
--- a/data/brands/shop/truck_repair.json
+++ b/data/brands/shop/truck_repair.json
@@ -5,6 +5,16 @@
   },
   "items": [
     {
+     "displayName": "Rush Truck Centers",
+     "locationSet": {"include": ["us"]},
+     "tags": {
+       "brand": "Rush Truck Centers",
+       "brand:wikidata": "Q120637525",
+       "name": "Rush Truck Centers",
+       "shop": "truck_repair"
+     }
+   },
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Do `Rush Truck Centers` have separate locations for their repair centers and showrooms?

DivyaB1:
Hi,

No, they do not. It's one location. How can we add multiple categories per brand?

Thanks,

bhousel:
> No, they do not. It's one location. How can we add multiple categories per brand?

For now we just have to put the item into 2 categories.

I'd just add this brand to both:
`shop=truck` and `shop=truck_repair`

Most of the car dealers are already set up like this also.  Mappers can place a point at the showroom and another point at the repair center.  This is beneficial anyway because these things often have separate phone numbers and opening hours and other attributes.

Cj-Malone:
There is https://github.com/osmlab/name-suggestion-index/pull/8578 doing that.

The alternative would be [service:vehicle:truck_repair](https://taginfo.openstreetmap.org/keys/service%3Avehicle%3Atruck_repair#overview). But the 2 pois makes sense to me, that's how I map fuel stations and shops.

PR #8584, opened 2023-09-04
Added several NYC based brands

hdevine825:
Added a selection of local NYC based chains.

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -2362,6 +2362,20 @@
       }
     },
     {
+      "displayName": "Matto Espresso",
+      "locationSet": {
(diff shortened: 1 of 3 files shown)


data/brands/amenity/cafe.json
```diff
@@ -3328,6 +3342,20 @@
...
     },
+    {
+      "displayName": "The Bean",
+      "locationSet": {
+        "include": ["new_york_city.geojson"]
+      },
+      "tags": {
+        "amenity": "cafe",
+        "brand": "The Bean",
+        "brand:wikidata": "Q122209700",
```
arch0345, line 3353: It looks like this brand only has 3 locations in NYC: https://www.thebean.nyc/hours-locations

In the future, I'd try to stick with adding brands with at least 10 locations.

PR #8591, opened 2023-09-05
Rewrite matcher of id-tagging-schema presets

serhii-muchychka:
**This is an alternative for https://github.com/osmlab/name-suggestion-index/pull/8586**

Now it takes into account all presets, not just those with no more than 3 tags

Now categories will be correctly defined for the following presets:

- amenity/pub/microbrewery
- amenity/recycling_container
- amenity/recycling_centre
- amenity/restaurant/steakhouse
- shop/car/second_hand
- shop/clothes/second_hand
- shop/laundry/self_service
- shop/supermarket/organic
- tourism/information/office
- tourism/information/map
- amenity/bicycle_parking/lockers
- amenity/parking/park_ride
- amenity/toilets/disposal
- man_made/mast/communication/mobile_phone
- natural/water/reservoir

### DIFF
--- a/scripts/dist_files.js
+++ b/scripts/dist_files.js
@@ -220,6 +220,13 @@ function buildIDPresets() {
     // NOTE: here we intentionally exclude `:wikidata`, in `matcher.js` we do not.
     const notName = /:(colour|type|left|right|etymology|pronunciation|wikipedia|wikidata)$/i;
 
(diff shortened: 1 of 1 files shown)


bhousel:
Nice!  This looks ok to me 👍 
Let's try it out and see if we get better matches..

PR #8592, opened 2023-09-05
Adding McLaren Brand to Car and Car_Repair

ckpiper:


arch0345:
Thanks @ckpiper and welcome!

PR #8598, opened 2023-09-07
Update bicycle_rental (french location) 

phd0:
update `{"include": ["fx"]}` for 8 operators, with only local coords of town, to prevent bad suggestion in iD.
image


phd0:
houps 
i made some error : i forget some spaces in some line after `:`

UKChris-osm:
Thanks @phd0 

Just an FYI, we no longer include any `wikipedia` references in the Name Suggestion Index, so this can be removed.

PR #8599, opened 2023-09-07
Update car_repair.json added 123 PARE-BRISE

jhutton1:
added 123 PARE-BRISE - https://www.123parebrise.fr/

### DIFF
--- a/data/brands/shop/car_repair.json
+++ b/data/brands/shop/car_repair.json
@@ -25,6 +25,16 @@
     }
   },
   "items": [
+            {
+      "displayName": "123 PARE-BRISE",
+      "locationSet": {"include": ["fr"]},
+      "tags": {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Thanks @jhutton1 

Is "_PARE-BRISE_" definitely all capital letters? I see there aren't any currently like that on the OSM Map, but a few written as "123 Pare-Brise"!

JaimeHutton-TomTom:
The website had all caps.  That's where I got that from.  Perhaps it's just the logo.  Do you suggest I change to mixed?

UKChris-osm:
Looking at their web site I can see they used mixed case when referring to themselves in text form (page <title>, text further down the page) and as other mappers have already used mixed case, I'd suggest using that instead of all capitals.

PR #8601, opened 2023-09-07
Update general.json added Müller

jhutton1:
added Müller - https://www.mueller.at/meine-filiale/

### DIFF
--- a/data/brands/shop/chemist.json
+++ b/data/brands/shop/chemist.json
@@ -533,6 +533,16 @@
       }
     },
     {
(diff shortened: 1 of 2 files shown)


UKChris-osm:
Is this the same as [this entry](https://nsi.guide/index.html?t=brands&k=shop&v=chemist&tt=m%C3%BCller)?

jhutton1:
that entry is a different url.  this is for https://www.mueller.at/

PR #8606, opened 2023-09-07
Update audiologist.json added Acustica Medica

jhutton1:
https://www.acusticamedica.pt/


arch0345:
It looks like this brand also sells hearing aids, would it make more sense for this to be in `shop/hearing_aids`?

PR #8608, opened 2023-09-07
Update employment_agency.json added AMS Jobseekers Austria

jhutton1:
https://www.ams.at/


arch0345:
It looks like the name listed on their website is 'Arbeitsuchende Österreich', the brand/name tags should probably be changed to this.

PR #8610, opened 2023-09-07
Update carpet.json added Top Cartpets & Floors

jhutton1:


arch0345:
It looks like this brand sells other types of flooring, would it make sense to move this to `shop/flooring`?

PR #8611, opened 2023-09-07
Update flooring.json added Top Carpets & Floors

jhutton1:


UKChris-osm:
Do _Top Carpets and Floors_ have separate stores for their floor and carpet services?

JaimeHutton-TomTom:
No.  It's just one store for both.  I created a pull request for each category.

PR #8612, opened 2023-09-07
Update car_parts.json add Big Brand Tire & Service

jhutton1:
https://bigbrandtire.com/


arch0345:
This brand seems to mostly sell tires. Can you move this to `shop/tyres`?

PR #8615, opened 2023-09-08
Update tyres.json remove Tiger Wheel & Tire.  Added to car_parts.json PR#8604

jhutton1:


UKChris-osm:
Personally, I think "_Tiger Wheel & Tire_" is currently correctly in the tyre shop category, so I'll wait to see what others think before merging.

PR #8624, opened 2023-09-09
fix: add french brand entry for Leons (Meubles Léon)

wolfy1339:


Cj-Malone:
Would this be better as 1 multilingual entry?

wolfy1339:
No. Since the display name is also localized it wouldn't work

There are other instances of brands having 2 separate entries, one in English and one in french

bhousel:
Ah, yeah this is ok.. we have a handful of other brands localized for Quebec..

The example I always think of is KFC / PFK:
https://nsi.guide/index.html?t=brands&k=amenity&v=fast_food#pfk-32490c
https://nsi.guide/index.html?t=brands&k=amenity&v=fast_food#kfc-9cadb7

PR #8628, opened 2023-09-10
Remove secondary names from template entries

serhii-muchychka:
Also would like to add functionality that will make the following valid:
`"operator*": "{source.tags.brand*}"`
but it takes more time ):

### DIFF
--- a/data/brands/advertising/totem.json
+++ b/data/brands/advertising/totem.json
@@ -11,7 +11,7 @@
         "amenity": "",
         "brand": "{source.tags.brand}",
         "brand:wikidata": "{source.tags.brand:wikidata}",
-        "name": ""
+        "R\\name": ""
       }
     }
   ]
--- a/data/brands/amenity/atm.json
+++ b/data/brands/amenity/atm.json
@@ -12,7 +12,7 @@
       "templateSource": "brands/amenity/bank",
       "templateTags": {
         "amenity": "atm",
-        "name": "",
+        "R\\name": "",
         "operator": "{source.tags.brand}",
         "operator:wikidata": "{source.tags.brand:wikidata}"
       }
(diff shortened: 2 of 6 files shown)


bhousel:
I kinda don't like this.. I wonder if we could just use normal regular expressions instead of introducing a special star syntax for matching?

I've been thinking about the idea of just building a separate library for matching tags, similar to how [MapCSS selectors](https://wiki.openstreetmap.org/wiki/MapCSS/0.2#Selectors) work.. But maybe less weird/complicated.

serhii-muchychka:
Ok. I will replace it with a regular expression

I did by analogy with the "mv" UNIX command. In this case, when  template group of tags corresponds to a fixed value, it is very easy to replace with regexp. But when we want to replace type `"operator*": "{source.tags.brand*}"` (which is not yet done), it will be more difficult with regexp

serhii-muchychka:
@bhousel maybe you will like this variant:
```
{
  "templateSource": "brands/amenity/bank",
  "templateTags": {
    "amenity": "atm"
  },
  "templateRecursiveTags": {
    "operator": "{source.tags.brand}"
  }
}
```
Operations will be performed in such order:
1. Copy all source tags
2. Performing `templateTags`
3. Performing `templateRecursiveTags`

"recursively" means base tag and all tags separated by ":"

For example `official_name` will not be recursive for the `name`, but `name:en` will be

Or maybe you have a different vision of what it should look like? I'll be happy for any idea 🙂

bhousel:
Sorry closing this as it's pretty stale at this point... I forgot about this, but I still don't like the proposed syntax and I've kind of forgotten what the problem was that this was supposed to solve.

bhousel:
Oh right.. it's that a template entry like this one removes `name` but not tags like `name:en`.

```js
    {
      "note": "Auto create ATMs for all banks (remove names per #7236)",
      "templateExclude": ["monobank"],
      "templateSource": "brands/amenity/bank",
      "templateTags": {
        "amenity": "atm",
        "name": "",
        "operator": "{source.tags.brand}",
        "operator:wikidata": "{source.tags.brand:wikidata}"
      }
    },
```

Ok now I do remember why the original PR was like `"name*"`, and why I thought maybe something like `"/name/"` would have been easier to work with.

PR #8630, opened 2023-09-11
Update vending_machine.json

Mannivu:
Fix brand name (see http://www.aqvagold.it)

### DIFF
--- a/data/brands/amenity/vending_machine.json
+++ b/data/brands/amenity/vending_machine.json
@@ -354,8 +354,7 @@
       "locationSet": {"include": ["it"]},
       "tags": {
         "amenity": "vending_machine",
-        "brand": "Distributore di acqua Aqvagold",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/vending_machine.json
```diff
@@ -354,8 +354,7 @@
       "locationSet": {"include": ["it"]},
```
arch0345, line 354: It looks like this brand was collected from the [planet file](https://taginfo.openstreetmap.org/search?q=Distributore+di+acqua+Aqvagold#values). A `matchName` should be added to prevent this from being automatically readded and to match this entry to existing features with this name.
```suggestion
      "locationSet": {"include": ["it"]},
      "matchNames": ["Distributore di acqua Aqvagold"],
```

arch0345 (changes requested):

PR #8640, opened 2023-09-14
Add The Paper Store

AntiCompositeNumber:
ref https://locations.thepaperstore.com/

### DIFF
--- a/data/brands/shop/gift.json
+++ b/data/brands/shop/gift.json
@@ -319,6 +319,28 @@
       }
     },
     {
+      "displayName": "The Paper Store",
+      "locationSet": {
+        "include":
+          [
+            "us-ct.geojson",
+            "us-fl.geojson",
+            "us-ma.geojson",
+            "us-me.geojson",
+            "us-nh.geojson",
+            "us-nj.geojson",
+            "us-ny.geojson",
+            "us-ri.geojson"
+          ]
+        },
+        "tags": {
+          "brand": "The Paper Store",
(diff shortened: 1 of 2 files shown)


AntiCompositeNumber:
Looks like `brand:wikidata=Q65068381` is already used for "The Paper _Shop_", which is something different. I don't see a relevant Wikidata item with that name, and I can't really find anything online about it.

AntiCompositeNumber:
Looks like the "The Paper Shop" addition was added in #7926 (cc @davidhicks). Based on alltheplaces/alltheplaces/pull/5058 it looks like this was probably a typo. In that case I still think The Paper Store should be marked as `shop=gift`, not `shop=variety_store`. It's been a while since I've been in a Paper Store, but ["a retail shop that sells inexpensive items, usually with a single price point for all items in the store"](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dvariety_store) doesn't really match.

AntiCompositeNumber:
There was only 1 entity matching `"brand:wikidata"=Q65068381 and brand="The Paper Shop" global`, which I confirmed is a The Paper Store using Mapillary and have corrected. So no need to add matchNames.

arch0345:
Thanks for catching this! It looks like most locations are currently tagged as `shop=gift`: https://overpass-turbo.eu/s/1AuD

davidhicks:
Thank you for the correction. I think you're right that I confused "The Paper Shop" (made up name) with "The Paper Store" (correct name) when proposing #7926

Issue #8576, opened 2023-09-01
Filters on nsi.guide are jumping around now


bhousel:
They move when you type in them 🤔 

![filters move](https://github.com/osmlab/name-suggestion-index/assets/38784/4a662db4-d771-408d-bbc0-818a40ca0769)

UKChris-osm:
It looks like the class setting is switched when you type, before it's `class="filters"` but when you type it becomes `class="filtersactive"`

vrun1208:
Hi @UKChris-osm, seems I figured it out, actually there is junk code in `docs/index.js` and I made a little change there and it is working. will soon open a PR

PR #8645, opened 2023-09-16
fixed-filters-class

vrun1208:
#8576 
fixed the `class=filters` issue that caused the css not set to `active` results in the inappropriate ui while typing.

It was look like `class=filtersactive` which should be looks like `class=filters active`


app/src/Filters.jsx
```diff
@@ -10,7 +10,7 @@ export function Filters() {
   const params = context.params;
   const setParams = context.setParams;
   const filters = getFilterParams(params);
-  const klass = 'filters' + (Object.keys(filters).length ? 'active' : '');
+  const klass = 'filters' + (Object.keys(filters).length ? ' active' : '');
```
UKChris-osm, line 13: Was this the only edit you made? Did you run _npm run appbuild_ as well?

vrun1208: @UKChris-osm No, I didn't run , but I just hit `npm run appbuild` in local and yes it is working!
also I manually made a change in the `docs/index.js`

UKChris-osm: What manual edits did you make to `index.js`? I'm not sure how that file is generated, but I thought it was all automated, and so manual edits might disappear once `npm run appbuild` is run again.

vrun1208: Screenshot 2023-09-19 203049
this one! 
Yes, you're right it is auto generated but the fix I made before `npm run appbuild` and after totally working fine.
making change in `filters.js` alone was enough as the issue was fixed, but to made sure I also dived into `index.js`.

UKChris-osm: As `npm run appbuild` should generate the desired changes, it's best to leave that file alone, and the change to the filters class should only be needed in the one file, the initial one you edited.

Unfortunately I'm without a computer at the moment, so can't run any of the npm software myself.

What changes were made to the files below, as I've lost track? It might be best to start afresh, and just edit the one line for the filters class, then run `npm run appbuild `and commit that as a PR.

vrun1208: I have done what you said in my latest commit. kindly check it

PR #8652, opened 2023-09-18
Update massage.json added Massage Addict

jhutton1:
https://www.massageaddict.ca/

### DIFF
--- a/data/brands/shop/massage.json
+++ b/data/brands/shop/massage.json
@@ -5,6 +5,16 @@
   },
   "items": [
     {
+      "displayName": "Massage Addict",
+      "locationSet": {"include": ["ca"]},
+      "tags": {
+        "brand": "Massage Addict",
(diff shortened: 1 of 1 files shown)


kjonosm:
I wonder whether `healthcare=alternative` + [`healthcare:speciality=acupuncture;reflexology;...`](https://wiki.openstreetmap.org/wiki/Key:healthcare:speciality#Specialities_for_healthcare=alternative) might be a better tag for this brand? What do you think @jhutton1?

jhutton1:
I believe they primarily do massage therapy.  Looks like they additionally have chiropractor/acupuncture/reflexology services.  I could add those to the massage.json as additional tags?   How would that look?

arch0345:
I think leaving those tags out is fine - some of the other existing brands we have in shop/massage also offer other services and don't have any additional tags. Merging as is for now

PR #8655, opened 2023-09-18
Move Peterbilt from shop=car to shop=truck

AntiCompositeNumber:
Per note on https://wiki.openstreetmap.org/wiki/Tag:shop=car preferring shop=truck

### DIFF
--- a/data/brands/shop/truck.json
+++ b/data/brands/shop/truck.json
@@ -43,6 +43,17 @@
       }
     },
     {
+      "displayName": "Peterbilt",
+      "id": "peterbilt-13eeb7",
+      "locationSet": {"include": ["001"]},
+      "tags": {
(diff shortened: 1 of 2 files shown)


data/brands/shop/truck.json
```diff
@@ -42,6 +42,17 @@
         "shop": "truck"
       }
     },
+    {
+      "displayName": "Peterbilt",
+      "id": "peterbilt-13eeb7",
+      "locationSet": {"include": ["001"]},
```
arch0345, line 48: It looks like most locations are still [tagged as `shop=car`](https://overpass-turbo.eu/s/1AH3). This will match existing locations tagged as `shop=car` with this entry.
```suggestion
      "locationSet": {"include": ["001"]},
      "matchTags" ["shop/car"],
```

PR #8658, opened 2023-09-19
Update company.json added SpeedPro

jhutton1:
https://www.speedpro.com/


arch0345:
Based on their website, wouldn't this brand fit better in `craft/signmaker`?

PR #8674, opened 2023-09-22
feat: use json response for facebook logo fetching

### LINKED ISSUE #8380: See whether we can avoid downloading the Facebook profile pics with `redirect=false`
>I'm getting a little off topic, but `fetchFacebookLogo` currently downloads the image. If we add `redirect=false` we get JSON [eg](https://graph.facebook.com/easyHotel/picture?redirect=false). It's probably smaller, and so faster.

_Originally posted by @Cj-Malone in https://github.com/osmlab/name-suggestion-index/issues/8373#issuecomment-1627824620_

kymckay:
- This speeds up the wikidata sync process since fetching the json requires less bandwidth than fetching images.
- This removes a bunch or erroneous default images that were previously being set on entries in the index (there are defaults other than the question mark).

closes #8380

### DIFF
--- a/scripts/build_wikidata.js
+++ b/scripts/build_wikidata.js
@@ -605,19 +605,19 @@ function fetchFacebookLogo(qid, username) {
   const m = username.match(/-(\d+)$/);
   if (m) userid = m[1];
 
-  return fetch(logoURL, fetchOptions)
-    .then(response => {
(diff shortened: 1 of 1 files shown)


bhousel:
This looks great!! Thanks @kymckay 👍

PR #8681, opened 2023-09-23
Link to YouTube when there is handle but not channel in wikidata

arrival-spring:
Noting for example https://www.wikidata.org/wiki/Q7190735, I thought that there could be some entities with a YouTube handle set, but without the YouTube channel ID set (I'm not sure how to find statistics on this). Also, I don't know that there is a bot automatically setting one from the other on Wikidata.

### DIFF
--- a/app/src/CategoryRowSocialLinks.jsx
+++ b/app/src/CategoryRowSocialLinks.jsx
@@ -46,8 +46,12 @@ export function CategoryRowSocialLinks(props) {
       <a key='pinterest' target='_blank' href={href}><FontAwesomeIcon icon={faPinterestSquare} size='lg' /></a>
     );
   }
-  if (props.youtube) {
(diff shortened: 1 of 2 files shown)


Cj-Malone:
The other option would be to have `build_wikidata.js` build the url. I have no strong preference either way.

tommylung:
Each YouTube channel must have the channel ID. The YouTube handle field will duplicate with the ID. You can use the [third party website](https://commentpicker.com/youtube-channel-id.php) to find and fill the channel ID on the Wikidata.

bhousel:
Thanks, I think this seems ok to collect them both if they are 2 different claims in Wikidata.  
I remember a while ago I mostly ignored Youtube in this code because I couldn't figure out how to easily pick out the channel ids.

PR #8685, opened 2023-09-25
Update doityourself.json add "Doras" (French only)

phd0:
https://www.doras.fr/agences/


kjonosm:
NSI doesn't store `website` tags...

Cj-Malone:
Also I think that's the wrong wikidata

PR #8688, opened 2023-09-26
Create place_of_worship.json and Add Uniting Church in Australia

Guerme:
Added place_of_worship.json to amenity category
Also added Uniting Church in Australia - https://uniting.church/


bhousel:
I think we're excluding the religious items for now, though we do still want this eventually.
#5960 was a first attempt at adding a tree for them - I should probably check it out and see what's missing from it.   I admit I don't remember now what the blockers were.

PR #8690, opened 2023-09-26
Update employment_agency.json Add Spherion

Guerme:
Adding Spherion - https://www.spherion.com/


kjonosm:
This is an [`office=employment_agency`](https://wiki.openstreetmap.org/wiki/Tag:office%3Demployment_agency)

PR #8691, opened 2023-09-26
Update cosmetics.json add Charlotte Tiilbury

jhutton1:


kjonosm:
The correct location code for the UK is `gb`

PR #8693, opened 2023-09-26
Update charity.json added Big Brothers Big Sisters Lone Star

jhutton1:


kjonosm:
Do they rather offer social or community services as an [amenity=social_facility](https://wiki.openstreetmap.org/wiki/Key:social_facility)?

PR #8694, opened 2023-09-26
Update company.json added Morgan, Lewis & Bockius

jhutton1:
https://www.morganlewis.com/


kjonosm:
How about [`office=consulting`](https://wiki.openstreetmap.org/wiki/Tag:office%3Dconsulting)?

PR #8695, opened 2023-09-26
Update doityourself.json added Beacon

jhutton1:


kjonosm:
Like Wikipedia the `brand=*` and `name=*` or `operator=*` tags usually don't include the legal entity type.

PR #8701, opened 2023-09-27
Add Eneos Frontier

DivyaB1:
Hi,

I'm adding Eneos Frontier - https://www.eneos-frontier.co.jp/

### DIFF
--- a/data/brands/shop/car_repair.json
+++ b/data/brands/shop/car_repair.json
@@ -31,7 +31,8 @@
 	      "tags": {
 	        "brand": "Eneos Frontier",
 	        "brand:wikidata": "Q11197713",
-	        "name": "Eneos Frontier",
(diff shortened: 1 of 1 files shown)


data/brands/shop/car_repair.json
```diff
@@ -25,6 +25,16 @@
     }
   },
   "items": [
+    	{
+	      "displayName": "Eneos Frontier",
+	      "locationSet": {"include": ["jp"]},
```
LaoshuBaby, line 30: If this is a Japan-only entry, I would recommend following https://wiki.openstreetmap.org/wiki/JA:Naming_sample to complete other multilingual tags. Of course, this is not mandatory, but it can reduce the workload of other maintainers who mainly maintain Japan-related entries.

PR #8705, opened 2023-09-27
Add Randstad

DivyaB1:
Hi,

I'm adding Randstad - https://www.randstad.com/

Thanks,

### DIFF
--- a/data/brands/office/consulting.json
+++ b/data/brands/office/consulting.json
@@ -6,11 +6,14 @@
   "items": [
     {
 	    "displayName": "Randstad",
-	    "locationSet": {"include": ["001"]},
+	    "locationSet": {"include": ["us","ca","mx","au","nz","in","cn","jp","my","sg","hk",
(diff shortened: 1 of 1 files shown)


arrival-spring:
Thanks for adding this, mostly looks good to me.

I think name and brand should both be capitalised, despite their logo having lower case r, in text on their website they use upper case 

Also, I think the location set for this needs to be limited, see https://www.randstad.com/find-randstad-in-your-country/

DivyaB1:
Updated brand name with a capital 'R' and added countries separately

bhousel:
thanks @DivyaB1 👍

kjonosm:
@DivyaB1 and @bhousel Randstad is now listed in NSI as both `office=consulting` and `office=employment_agency`.
Which would be the right office type to keep?

DivyaB1:
I believe it should be `office=employment_agency`.

PR #8709, opened 2023-09-28
Star transit networks france

QuentinS33:
In France, there are two transit networks called STAR, one in Rennes (NW of France) and the other one in Roanne (East of France).
To avoid confusing, i improved the locationSet of STAR Rennes and I added the network STAR Roanne
I changed the id of STAR Roanne and i am not sure if he's ok


bhousel:
Thanks @QuentinS33 , looks good 👍 
no need to worry about the `id` values, they get rebuilt automatically if needed.

PR #8717, opened 2023-10-01
Check for duplicating features between country-coder and NSI

serhii-muchychka:


LaoshuBaby (approved):

bhousel:
There's nothing really wrong with this, but it's kind of unnecessary, so I'm going to close it, sorry.

tl;dr: `location-conflation` has special support for Alaska and Hawaii, and NSI has added geojsons for all the US states, including Alaska and Hawaii, and uses the geojsons instead, which is technically redundant.

PR #8718, opened 2023-10-01
Rename "denn's Biomarkt" to "Denns BioMarkt"

nicolashohm:
See: https://de.wikipedia.org/w/index.php?title=Denns_Biomarkt&diff=prev&oldid=212753611

I have already updated [Wikidata](https://www.wikidata.org/wiki/Q48883773)!

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -1920,13 +1920,15 @@
       }
     },
     {
-      "displayName": "denn's Biomarkt",
+      "displayName": "Denns BioMarkt",
       "id": "dennsbiomarkt-6367d0",
       "locationSet": {"include": ["at", "de"]},
       "tags": {
-        "brand": "denn's Biomarkt",
+        "brand": "Denns BioMarkt",
         "brand:wikidata": "Q48883773",
-        "name": "denn's Biomarkt",
+        "name": "Denns BioMarkt",
+        "operator": "Dennree GmbH",
+        "operator:wikidata": "Q1189641",
         "organic": "only",
         "shop": "supermarket"
       }
--- a/data/operators/man_made/surveillance.json
+++ b/data/operators/man_made/surveillance.json
@@ -923,12 +923,12 @@
       }
     },
     {
-      "displayName": "Denn's Biomarkt?",
+      "displayName": "Denns BioMarkt",
       "id": "dennsbiomarkt-9c1a6c",
       "locationSet": {"include": ["001"]},
       "tags": {
         "man_made": "surveillance",
-        "operator": "Denn's Biomarkt?"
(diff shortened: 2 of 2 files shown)


data/operators/man_made/surveillance.json
```diff
@@ -923,12 +923,12 @@
...
     },
     {
-      "displayName": "Denn's Biomarkt?",
+      "displayName": "Denns BioMarkt",
       "id": "dennsbiomarkt-9c1a6c",
       "locationSet": {"include": ["001"]},
       "tags": {
         "man_made": "surveillance",
-        "operator": "Denn's Biomarkt?"
+        "operator": "Denns BioMarkt"
```
nicolashohm, line 931: I don't think this makes much sense, but I'm not sure, so I just updated it

data/operators/man_made/surveillance.json
```diff
@@ -923,12 +923,12 @@
       }
     },
     {
-      "displayName": "Denn's Biomarkt?",
+      "displayName": "Denns BioMarkt",
       "id": "dennsbiomarkt-9c1a6c",
       "locationSet": {"include": ["001"]},
```
nicolashohm, line 928: What does `001` means? Does it make sense to change it to `["at", "de"]` to be inline with the supermarket?

data/brands/shop/supermarket.json
```diff
@@ -1920,13 +1920,15 @@
...
+      "displayName": "Denns BioMarkt",
       "id": "dennsbiomarkt-6367d0",
       "locationSet": {"include": ["at", "de"]},
       "tags": {
-        "brand": "denn's Biomarkt",
+        "brand": "Denns BioMarkt",
         "brand:wikidata": "Q48883773",
-        "name": "denn's Biomarkt",
+        "name": "Denns BioMarkt",
+        "operator": "Dennree GmbH",
```
nicolashohm, line 1930: According to [Wikipedia](https://de.wikipedia.org/wiki/Denns_Biomarkt) (translated):
> Denns Biomarkt is an organic supermarket chain operated by the German Dennree GmbH.

data/operators/man_made/surveillance.json
arch0345: These 3-digit locationSets are [UN M49 codes](https://en.wikipedia.org/wiki/UN_M49). In this case, `001` would include the entire world. If this operator is only present in Austria and Germany, then it should be changed. You can use [this website](https://location-conflation.com/?referrer=nsi) to help visualize locationSets.

nicolashohm: thanks! updated.

arch0345:
Thanks for your first contribution to NSI!

PR #8723, opened 2023-10-02
added flags and wikidata ID

wvanderp:
In this PR, I've added the flags of the Belgian regions and provinces and added wikidata IDS to some more French provides

Some special attention is needed for BE-WLX - Luxembourg (BE). I could not get it to build without changing the flag's name to `province of Luxembourg`, I hope this is correct.

Looking forward to your feedback

### DIFF
--- a/data/flags/man_made/flagpole.json
+++ b/data/flags/man_made/flagpole.json
@@ -459,8 +459,8 @@
     },
     {
       "displayName": "BE-BRU - Brussels-Capital Region",
-      "id": "brusselscapitalregion-e5dc93",
-      "locationSet": {"include": ["001"]},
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks for this!  It looks pretty ok to me, although for regional flags we usually scope them to the country where they would normally be displayed.. so I'd change the `"include": ["001"]` to `"include": ["be"]`

bhousel:
> we usually scope them to the country where they would normally be displayed

Oh!  This would fix the Luxembourg issue too.  
You could make the flag name just be "Luxembourg" if it had a different locationSet.

The code would then be able to generate a different id for Belgian Luxembourg vs the country Luxembourg.

PR #8727, opened 2023-10-03
Update fashion_accessories.json added Bottega Veneta

jhutton1:


data/brands/shop/fashion_accessories.json
```diff
@@ -6,6 +6,16 @@
     }
   },
   "items": [
+    {
+      "displayName": "Bottega Veneta",
+      "locationSet": {"include": ["it"]},
```
arch0345, line 11: Based on their website, they have many other locations outside of Italy (https://www.bottegaveneta.com/en-us/storelocator)
```suggestion
      "locationSet": {"include": ["001"]},
```

arch0345 (changes requested):

Issue #6552, opened 2022-05-05
`Petrol` is not being recognized as brand


Dimitar5555:
When a gas station has `operator=Petrol`, iD doesn't suggest the correct brand (https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/amenity/fuel.json#L5351-L5363).
![image](https://user-images.githubusercontent.com/19364673/166923845-5912554b-802f-43c3-aca8-f439f51b3fb0.png)

When the I add `name=Petrol` it suggests to remove it since it's too generic.
![image](https://user-images.githubusercontent.com/19364673/166924068-d737b7e3-6791-4eaf-a4ac-fbf2f25b04c9.png)

Is it possible to make it match with the brand instead of triggering warnings?

bhousel:
Problem is this line in `buildMatchIndex()`, it means that the generic name never gets indexed.  I'll have to rethink this part of the code a bit.

https://github.com/osmlab/name-suggestion-index/blob/9424e0313feaa850e0b32fc1b522995357069e19/lib/matcher.js#L209

stefanb:
We'd have identical problem with a different "Petrol" brand based in Slovenia:
- https://www.wikidata.org/wiki/Q174824

with gas stations in several other countries (and not in Bulgaria):
- https://www.petrol.eu/gas-stations/map?lng=21.134580475616776&lat=43.3540214718437&zoom=5#
- https://location-conflation.com/?locationSet=%7B%20include%3A%20%5B%27si%27%2C%20%27hr%27%2C%20%27rs%27%2C%20%27ba%27%2C%20%27me%27%5D%20%7D&referrer=nsi

stefanb:
The generic name "Petrol" would conflict with exclusion regex:

https://github.com/osmlab/name-suggestion-index/blob/6ee28795fd7aa8f48ff8693d92888bf86db8a2af/data/brands/amenity/fuel.json#L7

stefanb:
Are we sure this is still the case?

In Slovenia it seems to work correctly after #7725:
![telegram-cloud-photo-size-4-5967484688878648627-y](https://user-images.githubusercontent.com/319826/215864768-d0ff4080-11be-4ce0-a912-b3abc2b1fa40.jpg)

And it creates the POI as expected: 
[...]

PR #8730, opened 2023-10-03
Update fuel.json added Petrol AD

jhutton1:


Dimitar5555:
Petrol is already in NSI. See https://nsi.guide/index.html?t=brands&k=amenity&v=fuel&tt=петрол#petrol-b450da

Cj-Malone:
@Dimitar5555 I suspect "petrol" as a match name was what "fixed" it in Slovenia, so this may fix #6552 too.

jhutton1:
Sorry, I missed that.  Closing PR as not needed.

Dimitar5555 (approved):
> @Dimitar5555 I suspect "petrol" as a match name was what "fixed" it in Slovenia, so this may fix #6552 too.

I'm not sure if it will work but it seems like it won't break anything. I will let you know what the result is.

PR #8734, opened 2023-10-04
Updated it.json Added Accenture

Guerme:
Adding Accenture - https://www.accenture.com/us-en

### DIFF
--- a/data/brands/office/it.json
+++ b/data/brands/office/it.json
@@ -5,6 +5,16 @@
   },
   "items": [
     {
+      "displayName": "Accenture",
+      "locationSet": {"include": ["001"]},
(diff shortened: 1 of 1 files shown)


kjonosm:
@Guerme Could you check whether Accenture should be added as `office=it` or rather `office=consulting`?

Guerme:
@kjonosm - I initally looked at their services page which listed mostly technology solutions.  However, reading it further, it does seem they are helping clients determine what technologies to use/build (i.e. consulting) and not actually providing the technology to the client (i.e. it).  I'll switch it to consulting.  Thanks.

kjonosm:
Thanks!

PR #8736, opened 2023-10-04
Update engineer.json Add Alstom

Guerme:
Adding Alstom - https://www.alstom.com/

### DIFF
--- a/data/brands/office/engineer.json
+++ b/data/brands/office/engineer.json
@@ -4,6 +4,16 @@
     "exclude": {"generic": ["^engineer"]}
   },
   "items": [
+  {
+      "displayName": "Alstom",
+      "locationSet": {"include": ["001"]},
+      "tags": {
+        "brand": "Alstom",
+        "brand:wikidata": "Q309084",
+        "name": "Alstom",
+        "office": "engineer"
+        }
+    },
     {
       "displayName": "WSP",
       "locationSet": {"include": ["us"]},


kjonosm:
@Guerme Sorry for nitpicking but according to OSM wiki `office=engineer` is used for an office that is offering (individual) engineering as a service. However, Alstom is a manufacturer in the infrastructure (mainly rail) industry.

Guerme:
@kjonosm - No worries.  I see that there is an OSM key for manufacturer.  Should I add a manufacturer.json file?  If so, should it go in the man-made folder?

kjonosm:
From my understanding there is no established tagging yet. Depending on the place it could be [`office=company`](https://wiki.openstreetmap.org/wiki/Tag:office=company), [`man_made=works`](https://wiki.openstreetmap.org/wiki/Tag:man_made=works) or [`industrial=*`](https://wiki.openstreetmap.org/wiki/Key:industrial). Any ideas @1ec5?

1ec5:
If Alstom’s facilities are that diverse, there would be multiple entries for the same operator, one per tag. It’s also possible that, say, a given company’s offices may be more relevant for inclusion in NSI than its factories or vice versa.

bhousel:
I'm just going to accept this as-is.  If there's a better category for it, we can move it, but calling them an engineering office isn't completely wrong.

PR #8740, opened 2023-10-04
Update trade.json Add Tradelink

Guerme:
Adding Tradelink - https://tradelink.com.au/

### DIFF
--- a/data/brands/shop/doityourself.json
+++ b/data/brands/shop/doityourself.json
@@ -14,6 +14,16 @@
   },
   "items": [
     {
+      "displayName": "Tradelink",
+      "locationSet": {"include": ["au"]},
+      "tags": {
+        "brand": "Tradelink",
+        "brand:wikidata": "Q120648432",
+        "name": "Tradelink",
+        "shop": "doityourself"
+      }
+    },
+    {
       "displayName": "Ace Hardware",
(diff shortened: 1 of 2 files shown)


kjonosm:
@Guerme Is this really a DIY store and not a `shop=kitchen`, `shop=bathroom_furnishing` or `shop=furniture`? You can check OSM wiki for reference.

Guerme:
@kjonosm - It sells fixtures for bathrooms and kitchens, so I don't think it can fit in one of those categories alone.  It also doesn't sell furniture, only appliances and fixtures, so not sure it fits in furniture.  So due to this, it seemed to me like it was more related to Lowe's/Home Depot, as it sells a mix of appliances/fixtures for both bathroom and kitchen.  Or can I add to both kitchen and bathroom furnishing?

kjonosm:
Current tagging is quite diverse. In my opinion it it best described as [`shop=trade`](https://wiki.openstreetmap.org/wiki/Tag:shop=trade) + `trade=plumbing`. On their website they call themselves 

> Australia’s oldest plumbing merchants

Guerme:
@kjonosm - Thanks for the suggestion.  Moved to trade.json and added trade=plumbing.

PR #8746, opened 2023-10-05
Add fastned brand

ckpiper:


serhii-muchychka:
It is already in the operators tree:
https://github.com/osmlab/name-suggestion-index/blob/main/data/operators/amenity/charging_station.json#L2169

ckpiper:
@serhii-muchychka Good catch.

PR #8752, opened 2023-10-05
Update books.json Add QBD Books

Guerme:
Adding QBD Books - https://www.qbd.com.au/

### DIFF
--- a/data/brands/shop/books.json
+++ b/data/brands/shop/books.json
@@ -12,6 +12,16 @@
   },
   "items": [
     {
+      "displayName": "QBD Books",
+      "locationSet": {"include": ["au"]},
+      "tags": {
+        "brand": "QBD Books",
+        "brand:wikidata": "Q118994358",
+        "name": "QBD Books",
+        "shop": "books"
+      }
+    },
+    {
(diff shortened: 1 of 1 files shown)


data/brands/shop/books.json
```diff
@@ -11,6 +11,16 @@
     }
   },
   "items": [
+    {
+      "displayName": "QBD Books",
+      "locationSet": {"include": ["au"]},
+      "tags": {
+        "brand": "QBD Books",
+        "brand:wikidata": "Q118994358",
```
serhii-muchychka, line 19: Appears to be a duplicate wikidata [Q7270909](https://www.wikidata.org/wiki/Q7270909)
```suggestion
        "brand:wikidata": "Q7270909",
```

Guerme: I didn't create either wikidata pages, but given that Q118994358 has the true brand name of QBD Books and more information than Q7270909, I think we should stick with Q118994358.

serhii-muchychka: Oh, I didn't notice that the newer QID was created by another user. Ideal variant - merge these wikidata QIDs. Maybe someone else will do it in the future

PR #8757, opened 2023-10-05
Add Zumba

caguilara:
Hi,
I'm adding Zumba - https://www.zumba.com/
Thanks,


arch0345:
Does Zumba have standalone locations? When searching for locations on their website, the only results I found are classes being held at affiliated gyms

PR #8758, opened 2023-10-05
Add Ecovis

caguilara:
Hi,
I'm adding Ecovis International - https://www.ecovis.com
Thanks,

### DIFF
--- a/data/brands/office/consulting.json
+++ b/data/brands/office/consulting.json
@@ -42,6 +42,16 @@
       }
     },
     {
+      "displayName": "Ecovis International",
+      "locationSet": {"include": ["001"]},
+      "tags": {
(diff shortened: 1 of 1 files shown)


data/brands/office/consulting.json
```diff
@@ -41,6 +41,16 @@
         "office": "consulting"
       }
     },
+    {
+      "displayName": "Ecovis International",
+      "locationSet": {"include": ["001"]},
```
arch0345, line 46: It looks like they only have locations in Europe.
```suggestion
      "locationSet": {"include": ["150", "tr"]},
```

arch0345 (changes requested):

PR #8774, opened 2023-10-09
Add Swedish petrol stations to fuel.json

fridokus:
Add two words commonly used for petrol stations in Sweden


arch0345:
Thanks for your first contribution to NSI!

PR #8775, opened 2023-10-09
Bosch Car Service has its own, dedicated Wikidata entry

mstsbe:
Instead of using the Wikidata entry of the huge parent company Robert Bosch, it seems more fitting to use the specific Wikidata entry for this business entity:
https://www.wikidata.org/wiki/Q894368

We also get the correct logo then, instead of the generic Bosch logo.


arch0345:
Thanks for catching this!

PR #8776, opened 2023-10-10
Add some Nordic brand generic names

fridokus:
All of these are used in Swedish, some in Danish and Norwegian too.


kjonosm:
Thanks @fridokus!

PR #8799, opened 2023-10-12
Add cooperative group

ckpiper:


Cj-Malone:
There is drama in this one, I'll think about it and update the PR.

Cj-Malone (approved):
The Co-operative Group missing was absolutely an issue, thank you. However it's an operator for the coop brand.

The "coop brand" is also a bit of a myth. There are several brands, each operated by several companies.

This is where I would like to leave things, with a generic "Co-op Food" (coop blue) and a specific "Co-op Food" by "The Co-operative Group". There may be another company operating coop blues (in which case NSI should be updated) however are probably a few small ones, where its hardly worth the time and probably wont meet Wikidatas notability requirements.

This is however an issue for NSI as it's classed as a duplicate. I'll remove the generic one as my hypotheticals above don't really matter until they are reality.

There is an issue with "The Co-operative Food" (coop green) as most people don't know who is operating it, they just know it's coop green. I think not having a generic "The Co-operative Food" in NSI is an issue. (This isn't introduced via this PR)

bhousel:
Just merged whatever this contains.. If there's still issues with it we can do a followup PR.

PR #8801, opened 2023-10-13
Update it.json added thales

jhutton1:


AntiCompositeNumber:
I'm not sure `office=it` fits here. Thales is mostly a manufacturing company, especially in defense and transportation. I did find one instance of it being tagged [`office=it`](https://www.openstreetmap.org/relation/11742461), but it seems like `office=company` is the most common tagging method so far.

data/brands/office/company.json
```diff
@@ -4,6 +4,16 @@
     "exclude": {"generic": ["^company$"]}
   },
   "items": [
+        {
+      "displayName": "Thales",
+      "locationSet": {"include": ["fr"]},
```
AntiCompositeNumber, line 9:
```suggestion
      "locationSet": {"include": ["001"]},
```
Thales is headquartered in France, but has offices in a number of countries: https://en.wikipedia.org/wiki/Thales_Group#Thales_international. Some of those have Wikidata items for the local organization (https://www.wikidata.org/wiki/Q1161666#P355 ), but I think it's fine to tag everything with the top-level group.

AntiCompositeNumber (approved):

PR #8802, opened 2023-10-13
Update beauty.json added Thérapie Clinic

jhutton1:


kjonosm:
@jhutton1 Could you add Ireland location set?

PR #8807, opened 2023-10-14
Change geo-location for Gold Coast City Council parks

Scottie0001:
Change geo-location for Gold Coast City Council from Victoria to Queensland


kjonosm:
Thanks @Scottie0001

PR #8816, opened 2023-10-17
Add dissolved Polish brands

starsep:


serhii-muchychka:
Hi, Filip
this should be done on the Wikidata side. Here it will be pulled up automatically during the release
For example, you can look at [Q7501155](https://www.wikidata.org/wiki/Q7501155) ([Property:P576](https://www.wikidata.org/wiki/Property:P576))

PR #8818, opened 2023-10-17
Remove boîte à livres

Bqleine:
The wikidata for the removed entry points to a lone osm node that is not a brand and not a public bookcase anymore.

I tried deleting it all-together but it comes back when I run npm run build.


Bqleine:
Thanks, using your query I was able to find that because boîte can be correctly written without an accent too the existing regex failed to match on a lot of nodes. I've updated it to support the multiple accents and it seems to be completely removed now.

Also, boîte à livre should be treated as a generic phrase because it's not a brand and is already represented by amenity=public_library

PR #8831, opened 2023-10-20
Add brand:ja to Ding Tea

Cj-Malone:
Currently [Ding Tea JP](https://nsi.guide/index.html?t=brands&k=amenity&v=cafe&tt=ding%20tea#dingtea-0cf217) has only ZH, presumably it should be JA?

cc @LaoshuBaby it was added in https://github.com/osmlab/name-suggestion-index/commit/590c34b0b7260fe01f18831243084dae4515e0bf

cc @arch0345 Ding Tea (Not JP) was added in https://github.com/osmlab/name-suggestion-index/commit/aed811fc3bfb66b320946e69de759208944c3eb2

(Values are presumably wrong, do not merge)

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -1533,10 +1533,12 @@
         "brand:en": "Ding Tea",
         "brand:ja": "薡茶",
         "brand:wikidata": "Q112123475",
+        "brand:zh": "薡茶",
         "cuisine": "bubble_tea",
         "name": "Ding Tea 薡茶",
         "name:en": "Ding Tea",
         "name:ja": "薡茶",
+        "name:zh": "薡茶",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
In fact the character "薡茶" can both be recognized as Chinese and Japanese, because those 2 East Asian language share many characters. At that time I split it and want to add a `name:ja` tag, but I've forgotten. I'd suggest both keep `name:zh` and `name:ja` in this JP region entry. `zh` is because this brand original from Taiwan (use Chinese), and `ja` is because this entry will be applied in Japan.

Thank you for your notice this!

Cj-Malone:
Thanks, I'll update the PR later when I get to a computer, or tomorrow.

If you have time before, feel free to do it.

PR #8838, opened 2023-10-23
Add Tacombi

hdevine825:


LaoshuBaby:
There is nothing wrong with this PR content itself (However, I haven’t checked whether its existing number of stores meets the requirements.)

But as far as Git usage habits are concerned, **I strongly recommend that you do not work directly on the main branch, but open a separate branch for each PR**, and always keep your main branch up to date with the upstream main branch.

In this way, every time you submit a PR, you can delete the temporary branch to ensure that your submission record does not have such a long commit history of past submissions!

it contains at least #8584 and #8772 in https://github.com/osmlab/name-suggestion-index/pulls?q=is%3Apr+author%3Ahdevine825+is%3Aclosed

PR #8839, opened 2023-10-23
Add British Heart Foundation Home Store

### LINKED ISSUE #6954: British Heart Foundation Home Stores
BHF run Home Stores, charity shops which just sell furniture and some white goods:
https://www.bhf.org.uk/shop/home-stores

Currently we don't have a separate tag for them, but it seems like it might be worthwhile assuming separate wikidata I guess?

I guess this is Tesco Metro/Superstore again effectively...

Not sure what to tag them as still charity shop but can we differentiate goods somehow?

peternewman:
Closes #6954

I've kept consistent and not added operator as per #4136 although it was common on the already mapped ones I found.

### DIFF
--- a/data/brands/shop/furniture.json
+++ b/data/brands/shop/furniture.json
@@ -331,6 +331,21 @@
       }
     },
     {
+      "displayName": "British Heart Foundation Home Store",
+      "locationSet": {"include": ["gb"]},
+      "matchNames": [
+        "british heart foundation furniture & electrical"
+      ],
+      "tags": {
+        "brand": "British Heart Foundation",
+        "brand:wikidata": "Q4970039",
+        "charity": "yes",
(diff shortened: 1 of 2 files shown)


data/brands/shop/furniture.json
```diff
@@ -330,6 +330,21 @@
...
+    {
+      "displayName": "British Heart Foundation Home Store",
+      "locationSet": {"include": ["gb"]},
+      "matchNames": [
+        "british heart foundation furniture & electrical"
+      ],
+      "tags": {
+        "brand": "British Heart Foundation",
+        "brand:wikidata": "Q4970039",
+        "charity": "yes",
```
peternewman, line 342: Charity is only proposed:
https://wiki.openstreetmap.org/wiki/Key:charity

PR #8843, opened 2023-10-23
Update trade.json added Mr. Electric

jhutton1:


kjonosm:
@jhutton1 Could you check whether [`craft=electrician`](https://wiki.openstreetmap.org/wiki/Tag:craft=electrician) is a better fitting tag?

jhutton1:
perhaps.  Does a .json exist for electrician under brands?

AntiCompositeNumber:
I agree that `craft=electrician` is the better tag. A file does not exist yet, but should be created.

PR #8845, opened 2023-10-23
Add cintas

DivyaB1:
Hi,

I'm adding Cintas - https://www.cintas.com/

Thanks,


1ec5:
#8916 removes Cintas for now. #8917 tracks restoring it and two other large chains as soon as OpenStreetMap settles on a tag that we can use for them.

Issue #8829, opened 2023-10-19
All caps La Compagnie du lit


Kalepom:
Hello, 
The name tag suggestion for La Compagnie du lit is written in all caps
![La Compagnie du lit](https://github.com/osmlab/name-suggestion-index/assets/12937457/ddef1def-d2e3-4e59-81de-298bb9002ee2)

PR #8846, opened 2023-10-24
fix caps

Outlet2048:
closes:  #8829


serhii-muchychka:
This brand was added in
https://github.com/osmlab/name-suggestion-index/pull/8420

Caps are not used in the brand's social networks:
https://www.facebook.com/lacompagniedulit

PR #8853, opened 2023-10-24
Rename "Bantierra" to "Caja Rural de Aragón"

miguel-latre:
Bantierra is the former brand of Caja Rural de Aragón. Bantierra brand was discontinued in 2019 ( https://es.wikipedia.org/wiki/Caja_Rural_de_Arag%C3%B3n )

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -2932,14 +2932,14 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


Robot8A:
Commit added to add the old name into a match name:
https://github.com/osmlab/name-suggestion-index/commit/43292ed696513691605269453052d87f547f904d

PR #8855, opened 2023-10-24
Added ALH Hotels Brand to Hotel.json

ckpiper:


arch0345:
It looks like ALH Hotels is an operator of hotels, each of which has a unique name https://www.alhgroup.com.au/venues

ckpiper:
Good catch. Not just hotels either. They have a lot of restaurants as well.

UKChris-osm:
In this case you can add `"preserveTags": ["^name"]` under the `locationSet` tag and this will allow mappers to add the unique name of each hotel without the NSI trying to overwrite it later.

bhousel:
> In this case you can add `"preserveTags": ["^name"]`

oh, that's not necessary here - it's already set that way for the whole `hotel.json` category at the top of the file..

UKChris-osm:
> oh, that's not necessary here - it's already set that way for the whole `hotel.json` category at the top of the file..

Ops, I didn't notice that, thanks 👍

ckpiper:
> this will allow mappers to add the unique name of each hotel without the NSI trying to overwrite it later.

@UKChris-osm I just got back from vacation, so that is why I am responding 2 weeks later. But I am a little confused on what you mean? Do you mean mappers will add the hotel name and this nsi_id and NSI will not overwrite the name with ALH Hotels?

PR #8859, opened 2023-10-25
Update university.json

lucumon:
Added "Universidade da Coruña" as amenity=university operator. Maybe it should be also added as brand?

### DIFF
--- a/data/operators/amenity/university.json
+++ b/data/operators/amenity/university.json
@@ -27,6 +27,18 @@
       }
     },
     {
+      "displayName": "Universidade da Coruña",
+      "locationSet": {"include": ["es"]},
+      "tags": {
+        "amenity": "university",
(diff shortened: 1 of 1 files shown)


lucumon:
Sorry I don't understand why you are not adding this info. Some universities here in Spain, despite being public and not private, operate in various campuses, so It's usefull to have this operator info on nsi.guide so tools like iD suggest wikidata tags, etc. when people are editing individual buildings and campuses. As an example, "Universidade da Coruña" operates two different campuses in Ferrol with a total of 16 buildings, and three campuses in A Coruña. Also, they may have some investigation offices on coworking spaces and so on.

bhousel:
This entry in the pull request would only match `amenity=university` - it wouldn't match buildings and offices.   The matcher doesn't actually match anything with the operator, only the specific key=value pair.

PR #8868, opened 2023-10-26
Update parking.json Add National Car Parks

Guerme:
Adding National Car Parks - https://www.ncp.co.uk/


Guerme:
Just discovered this is already in operators\amenity\parking.json, so doesn't need to be added to brands.

PR #8871, opened 2023-10-27
Rebrand Chuck E. Cheese's

Cj-Malone:
I was surprised to see "Chuck E. Cheese's", but then I actually looked at some images and the fascia does have the `'s`. So I agree with the name, but I think the brand is `Chuck E. Cheese`.

Anyway, I might be wrong so I'll leave it to an American to decide.

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -1157,12 +1157,12 @@
     {
       "displayName": "Chuck E. Cheese's",
       "id": "chuckecheeses-85c92c",
-      "locationSet": {"include": ["ca", "us"]},
+      "locationSet": {"include": ["ca", "pr", "us"]},
       "matchNames": ["chuck e cheese"],
       "matchTags": ["leisure/amusement_arcade"],
       "tags": {
         "amenity": "restaurant",
-        "brand": "Chuck E. Cheese's",
+        "brand": "Chuck E. Cheese",
         "brand:wikidata": "Q2438391",
         "cuisine": "pizza",
         "leisure": "amusement_arcade",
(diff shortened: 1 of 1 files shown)


AntiCompositeNumber:
Yes, the `'s` was dropped starting in 2019 when they rebranded to be more "modern". I don't know how far the roll out of the rebranding has progressed especially since they went bankrupt again in 2020. I think it does make sense to update `brand=` and `name=` like we have with Dunkin' Donuts and other slowly-rebranding franchises.

UKChris-osm:
I think this change would be better suited for all the tags, such as `display name` and `name`, rather than just the `brand` tag, as it seems the brand as a whole no longer includes the "'_s_".

Cj-Malone:
Thanks everyone, agreed.

How do we feel about 2 categories on these?

data/brands/amenity/restaurant.json
```diff
@@ -1155,18 +1155,18 @@
       }
     },
     {
-      "displayName": "Chuck E. Cheese's",
+      "displayName": "Chuck E. Cheese",
       "id": "chuckecheeses-85c92c",
-      "locationSet": {"include": ["ca", "us"]},
+      "locationSet": {"include": ["ca", "pr", "us"]},
```
arch0345, line 1160: According to their Wikipedia article, they have some locations outside of the US/Canada (Puerto Rico is included in the us country-coder polygon so it isn't necessary to include it): https://en.wikipedia.org/wiki/Chuck_E._Cheese#International_markets
```suggestion
      "locationSet": {"include": ["ae", "bh", "ca", "ch", "co", "gt", "hn", "mx", "pa", "pe", "qa", "sa", "sr", "sv", "tt", "us"]},
```

PR #8876, opened 2023-10-28
Updated Safeway in supermarkets

ckpiper:
Safeway Canada became independent in 2013. Is still uses same logo and branding, but has a different Wikidata page. Thus I think there should be 2 of them on for US and one for CA. 

https://github.com/alltheplaces/alltheplaces/pull/6208

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -31,6 +31,16 @@
   },
   "items": [
     {
+      "displayName": "Safeway",
+      "locationSet": {"include": ["ca"]},
+      "tags": {
+        "brand": "Safeway",
+        "brand:wikidata": "Q17111901",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
I see you have changed the existing entry from global (`001`) to just the USA (`us`) - are you confident the brand doesn't trade in any other Country (apart from Canada)?

Cj-Malone:
I've just done a quick spot check of the ones in OSM. There are a few, but I think they are tagging mistakes.

... Apart from Jordan. It does look like there are a couple "Safeway" threr.

ckpiper:
@UKChris-osm On the wikipedia site it says area served is ony 17 US States. And I could not find anywhere besides the US where they operated.

TheAdventurer64:
#9033

PR #8877, opened 2023-10-28
Added Clean Harbors as a waste water operator

ckpiper:
Not sure if this is exactly a operator (have not added an operator before), but I think it belongs here. It does many services, which you can see in the dropdown [cleanharbors](https://www.cleanharbors.com/). Also in the [wikipedia](https://en.wikipedia.org/wiki/Clean_Harbors) page it has waste services. Then lastly, [wikidata](https://www.wikidata.org/wiki/Q5130494) says an industrial and water waste management business.

### DIFF
--- a/data/operators/man_made/wastewater_plant.json
+++ b/data/operators/man_made/wastewater_plant.json
@@ -5,6 +5,15 @@
   },
   "items": [
     {
+      "displayName": "Clean Harbors",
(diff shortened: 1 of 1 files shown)


arch0345:
Based on their website, it does appear that they operate 7 wastewater treatment facilities in the US/Canada

PR #8881, opened 2023-10-29
fix Appalachian Power Company

Hiausirg:
The correct operator name is "Appalachian Power Company", not "Appalachian Power". ([Example](https://elibrary.ferc.gov/eLibrary/filelist?accession_number=20230412-8012&optimized=false))

### DIFF
--- a/data/operators/power/line.json
+++ b/data/operators/power/line.json
@@ -283,7 +283,7 @@
       }
     },
     {
-      "displayName": "APCO",
+      "displayName": "Appalachian Power",
       "id": "appalachianpower-33137a",
       "locationSet": {
         "include": [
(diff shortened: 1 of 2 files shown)


kjonosm:
Their website https://www.appalachianpower.com give a different impression...🤔

Hiausirg:
This dosen't mean anything. The correct legal name is "Appalachian Power Company".

kjonosm:
In that case, could you add "appalachian power" to `matchNames` so iD can update wrongly tagged items accordingly?

And is "APCO" a valid abbreviation as currently suggested? If not, it makes sense to also add it to `matchNames`.

Hiausirg:
> In that case, could you add "appalachian power" to `matchNames` so iD can update wrongly tagged items accordingly?
> 
> And is "APCO" a valid abbreviation as currently suggested? If not, it makes sense to also add it to `matchNames`.

added both to matchNames

PR #8884, opened 2023-10-29
added genghis grill to restaurant

ckpiper:


data/brands/amenity/restaurant.json
```diff
@@ -43,6 +43,17 @@
     }
   },
   "items": [
+    {
+      "displayName": "Genghis Grill",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "amenity": "restaurant",
+        "cuisine": "american",
```
AntiCompositeNumber, line 51: https://en.wikipedia.org/wiki/Genghis_Grill describes it as "Mongolian" (hence the name), but it looks like they've changed their menu since 2021. It seems like they only describe themselves as a "fast-casual bowl concept restaurant". I still wouldn't really call https://www.genghisgrill.com/menu `cuisine=american`, but I'm also having trouble thinking of a better alternative.

matkoniecz: no cuisine tag is better than a bad or dubious one

LaoshuBaby: FYI: #6791

AntiCompositeNumber: After thinking about it some more,
```suggestion
        "cuisine": "mongolian_grill;american",
```
seems to be the best option to me. The cooking method is largely still `mongolian_grill` (even though newer franchises use standard griddles), but the menu is about 50/50 Asian and American. I think keeping the `american` is useful to distinguish between the more traditional Mongolian barbecue restaurants.

1ec5: I guess the question is whether a diner would be surprised to visit one of these restaurant locations after a search for Mongolian grills. If we had gone with `mongolian_bbq` as I had originally suggested, Genghis would definitely be disqualified. As it is, I would probably still describe it as vaguely Asian fusion cuisine rather than a Mongolian grill, since people may want that special experience watching the meat go around in circles.

I’d suggest `cuisine=fusion;grill`, maybe with `theme=mongolia`, though in this case the theming doesn’t extend far beyond the name. As a point of reference, Outback Steakhouse (#2671) has more successfully appropriated Australian culture but still isn’t `australian`.

AntiCompositeNumber: `theme=mongolia` would be a stretch, at least for the post-2021 restaurants. They're more of a `theme=generic_fast_casual` or maybe `theme=bowl` (neither of those being serious tags, for the record). I'm not sure how useful it is to tag `cuisine=fusion` without some way to say what is being fused.

1ec5: If we want to get technical, this chain is definitely `cuisine=american`, because, well, bowling is quintessentially American. 😉 But my impression is that `cuisine=american` is somewhat associated with stereotypical American cuisine like mashed potatoes and gravy, macaroni and cheese, and scrambled eggs. I’m pretty sure that’s what a consumer would be expecting when they tap the button for “American” when searching for a restaurant in a consumer map application.

For what it’s worth, there are restaurants that describe themselves as fusion, period. The consumer application may not offer a button to conveniently search for fusion restaurants per se, but it might still label a restaurant as one, which you could think of as a warning to those with more traditional tastes.

data/brands/amenity/restaurant.json
```diff
@@ -43,6 +43,17 @@
     }
   },
   "items": [
+    {
+      "displayName": "Genghis Grill",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "amenity": "restaurant",
```
AntiCompositeNumber, line 50: Not to throw more confusion into the works, but per the definition in https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#restaurant-or-fastfood `amenity=fast_food` would probably fit this better, especially for the post-2021 design.

bhousel:
Going to merge, then move it to `fast_food` and leave the cuisine as `american` .

This shouldn't have been held up for 7 months because of the cuisine tag. 

["perfect is the enemy of good"](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good) and ["disagree and commit"](https://en.wikipedia.org/wiki/Disagree_and_commit), etc.

1ec5:
Thanks, totally forgot this one was still outstanding. We also merged #9455 recently despite similar questions. If anyone disagrees strongly with the tagging, feel free to open an issue to ex

PR #8887, opened 2023-10-29
create logistics to add averitt express

ckpiper:
OSM says office for forwarder/hauler. Not sure exactly sure if this is the correct location for this type of brand. Feel free to look https://www.wikidata.org/wiki/Q4828320.


Cj-Malone (approved):
Current OSM usage seems to be either on the landuse or on a generic office=yes/office=company.

I don't know a lot about this, but it seems fine to me.

Issue #6014, opened 2022-01-06
Future import assertions syntax for importing JSON files..


bhousel:
Currently this project does a lot of importing JSON files. 
This is complicated to do in a cross environment way - some info in https://github.com/osmlab/name-suggestion-index/commit/b21ac2eb66e6b7d371356b44500c7197e3223728  - tl;dr:

```js
import matchGroupsJSON from '../config/matchGroups.json';
```

- will "just work" in node if the `--experimental-json-modules` flag is enabled.
- will get bundled properly by rollup, webpack, esbuild if this code ever ends up in a browser bundle.

Unfortunately starting with Node 17, an [import assertion](https://nodejs.org/api/esm.html#import-assertions) is _required_, so the syntax for importing JSON has changed to:

```js
import matchGroupsJSON from '../config/matchGroups.json'  assert {type: 'json'};
```

@maxerickson tried adding these in f4a11709e, but needed to revert in 7e08ce1b6  when the LTS versions of node in our CI build failed on this newer syntax.

See also: https://github.com/tc39/proposal-import-assertions
Unflagging the JSON module importer:  https://github.com/nodejs/node/pull/37375
Rejected Pull Request for a flag preserving the legacy syntax: https://github.com/nodejs/node/pull/40210

It looks like esbuild, rollup, typescript etc are supporting the new syntax, or will soon:
https://github.com/evanw/esbuild/issues/1871#issuecomment-998383853

So this issue is mainly to keep an eye on things.. 
I don't think we can support Node 17 just yet, but we will eventually need to add these import assertions when we are targeting future node versions and can drop support for current node versions.

### UPDATE 2/16/2022: 
It surprised me to see that the import assertion requirement got [backported to node 16.14](https://github.com/nodejs/node/pull/41776).  This means we really can't wait to implement this, as our contributors are [already starting](https://github.com/osmlab/osm-community-index/pull/503) to run into parse errors  😠 

Also - eslint is not supporting the syntax yet until it reaches stage 4, so I guess we will be disabling eslint for a while.
https://github.com/eslint/eslint/discussions/15305

So today I will:
- [x] add this syntax anywhere we import a JSON file
- [x] drop support for older versions of node in package.json and the CI action
- [x] remove eslint from the test command for now 😢

PR #8891, opened 2023-10-29
fixed eslint and linted files

wvanderp:
Hello,

In this PR, I fixed eslint because it breaks on the new `assert {type: 'json'}` syntax. I fixed it by introducing Babel to interpret the syntax. I then solved the lining issues that arose.

There is a lot to like in this PR. But there are a few controversial decisions I made.

**Why do we need Babel?**  
https://github.com/eslint/eslint/discussions/15305 states that eslint will only be implemented when the TC accepts it, so we need Babel as a temporary stopgap.

**Why did you remove the commented-out code?**  
there were a lot of unused variables and imports that were only there for the unused code. So that either needed to be commented out or needed to be removed. I disliked "dead" code because it does not get updated, keeping it in a working state code-wise and idea-wise.

**You switched the fork of json-stringify-pretty-compact. Why?**  
I saw that typescript was complaining about the import. That is why I started looking at what was happening, and I saw that this was a fork of another repo. I don't know the original reason for changing, but the original repo started getting updates again. So I switched back, and that fixed the problem and introduced no new regressions.

**You added lining to the GitHub actions. This means that people who want to update the index will need to fix the lining errors in the glue code.**  
Because we now get a notification when there is a linting error, we can let the person who introduced it fix it.

**Why is cSpell now included in .vscode?**
cSpell is the most used spellchecker for vscode. So, capturing some OSM-specific words will help prevent error fatigue.

I am open to feedback and applying any changes you feel necessary.
-wvdp

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -6,6 +6,7 @@
   "extends": [
     "eslint:recommended"
   ],
+  "parser": "@babel/eslint-parser",
   "parserOptions": {
     "ecmaVersion": 6,
     "sourceType": "module"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,9 @@ jobs:
     - run: npm install
       env:
         FORCE_COLOR: 2
+    - run: npm run lint
+      env:
+        FORCE_COLOR: 2
     - run: npm run test
       env:
         FORCE_COLOR: 2
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,41 +1,68 @@
 {
-  "editor.formatOnSave": false,
-  "json.schemas": [
-    {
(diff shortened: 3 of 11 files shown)


bhousel:
Thanks for this, #6014 has some of the history from when the new `import assert` was introduced by node, and why eslint is disabled.  

Sorry to say I am uncomfortable accepting this PR, and would prefer to decline it.  I'm pretty ok leaving eslint disabled for now, and not comfortable introducing a dependency on babel just for this.  

Some of the other changes might make sense (like switching the json prettier) - we should open issues for these things if there is value there.

I don't understand the VSCode changes, and I don't use VSCode personally.  I don't understand why anyone would want spell checking in a code editor.

PR #8894, opened 2023-11-01
Add 'NTB' as an alias for 'National Tire and Battery'

CorruptComputer:
Most of the branding for this company just says "NTB", I had no idea their full name was "National Tire and Battery" until I found their Wikipedia page. Even their own website just says NTB on it, not the full name: https://www.ntb.com/home

Searching for "ntb" in iD does not find this brand:
image

I'm not super familiar with NSI, so I'm not exactly sure if this is right. I am just hoping to add "ntb" as an alias so it shows up in searches there.

### DIFF
--- a/data/brands/shop/car_repair.json
+++ b/data/brands/shop/car_repair.json
@@ -882,9 +882,10 @@
         "ntb"
       ],
       "tags": {
-        "brand": "National Tire and Battery",
+        "brand": "NTB",
         "brand:wikidata": "Q6978944",
-        "name": "National Tire and Battery",
+        "name": "NTB",
+        "alt_name": "National Tire and Battery",
         "shop": "car_repair"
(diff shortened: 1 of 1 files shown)


arch0345:
> Most of the branding for this company just says "NTB", I had no idea their full name was "National Tire and Battery" until I found their Wikipedia page. Even their own website just says NTB on it, not the full name: https://www.ntb.com/home

If this is the case, I'd say the `brand` and `name` tags should be changed to 'NTB' and 'National Tire & Battery' should be added as an `alt_name` tag

CorruptComputer:
Sure that sounds good to me, I made the requested changes. Please let me know if I should change anything else.

bhousel:
Thanks @CorruptComputer & welcome!

PR #8903, opened 2023-11-03
Add Budget Porta Potty

DivyaB1:
Hi,

I am adding Budget Porta Potty - https://budgetportapotty.com/

Thanks,

### DIFF
--- a/data/brands/shop/rental.json
+++ b/data/brands/shop/rental.json
@@ -4,6 +4,16 @@
     "exclude": {"generic": ["^rental$"]}
   },
   "items": [
+     {
+      "displayName": "Budget Porta Potty",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "brand": "Budget Porta Potty",
+        "brand:wikidata": "Q121318335",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Do they have any Facebook account which can be added to the Wikidata, so we can grab a logo?

AntiCompositeNumber:
Looking at https://budgetportapotty.com/, the locations list is just a bunch of SEO garbage for every town in every state. I also can't find any examples of this brand on OSM. How confident are we that this is a brand with a physical presence, as opposed to just a website and a phone number?

DivyaB1:
> Looking at https://budgetportapotty.com/, the locations list is just a bunch of SEO garbage for every town in every state. I also can't find any examples of this brand on OSM. How confident are we that this is a brand with a physical presence, as opposed to just a website and a phone number?

Looks like all the locations on their website are service locations and not physical presence. I am sure they have physical presence somewhere since they are providing services in almost all states but that information is not present on their website. So I will close the PR, Thanks.

PR #8910, opened 2023-11-05
Add Henry Laden

j13m126:
see https://www.roteskreuz.at/niederoesterreich/ich-will-helfen/second-hand-shops


arch0345:
Please note that since this entry is in the brands tree, a `brand:wikidata` tag is required for it to get generated as a preset

j13m126:
Thanks for the info, I added brand:wikidata.

PR #8921, opened 2023-11-08
Update college.json Added COMCAVE

mollimch:
Hello,

Adding COMCAVE

https://www.comcave.de/


AntiCompositeNumber (commented):
Title case is [generally preferred](https://wiki.openstreetmap.org/wiki/Names#Good_practice). https://www.comcave.de/impressum capitalizes it `ComCave`, https://www.facebook.com/COMCAVE.COLLEGE/ and various press releases use `Comcave.College`.

PR #8926, opened 2023-11-10
Use own engine for wikidata search

serhii-muchychka:
The main improvement here that the Google engine does not know how to search by aliases ("Also known as" fields) in Wikidata, unlike the built-in


bhousel:
Thanks @serhii-muchychka,  this seems ok.. 

I think I originally had it using the google site search because early on it was likely that the wikidata search would not turn up anything, and this was more to find out whether a page even existed at all.

Cj-Malone:
I still have issues with search sometimes, and would recommend searching by domain name before making a page.

But the search works out most of the time 👍

PR #8950, opened 2023-11-17
Add LHW

DivyaB1:
Hi,

I'm adding LHW - https://www.lhw.com/

Thanks,

### DIFF
--- a/data/brands/tourism/hotel.json
+++ b/data/brands/tourism/hotel.json
@@ -18,6 +18,17 @@
     }
   },
   "items": [
+      {
+      "displayName": "The Leading Hotels of the World",
+      "locationSet": {"include": ["001"]},
+      "matchNames": ["lhw"],
+      "tags": {
+        "brand": "The Leading Hotels of the World",
+        "brand:wikidata": "Q834396",
+        "name": "The Leading Hotels of the World",
+        "tourism": "hotel"
+      }
+    },
     {
       "displayName": "7天酒店",
       "id": "7daysinn-af4cd2",


AntiCompositeNumber:
Hmm... I'm not sure if this is a brand in the OSM sense. The hotels are independently owned and operated (or operated by a hotel management company). Each is independently named and branded, without reference to LHW except in the footer of the website. Most run their own websites in addition to being listed at https://www.lhw.com/find-a-hotel/browse-by-list. They do have a loyalty program that confers benefits when staying at LHW member hotels. It seems more like a trade association to me, and https://en.wikipedia.org/wiki/The_Leading_Hotels_of_the_World describes it as a "marketing association".

data/brands/tourism/hotel.json
```diff
@@ -18,6 +18,17 @@
...
   },
   "items": [
+      {
+      "displayName": "The Leading Hotels of the World",
+      "locationSet": {"include": ["001"]},
+      "matchNames": ["lhw"],
+      "tags": {
+        "brand": "The Leading Hotels of the World",
+        "brand:wikidata": "Q834396",
+        "name": "The Leading Hotels of the World",
```
Cj-Malone, line 28: No strong opinion on if this is added from me, it doesn't feel like a normal brand.But people have mapped it as such, eg: https://www.openstreetmap.org/way/734574023#map=19/37.44357/25.33006 https://www.openstreetmap.org/way/218656514#map=19/32.05685/34.76235

However I do have a strong opinion on the name, it shouldn't be here.
```suggestion
```

PR #8956, opened 2023-11-18
Update charging_station.json

jmarchon1:
Fixed name of the Rivian Adventure Network - was in index as "Rivian Adventure Charger". Do let me know if I'm doing this wrong, which is very possible.

### DIFF
--- a/data/brands/amenity/charging_station.json
+++ b/data/brands/amenity/charging_station.json
@@ -201,12 +201,12 @@
       }
     },
     {
-      "displayName": "Rivian Adventure Charger",
+      "displayName": "Rivian Adventure Network",
       "id": "rivianadventurecharger-5598dd",
       "locationSet": {"include": ["us"]},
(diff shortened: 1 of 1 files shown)


UKChris-osm:
I see there are two entries currently for [Rivian](https://nsi.guide/index.html?t=brands&k=amenity&v=charging_station&tt=rivian), so would it be better to simply drop the word "_network_" and "_charger_" from the Waypoint version, or possibly both?

It looks like they are also expanding into Canada.

jmarchon1:
There are two charger networks from Rivian - one is called the "Rivian Adventure Network" (for the DC fast chargers) and the other is "Rivian Waypoints" (for the AC chargers). 
I only attempted to fix the former in this PR, I'll try the latter later today. I can't really code :-)

jmarchon1:
I've made a few more edits now. Let me know what you think.

PR #8965, opened 2023-11-23
Kenyan Brands

bkissin:


UKChris-osm:
Thanks @bkissin 

The `id` tag is generated automatically a little later on when running `npm run build`, so you don't need to worry about adding a placement for it at the moment.

PR #8966, opened 2023-11-23
Update clothes.json added tentree brand

philchisholm:
Adding the brand tentree https://www.tentree.com/

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -14,6 +14,17 @@
     }
   },
   "items": [
+        {
+      "displayName": "tentree",
+      "locationSet": {"include": ["ca","us"]},
+      "tags": {
+        "brand": "tentree",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Do Tentree have their own shops?

I couldn't find any in OSM at all - are they just a brand of clothing that is sold in other shops?

AntiCompositeNumber:
The only "tentree" store listed in https://www.tentree.com/pages/store-locator is 
> tentree Flagship Store
1924 8th St
Suite 115
Northwest, Washington 20001, US

All others are independent, self-branded outodoor stores that sell "select tentree products". So it doesn't seem like this brand has enough of a physical presence to be included in NSI.

philchisholm:
Looking at it more in depth you folks are right about this one, the brand is just sold in stores but does not have any physical location. My mistake.

PR #8968, opened 2023-11-24
New charging_station operators : EVC, MER, amend Blink

zymurgic:


data/operators/amenity/charging_station.json
```diff
@@ -2250,6 +2261,18 @@
         "operator:wikipedia": "en:Meridian Energy"
       }
     },
+    {
+      "displayName": "MER",
```
Cj-Malone, line 2265: There is already [Mer](https://nsi.guide/index.html?t=brands&k=amenity&v=charging_station&tt=mer#mer-b20835). Are they the same?

zymurgic: I did spot that soon after I did the pull request. You're right. There were duplicates in the wikidata, which I've now flagged.

PR #8973, opened 2023-11-26
Update bus.json

alessandrocarmeli:


Cj-Malone:
That should have fixed the file, but I'll leave the review to someone who know more about public transit tagging than me.

serhii-muchychka:
@alessandrocarmeli
I have limited locationSet to the Canary Islands only according to information from the Wikipedia. Let me know if it isn't

PR #8974, opened 2023-11-26
Update tram.json

alessandrocarmeli:


serhii-muchychka:
@alessandrocarmeli
I have limited locationSet to the Canary Islands only according to information from the Wikipedia. Let me know if it isn't

PR #7604, opened 2023-01-02
 Create and update some brand entries in Taiwan and merge 04244972

iigmir:
Some brands with considerable branches in Taiwan do not exist, so I create them.

PR #8983, opened 2023-11-30
Remove discontinued brand Kohihan in TW

assanges:
> brands/amenity/cafe

The franchisee entity (no. 97363127) was dissolved in 2014, see https://www.twincn.com/item.aspx?no=97363127

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -5517,26 +5517,6 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Looks like this brand was added in https://github.com/osmlab/name-suggestion-index/pull/7604 early this year, so I guess the brand still continue to operate in another legal entieies?

cc original PR author: @iigmir

PR #8984, opened 2023-11-30
Added Anejo Restaurant

ckpiper:


AntiCompositeNumber:
https://www.anejo.ca/en/locations/index.html shows only 5 locations. It seems like the other usages aren't part of this brand.

PR #8988, opened 2023-11-30
Add spendless au to shoes

ckpiper:


AntiCompositeNumber:
This brand already appears in NSI at [spendlessshoes-373058](https://nsi.guide/index.html?t=brands&k=shop&v=shoes&tt=spendless#spendlessshoes-373058).

PR #8989, opened 2023-11-30
Add house au to houseware

ckpiper:


data/brands/shop/houseware.json
```diff
@@ -4,6 +4,17 @@
     "exclude": {"generic": ["^houseware$"]}
   },
   "items": [
+    {
+      "displayName": "House",
+      "locationSet": {"include": ["au"]},
+      "matchNames": ["House Bed & Bath"],
```
AntiCompositeNumber, line 10:
```suggestion
```
Looking at https://www.house.com.au/store-locator, it seems like `House Bed & Bath` should have its own NSI entry unless that branding is being phased out.

data/brands/shop/houseware.json
```diff
@@ -4,6 +4,17 @@
...
+      "tags": {
+        "brand": "House",
+        "brand:wikidata": "Q117921987",
+        "name": "House",
+        "shop": "houseware"
+      }
+    },
```
AntiCompositeNumber, line 17:
```suggestion
    },
    {
      "displayName": "House Bed & Bath",
      "locationSet": {"include": ["au"]},
      "tags": {
        "brand": "House",
        "brand:wikidata": "Q117921987",
        "name": "House Bed & Bath",
        "shop": "houseware"
      }
    },
```

AntiCompositeNumber (changes requested):

data/brands/shop/houseware.json
ckpiper: I do not think it is being phased out. I added it to the matchNames as the wikidata page has it under "Also Known As". In the store locator it is under the House brand. It looks like it is more specific for selling Bed and Bath linens and items. So I could see it being it's own entry. But at the same time in the store locator, if you click on all the drop down menus with all the stores in the different regions, there are only 13 of them. Which is barely over the inclusion limit. Also House Bed & Bath I think would still classify under the `shop=houseware` category.

AntiCompositeNumber (approved):

PR #8990, opened 2023-11-30
Added SpeeDee to oil change and Auto Service

ckpiper:


data/brands/shop/car_repair.json
```diff
@@ -25,6 +25,20 @@
     }
   },
   "items": [
+    {
+      "displayName": "SpeeDee",
+      "locationSet": {"include": ["us"]},
+      "matchNames": [
+        "SpeeDee Oil Change & Auto Service",
+        "SpeeDee Oil Change and Auto Service"
```
AntiCompositeNumber, line 33: The name matcher treats `&` and `and` as equivalent, so you don't need to include both. See https://github.com/osmlab/name-suggestion-index/wiki/Contributing#matchnamesmatchtags-optional

AntiCompositeNumber (approved):

PR #8993, opened 2023-12-01
Add IE to DFS

KasperFranz:
This pull request adds IE to the list of countries for DFS, it is using the same wikidata as GB


Cj-Malone:
Thank you, I imagine there are more like this, PRs or issues are very welcome.

KasperFranz:
I am sure there are, I will create PRs when I see them :)

PR #8994, opened 2023-12-02
Add Spanish furniture retailer Ofiprix

sergimn:
Adds [Ofiprix](https://www.ofiprix.com/), a Spanish furniture retailer and manufacturer with several brick and mortar stores all around the country.

### DIFF
--- a/data/brands/shop/furniture.json
+++ b/data/brands/shop/furniture.json
@@ -1464,7 +1464,6 @@
       "displayName": "Ofiprix",
       "id": "ofiprix-53177d",
       "locationSet": {"include": ["es"]},
-      "matchTags": ["building/warehouse"],
       "tags": {
         "brand": "Ofiprix",
         "brand:wikidata": "Q116970185",
--- a/data/brands/shop/furniture.json
+++ b/data/brands/shop/furniture.json
@@ -1461,6 +1461,18 @@
       }
     },
     {
+      "displayName": "Ofiprix",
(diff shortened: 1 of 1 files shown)


data/brands/shop/furniture.json
```diff
@@ -1460,6 +1460,18 @@
         "shop": "furniture"
       }
     },
+    {
+      "displayName": "Ofiprix",
+      "id": "ofiprix-53177d",
+      "locationSet": {"include": ["es"]},
+      "matchTags": ["building/warehouse"],
```
Cj-Malone, line 1467: I've not seen this done for `building` before

sergimn: This changeset (https://www.openstreetmap.org/changeset/28786083) classified one of the stores as a "warehouse" so I thought that if someone is trying to classify it as such again, the brand should still come up and suggest a shop instead.

Note that usually Ofiprix stores keep large amounts of stock in display, so I guess the confusion comes from this.

In that specific case, in addition, the mapping is wrong as there's lots of other stores in the building, but this is another topic.

sergimn:
```suggestion
```

sergimn: Fixed in 4973cc568f2cb6609630c5d9cfa88100dd9de75c

AntiCompositeNumber (approved):

Cj-Malone:
Thanks everyone

PR #9010, opened 2023-12-06
Added Main Event Entertainment

ckpiper:
Added this to restaurant as it is the same type of establishment as Dave & Busters. Where you can eat food and then play some arcade games and bowl as well.

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -46,6 +46,19 @@
   },
   "items": [
     {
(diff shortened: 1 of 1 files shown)


AntiCompositeNumber:
The company is Main Event Entertainment, but looking at https://www.mainevent.com/ and the reporting around the merger with D&B I think the brand is just `brand=Main Event`.

ckpiper:
@AntiCompositeNumber Should be good now.

AntiCompositeNumber (approved):

PR #9012, opened 2023-12-06
Added Fastrac to Fuel

ckpiper:


LaoshuBaby:
Interesting, it’s quite common for it to be a gas station and a convenience store at the same time, but it’s a gas station, a convenience store, and a pizza shop (and provides delivery?), so I’m very confused about which category it should be classified under.

ckpiper:
I changed it to `brand=Fastrac`. I think the convenience store could be branded as `brand=Fastrac Cafe`.

PR #9014, opened 2023-12-07
Update Costco Locations

Scottie0001:


data/brands/shop/wholesale.json
```diff
@@ -70,11 +70,19 @@
       "id": "costco-2c74f6",
       "locationSet": {
         "include": [
+          "au",
           "ca",
+          "cn",
```
LaoshuBaby, line 75: Please note that the Costco brand corresponding to ["brand:wikidata"="Q715583"] obviously already has other sub-entries in CN and TW, and they are in the local language:

https://github.com/osmlab/name-suggestion-index/blob/ac614ede2caf80f7640c58e05dbcc1936f26c3f2/data/brands/shop/wholesale.json#L322-L339

https://github.com/osmlab/name-suggestion-index/blob/ac614ede2caf80f7640c58e05dbcc1936f26c3f2/data/brands/shop/wholesale.json#L291-L305

Because the languages ​​and characters in these areas are completely different from the Latin alphabet commonly used in Europe and the Americas, and even in Mainland of China and Taiwan where Chinese characters are used in the same area, the localized name of this brand is very different, so NSI needs to ensure that in each region provides presets that match local conditions, what was called "on the ground".

Therefore you should not list them all in the Global version of Costco's location list. Otherwise, two Costco entries would appear in the above areas at the same time, one with Chinese characters and one with Latin letters.

I have not checked all the regions you added. I hope you can check them yourself to make sure they are not included in other existing entries.

LaoshuBaby (changes requested):

PR #9020, opened 2023-12-08
Auto-create light rail & subway stations from routes

k-yle:
I have no idea how to test this... It is based on the equivalent files for trams and trains, in the same folder. This PR tries to apply the same logic for light_rail and subway routes.


k-yle:
I figured out how to test this in iD. Seems to work:

bhousel:
Seems ok, thanks!  Sorry I missed this when you first sent it.. 😅

PR #7780, opened 2023-02-17
Split Gull AU and NZ

chuqtas:
Split Gull AU and NZ.
First NSI pull request in some time, hopefully I've done it right...

PR #9021, opened 2023-12-08
Change countdown to woolworths

k-yle:
In New Zealand, "Countdown" is rebranding to Woolworths, to match the branding used in Australia. 

I have tried to make the NZ presets as close as possible to the AU presets, based on what was done in #7780

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -1766,17 +1766,6 @@
       }
     },
     {
-      "displayName": "Countdown",
-      "id": "countdown-5dc426",
-      "locationSet": {"include": ["nz"]},
-      "tags": {
-        "brand": "Countdown",
-        "brand:wikidata": "Q5176845",
-        "name": "Countdown",
-        "shop": "supermarket"
-      }
-    },
-    {
       "displayName": "Covirán",
       "id": "coviran-d9bffc",
       "locationSet": {"include": ["es", "pt"]},
@@ -8275,7 +8264,7 @@
       }
     },
     {
-      "displayName": "Woolworths",
+      "displayName": "Woolworths (Australia)",
       "id": "woolworths-c254c9",
       "locationSet": {"include": ["au"]},
       "tags": {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
When is the rebranding happening?

When I go to https://woolworths.co.nz/ it redirects to https://www.countdown.co.nz/ and I couldn't see any obvious sign that a rebrand was imminent.

k-yle:
Hey, the problem we're having is that many stores have already changed their signage over the past few months. However, when we try to change `brand` and `name` in OSM, NSI tells us to undo our changes.

This means people who are trying to update the rebranded stores are being told by iD to stop it and put the tags back.... That's what I'm trying to solve here

data/brands/shop/supermarket.json
```diff
@@ -1765,17 +1765,6 @@
...
-      "id": "countdown-5dc426",
-      "locationSet": {"include": ["nz"]},
-      "tags": {
-        "brand": "Countdown",
-        "brand:wikidata": "Q5176845",
-        "name": "Countdown",
-        "shop": "supermarket"
-      }
-    },
     {
```
AntiCompositeNumber, line 1768:
```suggestion
    {
      "displayName": "Countdown",
      "id": "countdown-5dc426",
      "locationSet": {"include": ["nz"]},
      "tags": {
        "brand": "Countdown",
        "brand:wikidata": "Q5176845",
        "name": "Countdown",
        "shop": "supermarket"
      }
    },
    {
```

UKChris-osm:
> So we should add the new brands but not remove Countdown or add matchNames.

I think if the rebrand is to be quite a long process, then having both brands available in the index would be the best course of action for now, rather than removing `Countdown`, so current Countdowns aren't suggested in iD to be updated just yet, but still allows `Woolworths` to be available for the stores that have been rebranded.

k-yle:
> [...] having both brands available in the index would be the best course of action for now, [...]

Thanks team, that sounds great, I didn't realise you could have 2 entries with the same `brand:wikidata` tag. I've updated the PR

AntiCompositeNumber (approved):

PR #9026, opened 2023-12-11
Update restaurant.json: Moving Osmow's from "restaurant" to "fast food"

andrepoiy:
It's not really a sit-down restaurant, more like a takeout counter but still offering seating for those who want to eat in. Therefore moving to fast food section.


Cj-Malone:
Lets review this in #9025

PR #9031, opened 2023-12-14
Added Allied Universal

ckpiper:


data/brands/office/security.json
```diff
@@ -4,6 +4,16 @@
     "exclude": {"generic": ["^security$"]}
   },
   "items": [
+    {
+      "displayName": "Allied Universal",
+      "locationSet": {"include": ["001"]},
```
LaoshuBaby, line 9: > Allied Universal operates in the [United States](https://en.wikipedia.org/wiki/United_States), the [United Kingdom](https://en.wikipedia.org/wiki/United_Kingdom), [Canada](https://en.wikipedia.org/wiki/Canada), [Mexico](https://en.wikipedia.org/wiki/Mexico), [Honduras](https://en.wikipedia.org/wiki/Honduras), [Puerto Rico](https://en.wikipedia.org/wiki/Puerto_Rico), [U.S. Virgin Islands](https://en.wikipedia.org/wiki/U.S._Virgin_Islands), and [Nicaragua](https://en.wikipedia.org/wiki/Nicaragua).[[2]](https://en.wikipedia.org/wiki/Allied_Universal#cite_note-business1-2)
>
> Source: https://en.wikipedia.org/wiki/Allied_Universal

So I feel like it’s not appropriate to say it is 001? Because at least I’ve never heard of it in Asia
The scope of business of such security companies may usually be to travel with clients around the world, but they do not necessarily take orders around the world. 😥

ckpiper: I will change to only the ones in the drop down of https://www.aus.com/offices

PR #9035, opened 2023-12-17
Update fast_food.json: added Guac Mexi Grill and Fat Bastard Burrito

andrepoiy:
These two are tex-mex chains in Canada with a large enough number of locations that it warrants being added to NSI.

I left the id sections blank since I don't know what to put there


Snowysauce:
The Wikidata item for Guac Mexi Grill was deleted; see #10303

Issue #7692, opened 2023-01-25
kiosk as a brand


wvanderp:
In the Netherlands, we have a chain of shops called `kiosk`. They are commonly found at railway stations and sell drinks and other food and non-food items. 

It is already included in the nsi as `kiosk-c47323`, including the wikidata tag.

My problem is that if I open the ID editor, it states that Kiosk is a suspicious name.

https://www.openstreetmap.org/edit#map=20/51.69093/5.29400

This is because it is a suspicious name in most parts of the world, and that is the right call.

My question is can we do something about it?  
Moving it to the named section would lose the value of the check.  
Is this something that needs to be solved by the ID editor?
Can we extend the section to have overrides for certain places?

bhousel:
I'll check this - I think the expected behavior is that a match on a brand name should override a match from the generic list.  It might be something that needs a fix on the iD side 🤔

wvanderp:
I did some digging today, and this is what I found:

When we build the `matchIndex`, we exclude matches with generic names. And thus, we won't match on the name.

I need to investigate further to find the actual problem and solution.

Cj-Malone:
There is a bit of a hack where you can add generic names to matchNames.

https://github.com/osmlab/name-suggestion-index/pull/8730#issuecomment-1746614430

Snowysauce:
The problem in the original post was fixed with the workaround (as visiting the link and applying the preset deactivates the warning), but I'll leave this issue open in hopes that eventually we'll have the resources to code in a proper fix.

PR #9044, opened 2023-12-19
workaround for #7692 kiosk not recogised as shop

wvanderp:
A suggestion in #7692 said to add the generic name as the match name to workaround the issue of discarding the match.

This pr will try this "fix".


LaoshuBaby:
As the issue quoted above said, the matching of entry name should take precedence over the generic name, because the generic name is just to prevent others from mistaking the general name as a brand - but now it is indeed a brand
 Therefore, I think this hack is acceptable if no major refactor are made to the current code framework.

PR #9052, opened 2023-12-20
Add and fix supermarkets in dk

mikini:
* add Løvbjerg
* add ABC Lavpris

* change Coop 365discount to new branding without Coop
* remove fakta in dk; has been transformed into 365discount
* add fakta in de; border shops still fakta branded

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -1620,9 +1620,7 @@
       "locationSet": {"include": ["dk"]},
       "matchNames": [
         "coop 365",
-        "coop365",
-        "coop 365discount",
-        "coop 365 discount"
+        "coop 365discount"
       ],
       "tags": {
         "brand": "365discount",
@@ -1875,6 +1873,7 @@
       "displayName": "Dagli'Brugsen",
       "id": "daglibrugsen-072052",
       "locationSet": {"include": ["dk"]},
+      "matchNames": ["DagligBrugsen"],
       "tags": {
         "brand": "Dagli'Brugsen",
         "brand:wikidata": "Q12307017",
@@ -2540,16 +2539,6 @@
       }
     },
     {
-      "displayName": "fakta",
-      "locationSet": {"include": ["de"]},
-      "tags": {
-        "brand": "fakta",
-        "brand:wikidata": "Q123942180",
-        "name": "fakta",
-        "shop": "supermarket"
-      }
-    },
-    {
       "displayName": "famila",
       "id": "famila-5d2816",
       "locationSet": {"include": ["de", "it"]},
@@ -2802,6 +2791,10 @@
       "displayName": "Føtex",
       "id": "fotex-072052",
       "locationSet": {"include": ["dk"]},
+      "matchNames": [
+        "Foetex",
+        "Fotex"
+      ],
       "tags": {
         "brand": "Føtex",
         "brand:wikidata": "Q1480395",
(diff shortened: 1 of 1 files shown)


KasperFranz:
This brings up a question - what are the criteria for the number of shops required to be added to NSI - Fakta Germany only got three shops

LaoshuBaby:
> This brings up a question - what are the criteria for the number of shops required to be added to NSI - Fakta Germany only got three shops

For a canonical brand, 50 stores are usually needed, and at least 30 are recommended.  For countries where OSM data is seriously lacking and the country is small or the local mapper considers it important, we think at least 10 are needed.

However, if this brand operates in Germany and Denmark, the number of stores can be totaled

KasperFranz:
> > This brings up a question - what are the criteria for the number of shops required to be added to NSI - Fakta Germany only got three shops
> 
> For a canonical brand, 50 stores are usually needed, and at least 30 are recommended. For countries where OSM data is seriously lacking and the country is small or the local mapper considers it important, we think at least 10 are needed.
> 
> However, if this brand operates in Germany and Denmark, the number of stores can be totaled

I think it would be good to get that documented somewhere so we do not flood NSI with brands/operators etc that is less than that threshold.

In this instance Fakta denmark is closed

mikini:
Those numbers are a surprise to me. I have found no eligibility requirements anywhere in the current documentation. Who have decided those and how are they supposed to be upheld?

For a small country as [Denmark](https://en.wikipedia.org/wiki/Denmark) (~6 mio. inhabitants) adhering to those would make NSI sparsely populated.

Both ABC Lavpris and Løvbjerg added in this PR would not meet the 30 store threshold with their [15](https://www.abc-lavpris.dk/butikker) & [17](https://www.lovbjerg.dk/butikker-aabningstider) stores respectively, but I consider them as well-known chains.

I would probably not have added fakta DE by itself with its 3 stores but it came about from reusing the DK entry from the recent transformation to 365discount in DK (~300 stores).

LaoshuBaby:
> I think it would be good to get that documented somewhere

> Those numbers are a surprise to me. I have found no eligibility requirements anywhere in the current documentation.

It was not documented at first and was only used as a magic number in nsi-collector, but it has been recorded in this project for some time.

> but I consider them as well-known chains.

Yes, so I've said above that if it is well-known in the local or there are many factors such as population restrictions, it is really difficult to meet the number of branches, but it is well-known enough, at least 10 is the final bottom line.

KasperFranz:
> If the fakta shares same entry in Germany and Denmark (shares the same wikidata qid), it is acceptable, and the total number in the two countries can meet the notability requirements, it can be added.

They are two different qid [Q123942180](https://www.wikidata.org/wiki/Q123942180) and [Q3172238](https://www.wikidata.org/wiki/Q3172238).


The requirement for minimum shops are not very visible, so it might be worth getting it better documented

mikini:
I've removed fakta completely and polished some matchNames for Danish supermarkets, hopefully the PR is ok for merging now?

Just as a thought exercise to put things in an international perspective for the threshold values; the largest Danish national supermarket Bilka, known by all Danes, has [18 locations](https://www.bilka.dk/kundeservice/info/find-din-bilka/c/find-din-bilka/).

mikini:
Also added more elaborate instructions about this to the ["Judge Case" wiki](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case/_compare/209b66d963e4c85a631021bbc53a5cb673a90e6b...91928384011d47ba498854eec454e73c6443f241), and attempted a little more general cleaning.

KasperFranz:
LGTM

PR #9059, opened 2023-12-22
Add Les Petites Cantines

versgui:
French charitable organization

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -3188,6 +3188,18 @@
       }
     },
     {
+      "displayName": "Les Petites Cantines",
+      "locationSet": {
+        "include": ["fr"]
+      },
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
I searched that it is more like a public welfare organization, and that participants cook and exchange food with each other to enhance the connection between people. Is their any tag way better than tag it as `amenity=restaurant`?

versgui:
@LaoshuBaby Thanks for your feedback! I didn't find any more relevant tag.

You're right in your definition. The "preparing a meal together" dimension is added on top of that, which is a rather special principle.
In the end, the vast majority of participants pay more or less for their meal. Some people can just come in, eat a meal and pay for it, which is what a restaurant is all about.

bhousel:
I'm going to merge this and move it into the `social_facility` category..

PR #9065, opened 2023-12-26
Make Chili's worldwide

### LINKED ISSUE #9057: Include International Chili's Locations
According to [Wikipedia](https://en.wikipedia.org/wiki/Chili%27s#Locations), Chili's has a global presence, with locations all around the world and not just in the US. It would be really helpful if you could include these locations in the [index](https://nsi.guide/?id=chilis-96af40), or at least in Mexico, where the restaurant is quite popular outside of the US. Thank you.

andrewpmk:
Make Chili's (restaurant chain) worldwide

List of countries with Chili's locations: https://en.wikipedia.org/wiki/Chili%27s#Locations

Fix #9057

### DIFF
--- a/data/brands/amenity/pub.json
+++ b/data/brands/amenity/pub.json
@@ -49,6 +49,17 @@
       }
     },
     {
+      "displayName": "Antic",
+      "id": "antic-eaed91",
+      "locationSet": {"include": ["gb"]},
+      "note": "No name for Antic brand as each pub is named differently.",
+      "tags": {
+        "amenity": "pub",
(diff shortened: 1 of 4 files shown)


LaoshuBaby:
In #9066 (which I just merged) you also submitted the results of npm run build on other non-major changes to this PR.
Yes, it is necessary to execute formatter for normalization, but before each release, the maintainer responsible for the release will execute the checklist
I suggest that when you submit the PR later, do not modify non-mainly involved parts. This can reduce potential conflicts between different branches.

PR #9082, opened 2023-12-28
DriveTest

andrewpmk:
DriveTest driver examination centres in Ontario


data/operators/office/government.json
```diff
@@ -603,6 +603,19 @@
...
       }
     },
+    {
+      "displayName": "DriveTest",
+      "id": "drivetest-a69e09",
+      "locationSet": {
+        "include": ["ca-on.geojson"]
+      },
+      "tags": {
+        "name": "DriveTest",
```
AntiCompositeNumber, line 613:
```suggestion
        "government": "transportation",
        "name": "DriveTest",
```

AntiCompositeNumber (commented):
Looks like this is a pseudogovernmental office operated by Serco for the Ontario Ministry of Transportation. `office=government` is probably still the best tag for this, even though it has been privatized. I think this might actually fit better in the `brands/` tree rather than the `operators/` tree, which would allow us to specify `operator=Serco`.

andrewpmk:
> Looks like this is a pseudogovernmental office operated by Serco for the Ontario Ministry of Transportation. `office=government` is probably still the best tag for this, even though it has been privatized. I think this might actually fit better in the `brands/` tree rather than the `operators/` tree, which would allow us to specify `operator=Serco`.

There isn't a file office/government.json in brands. Only in operator

bhousel:
Thanks, it seems ok for this to be in the operator file if there isn't a brand file.

PR #9087, opened 2023-12-29
Add shop/deli to Subway

andrewpmk:
Add shop/deli to "matchNames" to Subway (amenity=fast_food). Some users are selecting deli by mistake (fixed a bunch of them).

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -7841,6 +7841,7 @@
         "subway sandwiches",
         "subway uk & ireland"
       ],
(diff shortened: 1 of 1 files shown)


AntiCompositeNumber:
I think it might be a better idea to just add it to the `food` group in [config/matchGroups.json](https://github.com/osmlab/name-suggestion-index/blob/523e8fab2cef0f43384b3c86b081b43400531967/config/matchGroups.json#L109). There are a number of sandwich fast food shops that are probably mis-tagged as delis.

PR #9096, opened 2023-12-30
Added a larger amount of missing brands in Sweden

matthiasfeist:


data/brands/amenity/cafe.json
```diff
@@ -3663,6 +3663,19 @@
...
       }
     },
+    {
+      "displayName": "Thelins konditori",
+      "id": "thelinskonditori-46f791",
+      "locationSet": {"include": ["se"]},
+      "tags": {
+        "amenity": "cafe",
+        "brand": "Thelins konditori",
+        "brand:wikidata": "Q124048792",
```
LaoshuBaby, line 3673: It only has 18 branches shop. Does it really satisfy the notability?

matthiasfeist: Yeah that's a good question. I've seen some other smaller chains in the index before, but I'm also fine with not adding it :)

bhousel: I think 18 is ok as a number, but the first thing I'm going to look at is Q124048792 and right now it looks pretty empty or spammy.  I worry that Wikidata admins might remove it if you don't add some supporting details from other sources.

matthiasfeist: that's all the data I have for them right now. What do you think would be missing?

Snowysauce:
The Wikidata item for Holy Greens was deleted; see #10303

PR #9100, opened 2023-12-31
Add Johannesburg City Parks and Zoo

mustaqimM:


LaoshuBaby:
It seems that [the operator you mentioned](https://en.wikipedia.org/wiki/Johannesburg_City_Parks_and_Zoo) is more like operating parks in and around Johannesburg, rather than a nature reserve.

https://wiki.openstreetmap.org/wiki/Tag:leisure%3Dnature_reserve

Maybe name-suggestion-index/data/operators/leisure/park.json suits you better?

Of course, if you think it is more suitable for ·leisure=nature_reserve·, then please ignore this comment. Because after all, you are a local mapper who knows more about the local area.

mustaqimM:
Hi @LaoshuBaby. It administers both parks and nature reserves. `"matchTags": ["leisure/park"],
      "tags": {
        "leisure": "nature_reserve",` matches both nature reserves and local parks, right? I did at first add it to `park.json` but moved it to include nature reserves. If that's not correct, please let me know.

Cj-Malone:
The current `matchTags` will recommend existing OSM `leisure=park` objects get turned into `leisure=nature_reserve`. If both categories are valid it should be listed in both files.

Check it out with iD in the UK with:
```
amenity=fast_food
name=Hungry Horse
```
gets recommended to
```
amenity=pub
```
because [match tags](https://nsi.guide/index.html?t=brands&k=amenity&v=pub&tt=hungry%20horse#hungryhorse-eaed91).
vs
```
amenity=atm
name=Bank of Scotland
```
which will be recommended brand tags, but not recommended to turn into a bank even though it's in bank.json because it's in both files https://nsi.guide/index.html?t=brands&tt=bank%20of%20scotland

PR #9102, opened 2023-12-31
Update ferry.json (Network Mistral)

Patchi-Osm:
Update the first entry with Wikidata and remove second (and "same") entry.


Cj-Malone:
If the are the same `"matchNames": ["Réseau Mistral"],` should be added to the first entry so iD recommends the updated.

PR #9104, opened 2024-01-01
Add Containers for Change

Scottie0001:
Add Containers for Change in Queensland and Western Australia

### DIFF
--- a/data/brands/amenity/recycling.json
+++ b/data/brands/amenity/recycling.json
@@ -93,6 +93,38 @@
       }
     },
     {
+      "displayName": "Containers for Change (Queensland)",
+      "id": "",
+      "locationSet": {"include": ["au-qld.geojson"]},
+      "preserveTags": ["^name"],
+      "tags": {
+        "amenity": "recycling",
+        "brand": "Containers for Change (Queensland)",
+        "brand:wikidata": "Q124060523",
+        "name": "Containers for Change",
+        "recycling_type": "centre",
+        "recycling:cans": "yes",
+        "recycling:glass_bottles": "yes",
+        "recycling:plastic_bottles": "yes"
+      }
+    },
+    {
+      "displayName": "Containers for Change (Western Australia)",
+      "id": "",
+      "locationSet": {"include": ["au-wa.geojson"]},
+      "preserveTags": ["^name"],
+      "tags": {
+        "amenity": "recycling",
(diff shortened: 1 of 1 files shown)


AntiCompositeNumber:
It looks like the Queensland and the WA programs use the same brand, but with different operators (Container Exchange and Western Australia Return Recycle Renew Limited, respectively). They have the same logo, same website, and same telephone shortcode. I think it would be better to use the same `brand=` and `brand:wikidata=` for both, but use separate `operator=` and `operator:wikidata=` tags for the two programs.

Scottie0001:
Each return point is operated by different companies. The Containers for Change program is managed by a government organisation in each state.

data/brands/amenity/recycling.json
```diff
@@ -92,6 +92,38 @@
...
       }
     },
+    {
+      "displayName": "Containers for Change (Queensland)",
+      "id": "",
+      "locationSet": {"include": ["au-qld.geojson"]},
+      "preserveTags": ["^name"],
+      "tags": {
+        "amenity": "recycling",
+        "brand": "Containers for Change (Queensland)",
```
AntiCompositeNumber, line 102:
```suggestion
        "brand": "Containers for Change",
```

data/brands/amenity/recycling.json
```diff
@@ -92,6 +92,38 @@
...
+      }
+    },
+    {
+      "displayName": "Containers for Change (Western Australia)",
+      "id": "",
+      "locationSet": {"include": ["au-wa.geojson"]},
+      "preserveTags": ["^name"],
+      "tags": {
+        "amenity": "recycling",
+        "brand": "Containers for Change (Western Australia)",
```
AntiCompositeNumber, line 118:
```suggestion
        "brand": "Containers for Change",
```

AntiCompositeNumber (commented):
In that case, let's just remove the disambiguators from `brand=`.

bhousel:
Yeah this looks like just adding 2 of the same thing..  If they had different operators then it would make sense, but these items don't have `operator` tags anyway.  I'll just merge and combine.

bhousel:
update: a Wikidata admin removed [Q124060545](https://www.wikidata.org/wiki/Special:Log?page=Q124060545) in April, so it sounds like they also agree that there doesn't need to be multiple entries for this brand.

PR #9108, opened 2024-01-02
Split EB Games AU into separate entry, update brand:wikidata

andrewtwelch:
I've split the AU/NZ EB Games from the Canadian one, they're both subsidiaries of GameStop but are two individual subsidiaries, and with separate WikiData entries.

### DIFF
--- a/data/brands/shop/video_games.json
+++ b/data/brands/shop/video_games.json
@@ -33,7 +33,7 @@
       }
     },
     {
-      "displayName": "EB Games",
+      "displayName": "EB Games (Canada)",
       "id": "ebgames-c56656",
       "locationSet": {
         "include": ["ca"]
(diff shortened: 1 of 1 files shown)


data/brands/shop/video_games.json
```diff
@@ -48,6 +48,21 @@
         "shop": "video_games"
       }
     },
+    {
+      "displayName": "EB Games",
```
LaoshuBaby, line 52: Sorry, the current displayName requirement for each entry is unique, so now there are two "EB Games" will cause an error

I suggest you change them to distinctive names such as "EB Games (Canada)" and "EB Games (AU & NZ)" respectively.

LaoshuBaby (changes requested):

data/brands/shop/video_games.json
andrewtwelch: I've just updated them both.

PR #9109, opened 2024-01-02
Update outdoor.json - Add Bever store chain in NL

Mitsjol:
Added the Bever outdoor goods store chain in the Netherlands.


Cj-Malone (approved):
LGTM, thanks

PR #9116, opened 2024-01-03
Added Trent Limited to Clothes

ckpiper:


AntiCompositeNumber:
Based on https://trentlimited.com/, it looks like Trent Limited is a company that owns brands, some of which have stores, rather than a brand in and of itself.

* https://www.westside.com/pages/store-locator appears to be broken or have no useful information.
* https://starbazaarindia.com/apps/s/storelocator looks like a decent candidate for inclusion
* https://misbu.com/apps/s/storelocators has only 3 stores
* http://bookerindia.net/Home.aspx has only 8 stores
* https://www.zudio.com/apps/s/storelocators?type=Zudio has 460 stores and would be a good candidate for inclusion

ckpiper:
@AntiCompositeNumber Good Catch. I think misbu and bookerindia do not need to be added, but I did Zudio and Star Bazaar.

AntiCompositeNumber (approved):

PR #9125, opened 2024-01-06
Update pastry.json

JoeyKentin:
Added Maison Dandoy a Belgian (mostly in Brussels) chain of biscuiteries


LaoshuBaby:
It seems that this pastry chain brand does not have many stores.

Also, if it's a brand concentrated around a specific city, maybe you could try a point and radius based constraint like this

图片

PR #9159, opened 2024-01-16
Add SuperHotel

DivyaB1:
Hi,

I'm adding SuperHotel - https://www.superhotel.co.jp/

Thanks,

### DIFF
--- a/data/brands/tourism/hotel.json
+++ b/data/brands/tourism/hotel.json
@@ -19,6 +19,16 @@
   },
   "items": [
     {
+      "displayName": "SuperHotel",
+      "locationSet": {"include": ["jp","mm"]},
+      "tags": {
+        "brand": "SuperHotel",
+        "brand:wikidata": "Q11313858",
+        "name": "SuperHotel",
+        "tourism": "hotel"
+      }
+    },
+    {
       "displayName": "7天酒店",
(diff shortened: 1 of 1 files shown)


data/brands/tourism/hotel.json
```diff
@@ -18,6 +18,16 @@
     }
   },
   "items": [
+    {
+      "displayName": "SuperHotel",
+      "locationSet": {"include": ["jp","mm"]},
+      "tags": {
+        "brand": "SuperHotel",
+        "brand:wikidata": "Q11313858",
+        "name": "SuperHotel",
```
AntiCompositeNumber, line 27:
```suggestion
      "displayName": "Super Hotel",
      "locationSet": {"include": ["jp","mm"]},
      "tags": {
        "brand": "Super Hotel",
        "brand:wikidata": "Q11313858",
        "name": "Super Hotel",
        "name:ja": "スーパーホテル",
```
Based on https://www.superhoteljapan.com/en/company.html and https://overpass-turbo.eu/s/1GeQ it looks like "Super Hotel" is the more common representation when not stylized as SUPERHOTEL.
It looks like the brand primarily uses the English name, but we should at least include the Japanese in `name:ja`.

AntiCompositeNumber (changes requested):

PR #9162, opened 2024-01-17
Add Blast Fitness

DivyaB1:
Hi,

Adding Blast Fitness - https://www.blastfitness.com/

Thanks,


AntiCompositeNumber:
Hrm, https://www.blastfitness.com/locations/ only lists 7 locations. It says that there are also locations in California and Massachusetts, but doesn't list any. I only see 1 already mapped. Do we know how many locations there actually are?

DivyaB1:
> Hrm, https://www.blastfitness.com/locations/ only lists 7 locations. It says that there are also locations in California and Massachusetts, but doesn't list any. I only see 1 already mapped. Do we know how many locations there actually are?

Hi,

So I see the 7 locations with their addresses in Arizona and Illinois, they do say there are locations in California and Mass but there is no list, same as you. Since there are 7 locations shouldn't they be merged?

bhousel:
Yeah 7 locations is really low, but I merged it anyway.  Maybe they'll grow?
For tiny brands like this it's easier to just update the 7 locations on OSM and be done.

PR #9165, opened 2024-01-17
Add Servpro

DivyaB1:
Hi

Adding Servpro - https://www.servpro.com/
This company provides multiple services, majorly cleaning and restoration. It could also belong under craft=restoration.
I think craft=cleaning is also reasonable, if you think otherwise let me know.

Thanks,

### DIFF
--- a/data/brands/craft/cleaning.json
+++ b/data/brands/craft/cleaning.json
@@ -4,6 +4,16 @@
     "exclude": {"generic": ["^cleaning$"]}
   },
   "items": [
+     {
+      "displayName": "Servpro",
+      "locationSet": {"include": ["ca", "us"]},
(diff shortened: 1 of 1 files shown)


AntiCompositeNumber:
`craft=cleaning` seems to be the most common tag of what's already been mapped, but I agree that there isn't an obvious tag for this sort of business.

DivyaB1:
> `craft=cleaning` seems to be the most common tag of what's already been mapped, but I agree that there isn't an obvious tag for this sort of business.

So is it ready to be merged?

bhousel:
I don't see anything on their website about them having branch locations, but they must have some locations to base their business out of, so ok.

PR #9187, opened 2024-01-24
Add Icelandic fuel stations

Cj-Malone:
https://orkan.is/english/orkan-stations/
https://www.n1.is/en/locations/
https://www.olis.is/english/stations


kjonosm:
Thanks!

PR #9188, opened 2024-01-24
Add subway transit networks in India

Vonter:


LaoshuBaby:
Thank you for making the first edit! Indeed, currently only Delhi's subway is recorded in NSI (because it is indeed large enough to be remembered), but we have even overlooked cities like Mumbai.

Subway networks usually have a significant impact not only on maps but also on future city planning. Therefore, even if some networks only have one operational line, they are still worth being included.

PR #9192, opened 2024-01-25
Remove PLN

NFarrasP:
"PLN" is the acronym of "Perusahaan Listrik Negara". Both refer to the same company owned by Indonesian government.


kjonosm:
Thanks!

PR #9197, opened 2024-01-25
Update hostel.json - Add 'Stayokay' hostel chain in NL

Mitsjol:
Adding 'Stayokay', a hostel chain in the Netherlands.


kjonosm:
Thanks

PR #9206, opened 2024-01-29
Update fitness_centre.json to add Active Fitness, Switzerland

### LINKED ISSUE #9205: Add Activ Fitness
https://www.activfitness.ch/studios/ lists 121 studios in Switzerland.

habi:
Fixes #9205

### DIFF
--- a/data/brands/leisure/fitness_centre.json
+++ b/data/brands/leisure/fitness_centre.json
@@ -73,6 +73,16 @@
       }
     },
     {
+      "displayName": "Activ Fitness",
+      "locationSet": {"include": ["ch"]},
+      "tags": {
+        "brand": "Activ Fitness",
+        "brand:wikidata": "Q123747318",
+        "leisure": "fitness_centre",
+        "name": "Activ Fitness"
+      }
+    },    
+    {
       "displayName": "Alex Fitness",
       "id": "alexfitness-2b2465",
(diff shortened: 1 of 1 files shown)


habi:
The chain is called "Activ Fitness", without an "e".
The JSON is correct, the commit message not :)

data/brands/leisure/fitness_centre.json
```diff
@@ -72,6 +72,16 @@
         "sport": "kickboxing"
       }
     },
+    {
+      "displayName": "Activ Fitness",
```
Cj-Malone, line 76: Since it's probably a common mistake, we can add a match name.

```suggestion
      "displayName": "Activ Fitness",
      "matchNames": ["Active Fitness"],
```

That way iD will ask users if [POIs with the spelling mistake](https://www.openstreetmap.org/node/7503188456) need updating.

Cj-Malone:
Thank you

data/brands/leisure/fitness_centre.json
habi: I didn't know that this is possible.
Then I would not have made https://www.openstreetmap.org/changeset/146830770 this afternoon, in which I corrected the wrong name of some Active Fitnesses to Activ Fitnesses.

PR #9208, opened 2024-01-29
Add Randstad Digital in /data/brands/office/it.json

cayenne17:
I haven't tested the build


bhousel:
Weird to use the note property to ask a question..
No, you don't usually need to make different Wikidata QIDs for each country.  
Each Wikidata QID represents a unique concept.  "Q2872338" is the entity "Randstad Digital".  It only needs different QID for different country if there is some reason why it's a meaningfully different thing in each country.

PR #9215, opened 2024-02-02
Add Clean Eatz

Cj-Malone:
OSM has most of these as `amenity=restaurant`, however it seems more like fastfood or cafes to me. Anyone have an opinion?

https://locations.cleaneatz.com/

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -587,6 +587,15 @@
         "takeaway": "yes"
       }
     },
+       {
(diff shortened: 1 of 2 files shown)


LaoshuBaby:
https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#is-an-amenity-a-restaurant-or-fast_food

It seems that there is currently only guidance on distinguishing `restaurant` and `fast_food`. How do you know how to distinguish `restaurant`s or `cafe`s?

PR #9219, opened 2024-02-04
Add some brands in China

quinncnl:


LaoshuBaby (commented):
杨国福麻辣烫 already existed
https://github.com/osmlab/name-suggestion-index/blob/58b328feaf3f3e01a7509cfe31ff47a7d020317a/data/brands/amenity/restaurant.json#L7964

PR #9222, opened 2024-02-06
Added What the Pitta to restaurant.json

rskedgell:


LaoshuBaby:

I agree with the point above, if only 4 stores are really not enough to have notability

rskedgell:
> https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#notability

Thanks, noted. I don't think I'd seen that page on the wiki.

PR #9226, opened 2024-02-07
Update Petrom fuel.json

dan13l9:
Actual Petrom brand is available only in Romania and Moldova. Additionally i added Morocco company Petrom.


UKChris-osm:
You have two entries with the same name @dan13l9 - you'll need to amend the **display name** to be unique for each entry.

PR #9228, opened 2024-02-07
add Marks & Spencer on clothes.json

dan13l9:
Internationally, Marks & Spencer is a clothing retail chain


bhousel:
Looks like we have it already as a "Department Store"
https://nsi.guide/index.html?t=brands&k=shop&v=department_store#marksandspencer-679c9d

PR #9232, opened 2024-02-08
Update jewelry.json - Add 'Schaap en Citroen' in NL

Mitsjol:
Adding 'Schaap en Citroen', a jewelry store chain in the Netherlands.


Cj-Malone:
Thank you

PR #9243, opened 2024-02-13
Added support for TECN and TECX (Belgium/Wallonia)

bxl-forever:
TEC is a public operator operating in about half of Belgium.  They have split their network in 6 groups: TECB, TECC, TECH, TECL, TECN and TECX.  The last 2 entries were missing in the validation list.
https://wiki.openstreetmap.org/wiki/WikiProject_Belgium/Conventions/Bus_stops#Belgian_transport_companies


LaoshuBaby:
Thank you very much for contributing your local knowledge to NSI, but we do not want you to edit the /dist folder because they are compilation products and will be regenerated every time you re-run the build. You should find the corresponding file to edit under the /data folder.

bxl-forever:
Interesting, but the corresponding entries are already in ``data/transit/route/bus.json``.

LaoshuBaby:
> Interesting, but the corresponding entries are already in `data/transit/route/bus.json`.

Yes, but in `/transit` tree folder, entry are required of `"network:wikidata"` as outside identifier, only a `"operator:wikidata"` is insufficient. You can try to add `"network:wikidata"` for each cities' bus system.

`/brands` corresponds to `brand:wikidata`, while `/operators` corresponds to `operator:wikidata`. Among the brands you submitted, if the build script cannot find the corresponding tag under this tree folder (`/transit`), it will consider it as an incomplete entry and will not be included in the preset.

bxl-forever:
Thanks for the explanation.  I’ll close this one and will open a new PR with the correct file.

PR #9247, opened 2024-02-14
Removed name on Robidog waste baskets

dafadllyn:
Removed name on waste baskets of brand "Robidog" as per https://wiki.openstreetmap.org/wiki/Good_practice#Don't_use_name_tag_to_describe_things, added missing tag "vending=excrement_bags".

### DIFF
--- a/data/brands/amenity/waste_basket.json
+++ b/data/brands/amenity/waste_basket.json
@@ -37,7 +37,7 @@
         "amenity": "waste_basket",
         "brand": "Robidog",
         "brand:wikidata": "Q2159689",
-        "name": "Robidog"
+        "vending": "excrement_bags"
       }
     }
   ]


data/brands/amenity/waste_basket.json
```diff
@@ -37,7 +37,7 @@
         "amenity": "waste_basket",
```
LaoshuBaby, line 37: Consider related #9248

The vending=excrement_bags tag may not necessarily apply to trash cans? Unless each trash can is also a vending machine, but in that case why use two entries?

dafadllyn: @LaoshuBaby:

> The vending=excrement_bags tag may not necessarily apply to trash cans?

[Robidog](https://de.wikipedia.org/wiki/Robidog) is a brand of waste baskets with integrated poop bag dispensers.

> Unless each trash can is also a vending machine, but in that case why use two entries?

I came across three ways of tagging Robidogs:

1. `amenity=waste_basket` + `vending=ecrement_bags` on a single node
2. `amenity=waste_basket` on one node and `amenity=vending_machine` + `vending=excrement_bags` on a separate node next to it
3. `amenity=vending_machine` + `vending=excrement_bags` + `bin=yes`

(Note that variant 3 has the disadvantage that `bin=yes` may or may not mean that waste basket is mapped separately (see https://wiki.openstreetmap.org/wiki/Key:bin). Besides people looking for waste baskets may not find it because `bin=yes` is a secondary (quality) tag, not a main (feature) tag.)

bhousel:
I don't want to add the `vending` tag to waste baskets, but I'm ok with removing the names.. I did this in efa9395c5

PR #9248, opened 2024-02-14
Fixed Robidog poop bag dispenser

dafadllyn:
Removed name from dog poop bag dispenser and waste basket of brand "Robidog" as per https://wiki.openstreetmap.org/wiki/Good_practice#Don't_use_name_tag_to_describe_things.


bhousel:
> Removed name from dog poop bag dispenser and waste basket of brand "Robidog" as per https://wiki.openstreetmap.org/wiki/Good_practice#Don't_use_name_tag_to_describe_things.

This isn't a case where the name is being used to describe things, but ok we can remove it if you really hate the name....  🙄

PR #9251, opened 2024-02-17
Add Pizza Salvatoré

wolfy1339:
Adds Pizza Salvatoré which has restaurant (sit-in with takeout and delivery) locations and fast-food (takeout only and some offer delivery) locations


Cj-Malone:
Thank you

PR #9257, opened 2024-02-19
Update bus.json

gosgtan:
Hi.
Changed from "TUC Cambrésis" to only "TUC", because "TUC" already means "Transports Urbains du Cambrésis". So this avoids the duplicate of "Cambrésis". I also added associated tags, the most important/common match names, and the current operator of the network.
EDIT : Deleted the "wikipedia" tag

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -16850,12 +16850,18 @@
       }
     },
     {
-      "displayName": "TUC Cambrésis",
+      "displayName": "TUC (Cambrésis)",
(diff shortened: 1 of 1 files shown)


gosgtan:
😔

LaoshuBaby:
Hello! Thank you for try to make your first changes in NSI！

_Besides, I noticed you former closed PR #9256, and found you are using a new branch for editing code. For example, when you realize that Wikipedia is not needed, you can continue editing on patch-1. All changes on patch-1 will be synchronized in the PR._

PR #9265, opened 2024-02-23
Update bus.json - added network:wikidata for TECN and TECX

bxl-forever:
TECN (TEC Namur) and TECX (TEC Luxembourg) are two regional networks operated by TEC (Public transport in the south of Belgium).
This commit adds "network:wikidata" for both of them.  (Yes, they point to the same wikidata entry, this is normal and expected.)

---

Related #9243

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -15534,6 +15534,7 @@
       "locationSet": {"include": ["be", "fx"]},
       "tags": {
         "network": "TECN",
+        "network:Q3512080": "Q3512080",
         "operator": "TEC",
         "operator:wikidata": "Q366922",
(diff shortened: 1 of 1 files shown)


data/transit/route/bus.json
```diff
@@ -15534,6 +15534,7 @@
       "locationSet": {"include": ["be", "fx"]},
       "tags": {
         "network": "TECN",
+        "network:Q3512080": "Q3512080",
```
LaoshuBaby, line 15537: Should be "network:wikidata" not "network:Q*" this place and the below

LaoshuBaby (changes requested):

PR #9268, opened 2024-02-24
Add Indomaret Point and Maxx Coffee

NFarrasP:
Indomaret Point and Maxx Coffee are coffee shop chains in Indonesia.

Indomaret Point brand is owned by Indomaret. Indomaret Point usually located inside the Indomaret itself.
Maxx Coffee stores are located in Indonesia's big cities.

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -2497,11 +2497,10 @@
     },
     {
       "displayName": "Maxx Coffee",
-      "locationSet": {
-        "include": ["id"]},
+      "locationSet": {"include": ["id"]},
       "tags": {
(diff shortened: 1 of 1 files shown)


kjonosm:
Is it `Max Coffee` oder `Maxx Coffee`? You use both variants...

UKChris-osm:
Is this the brand for Point Coffee: https://pointcoffee.id

If so, I'd suggest giving it its own Wikidata and branding.

NFarrasP:
> Is this the brand for Point Coffee: https://pointcoffee.id
> 
> If so, I'd suggest giving it its own Wikidata and branding.

Right, I will fix it

Issue #8406, opened 2023-07-17
Incorrect name=ALDI Nord/ALDI Süd for shop=supermarket in Germany


fabi003:
NSI for Germany contains incorrect date for

https://nsi.guide/index.html?t=brands&k=shop&v=supermarket#aldinord-0a839e
and
https://nsi.guide/index.html?t=brands&k=shop&v=supermarket#aldisud-0a839e

Data at the moment:

brand=ALDI Nord
brand:wikidata=Q41171373
name=ALDI Nord
shop=supermarket

right data ("ALDI Nord" brand):

brand=ALDI Nord
brand:wikidata=Q41171373
name=ALDI (as written on the shop) or name=Aldi (as written as German word)
website=https://www.aldi-nord.de/
shop=supermarket

Sample picture of a ALDI Nord brand shop in Germany: https://www.lebensmittelzeitung.net/galerien/So-sieht-die-neue-Filiale-Nummer-1-aus-1545

right data ("ALDI Süd" brand):

brand=ALDI Süd
brand:wikidata=Q41171672
name=ALDI (as written on the shop) or name=Aldi (as written as German word)
website=https://www.aldi-sued.de/
shop=supermarket

Sample picture of a ALDI Süd brand shop in Germany: https://www.focus.de/finanzen/news/koelner-filiale-aldi-stampft-kassen-service-ein-doch-es-liegt-nicht-am-diebstahl_id_184975891.html

"ALDI" or "Aldi" is an abbreviation for "Albrecht Diskont" so it could also be "AlDi", which is not used is Germany. Most people use "Aldi" for e.g. "I go to Aldi."

fabi003:
As additional background, why Aldi has two brands for the same shop name, as english Wikipedia don't has this:

In 1961, the discount chain was split into two, one operated by each Albrecht son, which divided Germany as food market into the north for "Aldi Nord" (Aldi North) and south "Aldi Süd" (Aldi South) before they expanded into the word. So the same shop name has two brands.

kjonosm:
@fabi003 Current numbers queried by overpass:

882 x `name=Aldi`
68 x `name=ALDI`
410 x `name=Aldi Süd`
1410 x `name=ALDI Süd`
191 x `name=Aldi Nord`
1447 x `name=ALDI Nord`

950 x `name=ALDI` + `name=Aldi`
[...]

PR #9274, opened 2024-02-26
Divide Germany (along the Aldi equator)

kiddikai:
This solves #8406.

I carefully traced the border between Aldi Nord and Aldi Süd, roughly following administrative boundaries where possible, and merged it with the Southern half of the "de" region. The resulting polygon covers every Aldi Süd supermarket and is used as locationSet for Aldi Süd.

In order to avoid redundancy, I decided not to add an additional GeoJSON file for Aldi Nord. Instead, I chose to include "de" and exclude the region for Aldi Süd. This way, updating one Aldi region directly updates the other as well.


bhousel:
Thanks @kiddikai for taking on this!  It looks great 👍

PR #9275, opened 2024-02-26
Update hotelF1, delete Formule 1

panasonic-mafia:
`Formule 1` is a former name of `hotelF1`, see https://en.wikipedia.org/wiki/HotelF1.
I'm not sure why it exists as a separate record, most prob from OSM import...

### DIFF
--- a/data/brands/tourism/hotel.json
+++ b/data/brands/tourism/hotel.json
@@ -1200,17 +1200,6 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Yes, I expect the old name "_Formula 1_" is from older OSM data.

We can import data based on what is currently in OSM, but we don't delete it automatically, it has to be done manually, such as via a PR like this.

PR #9277, opened 2024-02-27
Update fuel.json - Add 'OK' in NL

Mitsjol:
Adding 'OK', a Dutch chain of petrol stations in the Netherlands.

This company has the same logo as the Danish and Swedish OK, but the three companies seem to be disconnected as far as I can tell.

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -3639,6 +3639,16 @@
       }
     },
     {
+      "displayName": "OK",
+      "locationSet": {"include": ["nl"]},
+      "tags": {
+        "amenity": "fuel",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/fuel.json
```diff
@@ -3638,6 +3638,16 @@
         "name": "OK"
       }
     },
+    {
+      "displayName": "OK",
```
Cj-Malone, line 3642: They would have to have unique display names, eg:
```suggestion
      "displayName": "OK (NL)",
```

Mitsjol: Thanks!

PR #9285, opened 2024-02-29
Deleted Camaieu

ckpiper:
This brand has been dissolved since 2022 according to the [wikidata](https://www.wikidata.org/wiki/Q2934647). Thus I believe it is fine to delete the entry.

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -1713,19 +1713,6 @@
       }
     },
     {
-      "displayName": "Camaïeu",
-      "id": "camaieu-1cb4e7",
-      "locationSet": {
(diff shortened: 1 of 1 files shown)


gosgtan:
Hi, actually it seems like the brand is coming back in September (see https://www.sudouest.fr/economie/conso-distribution/l-enseigne-camaieu-est-de-retour-rachetee-par-celio-et-soutenue-par-lena-situations-16653719.php or https://www.capital.fr/conso/a-quoi-va-ressembler-le-nouveau-camaieu-1479102). I think it's better to wait until then to see if it comes to fruition?

bhousel:
We can close this if the brand is maybe coming back. 
It's ok for dissolved brands to exist in the index for a while.

PR #9286, opened 2024-02-29
MEVO: change operator

starsep:


UKChris-osm:
Hi @starsep 

isn't the operator [Inurba Mobility](https://inurbamobility.com/en/about-us/)??

starsep:
I suppose both are correct. CityBike Global is value used by local mappers. News seem to confirm that

https://rowermevo.pl/aktualnosci
> Gdy hiszpański operator, firma City Bike Global,

[...] When Spanish operator, company City Bike Global

Mentioned few times more

But in the footer
> Samoobsługa MEVO jest obsługiwana przez Inurba Gdańsk

Self-service od MEVO is operated by Inurba Gdańsk

PR #9287, opened 2024-02-29
Removed usless Inmedio brand separation

docentYT:


Dimitar5555:
Thank you! I've also updated the capitalisation since they use lowercase letters on their websites and store fronts.

PR #9291, opened 2024-03-02
Add bakeries and remove Lila Bäcker

britiger:
Adding two bakery brands and remove Lila Bäcker which is no longer in business. 
- Some of the Lila Bäcker shops are now reopened as these already exiting smaller brands.


kjonosm:
Thanks @britiger !

PR #9303, opened 2024-03-10
Reduce the display name of CTS

ynerant:
The full display name "Compagnie des Transports Strasbourgeois" is a bit long and takes a lot of space in editors. This is however unneeded since there is no ambiguity with any other public transport operator. Moreover, a lot of operators have only the acronym and not the full name, so this is also a compliance commit.

### DIFF
--- a/data/operators/amenity/ticket_validator.json
+++ b/data/operators/amenity/ticket_validator.json
@@ -42,7 +42,7 @@
       }
     },
     {
-      "displayName": "CTS (Compagnie des Transports Strasbourgeois)",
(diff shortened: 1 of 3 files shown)


UKChris-osm:
It might be worth limiting the [locationSet](https://location-conflation.com/?locationSet=%7B%22include%22%3A%5B%5B7.75%2C48.58%5D%5D%7D&referrer=nsi) as well.

kjonosm:
I shortened the display name to "CTS (Strasbourg)" an limited the location set as suggested.

PR #9305, opened 2024-03-11
Updated the Parkshuttle opperator

wvanderp:
The operator of the ParkShuttle Rivium changed. this PR reflects that change
https://nl.wikipedia.org/wiki/ParkShuttle

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -11119,8 +11119,9 @@
       "locationSet": {"include": ["nl"]},
       "tags": {
         "network": "Parkshuttle Rivium",
-        "network:wikidata": "Q108721420",
-        "operator": "Connexxion",
+        "network:wikidata": "Q2160206",
+        "operator": "Transdev",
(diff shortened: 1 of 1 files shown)


data/transit/route/bus.json
```diff
@@ -11119,8 +11119,9 @@
       "locationSet": {"include": ["nl"]},
       "tags": {
         "network": "Parkshuttle Rivium",
-        "network:wikidata": "Q108721420",
-        "operator": "Connexxion",
+        "network:wikidata": "Q2160206",
+        "operator": "Transdev",
```
LaoshuBaby, line 11123: I noticed that the P137 (operator) of Q2160206 is still Connexxion (Q591484), not Transdev (Q442637/Q104185516). It is recommended that you check this PR and wikidata at the same time.

wvanderp: You are right I've change the wikidata and added a source

PR #9307, opened 2024-03-12
Rename PKP Energetyka to PGE Energetyka Kolejowa

starsep:


data/operators/power/line.json
```diff
@@ -4878,21 +4878,22 @@
       }
     },
     {
-      "displayName": "PJM",
-      "id": "pjm-b654b4",
-      "locationSet": {"include": ["us"]},
+      "displayName": "PGE Energetyka Kolejowa",
+      "id": "pgeenergetykakolejowa-116894",
+      "locationSet": {"include": ["001"]},
```
LaoshuBaby, line 4883:
```suggestion
      "locationSet": {"include": ["pl"]},
```
It looks like it only operates in Poland, not worldwide?

In addition, taginfo shows that there are still many elements are using "PKP Energetyka" as operator. It is recommended that you add a matchNames to ensure that the entry "PGE Energetyka Kolejowa" can be applied to all existing data.

starsep: Thanks. I limited to Poland and added `matchNames` in https://github.com/osmlab/name-suggestion-index/pull/9307/commits/4bef91e4e4ac28bcc43d78e2855b5770140d1f47

PR #9317, opened 2024-03-19
Add Rostics (ex. KFC in Russia)

panasonic-mafia:
I'm was not sure about adding Cyrillic tags `brand:ru=` and `name:ru=Ростикс'с` as on all signs in reality I see `Rostic's`.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -9735,6 +9735,22 @@
       }
     },
     {
+      "displayName": "Rostic's",
+      "locationSet": {"include": ["ru"]},
+      "matchNames": [
+        "ростикс",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Are there any Facebook accounts or Wiki Commons logos available to add to the [Wikidata](https://www.wikidata.org/wiki/Q3442874) for this entry?

panasonic-mafia:
> Are there any Facebook accounts or Wiki Commons logos available to add to the [Wikidata](https://www.wikidata.org/wiki/Q3442874) for this entry?

Added 👍

panasonic-mafia:
See no logos in Wiki Commons, and they don't have a Facebook.

PR #9337, opened 2024-03-27
Correct spelling of PENNY

j13m126:


bhousel:
Oh yeah, @Contrib1043 put it to uppercase in c4611d7a in Aug 2021.
Then this came up for discussion in #6662 in Jul 2022.
I feel like people can't really decide how they want the brands that are styled uppercase.  

I don't really have a preference, but it would be great if you could link to some discussion where people want this change. Mostly I don't want people flipping the names back and forth between Camel Case and Uppercase all the time.

Contrib1043:
The company itself consistently uses the branding PENNY in uppercase, even in their imprint on their website they list themselves 'PENNY Markt GmbH' (see https://www.penny.de/impressum). Also most major other map providers uses PENNY as well. So I don't see evidence, that 'Penny' should be the correct or even a common spelling. On the ground the stores are spelled PENNY.

UKChris-osm:
Personally I'd leave PENNY as it is for now, as looking at the Overpass link it seems almost all of the current POI's are marked up as PENNY, and would all need updating, unless a discussion is held regarding a bulk / automated edit to change them all.

kjonosm:
> I've always thought that we should only use caps if the brand is an acronym.

A bit off topic... But I generally wonder how to deal with acronyms that are commonly pronounced as a "normal" word as opposed to pronouncing each letter separately?

English Wikipedia seems to have the policy to draw the handling of using all-caps acronyms along that line. Only for brand or company names where each letter is pronounced separately Wikipedia uses titles in all-caps, e.g. IBM, BMW, KPMG, HSBC, DKNY. There are of course exceptions, like e.g. IKEA.

Following that policy within the subject of supermarkets, English examples would be to generally use `Asda` instead of `ASDA` vs.  or `Tesco` instead of `TESCO`. The German equivalent would be to use `Aldi` instead of `ALDI` or `Rewe` instead of `REWE`. `Penny` in that regard would also be used instead of `PENNY` not being an acronym whatsoever.

Does it make sense to implement a similar policy in OSM (or NSI)? Or should this issue be ignored by just loosening spell checks on capital letters within NSI?

Contrib1043:
At wikipedia they've got a bunch of rules on that topic (for reference https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Capital_letters#Acronyms and https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style/Trademarks)

[Quote]
Trademarks in CamelCase are a **judgment call** ; the style may be used where it reflects general usage and makes the trademark more readable; however, usage should be consistent throughout the article.
OxyContin or Oxycontin  – editor's choice
however: PlayStation only (camelcase preferred because Playstation is not widely used.)

[...]

Initial lowercase in certain trademarks almost never written any other way, such as iPhone and eBay, are **accepted** on Wikipedia.

[/Quote]

Personally I would suggest to judge each individual case (based primarily on what branding is used on the ground) rather than a general ruleset, which might not cover all cases anyway.

urbalazs:
Underlying `"brand:wikidata": "Q284688"` refers to [Penny](https://www.wikidata.org/wiki/Q284688) so this tag with brand/name=PENNY is confusing. Either change the name in Wikidata, Wikipedia or merge the pull request to have a consistent naming.

Additionally, the Hungarian subsidiary calls itself as "Penny-Market Kft."

I also vote for **Penny**.

Contrib1043:
When I look at their hungarian homepage, they seem to consistently use the uppercase spelling (same as in Germany...)

image

If it helps, I can adjust the wikidata record - it would then be consistent with the uppercase spelling of their parent company REWE on wikidata btw ;-)

Cj-Malone:
The Wikidata label has no influence and no requirement on matching NSI. They serve different purposes.

Has there been a discussion somewhere supporting this? I wouldn't want this to be merged just for other people to come back and complain next week.

bhousel:
I'm just going to merge this and accept the idea that we'll probably be flipping the names of German brands back and forth forever.

PR #9338, opened 2024-03-29
Add Inova health system

wcedmisten:


wcedmisten:
Fixed this to use `us` instead. I believe they have some in maryland as well, so I couldn't just use `va`

Github wouldn't let me edit the existing PR since I used a different branch, sorry!

https://github.com/osmlab/name-suggestion-index/pull/9340

PR #9340, opened 2024-03-30
Add Inova Health System

wcedmisten:


UKChris-osm:
Hi @wcedmisten 

If they also have some in Maryland, you're able to include multiple `.geojson` references within the `locationSet` , such as:

```
"locationSet": {
       "include":["us-md.geojson","us-va.geojson"]
},
```

wcedmisten:
@UKChris-osm thanks! Changed it to only use those states

UKChris-osm:
Thank you @wcedmisten 👍

PR #9341, opened 2024-03-30
Add banks in Ghana

sammyhawkrad:
I added some banks in Ghana and ran build


UKChris-osm:
@arch0345 I've just noticed the template change in this code actually does the opposite of what was intended, and it's just displaying the current branch, rather than the template master.

nsi-template-error

Issue #9347, opened 2024-04-03
"Bar Sport" template


Mannivu:
The suggestion "[Bar Sport](https://nsi.guide/index.html?t=brands&k=amenity&v=cafe&tt=bar%20s#barsport-93bf5b)" is quite weird: it has no Wikidata entity, it's global except for China, it doesn't seem related to any currently existent brand and, weirdly enough, I can't find it in the code. What's up with this suggestion?

Cj-Malone:
It exists in [NSI](https://github.com/osmlab/name-suggestion-index/blob/69ae16a0d9e914432fdc37c960ce17fb3532bd18/data/brands/amenity/cafe.json#L384-L397), but not in iD because it doesn't have wikidata.

Strangely it gets added back if I remove it and run `npm run build`...

Mannivu:
I couldn't find it because I was looking in /brands/bar.json.

UKChris-osm:
**Bar Sport** appears in OpenStreetMap twice, as shown in the NSI Collection Data:
["amenity/bar|Bar Sport": 143,](https://github.com/ideditor/nsi-collector/blob/a2a019ba6edf1b09e1be7ba35ccfad5d4c00145d/dist/osm/names_all.json#L1017)
["amenity/cafe|Bar Sport": 77,](https://github.com/ideditor/nsi-collector/blob/a2a019ba6edf1b09e1be7ba35ccfad5d4c00145d/dist/osm/names_all.json#L1098)

... and because the NSI will add entries that appear more than 50 times, it will reappear each time `npm run build` is run, unless it's specifically excluded.

Using Google Translate it suggests that _Bar Sport_ is simply a translation of _Sports Bar_.

Mannivu:
This is not a great idea in my opinion (but I understand why it is used). However, I've added a PR to exclude it from Italy (here is a common name, which was used where cafes, which are called bar in Italy, were used as a spot where people would gather after a sport event), not deleting it entirely.

Cj-Malone:
Chris has manually excluded it so it won't show up in any country. Thank you both.

Mannivu:
@UKChris-osm can we have the same for "[Bar Roma](https://nsi.guide/index.html?t=brands&k=amenity&v=cafe&tt=bar%20r)"?

PR #9348, opened 2024-04-03
Update cafe.json

Mannivu:
Removed "Bar Sport" from Italy (no such brand).


Cj-Malone:
Rel: #9347 there may be a better solution

UKChris-osm:
I've removed Bar Sport from the index as it doesn't seem to me a large brand anywhere.

PR #9357, opened 2024-04-12
RTC Thailand Update bus.json

OumKaZa:


OumKaZa (commented):
lasted update.

LaoshuBaby:
Agree with @arch0345, if you find that locals are more familiar with using short names, you can use `"network":"RTC"` and use `Regional Transit Corporation` as matchNames

bhousel:
Oh yeah, I agree .. Unless there is a discussion somewhere that the community really wants to change the tagging of all their bus routes, we should leave this alone.

PR #9364, opened 2024-04-15
Remove `name` from Desjardins

wolfy1339:
Remove the `name` from Desjardins, as it is a federation of credit unions and not a single entity.


kjonosm:
Thanks @wolfy1339! I changed your pull request to still keep the name tag. The additional preserveTags=name prevents iD from suggesting existing name updates.

PR #9369, opened 2024-04-17
extend list of "dpd" parcel_locker countries

mnalis:
according to www.dpd.com and main linked [wikipedia entry](https://en.wikipedia.org/wiki/Geopost), where brand is called `DPD`


kjonosm:
Thanks @mnalis

PR #9379, opened 2024-04-20
Update industrial.json

gander:
Add meat processing plant Lenarcik in Poland

### DIFF
--- a/data/operators/landuse/industrial.json
+++ b/data/operators/landuse/industrial.json
@@ -17,6 +17,19 @@
       }
     },
     {
+      "displayName": "Lenarcik",
+      "id": "lenarcik-9ff3f0",
+      "locationSet": {"include": ["pl"]},
+      "tags": {
+        "brand": "Lenarcik",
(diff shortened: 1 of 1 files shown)


data/operators/landuse/industrial.json
```diff
@@ -16,6 +16,19 @@
         "operator:wikidata": "Q2265496"
       }
     },
+    {
+      "displayName": "Lenarcik",
+      "id": "lenarcik-9ff3f0",
+      "locationSet": {"include": ["pl"]},
+      "tags": {
+        "brand": "Lenarcik",
```
LaoshuBaby, line 24: Maybe brand is not necessary here?

gander: shop (one of 54): https://www.openstreetmap.org/way/866692988
plant: https://www.openstreetmap.org/way/866693458

product with logo: https://zmlenarcik.pl/wp-content/uploads/2019/10/smalec_dworski.jpg

I think it is brand, but not all shops are named the same, so then also operator

PR #9391, opened 2024-04-22
Split Heytea English entry

Cj-Malone:
I think the English seeking countries need a separate entry, @LaoshuBaby do you have any opinions?


LaoshuBaby:
Sorry for being late. I only focused on the country list when merging it. You did a correct job!

The suggestion made by arch0345 that using English in Singapore is also very reasonable.

Cj-Malone:
Thanks both

PR #9394, opened 2024-04-23
Update Lenarcik shop brand and operator

gander:


LaoshuBaby:
Thank you for explaining the relationship between the two more accurately and answering the questions in #9379 through patches. 😆

gander:
This was my first contribution to wikidata and NSI, and only when I saw the final result did I understand how I should have prepared this data for the index.

PR #9398, opened 2024-04-25
Fix bad locationSet

Nydragon:
Fixed bad locationSet value of "kr", it actually was meant to be "ko"


Nydragon:
I am confused now, "kr" is not a valid location, but OSM outputs the following data when I associate the place with the brand:

image

I don't know why it says language: Kanuri

Nydragon:
Kanuri = name:kr
Korean = name:ko

1ec5:
`kr` is the ISO&nbsp;3166 code for South Korea (**K**orean **R**epublic). `ko` is the ISO&nbsp;639 code for the Korean language.

LaoshuBaby:
You may noticed some place are using a locale called ko-KR, the ko means Korean language, and KR means South Korea the country. In locationSet there should be some place's country or region code, a geojson name, or a position with redius. 

You can see more details in https://github.com/osmlab/name-suggestion-index/wiki/Contributing#locationset-required

I truly found a `name:kr` but this looks like this is typo in entry, it should be `name:ko`

图片

and this was fixed in https://github.com/osmlab/name-suggestion-index/commit/8d3fe697352233ece75cd2fd82b123222da6f010 2 days ago, you can see wrong preset is because this repository build dist weekly (usually monday in UTC+9)

Therefore I'll close this PR, but still thank you for notice this strange tag.

Nydragon:
Alright, thanks for the clarification, so this issue should be deployed next Monday, right?

LaoshuBaby:
> so this issue should be deployed next Monday, right?

图片

Generally speaking, yes, but this is new version release date, there also need some time to fresh cache on CDN.

PR #9399, opened 2024-04-25
Add some Dutch parcel locker brands

RobinLinde:
Adds PostNL, localized version of DHL and de Buren.


bhousel:
Thanks @RobinLinde 👍

PR #9402, opened 2024-04-27
Add french universities

ErnestoLaPomme:


bhousel:
Sorry @ErnestoLaPomme , but we don't add individual colleges or universities to this index - we only add the kind of for-profit "universities” that operate like a branded business in multiple locations, such as DeVry or University of Phoenix. I know this is an unusual concept outside of the USA - the “universities” in the name-suggestion-index are really more like training centers than true universities.

PR #9406, opened 2024-04-28
Update clothes.json clothes=women for Millers

thetornado76:
Correction from clothes=woman to clothes=women for Millers


Cj-Malone:
Thank you

PR #9420, opened 2024-04-30
Added Italian clothing store chain Calliope

DnaX:
Calliope is a fashion label with various stores in Europe and Middle East.

See Wikidata item for more details: https://www.wikidata.org/wiki/Q125703464


kjonosm:
Thanks!

PR #9429, opened 2024-05-02
Add chocoZAP

tom-konda:
Add chocoZAP fitness center in Japan.


bhousel:
LGTM, thanks & welcome @tom-konda  👍

PR #9431, opened 2024-05-03
Update insurance.json - "Wüstenrot" also present in Hungary, Slovenia, Croatia.

mnalis:
see https://en.wikipedia.org/wiki/W%C3%BCstenrot-Gruppe


Cj-Malone:
Thank you

PR #9432, opened 2024-05-03
Add missing (and update existing) banks in Croatia

mnalis:
as officially listed at: 
https://www.hnb.hr/temeljne-funkcije/supervizija/popis-kreditnih-institucija

### DIFF
--- a/data/brands/amenity/payment_centre.json
+++ b/data/brands/amenity/payment_centre.json
@@ -81,7 +81,6 @@
         "amenity": "payment_centre",
         "bureau_de_change": "yes",
         "money_transfer": "western_union",
-        "office": "financial",
         "official_name": "Financijska agencija"
       }
     },
--- a/data/brands/amenity/payment_centre.json
+++ b/data/brands/amenity/payment_centre.json
@@ -72,6 +72,20 @@
       }
     },
     {
+      "displayName": "Fina - Financijska agencija",
+      "locationSet": {"include": ["hr"]},
+      "tags": {
+        "brand": "Financijska agencija",
+        "brand:wikidata": "Q12631015",
+        "name": "Fina",
+        "amenity": "payment_centre",
+        "bureau_de_change": "yes",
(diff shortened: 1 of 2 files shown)


data/brands/amenity/payment_centre.json
```diff
@@ -71,6 +71,20 @@
...
+      "displayName": "Fina - Financijska agencija",
+      "locationSet": {"include": ["hr"]},
+      "tags": {
+        "brand": "Financijska agencija",
+        "brand:wikidata": "Q12631015",
+        "name": "Fina",
+        "amenity": "payment_centre",
+        "bureau_de_change": "yes",
+        "money_transfer": "western_union",
+        "office": "financial",
```
Cj-Malone, line 84: I'm not fond of things getting multiple top level tags, in this case `amenity` and `office`. Are you sure this is correct tagging?

mnalis: Yes, it seems most precise... While most citizens will use them for paying bills (thus `amenity=payment_centre`), they do offer more advanced financial services to SMEs too. 

But if you think it is problematic, I can remove `office=financial`?

Cj-Malone: I would not have added `office=financial`, however it's not a hard rule, NSI does have entries with 2 categories, eg FedEx and Chuck e cheese.

I'll leave it for other people to provide opinions.

kjonosm: I would also be in favour of removing `office=financial`. Besides, OSM wiki [claims](https://wiki.openstreetmap.org/wiki/Tag:office%3Dfinancial_advisor#See_also) `office=financial` should be used for offices not open to the general public.

mnalis: Thanks for feedback, I've removed  `office=financial` now.

Cj-Malone:
Thank you, it can be restored of changed at anytime if people have different opinions.

PR #9433, opened 2024-05-03
Update README.md - add SCEE app

mnalis:
Add "SCEE" in "Currently used in" section


Cj-Malone (approved):
Makes sense to me, but should these be linking to code? 

`https://streetcomplete.app/` / `https://github.com/Helium314/SCEE` would probably be better.

mnalis:
> https://streetcomplete.app/ / https://github.com/Helium314/SCEE would probably be better.

Sure, makes sense to me too. Updated.

Cj-Malone:
Thanks

PR #9435, opened 2024-05-03
Update car_parts.json - "Tokić" add wikidata, country

mnalis:
according to https://www.tokic.hr/o-nama/ also present in Slovenia


Cj-Malone:
Thank you

PR #9441, opened 2024-05-05
Add Wolf, German bakery chain

Robot8A:


kjonosm:
Thanks for the pull request! There are several different bakery chains with that name in Germany. Could you limit the location to the actual area this chain operates in?

PR #9443, opened 2024-05-05
Added more Israel chains

Flatroy:


bhousel:
This seems ok, but:
- Items without a Wikidata identifier won't get included in any of the editor presets.
- A few items that seem mistagged, (e.g. Golf & Co. https://www.golfco.co.il/ seems to be more like houseware, not clothes)

But I'm happy to accept it as-is and people can change the categories later if they want to.

PR #9455, opened 2024-05-09
Update restaurant.json - Added BBQ Plaza in Thailand

ckpiper:
Now sure what to put for cuisine as about us page says food is grilled in a Mongolian and Japanase style. And not sure what was the decision from this https://github.com/osmlab/name-suggestion-index/pull/8884 conversation.

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -55,6 +55,7 @@
         "amenity": "restaurant",
         "brand": "Bar-B-Q Plaza",
         "brand:wikidata": "Q125861684",
+        "cuisine": "barbecue;mongolian;japanese"
         "name": "Bar-B-Q Plaza"
       }
     },
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -49,14 +49,19 @@
       "displayName": "Bar-B-Q Plaza",
       "locationSet": {"include": ["th"]},
       "matchNames": [
-        "BBQ Plaza"
+        "BBQ Plaza",
+        "Bar.B.Q Plaza"
       ],
       "tags": {
         "amenity": "restaurant",
         "brand": "Bar-B-Q Plaza",
+        "brand:en": "Bar-B-Q Plaza",
+        "brand:th": "บาร์บีคิวพลาซ่า",
         "brand:wikidata": "Q125861684",
         "cuisine": "barbecue;mongolian;japanese",
-        "name": "Bar-B-Q Plaza"
+        "name": "Bar-B-Q Plaza",
+        "name:en": "Bar-B-Q Plaza",
+        "name:th": "บาร์บีคิวพลาซ่า"
       }
     },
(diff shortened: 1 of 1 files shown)


1ec5:
The outcome of #6791 was to tag `cuisine=mongolian_grill` (though it should’ve been `mongolian_barbecue`). On the other hand, if this chain serves [real Mongolian-style barbecue like they eat in Mongolia](https://en.wikipedia.org/wiki/Khorkhog), then `cuisine=barbecue;mongolian` would be totally appropriate for this chain.

As for the Japanese part, you could append `;japanese`, but it isn’t clear whether this is a reference to _yakiniku_ or _teppanyaki_, which has a separate `cuisine=teppanyaki` tag already.

By the way, you’ll need to add some more statements to https://www.wikidata.org/wiki/Q125861684 to prevent it from getting deleted as meaningless spam. See [this guide on the wiki](https://wiki.openstreetmap.org/wiki/Key:brand:wikidata#Creating_a_Wikidata_item) for some tips.

data/brands/amenity/restaurant.json
```diff
@@ -45,6 +45,25 @@
     }
   },
   "items": [
+    {
+      "displayName": "Bar-B-Q Plaza",
+      "locationSet": {"include": ["th"]},
+      "matchNames": [
+        "BBQ Plaza",
+        "Bar.B.Q Plaza"
```
1ec5, line 53: NSI already recognizes these two spelling variants, because it strips punctuation and lowercases before comparing names.

data/brands/amenity/restaurant.json
```diff
@@ -45,6 +45,25 @@
...
+      ],
+      "tags": {
+        "amenity": "restaurant",
+        "brand": "Bar-B-Q Plaza",
```
1ec5, line 57: Just double-checking: does the Thailand OSM community expect the primary name to be in English or Thai? From news coverage, it seems like at least some locations primarily signpost the English name, but I couldn’t tell if that’s representative.

data/brands/amenity/restaurant.json
ckpiper: Understood how Bar.B.Q Plaza will already be recognized without punctuation. But How is BBQ Plaza recognized? It would need to strip the "ar" in "Bar" as well.

data/brands/amenity/restaurant.json
ckpiper: I was looking at locations in google street view. Most have the Engilsh version largely displayed, but the Thai version is also displayed as a header note or foot note above/below/next to the English. Thus I thought I should add both.

1ec5: Yes, I was just wondering which should be in the main `name` key. Sounds like it should be the English name.

1ec5 (approved):

Cj-Malone:
Thank you. If anyone from the Thailand community has an opinion, we will be happy to change it.

PR #9485, opened 2024-05-15
Update outdoor.json - added Can-AM

jhutton1:


bhousel:
Going to close this, as I can't tell if they actually have any stores.  
It looks like a brand that sells their products through dealers.. 
If I'm wrong and there are actual "Can-AM" retail locations, let me know and I'll reopen and merge.

(this is like Caterpillar #5647 and Polaris #5233)

PR #9492, opened 2024-05-16
Update supermarket.json - add match name for Big C

JanCallewaert-TomTom:


JanCallewaert-TomTom:
Closing since this isn't how match names work. Spaces get removed from match names, making the match name in this case the same as the brand name.

PR #9498, opened 2024-05-17
Update charging_station.json - Added QCode for reveo

ckpiper:


UKChris-osm:
Did you mean to replace `fx` with `fr`?

ckpiper:
I was not sure if fx was code for something that I was unaware of. Thus I left it.

UKChris-osm:
The Country Code `fx` refers to [mainland France](https://location-conflation.com/?locationSet=%7B%22include%22%3A%5B%22fx%22%5D%7D&referrer=nsi) only, whereas `fr` includes [mainland France and other French territories](https://location-conflation.com/?locationSet=%7B%22include%22%3A%5B%22fr%22%5D%7D&referrer=nsi). If `Reveo` only operate within mainland France, you can keep it as simply `fx`, if not, you can remove `fx` and use `fr` instead.

PR #9499, opened 2024-05-19
Update Omniva parcel lockers

markalex2209:
Omniva operates in all three Baltic states. Added Latvia and Lithuania to list of available locations.

Also, since all parcel lockers support both sending and receiving packages, added tags for it.


Cj-Malone:
Thank you

markalex2209:
No, thank you!)

PR #9500, opened 2024-05-19
Add Latvijas Pasts to parcel_locker.json

markalex2209:
Latvijas Pasts has something like 250 parcel lockers around Latvia: https://pasts.lv/lv/kategorija/pasta_nodalas/?type=pakomati

Thus, I believe it qualifies for addition to this list.


Cj-Malone:
Thank you

PR #9507, opened 2024-05-21
Add apogia hotels

ckpiper:


bhousel:
marketing assn ? all hotels have different names

PR #9513, opened 2024-05-21
Update payment_centre.json Added PayPoint

jhutton1:


Cj-Malone:
I don't think this is a right, shops offer this service but it's not a standalone thing. Right @UKChris-osm ?

jhutton1:
Do you suggest another category?

UKChris-osm:
I don't think PayPoint have their own payment centres / shops, and instead is a service offered by existing shops - but being able to add PayPoint to OSM seems like a good idea to me.

But, should it be via an NSI entry like this, or perhaps using the `payment` key added to existing shops, such as `payment:paypoint = yes`?

Cj-Malone:
Taginfo suggests `pay_point=yes` is the most popular with `paypoint=yes` behind it. I think this belongs in tagging presets, not NSI.

PR #9518, opened 2024-05-21
Update fuel.json - Added Enilive

ckpiper:


Cj-Malone:
@DnaX Please could you look at this? I know you did some Eni stuff recently.

DnaX:
Enilive, as I know, it's a greenwashing strategy of Eni. It's both an app for services available in the gas stops of Eni and umbrella company for some hydrogen raffinerie. I was looking for actual next-gen fuel gas stops but I didn't find any.

PR #9519, opened 2024-05-22
Update supermarket.json

Mannivu:
Add it to Carrefour Contact, see https://www.fruitbookmagazine.it/carrefour-contact-arriva-anche-in-italia-il-format-che-compete-con-il-discount/


Cj-Malone:
Thank you

PR #9523, opened 2024-05-22
Update car_parts.json - Added Magenti Marelli

ckpiper:


bhousel:
The brand makes parts but they don't operate stores..

PR #9526, opened 2024-05-22
Update hotel.json - Added Star Hotels

ckpiper:


bhousel:
This one too.. it's an operator maybe, but the hotels all have different names.

PR #9528, opened 2024-05-22
Update hotel.json - Added UNAHOTELS

ckpiper:


ckpiper:
@bhousel I reponed this and fixed the wrong qcode. Also here is the 
https://www.gruppouna.it/unahotels/hotels-resorts where they are named with UNAHOTELS <City>

PR #9536, opened 2024-05-23
Add romantik hotels

ckpiper:


Cj-Malone:
I think I'd have just done "Romantik" with a match name for "Romantik Hotels & restaurants", but I'll trust your opinion if if you've already considered this?

PR #9537, opened 2024-05-23
Update bank.json - Added Sparkasse

ckpiper:


Cj-Malone:
I think this one is a little more complex, I'd prefer it if a German checked it out.

bhousel:
Oh yeah there is some history on #3583, #2475 and the other issues that link there..

IIRC, Sparkasse is more like a generic word that describes the type of the bank (for example, like "Co-op"), not a specific name for a bank.  

Each region in Germany has their own Sparkasse banks in that region, and we've considered adding these sub-entities in the past.  NSI _does_ support geofencing local regions now (it didn't before) so this might be a thing to support eventually, if the German community wants it.

PR #9538, opened 2024-05-23
Update hotel.json - Added TOP International Hotels

ckpiper:


Cj-Malone:
I'm not convinced this is real, vs SEO trash

bhousel:
yeah looking more this seems like LHW on #9527 
The hotels aren't named "TOP"

PR #9541, opened 2024-05-24
Nordea is gone from Poland

matkoniecz:
I do not remember seeing it recently and per https://en.wikipedia.org/wiki/Nordea_Bank_Polska "Nordea Bank Polska ceased to exist as an independent company in 2014 after a takeover and legal merger with PKO BP. "

https://en.wikipedia.org/wiki/Nordea confirms "Nordea divested its Polish banking operations in 2013, with the sale to PKO Bank Polski for €694 million but retains a presence in Poland via operations and IT units supporting the Nordic banks"


matkoniecz:
see also https://www.openstreetmap.org/search?query=Nordea%2C%20poland (zero results)

Cj-Malone:
Thank you

PR #9545, opened 2024-05-24
Add Forevers from Latvia into butcher.json

markalex2209:
Butcher shop Forevers from Latvia has already 43 shops mapped on the OSM (and based on data from their site total of more than a hundred)


Cj-Malone:
Thank you

PR #9550, opened 2024-05-25
Add latvian bank Citadele bank.json

markalex2209:
Add Latvian Citadele bank.

It has relatively small number of branches, but [a lot of atms](https://overpass-turbo.eu/s/1M0S).


bhousel:
thanks @markalex2209 👍

PR #9552, opened 2024-05-25
Split baltic supermarket Rimi into 4 items

markalex2209:
Rimi has four types of stores: Rimi Hyper, Rimi Super, Rimi Mini, and Rimi Express. They operate under same brand, but differ in size.

This leads to mappers erroneously agreeing for suggestion to "correct tags" by editor. For example, `name=Rimi Mini` is suggested to be changed to 
```
name=Rimi
branch=Mini
```
which is against recommendations for use of tag [`branch`](https://wiki.openstreetmap.org/wiki/Key:branch). Misuse can be seen [here](https://overpass-turbo.eu/s/1M1U).

---
A couple additional notes regarding changes: 
- Express fits description of `shop=convenience`, so I extracted it there, but added `"matchTags": ["shop/supermarket"]` to ease fixes,
- For some unknown to me reason Lithuania's branch is missing Rimi Mini kind of shops (can be seen [here](https://www.rimi.lt/parduotuves), as opposed to [ee](https://www.rimi.ee/kauplused) and [lv](https://www.rimi.lv/veikali) ones), so I excluded it from `locationSet` of that item.

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -3928,6 +3928,19 @@
       }
     },
     {
+      "displayName": "Rimi Express",
+      "locationSet": {
+        "include": ["ee", "lt", "lv"]
+      },
+      "matchTags": ["shop/supermarket"],
+      "tags": {
+        "brand": "Rimi",
+        "brand:wikidata": "Q3741108",
+        "name": "Rimi Express",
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks 👍

PR #9553, opened 2024-05-25
Latvian grocery stores

markalex2209:
Added Latvian grocery store networks: top!, Aibe, Elvi, Beta and Vesco. All have more than 30 locations open.

Highlights of added:
- spelling of `top!` might be a bit controversial, but it was agreed on in local community,
- Latvian Aibe is a "daughter" of Lithuanian Aibė, but due distict spelling, it's in a separate item,
- Elvi exists in two separate forms: convenience store and supermarket. It operates under same brand name, and cannot be organically pushed into a singe type, thus two items added.

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -4920,6 +4920,15 @@
       }
     },
     {
+      "displayName": "Vesko",
+      "locationSet": {"include": ["lv"]},
+      "tags": {
+        "brand": "Vesko",
+        "name": "Vesko",
+        "shop": "convenience"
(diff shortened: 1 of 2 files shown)


bhousel:
Thanks this all seems ok.. 
Note that the items without `brand:wikidata` won't be converted to presets until that tag is added later.

markalex2209:
@bhousel, Thank you very much for explanation, I was not aware of that.
Can you, please, consider adding this prerequisite to the wiki or readme (unless, of course, I have missed this information being already there)?

PR #9556, opened 2024-05-28
Fix romantik hotel name

ckpiper:
ref: https://github.com/osmlab/name-suggestion-index/pull/9536
I think the brand itself should be Romantik Hotels & Restaurants as that is what the Website and Wikidata page has. On the website, "Romantik" is in larger font than "Hotels & Restaurants", but I think it is not a big enough difference that Hotels & Restaurants should be taken off completely.

Also should the name be Romantik Hotel and Romantik Restaurant? The [website](https://www.romantikhotels.com/en/hotels/#/map/) has Hotels named as `Romantik Hotel <City>`. Thus not sure what to do.

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -5344,7 +5344,7 @@
       "displayName": "Romantik",
       "id": "romantikhotelsandrestaurants-2fabb0",
       "locationSet": {"include": ["001"]},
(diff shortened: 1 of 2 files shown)


bhousel:
Apple Maps does this:

Screenshot 2024-05-28 at 2 41 55 PM


So I agree with these changes..
"Romantik Hotel <city>" sounds like a reasonable thing to do.

PR #9561, opened 2024-05-29
Move "CoCo壱番屋" to restaurant from fast_food; Set cuisine.

nabetaro:
In Japan, "CoCo ICHIBANYA" is considered a restaurant in OSM terms, since in most cases there are seats, waiter service, and the serving is not very fast.

see: https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#is-an-amenity-a-restaurant-or-fast_food

and specify cuisine to (Japanese) curry.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -1805,26 +1805,6 @@
       }
     },
     {
-      "displayName": "CoCo壱番屋",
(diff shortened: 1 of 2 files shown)


bhousel:
Looks good for me, thanks and welcome @nabetaro  👋

PR #9563, opened 2024-05-29
Update fuel.json Add Avia XPress

jhutton1:


bhousel:
Not sure if this really needs to be different from the existing [Avia item](https://nsi.guide/index.html?t=brands&k=amenity&v=fuel#avia-e0aaf8), but I'll accept it anyway..

PR #9564, opened 2024-05-29
Update car_rental.json - Added Cicar

ckpiper:
A lot of the shops have signs that say "Cabrera Medina" as well as "CICAR", which is why it is in matchNames. In 1989 CICAR joined the Cabrera Medina group, to expand into other Canary Islands.

### DIFF
--- a/data/brands/amenity/car_rental.json
+++ b/data/brands/amenity/car_rental.json
@@ -7,6 +7,17 @@
   },
   "items": [
     {
+      "displayName": "CICAR",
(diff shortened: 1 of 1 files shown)


kjonosm:
Should the location be limited to the Canary Islands? See https://ideditor.codes/?code=Q5813&map=5.23/28.7159/-18.6787

ckpiper:
@kjonosm It seems the bulk of its offices are in Canary Islands, but it also has some in Balearic Islands and Mainland Spain.

PR #9575, opened 2024-05-29
Extract DPD Latvia in parcel_locker.json

markalex2209:
Parcel lockers of DPD are not usually called DPD Pickup Station in Latvia.

Extracted instance for Latvia into a separate item, and since it's Latvia specific now, added DPD Latvia as operator.

Also @richlv confirmed at local office that all their parcel lockers are accepting packages, thus added `parcel_mail_in` tag.

### DIFF
--- a/data/brands/amenity/parcel_locker.json
+++ b/data/brands/amenity/parcel_locker.json
@@ -266,6 +266,19 @@
       }
     },
     {
+      "displayName": "DPD",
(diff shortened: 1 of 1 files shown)


bhousel:
thanks, this seems ok 👍

richlv:
Yay, thank you guys :)

PR #9594, opened 2024-05-30
Update convenience.json Add Press Shop

jhutton1:


kjonosm:
Would `shop=newsagent` or `shop=book` be a better fit for this chain?

jhutton1:
moved to shop=newsagent

PR #9605, opened 2024-06-04
[shop/car_repair] add Chrysler

panasonic-mafia:


kjonosm:
The Wikidata QID should reflect the brand, not the parent holding company. I would recommend to change the wikidata back to `Q29610` in both `shop=car` and `shop=car_repair`.

panasonic-mafia:
Thanks, makes sense. I had doubts after seeing [Chrysler dealers with the Stelantis q-code in OSM](https://taginfo.openstreetmap.org/search?q=brand%3Awikidata%3DQ181114#tags), but hopefully someone could change it.

kjonosm:
Thanks @panasonic-mafia!

PR #9608, opened 2024-06-04
Add Agricultural Bank of Taiwan

jhutton1:


Dimitar5555:
There is no need to duplicate the entry in two files. One entry in `bank.json` is enough, NSI will automatically create a copy for the ATMs.

LaoshuBaby:
@Dimitar5555  @jhutton1 I don't think this PR can be approved ^

This bank don't named "台湾农业银行" (and those translation are using Simplified character), it should be 全國農業金庫

You may confused it with 中国农业银行（Agricultural Bank of China）

https://zh.wikipedia.org/wiki/File:ROC_Agricultural_Bank_of_Taiwan_logo.svg

And I don't think it need to be added, because it only have 9 branch according to its website https://www.agribank.com.tw/Page?itemid=7&mid=9_

图片

----

I'm curious as to why you would notice this bank, as banks with influence in Taiwan have already been basically added by local mappers in https://github.com/osmlab/name-suggestion-index/pull/5268.

Cj-Malone:
Does the website say how many ATMs they have?

LaoshuBaby:
> Does the website say how many ATMs they have?

They haven't announced it, but as far as I know, ATMs will only be set up in places with branches, so the number also will not exceed 10

Dimitar5555:
According to Wikipedia they have 26 billion NTD (~800 million USD) in capital. Seems quite a lot for a bank with only 10 branches in a single country. Do we know if each branch has multiple locations/representative offices in each city?

LaoshuBaby:
> Seems quite a lot for a bank with only 10 branches in a single country.

https://t.me/OSMTaiwan/15011

> 图片
>
> 全國農業金庫國內分行管理辦法
> 
> 第3條
> 農業金庫每年申請增設分行，不得超過二處，其分行總數不得超過十處。但為配合農業金融監理政策協助處理經營不善之農、漁會信用部，不在此限。
> 
> -----
> 
> Article 3
> The National Agricultural Bank shall not apply to establish more than two branches per year, and the total number of branches shall not exceed ten. However, this does not apply when it is necessary to coordinate with the agricultural financial supervision policy to assist in dealing with poorly managed agricultural and fishing association credit departments.

That is to say, under normal circumstances, the number of branches of this bank has already been stipulated by legal provisions.

PR #9626, opened 2024-06-05
add PCC

keisisqrl:
PCC is a local supermarketmarket chain in Washington State:

https://www.pccmarkets.com

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -5933,6 +5933,18 @@
       }
     },
     {
+      "displayName": "PCC Community Markets",
+      "locationSet": {"include": ["us-wa.geojson"]},
+      "matchNames": ["PCC"],
+      "tags": {
+        "alt_name": "PCC",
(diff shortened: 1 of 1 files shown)


ckpiper:
Hello. From the website it looks like it only has 15 locations. https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case that is pretty far under the threshold for adding a brand to NSI.

Cj-Malone:
Yeah, this is a little low, but it has a Wikipedia page so it's Wikidata page isn't going to be deleted. Given that this is @keisisqrl first time, and it's not gonna get deleted, let's make an exception.

:tada: Welcome!

bhousel:
> Yeah, this is a little low, but it has a Wikipedia page so it's Wikidata page isn't going to be deleted.

I agree -maybe we should add this to our guidelines.  It's less about there being a specific number of branches to qualify, and more about whether the brand is notable enough to have a presence on Wikidata / Wikipedia.

PR #9634, opened 2024-06-08
Removed duplicate preset for subway stations

serhii-muchychka:
The other is in the file:
https://github.com/osmlab/name-suggestion-index/blob/main/data/transit/public_transport/station_subway.json
https://nsi.guide/index.html?t=transit&k=public_transport&v=station_subway

### DIFF
--- a/data/transit/railway/station.json
+++ b/data/transit/railway/station.json
@@ -7,16 +7,6 @@
   },
   "items": [
     {
-      "note": "Auto create railway/station for all route/subway too - #9320",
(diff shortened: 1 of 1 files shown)


bhousel:
oh good catch - I didn't realize these were in 2 places 🤔  
It's not ideal..  

I'll try to check how our presets match up with the presets from [id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema) .   Each of our files really should inherit from a preset that's already existing there.

PR #9643, opened 2024-06-10
[shop/car][shop/car_repar] add Bentley, Bugatti, Dacia, Ferrari

panasonic-mafia:


bhousel:
Hey @panasonic-mafia you've added a bunch of these pull requests in the past day or so - is the intention to just have a "car_repair" item for every "car dealer" item ?   We can accomplish that with a single template entry.

I agree that this is a pretty normal way to represent the separate dealers and service center - they often have slightly separate locations and different contact info / opening hours.

Here's an example of how they typically appear on Google Maps:

Screenshot 2024-06-10 at 11 34 58 AM

panasonic-mafia:
@bhousel, exactly, that would be great to have it as a template. 
I don't have experience with templates, but I'll try to add one within this PR.

panasonic-mafia:
@bhousel I added a template, may I kindly ask you to have a look on it when you have time please? I'm a bit not fluent in the codebase, so I might miss something.

bhousel:
> @bhousel I added a template, may I kindly ask you to have a look on it when you have time please? I'm a bit not fluent in the codebase, so I might miss something.

Thanks!  looks good to me 👍

PR #9655, opened 2024-06-16
fix: add package-lock.json

Venefilyn:
Package lock should exist in the project given that we use `npm` and
have commands set up. This makes it possible to ensure that the packages
we use are always the same and that any issues that occur will not be a
total surprise, in other words, we will get a better handle on
versioning

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@
 .DS_Store
 .cache
 npm-debug.log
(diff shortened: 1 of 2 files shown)


bhousel:
Yea this is ok 👍 
A few years ago this used to cause some problems, but not anymore.

PR #9656, opened 2024-06-16
fix: IDs and sorting

Venefilyn:
In #9652 I had accidentally entered a self-generated ID that was not a
base64 value. Hence it contained invalid characters. This PR fixes that
with running `npm run build` and sorts the wholesale shops JSON as well


Cj-Malone:
The build would have automatically fixed this, but thank you

PR #9658, opened 2024-06-17
Update DHL Poststation wikidata identifier

nicolashohm:
`Q1766703` Refers to "DHL Packstation" even though there is an actual wikidata identifier for "DHL Poststation".

A side effect of reusing the same wikidata identifier for both entities is that iD suggests renaming Poststations to Packstations.


nicolashohm:
let me know if you need more information about this PR to merge

PR #9669, opened 2024-06-20
Add Canadian Tire auto centres

### LINKED ISSUE #9667: Canadian Tire Auto Centres
Add an entry for Canadian Tire Auto Centres.
They are a part of the Canadian Tire stores, they can probably re-use the same wikidata entry

```
shop=car_repair
name:en=Canadian Tire Auto Centre
name:fr=Centre de l'auto Canadian Tire
```
This will require 2 entries, one for Quebec, one for the rest of Canada.

wolfy1339:
Fixes #9667

### DIFF
--- a/data/brands/shop/car_repair.json
+++ b/data/brands/shop/car_repair.json
@@ -267,6 +267,20 @@
       }
     },
     {
+      "displayName": "Canadian Tire Auto Centre",
+      "id": "canadiantire-8cbc80",
+      "locationSet": {
+        "include": ["ca"],
+        "exclude": ["ca-qc.geojson"]
+      },
+      "tags": {
+        "brand": "Canadian Tire",
+        "brand:wikidata": "Q106224674",
+        "name": "Canadian Tire",
+        "shop": "car_repair"
+      }
+    },
+    {
       "displayName": "Car-X",
       "id": "carx-878d90",
       "locationSet": {"include": ["us"]},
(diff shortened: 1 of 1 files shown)


Cj-Malone (changes requested):
Brand and name tags need to be localised as well, not just displayName

wolfy1339:
That is the brand name, there is no difference to the brand name in french/english.
Here is a streetview of a storefront in Quebec: https://www.bing.com/maps?toWww=1&redig=EAF66A16DECD47479ED6AA7D73E132E8&cp=46.104084%7E-71.312143&lvl=19.6&style=x&pi=-5.2&dir=137.2

The entry for the store is correct as it is, and
https://github.com/osmlab/name-suggestion-index/blob/081176f1d733ac436a7543f1c802462e4f64e47a/data/brands/shop/department_store.json#L223-L234

wolfy1339:
I'd also like to point out that they also have gas stations, that entry is not localized
https://github.com/osmlab/name-suggestion-index/blob/081176f1d733ac436a7543f1c802462e4f64e47a/data/brands/amenity/fuel.json#L910-L924

Streetview: https://www.bing.com/maps?toWww=1&redig=EAF66A16DECD47479ED6AA7D73E132E8&cp=46.817863%7E-71.330604&lvl=19.8&style=x&pi=2.2&dir=238.7

This is another issue, really, that is separate from this PR

wolfy1339:
@Cj-Malone I believe this should be fixed to your liking

PR #9673, opened 2024-06-23
Add àVélo bike share system in Quebec City

wolfy1339:


UKChris-osm:
Thank you @wolfy1339 

I have reduced the **locationSet** tag to just the area of Quebec City as the bikes only seem to be available in a very small area.

PR #9689, opened 2024-07-01
Add Holiday Oil (Utah Chain)

dknelson9876:
Needed as a disambiguation between Holiday (Q5880490) and Holiday Oil (Q125941013)


Cj-Malone:
Thank you

PR #9695, opened 2024-07-04
[PL] Add Pocztex

ireun:


Cj-Malone (changes requested):
We don't do operator:wikipedia anymore, nor am I confident that they all have the same opening hours.

Also if going with operator wikidata, not brand wikidata, this should be in the operator tree.

ireun:
> We don't do operator:wikipedia anymore
There is no brand wikidata for Pocztex yet, that's why I went with `operator:`, that's also what polish OSM wiki suggests https://wiki.openstreetmap.org/wiki/Pl:Tag:amenity%3Dparcel_locker#Przyk%C5%82ady_tagowania I'm going to create wikidata nonetheless 
> nor am I confident that they all have the same opening hours.
I am. Look at the request made on https://www.poczta-polska.pl/znajdz-punkt/ select "Pocztex AUTOMAT" click "pokaż filtry" you can then filter for points that are open on weekdays or 24/7, API returns 977 points in both cases.

Cj-Malone:
Thanks

PR #9699, opened 2024-07-06
Don't export `bic` values to taginfo.json

kjonosm:
Don't export `bic` values to taginfo.json


bhousel:
Makes sense 👍 
(For anyone else confused like me, these are bank ID codes used by Spanish banks).

PR #9706, opened 2024-07-09
Add USGS monitoring station preset

### LINKED ISSUE #9705: Add USGS monitoring stations
The USGS has well over 10,000 monitoring stations around the US, but most are not in OSM, and those that exist have spotty tagging patterns. It would be great to have standard tagging. These are [`man_made=monitoring_station`](https://wiki.openstreetmap.org/wiki/Tag:man_made%3Dmonitoring_station)

Most of these are known as streamgages, and while it might be nice to have a specific preset for these, there isn't an explicit streamgage tag in OSM. Instead, features are tagged with what they measure, such as [`monitoring:water_level=yes`](https://wiki.openstreetmap.org/wiki/Key:monitoring:water_level). USGS stations measure various conditions, such as water level, tides, water flow, precipitation, air pressure, salinity, pollution, and soil moisture. Even within the streamgage category, different gages measure different things. As such, I'd suggest the preset should be something generic like "USGS Monitoring Station".

Streamgage network: https://dashboard.waterdata.usgs.gov/app/nwd/en/

quincylvania:
Closes #9705. I haven't contributed to this project in years so I'm not sure if I'm doing this right. Thanks in advance for any assistance :)

### DIFF
--- a/data/operators/man_made/monitoring_station.json
+++ b/data/operators/man_made/monitoring_station.json
@@ -1,0 +1,23 @@
+{
+  "properties": {
+    "path": "operators/man_made/monitoring_station",
+    "preserveTags": ["^name"],
+    "exclude": {
+      "generic": ["^monitoring station$"]
+    }
+  },
+  "items": [
+    {
+      "displayName": "USGS Monitoring Station",
+      "id": "unitedstatesgeologicalsurvey-35adb7",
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Seems fine to me, thank you

PR #9716, opened 2024-07-12
data/operators/fast_food: init

UsefulRabbit:
Created with WKS to start, since they operate ~200 locations across the United States according to this map on their website: https://www.wksusa.com/brands/

Suggestions for other operators greatly appreciated.

### DIFF
--- a/data/operators/amenity/fast_food.json
+++ b/data/operators/amenity/fast_food.json
@@ -1,0 +1,25 @@
+{
+  "properties": {
+    "path": "operators/amenity/fast_food",
+    "preserveTags": [],
+    "exclude": {
+      "generic": [
+        "^[Ff]ast food?$",
+        "^[Rr]estaurant$",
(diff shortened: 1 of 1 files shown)


data/operators/amenity/fast_food.json
```diff
@@ -0,0 +1,25 @@
+{
+  "properties": {
+    "path": "operators/amenity/fast_food",
+    "preserveTags": [],
+    "exclude": {
+      "generic": [
+        "^[Ff]ast food?$",
```
peternewman, line 7: AFAIK this will match:
Fast food
fast food
Fast foo
fast foo

Did you mean foods?

bhousel:
This probably won't really work - we already track a lot of these brands under the brand tree, so we wouldn't also be able to match "WKS" in the 'operator' tag.  Matching is unfortunately per-tree.

PR #9725, opened 2024-07-16
Update fast_food.json Add HD Iskender

jhutton1:


data/brands/amenity/fast_food.json
```diff
@@ -34,6 +34,18 @@
     }
   },
   "items": [
+    {
+      "displayName": "HD Iskender",
+      "locationSet": {"include": ["tr"]},
+      "tags": {
+        "amenity": "fast_food",
```
AntiCompositeNumber, line 41: Are you sure this is fast food? I wasn't able to find a whole lot of information but there's plenty of videos of waiters dumping melted butter on people's plates, which I think would fit `amenity=restaurant` better.

jhutton1: Most locations I found are in shopping malls and the pictures appear to be fast food.  It does appear however there are servers present, suggesting a restaurant.  I agree and will move over to restaurant.

AntiCompositeNumber (approved):

PR #9726, opened 2024-07-16
Include brazilian supermarket chain called Supermarket

nighto:
Adds a brazilian supermarket chain called... Supermarket (in English). Here's their website: https://redesupermarket.com.br/

Btw I wasn't sure what to put in `id`, those seem generated by you so I left a placeholder.

Thanks for the suggestion @1ec5 , ref https://github.com/openstreetmap/iD/issues/10334#issuecomment-2230984728

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -7748,9 +7748,8 @@
     },
     {
       "displayName": "Supermarket (Brasil)",
-      "id": "supermarket-br",
+      "id": "supermarket-20f2a8",
       "locationSet": {"include": ["br"]},
-      "matchTags": ["shop/supermarket"],
       "tags": {
         "brand": "Supermarket",
         "brand:wikidata": "Q127375573",
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -7747,6 +7747,18 @@
       }
     },
     {
+      "displayName": "Supermarket (Brasil)",
+      "id": "supermarket-br",
+      "locationSet": {"include": ["br"]},
+      "matchTags": ["shop/supermarket"],
+      "tags": {
+        "brand": "Supermarket",
+        "brand:wikidata": "Q127375573",
+        "name": "Supermarket",
+        "shop": "supermarket"
+      }
+    },
+    {
       "displayName": "Supermercados BH",
       "id": "supermercadosbh-20f2a8",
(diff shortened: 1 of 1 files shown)


willemarcel (approved):

data/brands/shop/supermarket.json
```diff
@@ -7746,6 +7746,17 @@
         "shop": "supermarket"
       }
     },
+    {
+      "displayName": "Supermarket (Brasil)",
```
1ec5, line 7750: The disambiguating suffix might not be necessary, since we don’t have any other preset named “Supermarket”.

So far, I think all the other disambiguators have been in English (“Brazil”), but only because we don’t have a solution for localizing presets by language, only by geography. Fortunately, this preset is somewhat unlikely to be shown to users of languages other than Portuguese.

data/brands/shop/supermarket.json
```diff
@@ -7746,6 +7746,17 @@
...
+      "id": "supermarket-20f2a8",
+      "locationSet": {"include": ["br"]},
+      "tags": {
+        "brand": "Supermarket",
+        "brand:wikidata": "Q127375573",
```
1ec5, line 7755: [This Wikidata item](https://www.wikidata.org/wiki/Q127375573) may need [some additional information](https://wiki.openstreetmap.org/wiki/Key:brand:wikidata#Establishing_notability) on it. Otherwise, a Wikidata administrator may delete it as spam without realizing it’s legitimate.

nighto: Yeah, unfortunately a Wikipedia article has been created but then deleted last year, with the allegation that the brand "only exists" and therefore is not relevant to be on the encyclopedia.
https://pt.wikipedia.org/wiki/Wikip%C3%A9dia:P%C3%A1ginas_para_eliminar/Rede_Supermarket

nighto: Added a couple statements to Wikidata item, as well as copied [one photo from Mapillary](https://www.mapillary.com/app/?lat=-22.899917602539&lng=-43.202117919922&z=17.693165920333104&pKey=849641508964035&focus=photo) to [Wikimedia Commons](https://commons.wikimedia.org/wiki/File:Supermarket_na_Rua_Santo_Cristo,_Rio_de_Janeiro.jpg), but for whatever reason I wasn't able to link it on the Wikidata item.

PR #9732, opened 2024-07-18
Add Prime Storage (self storage)

mjoe999:
A US / CA self storage company with approx 40 instances in OSM, e.g.

https://www.openstreetmap.org/#map=19/39.27499/-76.67041

... and 300 or so in the "real world".


Cj-Malone (approved):
LGTM

PR #9735, opened 2024-07-19
Add Nioga Library System

Snowysauce:
Public library system for **Ni**agara, **O**rleans, and **G**enesee counties in New York, USA (hence the unusual name chosen by said system)

### DIFF
--- a/data/operators/amenity/library.json
+++ b/data/operators/amenity/library.json
@@ -1344,6 +1344,18 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


AntiCompositeNumber (approved):
While not every library network has enough control over member libraries to be counted as an `operator`, I think this one probably does based on their website. If this turns out to cause problems in the future we can adjust the tagging.

PR #9736, opened 2024-07-20
Remove Serbian Railways as it's a holding company

u-spec-png:
Serbian Railways is just a holding company Serbian Railways Infrastructure manages infrastructure assets.


UKChris-osm:
I don't know anything about the railways in Serbia, but the current entry within the NSI has lots of POI's added to the map already, whereas this suggested edit seems to have none so far.

Could this change cause a conflict within Serbian mappers??

Dimitar5555:
> I don't know anything about the railways in Serbia, but the current entry within the NSI has lots of POI's added to the map already, whereas this suggested edit seems to have none so far.

It's very common to tag the passenger company as operator of the infrastructure since people are not aware that they are operated by different entities. It's also understandable, since for the last 40 or more years there was a single company which managed infrastructure, rail freight and rail passengers.

P.S. @u-spec-png, if you are not going to update all `route=railway` relations in Serbia after this PR is merged, it would be nice include `"matchNames":["Железнице Србије"],`.

u-spec-png:
> P.S. @u-spec-png, if you are not going to update all `route=railway` relations in Serbia after this PR is merged, it would be nice include `"matchNames":["Железнице Србије"],`.

I will

u-spec-png:
> Could this change cause a conflict within Serbian mappers??

No, not rally as Infrastruktura Železnice Srbije is operator and not Železnice Srbije however it's still common for people to tag Železnice Srbije as an operator because they are not aware that it was broken up into smaller companies.

PR #9738, opened 2024-07-21
Update bank.json with old_name for GCB Bank

sammyhawkrad:


UKChris-osm:
Hi @sammyhawkrad , I think this would be better as a value of `matchNames` rather than using the tag `old_name`.

sammyhawkrad:
It's already added to the `matchNames` but I also want the `old_name` to be added.

UKChris-osm:
I wouldn't have thought this was needed, as that information is readily available via the wikidata item linked to it? It's not a tag we generally use within the Name Suggestion Index.

sammyhawkrad:
I usually add it manually on OSM so that users can still find them with old name. If it doesn't fit with NSI, this PR can be discarded.

Cj-Malone:
This is a weird one. `old_name` will massively help Nominatim and other search projects, but I don't think NSI is the place.

UKChris-osm:
If `old_name` is helpful then it could be added, but would many people be searching for it on OSM via an old name in a year or two after a rebranding has taken place, would we not be adding data that's unnecessary in the long run?

Cj-Malone (approved):
I think there should be a better place for this data. But it's not the first `old_name` in NSI. I have no major objection for it being here in the short term.

bhousel:
Yeah I'm pretty ok with this too.  We usually use `alt_name` for when a brand is in the process of switching names or uses both, but `old_name` is fine too if we think people might search for that.

UKChris-osm:
@sammyhawkrad would you be happy with using the `alt_name` tag instead of `old_name`?

sammyhawkrad:
Using alt_name would not be a true reflection of reality because it is their former name.

LaoshuBaby:
If we decide to allow the old name to appear in the preset, according to the semantics of the key, it is obvious that `old_name` is better than `alt_name`

Although this old name only works when used for matching and searching.

I support keeping this tag, just like even if you mention **𝕏** today, you are more accustomed to using its old name, **Twitter**.

PR #9745, opened 2024-07-23
added the right capitalisation to AVIA

wvanderp:


Dimitar5555:
Avia isn't an acronym as far as I know.

wvanderp:
I should have added more context for the change.

If we go to their cooperate website we see that their brand is always written in all caps.
https://avia-international.com/

the local website also concurs
https://avia.nl/

Wikipedia also writes it in all caps
and if we look at the logo then its also all caps

so I believe we should write it in all caps in osm

Dimitar5555:
[Billa](https://www.billa.at), [Lidl](https://www.lidl.bg), [Penny](https://www.penny.de), [Lukoil](https://lukoil.rs/sr), [Gazprom](https://www.gazprom-petrol.bg) and possibly other brands also style their names in all caps but in NSI they are in camel case.

AntiCompositeNumber:
In general, [the convention in OSM](https://wiki.openstreetmap.org/wiki/Names#Proper_name_spelling) is to use title case unless doing so would be confusing. While the [Dutch Wikipedia article](https://nl.wikipedia.org/wiki/AVIA) consistently capitalizes the name, the [German](https://de.wikipedia.org/wiki/Avia_International) and [Polish](https://pl.wikipedia.org/wiki/Avia_International) consistently title-cases it. The [French Wikipedia](https://fr.wikipedia.org/wiki/Avia_International) is inconsistent. 

I'm not seeing a compelling reason to all-caps this brand.

PR #9760, opened 2024-08-01
Add TravisMathew to brands/shop/clothes

impiaaa:
Four instances in OSM currently, out of 54 according to the website. Of the four, two are tagged with `clothes=men`, but other stores have options for women and children. One instance is tagged with `brand:wikidata=Q1027181`, which is the parent company. I added a new QID for the specific brand.


kjonosm:
Thanks @impiaaa!

PR #9761, opened 2024-08-01
Added match names for three banks

Snowysauce:
I closed and broke apart #9754 because I'm assuming it was too unwieldy to review/merge. This is the first of the partial pull requests.


kjonosm:
Looks good to me

PR #9781, opened 2024-08-10
Add Monroe County Parks, B&ECPL; fix Wikidata ID for Salt Mobile

Snowysauce:


Snowysauce:
Self-closing to exclude Monroe County Parks. There are only 20 parks ("only" in the context of the [suggested inclusion criteria](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case)), and since they're in one county, they can all be updated as needed in one OSM session. By contrast, the B&ECPL system has 37 branches, and meets the criteria for NSI inclusion.

PR #9782, opened 2024-08-10
More New York State library systems

Snowysauce:
These are the last five library systems in Western and Central New York State that are not in or proposed to be in the NSI.


kjonosm:
Looks good to me

PR #9845, opened 2024-08-27
Update sports.json Add Admiral Sports

jhutton1:


kjonosm:
@jhutton1: Do they run any actual shops?

jhutton1:
> @jhutton1: Do they run any actual shops?

Yes, they do.  Example from Greece:

image

kjonosm:
Thanks @jhutton1, I will limit the location to Greece...

PR #9851, opened 2024-08-28
Add operator presets for highway=street_lamp (CenterPoint Energy, Puget Sound Energy, and Seattle City Light)

CorruptComputer:
CenterPoint Energy owns and operates all of the street lights here, having a preset for this would make is easy to add that info. Used this as a reference for the values of that:
https://github.com/osmlab/name-suggestion-index/blob/8c39905f1d87f31bce9b40d4359bce36cafd1374/data/operators/power/line.json#L983

Discussed this in the OSM Discord as well, and someone had these values they were using for ones in their area, so I added them as well:
https://github.com/Lumikeiju/openstreetmap/blob/main/presets/seattle_city_light_pole.xml

### DIFF
--- a/data/operators/highway/street_lamp.json
+++ b/data/operators/highway/street_lamp.json
@@ -1,0 +1,39 @@
+{
+  "properties": {
+    "path": "operators/highway/street_lamp",
+    "exclude": {
+      "generic": ["^street_lamp$"]
+    }
+  },
+  "items": [
+    {
+      "displayName": "CenterPoint Energy",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "operator": "CenterPoint Energy",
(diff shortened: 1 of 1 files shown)


Snowysauce (approved):

Snowysauce:
I like this idea - so much so that I filed a feature request with iD to make adding street light operator names easier. I don't personally map them, but a quick overpass turbo search of my area showed a couple hundred mapped lights, all without operator details. I can't help but think that's linked to currently having to add the operator via the "tags" section.

CorruptComputer:
Thanks for that suggestion, I've opened a second PR over in the id-tagging-schema repo adding that as well:
https://github.com/openstreetmap/id-tagging-schema/pull/1327

PR #9856, opened 2024-08-30
Rename ČEZ to ČEZ Distribuce for electricity distribution

onovy:
See: https://wiki.openstreetmap.org/wiki/Automated_edits/onovy/%C4%8CEZ

### DIFF
--- a/data/operators/power/line.json
+++ b/data/operators/power/line.json
@@ -1086,18 +1086,18 @@
       }
     },
     {
-      "displayName": "ČEZ",
+      "displayName": "ČEZ Distribuce",
       "id": "cez-452a56",
       "locationSet": {
         "include": ["cz", "de", "sk"]
       },
       "matchNames": [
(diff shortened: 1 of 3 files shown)


Snowysauce:
The pull request as currently coded would change the suggested operator for ČEZ distribution entities in Germany and Slovakia along with the Czech Republic. Would the new operator still be correct in those countries, or does a different subsidiary of ČEZ handle distribution outside of the Czech Republic?

onovy:
As you can check here:
https://overpass-turbo.eu/s/1QsS

Operator ČEZ is not used outside of Czechia. I think de+sk countries are here only for power lines etc. near borders, so maybe we should remove them?

Snowysauce:
Thanks for providing the overpass turbo link. After reviewing the results, I'm comfortable merging this pull request as originally coded.

Snowysauce (approved):

PR #9857, opened 2024-08-30
Update convenience.json : add My Auchan

raphodn:
### What ?

Add a new supermarket "My Auchan"

### Why ?

In France, Auchan has a convenience-like chain called "My Auchan" (it used to be called "A2 Pas").

### Source

- https://www.auchan-retail.com/fr/nos-franchises/
- wikidata : https://www.wikidata.org/wiki/Q115800307
- similar : Carrefour City ([wikidata](https://www.wikidata.org/wiki/Q2940187)), Intermarché Express ([wikidata](https://www.wikidata.org/wiki/Q98278043))

### Extra info

It also sometimes spelled "MyAuchan" (without the space).
Is there a way to have both spellings ?

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -3549,6 +3549,16 @@
       }
     },
     {
+      "displayName": "My Auchan",
+      "locationSet": {"include": ["fr"]},
+      "tags": {
+        "brand": "My Auchan",
+        "brand:wikidata": "Q115800307",
+        "name": "My Auchan",
+        "shop": "convenience"
+      }
+    },
+    {
       "displayName": "myNEWS.com",
       "id": "mynewscom-3e5671",
       "locationSet": {"include": ["my"]},
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -3551,6 +3551,8 @@
     {
       "displayName": "My Auchan",
       "locationSet": {"include": ["fr"]},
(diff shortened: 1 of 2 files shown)


UKChris-osm:
Would this be more suitable in the `shop/convenience `category?

raphodn:
> Would this be more suitable in the `shop/convenience `category?

Good question.

In the `supermarket.json` I see chains like "Intermarché Express" or "Carrefour City/Market/Contact" or "U Express", which are pretty much the same size.

But in the `convenience.json` there is also "Franprix" or "Carrefour Express".

Ideally I think i would : 
- indeed put it in the convenience json instead
- move "U Express" & "Carrefour Express" there as well

UKChris-osm:
Those brands seem like they would also be more suitable in the `shop/convenience` category.

I'd recommend changing this PR for My Auchan to be part of the `shop/convenience` category, then submit another PR for any changes to existing entries so that others can review beforehand.

raphodn:
Sounds good ! Will update this PR this weekend.

For the other shops isn't there going to be an impact on the existing OSM POIs ? If yes should we create a seperate issue to discuss this ?

raphodn:
@UKChris-osm done :)

UKChris-osm:
There might be an impact on the other brands if they are switched to another category, but if they are more suitable to that category then iD and other mapping software will simply suggest the update to mappers :)

PR #9860, opened 2024-08-31
Bike Tower

starsep:


Snowysauce:
I looked at the links on the Wikidata entry, and the company's website and social media appear to refer to themselves as "Biketower", "BIKETOWER", or "BikeTower".

UKChris-osm:
I did a search using https://overpass-turbo.eu/s/1RUv and found that "BikeTower" is the most used text at the moment, so I'd suggest we stick with that for the time being.

starsep:
I have changed Bike Tower to BikeTower

PR #9881, opened 2024-09-01
Update frozen_food.json : Add Écomiam

remsoff:


UKChris-osm:
Thanks @remsoff - just to let you know we don't include the `brand:website` tag in the NSI.

remsoff:
Ok, thank you

PR #9886, opened 2024-09-01
Update butcher.json add Les Éleveurs de la Charentonne

remsoff:


kjonosm:
Thanks @remsoff! I limited `locationSet` to `fx`

PR #9891, opened 2024-09-02
Update of location for charging station Eldrive

neau:


Dimitar5555:
Is Eldrive a Bulgarian company? If it is, does it directly operate the stations in Romania and Lithuania or it has subsidiaries there?

kjonosm:
Anyway, `displayName` should somehow correspond with the actual `operator` or `brand` name.

Snowysauce:
The proposed locationSet appears to be correct per https://eldrive.eu/#charging__station. The company has individual offices in each country according to the contact info at the bottom of the page, but I don't know if they're just offices or subsidiary headquarters. I've refrained from merging this because the brand uses English for branding in some capacity, and so it should have at minimum an `operator:en` tag (along with the relevant local language tag to specify what language the default name is in).

Snowysauce:
This PR seems to have been abandoned, so I removed the questionable operator name change while merging in the locationSet change.

PR #9895, opened 2024-09-02
Update supermarket.json : add Biomonde

raphodn:
And the French supermarket chain "Biomonde"

https://www.wikidata.org/wiki/Q120801927

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -900,6 +900,17 @@
       }
     },
     {
+      "displayName": "Biomonde",
+      "locationSet": {"include": ["fr"]},
+      "matchNames": ["bio monde"],
+      "tags": {
+        "brand": "Biomonde",
+        "brand:wikidata": "Q120801927",
+        "name": "Biomonde",
+        "shop": "supermarket"
+      }
+    },
+    {
       "displayName": "BM",
       "id": "bm-0513f1",
       "locationSet": {"include": ["es"]},


kjonosm:
@raphodn  Could you please limit the `locationSet` to `fx`?
For reference, see https://fr.wikipedia.org/wiki/France_métropolitaine

raphodn:
> @raphodn Could you please limit the `locationSet` to `fx`? For reference, see [https://fr.wikipedia.org/wiki/France_métropolitaine](https://fr.wikipedia.org/wiki/France_m%C3%A9tropolitaine)

Why exactly ? In this particular case there are a couple of shops outside of fx... You can check the map here : https://magasins.biomonde.fr/

And how can we be sure that a franchise won't open in the DOM-TOM ?

kjonosm:
Hey @raphodn, thanks for clarifying. As a rule of thumb, we try to limit locationSets as much as possible to not suggest brands in areas they don't exist. In this case you're right though, it warns obvious by a quick glance on their websites stores map.

raphodn:
Ok @kjonosm but isn't replacing "fr" by "fx" such a small change but with possibly a big impact and to the detriment of local mappers in the DOM-TOM who won't have access to many French store brands in ID ?

I did a quick check, in the repo there are ~4500 occurences of "fr", and as much of "fx".

I went through a few, there are cases where "fx" should be replaced by "fr"..

Could you elaborate on the technical differences/optimizations ? Maybe in a Github Issue or Discussion ? Thanks !

kjonosm:
As already stated, we try to limit locationSets as much as possible to not falsely suggest brands in areas they don't exist. 
So, for items also present in the DOM-TOM, you're very welcome to use "fr". If you come across items in NSI which are present in the DOM-TOM but falsely list "fx" in their location set, please go ahead and fix it! 
However, for items only present within Metropolitan France, please use "fx" instead. Thanks again for your recent contributions!

PR #9896, opened 2024-09-02
Update convenience.json : add day by day

raphodn:
And the French convenience chain "day by day" (bulk_purchase only)

https://www.wikidata.org/wiki/Q48748765


kjonosm:
@raphodn  Could you please limit the `locationSet` to `fx`?
For reference, see https://fr.wikipedia.org/wiki/France_métropolitaine

PR #9900, opened 2024-09-03
Change Česká pošta name in man_made=surveillance to same as in amenit…

onovy:
…y=post_office

### DIFF
--- a/data/operators/man_made/surveillance.json
+++ b/data/operators/man_made/surveillance.json
@@ -704,12 +704,12 @@
       }
     },
     {
-      "displayName": "Česká Pošta",
+      "displayName": "Česká pošta",
       "id": "ceskaposta-5f01ff",
       "locationSet": {"include": ["cz"]},
       "tags": {
         "man_made": "surveillance",
-        "operator": "Česká Pošta",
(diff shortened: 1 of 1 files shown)


Snowysauce:
Should the name be "Česká pošta, s.p." or "Česká pošta"? Their logo and the relevant Wikidata page omit the suffix, which implies that the latter is the common name. (I'm aware that the `amenity=post_office` instance with the suffix has thousands of uses - however, that shouldn't be a factor in determining what the name value should be.)

onovy:
I asked on [talk-cz](https://lists.openstreetmap.org/pipermail/talk-cz/2024-September/026928.html) and community wants "Česká pošta, s.p.". TBH: I think "Česká pošta" is better too.

Snowysauce:
> I asked on [talk-cz](https://lists.openstreetmap.org/pipermail/talk-cz/2024-September/026928.html) and community wants "Česká pošta, s.p.". TBH: I think "Česká pošta" is better too.

Thanks for reaching out to the community. Since the consensus was to include the suffix, I'll merge this pull request into the repo.

PR #9906, opened 2024-09-05
Sort some list entries

harahu:
When running `npm run build` on `main`, I noticed these changes were made. I thought I might as well make a PR out of them.


UKChris-osm:
Thanks for this @harahu, but `npm run build` is run quite often by maintainers, and doesn't really need to be done via a PR.

If making any other edits or improvements however, you're welcome to also run `npm run build` with your PR's

PR #9907, opened 2024-09-05
Change import assertions to import attributes

harahu:
I couldn't get the project to build on Node v22, so I looked around and found this issue: https://stackoverflow.com/questions/78876691/syntaxerror-unexpected-identifier-assert-on-json-import-in-node-v22

After following one of the proposals, swapping import assertions to import attributes, the project builds on node v22 just fine. However, this means it won't build on Node v17 anymore.

### DIFF
--- a/lib/file_tree.js
+++ b/lib/file_tree.js
@@ -13,9 +13,9 @@ import { sortObject } from './sort_object.js';
 import { validate } from './validate.js';
 
 // JSON
-import treesJSON from '../config/trees.json' assert {type: 'json'};
+import treesJSON from '../config/trees.json' with {type: 'json'};
 const trees = treesJSON.trees;
-import categoriesSchemaJSON from '../schema/categories.json' assert {type: 'json'};
+import categoriesSchemaJSON from '../schema/categories.json' with {type: 'json'};
 
 
 // The code in here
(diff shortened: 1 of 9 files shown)


bhousel:
Oh these things again.. This relates to #6014

bhousel:
Yeah I think for now we might need to just stick to node 18 and 20.

Snowysauce:
I've been using modified JS files with the suggested code changes in order to run the build and Wikidata scripts on Node 22. What I was/am doing was loading the modded files into my local clone of the repository, running the script, then discarding the changed JS files prior to uploading the script output. I knew uploading the modded files would break compatibility for anyone running older versions of Node, and I didn't want to unilaterally force others to upgrade.

harahu:
> Yeah I think for now we might need to just stick to node 18 and 20.

Not sure I understand what you mean here. This PR is compatible with all versions >= 18, including 18 and 20, so that means you think this is fine to merge? Or do you mean that it is a goal to _avoid_ compatibility with v22?

bhousel:
oh nice!  I didn't realize that the `with` syntax was backported to v18 and 20

https://github.com/nodejs/node/issues/51622
https://github.com/nodejs/node/pull/51136

In that case, then yes, we should just switch to the accepted syntax.  Thanks for the PR!

PR #9908, opened 2024-09-05
Update BKK and Eviny power companies

harahu:
- Eviny is the parent company of BKK and Eviny Fornybar, a grid company and a power generation company, respectively. Eviny does not operate infrastructure themselves, only their child companies do. Hence why I've de-listed them.
- Eviny Fornybar operates power plants, generators and  grid infrastructure on the premises of their plants. Hence why I've added them to `power/generator` and `power/line`.
- BKK Produksjon is an old name for Eviny Fornybar, hence why I've de-listed them.

### DIFF
--- a/data/operators/power/generator.json
+++ b/data/operators/power/generator.json
@@ -3249,6 +3249,21 @@
       }
     },
     {
+      "displayName": "Eviny Fornybar",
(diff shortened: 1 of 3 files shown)


kjonosm:
Thanks @harahu! Looks good to me.

PR #9913, opened 2024-09-06
Fixup entries for RBC

wolfy1339:
The main RBC brand does not exist in the US anymore
> In 2012, RBC Bank USA's banking division for American clients was sold to PNC Financial Services[1]

There are some mistagged entries in the USA where there is a bank named Royal Bank which is different from RBC.

In the rest of the world, the banks tagged as RBC are most likely not banks in the sense that we tag them in OSM.
I can't find any evidence that there are bank branches operated elsewhere than Canada and the Carribean under the RBC brand name

[1]: https://en.wikipedia.org/wiki/RBC_Bank

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -10016,6 +10016,18 @@
       }
     },
     {
+      "displayName": "PNC",
+      "id": "pnc-ea2e2d",
+      "locationSet": {"include": ["us"]},
+      "matchNames": ["centura", "rbc"],
+      "tags": {
+        "amenity": "bank",
+        "brand": "PNC",
+        "brand:wikidata": "Q38928",
+        "name": "PNC"
+      }
+    },
+    {
       "displayName": "PNC Bank",
       "id": "pncbank-ea2e2d",
       "locationSet": {"include": ["us"]},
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Not being Canadian myself, I don't know how things work with regards to the French-Canadian language, but could the two entries be merged into one, using `name:fr` as well, which I noticed is present in all the entries when you select the Overpass Turbo link under [RBC Banque Royale](https://nsi.guide/index.html?t=brands&k=amenity&v=bank&tt=rbc#rbcbanqueroyale-e1345b)

bhousel:
I think in the past we have done separate entries for Quebec because their brands use French in the `name` tag.

Cj-Malone:
> I think in the past we have done separate entries for Quebec because their brands use French in the `name` tag.

Yeah, but I'd like to discuss this and the HK thing of both languages in `name` after OSM.org switches to vector tiles. It is mapping for the renderer.

wolfy1339:
It's not just the name tag, it's also the display name that gets translated

Snowysauce:
Just an FYI: significant parts of the code related to this PR were modified in https://github.com/osmlab/name-suggestion-index/commit/994bfeb3af47c7f84f2ad2fa970c1e4afd567f95.

Snowysauce:
Closing as an apparently abandoned PR.

PR #9928, opened 2024-09-09
Update Wikidata ID for Mobil

loganrosen:
The Mobil brand was previously under a separate Wikidata ID ([Q109676002](https://www.wikidata.org/w/index.php?title=Q109676002&redirect=no)) from the company ([Q3088656](https://www.wikidata.org/wiki/Q3088656)), but now [it has been merged into a single entry](https://www.wikidata.org/w/index.php?title=Q109676002&action=history). I've updated the IDs in the dataset accordingly and created a replacement to follow the redirect.

### DIFF
--- a/config/replacements.json
+++ b/config/replacements.json
@@ -102,9 +102,9 @@
       "note": "MEO - Wikidata redirect, https://github.com/osmlab/name-suggestion-index/issues/9543",
       "wikidata": "Q1365938"
     },
-    "Q3088656": {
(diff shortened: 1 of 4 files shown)


1ec5:
The changes on Wikidata are incorrect and should be reverted. A single item should not be simultaneously a former company and a current chain.

1ec5:
I’ve reverted the changes at Wikidata.

PR #9930, opened 2024-09-09
add "matchNames" in club=scout

tiptoptom:
In addition to #9889


Snowysauce:
Hi! Thanks for your contribution. Are these common abbreviations for the brands? If so, they could be added as `brand:short` tags instead. The difference is that the `brand:short` tag and value would become a suggested addition to OSM in editors using the NSI data (e.g., iD, JOSM), while `matchNames` are not seen outside of the NSI code. If the presence of a `brand:short` tag is undesired, these can indeed be added as `matchNames`.

PR #9931, opened 2024-09-09
Cyprus Island fuel stations

dmromanov:


Snowysauce:
"Fill n Go" and "StarOil (Cyprus)" are missing Wikidata tags. I assume the website tags are intended to be replacements for the Wikidata tags; however, tags like "website" are [automatically removed](https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#tags) whenever the index build script is run. In addition, entries without Wikidata tags are [not exported in releases](https://github.com/osmlab/name-suggestion-index/wiki/Technical-Details#availability-of-entries-in-downstream-projects), and will not become presets in OSM editors. Please [create Wikidata pages](https://github.com/osmlab/name-suggestion-index/wiki/Editing-Wikidata#creating-wikidata-pages) for both items if you would like to add them to the NSI.

PR #9939, opened 2024-09-11
Update optician.json "Les Opticiens Mutualistes" => "Écouter Voir"

cayenne17:
Remove OLD "Les Opticiens Mutualistes"
Add old_name on new

I don't know if that's how it works?

### DIFF
--- a/data/brands/shop/optician.json
+++ b/data/brands/shop/optician.json
@@ -271,12 +271,11 @@
       "id": "ecoutervoir-8cd79e",
       "locationSet": {"include": ["fr"]},
       "matchTags": ["shop/medical_supply"],
-      "matchNames": ["Les Opticiens Mutualistes"],
+      "matchNames": ["Les Opticiens Mutualistes", "opticiens mutualistes"],
       "tags": {
         "brand": "Écouter Voir",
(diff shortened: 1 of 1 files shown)


Cj-Malone:
We've agreed in principle for `old_name`. https://github.com/osmlab/name-suggestion-index/pull/9738 However it sounds like this rename happened 5 years ago. I think `matchNames` would be a better place for it.

cayenne17:
> We've agreed in principle for `old_name`. #9738 However it sounds like this rename happened 5 years ago. I think `matchNames` would be a better place for it.

`matchNames` has nothing to do with OSM?

`old_name` I use it on OSM so that when people search with the old name, they find it.

Cj-Malone:
`matchNames` is a NSI/iD thing so people can search for the old name and get recommend the entry. The value doesn't end up in OSM.

https://github.com/osmlab/name-suggestion-index/blob/bdaefe3ce78b949eff7d862593ce8321fd636eb3/data/brands/shop/optician.json#L137-140

cayenne17:
@Cj-Malone I've just added matchNames with the old name.

Cj-Malone:
Thanks

PR #9943, opened 2024-09-12
Update supermarket.json - Updated Netto City

ckpiper:
I see in the ATP spider some store names are called "Netto City" but when I checked google street view. They all look like 
https://commons.wikimedia.org/wiki/File:NETTO-Wandsbek.jpg
Thus I think the brand could be `Netto Marken-Discount` with the name still being Netto City. I cannot tell the difference between `Netto Marken-Discount` and `Netto City Marken-Discount`. It seems less obvious than Taco Bell and Taco Bell Cantina which both have same brand tag.

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -5783,7 +5783,7 @@
       "id": "nettocity-0a839e",
       "locationSet": {"include": ["de"]},
       "tags": {
-        "brand": "Netto City",
+        "brand": "Netto Marken-Discount",
         "brand:wikidata": "Q879858",
         "name": "Netto City",
         "shop": "supermarket"


ckpiper:
@Snowysauce do you think this is fine to merge?

Snowysauce:
**TL;DR version:** I'm not sure.

---

I'm not familiar with this brand (or with most things European), so I did some quick searching. My findings:

- NSI currently has three entries associated with [Wikidata ID Q879858](https://www.wikidata.org/wiki/Q879858): "Netto City", "Netto Marken-Discount" (shop/supermarket), and "Netto Getränke-Discount" (shop/beverages).
- The [store locator](https://www.netto-online.de/filialfinder) on the official website provided by the QID shows that stores operating under all three names do exist. "Netto Marken-Discount" is the primary brand and by far the most common (and is the QID's label); "Netto Getränke-Discount" focuses on drinks and is less prevalent; and "Netto City" is rare, appearing to be confined to the innermost areas of large cities.
- Thus, Netto City may be a downscaled version of Netto Marken-Discount for storefronts in urban areas with less square feet (floor space) than the standard Netto Marken-Discount.

From an American perspective, Netto City appears to be to Netto Marken-Discount what Dollar General tried/is trying to do with "DGX" - a smaller store in urban areas touting all the essential features of their full-line stores. NSI does not have an entry for DGX as of this writing (an [overpass turbo query](https://overpass-turbo.eu/s/1Ri5) only returns 9 locations), and there is no consistency to how each DGX location is tagged in OSM. Some are tagged as `brand=Dollar General`; others are not. None are tagged as `brand=DGX`.

However, where DGX differs from Netto City is that Dollar General [doesn't list DGX stores on their website any differently than their main stores](https://www.dollargeneral.com/store-directory/ny/rochester). DGX stores are clearly (IMO) `name=DGX`, `brand=Dollar General`, but as long as Netto Marken-Discount continues to clearly delineate which stores are "Netto Marken-Discount" and which are "Netto City", it's debatable as to whether the `brand` for Netto City stores is "Netto Marken-Discount" or "Netto City".

PR #9944, opened 2024-09-14
Add Ghana fuel stations

sammyhawkrad:


Snowysauce:
Are these stations operated by the gas companies or by independent station owners/franchises? Also, regarding `"old_name": "Aminaser Oil"`, when did the rename to Amser Oil occur?

sammyhawkrad:
Each brand added here is an independent gas company with multiple stations.

I can't really tell when the change occurred but if you look at their Facebook page the url (https://www.facebook.com/AminaserOilLtd) has the old name but it uses the new name. The profile picture also shows the logo for the new name.

The stations are branded with the new name but the operating company maintains the old name.

Snowysauce:
> Each brand added here is an independent gas company with multiple stations.

Yes, but are the individual stations owned and operated by the companies? If they're not, there shouldn't be preset `operator` tags.

sammyhawkrad:
Yes, the individual stations are owned and operated by the companies.

Snowysauce:
> I can't really tell when the change occurred but if you look at their Facebook page the url (https://www.facebook.com/AminaserOilLtd) has the old name but it uses the new name. The profile picture also shows the logo for the new name.
> 
> The stations are branded with the new name but the operating company maintains the old name.

Unless the name change was fairly recent, the `old_name` tag is unnecessary. Since the proposed entry has the old name as a match name, iD and other editors will still search for the old name when browsing presets.

The Wikidata ID in the `operator:wikidata` field for this brand also points to an unrelated page.

sammyhawkrad:
> The Wikidata ID in the operator:wikidata field for this brand also points to an unrelated page.

Thanks for catching that.

>  the old_name tag is unnecessary

The match name will be used by the editors but the old name will be useful for OSM end users. There have been discussions about `old_name` and it was deemed acceptable in NSI.

Snowysauce:
> The match name will be used by the editors but the old name will be useful for OSM end users. There have been discussions about `old_name` and it was deemed acceptable in NSI.

I'm aware of the referenced discussion; it was deemed acceptable on a short-term basis for recent renames. An [even more recent discussion](https://github.com/osmlab/name-suggestion-index/pull/9939) rejected a proposed use of `old_name` because the name change happened several years prior.

sammyhawkrad:
Alright no problem, I will remove the old name here and add them directly in OSM.

Snowysauce:
I would recommend dropping the operator tags per https://github.com/osmlab/name-suggestion-index/pull/10131#issuecomment-2471625459. If those tags were not present, I (and possibly others) would have merged this PR weeks ago.

PR #9947, opened 2024-09-14
Add Westport Public Schools

loganrosen:
Adding the [Westport Public Schools](https://www.wikidata.org/wiki/Q130300750) school district.

### DIFF
--- a/data/operators/amenity/school.json
+++ b/data/operators/amenity/school.json
@@ -50234,6 +50234,19 @@
       }
     },
     {
+      "displayName": "Westport Public Schools",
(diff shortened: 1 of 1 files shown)


Snowysauce:
The school's website only lists nine schools, and 10 instances of an item is generally the [minimum requirement](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#notability) for an item to be added to the NSI.

UKChris-osm:
@Snowysauce while you're correct, I'd say, in this case, I think 9 is close enough.

PR #9958, opened 2024-09-17
[brands/amenity/fuel][EKO (Serba)] correct locationSet

panasonic-mafia:
`sr` is Surinam, not Serbia.


Dimitar5555:
Thank you for catching it!

PR #9959, opened 2024-09-17
TRNC banks

dmromanov:


Snowysauce:
While filling out the banks' Wikidata pages, I noticed a couple things:

- Capital Bank appears to style their name as "CapitalBank"
- KoopBank appears to style their name as "Koopbank"
- Kıbrıs Faisal İslam Bankası/Central Bank of TRNC appears to only operate one branch of their own per https://www.kktcmerkezbankasi.org/tr/bilgiler/banka_subeleri. The rest of the listed branches appear to be those of other banks. (I could be completely wrong as I don't speak Turkish.)

dmromanov:
1&2 fixed, 3 removed

PR #9961, opened 2024-09-18
Merged two versions of `BKK-automata` (TVM)

gy-mate:
- Removed the old and incorrect version of `BKK-automata` (public transport ticket vending machine)
- Added common tags to the new version (`BKK jegy- és bérletkiadó automata`) – see [this node](https://www.openstreetmap.org/node/4339675749) for example


Snowysauce:
The merged entry does not have a `brand:wikidata` tag and will disappear from the list of brand presets if left as-is. (See [this section in our wiki](https://github.com/osmlab/name-suggestion-index/wiki/Technical-Details#availability-of-entries-in-downstream-projects) for details.) Since it has an `operator:wikidata` tag, the entry could be moved to `operators/amenity/vending_machine` if this is the only brand that this particular operator has.

gy-mate:
> Since it has an `operator:wikidata` tag, the entry could be moved to `operators/amenity/vending_machine` if this is the only brand that this particular operator has.

@Snowysauce Unfortunately this isn't the only kind of ticket vending machine BKK has so this isn't a viable option.

> The merged entry does not have a `brand:wikidata` tag and will disappear from the list of brand presets if left as-is. (See [this section in our wiki](https://github.com/osmlab/name-suggestion-index/wiki/Technical-Details#availability-of-entries-in-downstream-projects) for details.)

Thank you for the heads up! I've created a new Wikidata item for it and added that ID to the list.

Snowysauce:
> Thank you for the heads up! I've created a new Wikidata item for it and added that ID to the list.

Thanks! Everything now looks good to go, so I'll pull these commits into the repo.

PR #9976, opened 2024-09-23
Add Hotshots Sports Bar and Grill

tylerfilla:
This adds an American sports bar chain called Hotshots Sports Bar and Grill (https://hotshotsnet.com) Only two locations seem to be mapped. They're in 5 states, 14 locations, and another coming soon according to their website. I tagged it as a restaurant with cuisine=burger;pizza after looking at how other sports bar chains are tagged. Feel free to change that.

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -2993,6 +2993,18 @@
       }
     },
     {
+      "displayName": "Hotshots Sports Bar and Grill",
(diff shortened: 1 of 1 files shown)


Cj-Malone (approved):
15 locations is a little low, but OK

tylerfilla:
Whoops, I thought the threshold was 10. Can't remember where I read that. Sorry!

PR #9989, opened 2024-09-28
Add supermarkets in Ghana

sammyhawkrad:


Snowysauce:
I looked at `supermarket.json`, and `operator` as a preset seems to generally be reserved for when the operating entity has its own Wikidata page and is the operator of multiple chains. Other maintainers may feel differently, but unless there's something I'm not aware of, I don't see the need for the operator tag for the Melcom brand.

sammyhawkrad:
The Melcom Group of Companies consists of six separate entities: Melcom Limited (supermarket brand), Century Industries Limited, Crownstar Electronic Industries Limited, Melcom Hospitality, Melcom Travels, and Melcom Care.

So the parent company is what I have added as operator.

UKChris-osm:
Thanks for the submission @sammyhawkrad, but I'd focus on the `brand` only and not include the `operator` in this case, as the `brand` info is the main target within this category.

Issue #4426, opened 2020-10-09
Add France Literie


pyrog:
```
name=France Literie
shop=bed
website=https://www.france-literie.fr
wikidata=Q100249535
```
Currently 23 / 87 shops in France https://overpass-turbo.eu/s/YSZ

pyrog:
Thank you 😃

PR #9995, opened 2024-09-30
Corrects brand name for "France Literie"

yned:


serhii-muchychka:
This entry was added in https://github.com/osmlab/name-suggestion-index/issues/4426 .

PR #10012, opened 2024-10-06
Add TCAT (Tompkins Consolidated Area Transit)

loganrosen:
Adding [TCAT (Tompkins Consolidated Area Transit)](https://www.wikidata.org/wiki/Q7820360).

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -238,6 +238,16 @@
       }
     },
     {
+      "displayName": "Airport Shuttle Lelystad",
+      "id": "airportshuttlelelystad-2f3200",
+      "locationSet": {"include": ["nl"]},
+      "tags": {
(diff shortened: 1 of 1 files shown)


Snowysauce:
It seems strange to me to include the network's official name in the `displayName` without using it publicly in an OSM tag. "TCAT" is unquestionably the common name for the network, but [many of the existing instances in OSM](https://overpass-turbo.eu/s/1SgV) are also tagged with `operator = Tompkins Consolidated Area Transit`. I would either add that `operator` tag and value to the list of suggested tags, or remove the full name from the display name.

loganrosen:
Including the official name in the `displayName` was necessary to disambiguate it from another entry with a name of "TCAT." I've added the `operator` tag as you've suggested.

PR #10022, opened 2024-10-08
Fix Obs bygg linking to Obs supermarket's wikidata

andrebk:
Obs bygg wikidata entry was for the Obs supermarket, a different chain owned by the same parent organization.

Corrected the wikidata ID and added Obs supermarket as a separate entity.


kjonosm:
Thanks @andrebk!

PR #10030, opened 2024-10-12
Update bank.json - Add First Neighborhood Bank

### LINKED ISSUE #9927: First Neighborhood Bank (formerly First National Bank West Virginia)
First Neighborhood Bank (formerly First National Bank). I suppose it is a branch for West Virginia.

https://www.firstneighborhoodbank.com/about-us/office-locations

> In November of 2006, the bank changed from a national to a state-chartered bank, and marked the change with a new name – First Neighborhood Bank

https://www.firstneighborhoodbank.com/about-us/history

Wikipedia/wikidata and NSI seem to not mention this name/branch, while they have done prior research.
It does not have a wikidata page.

This was prompted by 'ATM / drive-through' at https://www.openstreetmap.org/way/1199252110 (see Bing Streetside)

danieldegroot2:
Fixes #9927.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -6807,6 +6807,20 @@
       }
     },
     {
+      "displayName": "First Neighborhood Bank",
+      "id": "firstneighborhoodbank",
+      "locationSet": {
+        "include": ["us-wv.geojson"]
+      },
+      "matchNames": ["First National Bank"],
+      "tags": {
+        "amenity": "bank",
+        "brand": "First Neighborhood Bank",
+        "brand:wikidata": "Q130383275",
+        "name": "First Neighborhood Bank"
+      }
+    },
+    {
       "displayName": "First Republic Bank",
       "id": "firstrepublicbank-ea2e2d",
       "locationSet": {"include": ["us"]},


Snowysauce:
The specified `matchName` will conflict with the following entry, as that entry and the proposed entry in this PR are both geofenced to West Virginia:

https://github.com/osmlab/name-suggestion-index/blob/8c55c57adea362566786b21ae076a9b94a3f190b/data/brands/amenity/bank.json#L6735-L6757

data/brands/amenity/bank.json
```diff
@@ -6806,6 +6806,20 @@
         "name": "First National Bank Texas"
       }
     },
+    {
+      "displayName": "First Neighborhood Bank",
+      "id": "firstneighborhoodbank",
+      "locationSet": {
+        "include": ["us-wv.geojson"]
+      },
+      "matchNames": ["First National Bank"],
```
danieldegroot2, line 6815:
```suggestion
      "matchNames": ["fnb"],
```
@Snowysauce Sorry, I used the old name here as I almost mistook it for the First National Bank (USA).
Is the current signed short name okay as match? It is the same matchname as First National Bank.

Snowysauce: This `matchName` would have the same issue as the first, in that editors using the NSI data for presets/suggestions wouldn't know which entry to present for a bank in West Virginia with the name "FNB". (In iD, it would cycle through both entries, forcing the use of both `brand:wikidata` and `not:brand:wikidata` to fully resolve all warnings.)

I personally think having a `matchName` with this entry does more harm than good. [Everything in West Virginia tagged with "FNB" or "First National Bank" in either name or operator](https://overpass-turbo.eu/s/1Szy) belongs to a different entity than the subject of this PR.

PR #10031, opened 2024-10-12
Zest charging stations

codeinabox:
Zest, in their own words: "operate EV charging solutions for local authorities, brands and destinations"

A map of current locations: https://map.zest.uk.com/

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -6860,7 +6860,7 @@
     {
       "displayName": "Zest",
       "locationSet": {
-        "include": ["gb-eng", "gb-sct"]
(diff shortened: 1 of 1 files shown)


data/operators/amenity/charging_station.json
```diff
@@ -6857,6 +6857,17 @@
         "operator": "ZES"
       }
     },
+    {
+      "displayName": "Zest",
+      "locationSet": {
+        "include": ["gb-eng", "gb-sct"]
```
Cj-Malone, line 6863: Can we just do GB? It's only a matter of time until they have one in Wales.
```suggestion
        "include": ["gb"]
```

PR #10033, opened 2024-10-13
Also match Pizza Express

peternewman:


Snowysauce:
Thanks for the submission, but slight variations in a name such as the one suggested here are already covered by the NSI's [built-in fuzzy matcher](https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#matchNames-matchTags). (I tested this by adding this specific suggestion to my local copy of the repository, and as expected, a run of the [build script](https://github.com/osmlab/name-suggestion-index/wiki/Glossary#build-script) removed the `matchNames` tag.)

peternewman:
Hi @Snowysauce ,

I was about to raise an issue with StreetComplete where I consume the NSI data, but having just tried in OSM via iD I get the same thing there. If I search for PizzaExpress I can add a node like that fine, but "Pizza Express" only finds "Pizza Hut Express".

So it seems like either the fuzzy matcher/build script need updating to allow this particular case or at least two of the NSI consumers need fixing...

Although I can also see a reasonably strong argument for now just allowing random whitespace anywhere in a name when trying to match, for risk of false positives.

Snowysauce:
> I was about to raise an issue with StreetComplete where I consume the NSI data, but having just tried in OSM via iD I get the same thing there. If I search for PizzaExpress I can add a node like that fine, but "Pizza Express" only finds "Pizza Hut Express".

This initially sounded like a `locationSet` problem, but I tried the same thing you did in the Philippines ([a location where the brand is supposed to be valid](https://location-conflation.com/?locationSet=%7B%22include%22%3A%5B%22ae%22%2C%22cy%22%2C%22es%22%2C%22gb%22%2C%22hk%22%2C%22id%22%2C%22ie%22%2C%22in%22%2C%22kw%22%2C%22mo%22%2C%22ph%22%2C%22sg%22%5D%7D&referrer=nsi)), and "PizzaExpress" doesn't appear in iD unless the space is omitted. The same thing happens with the local hotel brand "RedDoorz" (typing "RedDoorz" brings up the preset but "Red Doorz" doesn't, and attempts to add "red doorz" as a match name are removed by the build script). I've filed an issue for this apparent bug (#10040).

Snowysauce:
I'm closing this pull request for now based on further testing; please follow #10040 for updates on the underlying issue.

PR #10035, opened 2024-10-13
add Stanford University to brands/amenity/university

Shubham0523:
This PR adds **Stanford University** to the `brands/amenity/university` category in the NSI.

- **Wikidata ID**: Q41506
- **Reason for inclusion**: Stanford University is a globally recognized academic institution with a significant presence in various countries through research collaborations, satellite locations, and global academic influence.
- **Meets NSI requirements**: It has a valid Wikidata tag and fulfills the notability threshold due to its widespread impact and recognition.

### Additional Notes
Stanford's inclusion will improve the accuracy and comprehensiveness of university-related tagging in OpenStreetMap, particularly for editors using NSI presets like iD Editor.

### DIFF
--- a/data/brands/amenity/university.json
+++ b/data/brands/amenity/university.json
@@ -78,6 +78,18 @@
         "name": "University of Phoenix",
         "short_name": "UOPX"
       }
+    },
+    {
+      "displayName": "Stanford University",
+      "id": "stanforduniversity-d62529",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "amenity": "university",
+        "brand": "Stanford University",
+        "brand:wikidata": "Q1889100",
+        "name": "Stanford University",
+        "short_name": "Stanford"
+      }
     }
   ]
 }


Shubham0523:
please review the PR. @arch0345

data/brands/amenity/university.json
```diff
@@ -78,6 +78,18 @@
...
       }
+    },
+    {
+      "displayName": "Stanford University",
+      "id": "stanforduniversity-d62529",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "amenity": "university",
+        "brand": "Stanford University",
+        "brand:wikidata": "Q1889100",
```
1ec5, line 89: The universities tagged with `brand=*` are multi-campus for-profit university chains. Stanford has a few campuses besides the central campus, but these aren’t “brands”. Instead, the entry would need to go in the operator tree under data/operator/amenity/university.json.

NSI has [a different threshold of notability](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#notability) than other projects like Wikipedia and Wikidata, because the purpose of NSI is to enforce consistent tagging across a large set of features in OpenStreetMap. Stanford and its satellite facilities are already mapped in OSM, and there aren’t very many of them. I suppose we could still add it to the operator tree, though the layout of that file won’t be able to handle an influx of other single-campus universities.

On the other hand, NSI is currently missing entries for Stanford Medicine, Stanford Health Care, Stanford Health Care Tri-Valley, and Stanford Medicine Partners, all of which are local healthcare brands that would have many more eligible features in OSM.

Shubham0523: Sure, I would be interested in contributing to the missing entries.

Shubham0523: can you guide me as to how exactly I can help

bhousel:
Sorry @Shubham0523 , but we don't add individual colleges or universities to this index - we only add the kind of for-profit "universities” that operate like a branded business in multiple locations, such as DeVry or University of Phoenix. I know this is an unusual concept outside of the USA - the “universities” in the name-suggestion-index are really more like training centers than true universities.

PR #10050, opened 2024-10-15
Add Lova Yatak Turkish furniture brand in the same company as Kelebek

Guria:


Snowysauce:
This entry was added to `data/brands/shop/furniture.json` but is tagged as `"shop": "bed"`. Please change the tag value or move the entry to `data/brands/shop/bed.json`.

PR #10054, opened 2024-10-16
OVS extended to all Europe and added OVS Kids

DnaX:
OVS have at least one shop in every country of Europe.

OVS Kids it's a sub-brand of OVS, no different company, no different website, so I kept the same WIkidata item.

Ps. why do we need an OVS preset only for Georgia?

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -7688,12 +7688,24 @@
       "displayName": "OVS",
       "id": "ovs-8c6a50",
       "locationSet": {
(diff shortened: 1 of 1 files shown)


bhousel:
> OVS have at least one shop in every country of Europe.
> OVS Kids it's a sub-brand of OVS, no different company, no different website, so I kept the same WIkidata item.


Looks good to me, thanks 👍 

> Ps. why do we need an OVS preset only for Georgia?

I don't know - normally we do that if they have a different Wikidata QID or they write their name in a different language, or some other thing about that location is special.

PR #10009, opened 2024-10-05
update BBVA names

gedankenstuecke:
This updates the names of the different BBVA branches in the different countries, as suggested in #4650, to avoid stale suggestions. 

closes #4650

PR #10056, opened 2024-10-17
remove remaining bancomer entry

gedankenstuecke:
As noted by @Cj-Malone in #10009, there was a remaining _Bancomer_ entry left after that PR. This one remains it and instead folds it into the same `matchName` for the now-used `BBVA (México)`.


Cj-Malone:
Thank you

gedankenstuecke:
thanks for catching it!

PR #10057, opened 2024-10-17
Added 4 known retail chains - Monginis, The Sugarr & Spice, Khadim, Arambagh Foodmart

NLBRT:
The 4 brands are:

- Monginis, The Sugarr & Spice - Bakery chains, primarily focusing on pastries, light snacks and other baked goods (similar to Mio Amore, also added by me in my previous PR)
- Khadim - Footwear
- Arambagh Foodmart - Convenience store chain

With the exception of Monginis and Khadims (both of whom have significant presence all over India), The Sugarr & Spice and Arambagh Foodmart has presence primarily in West Bengal/Eastern India.

**Note:** 3 of the websites can be probably scraped for outlet locations.
<sub>P.S: I'm trying to figure out how to add the logos of these brands via facebook. In the meantime, would you folks mind adding it? I'd really appreciate if you could, but I'd probably figure it out by the time I add my next set of brands, or at least by the time this PR goes through, assuming there's no other issues.</sub>


Snowysauce:
Monginis is already in the NSI under [shop/bakery](https://nsi.guide/?t=brands&k=shop&v=bakery#monginis-db079e). If it should be under `shop/pastry`, please merge the existing `shop/bakery` entry with the proposed `shop/pastry` entry. If `shop/bakery` is fine, please merge the proposed `shop/pastry` entry with the existing `shop/bakery` entry.

> P.S: I'm trying to figure out how to add the logos of these brands via facebook.

This is done via the Facebook username property on the Wikidata pages. Once specified, the NSI's Wikidata sync script will automatically grab the Facebook profile picture (usually the logo) on its next run as long as the Facebook profile is public. (If the profile is set to "visible only while logged in", the script will not be able to get the logo.)

NLBRT:
> Monginis is already in the NSI under [shop/bakery](https://nsi.guide/?t=brands&k=shop&v=bakery#monginis-db079e).

Given the similarity of the nature of these businesses, I think it's better to go with `shop=pastry` than `shop=bakery`, which is further solidified by the fact that they advertise more about their cakes and pastries (and sometimes snacks) over their other baked goods. _(Sidenote: Mio Amore was formed in 2015 after Monginis franchises in East India, managed by Switz Foods, decided to split up and go with a different approach)_

As for The Sugarr & Spice,
1. I do realise that the difference in the name (&/And) might appear a bit confusing, but shouldn't be necessarily an issue per se.
2. IIRC, I'm yet to map their outlets present in my city. As for their presence, they most certainly have somewhere between at least 75-100 outlets present in the state and that's excluding transit hubs (railway stations, bus terminals, where they usually have 3/4 outlets each; kind of a bummer that they don't mention their outlets on the website). I hope to map about 4 or 5 of these in the coming week or two.

data/brands/shop/convenience.json
```diff
@@ -550,6 +550,25 @@
...
+      "tags": {
+        "brand": "Arambagh Foodmart",
+        "brand:wikidata": "Q130550508",
+        "name": "Arambagh Foodmart",
+        "name:en": "Arambagh Foodmart",
+        "name:bn": "আরামবাঘ ফুডমার্ট",
+        "shop": "convenience",
+        "air_conditioning": "yes",
+        "payment:cash": "yes",
+        "contact:website":"https://arambaghfoodmart.com/",
```
Snowysauce, line 567: These lines (and similar tags in the other entries) will be automatically removed by the build script if this PR is approved. Their values should be added to the brand's Wikidata page instead.

NLBRT: Got it. Will do.

Snowysauce:
> I do realise that the difference in the name (&/And) might appear a bit confusing, but shouldn't be necessarily an issue per se.

It isn't the most pressing issue with the brand by any means, but it is unusual. Is there a specific reason why the values are different?

NLBRT:
@Snowysauce can we go ahead with this?

Snowysauce:
> I assume using symbols for names is discouraged? If that's not the case, I guess we can revert to using "&" in the `name(:*)` tags as well.

It's not discouraged by any means. It's also not required for the `displayName` to match the `name` or `brand`/`operator`/etc., but in a case where the lone difference between the two is the word "and" and an ampersand, I think it's fair to ask "why" that difference exists.

NLBRT:
I guess the only reason might have been that they have the Bengali name signposted as "দি সুগার অ্যান্ড স্পাইস" (emphasis on "অ্যান্ড"), which literally translates to "And/and" and not "&". However, I went ahead and changed all "And" in the English names (`name`, `name:en` and `brand`) to "&".

Issue #6988, opened 2022-08-31
post_box provided by Deutsche Post have an non-canonical operator preset 


haukepribnow:
### Current behavior

If you create a post_box operated by Deutsche Post with iD, iD sets the following preset values:

- ```amenity=post_box```
- ```operator=Deutsche Post```
- ```operator:wikidata=Q157645```

### What is wrong?

Canonically, ```operator``` should match the one specified in the associated wikidata. So here, ```operator``` should be "Deutsche Post AG".

### Expected behavior

iD should set the following preset values:

- ```amenity=post_box```
- ```operator=Deutsche Post AG```
- ```operator:wikidata=Q157645```

### Other information

See a discussion about this topic (in German) here: https://www.openstreetmap.org/changeset/125505978

bhousel:
Oh yeah - I think most communities prefer to not include the corporation type in this value, but I have noticed that some Germans like to see the "AG" there. 

The latest planet file that we ingested 3 days ago collected these counts for the different values.
https://raw.githubusercontent.com/ideditor/nsi-collector/main/dist/osm/operators_all.json

<img width="396" alt="Screen Shot 2022-09-01 at 10 02 29 AM" src="https://user-images.githubusercontent.com/38784/187933469-a0e27a3c-2bbb-45f7-b43c-3738a0680c35.png">

So it looks like mappers prefer the version _without_ the "AG" about 3:1.. 

I took a look at the OSM Wikipedia page mentioned in that changeset discussion. It looks like the recommendation to switch from "Deutsche Post" to "Deutsche Post AG" was only changed last year:  https://wiki.openstreetmap.org/w/index.php?title=Deutsche_Postdienstleister&type=revision&diff=2118849&oldid=2103550

So that would explain why lots of post boxes were already mapped without the "AG".  The comment on that change really doesn't link to any other discussion about it, so it feels kind of just like one person editing the wiki to suit their preference.

I guess I'd like to see someone from the German community raise this on a more appropriate forum before we make the change in NSI..  A change like this will trigger validation issues on 27,000 post boxes, and sure we can do it, but I'd like some extra confirmation to make sure that someone else isn't going to show up tomorrow and ask us to change them back.

haukepribnow:
Understood. To be honest, I'm a very low frequency editor who's only doing very limited OSM editing as a hobby. So I wouldn't be the right one to trigger such a discussion.

However, independently from what "local preference" is about this topic, there is another thing to consider:

[...]

PR #10059, opened 2024-10-19
pipeline.json legal names

ZEE6de:
Apparently someone shortened the "operator" values of various pipeline companies by removing all kinds of ", LLC; Company;,L. P.; Operating" extensions & so on.
For pipeline companies, at least in the US, we use the full, exact legal name for the operator field, as this avoids ambiguity - for example, many pipelines are operated by shared operating companies with complex parent company structures, where such distinctions are essential.
It appears from that most european companies are also tagged using their proper legal name, so no reason to change that for US companies.
Those automated changes also break all existing mapping, which, in the US uses comprehensively legal names, and furthermore this extremely broad change has not been discussed anywhere, which violates the Automated Edit guidelines.

Greetings

### DIFF
--- a/data/operators/man_made/pipeline.json
+++ b/data/operators/man_made/pipeline.json
@@ -137,11 +137,11 @@
         "include": ["us-tx.geojson"]
       },
       "matchNames": [
-        "air liquide large indust u.s. lp"
+        "air liquide large indusries"
       ],
       "tags": {
         "man_made": "pipeline",
-        "operator": "Air Liquide Large Industries",
+        "operator": "Air Liquide Large Industries U.S. LP",
         "operator:wikidata": "Q113466577"
       }
     },
@@ -220,10 +220,10 @@
       "locationSet": {
         "include": ["us-tx.geojson"]
       },
-      "matchNames": ["amp texas pipelines, llc"],
+      "matchNames": ["amp texas pipelines"],
       "tags": {
         "man_made": "pipeline",
-        "operator": "AMP Texas Pipelines"
+        "operator": "AMP Texas Pipelines, LLC"
       }
     },
     {
@@ -231,11 +231,11 @@
       "id": "anadarkopetroleum-0a386a",
       "locationSet": {"include": ["us"]},
       "matchNames": [
-        "anadarko petroleum corporation"
+        "anadarko petroleum"
       ],
       "tags": {
         "man_made": "pipeline",
-        "operator": "Anadarko Petroleum",
+        "operator": "Anadarko Petroleum Corporation",
         "operator:wikidata": "Q483637"
       }
     },
@@ -274,12 +274,11 @@
       "id": "anrpipeline-0a386a",
       "locationSet": {"include": ["us"]},
       "matchNames": [
-        "anr pipeline co",
-        "anr pipeline company"
+        "anr pipeline"
       ],
       "tags": {
         "man_made": "pipeline",
-        "operator": "ANR Pipeline",
+        "operator": "ANR Pipeline Company",
         "operator:wikidata": "Q4653023"
       }
     },
@@ -298,20 +297,20 @@
       "locationSet": {
         "include": ["us-tx.geojson"]
       },
-      "matchNames": ["arcadia operating, llc"],
+      "matchNames": ["arcadia operating"],
       "tags": {
         "man_made": "pipeline",
-        "operator": "Arcadia Operating"
+        "operator": "Arcadia Operating, LLC"
       }
     },
     {
       "displayName": "Arena Offshore",
       "id": "arenaoffshore-0a386a",
       "locationSet": {"include": ["us"]},
-      "matchNames": ["arena offshore, lp"],
+      "matchNames": ["arena offshore"],
       "tags": {
         "man_made": "pipeline",
-        "operator": "Arena Offshore"
+        "operator": "Arena Offshore, LP"
       }
     },
     {
@@ -321,11 +320,11 @@
         "include": ["us-tx.geojson"]
       },
       "matchNames": [
-        "arrowhead gas services, llc"
+        "arrowhead gas services"
       ],
       "tags": {
         "man_made": "pipeline",
-        "operator": "Arrowhead Gas Services"
+        "operator": "Arrowhead Gas Services, LLC"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


Snowysauce:
> we use the full, exact legal name for the operator field

Names should be the [common name of the entity](https://wiki.openstreetmap.org/wiki/Key:name). Legal names can be stored in an operator's Wikidata.

> Those automated changes also break all existing mapping, which, in the US uses comprehensively legal names, and furthermore this extremely broad change has not been discussed anywhere, which violates the Automated Edit guidelines.

There's absolutely nothing automated about anything related to the NSI. The "S" stands for "suggestion" - it is up to individual editors to decide whether to use a preset or suggestion based on NSI data. Furthermore, many of the pipeline entries don't have `operator:wikidata` tags, meaning the entries and their suggested values aren't exported anyway.

ZEE6de:
> Names should be the [common name of the entity](https://wiki.openstreetmap.org/wiki/Key:name). Legal names can be stored in an operator's Wikidata.

We are talking about https://wiki.openstreetmap.org/wiki/Key:operator, not "name". The most correct "operator" variant is always the legal name, and it is practice to use it for pipelines in OSM.

> Can you point to a discussion where the consensus was to use legal names instead of common names?

(operator tag) I am changing things to the status quo. Where is the discussion to change countless operator values from their current way of mapping, which is the legal name, to apparently carelessly shortened variants? Many such short values which I try to fix here, such as "Kinder Morgan Treating", "Enterprise Products Operating", "Sheridan Production Company III" are simply made-up, and far from "common names". 
Also, those shortenings are very random such as "[ANR Pipeline Company](https://coraweb.sos.la.gov/CommercialSearch/CommercialSearchDetails.aspx?CharterID=111446_6E9CCB067A)" is shorted to "ANR Pipeline", but "[Dixie Pipeline Company LLC](https://coraweb.sos.la.gov//commercialsearch/CommercialSearchDetails.aspx?CharterID=94748_ABBC2F4DC7)" is shortened to only "Dixie Pipeline Company". "Company" is sometimes removed, sometimes left. Same with "Operating". How is this determined?

> The "S" stands for "suggestion" - it is up to individual editors to decide whether to use a preset or suggestion based on NSI data.

In theory yes, in practice no. The ID editor says that it fixes "outdated" tags, of course an inexperienced user is eager to "update" tags when given an easy way of doing so (e.g. https://www.openstreetmap.org/changeset/156804112)

> many of the pipeline entries don't have operator:wikidata tags, meaning the entries and their suggested values aren't exported anyway.

I assume sooner or later they will have wikidata tags, so they should still be correct.

Snowysauce:
> > Can you point to a discussion where the consensus was to use legal names instead of common names?
> 
> (operator tag) I am changing things to the status quo.

No then. Got it.

ZEE6de:
> No then. Got it.

Again, where is the discussion for the changes which I'm undoing here?
They affect _tens of thousands of kilometers of pipeline_, where carefully mapped operator information will be replaced with apparently random made-up values without any discussion whatsoever.

bhousel:
Hey @ZEE6de , I really don't appreciate how you're trying to weaponize this "language of consensus".  

The actual truth of what happened is that in 7efbbd416 @Snowysauce renamed some of the operators from things like "Eagle Ford Midstream, LP" to "Eagle Ford Midstream", and now you want them renamed back.  This is a pretty minor change.

When you say things like "break all existing mapping", " extremely broad change", and "apparently random made-up values", it's very hard for me to take you seriously. 

In fact, the convention on whether to use the full legal name with corporate form  _does_ get discussed sometimes.  
Here's one from a few years ago, that we've never really been able to close out because some people have such strong opinions.  https://github.com/osmlab/name-suggestion-index/issues/6988#issuecomment-1234342397

The data shows that most mappers prefer **not** to include corporate forms like "GmbH" or "LLC" and it's only a few places where a few people insist on keeping these around.  Maybe your pipelines are like this.  After an item in NSI has been linked to a Wikidata QID, it doesn't even matter at all what goes in the `operator` tag - the `operator:wikidata` identifies the operator uniquely - making this an unimportant and tedious thing to keep discussing.

So I feel like - If we do keep these around (I am fine with keeping or removing, I don't care), it's not because it's the right thing to do, and it's not because mappers prefer it, it would just be to appease anonymous trolls like yourself who keep showing up on our GitHub to bother us about this completely unimportant detail.

ZEE6de:
What I don't like is language like "just be to appease anonymous trolls like yourself" to good faith contributors. I have invested a lot of time into purely researching such operators to  be able to tag the most correct one, and now it turns out to be getting replaced with inaccurate stuff. 
If you don't see any significance of that that's totally fine, but then simply leave those values alone and don't call people who do care "anonymous trolls". Because then what's the significance of insisting to _not_ use the current mapping way, the legal name.

> When you say things like "break all existing mapping", " extremely broad change", and "apparently random made-up values", it's very hard for me to take you seriously.

I have explained for all three points why I think like that, none of then are just thrown in the room as insults, and I have not recieved any counter arguments for any of that. 
- It "breaks mapping" because of situations like [this](https://overpass-turbo.eu/s/1SZp)/[this](https://overpass-turbo.eu/s/1SZr) query. Before the NSI changes, there was one operator tag using the legal name, now there are two ways of tagging. Searching using the wikidata tags leaves out hundreds of miles shown in the first query, which lack the wikidata and have a different operator=*/vice versa. And this goes for all operators where replacements have started.
- It is "extremely broad" because it affects hundreds of values, basically every US pipeline.
- The new values _are apparently_ random, as demonstrated above with the "Company" and other extensions being sometimes removed, sometimes left.


> who keep showing up on our GitHub

Who shows up where? Is it really the fault of "trolling", or is it the fault of random NSI tag changes showing up in editors to established taggings? Those changes are the problem, not the people discussing them.

bhousel:
> What I don't like is language like "just be to appease anonymous trolls like yourself" to good faith contributors. I have invested a lot of time into purely researching such operators to be able to tag the most correct one, and now it turns out to be getting replaced with inaccurate stuff.

I'm not sure you are a "good faith contributor"..  You have no history of contributing anything on Github, this is your first comment here, and it's likely your account is just another alt for Hiausirg / Zluuzki, who has been blocked from OSM several times.

Everything else you said is just wrong.  It doesn't "break mapping" just because it broke your Overpass query.  A change that affects a few dozen pipelines isn't "extremely broad" or "basically every US pipeline".  The new values are not "apparently random" in any way - they are just the same thing as before without the "LP" on the end.  

Anyway.. I'm surprised I've spent even this much time replying to you.   We can end this now.

PR #10065, opened 2024-10-20
Add MDTA AET

cicku:
_Unsure about the display name._

### DIFF
--- a/data/operators/highway/toll_gantry.json
+++ b/data/operators/highway/toll_gantry.json
@@ -63,6 +63,18 @@
       }
     },
     {
+      "displayName": "Maryland Transportation Authority All-Electronic (Cashless) Tolling",
+      "locationSet": {
+        "include": ["us-md.geojson"]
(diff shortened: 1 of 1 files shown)


Snowysauce:
IMO the proposed `displayName` would only be necessary if MDTA had different "networks" (so to speak) of toll gantries. If all MDTA toll gantries are identical, the name of the authority is probably sufficient. (If MDTA has a mix of gantries and booths, and presets exist for both, iD and JOSM both make it clear to the end user which preset is which.)

Truthfully though, the value that is used for `displayName` doesn't really matter for OSM itself; AFAIK it's only used in the NSI for indexing and in some OSM editors as a preset label. (On the latter point, though, unless the editor has some mechanism to display long `displayName`s, there's a chance that a name such as this will be truncated for display purposes, somewhat negating the purpose of the detailed name.)

cicku:
All identical. Name updated

PR #10067, opened 2024-10-21
Add Enterprise GO

### LINKED ISSUE #10066: [CH] Add Enterprise GO
https://enterprise-go.ch is an 'office-less' car rental in Switzerland.
See https://community.openstreetmap.org/t/trying-to-bulk-import-locations/119788/ for details on an initial import of +100 stations.

habi:
The intention is to help keep the initial import of the 'stations' up to date, see: https://community.openstreetmap.org/t/trying-to-bulk-import-locations/119788/30

Fixes #10066

### DIFF
--- a/data/brands/amenity/car_rental.json
+++ b/data/brands/amenity/car_rental.json
@@ -185,6 +185,19 @@
       }
     },
     {
+      "displayName": "Enterprise GO",
+      "locationSet": {"include": ["ch"]},
+      "matchNames": ["enterprise car rental"],
+      "tags": {
+        "amenity": "car_rental",
+        "brand": "Enterprise GO",
+        "brand:wikidata": "Q130467239",
+        "operator": "Helvetic Motion AG",
+        "supervised": "no",
+        "website": "https://enterprise-go.ch/"
+      }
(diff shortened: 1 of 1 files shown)


Snowysauce:
I see there's still some ongoing discussion about the import, but unless I'm glossing over something, it seems to be mostly about details outside the scope of the NSI. Thus, the proposed PR looks fine to me. (I understand one of the points is what website to use, and that this PR includes a `website` tag; however, the build script will remove the `website` tag on its next run anyway - please see [Item Property Reference#tags](https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#tags) for more info.)

habi:
> I see there's still some ongoing discussion about the import, but unless I'm glossing over something, it seems to be mostly about details outside the scope of the NSI. Thus, the proposed PR looks fine to me. 

The import is done,  'we' are debating about the `website`-Tag to use.

> (I understand one of the points is what website to use, and that this PR includes a `website` tag; however, the build script will remove the `website` tag on its next run anyway [...]

I didn't know that, thanks for clarifying.

PR #10072, opened 2024-10-24
Remove Mobile Money brand from money_transfer.json

sammyhawkrad:
[Mobile Money](https://en.wikipedia.org/wiki/Mobile_Money) is not a brand but a technology or service type. There are several mobile money brands like [Orange Money](https://en.wikipedia.org/wiki/Orange_Money) (already in NSI), MTN Mobile Money ([MoMo](https://mtn.com.gh/momo/)) , [M-Pesa](https://en.wikipedia.org/wiki/M-Pesa) etc. and I believe this generic one shouldn't be in the NSI.

### DIFF
--- a/data/brands/amenity/money_transfer.json
+++ b/data/brands/amenity/money_transfer.json
@@ -45,21 +45,6 @@
       }
     },
     {
-      "displayName": "Mobile Money",
-      "id": "mobilemoney-1032f8",
-      "locationSet": {"include": ["001"]},
-      "matchNames": [
-        "mobil money",
-        "mobile money transfer"
-      ],
-      "tags": {
-        "amenity": "money_transfer",
-        "brand": "Mobile Money",
-        "brand:wikidata": "Q20312600",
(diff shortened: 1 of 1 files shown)


Snowysauce:
I read the wiki article, and I agree with the substance of the request. However, as an FYI, the sheer number of OSM items using "Mobile Money" as a name ([literally over 1000](https://overpass-turbo.eu/s/1TrS)) means we'll have to add that name to the generic name array post merge to keep it from being readded to the category file by the build script. This will, in turn, generate a _lot_ of "generic name" warnings in iD whenever the next NSI release is published.

I'm not sure if you'd like to proceed with the request as-is right now, or gradually change the OSM objects to refer to different brands over time before pulling the "Mobile Money" entry.

sammyhawkrad:
I think the changes should be applied right away. People will continue to add the "Mobile Money" brand to objects as long as it's an option. It's better to have warnings that can be fixed than to keep creating wrong data.

PR #10082, opened 2024-10-25
Update surveillance.json

JoeyKentin:
Updated the name to safe.brussels

### DIFF
--- a/data/operators/man_made/surveillance.json
+++ b/data/operators/man_made/surveillance.json
@@ -492,25 +492,22 @@
       }
     },
     {
-      "displayName": "Bruxelles Prévention & Sécurité (BPS) – Brussel Preventie & Veiligheid (BPV)",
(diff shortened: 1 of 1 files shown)


Snowysauce:
The build tests failed due to the edited entries having the same `displayName`. If the two entries belong to the same entity, please merge them together. Also, please add the old names of each entry to the relevant `matchNames` array(s).

UKChris-osm:
The first entry doesn't have a `wikidata` attached, so isn't appearing in the index - perhaps we could just delete that entry?

Snowysauce:
Since the OP didn't make any changes to the PR, I figured it was easier to address the issue in a new commit (https://github.com/osmlab/name-suggestion-index/commit/c18afce0952b6f3c166de0f0ad006c96cc223547) than to merge this PR as-is and have to clean up the resulting code.

PR #10119, opened 2024-11-03
Add Dome cafes

BudgieInWA:
Should I add something to aid with users searching for "dome" (without the decorated "o")?

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -2944,7 +2944,7 @@
     },
     {
       "displayName": "Dôme",
-      "locationSet": {"include": ["au"]},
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Hi @BudgieInWA 

You could add a `"matchNames": ["dome"],` which would help the matcher, which would be under the `locationSet` tag.

Also, if the cafes only appear in Western Australia, you can use `au-wa.geojson` as the value for `locationSet`.

BudgieInWA:
Thanks for that info.

Cj-Malone:
I'm pretty sure we don't need to add match names for fancy letters. Won't hurt though. Thanks.

PR #10121, opened 2024-11-03
Update RTC Bike Share and add state for Indego

quincylvania:
The bike share system in Las Vegas is branded as "RTC Bike Share". RTC of Southern Nevada is the name of the transit agency that owns it. RTC Bike Share is operated by Bicycle Transit Systems. BCycle is just the supplier. (BCycle is actually being acquired by BTS but it's not clear if they will rebrand).


UKChris-osm:
Thanks @quincylvania 👍

PR #10124, opened 2024-11-04
Update bus.json

mgageo:
update of bus networks in France/Bretagne

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -4296,11 +4296,11 @@
       }
     },
     {
-      "displayName": "CTRL",
+      "displayName": "IZILO",
       "id": "ctrl-add5eb",
       "locationSet": {"include": ["fx"]},
       "tags": {
-        "network": "CTRL",
+        "network": "IZILO",
         "network:wikidata": "Q2408086",
         "route": "bus"
       }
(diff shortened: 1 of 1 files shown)


Snowysauce:
Thanks for the submission. The update from CTRL to IZILO looks fine, but the suggested change to STAR Rennes is a bit unusual. It used to be common practice on OSM to prepend a location's abbreviation to the network name, but this appears to be deprecated: looking back through the history of `bus.json`, many network names have moved away from that convention, and STAR itself has had a `matchNames` value of `fr:star` to deprecate that name since commit https://github.com/osmlab/name-suggestion-index/commit/9f41e1cb8d2d89579e1c54080f4a83705dd7cf65 in March 2021.

mgageo:
The wiki (https://wiki.openstreetmap.org/wiki/Rennes/Transports_en_commun) has defined ‘FR:STAR’ for the network tag for many years. This distinguishes it from the Roanne/France ‘STAR’ network.

Snowysauce:
The change to CTRL has been made as requested in https://github.com/osmlab/name-suggestion-index/commit/8311ad89121be80eb14e6ab38a39fa5468a69bb2; the change to STAR has not. Modern OSM data consumers separate operators by the value of `operator:wikidata`, not `operator`, rendering the "FR:" prefix unnecessary.

PR #10131, opened 2024-11-09
easybox parcel locker: merge countries, remove operator & brand, add payment info

gy-mate:
This PR fixes _easybox_ parcel locker data.

- Split HU & RO variants
- Fix operators (_Sameday_ is just a brand, not the actual operator)
- Add tags to HU variant


UKChris-osm:
Thanks @gy-mate 

We generally don't include `contact:` or `payment:` tagging within the NSI however.

gy-mate:
@UKChris-osm Thank you for the quick reaction!

> We generally don't include `contact:` or `payment:` tagging within the NSI however.

I've removed them and [added this info to the wiki](https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference/_compare/99c641a8261eb9ffe8176a9b78c8dd2610870787...e19f428c11fa8d2252aff83b688b9d5b6b212194).

Dimitar5555:
I don't think that OSM likes having company types (EAD, Zrt, or S.A.) in operator fields (it may depend on the local community though). Also why does the Hungarian entry have `operator:addr` and company numbers tags? Is it common to place such tags there?

gy-mate:
> I don't think that OSM likes having company types (EAD, Zrt, or S.A.) in operator fields (it may depend on the local community though).

@Dimitar5555 The local community agrees on this, I've just asked them about it ([1](https://wiki.openstreetmap.org/wiki/Hungary/Jelölési_példák#Csomagautomata), [2](https://matrix.to/#/!gAwUAKJoCAuYnIoNeV:grin.hu/$lIBoEkfBj3n53lGbSpHe0bS5U3Gm9TW8KepXI93WL2A?via=grin.hu&via=matrix.org&via=cwt.grin.hu)). `ref:HU:company=*` and `ref:HU:vatin=*` is also _tied_ to the company name **and** type.

> Also why does the Hungarian entry have `operator:addr` and company numbers tags? Is it common to place such tags there?

Yes, it is. See [`operator:addr=*`](https://overpass-turbo.eu/s/1TT2), [`ref:HU:company=*`](https://overpass-turbo.eu/s/1TT3) and [`ref:HU:vatin=*`](https://overpass-turbo.eu/s/1TT4) on Overpass Turbo.

UKChris-osm:
I'd probably remove the _operator:addr_ and _ref_ tags as well from this index, as while it may seem handy having all these tags automatically added from within the NSI, I feel like it's out of scope for what we're trying to maintain, which is only really a canonical index of name / brand / operator.

It can also cause problems later on if an address or reference changes. If that information is available within the Wikidata article for each entry, then updating it there would make it available to those who need it.

Snowysauce:
Apart from the discussion about `payment:*`, I agree with everything that's been said by the other maintainers. The `operator:addr` and `ref:*` tags are a maintenance nightmare waiting to happen, and they should be excluded from the NSI for the same reasons why we don't include `contact:*` tags.

In the case of `payment:*`, those tags have [historically been allowed on a limited, case-by-case basis](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#adding-specific-keys).

UKChris-osm:
I hadn't realised `payment:` tagging was allowed in certain circumstances.

Personally I feel like that's an extra layer of maintenance we shouldn't be burdening ourselves with.

bhousel:
Oh yeah, my thought on this is that `payment` tags probably make sense for some limited kinds of POIs, like vending machines, or transit ticket machines, and it's probably fine for the parcel lockers too, and probably some other things I am not thinking about. 

I mostly want to avoid something like people adding `payment:cash=yes` to every fast food brand world wide. While technically correct, it doesn't help anyone and would just be annoying and generate millions of tag upgrade warnings.   There are a lot of tags on https://wiki.openstreetmap.org/wiki/Key:payment like this.

UKChris-osm:
> Oh yeah, my thought on this is that `payment` tags probably make sense for some limited kinds of POIs, like vending machines, or transit ticket machines, and it's probably fine for the parcel lockers too, and probably some other things I am not thinking about.

That makes sense regarding vending and ticket machines. I didn't actually think parcel lockers took payments directly though, I thought it was all done online or via an app?

gy-mate:
> I'd probably remove the _operator:addr_ and _ref_ tags as well from this index

@UKChris-osm @Snowysauce No problem, I've removed them! The wiki [has been updated](https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference/_compare/8e7e5cdc44d7514cd783f5287042116b669ca4e5...349ceb7f7ffed974f7dd40703036d67dc2fed9ba#diff-e2296f3f5be69877e661f88b40657c9260a13705b385d7c654f76712ca06ececL278-R287) as well.

gy-mate:
> I didn't actually think parcel lockers took payments directly though, I thought it was all done online or via an app?

@UKChris-osm That's the case at [Z-Boxes](https://nsi.guide/index.html?t=brands&k=amenity&v=parcel_locker&tt=z-box#zbox-5a86e9):

A red and white Packeta Z-BOX locker outdoors, surrounded by grass and trees. The locker is designed for parcel delivery and pickup.

But most other types of parcel lockers in Hungary _do_ have a POS terminal:

A blue GLS package locker with multiple compartments and digital screens, prominently featuring the number 2500 and directional text. The locker is located indoors with bright overhead lighting.

A large locker system labeled "alzaBox" with promotional graphics on its doors. It features a black Garmin Venu 2 watch advertisement on the front. The locker has a green interactive screen and keypad on one section. There is graffiti on the wall to the left and a sign labeled "easybox" on the right.

A package locker system with a colorful design, featuring the Magyar Posta logo, a touchscreen interface, and illustrated maps with pinpoint markers. A woman holds a parcel on the left side of the image.

A red and blue eMAG easybox parcel locker is positioned on a sidewalk, featuring multiple compartments and a digital interface for package pickup. A car is visible in the background.

data/brands/amenity/parcel_locker.json
```diff
@@ -410,8 +424,8 @@
         "amenity": "parcel_locker",
         "brand": "easybox",
         "brand:wikidata": "Q114496224",
-        "operator": "Sameday",
-        "operator:wikidata": "Q114417096",
+        "operator": "Delivery Solutions EOOD.",
```
Dimitar5555, line 427: Small change to match the spelling, according to the [company registar](https://portal.registryagency.bg/CR/en/Reports/ActiveConditionTabResult?uic=206957949).

```suggestion
        "operator": "Деливъри Солюшънс",
```

gy-mate: @Dimitar5555 Thank you!

The full legal name of the company would be _Деливъри Солюшънс ЕООД_ then, right?

Dimitar5555: Yes, but we avoid company types in the operator field.

Snowysauce:
I agree with @UKChris-osm; given all the discussion above, I'd recommend:

- dropping all of the operator tags from the preset
- making a Wikidata item for the operator (if one doesn't exist) and placing all the name variations (and address) there
- linking that page to the brand Wikidata page and vice versa (via the "operator"/"item operated" properties)

This way, the info in all its various forms is available from OSM to whoever wants it (via `brand:wikidata` and the links on its Wikidata page) without leading to prolonged discussions on OSM or here.

DnaX:
When i first created the preset, I added Sameday as operator just because some of those parcel lockers where sponsored by eMAG e-shop, so lots of people tagged as eMAG owned parcels (like Amazon Hubs).

BTW... ok to remove the operator tags.

sammyhawkrad:
>"what if the operator changes, or the brand is sold, what if the brand expands via another operator in another country", etc.

Then the entries are updated here. Does this not happen sometimes?

gy-mate:
> In the case of `payment:*`, those tags have [historically been allowed on a limited, case-by-case basis](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#adding-specific-keys).

> `payment` tags probably make sense for some limited kinds of POIs, like vending machines, or transit ticket machines, and it's probably fine for the parcel lockers too

I've put `payment:*=*` tags back in the list.

gy-mate:
> I'd recommend:
> 
> * dropping all of the operator tags from the preset

@UKChris-osm @Snowysauce @DnaX Done...

> * making a Wikidata item for the operator (if one doesn't exist) and placing all the name variations (and address) there

...done:

- [Delivery Solutions Zrt. (Q131107197)](https://www.wikidata.org/wiki/Q131107197)
- [S.C. Delivery Solutions S.A. (Q114417096)](https://www.wikidata.org/wiki/Q114417096)
- [Delivery Solutions EOOD. (Q131107345)](https://www.wikidata.org/wiki/Q131107345)

> * linking that page to the brand Wikidata page and vice versa (via the "operator"/"item operated" properties)

...and done (I hope I edited them correctly):

- [easybox (Q114496224)](https://www.wikidata.org/wiki/Q114496224)
- [Sameday (Q131107307)](https://www.wikidata.org/wiki/Q131107307)

gy-mate:
> Then the entries are updated here. Does this not happen sometimes?

@sammyhawkrad I also support the idea of not duplicating things in order to make them easier to maintain.

data/brands/amenity/parcel_locker.json
gy-mate: I've removed this tag completely due to other reasons: https://github.com/osmlab/name-suggestion-index/pull/10131#issuecomment-2480687657

UKChris-osm:
As this has now become quite a long thread, can I just confirm the aim of this PR:

Currently, easybox is listed twice in the NSI, one for Bulgaria and one for Romania & Hungary, with both using the same Wikidata number.

The PR splits Romania & Hungary into a separate entry, so the NSI has 3 entries, all using the same Wikidata?

CrowOk9830:
> Fix operators (Sameday is just a brand, not the actual operator)

Sameday is a well known international courier service. Iit's a company group as stated on the about us page of their website.

In a survey, mappers will see Sameday on the parcel locker (the blue sticker shown previously is an older version; refer to pictures on the main page of their website or on their delivery vehicles), which clearly assigns some importance to that word. We should map what's on the ground mainly. (If Sameday is not the operator, what is that, how it should be tagged, or is it unimportant?)

On the other hand, the country-specific subsidiary company is not presented on the machine (except on the receipt maybe, if printed at all) and not a well known operator name. Thus changing operator to that feels some opinionated nitpicking, and ambiguity can lead to many edit wars.

The OSM wiki for [Tag:amenity=parcel_locker](https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dparcel_locker) also encourages using only the main operator name, such as `operator=DPD`, `operator=DHL`, `operator=Amazon` etc., not the local subsidiary with company type appended. Perhaps it`s worth discussing operator tagging practice more widely and doing changes in sync with OSM wiki.

From the point of view of NSI I can agree to keep operator=Sameday or remove it but not to change.

(Also note that Packate's Z-BOX is a similar case, where the brand of the parcel locker is Z-BOX, the operator is the Packeta courier service company group. I'm note sure `brand=Packeta` and `name=Z-BOX` is ok in terms of naming, because it applies the same name to every Packeta operated parcel locker (compared to the AlzaBox). Also see the OSM wiki about parcel locker names.)

Snowysauce:
I think the back and forth and strong opinions in this thread underscores why NSI shouldn't be in the `operator` tag business when it comes to the brand tree. The guidance on the OSM wiki for "key:operator" allows for both the legal name and the trade name (i.e., branding) to be used, as long as one or the other is used consistently. This freedom to to choose a tagging method is fine for OSM as a whole but it presents a problem for projects like the NSI, where we're expected to pick one or the other as the "canonical" choice in cases where contributors submit `operator` tags with a `brand`. That inevitably leads to battles over what that canonical choice should be, and the whole ordeal has nothing to do with consistent _brand_ tagging in the most basic sense: `brand`, `brand:wikidata`, usually but not always `name`.

The emergence of Wikidata and the growing use of `*:wikidata` tags allows for an environment where things like operators or owners of a non-franchised `brand` can be stored and updated in one place, and still be accessed from OSM, without having to add or modify hundreds or thousands of OSM tags and objects. It was suggested earlier in the thread that NSI should keep tabs on `operator` data for brands, and just simply update the presets if something changes - but that overlooks a couple facts. There is a delay between when changes are made to the NSI code and when those changes go live in editors as presets, and even once the changes are live, _someone still has to decide to make the suggested changes to objects in OSM_. Absent someone doing that, there will be OSM objects with outdated data for months or years, and the blame for that will be pointed at the NSI - even though indexing `brand` operators is not part of our goal.

gy-mate:
@Snowysauce I totally agree. Wikidata is great as a central database for all kinds of data, especially those that change more often than brand names. One `*:wikidata=*` tag can create a link to a _lot_ of data about an element.

gy-mate:
> The PR splits Romania & Hungary into a separate entry, so the NSI has 3 entries, all using the same Wikidata?

@UKChris-osm That's a good point, they should be merged.

gy-mate:
```json
"payment:cash": "no",
"payment:contactless": "yes",
"payment:maestro": "yes",
"payment:mastercard": "yes",
"payment:mastercard_contactless": "yes",
"payment:mastercard_electronic": "yes",
"payment:visa": "yes",
"payment:visa_electron": "yes"
```

@DnaX @hackingorilla95 Can you please confirm that **Romanian** easyboxes have the above payment options (and only these)?

@Dimitar5555 Can you please confirm that **Bulgarian** easyboxes have the above payment options (and only these)?

Dimitar5555:
> @Dimitar5555 Can you please confirm that **Bulgarian** easyboxes have the above payment options (and only these)?

Personally I haven't used their services. Their website mentions that some parcel lockers don't have POS terminals, so it would be better to exclude such data from NSI.

From https://sameday.bg/pravila-i-usloviya-za-predostavyane-na-n/:

> Плащането на услугите се извършва чрез шкафчето
>  Услугите с наложен платеж могат да се заплащат само с банкова карта, през ПОС терминал на easybox, преди пратката да бъде взета реално от шкафа.
>  Някои шкафчета не позволяват плащане при доставка на колета с банкова карта.

> Payment of services performed via lockers
>  The cash on delivery services can only be paid using a bank card, through the easybox POS terminal, before the actual picking up of the parcel from the locker.
>  Some lockers do not allow payment upon delivery of the parcel by bank card.

DnaX:
@gy-mate About Romanian easyboxes I know you can pay with "cards", but not clear which circuit. Here we are using only Visa an MasterCard. So I think it's right the proposed payments tagging.

hackingorilla95:
Yes, thats those are the available paying options for Romania easyboxes.

There are also a few where you could pay with crypto like skrill or btc.

Sorry for the late response, feel free to ask me anything

Kudos~

gy-mate:
@Dimitar5555 @DnaX @hackingorilla95 Thank you for the info! :)

gy-mate:
> Payment of services performed via lockers
>  The cash on delivery services can only be paid using a bank card, through the easybox POS terminal, before the actual picking up of the parcel from the locker.
>  Some lockers do not allow payment upon delivery of the parcel by bank card.

@Dimitar5555 The Hungarian version of this website states the same but I've [asked](https://matrix.to/#/!gAwUAKJoCAuYnIoNeV:grin.hu/$3dGcvzrR31jq1hBZySXUq2hc2NVHdPRTMjRvGjTj7xY?via=grin.hu&via=matrix.org&via=cwt.grin.hu) fellow members of the local OSM community and no one has seen an easybox without a POS terminal. I'm pretty sure this is just a legal defence for the case when a POS terminal of an easybox is temporarily out of order.

Furthermore, https://sameday.bg/frequent-questions/easybox-dostavka/ says in the [_Easybox payment is not accepted. How can I resolve this?_](https://sameday.bg/frequent-questions/easybox-dostavka/#q3) point:

> To successfully process payments in Easybox, make sure that:
> • Your card balance covers the amount you need to pay.
> • If you want to pay with a different card than the one you tried initially, we recommend restarting the parcel collection process.
> • You have activated the contactless payment option.
> • **You hold the card near the payment device (POS)** for more than 5 seconds. You can also try using the insert card option if you are unable to make a contactless payment.

This means that Bulgarian easyboxes _do_ have POS terminals.

Dimitar5555:
> This means that Bulgarian easyboxes do have POS terminals.

That's what I've said. The question is whether all of them have.

Their location finder contains data about whether payment is allowed. Looking at https://sameday.bg/wp/wp-admin/admin-ajax.php?action=get_ooh_lockers_request&country=Bulgaria, there are 15 locations with `"supportedPayment":0` of 950 total locations.

Whether these 15 locations don't have POS terminal or it is temporarily out of order is unknown to me. What I do know is that a lot more of their locations didn't have POS terminals (which an older version of their website showed). I assume that we can assume that most of them have, so it should be okay.

gy-mate:
> so it should be okay.

@Dimitar5555 Thank you for the feedback! Could you please mark the PR as approved?

@UKChris-osm I think we've resolved all issues.

Dimitar5555 (approved):

UKChris-osm:
Thanks @gy-mate 👍

gy-mate:
@UKChris-osm Thank you! :)

gy-mate:
@Snowysauce Could you please add [these policies](https://github.com/osmlab/name-suggestion-index/pull/10131#issuecomment-2483691414) to the wiki so that future contributors (incl. me) can find them easily?

Snowysauce:
@gy-mate I [added a line](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case/_compare/d889cebe3b11b31733c86d8a33a67707d7449d38...2a05801212629ed9321f536cc235c7fb079781fc) to the tags table on the Judge Case page that links to this PR and to two other relevant PRs I remembered as I was making the change (one tangentially related, one directly related).

UKChris-osm:
@gy-mate this entry is now live and available within iD :)

gy-mate:
> I [added a line](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case/_compare/d889cebe3b11b31733c86d8a33a67707d7449d38...2a05801212629ed9321f536cc235c7fb079781fc) to the tags table on the Judge Case page that links to this PR and to two other relevant PRs I remembered as I was making the change (one tangentially related, one directly related).

@Snowysauce Thank you very much! :)

gy-mate:
> this entry is now live and available within iD :)

@UKChris-osm That was quick! Thanks! :)

PR #10132, opened 2024-11-09
Preserve `name=*` tags of CBA & Reál convenience stores

gy-mate:
The `name=*` tags of _CBA_ & _Reál_ convenience stores should be preserved, as these shops usually have a unique name by which they are commonly known.

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -1093,13 +1093,26 @@
       "displayName": "CBA",
       "id": "cba-a1e3fb",
       "locationSet": {
-        "include": ["bg", "hu", "ro", "sk"]
+        "include": ["hu", "ro", "sk"]
       },
       "preserveTags": ["^name"],
       "tags": {
         "brand": "CBA",
         "brand:wikidata": "Q779845",
         "name": "CBA",
+        "shop": "convenience"
+      }
+    },
+    {
+      "displayName": "CBA (България)",
+      "locationSet": {"include": ["bg"]},
+      "tags": {
+        "brand": "CBA",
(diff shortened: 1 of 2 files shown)


UKChris-osm:
When you say "_smaller_" shops, are you still talking about supermarkets, or the "_convenience stores_"?

I've also noticed the supermarket in Bulgaria is a separate entry from the rest, but this isn't the case for "_convenience stores_", do you think this should be changed as well?

gy-mate:
> When you say "_smaller_" shops, are you still talking about supermarkets, or the "_convenience stores_"?

@UKChris-osm My bad, I meant _convenience stores_. Fixed the title, description and the content of the PR. (Thank you for the question!)

gy-mate:
> I've also noticed the supermarket in Bulgaria is a separate entry from the rest, but this isn't the case for "_convenience stores_", do you think this should be changed as well?

Unfortunately I'm not familiar with Bulgarian _CBA_ stores.

Dimitar5555:
> I've also noticed the supermarket in Bulgaria is a separate entry from the rest, but this isn't the case for "_convenience stores_", do you think this should be changed as well?

Yes it should be a separate entry without `preserveTags`. They are named "CBA". Any other local names ([example](https://www.openstreetmap.org/node/3130696829)) are related to the branch.

gy-mate:
@Dimitar5555 Thank you for the info!

Lke this?

Dimitar5555:
Yes, exactly like that. Thanks!

gy-mate:
@UKChris-osm @Snowysauce I think this is ready for a final review.

PR #10135, opened 2024-11-10
[shop/convenience] add Lukoil

dan13l9:
added Lukoil convenience store brand for non-Cyrillic countries

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -3102,7 +3102,7 @@
       "tags": {
         "brand": "Lukoil",
         "brand:wikidata": "Q329347",
-        "name": "Lukoil"
+        "name": "Lukoil",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
How does [Lukoil Mini-Mart](https://nsi.guide/index.html?t=brands&k=shop&v=convenience&tt=Lukoil#minimart-f1e40b) factor into this, can it be removed or merged with the new addition?

Snowysauce:
Merging as overpass-turbo indicates that some kind of non-Cyrillic language option is needed. I'm not familiar with Lukoil Mini-Mart despite that brand being geolocated to the US, and as such I'll let it co-exist with a global "Lukoil" preset. I suppose time will tell if the Mini-Mart preset is genuinely needed.

PR #10137, opened 2024-11-12
Update ticket_validator.json - change to WSW mobil GmbH

Elefant-aus-Wuppertal:
The WSW operator should be written as WSW mobil GmbH, as it is also so for other German ticket validator operators like DB Vetrieb GmbH and Transdev Vertrieb GmbH.

### DIFF
--- a/data/operators/amenity/ticket_validator.json
+++ b/data/operators/amenity/ticket_validator.json
@@ -296,13 +296,13 @@
       }
     },
     {
-      "displayName": "WSW mobil",
+      "displayName": "WSW mobil GmbH",
       "id": "wswmobil-f980d0",
       "locationSet": {"include": ["de"]},
-      "matchNames": ["wsw mobil gmbh"],
+      "matchNames": ["wsw mobil"],
       "tags": {
         "amenity": "ticket_validator",
-        "operator": "WSW mobil",
+        "operator": "WSW mobil GmbH",
         "operator:wikidata": "Q44880072"
       }
     },


UKChris-osm:
What about [WSW](https://nsi.guide/index.html?t=operators&k=amenity&v=ticket_validator&tt=wsw#wsw-08753d)? Can these be merged, or one deleted?

Elefant-aus-Wuppertal:
Hi,

thank you for the question.

"WSW" is actually a duplicate (but with the wrong Wikidata).

The Wikidata there belongs to the umbrella organization, which is also called WSW. However, the operator is always the
WSW mobil GmbH.

Therefore, it would be good if the NSI would suggest retagging the "WSW" ticket validators to "WSW mobil GmbH".

Snowysauce:
If the name is changed, WSW's entry in [operators/emergency/phone.json](https://github.com/osmlab/name-suggestion-index/blob/1a89ae1ebef48b9a416f1ba49f00d12769378f57/data/operators/emergency/phone.json#L1091) should probably be changed as well for consistency.

Elefant-aus-Wuppertal:
> If the name is changed, WSW's entry in [operators/emergency/phone.json](https://github.com/osmlab/name-suggestion-index/blob/1a89ae1ebef48b9a416f1ba49f00d12769378f57/data/operators/emergency/phone.json#L1091) should probably be changed as well for consistency.

Oh yes, of course, true. Thank you for noticing.

I changes it in PR https://github.com/osmlab/name-suggestion-index/pull/10145

PR #10141, opened 2024-11-13
Add WSW mobil GmbH as a public transport ticket vending machine operator

Elefant-aus-Wuppertal:
Hi,

the operator WSW mobil GmbH does not only operates ticket validators, but also ticket vending machines. Until now it was not part in the NSI. Thank you.

### DIFF
--- a/data/operators/amenity/vending_machine.json
+++ b/data/operators/amenity/vending_machine.json
@@ -190,6 +190,21 @@
         "vending": "public_transport_tickets"
       }
     },
+        {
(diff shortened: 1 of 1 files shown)


Elefant-aus-Wuppertal:
This PR is being put on hold for now. I have just found out during an on-site inspection that there are also public transport ticket vending machines that have **two** operators - WSW mobil GmbH and simultaneously another one. I need to see what we do with them before adding this new operator.

But this only applies to the ticket machines. At the emrgency phones and the ticket validators there isn't this issue.

PR #10142, opened 2024-11-13
Update hotel.json

jhutton1:
Wyndham bought a number of hotels and added by Wyndham to end.  Updated official name of those where it had not been updated yet. 
[Article](https://www.travelmole.com/news/wyndham-adds-its-name-to-a-dozen-brands/?region=) and other research confirm this. 

It seems as though the brand name is actually changing for all of these hotels.  I am unsure on how to make this decision.  Should brand name be updated as well to include 'by Wyndham' at the end?  Or just updating official name as done in this PR.

### DIFF
--- a/data/brands/tourism/hotel.json
+++ b/data/brands/tourism/hotel.json
@@ -987,6 +987,7 @@
         "brand": "Days Inn",
         "brand:wikidata": "Q1047239",
         "name": "Days Inn",
+        "official_name": "Days Inn by Wyndham",
         "tourism": "hotel"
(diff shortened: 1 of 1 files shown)


UKChris-osm:
> It seems as though the brand name is actually changing for all of these hotels. I am unsure on how to make this decision. Should brand name be updated as well to include 'by Wyndham' at the end? Or just updating official name as done in this PR.

I think at some point we'll need to update the brand(s) but no rush, we'll see how this plays out.

PR #10165, opened 2024-11-22
Add Brigham Young University to operator/amentity/university

penguin1737:


bhousel:
Sorry @penguin1737 , but we don't add individual colleges or universities to this index - we only add the kind of for-profit "universities” that operate like a branded business in multiple locations, such as DeVry or University of Phoenix. The “universities” in the name-suggestion-index are really more like training centers than true universities.

This index is really for tracking the types of things where there are a lot of them, not just a few things (there are only 3 campuses for BYU). We are considering adding some university operated assets in #4779

PR #10172, opened 2024-11-25
Update parcel_locker.json

krystian-r:
Rebranding for DHL parcel machines (in Poland) - DHL POP BOX (not DHL Pop Box, which is incorrect) to DHL BOX 24/7


krystian-r:
Hello, @Snowysauce Will that update be merged in parcel_locker.json file? I made it a month ago, but no action from Your side.

PR #10187, opened 2024-11-29
Update fuel.json

BPowell76:
Add additional information for gas stations

## BP
- Add `brand:website=https://www.bp.com/`
- Add `self_service=yes`

## Marathon
- Add `brand:website=https://www.marathonfuel.com`
- Add `payment` tags
  - `payment:cash=yes`
  - `payment:contactless=yes`
  - `payment:credit_cards=yes`
  - `payment:debit_cards=yes`
- Add `self_service=yes`

## Sheetz
- Add `brand:website=https://www.sheetz.com/`
- Add `self_service=yes`

## Shell
- Add `brand:website=https://www.shell.com/`
- Add `self_service=yes`

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -5507,7 +5507,9 @@
         "amenity": "fuel",
         "brand": "Sheetz",
         "brand:wikidata": "Q7492551",
-        "name": "Sheetz"
+        "name": "Sheetz",
+        "brand:website": "https://www.sheetz.com/",
+        "self_service": "yes"
       }
     },
     {
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -5532,7 +5532,9 @@
         "amenity": "fuel",
         "brand": "Shell",
         "brand:wikidata": "Q110716465",
-        "name": "Shell"
+        "name": "Shell",
+        "brand:website": "https://www.shell.com",
+        "self_service": "yes"
       }
     },
     {
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -3815,7 +3815,13 @@
         "amenity": "fuel",
         "brand": "Marathon",
         "brand:wikidata": "Q458363",
-        "name": "Marathon"
+        "name": "Marathon",
+        "website": "https://www.marathonfuel.com/",
+        "payment:cash": "yes",
+        "payment:contactless": "yes",
+        "payment:credit_cards": "yes",
+        "payment:debit_cards": "yes",
+        "self_service": "yes"
       }
     },
     {
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -3816,7 +3816,7 @@
         "brand": "Marathon",
         "brand:wikidata": "Q458363",
         "name": "Marathon",
-        "website": "https://www.marathonfuel.com/",
+        "brand:website": "https://www.marathonfuel.com/",
         "payment:cash": "yes",
         "payment:contactless": "yes",
         "payment:credit_cards": "yes",
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -1019,7 +1019,9 @@
         "amenity": "fuel",
         "brand": "BP",
         "brand:wikidata": "Q152057",
-        "name": "BP"
+        "name": "BP",
+        "brand:website": "https://www.bp.com/",
+        "self_service": "yes"
       }
     },
     {


Snowysauce:
Thanks for your contribution. Please note that all `*:website` tags are outside the scope of the NSI and will be removed from the base code the next time the build script is run. ([further information here](https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#tags))

As for the remainder of the submission, I'm not sure if these tags are appropriate. The `self_service` values would be incorrect for any filling stations in New Jersey, and the payment tags for Marathon seem to be [the kind of payment tags the NSI tends to avoid](https://github.com/osmlab/name-suggestion-index/pull/10131#issuecomment-2468897999).

BPowell76:
> Thanks for your contribution. Please note that all `*:website` tags are outside the scope of the NSI and will be removed from the base code the next time the build script is run. ([further information here](https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#tags))
> 
> As for the remainder of the submission, I'm not sure if these tags are appropriate. The `self_service` values would be incorrect for any filling stations in New Jersey, and the payment tags for Marathon seem to be [the kind of payment tags the NSI tends to avoid](https://github.com/osmlab/name-suggestion-index/pull/10131#issuecomment-2468897999).

I guess I'm a little confused. Is the NSI independent of the Wikidata? How are the two related? I was thinking the NSI json data would push the tags to the wikidata page, but is that not the case?

As for the self_service tag, is there a way to filter out NJ?

UKChris-osm:
> I guess I'm a little confused. Is the NSI independent of the Wikidata? How are the two related? I was thinking the NSI json data would push the tags to the wikidata page, but is that not the case?

Wikidata is indeed independent to NSI. We used it as a reference to a brand name so that we can pull in a logo from Facebook or a Wiki Commons Logo where possible. We don't really add any information to Wikidata, except for an NSI reference that we generate, an example is [Somerset Bakehouse](https://www.wikidata.org/wiki/Q131143706) which if you scroll down, you'll see the value `OSM Name Suggestion Index ID` with the reference that we create.

This is a way to sow that we make use of a particular Wikidata article for our own needs. 

> As for the self_service tag, is there a way to filter out NJ?

You can use `locationSet` to define particular areas, using `include` as well as `exclude`.

Snowysauce:
Another concern I have with this PR is that the `self_service` presets would override any instances where someone personally surveyed a gas (petrol) station and determined that it had only self-service pumps (`self_service=only`) or only full-service pumps (`self_service=no`). We could, of course, make use of `preserveTags` to prevent this from happening, but coupled with the New Jersey issue, I do wonder if adding these tags to the presets would create more problems than it would solve.

BPowell76:
> Another concern I have with this PR is that the `self_service` presets would override any instances where someone personally surveyed a gas (petrol) station and determined that it had only self-service pumps (`self_service=only`) or only full-service pumps (`self_service=no`). We could, of course, make use of `preserveTags` to prevent this from happening, but coupled with the New Jersey issue, I do wonder if adding these tags to the presets would create more problems than it would solve.

I'm of the opinion that if this data is the US, then the "majority rules" so to speak. Gas stations are self service with the exception of New Jersey and some counties in Oregon (https://worldpopulationreview.com/state-rankings/self-service-gas-pumping-states).

I suppose if we use the locationSet exclude for NJ and OR so that they don't come up as self-service then I think that would be a compromise.

Snowysauce:
> I'm of the opinion that if this data is the US

Three of the four brands you submitted changes for have a global or nearly-global `locationSet`. So no, the affected data is not purely in the US.

> I suppose if we use the locationSet exclude for NJ and OR so that they don't come up as self-service then I think that would be a compromise.

That would require creating brand-new entries for each of the brands dedicated to New Jersey and/or Oregon that would be identical to the main entries in every way except for a tag (`self_service`) that the NSI really has no business being involved with.

bhousel:
Oh yeah I agree with the other maintainers that these additions don't make sense for NSI.
We did discuss the `payment` tags a bit on other issues, most recently a few weeks ago here, but our position has been consistent going back years:
https://github.com/osmlab/name-suggestion-index/pull/10131#issuecomment-2468897999

It is valuable to link every brand to a unique and stable Wikidata identifier, but beyond that we don't add tags for things like `brand:website` which anyone can get from the wikidata page, or `payment:cash=true` or `self_service=yes` which might be technically (mostly?) correct but don't really benefit anybody who uses OpenStreetMap data.  

Some of these tags aren't even all that stable, which adds to the maintenance burden of keeping OSM's data fresh.  (I wouldn't expect the OSM community to re-tag all 18,700 BP fuel stations every time they change their `brand:website`)

PR #10203, opened 2024-12-05
Update atm.json

rkdclgh:
add Postal Savings (operated by Korea Post) brand.


UKChris-osm:
This entry won't show up in the final index as it's in the `brand` category, but doesn't have a `brand:wikidata` attached to it.

PR #10215, opened 2024-12-10
add europtica to opticians

gedankenstuecke:
This PR adds the Argentinian chain of _Europtica_ opticians to the index. 

[According to their brand website](https://europtica.com.ar/nuestros-locales/), they currently have 38+ locations across 5 provinces, so the notability should be met I think :slightly_smiling_face:


gedankenstuecke:
Thanks so much!

PR #10220, opened 2024-12-12
MPL parcel locker: sync tags with Wiki

gy-mate:
Sync tags of _MPL_ parcel locker with https://wiki.openstreetmap.org/wiki/Hungary/Jelölési_példák#Csomagautomata and remove tags that are out of the NSI's scope.

- Remove `brand:short=*`, `name=*`, `short_name=*`
- Correct `brand` value
- Add `brand:wikidata=Q131431491`

### DIFF
--- a/data/brands/amenity/parcel_locker.json
+++ b/data/brands/amenity/parcel_locker.json
@@ -602,10 +602,8 @@
       ],
       "tags": {
         "amenity": "parcel_locker",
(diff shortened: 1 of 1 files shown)


vasony (approved):
Right. MPL is used only in short form.

gy-mate:
@Cj-Malone I just wanted to get an approval from @vasony first.

Thank you for the `displayName` and `matchNames` corrections!

PR #10222, opened 2024-12-13
Add SimplyFresh GB

TheUKHighStreet:
Convenience store chain with 25 branches. First time I've tried to add a brand so let me know if I've done something wrong.

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -3132,6 +3132,19 @@
       }
     },
     {
+      "displayName": "SimplyFresh",
+      "locationSet": {"include": ["gb-eng", "gb-wls"]},
+      "matchNames": [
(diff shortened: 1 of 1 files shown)


Cj-Malone (approved):
`Simply Fresh` and `SimplyFresh` are the same in the eyes of the matcher. It doesn't need to be (and can't be) used as a match name. Also may as well just use `gb` as it's only a matter of time until one opens in Scotland.

Adding social media to Wikidata is helpful (Specifically Facebook for the icon).

TheUKHighStreet:
Thanks - I'll add the social links.

TheUKHighStreet:
I see I was too late! Thanks.

Cj-Malone:
No worries, thank you.

PR #10224, opened 2024-12-14
Remove Hydro-Quebec from street lamp operators

wolfy1339:
Hydro-Quebec doesn't operate street lamps, that would be the transportation ministry, or the local municipality.

The reason it was being automatically added to the data is because certain power poles have street lamps and are tagged with
```
power=pole
highway=street_lamp
```

Ex: https://www.openstreetmap.org/node/9649669706
Relvant wiki entry: https://wiki.openstreetmap.org/wiki/Tag:power%3Dpole#Examples (See example from New Jersey, USA)

### DIFF
--- a/data/operators/highway/street_lamp.json
+++ b/data/operators/highway/street_lamp.json
@@ -88,18 +88,6 @@
       }
     },
     {
-      "displayName": "Hydro-Québec",
-      "id": "hydroquebec-0ef919",
-      "locationSet": {
-        "include": ["ca-qc.geojson"]
-      },
-      "tags": {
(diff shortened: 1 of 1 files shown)


Cj-Malone:
If it was automatically added from OSM data, it probably needs adding to the excludes to prevent it being re added.

CorruptComputer:
Apologies for the delayed reply here, these were not added automatically but by me in #9969. I had just been searching for already mapped operators using overpass to find them, I'll add a filter to my query to remove nodes that also contain a power pole so this doesn't happen again. Thanks for catching it!

PR #10225, opened 2024-12-14
Update sports.json

Fabe56:
Changing Wikidata item for Rapha to Q131446199, a dedicated entry for Rapha Clubhouses retail chain.

### DIFF
--- a/data/brands/shop/sports.json
+++ b/data/brands/shop/sports.json
@@ -737,7 +737,7 @@
         "brand": "Rapha",
         "brand:en": "Rapha",
         "brand:ja": "ラファ",
-        "brand:wikidata": "Q7293931",
+        "brand:wikidata": "Q131446199",
         "name": "Rapha",
         "name:en": "Rapha",
         "name:ja": "ラファ",


Snowysauce:
This has little to do with the specific code changed by this PR, but it seems unusual for an entry's `locationSet` to have both Latin script and non-Latin script countries. Should "Rapha" really be the default name for their stores in Japan, South Korea, and Taiwan? (There's also the discrepancy in names between "Rapha" (NSI) and "Rapha Clubhouse" (Wikidata), but one issue at a time, I suppose.)

Fabe56:
Some good points for sure! Languages selection on the Rapha website propose for each regions an English version. In [Traditional Chinese](https://content.rapha.cc/tw/zh/clubhouses) Rapha Clubhouses or Clubhouse still appear in English, I did not personally find any trademark, but I guess it's a company decision to uniform it.
About the discrepancy in names, I saw a consistence to name each store under this form Rapha [City] Clubhouse (from the [store webpage name](https://content.rapha.cc/au/en/clubhouses/taipei)) or Clubhouse [City] (from the [careers page](https://careers.rapha.cc/jobs)). However, this is not true everywhere: [Rapha London](https://www.instagram.com/rapha_london), [Rapha Seoul](https://www.instagram.com/raphaseoul) vs. [Rapha Hong Kong Clubhouse](https://www.instagram.com/rapha_hongkong).
And the name on the storefronts is [Rapha.](https://commons.wikimedia.org/wiki/File:DSC_5345_(37909666236).jpg) (with the dot to be precise).

PR #10227, opened 2024-12-15
add foodsharing.de to brands/amenity/food_sharing

### LINKED ISSUE #10226: move foodsharing.de from operators/amenity/social_facility to brands/amenity/food_sharing
[This](https://github.com/osmlab/name-suggestion-index/blob/5759a99c8c28b20d1128ddf2ba7776956ad17313/data/operators/amenity/social_facility.json#L743C1-L755C7) preset is [not contained in the OSM data](https://overpass-turbo.eu/s/1VIs). It is also not clear to me what it is actually intended for. I think a preset in brands/amenity/food_sharing is better suited for the foodsharing.de "Fairteiler".

I suggest deleting the preset in operators/amenity/social_facility and creating one in brands/amenity/food_sharing for “ Fairteiler".

tiptoptom:
Closes #10226


Snowysauce:
As stated at https://github.com/osmlab/name-suggestion-index/issues/10226#issuecomment-2663821790, half of this PR would be reverted on the next run of the build script post-merge. Unless there's a specific reason to explicitly exclude the `operators/amenity/social_facility` preset from our data, the addition to `brands/amenity/food_sharing.json` has to be considered a standalone change.

tiptoptom:
Is there anything left to do here before it can be merged?

Snowysauce:
Overpass-turbo returned 0 results for both `amenity=food_sharing` and `brand=foodsharing.de`, and `amenity=food_sharing` and a regex search for "foodsharing.de" anywhere in `name`. However, there are six results for `amenity=food_sharing` and `operator=foodsharing.de`. `operators/amenity/food_sharing` doesn't currently exist, but current OSM tagging indicates that this is where this preset would be best suited.

In either case, more research is needed on whether this preset meets the NSI's inclusion criteria - at least 10 instances as an `operator`, or around 30 instances as a `brand`.

bhousel:
> In either case, more research is needed on whether this preset meets the NSI's inclusion criteria - at least 10 instances as an `operator`, or around 30 instances as a `brand`.

These are soft requirements - its not like I invented these numbers to mean anything.  These numbers were mostly chosen to   have the collection script collect actual frequently used brand names from the OSM planet file, and not just descriptive strings like "beauty salon" or "bus stop".  If a businesses has fewer instances but is real and has a complete record in Wikidata, thats ok.

FWIW I consider NSI's notability requirements to be "whatever Wikidata's notability requirements are".  And also for efficiency, if a business only has 5 locations it is faster to just edit those 5 locations in OSM and move on with life.

tiptoptom:
According to [this page](https://wiki.openstreetmap.org/wiki/DE:Foodsharing), there are more than 800 foodsharing.de boxes.
These boxes are usually set up and maintained by individuals, only the organization is done via foodsharing.de. I think that `brand=*` is therefore more appropriate than `operator=*`. Besides, other `amenity=food_sharing` networks are already in `brand/amenity/food_sharing`.

bhousel:
There were conflicts, so I just added it myself in a91124c5c

PR #10232, opened 2024-12-18
Add "Company" to Pacific Gas and Electric Company operator value

spughettilinetti:
Added the word "Company" to the value of the operator tag. This is how the company calls itself on its company profile.

https://www.pge.com/en/about/company-information/company-profile.html

### DIFF
--- a/data/operators/power/line.json
+++ b/data/operators/power/line.json
@@ -6334,7 +6334,7 @@
         "pacific gas and electric company"
       ],
       "tags": {
-        "operator": "Pacific Gas and Electric",
+        "operator": "Pacific Gas and Electric Company",
         "operator:short": "PG&E",
         "operator:wikidata": "Q1815011",
         "power": "line"


data/operators/power/line.json
```diff
@@ -6334,7 +6334,7 @@
         "pacific gas and electric company"
       ],
       "tags": {
-        "operator": "Pacific Gas and Electric",
+        "operator": "Pacific Gas and Electric Company",
```
JesseWeinstein, line 6337: PG&E is mentioned in two other places in NSI, so this should change those, too.

https://github.com/osmlab/name-suggestion-index/blob/2cf9fbda54a4c9cd37ed627424ed6dfe64b67266/data/operators/man_made/pipeline.json#L6124

https://github.com/osmlab/name-suggestion-index/blob/2cf9fbda54a4c9cd37ed627424ed6dfe64b67266/data/operators/power/plant.json#L2551

Snowysauce: PG&E is also in ops/power/substation:

https://github.com/osmlab/name-suggestion-index/blob/346377d707794b8ab92a87f7a868213ffb3909d9/data/operators/power/substation.json#L7454

As @JesseWeinstein said, any PR that changes the operator value for PG&E in ops/power/line also has to change the value for PG&E in the other three categories in order to be considered for merging. (Each entry also needs to have the existing operator value - Pacific Gas and Electric - listed in `matchNames`.)

spughettilinetti: Sorry for the delay, I'll get 2 it this week

spughettilinetti: Sorry again for the delay! I've updated both.

https://github.com/spughettilinetti/name-suggestion-index/commit/80d35acf71170298191aeb96c7a0924e98a5a8f4
https://github.com/spughettilinetti/name-suggestion-index/commit/c918342e929284ed6ec6aa8079e317ff21958ef8

1ec5 (approved):

PR #10235, opened 2024-12-19
Sync with GE preset

balsoft:
Syncs with https://github.com/komachi/josm-ge-preset, mostly adding wikidata tags


Cj-Malone (changes requested):
Having a quick glance, the biggest thing is `operator:wikidata` without `operator`. But also I don't think we should be encouraging `operator:wikidata` to be a copy of `brand:wikidata`.

balsoft:
@Cj-Malone done, please squash before merging though

balsoft:
@Snowysauce agreed on both counts, fixed.

Cj-Malone (approved):
I'm not to fond of some of this, removing `name`, flip flopping on `payment:*`, adding `operator` which is the same as `brand`. However, lets go for now.

PR #10238, opened 2024-12-19
Add "Era Immobilier" for France

QuentinS33:
World group : Era Real Estate
France : Era Immobilier

On nsi, there is an entry for Era only in Bulgaria. The is the same world group

### DIFF
--- a/data/brands/office/estate_agent.json
+++ b/data/brands/office/estate_agent.json
@@ -408,9 +408,9 @@
       "id": "era",
       "locationSet": {"include": ["fx", "mq"]},
       "tags": {
-        "brand": "Era",
+        "brand": "ERA",
         "brand:wikidata": "Q121496901",
-        "name": "Era Immobilier",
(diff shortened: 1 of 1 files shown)


Snowysauce:
I'm not sure if Era should be in all caps or not. The global entry (added to the NSI after this request was created) uses all caps; however, that was a judgment call based on how the brand is presented in the US, which will use the global entry beginning with the next NSI release. (The name's origins is as an acronym for "Electronic Realty Associates".)

This situation may be completely different in France, but as an American I wouldn't have that knowledge. Overpass-turbo regex queries for `office=estate_agent` in OSM with the names "Era" and "ERA" returned roughly the same amount of results in France, further muddying the waters as to which spelling should be considered "canonical".

Snowysauce:
ERA France's official website (https://www.erafrance.com/) consistently uses "ERA", and as such I've edited the PR to use the all-caps format.

PR #10250, opened 2024-12-21
Update sports.json

Fabe56:
Adding Pas Normal Studios Flagship Store.

### DIFF
--- a/data/brands/shop/sports.json
+++ b/data/brands/shop/sports.json
@@ -641,6 +641,17 @@
       }
     },
     {
+      "displayName": "Pas Normal Studios Flagship Store",
+      "id":,
+      "locationSet": {"include": ["dk", "es", "de", "us", "kr", "sg", "tw"]},
+      "tags": {
+        "brand": "Pas Normal Studios Flagship Store",
+        "brand:wikidata": "Q131523198",
+        "name": "Pas Normal Studios Flagship Store",
+        "shop": "sports"
+      }
+    },
+    {
       "displayName": "Peloton",
       "id": "peloton-18eda3",
       "locationSet": {


UKChris-osm:
Would [Q131522667](https://www.wikidata.org/wiki/Q131522667) be more suitable? Also, do they include the words "_Flagship Store_" in the name of each store?

Fabe56:
> Would [Q131522667](https://www.wikidata.org/wiki/Q131522667) be more suitable? Also, do they include the words "_Flagship Store_" in the name of each store?

This is the item for the fashion label, while [Q131523198](https://www.wikidata.org/wiki/Q131523198) is for the retail chain. Different concept should not be merged in Wikidata.
For the name, the company name them with this Pas Normal Studios [City] Flagship Store in description, example with: [Pas Normal Studios Copenhagen Flagship Store](https://pasnormalstudios.com/eu/pages/copenhagen#:~:text=on%20Facebook%0A%0AThe-,Pas%20Normal%20Studios%20Copenhagen%20Flagship%20Store,-is%20located%20in) or [Pas Normal Studios San Francisco Flagship store](https://pasnormalstudios.com/eu/pages/san-francisco#:~:text=on%20Facebook%0A%0AThe-,Pas%20Normal%20Studios%20San%20Francisco%20Flagship%20store,-is%20located%20in). While shop sign show "Pas Normal Studios International Cycling Club".
What about just "Pas Normal Studios store"?

Snowysauce:
The existence of separate Wikidata items for "brand A" and a "brand of shops dedicated to selling brand A" is unusual, but not unprecedented. (I think a similar arrangement exists for the gas/petrol brand Shell, and I briefly considered something like this last night for Toys R Us based on how fragmented that brand has become since 2018.)

I think the `brand` value in both cases here is "Pas Normal Studios", with "Pas Normal Studios International Cycling Club" possibly being an `official_name` value. "Pas Normal Studios store" would be akin to (for example) "Kwik Fill gas station" if Wikidata items existed for both the gas station chain and the Kwik Fill-branded gasoline that they sell; in other words, tacking on "store" makes the name a description, not a formal brand name. As for "Flagship Store", in this case, with that phrase not being on signage and with it being used online as a descriptor for every store they operate, I don't consider it as being anything more than an overly posh way of saying "store location".

Another concern is that there are only seven stores with this specific nameplate. Generally, the minimum needed in order to be added to the NSI is 10-30.

PR #10284, opened 2024-12-30
comparable matchNames for aldi nord and süd

tiptoptom:


Cj-Malone:
This doesn't make sense. Adding Aldi Nord and Sud will no nothing as it's already the brand string, removing Aldi from ALDI Süd may have a negative outcome.

tiptoptom:
But it makes no sense at all to have “aldi” as a math name only at Aldi Süd. I think Aldi Nord and Aldi Süd should have comparable matchnames.

PR #10288, opened 2024-12-31
exclude two Aldi Nord branches

tiptoptom:
see https://github.com/osmlab/name-suggestion-index/commit/628624f3048e692c3d152de8f9cafe61b7c1a06f#commitcomment-150834234


Cj-Malone:
Thank you

PR #10289, opened 2024-12-31
Add Yettel banka

u-spec-png:
Rebranding see https://en.wikipedia.org/wiki/Yettel_Bank

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -9199,7 +9199,7 @@
       }
     },
     {
-      "displayName": "Yettel banka",
+      "displayName": "Yettel bank",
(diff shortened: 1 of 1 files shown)


Snowysauce:
https://github.com/osmlab/name-suggestion-index/pull/10289/files#diff-14bd1c952a27c2ef87db44bd9c4f546d2ec8220b7ae64a726dd53f10df0cdd68R9214-R9217 - per https://wiki.openstreetmap.org/wiki/Multilingual_names, there should be a suffixed name tag specifying which language "Yettel banka" is in.

u-spec-png:
It was a typo, fixed now.

PR #10291, opened 2025-01-01
[nz] Labtests -> Awanui Labs

k-yle:
I don't think we need to keep the old entry (?) I'll update [the remaining 25 nodes](https://taginfo.osm.org/tags/brand:wikidata=Q113312315) today which rebranded a while ago


bhousel:
Thanks @k-yle .. Sometimes it makes sense to add `old_name` but if you don't think we need it, I'm fine with merging as-is.

PR #10296, opened 2025-01-01
[nz] remove `operator:type` from DOC

k-yle:
There have been some counterproductive changes to DOC presets recently... 

1.  NSI suggests adding `operator:type=government` to thousands of protected areas & parks, which isn't helpful and just creates noise
2. there are some equally unhelpful "upgrades" which actually destroy the data when people blindly accept these suggestions:

image

image

### DIFF
--- a/data/operators/boundary/protected_area.json
+++ b/data/operators/boundary/protected_area.json
@@ -537,7 +537,6 @@
       "tags": {
         "boundary": "protected_area",
         "operator": "Department of Conservation",
-        "operator:type": "government",
         "operator:wikidata": "Q1191417"
       }
     },
--- a/data/operators/leisure/park.json
+++ b/data/operators/leisure/park.json
@@ -3470,7 +3470,6 @@
       "tags": {
         "leisure": "park",
         "operator": "Department of Conservation",
-        "operator:type": "government",
         "operator:wikidata": "Q1191417"
       }
     },


UKChris-osm:
Problem No.2 seems to be from the markers and information boards being a sub-tag of `tourist=information`, which is where the suggestion is coming from.

I don't know how to exclude sub-tags from a match, unless preserveTags would also work in this case?

Alternatively I wonder if we could also create a category for markers, boards and maps, using `tourist=information` as the template, but again, being that the sub-tag would be the focus, I don't know if that's possible?

@bhousel would you know??

Snowysauce:
Regarding issue 1, `operator:type` is one of the "quick access" fields for parks in the iD editor (see [here](https://github.com/openstreetmap/id-tagging-schema/blob/9aeafb1e4431f6bffc5f7c9438e568cac5d2996c/data/presets/leisure/park.json#L6)). Since iD is the most popular editor, it means that the "noise" values will probably be added eventually anyway, and thus it makes sense for NSI to provide a suggested best value for the tag. If you truly believe that `operator:type` is unnecessary data noise for parks, I would suggest submitting an issue to [@openstreetmap/id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema/tree/main) to have the field removed from the sidebar.

k-yle:
Regarding issue 1: @Snowysauce thanks for the explanation, I hadn't encountered an "error" like this before, the ones I usually see relate to the `(brand|operator):*` tags. Since this is intentional, I guess https://github.com/openstreetmap/iD/issues/6101 or https://github.com/facebook/Rapid/issues/1623 would be the long-term solution, so that users are not spammed with validator errors which they didn't cause (some people [get really frustrated](https://www.openstreetmap.org/changeset/157329851) by this)

---

Regarding issue 2: @UKChris-osm would it be acceptable to delete the DOC entry for `information=office` as an interim solution? I'm concerned that people who blindly fix these "errors" will start damaging the data (or maybe it's already happened)

UKChris-osm:
> would it be acceptable to delete the DOC entry for `information=office` as an interim solution?

I've removed the Wikidata reference from `information=office` so it'll stop being suggested, as an entry needs a Wikidata to make it into the final index.

bhousel:
> Alternatively I wonder if we could also create a category for markers, boards and maps, using `tourist=information` as the template, but again, being that the sub-tag would be the focus, I don't know if that's possible?
> 
> @bhousel would you know??

I'm not sure.  It sounds like a problem with the established tagging, if we need to look at a multiple tags (`tourism=*`+`information=*`) to know what a thing is.  I wrote about this on #4765.  My recommendation would be to remove everything from NSI that works this way.

Also `operator:type=*` is well established and the rail and infrastructure mappers like it.  It should probably stay.

k-yle:
thanks everyone, I see this was resolved in 9c7cf80a1f7d1442763bd5b09aa62c520b1eda2b so I'm gonna close this PR. appreciate the help!

PR #10297, opened 2025-01-01
Change Mr. Tire to Auto Service Center from Car Parts store

BPowell76:
Mr. Tire is wrongly marked as a Car Parts store. This company is a car service center and in this PR, I've moved the data from `car_parts.json` to `car_repair.json`.

Additionally, updated some of the information on Wikidata.


BPowell76:
Is there something I can do to help move this PR along?

PR #10298, opened 2025-01-02
Remove Flowbird as a brand

Crashillo:
[Flowbird](https://www.flowbird.com/) is an app for the parking tickets payment. That was included in the NSI long time ago as a brand for `amenity=vending_machine` with no `locationSet`. After that, some excludes were appended.

The point is Flowbird is NOT an actual brand, as we understand in the NSI, but a payment method. There's a `payment:flowbird=*` key indeed. Flowbird does not provide a vending machine itself (just do a google search: _flowbird vending machine_), that's something the local administration sets. Neither has a so-called name Flowbird.

That's why this brand should be removed from NSI, it should never be suggested when a user introduces a new vending machine.

Wiki: https://wiki.openstreetmap.org/wiki/DE:Tag:vending%3Dparking_tickets [DE]
Community: https://community.openstreetmap.org/t/parquimetros-con-name-flowbird/123573 [ES]

### DIFF
--- a/data/brands/amenity/vending_machine.json
+++ b/data/brands/amenity/vending_machine.json
@@ -10,6 +10,7 @@
         "^(自動)?[券販]売機$",
         "^automat na jizdenky$",
         "^casa dell'acqua$",
+        "^flowbird$",
         "^horodateur$",
         "^máquina de tickets$",
         "^milchtankstelle$",
--- a/data/brands/amenity/vending_machine.json
+++ b/data/brands/amenity/vending_machine.json
@@ -546,21 +546,6 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Flowbird appears over 700 times in [OSM Data](https://github.com/ideditor/nsi-collector/blob/1de2a8a25314ed059d8ca1a31ffe940004a1a429/dist/osm/brands_all.json#L2791) at the moment, and as such the NSI will automatically re-add it every time `npm run build` is run as any brand that appears more than 50 times is added.

At the top of the file you've already edited, you can add the name "_Flowbird_" to the exclusion list to prevent it reappearing.

PR #10310, opened 2025-01-06
Delete Total brand fuel station entry

ckpiper:
I think we can delete this entry from NSI as most places have been rebranded by this point. For any that are left we can add total to the matchNames of TotalEnergies (as well as "station total").

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -6346,19 +6346,6 @@
       }
     },
     {
-      "displayName": "Total",
-      "id": "total-b3d110",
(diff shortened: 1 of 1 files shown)


Cj-Malone (approved):

Snowysauce:
The locationSet for TotalEnergies should probably have an exception for the US per <https://totalenergies.com/united-states>: "We are active in the United States in all our business segments, _except retail fuel_."

ckpiper:
@Snowysauce Exclusion added

Snowysauce (approved):

PR #10326, opened 2025-01-12
Adding "Manolo Bakes" and "Pastelería Artesanal Manolo" Spanish cafes

aguafluye:


Snowysauce:
I read a translated version of the second "described at URL" for Manolo Artisanal Pastry, and I want to make sure I understand the relationship between the two brands correctly. From what I read, it seems like Manolo Artisanal Pastry is a smaller operation (apparently 11 locations) while Manolo Bakes is a more commercial endeavor (40-ish locations) that the owners of Manolo Artisanal Pastry once had an ownership share in but no longer do.

Also, I see that the Wikidata items for each brand use the paired properties "has cause"/"has effect" to link them together; however, I'm not sure if that's the best way to convey their relationship to one another. That said, if my description in the above paragraph is correct, I'm also not sure what the best way is. Perhaps a blurb in each of the items' description is appropriate, or perhaps the items don't need to be linked at all.

aguafluye:
> I read a translated version of the second "described at URL" for Manolo Artisanal Pastry, and I want to make sure I understand the relationship between the two brands correctly. From what I read, it seems like Manolo Artisanal Pastry is a smaller operation (apparently 11 locations) while Manolo Bakes is a more commercial endeavor (40-ish locations) that the owners of Manolo Artisanal Pastry once had an ownership share in but no longer do.
> 
> Also, I see that the Wikidata items for each brand use the paired properties "has cause"/"has effect" to link them together; however, I'm not sure if that's the best way to convey their relationship to one another. That said, if my description in the above paragraph is correct, I'm also not sure what the best way is. Perhaps a blurb in each of the items' description is appropriate, or perhaps the items don't need to be linked at all.

Good morning. That is exactly right, the Manolo Bakes brand is franchised whereas the original one is not AFAIK.

I would have some way of linking the two Q objects in wikidata just to be machine friendly and have them connected in a graph. I'd agree cause and effect are not the best properties, a "related to" property is good enough.

Snowysauce:
Thanks for the feedback. I'm still not sure what the best way forward is for the Wikidata situation; the timeline of developments and the discussion of how the two brands are and aren't related is a bit more complicated than what a statement-based database is designed to accommodate. The ideal situation would obviously be an encyclopedia article that covers everything in prose, which unfortunately doesn't and may never exist.

However, since this is a question of mine on the Wikidata end of things that has absolutely nothing to do with the NSI, I'll just let the Wikidata items stand as they are. What I can do is link this PR to the proposed NSI presets, which will then generate a link to this discussion on nsi.guide. Since our internal IDs for the brands will be linked from each Wikidata QID, editors and readers curious about the brands' relationship could then theoretically directly visit this discussion from Wikidata (with a stop to nsi.guide in between).

Now, for the NSI: the presets look fine, Manolo Bakes easily qualifies for the index, and while Manolo Artisanal Pastry is on the lower end of what we use as the minimum criteria for inclusion (10-30 locations), it's probably necessary to include that brand too so OSM editors aren't tempted to use the Manolo Bakes preset when they shouldn't.

PR #10328, opened 2025-01-12
add preserveTags to nsi-id-presets.json

### LINKED ISSUE #10083: Include `preserveTags` in `nsi-id-presets.json`
Currently, the `preserveTags` property is only added to the presets in `dist/nsi.json` but not to the presets in `dist/presets/nsi-id-presets.json`.

Since this property should trigger a change in behavior how the preset from the NSI is applied, it should be included in the `nsi-id-presets.json`, too, even if that property is not mentioned in the schema for presets.

I am using the `nsi-id-presets.json` in StreetComplete (via the [osmfeatures](https://github.com/westnordost/osmfeatures) library). It is convenient for me to use these because they are in the same format as the iD presets.

westnordost:
fixes #10083


Cj-Malone:
Thank you

PR #10335, opened 2025-01-13
Deleted Total Access

ckpiper:
Total Access is rebranding to TotalEnergies as with the larger "Total" company rebrand. I added it to the match names of Total Energies as I did with regular Total.

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -6366,18 +6366,6 @@
       }
     },
     {
-      "displayName": "Total Access",
-      "id": "totalaccess-3772a0",
-      "locationSet": {"include": ["fr"]},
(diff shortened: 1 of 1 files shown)


Snowysauce (commented):
A quick search online indicates that these stations were initially rebranded as "TotalEnergies Access". Based on this PR, I'm assuming that the TE Access brand has been folded into TotalEnergies, but I have no way of knowing since there are no TotalEnergies of any kind in the US. Beyond that question, I would consider adding "totalenergies access" to the list of matchNames since it was a matchName for the proposed-to-be-merged Total Access preset.

ckpiper:
@Snowysauce Good catch. I did not notice that. It is in the matchnames now.

PR #10345, opened 2025-01-14
Deleted Punto Simply and Simply Market Entries

ckpiper:
Simply Market Rebranded from 2016-2019 to Auchan Supermarche. Punto Simply merged into Simply Market.


UKChris-osm:
Punto Simply has a `locationSet` for Italy, do they still trade as `Auchan Supermarche` in Italy?

ckpiper:
From what I can see Auchan in Italy sold most of their stores to Conad. Thus, no longer in Italy at all.

PR #10346, opened 2025-01-14
Deleted Michelin and Pirelli

ckpiper:
Both of these brands do not have physical locations. Thus, I am wondering if it even makes sense to have them in NSI. Yes they are brands, but the point of NSI is to have common brand names for Places. Then POIs in OSM will have matching common names if they are the same company stores. Am I thinking about this correctly?

### DIFF
--- a/data/brands/shop/tyres.json
+++ b/data/brands/shop/tyres.json
@@ -3,6 +3,7 @@
     "path": "brands/shop/tyres",
     "exclude": {
       "generic": [
+        "^(michelin|pirelli)$",
         "^(tyres|wulkanizacja|vulcanizare)$",
         "^(вулканизация|шиномонтаж)$",
         "^borrachar[ií]a$",
--- a/data/brands/shop/tyres.json
+++ b/data/brands/shop/tyres.json
@@ -370,20 +370,6 @@
       }
     },
     {
-      "displayName": "Michelin",
-      "id": "michelin-c3bb11",
-      "locationSet": {
-        "include": ["001"],
-        "exclude": ["fr", "gb"]
-      },
-      "tags": {
-        "brand": "Michelin",
-        "brand:wikidata": "Q151107",
-        "name": "Michelin",
-        "shop": "tyres"
-      }
-    },
-    {
       "displayName": "OK Tire",
       "id": "oktire-20a245",
       "locationSet": {"include": ["ca"]},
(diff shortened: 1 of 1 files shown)


bhousel:
> Both of these brands do not have physical locations. Thus, I am wondering if it even makes sense to have them in NSI. Yes they are brands, but the point of NSI is to have common brand names for Places. Then POIs in OSM will have matching common names if they are the same company stores. Am I thinking about this correctly?

You are correct - we really only want to have brands with physical locations that should be mapped in OSM.

However I checked the Overpass queries on this page:
https://nsi.guide/index.html?t=brands&k=shop&v=tyres

...and I can see that for both these brands there are a lot of places in OpenStreetMap mapped like `shop=tyres`+`name=Michelin` or `name=Pirelli`

So maybe you're mistaken, or maybe these shops really are all mapped wrongly - I do not know.

But what it means is that even if we do delete these entries, our script that loads names from the OSM data will just put them back, because as far as the script is concerned they seem to be real names of tyre shops.

Oivo35:
I looked at the objects that have website and most of them are tagged incorrectly. These stores only sell Michelin and Pirelli tyres. So they should have the name of the stores.
And I also found a store in Australia called Michelin, so maybe for some stores the tags `name=Michelin` + `brand=Michelin` are correct.

BPowell76:
For what it's worth, I've never seen a tire shop that only sold one brand of tire. Shoot, even Firestone in the US sells multiple brands (even though I think Bridgestone and Firestone are the same parent company).

In my local area, there are no Michelin or Pirelli stores. Usually you get them through a dealer such as Mavis, Discount Tire, or Wal-Mart/Sam's Club.

ckpiper:
As @bhousel said above if the script which loads names from OSM data will end up putting Michelin and Pirelli back into NSI even after this was merged. I think it is safe to say that we can Close this PR.

BPowell76:
> As @bhousel said above if the script which loads names from OSM data will end up putting Michelin and Pirelli back into NSI even after this was merged. I think it is safe to say that we can Close this PR.

I disagree. If the data is wrong, it's wrong and needs to be fixed.

Cj-Malone:
It can be added to excludes to prevent it being re added automatically

ckpiper:
@Cj-Malone Do you mean add to exlude->generic at the top of tyres.json?

Cj-Malone:
Yes, adding them there will stop them being readded.

This is an assumption, but I think it's a safe bet:

1. People added `Michelin` to OSM, maybe because they sell Michelin branded tyres, maybe because there is a big advert on the side of the building. (Maybe `name`, probably `brand`)
2. NSI collection OSM
3. It becomes more popular because iD has it in search results

This feels pretty safe to me, at least in GB. Worst comes to worst, they can be re added.

ckpiper:
Awesome. Thanks for the info.

PR #10359, opened 2025-01-15
add Royal Rangers in brands/club/scout

tiptoptom:


data/brands/club/scout.json
```diff
@@ -77,6 +77,16 @@
         "club": "scout"
       }
     },
+    {
+      "displayName": "Royal Rangers",
+      "locationSet": {"include": ["001"]},
```
LaoshuBaby, line 82: I don't think this "royal" things is suitable for all over the world……maybe you can list some major countries in locationSet

tiptoptom: I have found an [overview](https://rri.world/Member_nations.php) of the countries in which the Royal Ragers are active. I think there are 76 countries. Should I add them all to the locationSet?

tiptoptom:
I have added the countries from the website.
https://rri.world/Member_nations.php

tiptoptom:
@LaoshuBaby any news on this? :see_no_evil:

Snowysauce:
I would say that a 73-country `locationSet` seems excessively detailed, but it wouldn't even be the largest `locationSet` in the index if added (that dubious distinction goes to [allianz-c74935](https://nsi.guide/index.html?t=brands&k=office&v=insurance&tt=allianz#allianz-c74935) and its 75-country set). Since precedent indicates that the `locationSet` is acceptable, and the fact that even one club in each listed country would easily push the brand over the NSI notability threshold, this PR seems OK.

PR #10378, opened 2025-01-20
Remove any related multilingual keys from templates

### LINKED ISSUE #7236: ATMs shouldn't have a name tag

### LINKED ISSUE #9260: name:XX left in templated entries
When an entry is created from the `brands/advertising/totem` template, it nukes `name`, but not `name:XX`.

Dimitar5555:
Resolves #7236
Resolves #9260

I might've missed or overlooked something so please test this change. Other than that it seems to produce correct results.

P.S. Quite a few banks have `official_name`, `alt_name` or `short_name`. If we are removing `name` from ATMs, we should remove the above mentioned tags as well.

### DIFF
--- a/lib/file_tree.js
+++ b/lib/file_tree.js
@@ -391,7 +391,7 @@ expandTemplates: (cache, loco) => {
           } else {
             delete tags[osmkey];
             // remove any related multilingual keys
-            const multilingual_keys = ['name', 'alt_name', 'official_name', 'short_name'];
+            const multilingual_keys = ['name', 'alt_name', 'official_name', 'short_name', 'full_name'];
             if (multilingual_keys.includes(osmkey)) {
               Object.keys(tags).forEach (key => {
                 if (key.startsWith(osmkey + ':')) {
--- a/lib/file_tree.js
+++ b/lib/file_tree.js
@@ -390,6 +390,15 @@ expandTemplates: (cache, loco) => {
             tags[osmkey] = tagValue;
           } else {
             delete tags[osmkey];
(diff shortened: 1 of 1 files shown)


bhousel:
I think it's better to use the [templateTags](https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#templateTags) property to control which tags get copied into the ATMs, rather than hardcoding the tags that you want to exclude in the file reader code.

Snowysauce:
I'm not well-versed in JavaScript, but my understanding is that this new code looks to see if any `name`/`*_name` tags have values of `""` in templates (erasing the tags), and if they do, then it also removes any subtags of the tags set to be erased (`name:*`/`*_name:*`). If that's what this code indeed does, it seems like a good thing to me.

Another tag to consider checking for is `full_name`; there's one use of that tag in `brands/amenity/bank`, complete with a set of companion multilingual tags.

Dimitar5555:
> I'm not well-versed in JavaScript, but my understanding is that this new code looks to see if any `name`/`*_name` tags have values of `""` in templates (erasing the tags), and if they do, then it also removes any subtags of the tags set to be erased (`name:*`/`*_name:*`). If that's what this code indeed does, it seems like a good thing to me.

Yes, this is what the code does. It might be a bit unclear so any suggestions for improvements are welcome!

> Another tag to consider checking for is `full_name`; there's one use of that tag in `brands/amenity/bank`, complete with a set of companion multilingual tags.

Added.

PR #10379, opened 2025-01-20
Update Mavis Tire Brake to car service center from tire store

BPowell76:
Mavis Discount Tire (aka Mavis Tires & Brakes) is a tire and car service center. They sell and do tire work, but also will perform oil changes, brakes inspections/work, and suspension work among a few others things.

Changed the main name to Mavis Tires & Brakes as it appeared more often in the stores list than Mavis Discount Tire, but kept both for name matching.

Also added some pre-selected service center options, but these can be removed if need be.

A list of locations to verify the services performed can be found [here](https://www.mavis.com/locations/all-stores/).


data/brands/shop/car_repair.json
```diff
@@ -756,6 +756,28 @@
...
+      "locationSet": {"include": ["us"]},
+      "matchNames": [
+           "mavis discount tire",
+           "mavis tires and brakes"
+      ],
+      "preserveTags": ["^name"],
+      "tags": {
+        "brand": "Mavis Tire",
+        "brand:wikidata": "Q65058420",
+        "name": "Mavis Tire",
```
Cj-Malone, line 771: Based on the link, I see `Mavis Discount Tire` and `Mavis Tires & Brakes` but not `Mavis Tire`.

Snowysauce:
Based on existing entries, `service:vehicle:*` is a permitted tag group; however, its usage is generally limited to a shop's defining specialty (e.g., `"service:vehicle:glass": "yes"` for a car repair shop that focuses on glass repairs). There's only one existing entry that has more than two `service:vehicle:*` tags (`tigerwheelandtyre-54a38f`), and said entry probably needs a tag cleanup.

data/brands/shop/car_repair.json
BPowell76: It's part of the site's logo

image

BPowell76:
> Based on existing entries, `service:vehicle:*` is a permitted tag group; however, its usage is generally limited to a shop's defining specialty (e.g., `"service:vehicle:glass": "yes"` for a car repair shop that focuses on glass repairs). There's only one existing entry that has more than two `service:vehicle:*` tags (`tigerwheelandtyre-54a38f`), and said entry probably needs a tag cleanup.

I can change it to just tires if we want to keep it limited to one tag, but I feel that might spark some discussion later on about why it was changed from a tire shop to a car repair shop.

data/brands/shop/car_repair.json
Snowysauce: I haven't had the opportunity to fully look into this, but if the company has only two brands, and all stores carry one of the two names, wouldn't it make more sense to split the existing entry into two and drop `preserveTags`? Otherwise, by using `preserveTags` and trying to shoehorn one brand into the other, we're not really standardizing anything, and possibly sending bad suggestions to OSM.

Snowysauce:
With the exception of [tigerwheelandtyre-54a38f](https://nsi.guide/?id=tigerwheelandtyre-54a38f), all-purpose car repair chains in the NSI typically do not have any `service:vehicle:*` tags in their preset.

BPowell76:
Ok. I'll update it and push a new commit this evening after work.

data/brands/shop/car_repair.json
BPowell76: I suppose so. What if we use Mavis Tire as a preserveTags name and use either of the existing ones as the primary? I think Mavis Tires & Brakes was the more common store name (and would be my vote the "primary name"), but it has been a couple of days since I last looked at it.

BPowell76:
It has been a while, but I wanted to let everyone involved in this PR that I pushed a new commit with changes to address the discussions about 2 weeks ago.

data/brands/shop/car_repair.json
Snowysauce: That approach still has the same problem as the original proposal. Two separate entries is the better way to go; even Mavis themselves considers the two names to be sibling brands ([Under FAQs: "Is Mavis Discount Tire the same as Mavis Tires and Brakes? - Yes! Mavis Tires and Brakes is a sub brand of Mavis Discount Tire."](https://www.mavis.com/mavis-family/))

Issue #9957, opened 2024-09-17
St. Vincent de Paul thrift stores


tylerfilla:
St. Vincent de Paul thrift stores are operated by independent nonprofits kind of like Goodwill shops. There's an [international organization](https://www.ssvpglobal.org), national units (e.g. [SSVP/SVdP USA](https://ssvpusa.org/)), and a bunch of local chapters (e.g. [SVdP St. Louis](https://svdpstlouis.org) and [SVdP Santa Clara County](https://www.svdp.org)). A chapter can operate a network of thrift stores (though not all do) and customize its branding quite a bit, so there isn't really a homogenous brand like what's currently modeled [in NSI](https://nsi.guide/index.html?t=brands&k=shop&v=charity&tt=vincent#saintvincentdepaulthriftstore-fca2e3) or even [on Wikidata](https://www.wikidata.org/wiki/Q66581837). There already are Wikidata items for [many of these local chapters](https://www.wikidata.org/w/index.php?search=society+of+st.+vincent+de+paul&title=Special:Search&profile=advanced&fulltext=1&ns0=1&ns120=1) (councils). I was getting ready to split [the Wikidata item](https://www.wikidata.org/wiki/Q66581837), as it conflates [SVdP USA](https://ssvpusa.org/) with this abstract retail chain identity, but I wanted to talk about it here first because it was created for NSI. I think St. Vincent de Paul thrift stores should be tagged like how Goodwill stores are tagged, as their organizational structures are very similar. Thoughts on this?

1ec5:
As you suspected, when I created this item, I had in mind the “chain” (more like a concept) more than the USA council per se. Another precedent would be AAA, which isn’t uniform across its many chapters. Feel free to split up the item more precisely, though it’ll probably be less disruptive if the current one remains focused on the chain.

Do you think we should introduce a bunch of operator entries to NSI, as we did with Goodwill? My rationale then was that Goodwill affiliates often put their name on the main sign or otherwise brand their region’s stores uniquely. I don’t know if that’s the case with SVdP, as I’m only familiar with one chapter’s thrift stores.

tylerfilla:
[...]

PR #10389, opened 2025-01-23
link shop brand, not entire religious order

matkoniecz:
see https://github.com/alltheplaces/alltheplaces/pull/12053

https://www.wikidata.org/wiki/Q7592120

https://en.wikipedia.org/w/index.php?title=St._Vincent_de_Paul_Thrift_Store&oldid=647143127 before it was merged

Disclaimer: I am not 1000% sure here

### DIFF
--- a/data/brands/shop/charity.json
+++ b/data/brands/shop/charity.json
@@ -595,7 +595,7 @@
       "locationSet": {"include": ["au"]},
       "tags": {
         "brand": "Vinnies",
-        "brand:wikidata": "Q117547405",
+        "brand:wikidata": "Q7592120",
         "name": "Vinnies",
         "shop": "charity"
       }


Snowysauce:
It seems like the right change to make, but the entry in question also has #9957 as an unresolved, potentially major issue. It'd be nice to roll out fixes for both the brand Wikidata QID and the store operators simultaneously, but it doesn't seem like the latter issue is anywhere close to a resolution. I'll leave this PR open for a couple of days in hopes of drawing renewed attention to #9957.

matkoniecz:
well, this PR does not resolve #9957 but makes it slightly less worse, I think

Though given #9957 it may be better to remove this entry outright to stop encouraging bad edits.

bhousel:
Oh yeah, let's just get this fix in now and people can decide if they really want to start setting `operator` tags later.  I personally think it's silly to have items for all the regional operators, but not going to stop people from adding them.

matkoniecz:
Regional operators of a shared brand are silly (I seen some people creating also Lidl Poland and using it as `brand:wikidata` instead of Lidl - despite it being exactly the same brand).

But if regional operators also have own, actually distinct brand then it is dubious to tag the same `brand:wikidata`

Snowysauce:
After this PR was merged, I came across [Q120646672](https://www.wikidata.org/wiki/Q120646672). This might be a more appropriate `brand:wikidata` value given that the stores in this PR are branded as "Vinnies", not "St. Vincent de Paul".

PR #10390, opened 2025-01-23
link Q21100245 (Anicura) for Anicura entry, not Q110299350 (Animates)

matkoniecz:


matkoniecz:
I also fixed bogus data induced by NSI: https://overpass-turbo.eu/s/26Pf

PR #10393, opened 2025-01-24
Add Bayne's the Family Bakers

ClarSco:
Bayne's is a chain of bakeries in Scotland.

According to their website (https://baynes.co.uk/our-shops/), they have 69 branches across the country.


data/brands/shop/bakery.json
```diff
@@ -425,6 +425,20 @@
         "shop": "bakery"
       }
     },
+    {
+      "displayName": "Bayne's",
+      "locationSet": {"include": ["gb"]},
```
tommylung, line 430:
```suggestion
      "locationSet": {"include": ["gb-sct"]},
```

You can use `gb-sct` location set for Scotland.

ClarSco: While Bayne's currently only operate in Scotland, there's no indication that they wouldn't expand to the rest of the UK, so not sure if it would be appropriate to limit the geographic scope within NSI, especially as there doesn't appear to be any other "Bayne's" in the database.

If by convention, the NSI would prefer the narrower scope until it's appropriate to expand it, I'll accept this suggestion.

UKChris-osm: > If by convention, the NSI would prefer the narrower scope until it's appropriate to expand it, I'll accept this suggestion.

I'd probably narrow the scope for now, we can always expand it later very easily and quickly, but if a different Bayne's appeared in England before this company expanded, we might find this entry being added by mistake before we can catch up.

UKChris-osm:
You shouldn't need that many `matchNames`, as that tag is really there to match names that might already exist in OpenStreetMap, rather than as a place for all possible keyword combinations.

ClarSco:
Ah. Wasn't sure if it was to help match newly tagged objects as they were being added to the OSM (eg. to help iD suggest a preset for a manually added object with just a `name=*` tag) or just existing ones.

I'm open to trimming the list down, but I'm not comfortable enough with Overpass to use it to filter out the irrelevant ones, and not familiar enough with how `matchNames` handles special characters (eg. does it ignore apostrophes?).

Are you able to do this, and make the approriate suggestions?

UKChris-osm:
> Are you able to do this, and make the appropriate suggestions?

Yes I can take care of this, just leave this PR as it is and I'll make any edit before merging 👍

PR #10399, opened 2025-01-26
Update PL Bank BPS

praszuk:
Remove `name` from Bank BPS.

Bank BPS is more likely a parent organization that unites local banks.
Each local bank in the BPS group has its own name, typically formatted as "Bank Spółdzielczy in [area]", and manages its own branches in the surrounding areas.

The main bank, currently named Bank BPS, operates only [12 branches](https://www.bankbps.pl/znajdz-placowke). In contrast, there are over [300 local banks with more than 2,300 branches](https://www.bankbps.pl/aktualnosci/zrzeszenie-bps-rosnie-w-sile).

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -2281,11 +2281,11 @@
       "displayName": "Bank BPS",
       "id": "bankbps-68054b",
       "locationSet": {"include": ["pl"]},
+      "matchNames": ["spółdzielczy"],
       "tags": {
         "amenity": "bank",
         "brand": "Bank Polskiej Spółdzielczości",
-        "brand:wikidata": "Q9165001",
-        "name": "Bank BPS"
+        "brand:wikidata": "Q9165001"
       }
     },
     {
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -2278,14 +2278,13 @@
       }
     },
     {
-      "displayName": "Bank BPS",
-      "id": "bankbps-68054b",
+      "displayName": "Banki spółdzielcze Grupa BPS",
       "locationSet": {"include": ["pl"]},
-      "matchNames": ["spółdzielczy"],
+      "matchNames": ["Bank BPS", "spółdzielczy"],
       "tags": {
         "amenity": "bank",
-        "brand": "Bank Polskiej Spółdzielczości",
-        "brand:wikidata": "Q9165001"
+	    "network": "Grupa BPS",
+	    "network:wikidata": "Q86673022"
       }
     },
     {
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -2278,13 +2278,14 @@
       }
     },
     {
-      "displayName": "Banki spółdzielcze Grupa BPS",
+      "displayName": "Bank BPS",
+      "id": "bankbps-68054b",
       "locationSet": {"include": ["pl"]},
       "matchNames": ["Bank BPS", "spółdzielczy"],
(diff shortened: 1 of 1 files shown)


Snowysauce:
Since there are banks with the name "Bank BPS", the best approach is probably to leave the name preset as is, but add `preserveTags` to allow editors to override the default name suggestion.

Snowysauce:
After reading the second link, I'm not sure if this preset should continue to be used in its current fashion. As you mentioned, the main bank of the network is Bank BPS, and its branches carry the brand of the same name . But, as you also mentioned, the network also includes over 300 other banks, which I presume have their own set of names and what the NSI would consider a brand name. Even if we consider the network as one unit, from what I understand, the correct brand name for it would be "Banki Spółdzielcze z Grupy BPS" and should be treated as a distinct entry from the main bank.

praszuk:
@Snowysauce Thank you for the detailed response and for analyzing the sources!

> After reading the second link, I'm not sure if this preset should continue to be used in its current fashion

That’s exactly why I want to change it!

> Even if we consider the network as one unit, from what I understand, the correct brand name for it would be "Banki Spółdzielcze z Grupy BPS" and should be treated as a distinct entry from the main bank.

I am currently looking for advice on what would be better option here.
I've noticed that we already have many entries for [local banks](https://www.wikidata.org/w/index.php?title=Special:Search&limit=500&ns0=1&ns120=1&offset=0&profile=default&search=bank+sp%C3%B3%C5%82dzielczy). However, maintaining over 300 brands would require too much effort and could also be problematic for mappers and data consumers.

Some of these already exist on the wikidata,  so I don’t think it would be an issue, and I also don’t see any reason to remove them.
I’m leaning more toward the `network` approach, as you mentioned. This wouldn’t conflict with the local brands.

What do you think about `network` along with `network:wikidata`? Could we define something like this in the NSI project?

```json
    {
      "displayName": "Banki spółdzielcze Grupa BPS",
      "id": "bankbps-68054b",
      "matchNames": ["Bank BPS", "spółdzielczy"],
      "tags": {
        "amenity": "bank",
	"network": "Grupa BPS",
	"network:wikidata": "Q86673022"
      }
    },
```
<https://www.wikidata.org/wiki/Q86673022>

Then mappers could add local name and brand according to the Wikidata.

Snowysauce:
This is a bit late, but I don't think that the desired solution is technically possible within the existing framework of the NSI. Even if a `network` tree was created, the presets in that tree would not be automatically suggested if a matching `brand` preset also exists. (See [here](https://github.com/osmlab/name-suggestion-index/wiki/Technical-Details#matching) for more details as to why.) In addition, the NSI's lead developer currently has many IRL commitments, meaning the addition of new features are not a priority nor is something that's feasible for a while.

In lieu of a more thorough solution, I could merge the PR as proposed, but this would come at the expense of name standardization for the 12 branches of the main bank.

praszuk:
@Snowysauce 

> This is a bit late

There are many open PRs, and this one is not a high priority – it's OK :)

> tree would not be automatically suggested if a matching brand preset also exists. (See here for more details as to why.)

Correct me if I'm wrong, but as long as someone doesn't want to add a new entry for "local bank," there won't be any problem, right? If that's true, I think it's OK.
And even if someone does add it, the only problem would be that iD editor won't suggest it?

> In lieu of a more thorough solution, I could merge the PR as proposed, but this would come at the expense of name standardization for the 12 branches of the main bank.

If we can standardize some group/shared tags for more than 2k branches and lose only 12 "main branches," which can be manually tagged, then I think it's worth changing it.

PR #10407, opened 2025-01-28
Add brand "Wasserwacht"

as-eu:
"Wasserwacht" is a part of German Red Cross and is involved in lifeguard service in Germany. It's registered as an own [brand in Germany](https://register.dpma.de/DPMAregister/marke/register/3020190119777/DE).

The `locationSet` for Bavaria is necessary, because the local red cross there is named differently. See also this link (only in German):
[https://styleguide.drk.de/downloads/wasserwacht/bayerisches-rotes-kreuz-wasserwacht](https://styleguide.drk.de/downloads/wasserwacht/bayerisches-rotes-kreuz-wasserwacht )

The lifeguarding stations, etc. are operated by local subdivisons/members from the German Red Cross.


Snowysauce:
I'm not sure I understand the PR as currently coded. Each proposed entry has an assortment of `matchNames` to catch a wide net of OSM objects; however, they also have `"preserveTags": ["^name"]`, which means only the `brand` and `brand:wikidata` tags will be standardized by these presets. (The same is true for `"preserveTags": ["^official_name"]` - any existing value of `official_name` in OSM will be left alone even though it looks like the intent of the PR is to standardize this value.)

`brand:official_name` should probably be removed; it's [only used 17 times globally](http://overpass-turbo.eu/s/1XVe), and the NSI tends to restrict tags in presets to those that are well established.

as-eu:
Sorry for the late reply.

> which means only the `brand` and `brand:wikidata` tags will be standardized by these presets

The main purpose is to introduce the brand. The original name should be preserved because the name is often used for a proper name of the location, for which one has no connection to the "Wasserwacht" as brand and/or the local/regional section of the Red Cross as operator. 

A fictive example:
a water rescue station next to the Sunshine Lake
name: Water Rescue Station Sunshine Lake
brand: Wasserwacht
operator: Red Cross District Association Sunshine County

Maybe it's easier to understand with some real world examples:

- pretty nice tagged: Way [938858159](https://www.openstreetmap.org/way/938858159)
- brand is missing: Node [4193424225](https://www.openstreetmap.org/node/4193424225)

In both examples, it wouldn't be useful to standardize the name, imo.

as-eu:
> `brand:official_name` should probably be removed; it's [only used 17 times globally](http://overpass-turbo.eu/s/1XVe), and the NSI tends to restrict tags in presets to those that are well established.

I can understand this, and I have no problem removing the tag. My intention was to represent that the Wasserwacht is commonly known as Wasserwacht, but the full brand name is "Wasserwacht des Deutschen Roten Kreuzes".
Is there any possibility to make this visible? `operator:official_name` is not possibly because there are a lot of different local chapters of the Red Cross that operate the water rescue service.

Snowysauce:
> The main purpose is to introduce the brand.

IMO, then the proposed preset shouldn't contain `name` tags at all - introduced or preserved. If the preset doesn't contain a tag - whether it's `name` or something else - the value of the unspecified tag is preserved by default when applying the preset.

> I can understand this, and I have no problem removing the tag. My intention was to represent that the Wasserwacht is commonly known as Wasserwacht, but the full brand name is "Wasserwacht des Deutschen Roten Kreuzes". Is there any possibility to make this visible?

I'd guess that the only way to do so in OSM while using established tags is to make the value of `brand` "Wasserwacht des Deutschen Roten Kreuzes" and moving "Wasserwacht" to `brand:short`. If those values/tags are undesired, moving that information to Wikidata is probably the only other option.

as-eu:
> the proposed preset shouldn't contain `name` tags at all

I removed all name tags. I initially thought that there is an automatic matching to existent OSM entities - my fault.

> the value of brand "Wasserwacht des Deutschen Roten Kreuzes" and moving "Wasserwacht" to brand:short  

Changed accordingly.

Should I open an new PR? Or can this one reopened? @Snowysauce

Issue #10354, opened 2025-01-15
Add N Sport


zenfiric:
clothes brand
Serbia
https://www.n-sport.net/

PR #10409, opened 2025-01-29
Add N Sport #10354

zenfiric:
This pull request resolves https://github.com/osmlab/name-suggestion-index/issues/10354 by adding the tags for the N Sport brand.


zenfiric:
Resolves #10354

PR #10412, opened 2025-01-29
Amend Tata Motors

karen-kerr:
Tata Motors is [international](https://www.tatamotors.com/organisation/our-global-presence/)


Snowysauce:
I attempted to determine where Tata actually has dealerships (not just operations), and the only two countries I could verify were India and South Africa. An [overpass-turbo query](https://overpass-turbo.eu/s/1Y4t) shows potential dealerships in a handful of additional countries, but still nowhere near a global presence.

PR #10416, opened 2025-01-30
remove Barbeques Galore from shop=outdoor to shop=bbq

andrewharvey:
https://wiki.openstreetmap.org/wiki/Tag:shop=bbq

see also https://github.com/openstreetmap/id-tagging-schema/pull/1430 for the preset

### DIFF
--- a/data/brands/shop/bbq.json
+++ b/data/brands/shop/bbq.json
@@ -1,0 +1,21 @@
+{
+  "properties": {
+    "path": "brands/shop/bbq",
+    "exclude": {
+      "generic": ["^bbq$"]
+    }
+  },
+  "items": [
+    {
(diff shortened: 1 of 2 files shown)


UKChris-osm:
Looking at their [website](https://www.barbequesgalore.com.au/) they seem to sell a great deal more than just BBQ's, is BBQ really the most suitable tag?

andrewharvey:
I've visited a few branches, they mainly sell BBQs and BBQ accessories, they do also sell some outdoor furniture, but still I would tag this as shop=bbq as the primary product focus. A lot of the things you see on their website might be "online only" or only available in selected stores or in a small part of the store.

shop=bbq is a much better fit than shop=outdoor which is a "Shop selling trekking, climbing, camping equipment".

PR #10439, opened 2025-02-01
[amenity/fuel][shop/gas] add Aygaz (TR)

panasonic-mafia:
https://www.aygaz.com.tr/. 

Almost no entries in OSM, but it's a biggest `shop=gas` dealer in Turkey, and also they have a [network of LPG filling stations for cars](https://kurumsal.aygaz.com.tr/otogaz/en-yakin-otogaz-istasyonlari).


panasonic-mafia:
Thanks for updates @kjonosm!

PR #10441, opened 2025-02-01
[brand/amenity/cafe] add Özsüt (TR)

panasonic-mafia:
https://ozsut.com.tr/en.

Around 80 cafes, worth adding.

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -1797,10 +1797,8 @@
       "tags": {
         "amenity": "cafe",
         "brand": "Özsüt",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Is `name:tr` needed if the `name` value also uses the native name?

panasonic-mafia:
> Is name:tr needed if the name value also uses the native name?

Probably not, thanks.

panasonic-mafia:
Don't know why I completely forgot [this guidance](https://wiki.openstreetmap.org/wiki/Multilingual_names#Repeating_name_with_language_specific_tag).

PR #10456, opened 2025-02-03
Added Teahouse chain.

NLBRT:
Bombay Cutting Chai is an Indian teahouse chain, primarily operating in Northern parts of the Indian State of West Bengal, as well as having an outlet in Qatar. Their online presence is quite shabby IMO, but they do exist and have >25 active outlets at the time of making this PR.


UKChris-osm:
We tend not to stray into tagging lots of things in the NSI, and focus on `brand` and `operator` tags only (with the odd exception) so tags like:

```
        "air_conditioning": "yes",
        "payment:cash": "yes",
        "contact:website":"https://arambaghfoodmart.com/",
        "contact:email": "customercare@arambaghfoodmart.com",
        "website:map": "https://arambaghfoodmart.com/our-footprints/"
```

... aren't something we'd include.

The tagging for `payment` is usually okay for things like payment terminals (such as vending or ticket machines), and if I'm honest, I can't remember what was decided about the `diet` tagging, so hopefully someone else can chip in with that information.

NLBRT:
Ah, I'll take note of that and remove those!

As for the key `diet:*=*` (more specifically `diet:vegetarian=*`, `diet:vegan=*` and `diet:jain=*`), I'd argue that in regions like India, many cafe/restaurant/fast food chains serve only vegetarian food and reflect the same through their branding, be it the name, or their logo, or other visual assets (such as ads). This can also go for `organic=*`.

UKChris-osm:
>  many cafe/restaurant/fast food chains serve only vegetarian food and reflect the same through their branding

I don't disagree, but we've had the same discussions regarding brands that always offer Wi-Fi, or free parking, etc, and it was more down to a concern that the index as a whole was getting a bit too big with too much data to manage. For example, what if a brand stopped offering Wi-Fi, or added a non-vegan range, and we didn't noticed?

NLBRT:
Only makes sense on the amenity side of things, not sure about the vegan restaurant adding non-vegan ranges though. Whatever it might be, I think this can be merged.

PR #10465, opened 2025-02-03
Update vending_machine.json

Mannivu:
Add brand:wikidata to Aqvagold brand


UKChris-osm:
Could the `displayName` be changed as well to match the `brand`?

Mannivu:
Sure, I've updated the file.

PR #10499, opened 2025-02-08
Added some Israel chains

Flatroy:
Hi
Added some more Israeli chains and aligned some existing ones

### DIFF
--- a/data/brands/shop/cosmetics.json
+++ b/data/brands/shop/cosmetics.json
@@ -469,13 +469,15 @@
     },
     {
       "displayName": "Lallin",
-      "id": "laline-e301c4",
-      "locationSet": {"include": ["il"]},
+      "id": "laline-255338",
+      "locationSet": {
+        "include": ["ca", "ge", "il", "jp", "tw"]
+      },
       "matchTags": ["shop/gift"],
       "tags": {
         "brand": "Laline",
         "brand:wikidata": "Q55649188",
-        "name": "ללין",
+        "name": "Laline",
         "name:en": "Laline",
         "name:he": "ללין",
         "shop": "cosmetics"
--- a/data/brands/shop/cosmetics.json
+++ b/data/brands/shop/cosmetics.json
@@ -473,14 +473,12 @@
       "locationSet": {
         "include": ["ca", "ge", "il", "jp", "tw"]
       },
-      "matchNames": ["lallin"],
+      "matchNames": ["lallin", "ללין", "ラリン"],
       "matchTags": ["shop/gift"],
       "tags": {
         "brand": "Laline",
         "brand:wikidata": "Q55649188",
         "name": "Laline",
-        "name:en": "Laline",
-        "name:he": "ללין",
         "shop": "cosmetics"
       }
     },
--- a/data/brands/shop/cosmetics.json
+++ b/data/brands/shop/cosmetics.json
@@ -468,11 +468,12 @@
       }
     },
     {
-      "displayName": "Lallin",
+      "displayName": "Laline",
       "id": "laline-255338",
       "locationSet": {
         "include": ["ca", "ge", "il", "jp", "tw"]
       },
+      "matchNames": ["Lallin"],
       "matchTags": ["shop/gift"],
       "tags": {
         "brand": "Laline",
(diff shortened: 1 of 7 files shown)


data/brands/shop/cosmetics.json
```diff
@@ -469,13 +469,15 @@
     },
     {
       "displayName": "Lallin",
```
kjonosm, line 471: `displayName` differs from `brand` and `name`

data/brands/shop/mobile_phone.json
```diff
@@ -340,6 +355,21 @@
         "shop": "mobile_phone"
       }
     },
+    {
+      "displayName": "Hot Mobile",
```
kjonosm, line 359: `displayName` differs from `brand` and `name`

data/brands/shop/mobile_phone.json
```diff
@@ -685,6 +715,36 @@
         "shop": "mobile_phone"
       }
     },
+    {
+      "displayName": "Partner",
```
kjonosm, line 719: `displayName` differs from `brand` and `name`

data/brands/shop/mobile_phone.json
```diff
@@ -685,6 +715,36 @@
...
+        "brand:he": "פרטנר",
+        "brand:wikidata": "Q2000500",
+        "name": "פרטנר",
+        "name:en": "Partner",
+        "name:he": "פרטנר",
+        "shop": "mobile_phone"
+      }
+    },
+    {
+      "displayName": "Pelephone",
```
kjonosm, line 734: `displayName` differs from `brand` and `name`

data/brands/shop/supermarket.json
```diff
@@ -10670,6 +10670,7 @@
       "displayName": "יינות ביתן",
       "id": "yeinotbitan-6b65f1",
       "locationSet": {"include": ["il"]},
+      "matchNames": ["yaynot bitan"],
       "tags": {
         "alt_name:en": "Yenot Bitan; Bitan Wines",
```
kjonosm, line 10675: Please remove space after semicolon

Flatroy: hi, thanks for review!
Where is this space? I don't see it

Snowysauce: The space following the semicolon is in the value for `alt_name:en`. Semicolons separating values of keys should not have a space following them per OSM conventions.

data/brands/shop/houseware.json
```diff
@@ -263,8 +263,8 @@
     },
     {
       "displayName": "FOX Home",
-      "id": "foxhome-a5093f",
-      "locationSet": {"include": ["il"]},
+      "id": "foxhome-43585e",
+      "locationSet": {"include": ["ca", "il"]},
```
Snowysauce, line 267: The Canadian stores should probably have a separate entry in the NSI that doesn't include the multilingual name tags. Also, the specified Wikidata item currently does not have any statements related to stores outside of Israel.

data/brands/shop/cosmetics.json
```diff
@@ -468,14 +468,17 @@
       }
     },
     {
-      "displayName": "Lallin",
-      "id": "laline-e301c4",
-      "locationSet": {"include": ["il"]},
+      "displayName": "Laline",
+      "id": "laline-255338",
+      "locationSet": {
+        "include": ["ca", "ge", "il", "jp", "tw"]
```
Snowysauce, line 474: This `locationSet` is highly problematic considering that each of the five countries listed not only have different primary languages, but different language scripts. If the brand does indeed operate in these countries, each country should probably have its own localized entry in the NSI.

data/brands/shop/supermarket.json
Snowysauce: Since the issue was in the NSI base code, not this PR, I removed the space after the semicolon as part of https://github.com/osmlab/name-suggestion-index/commit/d48cff6232c8591a36c45876e21ffd3ea980ee27. (Side note: apparently that was the only instance of a space after a semicolon-delimited value in the entire NSI - a project-wide search returned only this specific case.)

data/brands/shop/cosmetics.json
Flatroy: What if we remove this part for Hebrew:
`"name:he": "ללין"``

and leave only the English version of the brand name? will it be ok for all locations? because the brand is always Laline for all countries and we can add the Hebrew name to matchNames for search

Snowysauce: That solution would be OK.

PR #10508, opened 2025-02-10
Update Djak sport

zenfiric:


Snowysauce:
This brand was just added to the NSI by way of your PR in #10473. I can't imagine that many of us at the NSI are familiar with the brand, so a bit of context to the changes made here would be appreciated.

Also, although this has little to do with this PR, I'm curious how the name "Djak sport" is related to the brand when none of the name tags carry that value.

zenfiric:
> This brand was just added to the NSI by way of your PR in #10473. I can't imagine that many of us at the NSI are familiar with the brand, so a bit of context to the changes made here would be appreciated.

@Snowysauce I did a double-check, and the company is registered as "Đak d.o.o.", but they often go under "Đak sport" name. So I reverted back to plane Đak naming. The updates to `locationSet` are still valid

> Also, although this has little to do with this PR, I'm curious how the name "Djak sport" is related to the brand when none of the name tags carry that value.

Well, Đ in latin script is written as Dj if there are no diacritics, and this type of writing can occur in the countries mentioned in `locationSet`

Snowysauce:
Thanks for the added info. I looked deeper into their website, and they confirmed the suggested operating area in their "about us" page. Everything looks OK to merge; however, in order for the preset to still match for "Đak Sport", that name needs to be added to `matchNames` since it's no longer the `brand` value.

PR #10511, opened 2025-02-12
Expand countries of Insomnia Cookies

impiaaa:
Per their website: https://insomniacookies.com/countries

### DIFF
--- a/data/brands/shop/pastry.json
+++ b/data/brands/shop/pastry.json
@@ -161,7 +161,7 @@
     {
       "displayName": "Insomnia Cookies",
       "id": "insomniacookies-64e364",
-      "locationSet": {"include": ["us"]},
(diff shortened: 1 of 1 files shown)


Snowysauce:
I was able to confirm the chain's presence in Canada through their website, but all of the UK links I tried didn't work. That said, an overpass-turbo query shows one location in Canada (Toronto, verified by the chain's website) and in England (Manchester), so the change suggested by this PR is probably correct.

impiaaa:
ah sorry, just to clear up confusion, I added the one in Manchester, trusting their website. They do have a press release about it though https://www.restaurantonline.co.uk/Article/2023/08/30/Insomnia-Cookies-makes-UK-debut-with-double-Manchester-opening

Snowysauce:
The UK leg of their website is back up, and their store locator confirmed their presence in Manchester, England.

Issue #10507, opened 2025-02-10
Add recycling container operated by Viacor


andris86:
In Riga, Latvia I have found many of Viacor recycling containers which accept clothes, shoes and textiles. There are so many of them that I would like a preset created representing them. Many still remain unmapped.

Following tags should be present in this preset:

- amenity=recycling
- recycling_type=container
- operator=Viacor
- operator:email=viacorcharity@gmail.com
- operator:phone=+371 28 167 221
- operator:website=https://www.viacor.lv/
- recycling:clothes=yes
- recycling:shoes=yes
- recycling:textiles=yes

I don't know whether they are found in other countries so make this preset available only in Latvia for now.

Snowysauce:
A search of Wikidata for "Viacor" returns only [Q132104352](https://www.wikidata.org/wiki/Q132104352), which is apparently a French company by that name. I have no idea if that company has any relation to the containers in Latvia, but if not, another Wikidata item needs to be made to fulfill this issue. In any event, the suggested `operator:*` tags are not accepted by the NSI; those data points should be on its Wikidata item instead.

Snowysauce:
Looks good to me. I tweaked and added a couple things after browsing through the company's website.

PR #10514, opened 2025-02-12
Add ViaCor Charity

andris86:
https://github.com/osmlab/name-suggestion-index/issues/10507

### DIFF
--- a/data/brands/amenity/recycling.json
+++ b/data/brands/amenity/recycling.json
@@ -394,6 +394,19 @@
       }
     },
     {
+      "displayName": "ViaCor Charity",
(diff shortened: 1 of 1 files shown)


Snowysauce:
This PR would add the entry to `brands/amenity/recycling.json`; however, the entry has `operator` tags, not `brand` tags. Should the entry be added to `operators/amenity/recycling.json` instead?

Snowysauce:
Added entry to `operators/amenity/recycling.json` in https://github.com/osmlab/name-suggestion-index/commit/566e2fff021be01734091cc1fb40ba5016d0d78e

PR #10524, opened 2025-02-14
update some german power and grid operators

Johann150:
See comments in the individual commits for details/justification for the changes.


data/operators/man_made/pipeline.json
```diff
@@ -6914,6 +6903,7 @@
       "locationSet": {
         "include": ["de-sn.geojson"]
       },
+      "matchNames": ["ensonetz"],
```
kjonosm, line 6906: Please check whether the matchName is correct. It seems that ensoNetz is followed by Sachsennetze, see https://www.wikidata.org/wiki/Q124370321

data/operators/power/line.json
```diff
@@ -7145,7 +7132,8 @@
       "matchNames": [
         "enso",
         "enso netz",
-        "sachsennetze"
+        "sachsennetze",
```
kjonosm, line 7135: Please check whether the matchName is correct, see https://www.wikidata.org/wiki/Q124370321

data/operators/power/substation.json
```diff
@@ -8363,7 +8350,8 @@
       "matchNames": [
         "enso",
         "enso netz",
-        "sachsennetze"
+        "sachsennetze",
```
kjonosm, line 8353: Please check whether the matchName is correct, see https://www.wikidata.org/wiki/Q124370321

data/operators/power/line.json
Johann150: SachsenNetze is a 100% owned subsidiary of SachsenEnergie (grid operation and electricity distribution have to be legally separate).

SachsenEnergie was previously placed here so I thought it would have been previously established that this should be used.

Snowysauce:
Re Wesernetz Bremen GmbH: the `matchNames` value is identical to the `operator` value and will be removed by the build script post-merge. Is the `matchNames` value supposed to be looking for something else instead (e.g., "Wesernetz Bremen")?

Snowysauce:
The workflow tests indicate that these changes are missing `matchNames` that would cover two operators with 10 or more instances in OSM:

```
🏗   Merging items...
🍔  brands:	17653 total, 0 new
🚩  flags:	442 total, 0 new
💼  operators:	23784 total, 2 new
🚇  transit:	5757 total, 0 new
```

I can't be 100% positive about which changes are causing the "new" entries to be generated, but if I had to guess, I would say that it's the elimination of "Wesernetz" in `power/line` and the rename of "Wesernetz" in `power/substation`. If this PR was merged as-is, the "Wesernetz" entries would be re-added by the build script on its first run post-merge.

bhousel:
Because this had gotten pretty old, it wasn't able to merge cleanly, so I tried to add the same changes myself in 0be3f62c2.
I ran the build and it didn't create any new entries, so I think all the matching is probably working as expected. 👍

PR #10529, opened 2025-02-16
Fix ESLint after v9 upgrade; enforce in CI

loganrosen:
I noticed that the `lint` task wasn't working since ESLint was upgraded to v9, as it uses a new config format. I've followed [the migration guide](https://eslint.org/docs/latest/use/migrate-to-9.0.0) to migrate to the new config format and fixed the lint violations. I've also adjusted the GitHub Actions workflow to enforce ESLint in CI, so we can catch linting issues earlier in the development process.


Snowysauce:
This is a bit above my level of coding, but it looks like all the proper steps were taken.

PR #10531, opened 2025-02-16
Fix Pancheros name

aenet14:
Removed apostrophes from displayname and tags brand and name for Pancheros Mexican Grill to match wikidata and official Pancheros website

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -5127,15 +5127,15 @@
       }
     },
     {
-      "displayName": "Panchero's Mexican Grill",
+      "displayName": "Pancheros Mexican Grill",
       "id": "pancherosmexicangrill-96af40",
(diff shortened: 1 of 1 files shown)


Snowysauce:
Thanks for noticing the error. I corrected the name on Wikidata while evaluating the PR that added the brand to the NSI, but I completely failed to notice that the brand had the wrong values in the PR itself. The next NSI release isn't scheduled for a week, so hopefully there won't be too much damage to the OSM data in the interim.

aenet14:
I noticed it while I was correcting warnings on OSM. There’s a Pancheros
location near me that it said was spelled incorrectly, but Mapillary showed
it without the apostrophe, which led to me to check their website and the
wikidata, which both lined up with not having an apostrophe. So instead of
changing the location, I went I changed the NSI instead.

PR #10532, opened 2025-02-16
Add Waffle Cabin

akadouri:
US chain selling waffles at ski resorts https://wafflecabin.com/

I tried to create a wikipedia page but it wasn't approved https://en.wikipedia.org/wiki/Draft:Waffle_Cabin


Snowysauce:
Don't worry about the English Wikipedia; the only WMF "product" we use at the NSI is Wikidata, and the chain's item there looks good. The only notability standard we have here is outlined at the [Judge Case wiki page](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#notability), and this brand passes it easily.

PR #10533, opened 2025-02-17
Update atm.json (New Brand Entry)

Mukhtar27:
Adding New ATM brand in UAE
https://www.wikidata.org/wiki/Q20829802

PR #10534, opened 2025-02-17

Mukhtar27:
Adding New Brand Entry in UAE
https://www.wikidata.org/wiki/Q60572527

### DIFF
--- a/data/brands/amenity/atm.json
+++ b/data/brands/amenity/atm.json
@@ -517,4 +517,22 @@
       }
     }
   ]
+},
+{
+    "displayName": "Emirates Islamic",
+    "locationSet": {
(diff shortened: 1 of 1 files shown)


Snowysauce:
The banks in #10533 and this PR have branch locations in addition to ATMs, so it'd be highly unusual to add their ATM networks to the NSI without also adding their branch locations under `brands/amenity/bank`. (Adding the banks themselves would also make editing the ATM category file unnecessary as ATM presets are generated for banks by default.)

Snowysauce:
Closing per https://github.com/osmlab/name-suggestion-index/pull/10560#issuecomment-2671536734

PR #10539, opened 2025-02-18
add Emack and Bolio's

akadouri:
Ice cream chain with locations in the US and several other countries https://emackandbolios.com/locations/


Snowysauce:
The map up top shows six countries, but the location list itself shows locations in only three. Please edit the `locationSet` to only include the US, China, and South Korea.

PR #10541, opened 2025-02-18
Update fuel.json New Fuel Entr

Mukhtar27:
Added new fuel brand in UAE


Snowysauce:
This brand is already in the NSI: https://nsi.guide/index.html?t=brands&k=amenity&v=fuel&tt=emarat#emarat-b7987c

PR #10542, opened 2025-02-18
Update financial.json new entry

Mukhtar27:
Added new entry in Forex in UAE
https://www.wikidata.org/wiki/Q132524657


Snowysauce:
Brand was added in #10563

PR #10543, opened 2025-02-18
Update fuel.json New Fuel Entry

Mukhtar27:
Added new fuel brand in UAE
https://www.wikidata.org/wiki/Q3045998


Snowysauce:
This brand is already in the NSI: https://nsi.guide/index.html?t=brands&k=amenity&v=fuel&tt=enoc#enoc-7d34dd

PR #10545, opened 2025-02-18
Update car_repair.json/update office=association and added official name

BruniSchulz:
shop=car_repair was updated to office=association
official name 'Österreichischer Automobil-, Motorrad- und Touring Club' was added

### DIFF
--- a/data/brands/shop/car_repair.json
+++ b/data/brands/shop/car_repair.json
@@ -1029,7 +1029,8 @@
         "brand": "ÖAMTC",
         "brand:wikidata": "Q306057",
         "name": "ÖAMTC",
-        "shop": "car_repair"
+        "office": "association",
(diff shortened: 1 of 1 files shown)


Snowysauce:
Thanks for your contribution. In order to change the category of a preset, the entry needs to be moved from the old category's file to the new, desired category's file (in this case `brands/office/association.json`). Otherwise, the build script will automatically re-add the old category's tag to the preset and treat the desired category's tag as an extra tag.

BruniSchulz:
I've submitted a new request for this update:
 Update association.json (update office=association and added official name) #10551

Snowysauce:
Tagging as "needs discussion" per https://github.com/osmlab/name-suggestion-index/pull/10551#issuecomment-2665928041

BruniSchulz:
Sorry for some reason I opened multiple pull requests for the same brand, I close all of them and create a new one with correct edits.

PR #10546, opened 2025-02-18
[amenity/bank] add Volksbank Niederösterreich

BruniSchulz:
For Austria (AT) I have added a new bank brand 'Volksbank Niederösterreich'

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -19402,20 +19402,20 @@
         "name:en": "Kagoshima Bank",
         "name:ja": "鹿児島銀行"
       }
+    },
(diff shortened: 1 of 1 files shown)


Snowysauce:
The brand looks notable on its own, but I noticed that there's also an Austrian bank called "[Volksbank](https://www.wikidata.org/wiki/Q695110)". Is there any relation between Volksbank Niederösterreich and Volksbank, and if so, what makes the two distinct enough to be separate brands?

BruniSchulz:
@Snowysauce thanks for the feedback, I totally agree. I adjusted the file with a central entry for Austria.

PR #10550, opened 2025-02-18
Update bank.json (added for AT)

BruniSchulz:
For AT I added bank brand '3 Banken Gruppe'


Snowysauce:
All three of the banks that comprise 3 Banken Gruppe have their own NSI presets, and an overpass-turbo query for "3 Banken Gruppe" returns no results.

PR #10551, opened 2025-02-18
Update association.json (update office=association and added official name)

BruniSchulz:
as explained in the comment in request 'Update car_repair.json/update office=association and added official name #10545'
shop=car_repair was updated to office=association
official name 'Österreichischer Automobil-, Motorrad- und Touring Club' was added

I moved the entry from the old category's file to the new category's file brands/office/association.json.

### DIFF
--- a/data/brands/office/association.json
+++ b/data/brands/office/association.json
@@ -203,4 +203,24 @@
       }
     }
   ]
+},
+{
+    "displayName": "ÖAMTC",
+    "id": "oamtc-d1144d",
+    "locationSet": {
+        "include": [
+            "at"
+        ]
+    },
+    "preserveTags": [
+        "^name",
+        "^official_name"
+    ],
+    "tags": {
+        "brand": "ÖAMTC",
+        "brand:wikidata": "Q306057",
+        "name": "ÖAMTC",
+        "office": "association",
+        "official_name": "Österreichischer Automobil-, Motorrad- und Touring Club"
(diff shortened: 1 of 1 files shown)


Snowysauce:
```
    "preserveTags": [
        "^name",
        "^official_name"
    ],
```

will cause OSM editors to ignore any existing values of `name` and `official_name` when applying this preset. Since the `official_name` tag is being added by this PR, I'm assuming that the intention is to have a standardized value for the tag. Using `preserveTags` would have the opposite effect.

Snowysauce:
Also, the [overpass-turbo query](https://overpass-turbo.eu/s/1YYM) I linked to in #10545 returns only one result for `name=ÖAMTC` and `office=association`, and that result is co-tagged as `shop=car_repair`.

BruniSchulz:
Closing this one as well, sorry, still learning GitHub basics.

BruniSchulz:
> Also, the [overpass-turbo query](https://overpass-turbo.eu/s/1YYM) I linked to in https://github.com/osmlab/name-suggestion-index/pull/10545 returns only one result for name=ÖAMTC and office=association, and that result is co-tagged as shop=car_repair.

As a local I see that it is not really a car_repair but rather an office which does inspections. similar to [ADAC](https://nsi.guide/index.html?t=brands&k=office&v=association&tt=adac) in Germany.

PR #10552, opened 2025-02-18
Update Zegna

cicku:
https://www.zegnagroup.com/wp-content/uploads/2021/12/ZEGNA-THE-REBRANDING-PRESS-RELEASE-ENG.pdf


Snowysauce:
This change seems fine, but "Ermenegildo Zegna" should (and probably needs) to be added as a matchName.

cicku:
Done

PR #10556, opened 2025-02-19
Update bank.json New bank add

Mukhtar27:
Added new bank in UAE
https://www.wikidata.org/wiki/Q130726987

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -5888,8 +5888,7 @@
       "displayName": "Dubai Islamic Bank",
       "id": "dubaiislamicbank-2a7aac",
       "locationSet": {
-        "include": ["ae"],
-        "exclude": ["bg"]
+        "include": ["ae","pk","ke","jo"]
       },
       "tags": {
         "amenity": "bank",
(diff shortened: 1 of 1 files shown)


Snowysauce:
Please stop placing new code at the very end of the file; this is the cause of the failed code checks. For guidance on where the code can/should be placed, please look at my edit to #10544.

Mukhtar27:
@Snowysauce , thanks for the help. I am very sorry for adding it like that. Though, i would place the way you have done, do i need to add at the end.
  ]
}

Snowysauce:
I think the easiest way to explain where new code can go is by showing an example. As of this writing, here is the last entry in `brands/amenity/bank.json`:

https://github.com/osmlab/name-suggestion-index/blob/2e9e16f10eae6544caa165c5c98fd6551df7aac8/data/brands/amenity/bank.json#L19466-L19480

A new entry can be added between lines 19480 and 19481 by adding a comma to the end of line 19480, and inserting the new code right afterward (either right after the comma or on a new line below it - the code tester and build script can handle either case). Your code was structurally fine, so if you place future requests in the suggested location (or its equivalent in the other category files), everything should be good.

As for the "] }" down at the end (lines 19481 and 19482 in the example above), those are the two bits of code that essentially close out or "seal up" the file, and so anything placed in between or below those braces will cause code errors.

Mukhtar27:
Thanks @Snowysauce for your support.

PR #10560, opened 2025-02-19

Mukhtar27:
Added ATM in UAE


Snowysauce:
I've tried to be patient, but at this point I'll be blunt - this brand has to be added under `brands/amenity/bank`. Closing.

PR #10561, opened 2025-02-19
[office/association] add ÖAMTC

BruniSchulz:
referring to request Update association.json (update office=association and added official name) #10551
ÖAMTC needs to be captured as an association , currently it has category shop=car_repair which is incorrect

### DIFF
--- a/data/brands/office/association.json
+++ b/data/brands/office/association.json
@@ -203,4 +203,19 @@
       }
     }
   ]
+},
+{
+    "displayName": "ÖAMTC",
+    "locationSet": {
+        "include": [
+            "at"
+        ]
+    },
+    "tags": {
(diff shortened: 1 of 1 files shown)


Snowysauce:
I moved the code to a valid location in the file.

BruniSchulz:
This statement '_Also, the suggested code addition is placed in an invalid location in the file. If I let the code check workflow run, the checks would fail. Please see the edit I made to a similar error in https://github.com/osmlab/name-suggestion-index/pull/10544 for guidance on where new code needs to be placed'_ is linked to a request that was created by another user.
Regarding 'ÖAMTC': I understand thta it is not possible to change the category.
Therefore I will close this PR.

PR #10564, opened 2025-02-19
Update car.json add brand Mercedes-Benz Vans

BruniSchulz:
add brand Mercedes-Benz Vans which is an own brand within the Mercedes-Benz Group

### DIFF
--- a/data/brands/shop/car.json
+++ b/data/brands/shop/car.json
@@ -1,4 +1,20 @@
 {
+    "displayName": "Mercedes-Benz Vans ",
+    "locationSet": {
+        "include": [
+            "de",
+            "at",
+            "ch"
+        ]
+    },
+    "tags": {
+        "brand": "Mercedes-Benz Vans",
+        "brand:wikidata": "Q55383469",
+        "name": "Mercedes-Benz Vans",
+        "shop": "car"
+    }
+},
+{
   "properties": {
     "path": "brands/shop/car",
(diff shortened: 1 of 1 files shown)


Snowysauce:
An overpass-turbo query for "Mercedes-Benz Vans" returned no results. Are the vans sold at specialized dealerships or at Mercedes-Benz dealers that also sell passenger cars?

BruniSchulz:
the vans are sold at Mercedes-Benz dealers that sell passenger car and vans and there are dealers who only sell vans

Snowysauce:
> there are dealers who only sell vans

How many of these dealerships exist, and are the dealerships themselves branded any differently than a typical Mercedes-Benz dealer? There needs to be [at least 10-30 locations](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#notability) with the "Mercedes-Benz Vans" brand in order for this brand to be included in the NSI. The NSI is an OSM editing "assistant", for lack of a better term, and our purpose is to collect data only for entities whose presets can and will be used in OSM.

panasonic-mafia:
I see Mercedes-Benz Vans more as a product that is sold at Mercedes-Benz dealerships, rather than a separate dealerships. So probably it shouldn't be a separate NSI entry for this.

BruniSchulz:
@Snowysauce you are right from this perspective it is more a product, I agree.
I will close this request.

PR #10569, opened 2025-02-20
Update interior_decoration.json (new brand entry)

Amar2375:
Adding new entry for a brand in Malaysia

https://www.tupperwarebrands.com.my/index
and
https://www.wikidata.org/wiki/Q1129117


Snowysauce:
The brand meets NSI inclusion criteria per https://shop.tupperwarebrands.com.my/a/store-locator/list, but it may be better under `brands/shop/houseware`.

PR #10571, opened 2025-02-20
[amenity/cafe] add Malaysia to Boost Juice

Amar2375:
Adding new country to existing brand in Malaysia
https://www.boostjuicebars.com.my/
and
https://www.wikidata.org/wiki/Q4943789


Snowysauce:
Countries are added to existing brands by editing the `locationSet` of the existing preset, not by creating a duplicate preset.

PR #10573, opened 2025-02-20
Add Kenzo (LVMH)

cicku:


Snowysauce:
Adding as proposed per https://www.kenzo.com/on/demandware.store/Sites-KENZO_US2-Site/en_US/Stores-AllStores. The `locationSet` could theoretically be narrowed to seven items, but that's bordering on being a bit cumbersome.

PR #10574, opened 2025-02-20
Update medical_supply.json (New brand entry)

Amar2375:
Adding new entry for Brand in Malaysia for Medical supply
https://www.euyansang.com.my/
and
https://www.wikidata.org/wiki/Q4207246


Snowysauce:
This brand is already in the NSI as https://nsi.guide/?t=brands&k=shop&v=herbalist#euyansang-7be159.

PR #10150, opened 2024-11-18
Merge duplicated Yves Rocher

urbalazs:

PR #10575, opened 2025-02-20
Add Vilebrequin

cicku:


Snowysauce:
Is this request referring to [Q3558391](https://www.wikidata.org/wiki/Q3558391)? The specified QID belongs to another brand that's already in the NSI.

If this is for Q3558391, the global `locationSet` is fine; their store locator returned far too many countries and regions to individually list.

cicku:
Somehow I did not find that existing entry. However, I personally think that putting it to sporting goods is better than generic clothes, since they are mostly a swimwear company.

What do you think?

Snowysauce:
Personally, I'd probably leave the existing preset as-is (but expand the `locationSet` to worldwide as established by this discussion). The website does make it clear that their roots are in swimwear, but they also appear to carry a full selection of non-swimwear outerwear designed for "the beach life" (e.g., sweatshirts).

However, I'm probably not the best person to offer an opinion on whether to use an umbrella category (clothes) or something more specific (sports, clothes=swimwear). We've had an open PR for Yves Rocher since November over whether that brand should be "shop=beauty" (umbrella) or "cosmetics" (specific) - #10150.

PR #10576, opened 2025-02-20
Update restaurant.json (New brand entry)

Amar2375:
Adding new entry for Brand in Malaysia for restaurant
https://www.4fingers.com.my/
and
https://www.wikidata.org/wiki/Q23043391


Snowysauce:
This brand is already in the NSI as https://nsi.guide/?t=brands&k=amenity&v=fast_food#4fingerscrispychicken-aadea7.

PR #10578, opened 2025-02-20
[shop/electronics] add Harvey Norman

Amar2375:
Adding new entry for Brand in Malaysia for electronics
https://www.harveynorman.com.my/
and
https://www.wikidata.org/wiki/Q4040441

### DIFF
--- a/data/brands/shop/electronics.json
+++ b/data/brands/shop/electronics.json
@@ -2468,7 +2468,8 @@
       "displayName": "Harvey Norman",
       "locationSet": {
           "include": [
-              "my"
+              "my",
+              "au"
           ]
       },
       "tags": {
--- a/data/brands/shop/electronics.json
+++ b/data/brands/shop/electronics.json
@@ -2465,4 +2465,20 @@
       }
     }
   ]
+},
+{
+    "displayName": "Harvey Norman",
+    "locationSet": {
+        "include": [
+            "my"
+        ]
+    },
+    "tags": {
+        "brand": "Harvey Norman",
+        "brand:en": "Harvey Norman",
+        "brand:wikidata": "Q4040441",
+        "name": "Harvey Norman",
+        "name:en": "Harvey Norman",
+        "shop": "electronics"
+    }
 }
--- a/data/brands/shop/electronics.json
+++ b/data/brands/shop/electronics.json
@@ -2463,22 +2463,20 @@
         "name:ja": "駿河屋",
         "shop": "electronics"
       }
+    },
+    {
+      "displayName": "Harvey Norman",
+      "locationSet": {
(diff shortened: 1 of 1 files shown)


Snowysauce:
This brand is already in the NSI as https://nsi.guide/?t=brands&k=shop&v=department_store#harveynorman-b73e73. The `locationSet` of the existing preset does not include Malaysia, so I'll leave this PR open in case the brand should be tagged differently there (it's set as `shop=department_store` elsewhere).

andrewharvey:
The wiki say `shop=electronics` is

> Shop selling consumer electronics (TV, radio, fridge)
> 
> May focus on big household appliances washing machines, fridges, ovens, etc. Most electronics shops will also sell related peripheral devices, cables, batteries, etc. May sell second-hand electronics. May offer repair services.

In Australia that's what they do, household appliances, TVs, fridges etc. They are sometimes located inside a Domayne shop which does the furniture. I think we could probably change the AU entry to `shop=electronics` for the locations I'm familiar with, I just don't know if this can apply to every store.

andrewharvey:
Our local discord channel indicated that not all Harvey Norman stores in Australia are the same. Some only do electronics (`shop=electronics`) others to both electronics and furniture (hence `department_store`).

Based on this I'm fine for the MY entry to have it's own entry based on how their stores are, so I think this PR is good.

As for the AU entries, I wonder if we can have the brand appear in both shop/electronics and shop/department_store? So in iD both are presented as options and user would pick, and then also users aren't asked to mass retag from one to the other (only as a suggestion to add `operator:wikidata`) Does that work?

Snowysauce:
> As for the AU entries, I wonder if we can have the brand appear in both shop/electronics and shop/department_store? So in iD both are presented as options and user would pick, and then also users aren't asked to mass retag from one to the other (only as a suggestion to add `operator:wikidata`) Does that work?

I'm fairly confident that it's doable from a technical standpoint per #10150, where we've (unintentionally) had duplicate Yves Rocher presets in very similar categories for at least several months.

From the "should we do this" perspective, the results of the Discord discussion push me toward "yes". I'm pretty sure we have some European brands that operate both convenience stores and supermarkets, and we have presets in both categories so that editors aren't tempted to shoehorn everything into one category or the other. The Aussie situation with Harvey Norman (as described above) reminds me of that, and I think it's reasonable to provide `shop=electronics` as a preset tagging option for locations where that tag is more appropriate.

andrewharvey (approved):

andrewharvey:
Great. I've update this one to include `AU` as well since that I know of. Other countries included in the department story entry I'm not sure about. Based on this I'm happy to have this merged.

PR #10580, opened 2025-02-20
[shop/furniture] add OSIM

Amar2375:
Adding new entry for Brand in Malaysia for furniture
https://www.osim.com.my/en/
and
https://www.wikidata.org/wiki/Q4207288

### DIFF
--- a/data/brands/shop/furniture.json
+++ b/data/brands/shop/furniture.json
@@ -2435,22 +2435,20 @@
         "name:ja": "ニトリ",
         "shop": "furniture"
       }
+    },
+    {
+        "displayName": "OSIM",
+        "locationSet": {
+            "include": [
+                "my"
+            ]
+        },
+        "tags": {
+            "brand": "OSIM",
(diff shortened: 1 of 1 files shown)


Snowysauce:
I was unable to conclusively determine how many locations this brand has. An overpass-turbo query only returned four locations, one tagged as `shop=furniture` and two tagged as `shop=electronics`. (The fourth had just `shop=yes`.) `shop=electronics` seems to be the better tag based on the brand's description on the English Wikipedia.

Snowysauce:
> I was unable to conclusively determine how many locations this brand has.

It turned out that this was caused by visiting the site linked on Wikidata, which redirected me to their US-specific online store, instead of directly visiting the Malaysian site linked at the start of this PR. The Malaysian website and domain has no such redirect, and a look at their store locator indicates that the brand has more than 30 stores. I've added the Malaysian links (main page and store locator) to the Wikidata item with appropriate qualifiers.

PR #10582, opened 2025-02-20

Amar2375:
Adding new entry for Brand in Malaysia for atm
https://www.alliancebank.com.my/
and
https://www.wikidata.org/wiki/Q16827441

### DIFF
--- a/data/brands/amenity/atm.json
+++ b/data/brands/amenity/atm.json
@@ -517,4 +517,20 @@
       }
     }
   ]
+},
+{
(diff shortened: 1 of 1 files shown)


Snowysauce:
Closing per https://nsi.guide/?t=brands&k=amenity&v=atm#alliancebank-bc168a - its ATM preset was automatically generated from its entry in `brands/amenity/bank`. If there's an issue with the existing preset, please file a specific request detailing where it needs improvement.

PR #10584, opened 2025-02-20
[shop/cosmetics] add Cosway

Amar2375:
Adding new entry for Brand in Malaysia for cosmetics
https://shop.cosway.com.my/
and
https://www.wikidata.org/wiki/Q15916839

### DIFF
--- a/data/brands/shop/cosmetics.json
+++ b/data/brands/shop/cosmetics.json
@@ -1190,23 +1190,21 @@
         "name:zh": "莎莎",
         "shop": "cosmetics"
       }
+    },
+    {
+        "displayName": "Cosway",
+        "locationSet": {
+            "include": [
+                "my"
+            ]
+        },
+        "tags": {
+            "brand": "Cosway",
+            "brand:wikidata": "Q15916839",
+            "name": "Cosway",
+            "shop": "cosmetics"
+        }
     }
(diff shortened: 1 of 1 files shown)


Snowysauce:
The brand is not yet in the NSI and is notable enough to be included, but not as `shop=cosmetics`. Their website shows that they sell far more than cosmetics, while an overpass-turbo query returns 40 instances of "Cosway" with a `shop=*` tag but none with `shop=cosmetics`. (It's not clear what the `shop` value should be, either; the existing POIs have a wide variety of values.)

Snowysauce:
Revisiting this PR, the most applicable tag might be `shop=convenience`. Their product selection is more along the lines of a variety store, general store or department store, but:

- the stores don't appear to be lower price-point oriented, ruling out `shop=variety_store`
- they don't appear to be an all-encompassing one stop shop as defined by the OSM wiki page for `shop=general`
- they don't appear to be anywhere near the size of a standard department store, one of the key criteria for `shop=department_store` as defined by OSM wiki

Five of the 39 relevant POIs are currently tagged as `shop=convenience`. I completely understand why the existing POI tags are so varied - this is a difficult brand to categorize. (To that point, no single key-value combo has more than a half-dozen uses.)

PR #10593, opened 2025-02-20
Add Tom Ford

cicku:


Snowysauce:
I looked at both https://www.tomfordfashion.com/en-us/stores and https://www.tomfordbeauty.com/store-locator to gauge the brand's operating area. The latter appears to cover the brand's boutique arm (fragrances, cosmetics, and whatnot), while the former appears to be the actual subject of this PR (clothing and accessories). Thus, I'll only consider results I got from the former link.

I initially entered a large number of countries from around the world into the Tom Ford Fashion store locator, and it only returned results from countries in western Europe and southeast Asia. However, [an overpass-turbo query](https://overpass-turbo.eu/s/1Zdl) hinted at a larger operating area than that, and indeed, more granular searches (like "California" instead of "United States") returned stores in other areas than Europe and Asia.

I'm not convinced that the brand is truly present worldwide, but without a better store locator, it's impossible to definitively say where they don't have stores. Adding as proposed.

PR #10598, opened 2025-02-21
Split Vienna Hotels

kai-project:
This entry conflated Vienna Hotels (of Jinjiang Group) and Vienna House (of Wyndham Group).

### DIFF
--- a/data/brands/tourism/hotel.json
+++ b/data/brands/tourism/hotel.json
@@ -4073,16 +4073,16 @@
         "brand": "Vienna Hotels",
         "brand:wikidata": "Q7928333",
         "name": "Vienna Hotels",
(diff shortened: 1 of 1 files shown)


Snowysauce:
Thanks for the contribution. This PR looks fine, and if overpass-turbo is accurate, the existing preset was never used because the `locationSet` was incorrect (so that's a silver lining). I have a few suggestions:

- remove `official_name` from Vienna Hotels since it has the same value as the `name` and `brand` tags
- change the `locationSet` for Vienna House to "150" (Europe) per overpass-turbo and https://www.wyndhamhotels.com/vienna-house/locations
- add `"brand:wikidata": "Q2523363",` to Vienna House as this appears to be the brand's Wikidata item

PR #10601, opened 2025-02-22
add `doc_url` property to the generated taginfo file

k-yle:
This allows people to go from taginfo directly to the NSI website to see how exactly the tag is used. I've committed the changes to `dist/taginfo.json` to show the impact of this change. 

Also fixed a broken URL for the project icon.

### DIFF
--- a/dist/taginfo.json
+++ b/dist/taginfo.json
@@ -15,11 +15,31 @@
     {"key": "access", "value": "permit"},
     {"key": "access", "value": "private"},
     {"key": "access", "value": "yes"},
-    {"key": "advertising", "value": "billboard"},
-    {"key": "advertising", "value": "column"},
-    {"key": "advertising", "value": "poster_box"},
-    {"key": "advertising", "value": "screen"},
-    {"key": "advertising", "value": "totem"},
+    {
+      "key": "advertising",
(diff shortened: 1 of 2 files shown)


dist/taginfo.json
```diff
@@ -43,86 +63,362 @@
...
+      "key": "amenity",
+      "value": "prep_school",
+      "doc_url": "https://nsi.guide/?t=brands&k=amenity&v=prep_school"
+    },
+    {
+      "key": "amenity",
+      "value": "prison",
+      "doc_url": "https://nsi.guide/?t=operators&k=amenity&v=prison"
+    },
+    {"key": "amenity", "value": "pub", "doc_url": "https://nsi.guide/?t=operators&k=amenity&v=pub"},
```
peternewman, line 321: What decides which tag of operators and brands gets picked? There are more brands (58) than operators (17) yet the link goes to the latter? Assuming you can't put two links in, or two entries, maybe linking to the bigger one, or defaulting to brand if present or something might make sense?

k-yle: I've removed the `&t=...` part from the URL to solve this issue. Therefore, this PR is blocked by #11120

PR #10603, opened 2025-02-22
[amenity/cafe] expand locationSets of Chagee and ChaTraMue

MASangys:
ADD CHAGEE intl, update chatrume


Snowysauce:
ChaTraMue is operating in 10 countries (the two in the PR plus eight others), while Wikidata only lists two countries for Chagee when excluding China.

PR #10604, opened 2025-02-22
[amenity/fast_food] add A&W Malaysia

MASangys:
a&w available in Malaysia also now

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -110,9 +110,9 @@
       }
     },
     {
-      "displayName": "A&W (USA)",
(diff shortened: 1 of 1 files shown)


Snowysauce:
A&W Malaysia probably needs its own Wikidata item (and thus entry). Unlike A&W Canada, A&W Malaysia uses almost the same logo/branding as the American chain; however, the American chain's logo carries the tagline "All American Food" while the Malaysian logo does not (for obvious reasons). All three chains have different ownership, but as we've seen with Shell, that isn't necessarily a deal-breaker when it comes to preset-sharing.

PR #10611, opened 2025-02-23
Add Rush

peternewman:
Lots of stores:
https://www.rush.co.uk/salon-finder

### DIFF
--- a/data/brands/shop/hairdresser.json
+++ b/data/brands/shop/hairdresser.json
@@ -772,6 +772,10 @@
       "locationSet": {
         "include": ["gb-eng"]
       },
+      "matchNames": [
+        "rush hair & beauty",
+        "rush hairdressers"
+      ],
       "tags": {
         "brand": "Rush Hair",
         "brand:wikidata": "Q132737191",
--- a/data/brands/shop/hairdresser.json
+++ b/data/brands/shop/hairdresser.json
@@ -768,14 +768,14 @@
       }
     },
     {
-      "displayName": "Rush",
+      "displayName": "Rush Hair",
       "locationSet": {
         "include": ["gb-eng"]
       },
       "tags": {
-        "brand": "Rush",
+        "brand": "Rush Hair",
         "brand:wikidata": "Q132737191",
-        "name": "Rush",
+        "name": "Rush Hair",
         "shop": "hairdresser"
       }
     },
--- a/data/brands/shop/hairdresser.json
+++ b/data/brands/shop/hairdresser.json
@@ -768,6 +768,18 @@
       }
     },
     {
+      "displayName": "Rush",
+      "locationSet": {
+        "include": ["gb-eng"]
+      },
+      "tags": {
+        "brand": "Rush",
+        "brand:wikidata": "",
+        "name": "Rush",
+        "shop": "hairdresser"
+      }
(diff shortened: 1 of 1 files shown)


data/brands/shop/hairdresser.json
```diff
@@ -767,6 +767,18 @@
...
     },
+    {
+      "displayName": "Rush",
+      "locationSet": {
+        "include": ["gb-eng"]
+      },
+      "tags": {
+        "brand": "Rush",
+        "brand:wikidata": "",
+        "name": "Rush",
```
peternewman, line 778: Not sure what exactly the brand/name should be, might need to double-check one, they seem to go by various names on their website:
https://www.rush.co.uk/

I guess I'll need match names for the ones we don't pick too:
Rush Hair & Beauty Salon
Rush Hairdresser

data/brands/shop/hairdresser.json
```diff
@@ -767,6 +767,18 @@
...
       }
     },
+    {
+      "displayName": "Rush",
+      "locationSet": {
+        "include": ["gb-eng"]
+      },
+      "tags": {
+        "brand": "Rush",
+        "brand:wikidata": "",
```
peternewman, line 777: Also needs Wikidata too please...

Snowysauce: https://www.wikidata.org/wiki/Q132737191

data/brands/shop/hairdresser.json
Snowysauce: When I created the Wikidata item, I went with "Rush Hair & Beauty" as the title (brand) and "Rush Hairdressers" as an alias (as that seems to be the name of each individual salon). The chain's social media accounts all have usernames with some variation of Rush Hair & Beauty, so I assume they consider that to be their brand as well.

data/brands/shop/hairdresser.json
peternewman: Thanks @Snowysauce . Added.

data/brands/shop/hairdresser.json
peternewman: The two shops I've seen are Rush Hair or more specifically:
```
RUSH
HAIR
```

But I don't think I've seen the newline elsewhere, I'm less clear if we should use the capitals or not?

Snowysauce: I revisited the link at the start of the PR, and aside from the logo, every mention of "Rush" in a mixed-case context was in title case. The all caps brand presentation is probably purely stylistic, such as how Burger King's logo has the name in all caps but their brand is well established to be "Burger King".

peternewman: Cool, thanks for checking. I guess this is good to merge now then?

peternewman:
Let me know if you think anything is outstanding @Snowysauce ? Otherwise do you want to approve and merge?

Snowysauce:
I would add "Rush Hair & Beauty" and "Rush Hairdressers" as `matchNames`, but otherwise this preset looks fine to me.

peternewman:
IIRC we don't need to do both `&` and `and`? I assume `Rush` is felt too generic as a matchName?

Snowysauce (approved):

PR #10614, opened 2025-02-24
Adding official name for Rapid City SD school district and adding Sioux Falls SD school district

the-spui-ninja:


Snowysauce:
Although `operator:official_name` is [an established tag](http://overpass-turbo.eu/s/1ZiA), we don't currently have any presets that use it and I'm not sure if we should begin utilizing it. I can see people using this PR as precedent to include tags for the full legal names of operators (complete with legal form suffixes), and that's data better suited for Wikidata TBH.

the-spui-ninja:
So the full legal names are used more often than you'd think, for instance the school buses will usually have the full school district name including the designation number printed on them. My intent with this was to have the full name show up consistently across the state (I've been adding the full name to the operator tag as I've been mapping), but if people are strongly against it (especially since Rapid doesn't make a big deal out of it, unlike say Spearfish) I'm okay with taking it out.

Snowysauce:
> the full legal names are used more often than you'd think, for instance the school buses will usually have the full school district name including the designation number printed on them

IMO this reasoning could be used to justify changing the main `operator` value itself. However, for now, I'll just remove the potentially problematic `operator:official_name` tags in order to approve the Sioux Falls addition. A new PR can be filed to change the main `operator` tag values if so desired.

PR #10617, opened 2025-02-25
Replace Coles Express with Reddy Express

lkernan:
Replace Coles Express convenience store with Reddy Express.

Reddy Express is the convenience store brand while Shell is the main fuel station brand since Viva bought the servos from Coles. 
Wiki has been renamed previously.


Cj-Malone:
If you don't mind, we'll do this in #10618

PR #10623, opened 2025-02-26
Update locationSet on Simpo

zenfiric:
there are locations in 
- https://simpo.bs
- https://simpo.me
- https://simpo.mk
- https://simpo.rs


Snowysauce:
`"bs"` is the country code for the Bahamas. Was it supposed to be `"ba"` instead?

PR #10626, opened 2025-02-26
add Sydney Water as operator for man_made=pumping_station,water_tower

andrewharvey:


Snowysauce:
How many offices are there? Overpass-turbo only returned one relevant result, and Sydney Water's jobs page (the only official page I found that even refers to multiple locations) had listings for just four distinct offices.

andrewharvey:
I believe they only have the one corporate office at https://www.openstreetmap.org/way/344278413

I added them here since `operators/man_made/pumping_station`, `operators/man_made/water_tower`, `operators/man_made/water_works` all inherit from `operators/office/water_utility`.

I would prefer to keep their infrastructure operate to their corporate office though, but then should I just add them to each of those `man_made` entries?

I'm not even sure adding operator on the corporate office makes sense.

Currently it's tagged on the building, but the building may have it's own operator and simply lease the office space to Sydney Water, I'm not sure.

Snowysauce:
> I added them here since `operators/man_made/pumping_station`, `operators/man_made/water_tower`, `operators/man_made/water_works` all inherit from `operators/office/water_utility` and I wanted to add them as the operator for those infrastructure items.

Ahh, gotcha. I didn't realize that `office/water_utility` fed data to the other three categories.

> I would prefer to keep their infrastructure operator separate to their corporate office, but then should I just add them to each of those man_made entries?

I'm not sure. It'd be easier from a maintenance standpoint to add the data to `office/water_utility` and let it trickle down (no pun intended), but then we'd have a potentially undesirable and unused preset for Sydney Water in the `office/water_utility` category. Meanwhile, `man_made/pumping_station`, `man_made/water_tower`, and `man_made/water_works` all have at least one entry that was added independently of the `office/water_utility` feeder template.

andrewharvey:
I might just add them separately then, it shouldn't add much to maintenance, they don't change very often and then the decision about office/water_utility can be made independently.

andrewharvey:
I decided to just add the operator to the respective infrastructure items under `man_made`. We already had an entry under `man_made=pipeline` and `man_made=water_works`.

If #10628 is accepted, then I'll also add it there.

PR #10628, opened 2025-02-26
add man_made=reservoir_covered operator from office=water_utility

andrewharvey:


andrewharvey:
I think this makes sense because water utilities usually operate water delivery infrastructure which can include water towers, pipelines and covered reservoirs.

However if we're not confident of this globally, I can just add this for Sydney Water initially rather than inherit from `office=water_utility`.

Snowysauce:
This proposal seems fine. I checked overpass-turbo to see if the operators in `office/water_utility` were tagged as operators of any covered reservoirs, and there were results for at least a half-dozen of the operators.

Issue #10553, opened 2025-02-18
Update KFC Entry to Include Vietnamese Name in fast_food.json


ditrungduong:
The JSON file does not include the correct Vietnamese translation for KFC.
Add 'Gà rán KFC' to the name:vi field

1ec5:
Should it be “Gà Rán Kentucky”? Also, I wonder if it should be `alt_name:vi=*` instead, because [the signs](https://garanchickenfood.vn/top-6-nhung-thuong-hieu-ga-ran-nen-nhuong-quyen-tai-viet-nam) seem to only say “KFC”.

ditrungduong:
Thank you for your input! Yes, I agree that “Gà Rán Kentucky” would be a better fit for alt_name:vi.

Regarding alt_name:vi vs. name:vi, I also agree that “KFC” should remain as the primary name (name:vi), since both outdoor and indoor signage prioritize "KFC" in Vietnam. Using "Gà Rán Kentucky" in alt_name:vi follows the same approach as in English-speaking countries, where "Kentucky Fried Chicken" is listed under alt_name because "KFC" is the more commonly recognized brand.

ditrungduong:
Thank you. I will do that.

1ec5:
Fixed in #10636.

PR #10636, opened 2025-02-28
Updated KFC entry with alt_name:vi as 'Gà Rán Kentucky

ditrungduong:
- Updated KFC entry in `fast_food.json`
- Added `alt_name:vi` as "Gà Rán Kentucky" while keeping `name:vi` as "KFC"
- This follows the localization guidelines and maintains consistency with branding
- Related to Issue #10553

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -13607,6 +13607,23 @@
         "name:zh": "齊柏林熱狗",
         "takeaway": "yes"
       }
-    }
+    },
+    {
+      "displayName": "KFC",
+      "id": "kfc-vn",
+      "locationSet": {
+        "include": ["vn"]
+      },
+      "tags": {
+        "amenity": "fast_food",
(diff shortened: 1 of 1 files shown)


Snowysauce:
The build tests failed because "(Vietnam)" (or the appropriate localization thereof) needs to be added to the end of the new entry's display name.

Other suggestions:

- `brand:vi` is unnecessary since its value is identical to `brand`
- Vietnam needs to be excluded from the global KFC entry; otherwise, Vietnamese editors will be given an option to use this proposed preset or the global one

ditrungduong:
Thanks for the feedback! I’ve made the following corrections based on your suggestions:

Updated displayName to "KFC (Vietnam)" to meet localization requirements.
Removed brand:vi since it was redundant.
Updated the global KFC entry to exclude "vn" so that Vietnamese editors will only see the localized option.
Please let me know if any further changes are needed!

PR #10644, opened 2025-03-03
Updated Spanish fuel stations

paunofu:


paunofu:
I don't know why github is committing Cepsa if it was merged two weeks ago :)

Robot8A:
As I can see, the only change left to merge is the change from Plenoil to Plenergy (the website https://plenoil.es/ shows that this is indeed true). I am merging so we can close this issue.

PR #10651, opened 2025-03-04
change operator name of ÜSTRA in GVH to be the currently used one

t-aus-m:
ÜSTRA does not call herself "Überlandwerke und Straßenbahn Hannover" anymore since 1960.
Source: https://de.wikipedia.org/wiki/%C3%9Cstra_Hannoversche_Verkehrsbetriebe

### DIFF
--- a/data/transit/route/tram.json
+++ b/data/transit/route/tram.json
@@ -209,7 +209,7 @@
         "network": "Großraum-Verkehr Hannover",
         "network:short": "GVH",
         "network:wikidata": "Q1549516",
-        "operator": "Überlandwerke und Straßenbahnen Hannover",
+        "operator": "ÜSTRA Hannoversche Verkehrsbetriebe Aktiengesellschaft",
         "operator:short": "ÜSTRA",
         "operator:wikidata": "Q265625",
         "route": "tram"
--- a/data/transit/route/tram.json
+++ b/data/transit/route/tram.json
@@ -209,8 +209,7 @@
         "network": "Großraum-Verkehr Hannover",
         "network:short": "GVH",
         "network:wikidata": "Q1549516",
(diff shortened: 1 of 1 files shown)


Snowysauce:
Your first instinct was correct; taking care of the data after the correction is added to the NSI and released to editors is better because the OSM editors won't suggest the old value as you're cleaning up the data. I suspect the reason for the delay in processing this PR is that the majority of our project maintainers haven't been around as of late, including the ones who normally process European-related PRs.

This change seems straightforward to me, but as a native English speaker with very little knowledge of other languages, I tend to stick to changes involving entities in the English language (lest I get stuck in another overseas quagmire like #10226).

FWIW, while the text of the "fix data" tag indicates a need to cleanup the OSM data post-merge, it's more of a suggestion than a hard requirement. These days, the tag is mostly used as an at-a-glance indicator that the PR/issue is suggesting a change to the existing NSI data.

kjonosm:
We usually don't add the legal form. Could you delete "Aktiengesellschaft" from the new operator value?

t-aus-m:
Hm, I am not a fan of that. The ÜSTRA is usually only referred to as either "ÜSTRA" or "ÜSTRA Hannoversche Verkehrsbetriebe Aktiengesellschaft", "ÜSTRA Hannoversche Verkehrsbetriebe" is not used by anyone including the company itself. Most of the currently existing data is also tagged with the legal form, either as "AG" or as "Aktiengesellschaft". It omitting the legal form a hard requirement for merging this PR?

Snowysauce:
> Is omitting the legal form a hard requirement for merging this PR?

By rule, no, but in practice, it appears to be since this PR has sat idle for the last month.

PR #10306, opened 2025-01-03
Correct `parcel_locker`s and add `hu` versions

gy-mate:
## DHL

- Remove out-of-scope tags (`operator` and `operator:wikidata`) after linking them to the `brand:wikidata` value on Wikidata
- Correct a Wikidata tag (its value was the same as `operator:wikidata`'s)
- Add 'DHL csomagautomata' for `hu`. Sources: https://wiki.openstreetmap.org/wiki/Hungary/Jelölési_példák#Csomagautomata and https://dhlexpress.hu/hu/

## AlzaBox

- Add `parcel_mail_in=yes`. Source: https://dhlexpress.hu/hu/alzabox/
- Add `parcel_pickup=yes`

## DPD Pickup Station

- Remove out-of-scope tags (`operator` and `operator:wikidata`) because the operator is linked to the `brand:wikidata` item on Wikidata
- Add `parcel_mail_in=no`. Source: https://www.dpd.com/hu/hu/visszakuldes/

## Express One

- Add 'Express One csomagpont'. Source: https://expressone.hu/csomagpont

## GLS

- Add `parcel_mail_in=yes` & `parcel_pickup=yes`. Source: https://matrix.to/#/!gAwUAKJoCAuYnIoNeV:grin.hu/$dxwb8-1loFrzP38KeuvodWb0Xdmo4hIHK34AOyi7ahQ?via=grin.hu&via=matrix.org&via=cwt.grin.hu

## IKEA

- Add 'IKEA csomagpont'. Source: https://www.ikea.com/hu/hu/customer-service/services/kattints-es-vedd-at-csomagpontok-puba9241fe7/

## MPL

- Add `parcel_mail_in=yes` & `parcel_pickup=yes`. Source: https://matrix.to/#/!gAwUAKJoCAuYnIoNeV:grin.hu/$dxwb8-1loFrzP38KeuvodWb0Xdmo4hIHK34AOyi7ahQ?via=grin.hu&via=matrix.org&via=cwt.grin.hu

## Packeta

- Add `parcel_mail_in=yes` & `parcel_pickup=yes`. Source: https://matrix.to/#/!gAwUAKJoCAuYnIoNeV:grin.hu/$dxwb8-1loFrzP38KeuvodWb0Xdmo4hIHK34AOyi7ahQ?via=grin.hu&via=matrix.org&via=cwt.grin.hu

PR #10657, opened 2025-03-05
fix wrong suggestion

matkoniecz:
@gy-mate - fixes (maybe part of) #10306

as reported by Polish mapper this is not always true (picture below advertises point-to-point delivery)

screen04

in general I would be really careful with `parcel_mail_in=no` as it changes dynamically

### DIFF
--- a/data/brands/amenity/parcel_locker.json
+++ b/data/brands/amenity/parcel_locker.json
@@ -398,7 +398,6 @@
         "amenity": "parcel_locker",
         "brand": "DPD Pickup Station",
         "brand:wikidata": "Q114273730",
-        "parcel_mail_in": "no",
         "parcel_pickup": "yes"
       }
     },


gy-mate (changes requested):
Could you please extract the `hu` `locationSet` of this item to a separate one in order to keep the `"parcel_mail_in": "no"` value there as described at https://wiki.openstreetmap.org/wiki/Hungary/Jelölési_példák#Csomagautomata?

> in general I would be really careful with parcel_mail_in=no as it changes dynamically

It doesn't change often in Hungary. I've found no traces of local DPD parcel lockers accepting `mail_in` since they've been introduced.

bhousel:
> Could you please extract the `hu` `locationSet` of this item to a separate one in order to keep the `"parcel_mail_in": "no"` value there as described at [https://wiki.openstreetmap.org/wiki/Hungary/Jelölési_példák#Csomagautomata](https://wiki.openstreetmap.org/wiki/Hungary/Jel%C3%B6l%C3%A9si_p%C3%A9ld%C3%A1k#Csomagautomata)?

No. We should not be setting this `parcel_mail_in` tag in NSI.  https://github.com/osmlab/name-suggestion-index/pull/10306#issuecomment-2699169091
If you want all the parcel lockers in Hungry to have the tag, you can use Overpass and JOSM to do it.

gy-mate (approved):

Snowysauce:
I added a couple of new discussion links to the Judge Case tag section, but not a specific row for `parcel_mail_in`. To me, the tag fits into the "tags that require an on ground survey" umbrella, a row which was deliberately left open-ended because there are probably dozens of tags that we could list there. This approach to these tags may not provide at-a-glance, clear-cut guidance on everything, but I think it's easier from a maintenance standpoint to just link to discussions that talk about these sorts of tags. (These days, anything that cuts down on maintenance is a good thing - most of the project maintainers are currently inactive.)

gy-mate:
> I added a couple of new discussion links to the Judge Case tag section

@Snowysauce Thanks a lot! :)

> but not a specific row for `parcel_mail_in`. To me, the tag fits into the "tags that require an on ground survey" umbrella

If I were to add _DPD csomagautomata_ to the NSI with a `hu` `locationSet`, why would it need ground survey? No DPD parcel lockers in Hungary accept `parcel_mail_in` as they are all the same.

matkoniecz:
> If I were to add DPD csomagautomata to the NSI with a hu locationSet, why would it need ground survey? No DPD parcel lockers in Hungary accept parcel_mail_in as they are all the same.

because maybe there not actually the same and this claim is as wrong for Hungary as it was for Poland?

this is also detail that is very very likely to change and not being noticed, like it was not noticed that NSI was making false claims for Poland

PR #10661, opened 2025-03-06
Add more Swedish shop brands

sp1187:
Changes:
* Add shop brands 3Butiken, Dogman, Löplabbet, Panduro (formerly Panduro Hobby), Team Sportia, Ur&Penn.
* Add Sweden and Norway to jem&fix location set.
* Add Norway to NetOnNet location set.
* Update name of Hööks (formerly Hööks Hästsport) and add Denmark, Finland and Norway to location set.

### DIFF
--- a/data/brands/shop/craft.json
+++ b/data/brands/shop/craft.json
@@ -99,6 +99,21 @@
       }
     },
     {
+      "displayName": "Panduro",
+      "locationSet": {"include": ["dk", "no", "se"]},
+      "matchNames": ["panduro hobby"],
+      "matchTags": [
+        "shop/doityourself",
+        "shop/hobby"
+      ],
+      "tags": {
+        "brand": "Panduro",
+        "brand:wikidata": "Q265941",
+        "name": "Panduro",
+        "shop": "craft"
+      }
+    },
+    {
       "displayName": "pipoos",
       "id": "pipoos-d918ce",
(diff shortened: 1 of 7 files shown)


Snowysauce:
Science Fiction Bokhandeln and Inet appear to have five or less locations. We generally don't add brands to the index unless they have 10-30 locations per [our notability criteria](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#notability). Since those were the first two brands in the PR, I didn't look at the rest.

sp1187:
I removed those two now, the rest of them should be larger. Löplabbet is perhaps an edge case though they seem to have around 25 shops in Sweden and Norway combined. 

Also updated Hööks a bit with an expanded location set and more tags.

Snowysauce:
> I removed those two now, the rest of them should be larger. Löplabbet is perhaps an edge case though they seem to have around 25 shops in Sweden and Norway combined.

Thanks. All of the remaining brands look notable enough to be in the NSI, Löplabbet included.

> Do you think I should split this up into several PRs?

I personally don't mind large PRs - I sent in some wide-ranging ones to the NSI back before I became a maintainer - but I will admit that they take longer to properly vet and process. Smaller PRs normally get approved faster; however, we currently have a bit of a maintainer shortage, so I'm not sure that the size of the PRs would make a difference right now.

PR #10664, opened 2025-03-08
Move Carrefour City/Contact/Market from supermarket.json to convenience.json 

### LINKED ISSUE #9869: Move some French shops from supermarket to convenience
### What

Following a discussion in #9857, there are a couple of French `shop/supermarket` that should be moved to `shop/convenience`.

- [x] Intermarché Express : https://github.com/osmlab/name-suggestion-index/commit/e7e1d9fbe02c67c1f1561688c37db36a86427e00
- [x] U Express : https://github.com/osmlab/name-suggestion-index/commit/3d831ed5a429c030f02d982797ad66fd07491c8d
- [ ] Carrefour City/Market/Contact

### Extra info

- https://wiki.openstreetmap.org/wiki/Tag:shop%3Dsupermarket
- https://wiki.openstreetmap.org/wiki/Tag:shop%3Dconvenience

> shop:convenience : Typically much smaller than a supermarket, and typically carries a more limited selection of food items. One suggestion for distinguishing between a supermarket and a convenience shop, is that a supermarket is large enough to use trolleys.

### Impact ?

> There might be an impact on the other brands if they are switched to another category, but if they are more suitable to that category then iD and other mapping software will simply suggest the update to mappers, so the impact won't be a big deal.

raphodn:
Closes #9869


Snowysauce:
I have two concerns, both regarding "Carrefour Market":

- separate presets exist for the brand for Taiwan and Georgia, but as proposed they would be covered up by the global preset (i.e., the global preset should have `"exclude": ["ge", "tw"]` as part of its `locationSet`)
- the preset being global is potentially problematic in itself; for example, according to https://www.britannica.com/money/Carrefour-SA, Carrefour left the US market in the 1990s and never returned

PR #10674, opened 2025-03-11
Update operator and network tags of tram routes in Lodz and add similar one item to buses

codemaster-arch:
**1. trams**

Changed _network_ name from "LTZ Łódź" to "Lokalny Transport Zbiorowy w Łodzi", aligning with official terminology. Wherever this name appears, it is given in full words. When you google this abbreviation, results refer to "ŁTŻ Łódź" (motorcycle speedway club) rather than mass transit service.

Updated _network:wikidata_ tag to link to the correct Wikidata item representing the transit network.

Fixed _operator_ name from "MPK Łódź" to "MPK-Łódź", matching the official spelling. The company consistently uses a hyphen instead of blank space in its name.

**2. buses**

Created a separate entry for "Lokalny Transport Zbiorowy w Łodzi" with _route=bus_ as an equivalent to existing _route=tram_ entry.

### DIFF
--- a/data/transit/route/tram.json
+++ b/data/transit/route/tram.json
@@ -345,13 +345,13 @@
       }
     },
     {
-      "displayName": "LTZ Łódź",
+      "displayName": "Lokalny Transport Zbiorowy w Łodzi",
       "id": "ltzlodz-ba1938",
       "locationSet": {"include": ["pl"]},
       "tags": {
-        "network": "LTZ Łódź",
-        "network:wikidata": "Q1149987",
-        "operator": "MPK Łódź",
+        "network": "Lokalny Transport Zbiorowy w Łodzi",
+        "network:wikidata": "Q133202246",
+        "operator": "MPK-Łódź",
         "operator:wikidata": "Q11780295",
         "route": "tram"
       }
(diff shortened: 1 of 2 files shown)


data/transit/route/tram.json
```diff
@@ -345,13 +345,13 @@
       }
     },
     {
-      "displayName": "LTZ Łódź",
+      "displayName": "Lokalny Transport Zbiorowy w Łodzi",
       "id": "ltzlodz-ba1938",
       "locationSet": {"include": ["pl"]},
       "tags": {
-        "network": "LTZ Łódź",
```
Snowysauce, line 352: "LTZ Łódź" needs to be added as a match name for the preset. Based on the workflow tests, our build script will create a new preset titled "LTZ Łódź" otherwise.

codemaster-arch: I've added matchname as requested. I understand that I didn't have to add this line in **bus.json** file? I hope everything is alright now.

Snowysauce: After looking at `bus.json`, there is an unused preset for "LTZ Łódź" that should be merged with the new, proposed preset. The easiest way to do so would be to delete that preset and add the same `matchNames` line that was added to `tram.json` to the new preset.

codemaster-arch: I've merged the old bus preset into the new one by removing the duplicate entry and adding matchNames as suggested. I apologize for the numerous revisions, I'm new to this.

Snowysauce: No worries; there's definitely a learning curve to reading and working with the category files, and it gets especially tricky when preset changes involve the NSI Collector (OSM planet) data. Everything looks good now, and the workflow tests also came back clean.

PR #10675, opened 2025-03-12
Add 9 more locations to Mixue ice cream café

LuisAPI:
1. Add seven more ASEAN countries (`ID`, `KH`, `LA`, `PH`, `SG`, `TH`, `VN`), South Korea and Australia to list of countries where the Mixue brand now operates.
2. Replace English-language name of brand from unused capitalized Romanized form of Chinese name (`MIXUEBINGCHENG`) to actual brand name used [on the ground](https://wiki.openstreetmap.org/wiki/Ground_truth) (i.e. `Mixue Ice Cream & Tea`).


Snowysauce:
The links currently available on the Wikidata item provide the following information:

- countries where a full location list is provided for the brand: Thailand, Indonesia, Singapore, Vietnam, Malaysia
- countries where the brand is either mentioned or implied to operate in, but no locations are explicitly listed: China (covered by a different preset), Japan, Cambodia, Australia, Philippines

The brand's article on the English Wikipedia mentions operations in South Korea and Laos, completing the suggested `locationSet` additions. As such, the `locationSet` change looks OK. (The Wikidata item could use quite a bit of updating in terms of operating area and related verification, but that's irrelevant to the PR.)

As for the `brand` value, please note that the China-specific preset also uses "MIXUEBINGCHENG" as the value for `brand:en` - this tag's value should probably be changed to "Mixue Ice Cream & Tea" for consistency.

Finally, the proposed location set covers countries with a variety of language scripts. Is the brand presented in English in all of them?

LuisAPI:
Admittedly I haven't checked much regarding this brand's locations in all those countries that Wikipedia claims it operates in.

In the case of the Philippines specifically though, I can confirm it does operate under the English brand name, and so that was the reason why I made this PR in the first place.

About the Romanized name, actually I was thinking a properly spelled form of it should instead be under `name:zh-Latn-pinyin=`, per
the [Multilingual names](https://wiki.openstreetmap.org/wiki/Multilingual_names) page on the OSM Wiki.

I think it's much better to add it now the way it is, and simply wait for a volunteer interested in this brand's operations in any of the other countries to submit an issue/PR correcting their specific location's data... rather than just not having Mixue listed *at all* on iD in those countries.

Snowysauce:
> I think it's much better to add it now the way it is, and simply wait for a volunteer interested in this brand's operations in any of the other countries to submit an issue/PR correcting their specific location's data... rather than just not having Mixue listed _at all_ on iD in those countries.

This was kind of discussed at #10757, and I think each project member and maintainer have their own opinions on the topic. I tend to agree with those who are hesitant to adopt PRs where follow-up work may be needed, and as such I'll leave the fate of this PR to be decided by others.

Snowysauce:
I attempted to look into the multilingual question for the English preset, and at a cursory glance it seems as if the brand is co-signed in English and Chinese in Japan but not in Malaysia, where it's apparently only signed in English. To me, that calls for separate presets: one for Japan only (the existing preset) and one for other countries where English is the primary spoken language that doesn't have any of the `*:zh` multilingual tags. The Chinese multilingual tags aren't wrong for the latter countries, but they'd be a questionable suggestion on the part of the NSI.

Snowysauce:
> ... one [preset] for other countries where English is the primary spoken language that doesn't have any of the `*:zh` multilingual tags. The Chinese multilingual tags aren't wrong for the latter countries, but they'd be a questionable suggestion on the part of the NSI.

I added this in https://github.com/osmlab/name-suggestion-index/commit/b6eab8dc439157894e9b450fc6dfe90dbeda1b42 after being pinged yet again about this brand. With this change, here's the status of this brand against the proposed `locationSet`:

- Australia, Indonesia, Malaysia, Philippines, Singapore, Vietnam: covered by an English-only preset
- Japan: covered by a preset with English primary tags and Chinese secondary tags
- Thailand and Cambodia: brand presence confirmed, but it's not known what language the store names use (the countries mostly use language scripts other than Latin)
- South Korea and Laos: brand presence unconfirmed, and it's not known what language the store names use (the countries use language scripts other than Latin)

Snowysauce:
Cambodia, Laos, Thailand, and South Korea have been added to the Latin-only preset per:

- Thailand: https://th.mixue.com/
- Cambodia: https://www.facebook.com/MIXUECAMBODIA/
- South Korea: https://www.instagram.com/mixuekr/
- Laos: https://www.facebook.com/p/MIXUE-Ice-CreamTea-LAOS-100082987039915/

This change should complete this PR.

LaoshuBaby:
> * Japan: covered by a preset with English/Latin primary tags and Chinese secondary tags

There is only 1 store in Japan, use global one (if have) is enough I think.

PR #10676, opened 2025-03-12
[shop/clothes] add JAM Clothing

RicsPOITT:
Adding new entry for a Brand in South Africa for Clothing Store

https://jamclothing.co.za/
and
https://www.wikidata.org/wiki/Q129814709

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -12560,16 +12560,17 @@
         "shop": "clothes"
       }
     },{
-    "displayName": "JAM Clothing",
+    "displayName": "JAM",
     "locationSet": {
         "include": [
             "za"
         ]
     },
+      "matchNames": ["JAM Clothing"],
     "tags": {
-        "brand": "JAM Clothing",
+        "brand": "JAM",
         "brand:wikidata": "Q129814709",
-        "name": "JAM Clothing",
+        "name": "JAM",
         "shop": "Clothes"
     }
 }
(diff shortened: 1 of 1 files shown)


RicsPOITT:
If you look at the store locator you will clearly see that there is alot of stores in south africa and not just a few. 
image

The Naming on the storefornt and websit is capital letters JAM Clothing

Snowysauce:
> If you look at the store locator you will clearly see that there is alot of stores in south africa and not just a few. 

Indeed; store count is the very first thing I check for any issue or PR. If the brand didn't have enough stores to qualify for the NSI, my comments would have focused on that. Since the chain passes our notability criteria, the next step is determining what the best tag values are for the brand, hence the check of how the chain is currently mapped.

> The Naming on the storefornt and websit is capital letters JAM Clothing

In terms of OSM, determining whether the name should be in all caps or not requires a deeper look than the logo or branding; otherwise, we'd have presets for "BURGER KING", not "Burger King". A deep dive into the website indicates that "JAM" started out as an acronym/separate letters, so using "JAM" seems appropriate here. What's less clear is whether `name` (and possibly `brand`) should be "JAM" or "JAM Clothing". The latter is obviously the company's full formal trading name, but the photos of storefronts on the website and all mentions of the brand in non-formal contexts omit the word "Clothing". This is probably why eight of the 10 existing POIs in OSM use "JAM" or "Jam".

RicsPOITT:
Agreed, on JAM and not JAM Clothing as best tag value.

PR #10678, opened 2025-03-12
[shop/furniture] add The Bed Shop

RicsPOITT:
Adding new brand for Furniture store in south Africa called The Bed Shop

https://www.wikidata.org/wiki/Q129571574
https://thebedshop.co.za/


RicsPOITT:
It would actually make more sense to have it as `shop=bed` reason i added it as `shop=furniture` was because some if these brands locations also specializes in furniture. Now thinking of it definately under `shop=bed`

do you need me to close the pr and reopen new one under correct json?

Snowysauce:
> do you need me to close the pr and reopen new one under correct json?

Not at all; if you move the preset to the `shop=bed` json file in your local copy of the NSI, GitHub will detect that change and automatically update this PR with the new commits.

PR #10679, opened 2025-03-12
[route/bus] add Hop Bus

gosgtan:
Added a new bus network (Hop Bus)


Snowysauce:
According to the network's Wikidata page, the network won't be in service for another four months. No website is listed for the network, and their Facebook page has no activity whatsoever. I don't know what the NSI has done in the past for entities that don't yet exist, but this seems like a PR better suited for when the network launches in July.

PR #10683, opened 2025-03-13
Include Rio de Janeiro Prefecture as school and kindergarten operator

nighto:


data/operators/amenity/clinic.json
```diff
@@ -1789,23 +1789,6 @@
...
-      "locationSet": {
-        "include": ["br-rj.geojson"]
-      },
-      "matchNames": [
-        "prefeitura do rio de janeiro"
-      ],
-      "tags": {
-        "amenity": "clinic",
-        "healthcare": "clinic",
-        "operator": "Prefeitura da Cidade do Rio de Janeiro",
```
Snowysauce, line 1804: This name needs to be retained as a match name for the renamed entry according to the workflow tests. Otherwise, the build script will recreate an entry with this name on its first run after this PR is merged.

PR #10684, opened 2025-03-13
Add Craft Union GB [Over 600 pubs]

TheUKHighStreet:


data/brands/amenity/pub.json
```diff
@@ -57,6 +57,16 @@
         "brand:wikidata": "Q123913080"
       }
     },
+    {
+      "displayName": "Craft Union",
+      "locationSet": {"include": ["gb"]},
```
codeinabox, line 62: They only have pubs in England and Wales

```suggestion
      "locationSet": {"include": ["gb-eng", "gb-wls"]},
```

Cj-Malone: For now

PR #10685, opened 2025-03-13
Add forthcoming locations for hotel brand lyf

codeinabox:
The hotel brand lyf is opening locations in Bali, Paris, and Vietnam

https://www.discoverasr.com/en/find-residence/property?view=property&brand=lyf


Snowysauce:
I have the same concerns about this PR as I do about #10679. The French `locationSet` addition might be fine as that location is expected to open in May (two months from now), but the locations in the other countries aren't scheduled to open until sometime next year or later. A lot can happen between then and now, including the possibility of those planned locations being cancelled and never opening.

codeinabox:
Good point, I don't mind restriction this to the addition of Paris in the location set, as the rest don't open till at least 2026

Issue #5371, opened 2021-09-20
Hypermarket Target locations


nchristensen:
Certain Target locations, formerly known as SuperTargets, have full grocery stores making them hypermarkets like Walmart Supercenters. The OSM wiki indicates hypermarkets are typically tagged as `shop=supermarket`. Should we add a supermarket option for these Target locations alongside the `shop=department_store` option already present?

bhousel:
Seems good to me, thanks for the suggestion 👍

PR #10691, opened 2025-03-14
Drop Target shop=department_store

Cj-Malone:
I'm not super confidant on this, I've never been to the US, never been in a Target.

I am confidant the vast majority in OSM are miss tagged:

> Target (only stores that do not carry a full selection of groceries; stores that sell groceries should be [shop](https://wiki.openstreetmap.org/wiki/Key:shop)=[supermarket](https://wiki.openstreetmap.org/wiki/Tag:shop%3Dsupermarket))

https://wiki.openstreetmap.org/wiki/Tag:shop%3Ddepartment_store

Briefly discussed on [OSM US slack](https://osmus.slack.com/archives/CDJ4LKA2Y/p1736382081486899), they basically all offer food nowadays. And a lot of the `shop=department_store` in OSM may have been encouraged because that was in NSI first.

> basically all

I don't know. And the website doesn’t differentiate.

### DIFF
--- a/data/brands/shop/department_store.json
+++ b/data/brands/shop/department_store.json
@@ -1452,17 +1452,6 @@
       }
     },
     {
-      "displayName": "Target (USA)",
-      "id": "target-592fe0",
-      "locationSet": {"include": ["us"]},
-      "tags": {
-        "brand": "Target",
-        "brand:wikidata": "Q1046951",
-        "name": "Target",
-        "shop": "department_store"
-      }
-    },
-    {
       "displayName": "The Warehouse",
       "id": "thewarehouse-98a4a8",
       "locationSet": {"include": ["nz"]},


Snowysauce:
All Target stores I've been to do have a food selection, but while they have made their grocery departments larger in the last decade-plus, I wouldn't consider the departments to be full selections on par with a Walmart Supercenter or a standalone supermarket. Walmart Supercenters are about one-third supermarket and two-thirds other departments, whereas the store composition at my local Targets (as such, YMMV) is more like one-eighth supermarket to seven-eighths everything else.

All of that said, I think OSM defines stores like these differently than how they're commonly perceived, as I can't imagine that most Americans would consider either Walmart (supercenter or not) or Target as supermarkets. In common parlance, that term is reserved exclusively for chains like Wegmans, Publix, etc.

bhousel:
#5371 was where we last discussed it, I think.
OSM doesn't have a `hypermarket` tag, so for stores in that format we've used `supermarket`.
see https://en.wikipedia.org/wiki/Hypermarket

Snowysauce:
Another theory I have as to why `shop=department_store` is a common tag in the US is that in common parlance, the stores that replaced traditional department stores like Sears (Walmart, Target, etc.) seem to have inherited the "department store" label/perception, and that name stuck even if it's not 100% accurate these days. I've only become familiar with the term "hypermarket" through my time here at the NSI, and I agree that it's the most applicable label for chains like Walmart and Target.

I think it almost goes without saying that, like Walmart, `shop=department_store` has to become a match tag for a `shop=supermarket` Target preset if this PR is accepted, both for resolving conflicts with the OSM planet data and for bridging the gap between common parlance and OSM tagging conventions.

Cj-Malone:
I don't think it's good to compare this to Walmart. Walmart is really clear with it's different store formats, and OSM is perfect. As far as I can tell, Target used to have different store formats, but stopped saying them publicly (if they even exist internally any more) and OSM is just random.

> #5371 was where we last discussed it, I think. OSM doesn't have a `hypermarket` tag, so for stores in that format we've used `supermarket`. see https://en.wikipedia.org/wiki/Hypermarket

I think that's fine, "Americans do everything bigger". I'm curious if Target has locations that don't match the supermarket/hypermarket definition.

Snowysauce:
Target is based in the Twin Cities, so I checked out the page for a random store there and compared it to those of my local stores. [Here's a store in St. Paul, Minnesota](https://www.target.com/sl/st-paul-midway/2229) that lists "Expanded Grocery", "Bakery", "Deli", and "Wine, Beer & Spirits" under store features. By comparison, all of my local Targets (Rochester, New York) just have "Fresh Grocery" and "Beer Available". I'm assuming that means that the store in St. Paul has a considerably larger food selection than any Target I've personally seen.

bhousel:
I agree with @Snowysauce  - From what I can tell on their store locator "Fresh Grocery" is the checkbox that says whether they have grocery items and would be a `shop=supermarket` vs just a basic `shop=department_store`.

All of the ones in my area around New Jersey have "Fresh Grocery", but I wouldn't be surprised if there are others that still don't offer groceries.

Screenshot 2025-03-14 at 2 53 06 PM

As for what tag to use, I really don't have a preference.  
I think OSM has too many tags and people assign too much importance to the actual `shop=` value.

1ec5:
> Another theory I have as to why `shop=department_store` is a common tag in the US is that in common parlance, the stores that replaced traditional department stores like Sears (Walmart, Target, etc.) seem to have inherited the "department store" label/perception, and that name stuck even if it's not 100% accurate these days.

I think this is a good insight. `shop=department_store` was originally intended for traditional full-service department stores, like a Montgomery Ward, or at least a Macy’s or Sears inside a shopping mall. The clearest test is that each department has its own checkout counter. But the U.S. hardly has any full-service department stores anymore. Instead, when people hear department store in a present-day context, they might well think of a _discount_ department store such as Kmart, Kohl’s, Marshalls, or Ross, all of which we’re tagging as `shop=department_store`. Walmart was clearly a chain of discount department stores before they added groceries.

In the U.S. retail industry, a supermarket is a combination of grocery store + pharmacy.[^wall] A hypermarket, also known as a discount superstore, is a combination of discount department store + supermarket. Clear examples of hypermarkets include present-day Walmart and Meijer, which we’re tagging as `shop=supermarket`. In my own experience with hypermarkets, these are the kinds of places you go primarily for your grocery shopping, mostly forgetting about the rest of the departments until you need something over there. 

Target has two formats, both of them quite large:

* The smaller format has half a dozen aisles of packaged foods, a freezer aisle in the back, and two half-aisles of “fresh groceries”, all in zip-lock bags. This is discount department store + convenience store. I’ve been to gas stations that had more fresh produce.
* The larger format, originally and still sometimes branded as SuperTarget, has a full-service supermarket including a bakery and deli counter. This is a hypermarket, equivalent to a Meijer or Walmart Supercenter.

I think it makes sense to keep distinguishing the two formats because of the significant difference in selection, but unfortunately it won’t be easy to make this choice intuitive for users. Users who’ve been to a hypermarket and a smaller Target will know the difference; others, not so much.

[^wall]: Historically, a chain pharmacy would always be next door to a chain grocery store, but they started tear down the divider wall once the laws allowed for it.

Cj-Malone:
> I agree with @Snowysauce - From what I can tell on their store locator "Fresh Grocery" is the checkbox that says whether they have grocery items and would be a `shop=supermarket` vs just a basic `shop=department_store`.
> 
> All of the ones in my area around New Jersey have "Fresh Grocery", but I wouldn't be surprised if there are others that still don't offer groceries.

It appears they all have this flag. Josh noticed in tho OSM US thread too.

> OSM is perfect.

Was perfect. #10694

bhousel:
stale

PR #10702, opened 2025-03-16
amenity/bicycle_rental: Nibelungen Bike rebranded to VELOLEO

rohieb:
See the [announcement](https://www.braunschweig.de/politik_verwaltung/nachrichten/veloleo.php) (in German)

Not sure if the ID should change too, but I guess it should stay for backwards compatibility…?

I've already updated the Wikidata item accordingly.

### DIFF
--- a/data/brands/amenity/bicycle_rental.json
+++ b/data/brands/amenity/bicycle_rental.json
@@ -2763,16 +2763,17 @@
       }
     },
     {
-      "displayName": "Nibelungen-Bike",
+      "displayName": "VELOLEO",
       "id": "nibelungenbike-497979",
       "locationSet": {
         "include": [[10.52, 52.26, 10]]
       },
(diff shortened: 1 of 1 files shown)


UKChris-osm:
The ID will automatically change if merged.

I'd add a [matchNames](https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#matchnames-matchtags) for the former name as well.

Do we need to have a `brand` and `network` if both have the same values?

rohieb:
> I'd add a matchNames for the former name as well.

Right, I've added it.

> Do we need to have a `brand` and `network` if both have the same values?

I'm not sure. A lot of other templates in this file are doing it the same way too. I cannot find good documentation on the OSM Wiki what "network" is actually supposed to mean. I think it's a different issue better solved in another pull request; here I just used search-and-replace for the rebrand.

PR #10717, opened 2025-03-18
Add LKQ Pick Your Part to operator/landuse/industrial

EvanCarroll:
Only added LKQ Pick Your Part. This is a salvage yard in the USA with 63 yards.

The branch name is wrong. Doesn't matter. Caught it. This should be in operators not brands.

For one such location, https://www.openstreetmap.org/way/1368619252

### DIFF
--- a/data/brands/landuse/industrial.json
+++ b/data/brands/landuse/industrial.json
@@ -1,0 +1,23 @@
+{
+  "properties": {
+    "path": "brands/landuse/industrial",
+    "preserveTags": ["^name"],
+    "exclude": {"generic": ["^industrial$"]}
+  },
+  "items": [
+    {
+      "displayName": "LKQ Pick Your Part",
+      "locationSet": {"include": ["us"]},
+      "preserveTags": ["^name"],
+      "tags": {
+        "brand": "Pick Your Part",
+        "brand:short_name": "PYP",
+        "brand:wikidata": "Q133309706",
+        "industrial": "auto_wrecker",
+        "landuse": "industrial",
+        "name": "Pick Your Part",
+        "official_name": "LKQ Pick Your Part",
+      }
+    }
+  ]
+}


Snowysauce:
From the workflow tests:

```
⚠️   Warning - Unusual OpenStreetMap tag:
----------------------------------------------------------------------
  To resolve these, make sure the OpenStreetMap tag is correct.
----------------------------------------------------------------------
  "LKQ Pick Your Part (lkq-e0f808)" -> unusual tag? -> "wikidata = Q133309706"
```

Plain Wikidata tags (`wikidata=*`) are flagged as errors by our build script because these are intended for standalone locations, which are outside the scope of the NSI. The contents of the Wikidata QID in question are more in line with a brand and as such the appropriate tag would be `brand:wikidata`.

However, since this preset is being proposed for the operators tree, most of the proposed values only work if these salvage yards are the only entity that LKQ operates.

EvanCarroll:
LKQ operates other non-salvage yard entities too. Like purely parts stores. They have multiple businesses. So how would you model it if the operator of a salvage yard is a corporation with multiple "brands" but only one of them is vehicle salvage?

I an see your point though. Can we have have operator, operator: wikidata set to LKQ Corporation. And have brand: brand:wikidata set to Pick Your Part. Then we can just leave wikidata empty?

If this is acceptable, I'll fix. To put it another way, why can't an operator/landuse/industrial have a brand too?

EvanCarroll:
@Snowysauce updated, opinions?

Cj-Malone:
`landuse` for a shop feels really weird.

bhousel:
> `landuse` for a shop feels really weird.

Yeah this is how scrap yards work in OSM, for some reason.
Actually this item isn't even using the tag I thought it would. 🤷 
https://wiki.openstreetmap.org/wiki/Tag:industrial%3Dscrap_yard

EvanCarroll:
> `landuse` for a shop feels really weird.

Doesn't feel weird to me. It's not a shop in the sense that I see shop used on OSM. And it does follow other OSM conventions. For example, a dealership has a "shop" the building. But you can tag the containing parking lot with `landuse=retail` to show which lots belong to the dealer.

data/operators/landuse/industrial.json
```diff
@@ -26,6 +26,22 @@
...
+      "tags": {
+        "brand": "Pick Your Part",
+        "brand:short_name": "PYP",
+        "brand:wikidata": "Q133309706",
+        "industrial": "auto_wrecker",
+        "landuse": "industrial",
+        "name": "Pick Your Part",
+        "official_name": "LKQ Pick Your Part",
+        "operator": "LKQ",
+        "operator:wikidata": "Q20711490"
```
1ec5, line 42: [Pick Your Part <small>(Q133309706)</small>](https://www.wikidata.org/wiki/Q133309706) says the chain is a subsidiary of [LKQ Corporation <small>(Q20711490)</small>](https://www.wikidata.org/wiki/Q20711490). We don’t normally use `operator=*` to tag corporate hierarchies. It’s primarily for situations where the operator is orthogonal to the brand (think franchisee), which is why `operator=*` rarely occurs in the brand/ tree and vice versa. Since there is a brand involved, consider moving this entry to a new data/brands/landuse/industrial.json.

1ec5: To be clear, `operator=*` also happens to be correct tagging on these scrapyards. However, it isn’t the job of NSI to standardize both `brand=*` and `operator=*` at the same time, or even if so, `brand=*` takes precedence. Otherwise, we’d have a lot more owned-and-operated chains under the brand/ tree to add `operator=*` to.

EvanCarroll: @1ec5 to speak to how intrusive LKQ is in the brand....

Screenshot_2025-03-20_15-30-51

That's the kind of stuff on the building that i'm mapping

1ec5: I would love it for NSI to be more granular so that we can standardize `brand=*` and `operator=*` individually with their QIDs on a given feature. #10181 is a prime example why. But that would require significant work in this repository and upstream and downstream.

EvanCarroll: Ah, so this isn't a taxonomy issue as much as it is an NSI shortcoming now, that i'll have to read more about. Cool cool. I'll get this moved over to brand today then. We're good.

EvanCarroll:
@1ec5 how's that look?

1ec5 (approved):
Looks great, thanks!

PR #10719, opened 2025-03-18
[shop/electronics] add brand RED ZAC for AT

BruniSchulz:
add brand RED ZAC for AT
reference website:
https://www.redzac.at/info/Home
more than 200 shops in AT

### DIFF
--- a/data/brands/shop/electronics.json
+++ b/data/brands/shop/electronics.json
@@ -2496,14 +2496,14 @@
       }
     },
     {
-      "displayName": "RED ZAC",
+      "displayName": "Red Zac",
       "locationSet": {
         "include": ["at"]
       },
       "tags": {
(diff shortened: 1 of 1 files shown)


Snowysauce:
It looks like 15 of the brand's locations are currently mapped in OSM, and 14 of them use the mixed-case "Red Zac" or a variant thereof for the name.

BruniSchulz:
So should I update the spelling to 'Rec Zac'?

Snowysauce:
Perhaps. If the preferred name value came about organically (e.g., multiple different editors added them at various times), then I'd say yes. However, if one editor changed them en masse at a single time (like what happened in #10773), then what's in OSM doesn't matter much and it becomes a judgment call on what value is really best.

PR #10723, opened 2025-03-18
[amenity/bank] add Abu Dhabi Islamic Bank in Egypt

ZiadTT:
https://www.wikidata.org/wiki/Q130726987
And
https://www.adib.eg/

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -19479,6 +19479,23 @@
         "name:en": "Kagoshima Bank",
         "name:ja": "鹿児島銀行"
       }
+    },{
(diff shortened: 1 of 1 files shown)


Snowysauce:
This brand is already in the NSI as part of [this preset](https://nsi.guide/index.html?t=brands&k=amenity&v=bank&tt=abu%20dhabi%20islamic%20bank#abudhabiislamicbank-92591a) - we don't break up brands by their subsidiaries/country-specific Wikidata QIDs unless there's a need for different tagging. This PR proposes the same tagging as the existing preset except for the Egypt-specific QID.

PR #10724, opened 2025-03-18
Update bank.json add new bank brand in Egypt

ZiadTT:
https://www.qnb.com.eg/sites/qnb/qnbegypt/page/en/en-home.html
And
https://www.wikidata.org/wiki/Q12184634


Snowysauce:
This appears to be the Egyptian subsidiary of an international bank based in Qatar, with no difference in branding between the two. As such, the preset for this bank should be based around the parent bank, including using its QID for `brand:wikidata`.

PR #10725, opened 2025-03-18
Add NLB branches distinct by country

zenfiric:
NLB purchased a couple of local banks, but they have a slightly distinct branding

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -8453,19 +8453,6 @@
       }
     },
     {
-      "displayName": "Komercijalna banka",
-      "id": "komercijalnabanka-0c399e",
-      "locationSet": {
-        "include": ["ba", "me", "rs"]
-      },
-      "tags": {
-        "amenity": "bank",
-        "brand": "Komercijalna banka",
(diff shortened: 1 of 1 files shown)


Snowysauce:
Quick question: Komercijalna banka had Bosnia in its location set. Was that an error, or were the branches in that country rebranded to just NLB?

data/brands/amenity/bank.json
```diff
@@ -9500,6 +9487,48 @@
...
+      "locationSet": {
+        "include": ["rs"]
+      },
+      "tags": {
+        "amenity": "bank",
+        "brand": "NLB",
+        "brand:wikidata": "Q20566771",
+        "name": "НЛБ Банка Скопје",
+        "name:mk": "НЛБ Банка Скопје",
+        "name:al": "NLB Banka Skopje"
```
Snowysauce, line 9515: `name:al` is a non-standard language tag per https://wiki.openstreetmap.org/wiki/Key:name#Variants and https://en.wikipedia.org/wiki/en:List_of_ISO_639-1_codes. If the language in question is Albanian, the preferred tag is `name:sq`.

PR #10728, opened 2025-03-18
[amenity/Bank] add Societe Arabe Internationale de Banque

ZiadTT:

https://www.saib.com.eg/en/about-us/
And
https://www.wikidata.org/wiki/Q59050464

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -19479,6 +19479,23 @@
         "name:en": "Kagoshima Bank",
         "name:ja": "鹿児島銀行"
       }
+    },{
+    "displayName": "Societe Arabe Internationale de Banque-saib",
+    "locationSet": {
+        "include": [
+            "eg"
+        ]
+    },
+    "tags": {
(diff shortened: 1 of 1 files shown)


Snowysauce:
The bank is notable enough for inclusion in the NSI, but it's not clear at all what the `name` and `brand` tag values should be. I entered every search query I could think of into overpass-turbo, and they all returned nothing. If any of the bank's branches are currently mapped in OSM, I couldn't find them.

For the record, these are the search queries I attempted:
<details>
<summary>Full list</summary>

- "name=Societe Arabe Internationale de Banque-saib"
- "name=Societe Arabe Internationale de Banque"
- "name=saib", "name=Saib", "name=SAIB"
- "name=بنك الشركة المصرفية العربية الدولية"
- all of the above values with the "brand" key
- regex search for any key containing "wikidata" with a value of "Q59050464"
</details>

The bank's website appears to exclusively use "saib" or "SAIB" - even for mentions of the bank in Arabic text - and one of the two should likely be the name/brand value. The formal names in French and Arabic can be match names.

PR #10729, opened 2025-03-18
[amenity/Bank] add Agricultural Bank Of Egypt

ZiadTT:

https://www.abe.com.eg/
And
https://www.wikidata.org/wiki/Q54874868

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -19479,6 +19479,23 @@
         "name:en": "Kagoshima Bank",
         "name:ja": "鹿児島銀行"
       }
+    },{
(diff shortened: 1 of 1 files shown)


Snowysauce:
The bank's website appears to be only viewable to users from certain countries, and maybe only from Egypt. I've tried to visit it at various times since Saturday with no success, and the most recent cached version of the site on the Internet Archive's Wayback Machine is from 2021. As a result, I'm unable to verify whether or not this bank meets the NSI's notability criteria.
(For what it's worth, Overpass-turbo returns only one result from OSM.)

Snowysauce:
Closing due to being unable to verify whether or not this bank meets the NSI's notability criteria.

PR #10735, opened 2025-03-18
Update fuel.json [amenity/fuel] Add Egypt country to Mobil fuel brand

ZiadTT:
Add Egypt country to Mobil fuel brand 
https://www.wikidata.org/wiki/Q109676002
And 
https://www.mobil.com/en

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -4024,7 +4024,7 @@
       "id": "mobil-00679c",
       "locationSet": {
         "include": ["001"],
-        "exclude": ["jp"]
+        "exclude": ["jp","eg"]
       },
       "matchNames": ["mobile"],
       "note": "This wikidata isn't Q3088656",


Snowysauce:
This PR would remove the Mobil preset from Egypt, not add it. Is this the intended outcome?

ZiadTT:
> This PR would remove the Mobil preset from Egypt, not add it. Is this the intended outcome?

Hi No I need to add Egypt not remove it

panasonic-mafia:
@ZiadTT Mobil currently has:
```
"locationSet": {
         "include": ["001"],
         "exclude": ["jp"]
       }
```
It means include whole world (`001`) and exclude Japan. Egypt is already included by `001`. You ca read more about locationSet [here](https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#locationset-tips).
I think this PR can be closed.

PR #10738, opened 2025-03-18
Mixed upper and lower case brand names for supermarkets

map-per:
Changed the brand names of Aldi, Edeka, Norma and Rewe to user mixed upper and lower case instead of upper case only. The change is similar to the change for _Penny_ in #9337.

This makes `name` and `brand` more consistent with other brands.

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -257,7 +257,7 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


bhousel:
Cool yes, time for the annual changing of the brand capitalization.  This is becoming just like daylight savings time.

Issue #10372, opened 2025-01-18
Change brand Orange


Oivo35:
Change brand [Orange](https://nsi.guide/index.html?t=brands&k=shop&v=mobile_phone#orange-ee5434) from `shop=mobile_phone` to `shop=telecommunication` because Orange sells not only mobile phones and contracts but also landline, internet and IPTV services.

Snowysauce:
This suggestion makes sense to me, but per #10829 a change to a brand this large requires a discussion with a larger audience (e.g., the OSM community forums). There are [almost 1700 matches](https://overpass-turbo.eu/s/21JW) for `name=Orange` and `shop=mobile_phone`.

Cj-Malone:
https://community.openstreetmap.org/t/nsi-change-orange-brand-from-shop-mobile-phone-to-shop-telecommunication/128392 🙃

Snowysauce:
Well, that conversation was... interesting. Closing this issue as "no consensus".

ilias52730:
I [restarted](https://community.openstreetmap.org/t/nsi-change-orange-brand-from-shop-mobile-phone-to-shop-telecommunication/128392/14) that discussion. It seems there's a huge misunderstanding where people tagged telecommunication shops as mobile phone shops.

bhousel:
> I [restarted](https://community.openstreetmap.org/t/nsi-change-orange-brand-from-shop-mobile-phone-to-shop-telecommunication/128392/14) that discussion. It seems there's a huge misunderstanding where people tagged telecommunication shops as mobile phone shops.

FWIW, I think both of these tags are so similar that it doesn't really matter much what people choose - but it does sound like they could be retagged if they also offer home internet service.

It's a shame that the OSM forums are so bad. 😕

ilias52730:
It's a shame indeed. Only sarcasm (*"maybe they sell vegetables"*) and no actual research whatsoever. In any case, thanks for giving your vote. I'll create a pull request changing only Orange from mobile shop to telecom shop in the next few days.

ilias52730:
See https://community.openstreetmap.org/t/rfc-feature-proposal-shop-telecommunication/135231

PR #10743, opened 2025-03-19
[Office/telecommunication] add Orange Egypt to telecommunications bra…

ZiadTT:
…nd in EGY

Add new entry for telecommunications brand in Egypt https://www.orange.eg/en/
And 
https://www.wikidata.org/wiki/Q266341

### DIFF
--- a/data/brands/office/telecommunication.json
+++ b/data/brands/office/telecommunication.json
@@ -552,6 +552,23 @@
         "name:en": "Mobilis",
         "office": "telecommunication"
       }
+    },{
+    "displayName": "Orange Egypt",
+    "locationSet": {
+        "include": [
+            "eg"
+        ]
+    },
+    "tags": {
+        "brand": "Orange Egypt",
+        "brand:en": "Orange Egypt",
+        "brand:ar": "اورنچ مصر",
+        "brand:wikidata": "Q266341",
+        "name": "Orange Egypt",
+        "name:en": "Orange Egypt",
+        "name:ar": "اورنچ مصر",
+        "office": "telecommunication"
     }
+}
   ]
 }


Oivo35:
Orange operates worldwide and is now in NSI as [Orange](https://nsi.guide/index.html?t=brands&k=shop&v=mobile_phone#orange-ee5434) shop=mobile_phone and [Orange Mali](https://nsi.guide/index.html?t=brands&k=shop&v=telecommunication#orangemali-478897) as shop=telecommunication.
In Mali, Orange has the same corporate identity as worldwide.
See [https://www.facebook.com/orange.mali](url)

Both of these presets should be merged under one brand - Orange as shop=telelcommunication.
Similar issue as #10372

ZiadTT:
> Orange operates worldwide and is now in NSI as [Orange](https://nsi.guide/index.html?t=brands&k=shop&v=mobile_phone#orange-ee5434) shop=mobile_phone and [Orange Mali](https://nsi.guide/index.html?t=brands&k=shop&v=telecommunication#orangemali-478897) as shop=telecommunication. In Mali, Orange has the same corporate identity as worldwide. See [https://www.facebook.com/orange.mali](url)
> 
> Both of these presets should be merged under one brand - Orange as shop=telelcommunication. Similar issue as #10372

Thank you 
I think Vodafone (Egypt) Also the same case , it is under shop=mobile_phone in NSI , but it should be under shop=telelcommunication

Snowysauce:
I agree with @Oivo35 that the primary preset for Orange should be enough to cover the brand wherever it operates. However, I'm refraining from using the word "global" to describe the preset because [they apparently only have consumer retail operations in 26 countries](https://www.orange.com/en/orange-world). Whatever the Orange preset ends up being tagged as (`shop=mobile_phone` or `shop=telecommunication`), its `locationSet` needs to be reined in - as it currently is, the brand's name is so generic that the chance of false positives in places like North America is high.

PR #10749, opened 2025-03-19
Update furniture.json for Poland new brand

katslodz:
Adding new brand for shop=furniture in Poland called Komfort

https://www.wikidata.org/wiki/Q127275164
https://komfort.pl/


UKChris-osm:
Added via [bce57a7](https://github.com/osmlab/name-suggestion-index/commit/bce57a7d0d5b56fac585580815c94cbc8f8d05b2)

Issue #10011, opened 2024-10-06
Consignaction beverage container return in Québec, Canada


une-abeille-osm:
https://consignaction.ca/

Two entries are needed: `name=Consignaction` and `name=Consignaction+`. The only difference between the facilities is that the "plus" variety offer a bulk service. I do not believe that there is a tag for this in OSM, so the rest of their tags could be the same:

```
amenity=recycling
operator=Consignaction
recycling:beverage_cartons=yes
recycling:cans=yes
recycling:glass_bottles=yes
recycling:plastic_bottles=yes
recycling_type=centre
```

According to the operator's website, all facilities have the same hours of operation, so perhaps those should also be added?
```
opening_hours=Mo-We 08:00-18:00; Th-Fr 08:00-19:00; Sa-Su 08:00-18:00
```

Snowysauce:
A search on Wikidata for "Consignaction" returned no results. Entries need a Wikidata page and tag in order to be shown as presets/suggestions in OSM editors. Please see [Technical Details](https://github.com/osmlab/name-suggestion-index/wiki/Technical-Details#availability-of-entries-in-downstream-projects) for more information, and [Editing Wikidata](https://github.com/osmlab/name-suggestion-index/wiki/Editing-Wikidata#creating-wikidata-pages) for a guide to creating and editing Wikidata pages.

Snowysauce:
https://www.wikidata.org/wiki/Q132858940

After re-reading the issue, it sounds like the requested presets are better suited for the brand tree, not the operator tree. The relevant file in the operator tree (`operators/amenity/recycling`) is designed to have only one entry for a specific operating entity, and since Consignaction operates both "Consignaction" and "Consignaction+", the two types of centers would not be able to have separate presets. In addition, we'd likely have to omit the name tag to allow Consignaction to be listed in that file at all per past precedent.

On the other hand, while listing Consignaction in the brand tree's file (`brands/amenity/recycling`) would allow for two separate entries for "Consignaction" and "Consignaction+" with the desired name tags in place, it would also mean that the `operator=Consignaction` tag would be omitted from the presets per [general NSI practice for brands](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#adding-specific-keys).

In either case, `opening_hours` probably wouldn't be included as a preset tag; see #10170 and #10589.

wolfy1339:
Added Consignaction in #10754
[...]

PR #10754, opened 2025-03-19
Add Consignaction in Quebec

wolfy1339:


Snowysauce:
Related discussion: #10011 

The concerns I had there seem to be addressed by this PR, so this seems good to merge.

PR #10758, opened 2025-03-21
Update restaurant.json - Wrong key value 'bar&grill' (tag cuisine)

mutipg:
Wrong key value 'bar&grill' changed to 'bar_and_grill' (tag cuisine)

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -117,7 +117,7 @@
         "amenity": "restaurant",
         "brand": "54th Street Grill and Bar",
         "brand:wikidata": "Q131290500",
-        "cuisine": "american;bar&grill",
+        "cuisine": "american;bar_and_grill",
         "name": "54th Street Grill and Bar"
       }
     },
@@ -1063,7 +1063,7 @@
         "amenity": "restaurant",
         "brand": "Bubba's 33",
         "brand:wikidata": "Q119359352",
-        "cuisine": "pizza;burger;bar&grill",
+        "cuisine": "pizza;burger;bar_and_grill",
         "name": "Bubba's 33"
       }
     },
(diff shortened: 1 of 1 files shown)


Cj-Malone:
`bar&grill` has way more uses.

https://taginfo.openstreetmap.org/keys/cuisine#values

mutipg:
This is due to a mistake in the wiki, which I have already corrected. Key values ​​should not contain special characters, and 'ampersand' is one of them. I have also changed all occurrences of bar&grill in the osm data to bar_and_grill.

Snowysauce:
This is tricky. I agree that the changed value is in line with general OSM wiki guidance, but it's also a change I would have preferred to see happen organically (i.e., the cuisine wiki page gets updated, the change sticks, and then the values gradually change over as editors consult the page for guidance). At the same time, I realize that keeping the previous value in the NSI presets doesn't encourage that at all, and actually encourages editors to switch back to the previous value.

I suppose this can be merged if we get to the time of the next release and the edit to the cuisine wiki page hasn't been reverted,

EvanCarroll:
What i'm confused about here is "grill" is explicitly documented as "bar and grill"

> A restaurant serving grilled food. Sometimes called a "Bar & Grill" in the United States.	

Why not just delete both "bar&grill" and "bar_and_grill" and rename to grill with `alcohol=yes` and `drink=beer`.

That they serve alcohol is totally irrelevant to the cuisine.

Snowysauce:
The changes to the wiki page and the OSM data are still in place as of this writing, and since the next release is imminent, I'll merge this PR.

> Why not just delete both "bar&grill" and "bar_and_grill" and rename to grill with `alcohol=yes` and `drink=beer`.

This could be a discussion worth having, but since there are almost 700 uses of `cuisine=bar_and_grill` (either by itself or in combination with something else), a better venue for the discussion would probably be something like the OSM discussion forums.

PR #10764, opened 2025-03-23
[amenity/charging_station] add circontrol

Mukhtar27:
Adding New Brand Entry in UAE and Spain
https://www.wikidata.org/wiki/Q104543367

### DIFF
--- a/data/brands/amenity/charging_station.json
+++ b/data/brands/amenity/charging_station.json
@@ -535,6 +535,22 @@
         "name:en": "NIO Power Charger",
         "name:zh": "蔚来超充站"
       }
+    },{
+    "displayName": "circontrol",
+    "locationSet": {
+        "include": [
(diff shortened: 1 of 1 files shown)


Snowysauce:
The company's website seems to indicate that they're more of what OSM would consider a "manufacturer" than a "brand" (see #10181). Unfortunately, due to [how the most common data consumers use the NSI data](https://github.com/osmlab/name-suggestion-index/wiki/Technical-Details#matching), we have to choose whether we match for the `brand` of an object or for the entity that operates it (`operator`).

Snowysauce:
> we have to choose whether we match for the `brand` of an object or for the entity that operates it (`operator`).

Closing this PR as it's generally preferred to match for the operator of items such as charging stations or surveillance cameras instead of their manufacturer.

PR #10765, opened 2025-03-23
[office/telecommunication] add etisalat and

Mukhtar27:
Added new brand in UAE
https://www.wikidata.org/wiki/Q543449

### DIFF
--- a/data/brands/office/telecommunication.json
+++ b/data/brands/office/telecommunication.json
@@ -552,6 +552,23 @@
         "name:en": "Mobilis",
         "office": "telecommunication"
       }
+    },{
+    "displayName": "e&",
+    "locationSet": {
+        "include": [
+            "ae"
+        ]
+    },
+    "tags": {
+        "brand": "etisalat and",
+        "brand:en": "etisalat and",
+        "brand:ar": "اتصالات والمزيد",
+        "brand:wikidata": "Q543449",
+        "name": "Etisalat",
+        "name:en": "Etisalat",
+        "name:ar": "اتصالات",
+        "office": "telecommunication"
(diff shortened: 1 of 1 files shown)


Oivo35:
If it is shop which sell phone contracts consider use [https://wiki.openstreetmap.org/wiki/Tag:shop%3Dtelecommunication](url).

Snowysauce:
`name=Etisalat` and `office=telecommunication` returned 14 results in the UAE, but the POIs are in locations that make me think that they're retail stores mistagged as offices.

bhousel:
> `name=Etisalat` and `office=telecommunication` returned 14 results in the UAE, but the POIs are in locations that make me think that they're retail stores mistagged as offices.

This might make sense - a long time ago the OSM wiki recommended `office=telecommunication` even for retail stores, but I think the wiki guidance was fixed around sometime around when we did https://github.com/osmlab/name-suggestion-index/pull/3120

Snowysauce:
This brand also operates in Egypt under the same name; see https://www.eand.com/en/international-telecoms.html and #10769.

Snowysauce:
The brand has been added to `shop/communication` per https://github.com/osmlab/name-suggestion-index/pull/10765#issuecomment-2748992611 and https://github.com/osmlab/name-suggestion-index/pull/10765#issuecomment-2748530857, and added under the name/brand "e&" per the brand's logo in English and Arabic (both of which use "e&") and the brand's English Wikipedia article (which opens with the phrase "e&, formerly branded as Etisalat"). "Etisalat", "etisalat and", and their Arabic equivalents have been included as match names.

PR #10766, opened 2025-03-23
Lululemon China location

peternewman:
From https://www.lululemon.cn/


LaoshuBaby:
LGTM, because lululemon don't use localized Chinese name in China but just those alphabets as brand, so using global preset is acceptable I think.

Besides, If possible this PR can be submitted with #10760 #10761 at one time because they are related and similar.

peternewman:
> LGTM, because lululemon don't use localized Chinese name in China but just those alphabets as brand, so using global preset is acceptable I think.

Thanks, I hadn't even considered the language side of it TBH.

> Besides, If possible this PR can be submitted with #10760 #10761 at one time because they are related and similar.

Yeah sorry, @UKChris-osm kind of beat me to it on #10760 and then I only realised China was on there after looking at the Wikidata post #10761 .

I've also opened https://github.com/alltheplaces/alltheplaces/issues/12690 which I guess may shake out any other countries I've missed.

PR #10769, opened 2025-03-24
[office/telecommunication] add e& egypt to the telecommunications EGY

ZiadTT:
Adding new entry for a brand in Egypt for telecommunication
 http://www.eand.com.eg/StaticFiles/portal2/etisalat/index_en.html 
 And
https://www.wikidata.org/wiki/Q4120269

### DIFF
--- a/data/brands/office/telecommunication.json
+++ b/data/brands/office/telecommunication.json
@@ -552,6 +552,23 @@
         "name:en": "Mobilis",
         "office": "telecommunication"
       }
+    },{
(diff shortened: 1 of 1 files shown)


Snowysauce:
If this is the same brand as #10765, this PR should be merged with that one.

Snowysauce:
Closing per https://www.eand.com/en/international-telecoms.html and https://github.com/osmlab/name-suggestion-index/pull/10769#issuecomment-2748535398.

Issue #10505, opened 2025-02-10
Add truck brand Iveco


Oivo35:
[https://www.iveco.com/global/welcome/](url)

    {
      "displayName": "Iveco",
      "locationSet": {"include": ["001"]},
      "preserveTags": ["^name"],
      "tags": {
        "brand": "Iveco",
        "brand:wikidata": "Q29913",
        "name": "Iveco",
        "shop": "truck"
      }

Snowysauce:
I'm personally always a bit suspicious of "new" global brands because this project periodically reads the OSM planet file and automatically picks up potential brands that have at least 50 instances globally. Thus, if there's a brand (such as this one) that touts a global presence without coming onto our radar through that method, it's worth looking into whether their claims are exaggerated.

Based on the linked website, the company does not operate in most of North America, and as such the continent's largest countries (Canada, Mexico, and the US) should be excluded from the `locationSet` at minimum.

These are the overpass-query results for "[Iveco](https://overpass-turbo.eu/s/1Yx0)" and "[IVECO](https://overpass-turbo.eu/s/1Yx1)"; instances of `shop=truck` are in green, other values of `shop` are in yellow, and POIs without a `shop` tag (potentially unrelated) are in red. They confirm my concerns about North America, but they also show that "global" status may be warranted for everywhere else. There are 91 POIs in OSM scattered across the world with one name or the other and a `shop` tag, but there isn't a single key-value pairing (`shop=*`) that's common enough to get noticed in the OSM planet file. `shop=truck` may indeed be the ideal key-value pair for the brand, but I'm not familiar with the brand at all (again, no North American presence from Mexico north) and thus can't provide an educated opinion on the matter.

Snowysauce:
OSM stats as of this writing (including both casing variants of "Iveco"):

| tag | quantity | note |
| --- | --- | --- |
| shop=car_repair | 40 | |
| shop=car | 34 | |
| shop=truck | 12 | |
| shop=car_parts | 4 | |
| shop=convenience | 1 | potentially unrelated |
| shop=yes | 1 | potentially unrelated |
| **Total** | **92** | 90 likely relevant

[...]

PR #10772, opened 2025-03-24
[shop/truck_repair] add Iveco

panasonic-mafia:
rel: #10505

I kept same location set as in shop/truck entry for Iveco because I think it's sensible.


panasonic-mafia:
Thank you both!

PR #10773, opened 2025-03-24
Change the Wikidata item for A+ Glass

Albynton:
Wikidata item [Q123409768](https://www.wikidata.org/wiki/Q123409768) is for a different shop named "Active Green + Ross", located in Canada. The item is [already linked to the correct brand in NSI](https://nsi.guide/?t=brands&k=shop&v=car_repair&tt=Active%20Green%20%2B%20Ross#activegreenross-9c9847), with ID activegreenross-9c9847.

I could not find a matching Wikidata item for A+ Glass.

### DIFF
--- a/data/brands/shop/car_repair.json
+++ b/data/brands/shop/car_repair.json
@@ -88,13 +88,13 @@
       }
     },
     {
-      "displayName": "A+ Glass",
+      "displayName": "A+Glass",
       "id": "aglass-98e3a2",
       "locationSet": {"include": ["fr"]},
       "tags": {
-        "brand": "A+ Glass",
+        "brand": "A+Glass",
         "brand:wikidata": "Q116688243",
-        "name": "A+ Glass",
+        "name": "A+Glass",
         "service:vehicle:glass": "yes",
         "shop": "car_repair"
       }
--- a/data/brands/shop/car_repair.json
+++ b/data/brands/shop/car_repair.json
@@ -93,7 +93,7 @@
       "locationSet": {"include": ["fr"]},
       "tags": {
         "brand": "A+ Glass",
-        "brand:wikidata": "Q123409768",
+        "brand:wikidata": "Q116688243",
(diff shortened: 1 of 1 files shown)


Snowysauce:
Thanks for spotting the error. Does [Q116688243](https://www.wikidata.org/wiki/Q116688243) appear to be the correct QID for A+ Glass?

Albynton:
> Does [Q116688243](https://www.wikidata.org/wiki/Q116688243) appear to be the correct QID for A+ Glass?

Yes, it appears to be correct.

Though I see that they write it "A+Glass" in they branding, but NSI's entry is "A+ Glass" (with a space). Should we change the tags to match them?

Snowysauce:
> Though I see that they write it "A+Glass" in they branding, but NSI's entry is "A+ Glass" (with a space). Should we change the tags to match them?

I'm genuinely not sure. OSM has the brand consistently mapped as "A+ Glass", even on POIs not touched by our preset (i.e., `brand:wikidata` is missing). It could be a case of editors consistently copy/pasting bad data, of course, or I could be missing something. I agree that "A+Glass" is the name/brand value I'd use based on their website.

Albynton:
I checked a few POIs for the brand, almost all were renamed last year by the account [votre_solution_de_visibilite_digitale_locale](https://www.openstreetmap.org/user/votre_solution_de_visibilite_digitale_locale), who claims to act in the the shop owners' name.

In a previous edit, the same account wrote it "A+Glass".
Example: https://www.openstreetmap.org/node/11064497105/history

With a single account responsible for most of the names, I'd hardly call it a consensus.

Snowysauce:
Thanks for looking into the situation. Given that additional info, I have no objections to the NSI preset tags being changed to "A+Glass". If you want to update this PR with the correct Wikidata QID and the changed tag values, I think it'd be fine to merge.

Albynton:
Thank you for your help @Snowysauce!

PR #10778, opened 2025-03-26
Update bank.json adding Compass in the category bank in Italy

vinzdb:
Added new bank company for Italy
https://www.compass.it/
Compass Banca (Q127513540)


Snowysauce:
The bank appears to be notable enough for inclusion (30+ branches), but I'm not sure if the tag values are correct. Overpass-turbo returns 14 results in OSM for `amenity=bank` and `name=Compass`, and 0 results for `amenity=bank` and `name=Compass Banca`. (There is only one POI named "Compass Banca" in OSM at all; it's tagged as `shop=money_lender`.)

vinzdb:
Compass Banca in Italy is a financial institution that primarily offers banking services, including personal loans, credit cards, and financing solutions. It is not categorized as a traditional money lender but operates as a bank within the banking system. Regarding the name I agree that "Compass" would fit better as it is how it is displayed on the official website and on the storefont in reality.

PR #10787, opened 2025-03-26
[amenity/post_office] add Purolator

meyerk1456:
Adding a new brand for a post office in Canada.

https://www.wikidata.org/wiki/Q7261372
https://www.purolator.com/en

### DIFF
--- a/data/brands/amenity/post_office.json
+++ b/data/brands/amenity/post_office.json
@@ -611,6 +611,16 @@
         "name": "顺丰速运",
         "name:en": "SF Express",
         "name:zh": "顺丰速运"
+      } 
+    },  
+    {   
+        "displayName": "Purolator",
+        "locationSet": {"include": ["ca"]},
+        "tags": {
+          "brand": "Purolator",
+          "brand:wikidata": "Q7261372",
+          "name": "Purolator",
+          "amenity": "post_office"
       }
     }
   ]


Snowysauce:
Their [store locator](https://www.purolator.com/en/shipping/find-shipping-centre) does not make it easy to verify how many standalone locations this brand has. Its article on the English Wikipedia states that there were 104 "shipping centres" in 2021, but a lot can change in four years. If we were to consider just drop boxes, the brand would unquestionably qualify for the NSI due to the sheer number of those locations.

meyerk1456:
Thank you for your feedback.
The Purolator website confirms:
104 Retail Locations (physical stores), 250 Drop Boxes, 41 Parcel Lockers, and Additional locations within retail stores like Staples
Source: https://www.purolator.com/en/shipping/shipping-locations-access-points
 
Given these numbers, would Purolator be a better fit under office=courier?
Currently, NSI does not have data/brands/office/courier, but adding it might be a better long-term solution. Alternatively, if drop boxes are enough to qualify, should this be added under operators/amenity/post_box.json instead?

Happy to adjust based on what fits best and makes more logical sense for NSI.

Snowysauce:
Thanks for finding that link. At 104 active locations, the brand easily clears the NSI's notability criteria.

As for `amenity=post_office` vs. `office=courier`, that seems to be a judgment call based on quick reads of https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dpost_office and https://wiki.openstreetmap.org/wiki/Tag:office%3Dcourier. To date, OSM editors have seem to have chosen `amenity=post_office` ([11 instances](http://overpass-turbo.eu/s/21Gn)) instead of `office=courier` ([0 instances](http://overpass-turbo.eu/s/21Gp)).

PR #10796, opened 2025-03-27
[shop/convenience] add Corner Store

meyerk1456:
Adding a new brand for shop=convenience in Canada.

https://www.wikidata.org/wiki/Q132801808
https://www.corner-store.ca/en/

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -8317,6 +8317,16 @@
       }
     },
     {
+      "displayName": "Corner Store",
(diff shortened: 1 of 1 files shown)


Snowysauce:
The provided URL (and thus also the one on the Wikidata item) redirects to <https://journie.ca/on-the-run-ca/on-en>, which belongs to the brand "On the Run". That brand is [already in the NSI](https://nsi.guide/?t=brands&k=shop&v=convenience#ontherun-4d6b4b).

PR #10802, opened 2025-03-27
[shop/mobile_phone] change brand value for Koodo

meyerk1456:
Update brand to Koodo Mobile to match wikidata page.

https://www.wikidata.org/wiki/Q6430529
https://www.koodomobile.com/en

### DIFF
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -434,7 +434,7 @@
       "id": "koodo-947b24",
       "locationSet": {"include": ["ca"]},
       "tags": {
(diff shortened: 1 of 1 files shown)


Snowysauce:
This change is debatable; while their legal name appears to be Koodo Mobile, they seem to use just "Koodo" as their brand name.

matkoniecz:
from looking at banner on their website and https://duckduckgo.com/?t=ffab&q=koodo+shop&iax=images&ia=images it seems that it should not be changed

> to match wikidata page

maybe wikidata prefers official name, maybe wikidata is wrong

PR #10815, opened 2025-03-27
Update social_centre.json(Update brand CAN)

meyerk1456:
Updated the name for Royal Canadian Legion to match Wikidata and official website.

https://www.wikidata.org/wiki/Q3270231
https://legion.ca/

### DIFF
--- a/data/brands/amenity/social_centre.json
+++ b/data/brands/amenity/social_centre.json
@@ -173,15 +173,16 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


Snowysauce:
This preset is referring to the Royal Canadian Legion in the context of their meeting locations. Based on the Legion's website, a case could be made that they should be referred to as "branches" instead of "halls", but removing "Hall" without a replacement is incorrect.

PR #10817, opened 2025-03-28
[shop/hardware] add BMR

meyerk1456:
Adding a new brand for shop=hardware in Canada.

https://www.wikidata.org/wiki/Q3117331
https://www.bmr.ca/en/

### DIFF
--- a/data/brands/shop/hardware.json
+++ b/data/brands/shop/hardware.json
@@ -479,6 +479,16 @@
         "name": "Мосхозторг",
         "shop": "hardware"
       }
+    },
+    {
+      "displayName": "BMR",
+      "locationSet": {"include": ["ca"]},
+      "tags": {
+        "brand": "BMR",
+        "brand:wikidata": "Q3117331",
+        "name": "BMR",
+        "shop": "hardware"
+      }
     }
   ]
 }
--- a/data/brands/shop/hardware.json
+++ b/data/brands/shop/hardware.json
@@ -483,9 +483,10 @@
     {
       "displayName": "BMR",
       "locationSet": {"include": ["ca"]},
+      "preserveTags": ["^name"],
       "tags": {
         "brand": "BMR",
-        "brand:wikidata": "Q3117331",
+        "brand:wikidata": "Q126810983",
         "name": "BMR",
         "shop": "hardware"
       }


kjonosm:
There is also https://www.wikidata.org/wiki/Q126810983
Which wikidata item should be used here?

Snowysauce:
[Q3117331](https://www.wikidata.org/w/index.php?title=Q3117331&oldid=2245051653) appears to be intended for the company, which operates at least two brands according to the French Wikipedia's article. Q126810983 is probably more appropriate as the `brand:wikidata` value.

Snowysauce:
In addition, the chain's store locator indicates that each location has a unique name, meaning `"preserveTags": ["^name"],` has to be added to the preset.

wolfy1339:
I wouldn't put too much importance to the name given on the store locator.
> BMR dealers are cooperatives and independent entrepreneurs

Most of those names are made up of the brand and operator, or brand, operator and branch, or simply operator
ex: BMR Avantis Coopérative (St-Georges)
```
brand=BMR
operator=Avantis Coopérative
branch=St-Georges
```
ex: Matériaux Lavoie
```
brand=BMR
operator=Matériaux Lavoie
```

The most prominent name displayed is BMR and people would search for that and not  BMR Avantis Coopérative (St-Georges).

Snowysauce:
> Most of those names are made up of the brand and operator, or brand, operator and branch, or simply operator

> The most prominent name displayed is BMR and people would search for that and not BMR Avantis Coopérative (St-Georges).

Noted, but this doesn't match with how most of the brand is currently mapped in OSM. There are [15 results in overpass-turbo](http://overpass-turbo.eu/s/21Gj) for `shop=hardware` and `name=BMR`, and [33 results](http://overpass-turbo.eu/s/21Gi) for `shop=hardware` and a regex search where "BMR" is part of the value of `name` but `name` is not just "BMR". It's entirely possible that the 33 latter POIs should have `name=BMR` and `branch=(remainder of the existing name value)`, but that'd require research on each individual POI by someone familiar with the brand.

meyerk1456:
Thank you for the feedback. I can update the brand:wikidata to Q126810983 as well as add "preserveTags": ["^name"] if it is more preferable.

PR #10818, opened 2025-03-28
[office/tax_advisor] add MNP

meyerk1456:
Adding a new brand for office": "tax_advisor in Canada.

https://www.wikidata.org/wiki/Q6826483
https://www.mnp.ca/

### DIFF
--- a/data/brands/office/tax_advisor.json
+++ b/data/brands/office/tax_advisor.json
@@ -132,6 +132,16 @@
         "name": "Vereinigte Lohnsteuerhilfe",
         "office": "tax_advisor"
       }
+    },
+    {
+      "displayName": "MNP",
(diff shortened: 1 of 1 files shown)


Snowysauce:
The brand is notable enough for inclusion in the NSI, but not as `office=tax_advisor` as tax services are just part of the brand's offerings. This is backed up by the overpass-turbo search results for the brand, which shows that none of the 13 currently mapped locations (out of 174 overall) are tagged as `office=tax_advisor`. Most are tagged as `office=accountant`, but some are tagged as `office=financial` or `office=financial_advisor`. Please do more research on the brand to determine what the proper tag should be - `office=tax_advisor` isn't it.

PR #10820, opened 2025-03-28
[shop/stationery] add Calendar Club

meyerk1456:
Adding a new brand for "shop": "stationery" in Canada.

https://www.wikidata.org/wiki/Q132798539
https://www.calendarclub.ca/

### DIFF
--- a/data/brands/shop/stationery.json
+++ b/data/brands/shop/stationery.json
@@ -452,6 +452,16 @@
         "name:zh": "金玉堂",
         "shop": "stationery"
       }
+    },
+    {
+      "displayName": "Calendar Club",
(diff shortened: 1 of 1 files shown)


kjonosm:
seems more like a shop=gift. See e.g. mission statement on the website...

Snowysauce:
Agreed, and overpass-turbo shows that existing OSM tagging for the brand is almost unanimously `shop=gift`. The other issue is that [the brand apparently only has 10 physical locations](http://calendar-club.myshopify.com/pages/store-locator). That's at the very limit of what we even consider to be notable enough for the NSI, and far shy of the 30 location recommendation.

PR #10824, opened 2025-03-28
[amenity/social_facility] add Meals on Wheels

meyerk1456:
Adding a new brand for "amenity": "social_facility" in Canada and USA.

https://www.wikidata.org/wiki/Q365857
https://www.mealsonwheels.ca/

### DIFF
--- a/data/brands/amenity/social_facility.json
+++ b/data/brands/amenity/social_facility.json
@@ -726,6 +726,16 @@
         "name:en": "Caritas",
         "name:zh": "明愛會"
       }
+    },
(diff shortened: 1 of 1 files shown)


kjonosm:
Wikidata item seems to be a mix of both the actual business/brand and the general concept of meals on wheels.

Snowysauce:
Agreed. The Wikipedia articles that the QID is linked to discuss the general concept, not a specific brand or brands, so using this QID for any brand wouldn't be appropriate.

PR #10832, opened 2025-03-28
Remove `preserveTags: ["^name"]` from ATMs

Dimitar5555:
Since #10378 and https://github.com/osmlab/name-suggestion-index/commit/91ed2e449273afe3db415b2eef082724cfa197f6 we are no longer in the `name` business of ATMs. 

NSI is generating a preset for it's ATMs that includes a `name` field. This field now stays empty on the top of the list in iD which is likely unwanted. The bug seems to be caused by https://github.com/osmlab/name-suggestion-index/blob/main/scripts/dist_files.js#L368-L372. iD's preset seems to include all necessary tags so it should work fine in theory (see https://github.com/openstreetmap/id-tagging-schema/blob/main/data/presets/amenity/atm.json).

I'm not sure if making this change would break anything on iD's side... Is there any way to test it?

|Branded | Unbranded|
|--|--|
|image|image|

### DIFF
--- a/data/brands/amenity/atm.json
+++ b/data/brands/amenity/atm.json
@@ -2,7 +2,6 @@
   "properties": {
     "path": "brands/amenity/atm",
     "skipCollection": true,
-    "preserveTags": ["^name"],
     "exclude": {"generic": ["^atm$"]}
   },
   "items": [


bhousel:
Yeah, this seems ok to me.
Adding `"preserveTags": ["^name"]` in 91ed2e4 is like saying "we expect these features to have unique names for the user to supply.  That's why `dist_files` adds that field - it's not a bug.

So it sounds like remove that `preserveTags` property should produce the behavior you are looking for.

Dimitar5555:
Confirming that it works as expected:

image

PR #10834, opened 2025-03-30
Fix "France Travail" vs "Pôle Emploi"

nlehuby:
"Pôle Emploi" brand does not exist anymore, it has been replaced by "France Travail" (since 2024-01-01).

And it is a public service, not an employment agency.

Sources:
- https://wiki.openstreetmap.org/wiki/France/Services_Publics#France_Travail_/_P%C3%B4le_Emploi
- https://forum.openstreetmap.fr/t/de-pole-emploi-a-france-travail/20397

### DIFF
--- a/data/brands/office/employment_agency.json
+++ b/data/brands/office/employment_agency.json
@@ -117,19 +117,6 @@
       }
     },
     {
-      "displayName": "France Travail",
-      "id": "francetravail-9e0ae6",
-      "locationSet": {"include": ["fr"]},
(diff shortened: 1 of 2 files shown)


Snowysauce:
I read through the linked discussion on the OSM forum, and if the community's fine with the tag change, there probably won't be any objections here. However, even though it was proposed and accepted by the community as part of the conversion, the `old_name` tag is [typically outside of our scope](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#adding-specific-keys).

PR #10847, opened 2025-04-01
Update of Catalana Occidente due to rebranding to Occident

paunofu:
Catalana Occidente has fusionante all its insurance brands and named as "Occident". Additionally, the Wikidata reference is updated from Q5936881 to Q124218333. Q5936881 makes reference to the parent company and Q124218333 only to occident as an insurance company.

Sources: 
https://www.occident.com/
https://brandemia.org/catalana-occidente-cambia-nombre-occident

@Robot8A

### DIFF
--- a/data/brands/office/insurance.json
+++ b/data/brands/office/insurance.json
@@ -728,13 +728,14 @@
       }
     },
     {
-      "displayName": "Catalana Occidente",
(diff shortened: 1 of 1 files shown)


Robot8A:
Looks good!

PR #10850, opened 2025-04-01
Add Nacex franchises 

Crashillo:
Spanish logistics brand


Snowysauce:
The specified Wikidata QID belongs to Speedy. Was it intended to be [Q126366376](https://www.wikidata.org/wiki/Q126366376)?

Crashillo:
Ouch! bad copy-pasting. Fixed!

PR #10851, opened 2025-04-02
[amenity/bicycle_rental] add CapMetro Bikeshare, feature for Austin, TX

piperswe:
There are 77 CapMetro Bikeshare docking stations, according to https://www.capmetro.org/bikeshare/bikeshare-stations. I will soon be ensuring they are all tagged similarly, following the recent rebrand of the service. Having the brand in iD would make things much easier.


Snowysauce:
Everything proposed here seems to pass the eye test. Some of the tags seem like things the NSI shouldn't be dabbling in, but at the same time none of those tags would be unprecedented additions to the index. The creation of a feature dedicated to Austin is probably OK if more uses can be found for it among existing NSI entries - we try to avoid single-use features as much as possible.

piperswe:
Thank you for the feedback! I've pushed an additional commit which uses the new Austin feature for other City of Austin presets.

PR #10852, opened 2025-04-02
[amenity/vending_machine] add CapMetro TVMs

piperswe:
These TVMs are present at most MetroRail stations.

### DIFF
--- a/data/brands/amenity/vending_machine.json
+++ b/data/brands/amenity/vending_machine.json
@@ -354,6 +354,19 @@
       }
     },
     {
+      "displayName": "CapMetro",
+      "id": "capmetro-4ef908",
+      "locationSet": {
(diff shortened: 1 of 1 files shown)


Snowysauce:
Overpass-turbo returns [0 results](http://overpass-turbo.eu/s/21Da) for `amenity=vending_machine` and `brand=CapMetro`. Is this preset being suggested to help add these machines to OSM, and if so, will there be at least 20-ish instances when most/all of the POIs are mapped?

Snowysauce:
However, there are [four results](https://overpass-turbo.eu/s/21Gg) for `amenity=vending_machine` and `operator=CapMetro`. Would this preset be better suited for `operators/amenity/vending_machine`?

PR #10853, opened 2025-04-02
Add Quironsalud healthcare operator

Crashillo:


Snowysauce:
Clerical note: both categories are ones where iD displays an `operator:type` selector. Omitting a suggested value for this tag leaves the value up to the whim of individual editors, when ideally a given operator should have a consistent `operator:type`.

Issue #10700, opened 2025-03-15
Change GLS Paketomat


Oivo35:
Add wikidata to [GLS Paketomat](https://nsi.guide/index.html?t=brands&k=amenity&v=parcel_locker#glspaketomat-985bb3). All GLS Paketomat lockers in Croatia and Slovenia offers parcel pickup, but not all parcel mail in.
Croatia and Slovenia should also be excluded from global preset [GLS](https://nsi.guide/index.html?t=brands&k=amenity&v=parcel_locker#gls-6a219a).

Suggestion:

    {
      "displayName": "GLS Paketomat",
      "id": "glspaketomat-985bb3",
      "locationSet": {"include": ["hr", "si"]},
      "tags": {
        "amenity": "parcel_locker",
        "brand": "GLS Paketomat",
        "brand:wikidata": "Q366182",
        "name": "GLS Paketomat"
        "parcel_pickup": "yes"
      }
    },

Snowysauce:
Everything suggested was added in https://github.com/osmlab/name-suggestion-index/commit/c3d62e9e4dbcda7f258614971e0bff0aa233ffab except for the `parcel_pickup` tag, which I'm not sure should be in the NSI's scope (see #10657).

Oivo35:
Thanks for clarifying the use of `parcel_pickup` tag. I' see this tag at other parcel pickup presets in NSI and that's why I suggested it.
Now I agree with you.

PR #10855, opened 2025-04-02
remove parcel_mail_in and parcel_pickup

matkoniecz:
see https://github.com/osmlab/name-suggestion-index/pull/10657#issuecomment-2701442463

this changes and NSI cannot keep up with such changes

it already resulted in bogus data, see https://github.com/openstreetmap/id-tagging-schema/issues/1474

fixes https://github.com/openstreetmap/id-tagging-schema/issues/1474

### DIFF
--- a/data/brands/amenity/parcel_locker.json
+++ b/data/brands/amenity/parcel_locker.json
@@ -37,7 +37,7 @@
         "brand": "AlzaBox",
         "brand:wikidata": "Q115254158",
         "operator": "Alza",
-        "operator:wikidata": "Q10786832",\
+        "operator:wikidata": "Q10786832",
       }
     },
     {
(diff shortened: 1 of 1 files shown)


matkoniecz:
if parcel_mail_in should be removed and parcel_pickup should stay  let me know and I will make a new PR

Snowysauce:
I personally have no problems with a mass removal of both - in my view, they're outside the scope of our project. FWIW, #10700 requested that `parcel_pickup` be added to a preset that didn't have it, and I cited #10657 as the reason why I found that change problematic.

matkoniecz:
> [needs discussion ](https://github.com/osmlab/name-suggestion-index/labels/needs%20discussion)

I think that https://github.com/osmlab/name-suggestion-index/pull/10306#issuecomment-2699169091

> NSI should stop setting tags like parcel_mail_in - we should rely on users to ground survey this detail.

> I'm sorry I haven't been paying enough attention to this project and tags like this have slipped into the suggestions. In the past we've been very consistent about not suggesting tags that aren't true everywhere.

is relatively clear?

(though I am not sure about exact governance structure in this project)

Cj-Malone (approved):

PR #10856, opened 2025-04-02
Remove invalid operator:type=private for BKK

gy-mate:
BKK (BKK Budapesti Közlekedési Központ Zrt.) is a public company owned by the local government of Budapest: https://bkk.hu/rolunk/bemutatkozas/kik-vagyunk/#:~:text=A%20BKK%20100%20sz%C3%A1zal%C3%A9kban%20f%C5%91v%C3%A1rosi%20tulajdon%C3%BA%20k%C3%B6zszolg%C3%A1ltat%C3%B3%20t%C3%A1rsas%C3%A1g

Also, operator:* keys are generally not allowed in the NSI: https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#adding-specific-keys

### DIFF
--- a/data/brands/amenity/bicycle_rental.json
+++ b/data/brands/amenity/bicycle_rental.json
@@ -2601,7 +2601,6 @@
         "network": "MOL Bubi",
         "network:wikidata": "Q16971969",
         "operator": "BKK",
-        "operator:type": "private",
         "operator:wikidata": "Q608917"
       }
     },


Snowysauce:
> Also, operator:* keys are generally not allowed in the NSI: https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#adding-specific-keys

`operator:type` is an exception, and is implicitly expected by iD to be supplied by the NSI for [any presets in these categories](https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema%20operator%2Ftype&type=code).

bhousel:
> operator:type is an exception, and is implicitly expected by iD to be supplied by the NSI for [any presets in these categories](https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema%20operator%2Ftype&type=code).

Is there a good reason for this?  I’ve never understood why people want to store `operator:type` in the OSM data.

Snowysauce:
That's a question better suited for id-tagging-schema. I don't have a personal preference one way or the other.

gy-mate:
> `operator:type` is an exception, and is implicitly expected by iD to be supplied by the NSI for [any presets in these categories](https://github.com/search?q=repo%3Aopenstreetmap%2Fid-tagging-schema%20operator%2Ftype&type=code).

@Snowysauce I see—thanks for the commit!

I couldn't find this info on the wiki. Could you please add it there?

gy-mate:
@Snowysauce [Thanks a lot](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case/_compare/8a402b43ddac764aaf4b07fccd7c1d4d459a0113...c0b28162865c451eb75de6448265e0766ce49abc)! :)

Issue #10154, opened 2024-11-18
Aldi Suisse should use Q111030009


brawer:
NSI currently assigns [Q41171672 (Aldi Süd)](https://www.wikidata.org/wiki/Q41171672) to Aldi stores in Switzerland. However, Aldi Switzerland is a separate company with its own Wikidata ID, [Q111030009](https://www.wikidata.org/wiki/Q111030009). Proposal: Change NSI to assign the Wikidata ID  Q111030009 to Aldi stores in Switzerland.

UKChris-osm:
Personally I feel we should stick to the basics within the NSI and focus on just the `brand` and not the underlying operator.

McDs, for example, is operated as a franchise model, with many different operators around the world, and even multiple operators within the same regions, but by focusing on just the `brand`, we can tag it as simply being "_a McDs_", and I think we should do the same with most big brands, unless there is a language specific version, or if the branding looks differently across different areas.

But I've just noticed McDs is split into multiple countries as well for France and Norway, but are using the same tagging, which seems unnecessary.

brawer:
For what it’s worth, Aldi Süd and Aldi Suisse have different logos. So it’s a more user-visible difference than McDonald’s operators.

bhousel:
> But I've just noticed McDs is split into multiple countries as well for France and Norway, but are using the same tagging, which seems unnecessary.

Yeah that does seem redundant - I remember Norway wanted more `"preserveTags": ["^name"]` so it's probably related to that.  And it's ok for different parts of the world to use different spelling or local tags.

But at least they are all still using `"brand:wikidata": "Q38076"` - What we're talking about here would be losing that also, which is the thing I would be sad about.

>For what it’s worth, Aldi Süd and Aldi Suisse have different logos. So it’s a more user-visible difference than McDonald’s operators.

Ah that is good to know - yes if they really have different names or logos, then I could see a case for splitting it into a new brand.   The logo looks the same to me, but with the word "Suisse" below it.  I don't really consider that _different_, but I dunno, maybe Swiss people do 😆

brawer:
So what’s the verdict on Aldi Suisse?

[...]

PR #10857, opened 2025-04-05
Add more Coach countries

peternewman:


data/brands/shop/bag.json
```diff
@@ -64,6 +64,19 @@
...
+      "matchNames": [
+        "coach new york",
+        "coach ny"
+      ],
+      "tags": {
+        "brand": "Coach",
+        "brand:wikidata": "Q117745958",
+        "name": "Coach",
+        "shop": "bag"
+      }
```
peternewman, line 79: For some reason there's a separate entry for this one on Wikidata?!?

data/brands/shop/bag.json
```diff
@@ -53,7 +53,7 @@
     {
       "displayName": "Coach",
       "id": "coach-bdf41b",
-      "locationSet": {"include": ["ca", "us"]},
+      "locationSet": {"include": ["at", "be", "ca", "de", "es", "fr", "gb", "id", "ie", "it", "jp", "mx", "nl", "nz", "us"]},
```
peternewman, line 56: The European site gives a list of a number of countries, others have just been random guesswork. I guess we could try every country code or something?

data/brands/shop/bag.json
peternewman: I can remove the Australian one for now if its a bit controversial and do it in a separate PR to get the simpler one across the line?

data/brands/shop/bag.json
peternewman: Have now merged @Cj-Malone 's ATP data dump into this list, so closing this one for now.

data/brands/shop/bag.json
Snowysauce: I'd imagine the reason this PR is sitting idle is that the location set size is being expanded nine-fold, from two countries to 18, and that at least two of the 18 countries are places where Latin is not the common writing script. This kind of complex change requires a fair amount of verification and research on the part of maintainers, and we've all been pressed for time to some degree in recent weeks.

peternewman: Sorry I'd not considered the non-Latin aspect at all.

For future reference, what sort of verification are you going to do, as it may well be that I've done some of it already. I've personally seen and mapped them in at least the UK (and possibly another country too, I'd have to double-check). Apart from India (where I did a bit of searching) I added the other countries based on store listings on their websites. Although if I'm honest I didn't check the stores actually exist in the places they say they do, but it would suggest they've at least had them in those places at some point.

davidhicks:
Q117745958 (Coach Australia) is for `COACH STORES AUSTRALIA PTY LTD` (https://abr.business.gov.au/ABN/View?abn=48622183224) which is the brand's company in Australia. The problem is Wikidata's typical conflation of brands and companies where they should be separate items. Most multinational brands will have a different company registered in each jurisdiction they operate within, and the various companies in each country license from the parent company somewhere else in the world the trademarks/brand name.

Q727697 (Coach) as a brand (not company) should probably be used by ATP and NSI for the `brand:wikidata` key no matter which country the Coach brand operates within. If anyone felt particularly bored and NSI wanted to go down the path of doing so, individual NSI entries could then be created and differentiate on `operator:wikidata=Q117745958` for Australia, and other operating companies for different jurisdictions.

bhousel:
If I'm understanding correctly - I agree we can probably just use `Q727697` everywhere and close this PR?

I don't think it makes sense for NSI to use the per-country Wikidata items for global brands.  We discussed recently this trend on #10154 too.

davidhicks:
I suggest the only time it may make sense to split brands in NSI on a country-by-country basis is if the brand-specific tags change depending on country. For example, all ACME stores in the UK sell womens clothing only, but all ACME stores in the US sell both womens and mens clothing. I don't think that happens in practice very often though. If NSI were interested in operator-specific tags such as the franchise owner of a store/restaurant, it'd get messy very quickly.

Usually it is Wikidata not NSI nor ATP that's wrong in conflating a multinational company, its various businesses in each country it operates, and the brands owned and/or licensed for use by the multinational company all into a single item. It's not an easy situation either for Wikidata. Wikidata has property https://www.wikidata.org/wiki/Property:P1716 which is meant to map to OSM/NSI's `brand:wikidata` key and the value of this property is meant to be a distinct Wikidata item for the brand common across one or more retail chains (https://www.wikidata.org/wiki/Q507619). Wikidata's definition of a retail chain is loose but appears to be intended to not only differentiate retail outlets on branding, but also their management and ownership. Particularly in complex franchise models, there are sometimes two or three companies operating in a country under the same licensed brand name, with some consistency in management and marketing provided commonly by the brand owner across all licensed franchisee companies.

Wikidata therefore may have an item for the brand, an item for the company which owns the brand, an item for 3 different retail chains using the brand, an item for 3 different companies which operate the 3 retail chains (and possibly other retail chains of different brands), etc. But NSI only cares about an item for the brand.

Even very obvious examples such as KFC (https://www.wikidata.org/wiki/Q524757) remain heavily conflated on Wikidata though. Q524757 (KFC) is a brand. Q668737 (Yum! Brands) is the brand owner. Q28224589 (Collins Foods) is one of the numerous licensed companies operating a chain of KFC restaurants. Q524757 (KFC) is also conflated to potentially mean a subsidiary company of Yum! Brands which is used to hold the KFC brand IP. Q524757 (KFC) is also conflated as an abstract concept of a chain of every KFC restaurant in the world regardless of country of operation and licensed operator/franchise owner (maybe fair enough, per loose definition of a retail chain on Wikidata).

bhousel:
> I suggest the only time it may make sense to split brands in NSI on a country-by-country basis is if the brand-specific tags change depending on country. For example, all ACME stores in the UK sell womens clothing only, but all ACME stores in the US sell both womens and mens clothing. I don't think that happens in practice very often though. If NSI were interested in operator-specific tags such as the franchise owner of a store/restaurant, it'd get messy very quickly.

Yes, there are a few interesting examples of this but they are rare.
For example, the American restaurant chain "Denny's" went to Japan and spun off into its own unique thing there.
https://nsi.guide/index.html?t=brands&k=amenity&v=restaurant&tt=denny
https://yesshewentthere.com/2018/04/10/dennys-in-japan-much-better-food-than-youd-expect/

peternewman:
> If I'm understanding correctly - I agree we can probably just use `Q727697` everywhere

Okay that's fine with me, I've updated the PR to match.

> close this PR?

No sorry, there are still lots of countries that have Coach shops but aren't listed in the location set in NSI, which is the main part of this PR.

Originally I went to map a Coach in the UK and found it wasn't in SC, got home, found the preset was in NSI but only in two countries, assumed there may be more if the UK was missing and found lots via different sources (including the alternative Australian Wikidata). Added those and hence this PR. There's a query over non-Latin countries (which I'm happy to pull out to a separate PR), although at least India has Coach stores already mapped with the Wikidata ID as "Coach". Aside from India (which I checked with a search engine and already has some mapped too), all other countries have stores listed on their Coach website/that's where I sourced the country from.

data/brands/shop/bag.json
Snowysauce: Specifically regarding this PR, the research required is two-fold: whether the brand actually has physical stores in the 16 "new" countries, and whether the stores in the non-Latin countries still have "Coach" on the storefronts and/or as the brand name. I don't have any reason to suspect that anything is amiss with the PR as proposed, but I personally approach location set expansion proposals with an abundance of caution.

(An entity's location set can easily be expanded with no immediate impact to OSM; however, if an entity is given too broad of a location set and has too generic of a name, it usually results in OSM receiving a lot of bad data - bbfeb070738181c24d1e28fc9e2067bd1e1bbfc0 is a perfect example of the latter.)

peternewman: Sorry I thought I'd posted this before, but here's the stores listing for all but India (as mentioned previously) in terms of physical stores:
https://github.com/alltheplaces/alltheplaces/issues/12857

> I personally approach location set expansion proposals with an abundance of caution.

Looking at the Git blame for this, it seems the restricted locationsets were generate by @matkoniecz in https://github.com/osmlab/name-suggestion-index/commit/94bcf90a03271c25b7f1b5ce7dec387c1a5e4c2c presumably based on what was live in OSM at the time. I wonder if there should be some discussion (maybe in a separate issue) of whether that should be re-run a bit like some of the other stuff is pulled in from OSM, as clearly more of this store, and I'm sure others, have been mapped since that was last done.

ilias52730:
This has been sitting idle for 4.5 months. _Coach_ is indeed present in all countries added in this PR. It uses the same branding even in Thailand and Japan (source: [Google Maps images](https://lh3.googleusercontent.com/gps-cs-s/AC9h4nqHYC3zAzqoq-UOhDeMUZ7JbE_Ol4YSRU2UTcgHNyxiQHu5p32ne1U-M1CHel8tOpAaaxBZG5-TKPGPWyyFhNGCTyFX-vna5q04DQ-c6hOlQ8fH6ugyueqJ8sXGuQRXEUmTjbwN6g=w520-h350-n-k-no)). Merging PR.

Issue #10109, opened 2024-10-30
Proper spelling for wholesale METRO (should be Metro)


dobridabar:
Please change spelling for wholesale METRO to Metro, because METRO is not an acronym.

After that, merge with Bulgarian "Metro", which was the only one that was spelled correctly.

Rationale (from: https://wiki.openstreetmap.org/wiki/Names): "Names are adapted to the local capitalisation rules; [title case](https://en.wikipedia.org/wiki/Title_case) is generally preferred, but regional rules have preference over general rules."

It is true that Metro brands themselves in uppercase (METRO), but branding decision does not override OSM naming convention, which is to use title case (Metro). Too many companies brand themselves in uppercase, and writing their names in uppercase on OSM map would give them unfair advantage (*) over other companies which properly capitalize their brands.

(*) by taking more space on the map and being more visible

Cj-Malone:
See #9337 / @Contrib1043

dobridabar:
Hi, I found two more stores in Serbia for which I would like to fix the case:

1. IDEA -> Idea
2. JYSK -> Jysk

Should I make the changes myself and create a PR?

matkoniecz:
> Should I make the changes myself and create a PR?

no idea are these correct changes, but I can say that from my experience making PR greatly increases chance that NSI will make change that someone proposed, compared to just opening an issue

Snowysauce:
Historically, there hasn't been any consensus or consistency one way or the other when it comes to German brands. We've had attempts to change all-caps values to mixed-case values (#9388, #9664), a reversion of some of those attempts (#9596), and in the past month adoptions of both case changes (#10738) and all-caps values (#10875).

dobridabar:
Related: https://github.com/osmlab/name-suggestion-index/issues/10862

dobridabar:
All in all, this is a mess.

I already quoted OSM Wiki page that, in my opinion, clearly favors "local capitalization rules", which is (for Serbia, at least), use of uppercase is allowed only for acronyms.

I am not a language expert, but I believe that most local capitalization rules around the world are similar.

We really need NSI admins to take a clear stance on this question, otherwise this will become a meaningless edit war.

Cj-Malone:
Thanks for the kind words.

bhousel:
> We really need NSI admins to take a clear stance on this question, otherwise this will become a meaningless edit war.

[...]

PR #10861, opened 2025-04-07
Mixed upper and lower case brand names for Edeka and Rewe beverage shops

map-per:
Changed the brand names of Edeka and Rewe beverage shops to use mixed upper and lower case instead of upper case only. The change is a follow up of the change for supermarkets of the same brands in #10738.

This makes `name` and `brand` more consistent with other brands.

### DIFF
--- a/data/brands/shop/beverages.json
+++ b/data/brands/shop/beverages.json
@@ -40,14 +40,14 @@
       }
     },
     {
-      "displayName": "EDEKA Getränkemarkt",
+      "displayName": "Edeka Getränkemarkt",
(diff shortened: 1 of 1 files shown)


Snowysauce:
Closing per https://github.com/osmlab/name-suggestion-index/issues/10109#issuecomment-2800267007 - the OSM community needs to decide which tagging style to use, then come back to us with the consensus decision. The tug of war based on personal preferences (referenced by https://github.com/osmlab/name-suggestion-index/issues/10109#issuecomment-2803297473) is ridiculous.

Issue #10865, opened 2025-04-07
Add support for trail operators


quincylvania:
Okay here's a tricky one. Over at the Trails Stewardship Initiative [we're recommending](https://wiki.openstreetmap.org/wiki/United_States/Trails_Stewardship_Initiative#Trail_Tagging_Chart) that all trails be tagged with `operator` tags if applicable, and otherwise be tagged `informal=yes`. This is an important distinction since use of informal trails can lead to public endangerment and ecological damage. Adding an operator can be relatively tedious, but maybe the NSI can help with that!

Trails are a different kind of data than the POIs that are typical of the NSI. They can have various primary tags, usually one of `highway=path/footway/cycleway/bridleway/track`. Trails are often grouped into routes like `route=foot/hiking/bicycle/mtb/horse`, and these also have operator tags. But trail operators are often land managers who run all kinds of other features that people might want to map with `operator`, like service roads, campgrounds, picnic sites, information boards, slipways, etc.

While an example of an existing tag would be `operator=National Park Service`, it's likely that people will want to start tagging specific units of the NPS as operators, like `operator=Badlands National Park`. There are thousands of park units, ranger districts, land conservancies, etc. in the US alone.

The NSI would be an amazing place to standardize tagging for operators of trails and recreation features, but the existing model of defining iD presets won't really work here. A better model would be to define a list of independent operators here, and then surface them in iD through a specialized Operator field.

What do you all think? I'd be happy to help build these features.

Snowysauce:
> The NSI would be an amazing place to standardize tagging for operators of trails and recreation features, but the existing model of defining iD presets won't really work here.

If this new feature requires a framework beyond what the NSI currently utilizes, it would likely have to be independently developed by one or more volunteer contributors. It's unlikely that any new project features on a large scale will be built in-house anytime soon (see https://github.com/osmlab/name-suggestion-index/discussions/10757#discussioncomment-12567659).

bhousel:
I think #4768 and #9211 cover these cases of adding something that looks like a highway or a route to this index, so I'm going to close this as a duplicate.  The trails themselves are identified by `wikidata`, and the `operator` is an added detail, not the identifying tag.

Because the accepted tagging would be for a new tree it would require code changes in iD and a version bump in NSI, which for me means it probably won't happen.  I think we are stuck with NSI doing what it does for the foreseeable future.  PM me if you want to chat about that!

PR #10866, opened 2025-04-07
Add Appalachian Trail club operators

quincylvania:
This is a draft PR to help illustrate and open up discussion around #10865. The data format would need to be better defined, but this is a start. The data is for the 30 member clubs that maintain the Appalachian Trail ([relation/156553](https://www.openstreetmap.org/relation/156553)), all mapped in OSM and linked to Wikidata.

### DIFF
--- a/data/operators/trail.json
+++ b/data/operators/trail.json
@@ -1,0 +1,394 @@
+{
+  "items": [
+    {
+      "locationSet": {
+        "include": [
+          "us"
+        ]
+      },
+      "tags": {
+        "operator": "Green Mountain Club",
+        "operator:type": "private",
+        "operator:short": "GMC",
+        "operator:wikidata": "Q16994469"
+      }
+    },
+    {
+      "locationSet": {
+        "include": [
+          "us"
+        ]
+      },
+      "tags": {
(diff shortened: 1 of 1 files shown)


bhousel:
This might be ok if we can get consensus what features these operators would go on, and if `operator` really is the right tag for what these organizations do.

Do you see it as like:

A big route relation with 
```js
name = Appalachian Trail
type = route
route = ?
network = nhn?
wikidata = Q620648
operator = Appalachian Trail Conservancy
operator:wikidata = Q2858589
```

And that route holds a bunch of member routes like (the things in this PR):
```js
name = ?
type = route
route = ?
network = rhn?
wikidata = ?
operator = New York-New Jersey Trail Conference
operator:type = private
operator:short = NYNJTC
operator:wikidata = Q17083031
```

And then those member routes hold the different `highway=whatever` members?

There some pieces of this that NSI wouldn't be able to match without code changes (the superroute and the individual segments), and some pieces that it might be able to support today (the regional routes? idk).

I'm not sure that the `operator` is the thing that defines these features, so I'm handwaving a bit.

PR #10870, opened 2025-04-08
Add RNV  Transit Ticket vending machines in Germany

coolultra1:
More info: https://en.wikipedia.org/wiki/Rhein-Neckar-Verkehr - Large transit operator in the Rhine-Main area with surely hundreds of machines.

All machines are consistent, so payment tags are correct. Should be ready to merge.


Snowysauce:
I'm not personally a fan of including the ancillary tags suggested here (e.g., `currency:EUR`), but past precedent/the existing presets indicates that they're acceptable for this category.

PR #10863, opened 2025-04-07
Rename Sixt

zenfiric:
Resolves #10862

PR #10875, opened 2025-04-09
Add METRO GASTRO

ginnyTheCat:
Add [METRO GASTRO](https://www.metro.de/unternehmen/metro-gastro) (formerly "C+C Schaper") as a part of the METRO brand, as it's often being mistagged as `branch=Gastro`.

From my research it looks like these only exist in Germany, there, however, they make up a big chunk of the METRO lineup ([47 out of all 102 stores](https://www.metro.de/standorte)).

### DIFF
--- a/data/brands/shop/wholesale.json
+++ b/data/brands/shop/wholesale.json
@@ -240,6 +240,17 @@
       }
     },
     {
+      "displayName": "METRO GASTRO",
+      "id": "metrogastro-ad6aeb",
+      "locationSet": {"include": ["de"]},
+      "tags": {
+        "brand": "METRO",
+        "brand:wikidata": "Q13610282",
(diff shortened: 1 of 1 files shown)


Snowysauce:
The casing of the values of the name tag (and possibly also the brand tag) seem to contradict the style changes introduced in #10738 and proposed to be expanded upon in #10861 and #10863.

ginnyTheCat:
In this case I changed to to be consistent with how it's currency tagged and how METRO itself is represented in the NSI. I don't have anything against changing both METRO and METRO GASTRO to the new style changes.

Cj-Malone:
https://gist.github.com/Cj-Malone/23eddaafabb0ab7234e0cb420ffd35d9

PR #10876, opened 2025-04-10
[amenity/pharmacy] fixed name:ru variant for Pharmadepot pharmacy

a-ney:


panasonic-mafia (approved):
Better translation. LGTM. Thanks.

PR #10877, opened 2025-04-10
Make Fix Price available in LV

somehundred:
Fix Price is operating also in Latvia. Here is the website: https://lv.fix-price.com/

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -2942,6 +2942,7 @@
       },
       "tags": {
         "brand": "Fix Price",
+        "brand:wikidata": "Q4038791",
         "name": "Fix Price",
         "shop": "supermarket"
       }


Snowysauce:
The preset changed by this PR does not have a Wikidata tag, so at the moment it wouldn't be available anywhere.

Snowysauce:
Thanks. Looking at the Wikidata page, we have already two presets for this brand under `shop=variety_store` that cover Georgia, Belarus, and Russia. Would that be a more accurate tag for Latvia, or is `shop=supermarket` a better fit for their Latvian operations?

panasonic-mafia:
IMO it probably should be merged into shop=variety store. OSM currently has 18 POIs for Fix Price in `shop=supermarket`: https://overpass-turbo.eu/s/22hA. Seems like supermarket entry was added by NSI collector in 
6e656c7.

bhousel:
Sounds like there is some followup work to do, but this PR is fine.

PR #10878, opened 2025-04-10
Add RD Electronics

andris86:
RD Electronics is an electronics store chain operating in Latvia. The company itself is called MK Trade.

### DIFF
--- a/data/brands/shop/electronics.json
+++ b/data/brands/shop/electronics.json
@@ -1409,6 +1409,16 @@
       }
     },
     {
+      "displayName": "RD Electronics",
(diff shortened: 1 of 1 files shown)


Snowysauce:
The specified Wikidata item does not have any websites listed, and as such the prominence of the brand (i.e., number of locations) cannot be verified.

Snowysauce:
Thanks. I think I found the list of locations at https://www.rdveikals.lv/site/content/lv/166/Kontakti.html, and based on that the brand meets our inclusion criteria.

PR #10881, opened 2025-04-11
Add One NZ, A New Zealand cell phone company

catgirlseraid:
Vodafone sold their NZ division off in 2023 and then they rebranded to One NZ. I'm not sure if i should also include vodafone in the match names as it is the name many people still know them by, please leave a comment if not. Thanks <3

### DIFF
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -13,7 +13,7 @@
       "displayName": "One NZ",
       "id": "",
       "locationSet": {"include": ["nz"]},
-      "matchNames": ["one nz, vodafone"],
+      "matchNames": ["one nz", "vodafone"],
       "preserveTags": ["^name"],
       "tags": {
         "brand": "One NZ",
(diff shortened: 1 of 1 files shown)


Snowysauce:
> I'm not sure if i should also include vodafone in the match names as it is the name many people still know them by

I'd recommend doing so if Vodafone has completely ceased operations in New Zealand and has been completely replaced by One NZ.

I have a couple of tagging questions:

- `"matchNames": ["one nz, vodafone"],` - is this intended to match "one nz, vodafone" literally, or the separate terms "one nz" and "vodafone"? If it's the latter, the array needs to be formatted as `["one nz", "vodafone"]`.
- `"preserveTags": ["^name"],` - does each store have a unique name? If not, this tag should be removed.

catgirlseraid:
Thanks for the feedback! <3 I've updated match names as it was intented. Should `"preserveTags": ["^name"],` be kept if some stores have their location as part of the name, for example `One NZ Newmarket Westfield`?

Snowysauce:
> Should `"preserveTags": ["^name"],` be kept if some stores have their location as part of the name, for example `One NZ Newmarket Westfield`?

Assuming all of the store names follow the format "One NZ (descriptor)", I would tag them in OSM as `name=One NZ`, `branch=(descriptor)`, and omit `"preserveTags": ["^name"],` from the NSI preset.

catgirlseraid:
I've removed the `"preserveTags": ["^name"],` tag, thanks!

Issue #10639, opened 2025-03-01
Add brand Česká podnikatelská pojišťovna


Oivo35:
Add Česká podnikatelská pojišťovna insurance brand with more than 100 locations in Czech Republic:
https://www.cpp.cz/pobocky

    {
      "displayName": "Česká podnikatelská pojišťovna",
      "locationSet": {"include": ["cz"]},
      "tags": {
        "brand": "Česká podnikatelská pojišťovna",
        "brand:short": "ČPP",
        "brand:wikidata": "Q14949324",
        "name": "Česká podnikatelská pojišťovna",
        "office": "insurance"
        "short_name": "ČPP"
      }
    },

Snowysauce:
`short_name` was omitted from the added preset since `brands/office/insurance` has `"preserveTags": ["^name"]` applied to the entire category, and the `short_name` value may not always be applicable if the `name` value is modified in OSM. `brand:short` is a better way to convey the desired information in this case, and thus it was included in the preset as originally proposed even though it's an uncommon tag for the NSI.

Oivo35:
Thank you for clarifying.

PR #10885, opened 2025-04-11
Adding/updated hostels associated with Hostelling International

ShinyDog2636:
Adding/updating the hostelling organizations in England, Wales, Scotland, Canada, USA, India, and Australia that are associated with Hostelling International

This is an incomplete list but captures many of the largest national organizations.


Snowysauce:
`brand:short` is preferred over `short_name` for presets in categories with `"preserveTags": ["^name"],` as `short_name` is not guaranteed to be a short version of the `name` value. See https://github.com/osmlab/name-suggestion-index/issues/10639#issuecomment-2692550080.

PR #10898, opened 2025-04-19
Update ice_cream.json

NFarrasP:
Add Mixue, an ice cream chain in East and Southeast Asia. I added different "name" since that is more popular (at least in Indonesia) than the long brand name. I don't know if that's allowed.


kjonosm:
Please see #10675

Snowysauce:
Closing as a duplicate of #10675. That pull request needs localization research before it can be merged; you may be able to help.

Issue #10902, opened 2025-04-20
Duplicate brand objects in Turkey


burder5334:
Bim brand only supermarket in Turkey.

{
      "displayName": "BİM",
      "id": "bim-8af1d3",
      "locationSet": {"include": ["ma", "tr"]},
      "tags": {
        "brand": "BİM",
        "brand:wikidata": "Q1022075",
        "name": "BİM",
        "shop": "supermarket"
      }
    }

But have a duplicate object in

{
      "displayName": "BİM",
      "id": "bim-6cb8de",
      "locationSet": {
        "include": ["eg", "ma", "tr"]
      },
      "tags": {
        "brand": "BİM",
        "brand:wikidata": "Q1022075",
        "name": "BİM",
        "shop": "convenience"
      }
    }

bhousel:
One is `shop=convenience` the other is `shop=supermarket` , so that’s not really a duplicate.  Am I correct understanding that you want the convenience entry deleted?

PR #10903, opened 2025-04-21
Update convenience.json

ddpasa:
BİM in Turkey is a supermarket

fixes [10902](https://github.com/osmlab/name-suggestion-index/issues/10902)


bhousel:
Thanks @ddpasa

PR #10904, opened 2025-04-21
Add Sézane

Crashillo:
French fashion brand

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -9686,7 +9686,7 @@
     },
     {
       "displayName": "Sézane",
-      "locationSet": {"include": ["001"]},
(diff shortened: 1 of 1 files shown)


Snowysauce:
I checked the website listed for the brand on Wikidata (which redirected me to their US site based on location), and https://www.sezane.com/us/our-locations only lists locations in the US and Europe. If this truly is the extent of where they have locations, the proposed `locationSet` of "001" (global) is far too broad.

Crashillo:
I included the global set, since the current ones are fr, us, uk, de, ch and es... is a fast growing company, and some markets usually come together, as the EU market, to avoid future editions... nevermind, i'll include the subset

PR #10906, opened 2025-04-22
adding several East Texas chains

UltimateRiff:
my first actual PR, lemme know if I missed anything

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -147,7 +147,6 @@
         "brand:wikidata": "Q114572970",
         "brand:website": "https://www.origin.bank",
         "name": "Origin Bank"
-        "brand:website": "
       }
     },
     {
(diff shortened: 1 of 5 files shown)


Snowysauce:
I haven't had a chance to look at the specifics of the brands yet, so these are just some general notes:

- the `brand:website` tags will be automatically removed by the build script once this PR is merged into the NSI, so be sure that the websites are stored on the Wikidata items
- regarding "Kim's Convenience Store", the suggested scope (country-wide) makes the suggested `matchNames` value ("Kim's") problematic. This will cause iD to suggest this preset for any American convenience store with a name starting with "Kim's", even if the remainder of the name makes it clear that it's not part of the chain

UltimateRiff:
actually, I think Kim's might be Texas-only at the moment, I'll have to double-check on that tho... either way, their signs only says Kim's, but their website says Kim's Convenience Store, which is why I added that as a match name

UltimateRiff:
aight, fixed~

PR #10909, opened 2025-04-23
Update hospital.json - NHS trusts are operator:type=public, not private

rskedgell:
NHS trusts in the UK are publicly owned, so should be operator:type=public, not private. The wiki seems fairly unambiguous on this.

https://wiki.openstreetmap.org/wiki/Key:operator:type

This has resulted in mis-tagging by inexperienced users who unquestioningly accept iD's suggestions for tag upgrades.


bhousel:
Thanks @rskedgell 👍

PR #10920, opened 2025-04-29
[operators/man_made/surveillance] add France to main McDonalds entry

panasonic-mafia:
rel to: https://github.com/osmlab/name-suggestion-index/commit/10af0856a3ee6dcdde63d035a8adb1d18fe41c2d

[Q60766811](https://www.wikidata.org/w/index.php?title=Q60766811) has zero usages in OSM

- https://taginfo.geofabrik.de/europe:france/search?q=brand%3Awikidata%3DQ60766811
- https://taginfo.geofabrik.de/europe:france/search?q=operator%3Awikidata%3DQ60766811

seems France uses Q38076 for both brands and operators:

- https://taginfo.geofabrik.de/europe:france/search?q=operator%3Awikidata%3DQ38076
- https://taginfo.geofabrik.de/europe:france/search?q=brand%3Awikidata%3DQ38076

### DIFF
--- a/data/operators/man_made/surveillance.json
+++ b/data/operators/man_made/surveillance.json
@@ -3914,7 +3914,6 @@
           "by",
           "cn",
           "eg",
-          "fr",
           "iq",
           "jo",
           "jp",
@@ -3933,17 +3932,6 @@
           "tw"
         ]
       },
-      "tags": {
-        "man_made": "surveillance",
-        "operator": "McDonald's",
-        "operator:wikidata": "Q38076"
-      }
-    },
-    {
-      "displayName": "McDonald's (France)",
-      "id": "mcdonalds-e10d05",
-      "locationSet": {"include": ["fr"]},
-      "matchNames": ["mac donald"],
       "tags": {
         "man_made": "surveillance",
         "operator": "McDonald's",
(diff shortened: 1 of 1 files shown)


panasonic-mafia:
@Snowysauce was there are specific reason for Q60766811 in this entry?

Snowysauce:
This stems from France being excluded from the main McDonald's surveillance preset for some reason (a copy-paste artifact from the brands tree?), and the latest OSM data import finding surveillance cameras at McDonald's in France (erroneously tagged as "Mac Donald"). In case there was a specific reason to exclude France from the main preset, I created the French-specific surveillance preset and chose Q60766811 as the most relevant QID.

If the French exclusion from the main preset was nothing more than a copy-paste error, I have no objections to the deletion of the French-specific surveillance preset as long as the French exclusion to the main preset is also deleted.

Snowysauce:
I should probably clarify that I do have an objection to the PR as proposed; by changing the QID, the preset becomes identical to the main preset, at which point the two presets should be merged instead. My earlier comments still stand as well.

panasonic-mafia:
Thanks for your comment! Seem a copy-paste thing to me. I'm pretty convinced to put France to main preset for survellance.

panasonic-mafia:
Should be good to go.

LaoshuBaby (approved):

PR #10924, opened 2025-04-30
[amenity/hospital] add legal entity type to Helios Kliniken

git-tgo:
Add extension "GmbH" to operator, see https://www.helios-gesundheit.de/rechtliches/impressum


Snowysauce:
Re this and #10923: legal entity suffixes are typically omitted from operator names in the NSI.

PR #10931, opened 2025-05-04
[office/telecommunication] Add Telkom Indonesia

NFarrasP:
Please add Telkom Indonesia, a state-owned telecommunication company in Indonesia. It is the parent company of Telkomsel, which sometimes share office with Telkom Indonesia, but not always.

### DIFF
--- a/data/brands/office/telecommunication.json
+++ b/data/brands/office/telecommunication.json
@@ -272,6 +272,17 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


Snowysauce:
Is this preset intended for offices or retail stores?

NFarrasP:
> Is this preset intended for offices or retail stores?

Offices.

PR #10932, opened 2025-05-04
Add BreadLife bakery chain

NFarrasP:
Please add BreadLife bakery chain from Indonesia

### DIFF
--- a/data/brands/shop/bakery.json
+++ b/data/brands/shop/bakery.json
@@ -526,6 +526,17 @@
       }
     },
     {
+      "displayName": "BreadLife",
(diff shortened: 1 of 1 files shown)


Snowysauce:
How many locations does the chain have? The website listed on the Indonesian Wikipedia article is dead and only one location is mapped on OSM.

NFarrasP:
After further research, it only has around 30 stores. Might be too early to request naming.

Snowysauce:
30 locations is fine - see https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#notability

PR #10076, opened 2024-10-25
Poczta Polska is branded, no need for operator tag

matkoniecz:
fixes #10028

Poland has no USPS/USPO dychotomy

PR #10942, opened 2025-05-08
Change Canada Post post_office to brand, not operator

jarek:
Similar to #7100 and #10076: at least some Canada Post outlets are not operated by Canada Post (possibly most), however they are prominently branded as Canada Post.

See https://community.openstreetmap.org/t/canada-post-outlets-not-operated-by-canada-post/129748 for discussion and a picture. See also https://wiki.openstreetmap.org/wiki/Canada_Post for a project where many post offices were added or amended.

Draft PR while I wait for further comments on the forum.

Sorry for making this two commits, the repo is huge to clone and I couldn't figure out how to edit two files in one commit in the Github web interface. Please squash when merging.

### DIFF
--- a/data/brands/amenity/post_office.json
+++ b/data/brands/amenity/post_office.json
@@ -38,6 +38,20 @@
       }
     },
     {
+      "displayName": "Canada Post",
+      "id": "canadapost-17c815",
+      "locationSet": {
+        "include": ["ca"],
+        "exclude": ["ca-qc.geojson"]
+      },
+      "matchNames": ["canada post", "canada post/ poste canada"],
+      "tags": {
+        "amenity": "post_office",
+        "brand": "Canada Post",
+        "brand:wikidata": "Q1032001"
+      }
+    },
+    {
       "displayName": "City Express (Србија)",
       "id": "cityexpress-59a7c7",
       "locationSet": {"include": ["rs"]},
@@ -330,6 +344,23 @@
         "amenity": "post_office",
         "brand": "Posten",
         "brand:wikidata": "Q1815701"
+      }
+    },
+    {
+      "displayName": "Postes Canada",
+      "id": "postescanada-525cac",
+      "locationSet": {
+        "include": ["ca-qc.geojson"]
+      },
+      "matchNames": [
+        "canada post/ poste canada",
+        "poste canada",
+        "postes canada"
+      ],
+      "tags": {
+        "amenity": "post_office",
+        "brand": "Postes Canada",
+        "brand:wikidata": "Q1032001"
       }
     },
     {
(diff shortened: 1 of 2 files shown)


data/brands/amenity/post_office.json
```diff
@@ -332,6 +346,23 @@
...
+    {
+      "displayName": "Postes Canada",
+      "id": "postescanada-525cac",
+      "locationSet": {
+        "include": ["ca-qc.geojson"]
+      },
+      "matchNames": [
+        "canada post/ poste canada",
+        "poste canada",
+        "postes canada"
```
jarek, line 358: proper spelling of the brand in French is Postes Canada, however there may be some cases misspelled as "Poste Canada"

jarek:
As a note about languages. Branding in both English and French is used throughout Canada, however in Quebec the French is on the left, and in rest of Canada the English is on the left:
* https://commons.wikimedia.org/wiki/File:Canadian_Post_Office.jpg
* https://commons.wikimedia.org/wiki/File:800_Rene-Levesque_Ouest_-_04.jpg

I was curious which would be used in New Brunswick, which is officially bilingual, and through spot checks on a Major Commercial Map Provider With User-Submitted Photos And Streetside Imagery found that English is on the left, including in heavily French-speaking areas like Bathurst, Campbellton, and Edmundston.

This PR applies the English brand "Canada Post" to `brand` tag in all areas except Quebec (as was already the case with `operator` tags). The Wikidata is the same, of course.

data/brands/amenity/post_office.json
1ec5, line 1: Just double-checking: most of the entries in this file also set a `name=*` tag. Should these entries have `name=*` set to the same value as `brand=*` as a placeholder? `preserveTags` already includes `name`, so the user is free to manually specify a branch-specific name if they know it.

After we moved post offices from the brands tree to the operators tree in 0b91e23310bc8ddeda7c121c30d49e71f5b0d448 for #4638, we deleted the entries’ `name=*` tags in 2cc5cb939b1fb60b6a29b0e930c7c713d8edc59b for #4783, presumably because there was a consensus that `operator=*` usually shouldn’t influence `name=*` on its own. However, when post office brands were moved to the operator tree in #5613, these `name=*` tags were not restored. I think this accounts for most of the entries that don’t currently have a `name=*` tag.

jarek:
[As I summarized in the forum thread](https://community.openstreetmap.org/t/canada-post-outlets-not-operated-by-canada-post/129748/25), after 7 days there was general agreement that `operator` is incorrect and `brand` is correct, so I'm marking the PR ready for review.

data/brands/amenity/post_office.json
Cj-Malone: I'd leave name off for now, like GB. OSM seems it be "xx Post Office".

1ec5 (approved):
Thanks!

PR #10947, opened 2025-05-09
Added new countries for Mix Markt

alnzrv:


bhousel:
> Should I attach an examples from Google Maps to make the verification easier?

Nah, I trust you.. I'm gonna just merge this as-is.  
If it turns out to be wrong, someone else will come along and change it back.

PR #10955, opened 2025-05-14
Update alcohol.json

kamasytrix:
Poltavpivo add


Snowysauce:
How many retail locations does this brand have?

kamasytrix:
more than 40 branded stores in the Poltava region, and plans to expand to other regions of Ukraine

kamasytrix:
IMG_1747
IMG_1749
IMG_1750

kamasytrix:
Poltava is the name of the city where this beer works, only names are used in logos, the brand Poltava beer

PR #10956, opened 2025-05-14
Update charging_station.json

kamasytrix:
autoenterprise update name EVA Chargers

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -505,7 +505,7 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


Snowysauce:
This change seems OK on the surface; however, I looked at the website of EVA Chargers, and I came away with the impression that their specialty is the production of EV chargers. Do they actually operate EV chargers, or is the company solely a manufacturer of the stations?

kamasytrix:
IMG_1746
IMG_1745
one of the largest brands of charging stations in Ukraine

PR #10957, opened 2025-05-14
Regions in public transport networks of Spain

swaree:
Also expanded `transportpublicgeneralitatvalenciana-cbced8` and `metrobus-a154e2`.


Snowysauce:
https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#locationset

> A current list of supported codes can be found at https://ideditor.codes/

swaree:
Hi, I think that was all, now it's ready for review. Sorry for the amount of commits, please squash when merging as with the other one.

PR #10958, opened 2025-05-14
Spain features (provinces and autonomous communities)

swaree:
This uses simplified versions of datafiles provided by the © Instituto Geográfico Nacional, with permission for use in OSM.


bhousel:
Hey @swaree thanks for this work!

I took a quick look and it does seem like:
* Although they were simplified, the files are still pretty large, particularly including unnecessary coastal detail
* There are incongruities and seam issues between the boundaries (see below).  I don't know how much this matters really, but we have tried to avoid it in the past when constructing sub-country boundaries.  Checkout @jdcarls2 awesome tutorial on https://github.com/osmlab/name-suggestion-index/pull/5449 for what he did.

Screenshot 2025-05-14 at 4 51 48 PM

Ideally I think we should fix this stuff before merging..

swaree:
Regarding the weird overlap, it's because the provinces and autonomous communities (CCAA) come from different datasets. I added CCAA with only one province from the latter dataset, so if you choose ES-AL and ES-MU, it's a clear cut. Maybe it's best to generate CCAA as a merge of provices to avoid disparity.

Thanks for the insight, I'll try to tackle the issues.

swaree:
Hi,

I wasn't able to reduce the filesize much but now the coastlines follow roughly the shape and the CCAA are merged versions of the provinces.

Snowysauce:
Quick thoughts after looking at the suggested features in QGIS:

As suggested by @bhousel, I would take a good look at nodes in areas where the borders run through areas with little or no points of interest. Here's an example of a stretch of border nodes where everything between the first and last selected nodes can be removed with no impact to the features' geofencing scope as far as the NSI is concerned:

Screenshot 2025-05-15 063303

The proposal also currently includes redundant features for autonomous communities that are coterminous with a single province. The redundant features should be eliminated, preferably with the provincial feature being the one that remains. The suggested changes to the readme file already note cases where an autonomous community has the same boundaries as a province.

Crashillo:
One question, according the docs it's possible to use the wikidata:

> Strings recognized by the [country-coder library](https://github.com/rapideditor/country-coder#readme).
> These include [ISO 3166-1 2 or 3 letter country codes](https://en.wikipedia.org/wiki/List_of_countries_by_United_Nations_geoscheme), [UN M.49 numeric codes](https://en.wikipedia.org/wiki/UN_M49), and supported Wikidata QIDs.
> Examples: "de", "001", "conus", "gb-sct", "Q620634"
> 👉 A current list of supported codes can be found at https://ideditor.codes/

This property (P3896), called `geoshape` (ex. https://www.wikidata.org/wiki/Q5772#P3896) might be used as boundary, right? and the same for other sub-regions

bhousel:
> I wasn't able to reduce the filesize much but now the coastlines follow roughly the shape and the CCAA are merged versions of the provinces.

Thanks!  I checked a few of them and they look much better now 👍

bhousel:
> This property (P3896), called `geoshape` (ex. https://www.wikidata.org/wiki/Q5772#P3896) might be used as boundary, right? and the same for other sub-regions

That is cool that this exists, and I didn't know about it! 
That said - what we really need for shapes is something a lot simpler than this, as these will get fetched by Rapid / iD at start up and put into an index.   If we tried building the geometry index on-the-fly by fetching the geoshapes from Wikidata, it would take Rapid a while to start up.

Snowysauce:
Another thought I had regarding the autonomous communities: are there any cases where the borders don't fall along provincial boundaries? If there aren't, we probably don't need features for them - this PR would complete the set of Spanish provincial features, and the autonomous communities can then be assembled in location sets if needed by listing the member provinces. (Again, this only holds true if the borders of autonomous communities run along provincial boundaries.)

swaree:
Hi, any comments on this last version? I also modified the other pull request to account for autonomous communities not having a separate geojson file. Thanks!

Snowysauce:
The current version of this PR seems fine to me as a good base for future refining. There are some mismatches with country-coder's boundaries for France and Andorra, but at this point those adjustments are probably easiest done post-merge.

Issue #10949, opened 2025-05-11
Inaccurate generated Overpass Turbo query


Vectorial1024:
While investigating the tagging for a Hong Kong supermarket brand "Market Place by Jason's", I decided to lookup an NSI entry: https://nsi.guide/?t=brands&k=shop&v=supermarket#marketplace-745f9c

To determine whether the NSI entry is in use, I clicked on "Search Overpass Turbo", which generated the following query:

```
[out:json][timeout:100];
(nwr["name"="Market Place"];);
out center;

{{style:
node,
way,
relation
{ color:red; fill-color:red; }
node[shop=supermarket],
way[shop=supermarket],
relation[shop=supermarket]
{ color:yellow; fill-color:yellow; }
node[shop=supermarket][brand=Market Place][brand:wikidata=Q10851178],
way[shop=supermarket][brand=Market Place][brand:wikidata=Q10851178],
relation[shop=supermarket][brand=Market Place][brand:wikidata=Q10851178]
{ color:green; fill-color:green; }
}}
```

The above query is actually a generic query trying to look for anything named "Market Place" which returns too many results from all over the world, but recall that we are looking for a supermarket brand in Hong Kong with `shop=supermarket`. This `shop` tag is not reflected in the initial query.

------

The generated query can be fixed by the following:

```
[out:json][timeout:100];
(nwr["name"="Market Place"]["shop"="supermarket"];);
out center;
...
```

However, is this a valid problem? Is it intended for the NSI lookup to be "specific" enough to pinpoint exact matches, or should it be "general" enough to (in this case) include possible false positives?

Snowysauce:
https://github.com/osmlab/name-suggestion-index/issues/4805#issuecomment-750372928

> [The OPT links are] more for doing research on an unknown brand, less for seeing where a known brand is used.

This comment is over four years old, but it's still accurate. In any event, I wouldn't expect to see any major change to the queries in the near future - the project's lead developer is pressed for time due to IRL developments, and code contributions from other sources are basically nonexistent.

Cj-Malone:
The way I see it, this isn't even a developer time issue. Just a miss understanding and a generic name.

The query is meant to find all items matching the name, and colour code them category and brand match. And it does that. The generic name of "Market Place" makes using that specific query kinda useless, using it with "Market Place by Jason's" gives you a more accurate view on how it usually works. You can quickly confirm the location set is accurate, and OSM is matching the NSI tagging reasonably well.

[...]

PR #10967, opened 2025-05-19
Merge Market Place by Jasons into Market Place

Cj-Malone:
Rel: https://github.com/osmlab/name-suggestion-index/issues/10949#issuecomment-2869954634

According to [Wikipedia](https://en.wikipedia.org/wiki/Market_Place_(supermarket)), this is clear that "Market Place by Jasons" was rebranded to "Market Place". I imagine the Wikidata entry will be merged at some point too, I'm not sure which ID they'll keep.

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -5040,25 +5040,13 @@
       "displayName": "Market Place",
       "id": "marketplace-745f9c",
       "locationSet": {"include": ["hk"]},
+      "matchNames": ["market place by jasons"],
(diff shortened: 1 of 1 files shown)


ilias52730:
The correct Wikidata item is now technically [Q10851178](https://www.wikidata.org/wiki/Q10851178). It's been 3 months and the old one still hasn't been merged with the new one, so keeping the old one here would be wrong. If they merge them, they'll probably use the new, correct one. If not, we'll update it accordingly.

PR #10971, opened 2025-05-20
Updated 5 banks and added a bank in Morocco

ilias52730:
Updated the following Moroccan banks' display names, brand names or names following conventions in Morocco (`name` with French and Arabic, `name:fr` with the name in French and `name:ar` with the name in Arabic).

**Bank Al-Maghrib**: removed parenthesis from the display name.
**BMCE Bank**: this bank changed its name to Bank of Africa ([source](https://www.leconomiste.com/flash-infos/bmce-bank-africa-devient-officiellement-bank-africa)). Made sure to add `old_name`.
**Crédit Agricole du Maroc**: Updated the name. Was missing a word in Arabic.
**Société Générale**: Also updated the name. It was called _Société générale du Maroc_ on the NSI, but the reality is that they are simply called Société Générale.

I also added [**CFG Bank**](https://en.wikipedia.org/wiki/CFG_Bank), an important Moroccan bank that was missing.

This is my first pull request ever, hope everything is good. I'm trying to improve the NSI in Morocco, where I'm one of the main mappers. Feel free to ask any questions.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -12209,7 +12209,7 @@
       "matchNames": [
         "société générale سوسيتيه جنرال",
         "société générale maroc سوسيتيه جنرال",
-        "سوسيتيه جنرال"
+        "سوسيتيه جنرال",
         "SGMA"
       ],
       "tags": {
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -3937,7 +3937,7 @@
         "brand:ar": "بنك أفريقيا",
         "brand:fr": "Bank of Africa",
         "brand:wikidata": "Q2300433",
-        "name": "Bank of Africa بنك أفريقيا"
+        "name": "Bank of Africa بنك أفريقيا",
         "name:ar": "بنك أفريقيا",
         "name:fr": "Bank of Africa",
         "old_name": "Banque Marocaine du Commerce Extérieur البنك المغربي للتجارة الخارجية",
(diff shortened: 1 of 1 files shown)


Snowysauce:
I haven't had a chance to look at the specifics of the PR, but I have two quick notes.

1. `old_name` is [generally outside of our scope](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#adding-specific-keys)
2. "Bank of Africa" is an English phrase but is tagged as French ("name:fr"). I assume this is for consistency with local conventions and/or [local OSM wiki guidance](https://wiki.openstreetmap.org/wiki/Multilingual_names#Morocco)?

ilias52730:
Hi! Thanks for taking the time to take a quick look.
1. A lot of people in the country still call the bank "BMCE" when talking about Bank of Africa. The change has been made less than 5 years ago and a lot of people are still accustomed to calling it BMCE. That's why I though it would be useful. But if you think it should go into matchnames, I'm fine with that.
2. It is indeed an English phrase, so this one is tricky. It is called "Bank of Africa" even in French conversations. If you go to a branch, you'll hear "Bienvenue à Bank of Africa, que puis-je faire pour vous?". So you tell me what should be done in this case: name:fr because it's one of the languages used in the country or name:en because it's technically an English phrase.

I also just realized, after looking at other PRs as examples, that I made a little too many changes in a single PR. I thinks it's better this one be discarded and I'll push some with case by case changes so that it's easier to process.

PR #10974, opened 2025-05-21
Rebrand EB Games in Canada

Cj-Malone:
I'll leave as a draft/reminder for now, but it seems like GameStop is going to become EB Games again in CA.

https://www.gamestop.ca/promo/landing/ebgames-is-back


Cj-Malone:
#11222

PR #10977, opened 2025-05-22
Update Cityflyer (HK Bus)

Vectorial1024:
Color is added.

I cannot seem to run the build script locally. In case running the build script is required before merging, please do so on my behalf, thanks!

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -25085,6 +25085,7 @@
       "id": "cityflyer-5398e8",
       "locationSet": {"include": ["hk"]},
       "tags": {
+        "colour": "#cc0000",
         "network": "城巴機場快線 Cityflyer",
         "network:en": "Cityflyer",
         "network:wikidata": "Q5124066",


Snowysauce:
Network color is not something any of our bus presets currently have, and TBH I'm not sure if they should.

LaoshuBaby:
> I cannot seem to run the build script locally.

Don't need to do this when open a PR, because it will let diff contain many changes not related to the brand you changed.

> In case running the build script is required before merging, please do so on my behalf, thanks!

This is on [publish checklist,](https://github.com/osmlab/name-suggestion-index/wiki/Release-Checklist) don't worry.

LaoshuBaby:
> and TBH I'm not sure if they should.

https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case

This problem is the first time been talked in NSI, for my opinion I don't think this is needed because many city's public transport network use colour to distinguish different route/line. Yes, a specific operator may also have a representative colour but I guess it can't represent a network?

Vectorial1024:
re @Snowysauce , the situation is that I found some such Cityflyer relations have `colour=...` while some don't, so this seemed like a tagging that could be standardized.

bhousel:
>  I don't think this is needed because many city's public transport network use colour to distinguish different route/line.

Agree with @LaoshuBaby ... This project is tracks whole transit networks, and `colour` is usually different for each route within the network.  

After all the routes/stops/etc in your network are tagged with `network:wikidata`, you can find stuff easily with Overpass and assign other tags that you know are true across the entire network, if that's a thing you really want to do.

Vectorial1024:
The thing is, the actual network is Citybus, but Citybus itself deliberately selects a few routes to be under its specific brand "Cityflyer", which itself is not a network but has the characteristic reddish color.

The "no color for whole networks" principle does not seem to apply here because Cityflyer is not a network (at least this is how I understand it).

bhousel:
Merging anyway, if it's wrong, somebody will eventually tell us.

PR #10589, opened 2025-02-20
Add bicycle_repair_station.json with "RadKULTUR" entry

mcliquid:
I would like to add the “RadService-Punkt” bicycle repair stations of the “RadKULTUR” brand/initiative. These are limited to the federal state of Baden-Württemberg in Germany. There are currently over 600 stations and the number is constantly increasing due to an ongoing funding program. The stations all have the same properties, even if the manufacturer of the station alternates.

Since there was no file for bike repair stations yet, I'd be grateful for any feedback on this!

PR #10982, opened 2025-05-26
Adding new EV elements

GA-Kevin-Codes:
Added 2 new entries, both to the brand category:

- IONNA: A new EV charging network
- HYC400: Model of dispenser IONNA and others use

I have added as much information as I can to the IONNA and HYC400 Wikidata pages to make it useful to NSI. Thank you for your consideration.


Snowysauce:
The proposed change to `man_made/charge_point` is invalid (fails the workflow checks) for two key reasons: it lacks a `locationSet` and a `brand` tag. This entry is clearly aimed at a would-be `manufacturer` preset tree, something that was proposed in #10181 but is a long time away from becoming a reality, if ever. As such, the change to `man_made/charge_point` should be discarded entirely.

The proposed change to `amenity/charging_station` is in scope and valid, but there's a couple points with it that probably warrant further discussion. Some of the tag suggestions likely fall outside the scope of the NSI (see https://github.com/osmlab/name-suggestion-index/pull/10589#issuecomment-2766589567 and [our Judge Case page](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#adding-specific-keys)), and there are only 13 charge points in the entire US according to the Wikidata entry, which means [the brand probably doesn't meet our notability criteria at the moment](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#notability).

PR #10987, opened 2025-05-26
Update greengrocer.json : add Provenc'halles

raphodn:


Snowysauce:
How many locations does the chain have? I visited https://www.provenc-halles.fr/nos-magasins/, but failed to get the locator to produce any results.

PR #10988, opened 2025-05-26
Add 2 French fast foods

raphodn:


Snowysauce:
How many locations does DĒVOR have? https://www.devor.me/restaurants lists only 11 restaurants, but I have no idea if that's a complete list. (I was unable to get the site to translate into English in real-time, and the English-language site option at the top is a dead link.)

Mezzo di Pasta has enough locations to be included in the NSI per https://www.mezzodipasta.fr/restaurants.

Snowysauce:
> I never know if the threshold is 30 or 10 ?

This is probably the most authoritative answer that exists: https://github.com/osmlab/name-suggestion-index/pull/10227#issuecomment-2779167009. The threshold for inclusion is less a hard and fast number and more of asking "is the maintenance of this brand's/operator's POIs something that will be significantly assisted by inclusion in the NSI".

Pulling that concept back to this PR, if there are 11 locations but they're scattered all over France, NSI inclusion could help. However, if the 11 locations are concentrated in one region or city, it might be simpler to load them into overpass-turbo and JOSM and batch edit them that way. Just as an example: there is a restaurant/fast food chain in my area that has 14 locations, and has not only Wikidata but an article on the English Wikipedia. It's not in the NSI because all 14 locations are in the same metropolitan area and any brand-wide edits can be handled with a single batch edit.

> should i split my PR ?

I would say yes given the very low level of maintainer activity here in recent weeks. One half of this PR is good to go while the other does genuinely need feedback from other maintainers, and that feedback might not be coming for a while, unfortunately.

ilias52730:
According to [their website](https://www.devor.me/restaurants), Devor only has 11 restaurants, which is not enough.
Mezzo Di Pasta has [27 restaurants](https://www.mezzodipasta.fr/restaurants), which is close enough to 30.
I removed Devor and left Mezzo Di Pasta.
Merging PR.

PR #10990, opened 2025-05-26
Fix fees on indoor play

EvanCarroll:
The wiki documents two different types of indoor_play

* play that's free without a fee
* centers for indoor play that require a fee.

We have both on this list. The IKEA play centers do not require a fee, as far as I can see. The others clearly do. This is in line with the wiki proposal.


Snowysauce:
This seems like less of a "fix" and more like a proposal to add tags that I'm not sure the NSI should be involved with.

bhousel:
This seems unnecessary but i'm merging it anyway

PR #10995, opened 2025-05-28
 Special Civil Courts (JECs) in Brazil

mapaslivre:
Special Civil Courts (JECs) are bodies of the State Common Justice System, part of the Judiciary that operates in all Brazilian states.

History of the Special Civil Courts (JECs) https://pt.wikipedia.org/wiki/Juizado_Especial_C%C3%ADvel and their foundation.

### DIFF
--- a/data/operators/amenity/courthouse.json
+++ b/data/operators/amenity/courthouse.json
@@ -158,6 +158,16 @@
       }
     },
     {
+      "displayName": "Juizados Especiais Cíveis (JECs)",
+      "id": "poderjudiciário-3a5e8d",
+      "locationSet": {"include": ["br"]},
+      "tags": {
+        "amenity": "courthouse",
+        "operator": "Poder Judiciário",
+        "operator:wikidata": "Q10312802"
+      }
+    },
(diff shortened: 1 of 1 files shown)


mapaslivre:
@Snowysauce  In Brazil, the judiciary controls and coordinates these special courts. 

https://www.wikidata.org/wiki/Q105985 

Perhaps the judiciary is the same throughout the world, but in Brazil it is called the judiciary, which controls the Brazilian justice system.

mapaslivre:
Yes, the correct entry would be Q105985. 
Please correct the entry I made in “operator:wikidata”: “”Q10312802“ to 

”operator:wikidata": ‘’Q105985 ”
which is the correct entry.

Snowysauce:
I'm sorry, but I gave you a link to the Portuguese Wikipedia's article on the Brazilian judiciary and the ID of the Wikidata item attached to said article, and you're insisting that we use the Wikidata QID attached to another set of "explains the concept, but not the operator" articles. I'll let other maintainers deal with this.

PR #10996, opened 2025-05-29
Update insurance name to GRAWE

j-px:
The company's official website https://www.grawe.at ([here the latest archived version because somehow it's not really loading…](https://web.archive.org/web/20250510120138/https://www.grawe.at/)) only uses capitalized GRAWE.

Also, what to do with all the existing tags that use 'Grawe'? Will they be automatically changed after merge, or should I do that?

### DIFF
--- a/data/brands/office/insurance.json
+++ b/data/brands/office/insurance.json
@@ -1313,7 +1313,7 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


j-px:
Ah, thanks. Sorry I didn't notice.

PR #10998, opened 2025-05-30
Correct brand:wikidata reference for Michael Kors clothing brand

rjw62:
It seems that the original wd item has morphed into representing "Capri Holdings" the parent company of the Michael Kors brand. So I've created a new wd item for the brand itself. NSI should switch to that item.

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -7063,7 +7063,7 @@
       "locationSet": {"include": ["001"]},
       "tags": {
         "brand": "Michael Kors",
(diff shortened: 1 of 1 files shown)


Snowysauce:
This change looks OK to me. The edit history of Q19572998 is very strange; it looks like the change in focus from the brand to the holding company started as far back as 2021, if not earlier.

I moved a couple statements that explicitly belonged to the brand from Q19572998 to Q134612138, and there's probably still more work to be done in that department. Of course, that's a matter that's entirely independent of this PR.

PR #11004, opened 2025-05-30
[brands/office/government] rename Câmara de Vereadores to Câmara Municipal

mapaslivre:
city council, known in Brazil as the chamber of councilors.  https://www.wikidata.org/wiki/Q25550691

In Brazil, the city council is the legislative body of each municipality, acting as the assembly of representatives of the citizens residing there.  https://pt.wikipedia.org/wiki/C%C3%A2mara_municipal_(Brasil)

### DIFF
--- a/data/brands/office/government.json
+++ b/data/brands/office/government.json
@@ -37,12 +37,14 @@
       }
     },
     {
-      "displayName": "Câmara de Vereadores",
-      "id": "camaradevereadores-efb6ac",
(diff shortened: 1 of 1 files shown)


Snowysauce:
This seems like a generic term, and the chosen Wikidata QID underscores that impression.

mapaslivre:
the main name is city council and the alt name is council chamber

PR #11005, opened 2025-05-30
Update government.json

mapaslivre:
The Superior Electoral Court (TSE) is the highest legal authority of the Brazilian Electoral Justice System, with national jurisdiction.
 The other authorities are represented during elections by the Regional Electoral Courts (TRE), electoral judges, and electoral boards throughout Brazil. 
The TSE works in conjunction with the TREs, which are directly responsible for administering the electoral process in the states and municipalities.  https://pt.wikipedia.org/wiki/Tribunal_Superior_Eleitoral

https://www.wikidata.org/wiki/Q7309051

### DIFF
--- a/data/brands/office/government.json
+++ b/data/brands/office/government.json
@@ -76,6 +76,18 @@
         "office": "government"
       }
     },
+     {
+      "displayName": "Cartorio Eleitoral",
+      "id": "cartorioeleitoral-3a5e8d",
+      "locationSet": {"include": ["br"]},
+      "tags": {
+        "government": "register_office",
+        "office": "government",
+        "name": "Cartorio Eleitoral",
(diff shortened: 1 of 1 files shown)


Snowysauce:
The workflow tests failed because the proposed preset is in the `brands` tree but does not have a `brand` tag. The preset also has an `operator:wikidata` value without an `operator` tag, which doesn't generate errors but is nonetheless a nonstandard practice.

mapaslivre:
I made a new commit inserting the tag “brand”: “Cartorio Eleitoral” in the file data/brands/office/government.json.

bhousel:
This was kinda old and generating some conflicts, so I added it myself in 86c3ec241

PR #11012, opened 2025-06-01
Fix Idexe name and better specify clothes types

DnaX:
I corrected the name, "Idexè", not "Idexe" and I specified better the types of clothes sold.

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -5038,14 +5038,14 @@
       }
     },
     {
-      "displayName": "Idexe",
+      "displayName": "Idexè",
       "id": "idexe-d72191",
       "locationSet": {"include": ["it"]},
       "tags": {
-        "brand": "Idexe",
+        "brand": "Idexè",
         "brand:wikidata": "Q129995614",
-        "clothes": "children",
-        "name": "Idexe",
+        "clothes": "babies;children",
+        "name": "Idexè",
         "shop": "clothes"
       }
     },


Snowysauce:
I looked at the brand's website and their social media accounts, and they all used "Idexe'" (i.e., a trailing apostrophe instead of an accent over the last E). Is the brand name pronounced using the accent, or is the apostrophe purely stylistic?

DnaX:
Usually people don't know how to make uppercase "È" because it's not directly available on the Italian keyboard, so people (me as well with another laptop with US keyboard) uses trailing apostrophe instead of accent.

The point of my MR it's because the preminend tagging with "Idexè" already on OSM: https://overpass-turbo.eu/s/25jJ

Instead on Google Maps those stores are named IDEXE' (uppercase with trailing apostrophe).

Mannivu:
Company official website uses `IDEXE'` both in logo and headers, I think that should be used, but maybe we can add `Idexè` as an `alt_name`

AntiCompositeNumber:
> Usually people don't know how to make uppercase "È" because it's not directly available on the Italian keyboard, so people (me as well with another laptop with US keyboard) uses trailing apostrophe instead of accent.

That doesn't explain the logo though, where the mark is below and to the right of the second E. That looks a whole lot more than E' than É. The company seems to consistently use `IDEXE'` as Mannivu noted, and we'd usually de-caps that down to `Idexe'`.

It seems like this issue needs further discussion among the Italian OSM community to agree on a consistent spelling.

Mannivu:
> It seems like this issue needs further discussion among the Italian OSM community to agree on a consistent spelling

The discussion is underway in https://community.openstreetmap.org/t/quale-name-per-questo-brand/132392

DnaX:
After the discussion, we have chosen "IDEXE'" name.

Mannivu:
@DnaX wouldn't it be better lowercase (that is `Idexe'`), since it's not an acronym and all-caps it's just a marketing embellishment?

ilias52730:
> @DnaX wouldn't it be better lowercase (that is `Idexe'`), since it's not an acronym and all-caps it's just a marketing embellishment?

I also agree. Countless brands have uppercase logos/naming on their website but that's just a marketing technique.

PR #11013, opened 2025-06-02
Spain rail transport: New services and names

swaree:
- Cercanías Santander was renamed to Cantabria in 2019.
- Old FEVE services now operate under Renfe Cercanías AM
- Araba got a new commuter service.

Some of the changes I missed in my previous pull request. Thanks.

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -230,7 +230,7 @@
         "network": "Cercanías León",
         "network:wikidata": "Q107428854",
         "operator": "Renfe Cercanías AM",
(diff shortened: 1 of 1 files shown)


Snowysauce:
Cercanías Cartagena needs a `network:wikidata` tag to be exported as a preset.

swaree:
I see. It is a one-line network, and the line does have a Wikidata ID, with Cercanías Cartagena as an alias. Is that one appropiate?

Snowysauce:
That solution makes sense to me.

PR #11020, opened 2025-06-03
Add Ashford Communities

EvanCarroll:
Slum lord that owns 22 apartment complexes in Texas.

Nice website though... https://ashfordco.com/about-ashford-communities/


EvanCarroll:
I can break this up if needed but because they're _very_ similar additions it seems like it can all go together. Probably upwards of a million residences are covered in this.

data/operators/landuse/residential.json
```diff
@@ -4,10 +4,78 @@
...
+      "matchNames": [
+        "avalon"
+      ],
+      "tags": {
+        "landuse": "residential",
+        "operator": "AvalonBay",
+        "operator:wikidata": "Q4827537",
+        "residential": "apartments"
+      }
+    },
```
1ec5: Avalon is already listed under the brand tree. AvalonBay does own and operate the properties, but unlike some other REITs, they brand the properties quite heavily as an apartment chain. If you add AvalonBay to the existing entry as an `operator=*`, that would make it more consistent with the Modera entry. But listing it as only a brand is probably sufficient, since that’s the more apparent characteristic of each apartment complex in the chain.

https://github.com/osmlab/name-suggestion-index/blob/49a0fe1c7245dd030b394ee43e13f1c9866bc1d2/data/brands/landuse/residential.json#L8-L19

EvanCarroll: What is the guiding criteria for determining if it's a brand or not? I know it's either-or, but isn't it more useful to have the operator information for an apartment complex?

Not to mention consider these complexes are also Avalon,

* [AVA Pacific Beach](https://www.avaloncommunities.com/california/san-diego-apartments/ava-pacific-beach/)
* [eves North Dallas](https://www.avaloncommunities.com/texas/dallas-apartments/eaves-north-dallas/)
* [Kanso Twinbrook](https://www.avaloncommunities.com/maryland/rockville-apartments/kanso-twinbrook/)

There all under the same domain. They have the same operator. They have different brands.

1ec5: AvalonBay owns several chains (brands): Avalon, AVA, Eaves, and Kanso. It doesn’t matter at all that the websites are under the same domain name. A good analogy would be one of the big hotel companies: for example, Mariott owns multiple chains, including one named Mariott. The only difference is that AvalonBay owns and operates their apartments, but that isn’t particularly unusual among brands. Feel free to add AVA, Eves, and Kanso as brands too.

bhousel: > What is the guiding criteria for determining if it's a brand or not? I know it's either-or, but isn't it more useful to have the operator information for an apartment complex?

https://github.com/osmlab/name-suggestion-index/issues/5625#issuecomment-958218478 lists the guidance on how to decide between brand and operator.  I could see it being either, but I agree with @1ec5 that these these apartment names seem like brands, similar to what hotels are doing.    You can add `operator` tags alongside the brand tags if you want, that's allowed.

EvanCarroll: @bhousel @1ec5 see if that is more what you're looking for.

data/brands/landuse/residential.json
```diff
@@ -5,15 +5,48 @@
...
+        "brand:wikidata": "Q134701622",
+        "residential": "apartments"
+      }
+    },
     {
       "displayName": "Avalon",
       "id": "avalon-4e5c5e",
       "locationSet": {"include": ["us"]},
+      "matchNames": [
+        "avalon"
```
1ec5: The matcher already knows about this name from `name`, `brand`, and `displayName`, so we can safely omit it.

1ec5 (approved):
Looks good. I spot-checked [Ashford](https://www.bing.com/maps?&cp=30.373272~-97.712638&lvl=17&dir=130.91422&pi=-5.4008236&style=x&v=2&sV=2&form=S00027), [Camden](https://www.mapillary.com/app/?pKey=1959143534238686), and [MAA](https://www.mapillary.com/app/?pKey=476833686882409) and found that they post branded monument signs and/or flags at their leasing offices, so they do match the other entries in this file. Feel free to add the other Avalon brands you’ve mentioned either as part of this PR or in a separate one.

data/brands/landuse/residential.json
EvanCarroll: @1ec5  I assume it doesn't need it then for camden, or maa either? if the name is the same it's good? now that we're only using the brand, and not the operator they're all the same.

1ec5: Yes, you can nix that too. 👍

EvanCarroll: @1ec5 I also nixed the "name" field because none of these complexes are named only the brand. It doesn't make sense to have that there as well as a tag that's set. That has to be wrong. This is a deviation from hotels.

1ec5: You’re right that the correct `name=*` of each of these complexes should ideally include the individual location name (which can also go in `branch=*`). That said, having lived at two of these properties over the years, I’d say it’s not that uncommon for residents to use the brand name in isolation, in the same way you’d refer to an individual hotel by its brand. It’s not a big deal either way, since the user is probably pretty likely to have the full name on hand if they already know the brand.

EvanCarroll: @1ec5 cool cool, I'm done with the patch then -- it's ready to merge.

PR #11021, opened 2025-06-04
Add entries for Sépaq

wolfy1339:
Add 3 entries for Sépaq, a Quebec government agency that operates the provincial parks and nature reserves.

* Adds an entry into `leisure=nature_reserve`
* Add an entry for `boundary=national_park`
* Adds an entry for `boundary=protected_area`

In many cases, more than one preset would apply at the same time, as all national parks are also tagged as `leisure=nature_reserve`, and all nature reserves are also tagged `boundary=protected_area`.

Example Nature Reserve: https://www.openstreetmap.org/relation/8008866
```
boundary=protected_area
leisure=nature_reserve
name=Réserve faunique de Portneuf
operator=Sépaq
operator:wikidata=Q3488215
operator:type=government
```
Example National Park: https://www.openstreetmap.org/relation/15421717
```
boundary=national_park
leisure=nature_reserve
name=Parc national de la Jacques-Cartier
operator=Sépaq
operator:type=government
operator:wikidata=Q3488215
```

Would this cause any issues downstream in editors?

### DIFF
--- a/data/operators/boundary/national_park.json
+++ b/data/operators/boundary/national_park.json
@@ -257,6 +257,20 @@
       }
     },
     {
+      "displayName": "Sépaq",
+      "id": "societedesetablissementsdepleinairduquebec-77ca76",
+      "locationSet": {
+        "include": ["ca-qc.geojson"]
+      },
+      "tags": {
+        "boundary": "national_park",
+        "operator": "Société des établissements de plein air du Québec",
+        "operator:short": "Sépaq",
+        "operator:type": "government",
+        "operator:wikidata": "Q3488215"
+      }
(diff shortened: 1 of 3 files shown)


Snowysauce:
> In many cases, more than one preset would apply at the same time, as all national parks are also tagged as `leisure=nature_reserve`, and all nature reserves are also tagged `boundary=protected_area`.
> 
> Would this cause any issues downstream in editors?

I don't believe this would lead to conflicts as long as all the `operator`/`operator:*` tags and values are identical, but there are certain editors (most notably iD) where only one of the matching presets will be automatically presented.

PR #11022, opened 2025-06-04
Add other Avalon brands

EvanCarroll:


EvanCarroll:
@1ec5 as requested.

data/brands/landuse/residential.json
```diff
@@ -43,6 +56,33 @@
         "residential": "apartments"
       }
     },
+    {
+      "displayName": "eaves by Avalon",
+      "id": "eaves-4e5c5e",
+      "locationSet": {"include": ["us"]},
+      "tags": {
+        "brand": "eaves",
+        "brand:official": "eaves by Avalon",
```
1ec5, line 65: Hah, they really are going for hotel naming practices here…

1ec5 (approved):
Thanks!

data/brands/landuse/residential.json
1ec5: Oh, I jumped the gun: looks like `brand:official=*` isn’t in use at all. a4811375961c988d193d8631def1a67c848d259b moves the “by Avalon” into `brand=*` but leaves “eaves” as `brand:short=*`, which is a common key. Sorry for missing that.

PR #11028, opened 2025-06-08
Add a few more Coach countries

peternewman:
Given #10857 has got stuck for now, here are a few more Coach countries where I've personally seen a shop and modified a node in that country


peternewman:
Thanks @Snowysauce

Issue #11030, opened 2025-06-09
Operator or brand?  charging_stations


Rovastar:
I loved this project since it was introduced and a much needed part of OSM (I was always sceptical that the power that be in OSM would ever allow such an attempt at standardization and I gave up a decade ago trying to convince them otherwise)

Anyway it is my first time (i think) I am dabbling into it to change stuff.

I was looking to do some editing to eV charging stations. Got a bit confused as there are 2 sections they can occur 
/data/brands/amenity/charging_stations.json
/data/operators/amenity/charging_stations.json

Most of the time the operator is the brand. I can think of only a couple cases where it would not be and they are not in the lists from what I can see.

I presume you don't want this to be like this the duplicate entries for both. If there is a valid reason let me know.

I propose to merge these together probably under brand.

Is that something you would be interested in?
To be honest for any open source project like GitHub I don't want to enter spend a lot of my time doing stuff and it not go anywhere so I would rather be told now that you will not consider merging these.

bhousel:
hey @Rovastar thanks for the kind words!

We do have some guidance on previous tickets about when to choose `operator` vs when to choose `brand`:
https://github.com/osmlab/name-suggestion-index/issues/5625#issuecomment-958218478

Charging stations are one of those categories that does live in both places
https://nsi.guide/index.html?t=brands&k=amenity&v=charging_station
https://nsi.guide/index.html?t=operators&k=amenity&v=charging_station

I think our intent was that the "branded" stations should include things like Tesla , and "operated" stations should include situations where a business or municipality just operates a lot of public charging stations.   But reviewing the lists, it does look like today things are inconsistent. 

Based on that guidance, I think a many of the entries in the `operator` file should be moved over to the `brand` file.

We should discuss more before you start changing things - it sounds like you want to move everything over to the `operator` file, and that's kind of the opposite of where I think things belong.

Rovastar:
I don't mind what way you do it. In my original edit of this issue I did say I wanted it brand but then reading more posts it seemed that brand was there as an exception so I swapped it to operator as that where most of the entries were.

Either way it would benefit from one place. Happy for it to be in brand.

The questions I have for that is the vast majority of those in operator don't have a brand. 
Are you happy for the brand directory for this json to contain loads with no brand?
will it work if there are entries with no brand tags in the brand section?
Or do you want all the operators without a brand to have the added with brand the same as the operator?

[...]

PR #11036, opened 2025-06-11
Charging_station cleanup

Rovastar:
Following on from https://github.com/osmlab/name-suggestion-index/issues/11030 

I Merged about 50 charge station operators into brand where they had both tags

Also/alongside

added missing brand where it was an obvious known brand

applegreen Electric
E.On Drive (renamed too)
Elecktra
Electrify America
ev connect
evyve
fastned
geniepoint
gridserve
Mercedes benz
Osprey
podpoint
porsche
TotalEnergies
zest

added missing operator

Ionity
mer
Nio Power to non-chinese  nio power, nio power Destination, nio swap

updated countries
applegreen electric added us
be.ev updated to gb
bp pulse added us

excluded us, au, nz, gb from BP
(BP pulse is teh brand in these countries)

removed e.on danmark

### DIFF
--- a/data/brands/amenity/charging_station.json
+++ b/data/brands/amenity/charging_station.json
@@ -2,31 +2,18 @@
   "properties": {
     "path": "brands/amenity/charging_station",
     "skipCollection": true,
-    "preserveTags": [
-      "^name"
-    ],
+    "preserveTags": ["^name"],
     "exclude": {
-      "generic": [
-        "^charging station$"
-      ],
-      "named": [
-        "^tesla, inc.$"
-      ]
+      "generic": ["^charging station$"],
+      "named": ["^tesla, inc.$"]
     }
   },
   "items": [
     {
       "displayName": "7-Eleven",
-      "id": "7eleven-994ed4",
-      "locationSet": {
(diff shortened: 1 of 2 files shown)


Snowysauce:
Several of the moved presets lack `brand:wikidata` tags, meaning those presets would not be exported in NSI releases and would effectively disappear from the list of available presets.

Rovastar:
I have updated this now.

Please let me know if there are any other issues.

bhousel:
Looks generally ok to me. 
I didn't check every line item, but my thoughts are: - if a charging station _has_ a brand then it should really be moved from the operator file to the brand file, and that looks like what you did 👍 

(adding the https://github.com/osmlab/name-suggestion-index/issues/5625#issuecomment-958218478 brand vs operator guidance here, so the tickets all get linked and people can refer back to it)

Rovastar:
Yeah. I want to move more across but need to do some research on each one but it is a good start. Brand should where most new one are entered unless there are exceptions going forward.

I'll do another batch once this merge has gone through.

Rovastar:
thanks!

PR #11040, opened 2025-06-13
update scouting brand names

tiptoptom:
See the wiki data column at https://nsi.guide/index.html?t=brands&k=club&v=scout.


AntiCompositeNumber:
Unless you are planning to update the existing data yourself, or have already done so, this should probably include `matchNames` to allow existing tagging to be upgraded.

tiptoptom:
Good point! I added the old names in `matchNames`.

PR #11043, opened 2025-06-14
Update stationery.json

swaree:
Bureau Vallée is now present outside of France.

Reflected on their website:
image


Snowysauce:
Merging per the localized websites' store lists.

PR #11016, opened 2025-06-03
[shop/supermarket] add Mix Mateus

mapaslivre:
Mix Mateus
Retail chain with stores throughout Brazil.

https://www.grupomateus.com.br/

PR #11048, opened 2025-06-16
Update lottery.json

mapaslivre:
name duplication correction

### DIFF
--- a/data/brands/shop/lottery.json
+++ b/data/brands/shop/lottery.json
@@ -28,7 +28,7 @@
       "tags": {
         "brand": "Loterias Caixa",
         "brand:wikidata": "Q134621462",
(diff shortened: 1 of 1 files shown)


Snowysauce:
A query for "Loterias Caixa" returns [14 results](http://overpass-turbo.eu/s/26aA); a query for "Loterias Caixa Economica" returns [no results](http://overpass-turbo.eu/s/26aB). (If "name duplication correction" is referring to https://github.com/osmlab/name-suggestion-index/pull/11016#issuecomment-2950629240 and https://github.com/osmlab/name-suggestion-index/pull/11016#issuecomment-2952731126, that's a completely different issue caused by using outdated snapshots of the NSI for PRs.)

PR #11057, opened 2025-06-17
Removed duplicate "ma" from total-389b37, kept in totalenergies-ff2f90

ilias52730:
Morocco was present in both total-389b37 (which seems to be an entry with the old name (Total) that was kept on the NSI for Morocco, Algeria and Eritrea) and totalenergies-ff2f90.

I removed it from the former and kept it in the latter. Gas stations here are branded "TotalEnergies", just like the rest of the world. Have no idea about Algeria and Eritrea so I did not move them.


panasonic-mafia (approved):
LGTM, thanks!

PR #11067, opened 2025-06-17
Moved Orange from mobile_phone to telecommunication

ilias52730:


ilias52730:
Closing this. Will go through it case by case.

Issue #11025, opened 2025-06-06
Tamoil express coupled to incorrect Wikidata-entry and typo in name-tag


A67-A67:
The entries for Tamoil express are coupled to the general Tamoil Wikidata-entry [Q706793](https://www.wikidata.org/wiki/Q706793) instead of the entry for Tamoil express [Q124658477](https://www.wikidata.org/wiki/Q124658477).

Secondly, the name-tag capitalizes the 'express' part, while this isn't done in reality ([example](https://commons.wikimedia.org/wiki/File:Tamoil_Express_petrol_station,_Winschoten_(2019)_02.jpg)). Therefore NSI-created brand-tags are inconsistent with the existing brand-tags for Tamoil express.

The entries are:
- [tamoilexpress-6cdea6](https://nsi.guide/?t=brands&k=amenity&v=fuel#tamoilexpress-6cdea6)
- [tamoilexpress-ac4a89](https://nsi.guide/?t=brands&k=advertising&v=totem#tamoilexpress-ac4a89)

RobinLinde:
Considering the [Tamoil locator](https://tamoil.nl/tankstations) and [news messages](https://tamoil.nl/nieuws-en-acties) also very consistently capitalize it like "Tamoil Express" I think this could be considered the correct spelling. I have opened a PR (#11071) which corrects the Wikidata entry (among others)

PR #11071, opened 2025-06-18
Change Dutch parcel locker names, add two entries

RobinLinde:
Removes the "pakketautomaat" text from two entries, which can be seen as descriptive rather than being part of the brand. Also added entries for Fiets & Service and Decokay, plus changed the Tamoil Express entry to the separate Wikidata identifier (#11025).


RobinLinde:
Closed in favour of #11104 and #11105

Issue #11072, opened 2025-06-18
Japanese ATM brands


momiji-takinogawa:
cc: @LaoshuBaby 
I want to organize the popular ATM brands in Japan, located within stations, konbini stores, and so on.
The brands can be grouped by their operator into 3 categories: 
1. ATMs operated by non-bank financial companies.
   - E-net/イーネット
   - PatSat/パッとサッと
   - View Altte/ビューアルッテ (merged in #11059)
2. ATMs operated by the traditional banks. They have bank offices and usually run ATMs with their bank brand.
   - AEON Bank/イオン銀行
   - JP Bank/ゆうちょ銀行
   - Most of the traditional banks
3.  ATMs operated by neobanks that run online-exclusive banking services. They do not have physical counters for customer service.
    - Seven Bank/セブン銀行
    - Lawson Bank/ローソン銀行

For Category 1, we can simply add an entry in amenity/atm. 
For Category 2, we add an entry in amenity/bank, and then amenity/atm entry will be automatically generated.
For Category 3, I think we shouldn't add an entry for amenity/bank as there should be no POIs to be tagged as amenity=bank. Instead, we addan  NSI entry for amenity/atm. I found several mistagging for セブン銀行 (ATMs in 7-Eleven are tagged as banks) https://overpass-turbo.eu/s/26jb

LaoshuBaby:
> 3\. ATMs operated by neobanks

I'd suggest that don't record none-real bank are supported by whom, for example, I've found e-net ATM can both supported by MUFG or SMBC.

I try to sort this in a draft table, and I think we need to discuss traditional bank seperatedly.

<table border="1">
  <thead>
    <tr>
      <th rowspan="2">Category</th>
      <th rowspan="2">Example</th>
      <th colspan="4">Judgment Question</th>
      <th rowspan="2">Conclusion</th>
    </tr>
    <tr align="left">
      <th>① <br />Is it an 金融機関?</th>
      <th>②-A <br />Does it have its own bank branch offices?</th>
      <th>②-B <br />Is it rely on specific none-bank commercial establishments?</th>
      <th>③ <br />Does it have physical ATMs?</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td rowspan="3">Category I</td>
      <td>イーネット<br />(E-net)</td>
      <td rowspan="3" align="center">X</td>
      <td rowspan="3" align="center">X</td>
[...]

PR #11076, opened 2025-06-19
Change Seven Bank to amenity=atm

momiji-takinogawa:
セブン銀行 (Seven Bank) used to have customer service counters in several shopping malls in Japan, but today it only operates ATMs.

issue: #11072


LaoshuBaby:
> セブン銀行 (Seven Bank) used to have customer service counters in several shopping malls in Japan, but today it only operates ATMs.

Sure, you means itoyokado? They now only service nanaco related stuff and become a pure internet bank like rakuten bank. None-real estate bank can't been shown on OSM.

----

BTW, I just searched [itoyokado](https://nsi.guide/?t=brands&tt=itoyokado) and haven't found it in nsi, I'll add it later

PR #11077, opened 2025-06-19
Add Cotton On brands

TheUKHighStreet:


Snowysauce:
The various Cotton On brands may have excessively broad `locationSet`s - according to https://cottonon.com/AU/store-finder/ their presence in North America, South America, and Europe is limited to one country on each continent.

AntiCompositeNumber:
I'm also questioning the addition of `"operator": "Cotton On Group"`. Do we know, for sure, that all stores are corporate-owned and operated? Normally these cross-brand relationships would be stored on Wikidata only.

PR #11078, opened 2025-06-19
Removed "Épicier" - literally means "grocery store" - not a brand and no wikidata

ilias52730:
Don't know who added Épicier or Épicerie as brands of department stores, convenience stores and copyshops in Morocco. They're just the generic French word for grocery store and do not have wikidata.

Someone already mentioned it in [commit dfacf64](https://github.com/osmlab/name-suggestion-index/commit/dfacf64528c4ab5aee4d10058cbefb5e21429ffd).

### DIFF
--- a/data/brands/shop/department_store.json
+++ b/data/brands/shop/department_store.json
@@ -461,16 +461,6 @@
       }
     },
     {
-      "displayName": "Épicerie",
-      "id": "epicerie-e32af0",
-      "locationSet": {"include": ["ma"]},
-      "tags": {
-        "brand": "Épicerie",
-        "name": "Épicerie",
-        "shop": "department_store"
-      }
-    },
-    {
       "displayName": "Éxito",
       "id": "exito-96a609",
       "locationSet": {"include": ["co"]},


Cj-Malone:
They will be auto generated from OSM. They'll need to be added to the excludes at the head of the file, or they'll comeback. Ideally OSM would also be fixed.

Snowysauce:
An explanation of how the entries were created can be [found here](https://github.com/osmlab/name-suggestion-index/wiki/Contributing#generic-name-filtering). We try our best to filter out generic names after runs of the NSI Collector, but if none of the maintainers speak a certain language - in this case French and/or Arabic - it's completely possible that some generic names can elude detection for a long time.

As for the commit that removed this specific name from `shop/copyshop` once before, it's from an earlier version of the NSI (v5) and perhaps the word's generic status wasn't carried over when the project transitioned to v6. 

In any event, thanks for catching these entries. They probably need to be added to the category files' exclude Array to keep them from returning, but that can be done post-merge.

ilias52730:
Thanks a lot for the explanations, I actually had no idea how it works.

PR #11081, opened 2025-06-20
Add Game Nation

TheUKHighStreet:


ilias52730:
Extra detail: they have 39 branches according to their [official website](https://game-nation.co.uk/our-venues/).

TheUKHighStreet:
Sorry - do I need to do something? I've seen that you've changed the locationSet to gb-eng rather than gb. Thanks.

UKChris-osm:
> Sorry - do I need to do something? I've seen that you've changed the locationSet to gb-eng rather than gb. Thank.

Nothing to change, when checking I just noticed they only trade in England so thought it would make sense to change it.

PR #11082, opened 2025-06-20
Separated Banque Populaire entries

ilias52730:
Banque Populaire is the second largest Moroccan bank.

image

It has a subsidiary in Europe which operates in Spain, France, Belgium, the Netherlands, Italy and Germany, but as a separate entity with a completely different name: Chaabi Bank (see image below of Chaabi Bank in Paris).

image

On the NSI, both the Moroccan entity and the European entity were under one single entry. I separated them by updating the Moroccan entry (Banque Populaire) and adding the European entry (Chaabi Bank). They already have separate Wikidata entries.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -3330,22 +3330,35 @@
       }
     },
     {
-      "displayName": "Banque Populaire (البنك الشعبي)",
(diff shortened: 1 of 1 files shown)


ilias52730:
Phew, should be good now, I forgot a comma...

PR #11083, opened 2025-06-21
Fix Tenya QID

momiji-takinogawa:
Tenya entrie's QID was set to the former operating company, and the entry was incorrectly marked as dissolved.

This PR replaces the brand:wikidata with the QID for the brand.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -13087,7 +13087,7 @@
         "brand": "てんや",
         "brand:en": "Tenya",
         "brand:ja": "てんや",
-        "brand:wikidata": "Q11319830",
+        "brand:wikidata": "Q124282698",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
image
I click into the QID you provided and found the picture, more exactly meet with my daily memory, "What is Tenya? Sure enough, Tenya is indeed this Tendon Tenya"
Ah, It's this shop, I often found it near station, but I guess the brand and name should be written as "天丼てんや"?

Because this will fit what the sign said, and "on the ground" is a important principle in OSM. If you or other mapper agree with this change I'll commit it tomorrow. 

---
For reply for banks, I'll reply in Monday😭

momiji-takinogawa:
I think 天丼てんや is a better name for the chain. Could you take care of the change?

PR #11087, opened 2025-06-22
add New Zealand Territorial Authority districts

cloventt:
These locationSets represent the approximate boundaries of NZ territorial authorities. These are distinct from the approximate regions described by ISO 3166-2:NZ. My ultimate goal is to be able to, for example, recommend that tags be upgraded for substations within the Christchurch City territorial authority boundary be upgraded to reflect that the owner/operator is Orion New Zealand Limited.

This data was generated in the following way:

1. the boundaries were taken from the Christchurch City Council Open Spatial Data portal ([here](https://opendata-christchurchcity.hub.arcgis.com/datasets/cfd99593548a484db730a092a7097a49_0/explore)) (N.B. this dataset features an open license)
1. simply the polygons using the Shapely lib `simplify` method with a precision of 0.01 tolerance
1. convert to the format used in this dataset


cloventt:
Script used to automatically generate the output from the downloaded GeoJSON dataset. Some code adapted from [anishdhakal15/geojson-simplifyer](https://github.com/anishdhakal15/geojson-simplifyer).

```python
import json
from pathlib import Path
from shapely.geometry import LineString, Point, Polygon

if __name__ == '__main__':
    with open(Path('~/Territorial_Local_Authority_(OpenData).geojson').expanduser(), 'r') as f:
        loaded_json = json.load(f)
    for authority in loaded_json['features']:
        authority_name = authority['properties']['TerritorialAuthorityName'].title()
        authority_name_kebab = authority['properties']['TerritorialAuthorityName'].lower().replace(' ', '-')
        file_name = f'nz-{authority_name_kebab}.geojson'.replace('\'', '')
        authority['id'] = file_name
        authority['properties'] = {}

        # simplify geometry
        tolerance = 0.01  # precision of simplification

        simplified_polygons = []
        for polygon in authority['geometry']['coordinates']:
            simplified_exterior = LineString(polygon[0]).simplify(
                tolerance, preserve_topology=True)
            simplified_interiors = [LineString(interior).simplify(
                tolerance, preserve_topology=True) for interior in polygon[1:]]
            simplified_polygons.append(
                [simplified_exterior.coords[:]] + [i.coords[:] for i in simplified_interiors])
        authority['geometry']['coordinates'] = simplified_polygons

        with open(Path(f'~/development/name-suggestion-index/features/nz/{file_name}').expanduser(), 'w') as out_file:
            json.dump(authority, out_file, indent=2)

        print(f'- [{file_name}](https://location-conflation.com/?locationSet=%7B%22include%22%3A%5B%22{file_name}%22%5D%7D&referrer=nsi) - {authority_name}')


```

Snowysauce:
I'm sorry, but I'm not at all convinced that these features are necessary. This proposal would add 70 features for a country that already has a complete set of subdivision features, and when there's this many features involved, I wonder how much use each feature would actually see. Would it make more sense for whatever entries that absolutely require these boundaries to use a set of coordinate-based radial definitions?

The proposed 11% increase in the amount of features we have also affects the size of our releases, which in turn impacts the performance of consumer projects that use the features for geographic processing. It'd be one thing if we were adding features for countries sorely lacking them (like China or Russia), but this proposal doesn't really add any groundbreaking functionality.

PR #11090, opened 2025-06-22
Add initial military operators category (landuse=military)

wvanderp:
# Add military.json to define military operators and their attributes

I have always been interested in creating a map of all the military bases in the world, organized by country. For that, you need good data.

That is why I propose adding `landuse=military` to the NSI. This is just an initial step to get the ball rolling. I welcome discussion about the choices I've made, and I am sure there are some errors in this first attempt at creating this new category.

I've started with all the operators that have around 10 uses in Taginfo.

I also have some issues that need solutions, and I would like to hear your thoughts on them:

- Arma dei Carabinieri – also in police.json
- Gendarmerie nationale – also in police.json
- United States Army Corps of Engineers – also in protected_areas.json
- Guardia Costiera – also in police.json
- Guardia Civil – also in police.json

We can move them here or keep them in their current files. Additionally, I'm still debating whether the gendarmerie should be classified as a police force or a military force, as they serve in this capacity in some countries and others.

I hope this is something we want to add to the NSI.
Any feedback or suggestions for improvement are welcome.

### DIFF
--- a/data/operators/landuse/military.json
+++ b/data/operators/landuse/military.json
@@ -24,7 +24,7 @@
         "landuse": "military",
         "military_service": "army",
         "operator": "Australian Army",
-        "operator:wikidata": "Q122191"
+        "operator:wikidata": "Q781360"
       }
     },
     {
@@ -68,7 +68,7 @@
         "landuse": "military",
         "military_service": "intelligence",
         "operator": "Central Intelligence Agency",
-        "operator:wikidata": "Q206613"
+        "operator:wikidata": "Q37230"
       }
     },
     {
@@ -169,7 +169,7 @@
         "landuse": "military",
         "military_service": "gendarmerie",
         "operator": "Guardia di Finanza",
-        "operator:wikidata": "Q54853"
+        "operator:wikidata": "Q1552861"
       }
     },
     {
@@ -268,7 +268,7 @@
         "military_service": "navy",
         "operator": "海上自衛隊",
         "operator:en": "Japan Maritime Self-Defense Force",
-        "operator:wikidata": "Q830306"
+        "operator:wikidata": "Q731647"
       }
     },
     {
@@ -303,7 +303,7 @@
         "landuse": "military",
         "military_service": "intelligence",
         "operator": "National Security Agency",
-        "operator:wikidata": "Q180057"
+        "operator:wikidata": "Q121194"
       }
     },
     {
@@ -347,7 +347,7 @@
         "landuse": "military",
         "military_service": "air_force",
         "operator": "Royal Air Force",
-        "operator:wikidata": "Q222596"
+        "operator:wikidata": "Q165862"
       }
     },
     {
@@ -371,7 +371,7 @@
         "landuse": "military",
         "military_service": "navy",
         "operator": "Royal Australian Navy",
-        "operator:wikidata": "Q122190"
+        "operator:wikidata": "Q741691"
       }
     },
     {
@@ -433,7 +433,7 @@
         "landuse": "military",
         "military_service": "coast_guard",
         "operator": "United States Coast Guard",
-        "operator:wikidata": "Q11222"
+        "operator:wikidata": "Q11224"
       }
     },
     {


Cj-Malone:
I never really mapped military stuff, and when I did, the tagging confused me. Is `military` a subtag of `landuse=military`, or it's own top level tag?

Anyway, my first comment on this is that `military_service=intelligence` has 0 uses, and `military_service` doesn’t have loads in general. NSI should be using well established tags IMO. Quality + uncontroversial.

LaoshuBaby:
I don't think provide preset for military landuse is a good idea. If we approve this, I can see the future we will receive PR that want to add US Army, add PLA, add other countries real national armed force to a preset.

The first, mapping landuse is illegal in many countries, handle a project containing those information(especially if there have data related to their countries) may make those contributor in danger.

The second, those feature can do only less with improving general mapping quality, for military information is hard to comfirm and ensure, this violate a important "verifiable" osm principle. Also, some data source can't be access via open and license compatible method.

The third is for my opinion, our world is in danger, those years many regional conflict, or we can call it war, had happened. Ukraine with Russian, Israel with Palestine, and maybe something unbelievable in future Asian Pacific. Mapping military landuse will contribute nothing to world peace.

Snowysauce:
> If we approve this, I can see the future we will receive PR that want to add US Army, add PLA, add other countries real national armed force to a preset.

This PR already proposes that to an extent, with various agencies and armed forces departments from the US, Britain, India, etc. being in the initial batch of proposed presets. 

> The first, mapping landuse is illegal in many countries, handle a project containing those information(especially if there have data related to their countries) may make those contributor in danger.
> 
> The second, those feature can do only less with improving general mapping quality, for military information is hard to comfirm and ensure, this violate a important "verifiable" osm principle. Also, some data source can't be access via open and license compatible method.

I tend to agree with this take - there are a significant number of countries where any knowledge of military or intelligence properties can put someone or something in a legal quagmire, if not worse. Even American agencies have classified facilities that, if mapped, would open up the OSMF or individual contributors to severe scrutiny.

I guess the question I have is whether providing presets for operators of publicized, verifiable, and previously mapped facilities falls into the same legal grey area as what I described above. I don't know, and TBH this entire topic may be better suited for the OSM forums or other, more mainstream discussion channels. @cj-malone implied above that military tagging in general is spotty, and if the OSM community is still hashing out what the basic tagging conventions should be, then IMO those tags aren't ready to be assisted by NSI presets.

> The third is for my opinion, our world is in danger, those years many regional conflict, or we can call it war, had happened. Ukraine with Russian, Israel with Palestine, and maybe something unbelievable in future Asian Pacific. Mapping military landuse will contribute nothing to world peace.

I obviously agree with this take too, but I'm torn as to how much this reality affects the core question of whether we should provide presets for operators of publicized, verifiable, and previously mapped facilities. On one end, it can be seen as simply standardizing what already exists in OSM, but on the other, the mere presence of the presets could easily send a signal to editors - rightly or wrongly - that "yes, we should be mapping every one of this armed forces'/agency's facilities".

I look at #10181 as an example of good intentions gone bad. Originally, a small bunch of cameras were (likely) innocently but erroneously mistagged as `operator=Flock Safety`, then it became a preset though OSM planet imports, and then the preset's presence caused the problem to snowball until it was finally deactivated in March. The OSM tag history tracker shows a [clear correlation between the preset's deactivation and the end of the problem's growth](https://taghistory.raifer.tech/#***/operator/Flock%20Safety). My point is that what we, the NSI, send to iD can have a powerful impact on tagging, and not necessarily in good ways or the ways that were intended.

bhousel:
This seems ok, so I'm just going to merge it as-is.  If anything breaks someone will tell us.

LaoshuBaby:
图片

I strongly recommend check it’s really necessary whether to add this tree, because today I noticed that the previously highly controversial PR was ultimately merged for the sake of efficiency.

Now, in the iD editor, if you accidentally click on a military area you will be prompted to add more detailed information. This can lead mappers to mistakenly think that OSM is encouraging them to add more military data, which might put them at risk of violating local laws.

daishu0000:
I think this PR will put some unsuspecting mapper at serious legal risk, because drawing military zones is illegal in many countries.

PR #11095, opened 2025-06-23
Add brand "Wasserwacht" (edited version)

as-eu:
This is the edited version of #10407 containing the improvements proposed by @Snowysauce.


AntiCompositeNumber:
@Snowysauce, do you have comments?

----
This feels more like an `operator=` than a `brand=`. That's where the rest of the lifeguard entries are, including an existing one for the Bayerisches Rotes Kreuz. We do have some ambulance stations in the brand tree, but other emergency response POIs are all `operator=`. I don't claim to understand the German Red Cross system though, so I might be missing a good reason that these should be brands.

Snowysauce:
> @Snowysauce, do you have comments?
> 
> This feels more like an `operator=` than a `brand=`. That's where the rest of the lifeguard entries are, including an existing one for the Bayerisches Rotes Kreuz. We do have some ambulance stations in the brand tree, but other emergency response POIs are all `operator=`. I don't claim to understand the German Red Cross system though, so I might be missing a good reason that these should be brands.

I have the exact same opinion, right down to the caveat that maybe I'm missing something as to why this should be a brand instead of an operator.

as-eu:
I touched on this in a comment I made earlier: https://github.com/osmlab/name-suggestion-index/pull/10407#issuecomment-2709136815

The German Red Cross is divided into various regional and local subdivisions (regional associations -> district associations -> local associations where applicable).
These are all independent registered associations and are usually the operators of local services.

I hope this information is helpful.

Am 16. Juli 2025 06:35:34 MESZ schrieb Snowysauce ***@***.***>:
>Snowysauce left a comment (osmlab/name-suggestion-index#11095)
>
>I have the exact same opinion, right down to the caveat that maybe I'm missing something as to why this should be a brand instead of an operator.
>

Snowysauce:
With that explanation, I'm going to stick with my original thoughts and say that these entries belong in the `operators` tree.

as-eu:
Could you explain your toughts a bit further? It's not understandable for me at the moment.

Thx.

>With that explanation, I'm going to stick with my original thoughts and say that these entries belong in the `operators` tree.

bhousel:
I just added a basic "Wasserwacht" entry to the existing `operators/emergency/lifeguard.json` file in 0555b76b1

as-eu:
> I just added a basic "Wasserwacht" entry to the existing `operators/emergency/lifeguard.json` file in [0555b76](https://github.com/osmlab/name-suggestion-index/commit/0555b76b18b2718fdff0e1c2f86daa3a15642bde)

But this doesn't solve the discussion. In my opinion, this suggests a false tagging to users.
There is no "Wasserwacht" operator! The services are always run by local groups. I translated the relevant section from the German [Wikipedia article](https://de.wikipedia.org/wiki/Wasserwacht#Ortsgruppen):

> Local groups
> The local groups form the basis of the water rescue service's work. They run the water rescue service and provide training in swimming, lifesaving, first aid, and nature conservation.

This is pretty comparable to the Malteser Hilfsdienst, as discussed here: #6909

bhousel:
> But this doesn't solve the discussion. In my opinion, this suggests a false tagging to users.
> There is no "Wasserwacht" operator! The services are always run by local groups.

ok, sorry, I will move it to the `brands` tree

PR #11096, opened 2025-06-23
Update Boseh in bicycle rental

NFarrasP:
Boseh is ran by UPT Angkutan Dinas Perhubungan Kota Bandung, a local government's department in Bandung City. So it is a government operator, not a private entity. The Wikidata item is for Dinas Perhubungan Kota Bandung (parent organization) since creating UPT Angkutan is a bit unnecessary right now.

### DIFF
--- a/data/brands/amenity/bicycle_rental.json
+++ b/data/brands/amenity/bicycle_rental.json
@@ -985,8 +985,8 @@
         "network": "Boseh",
         "network:wikidata": "Q56392276",
         "operator": "Boseh",
-        "operator:type": "private",
-        "operator:wikidata": "Q56392276"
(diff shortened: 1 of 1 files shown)


Snowysauce:
> Boseh is ran by UPT Angkutan Dinas Perhubungan Kota Bandung

Shouldn't this be the `operator` value then, regardless of whether it has a Wikidata item?

NFarrasP:
> > Boseh is ran by UPT Angkutan Dinas Perhubungan Kota Bandung
> 
> Shouldn't this be the `operator` value then, regardless of whether it has a Wikidata item?

You are right. I have modified it.

PR #11099, opened 2025-06-23
Add swimming pools and community centers for City of Houston

EvanCarroll:
This adds a new file useful for non-branded municipal pools, and also the community centers for City of Houston

* Website for pools, https://www.houstontx.gov/parks/swimming.html
* Website for community centers, https://www.houstontx.gov/parks/communitycenters/index.html

### DIFF
--- a/data/operators/amenity/community_centre.json
+++ b/data/operators/amenity/community_centre.json
@@ -250,6 +250,20 @@
       }
     },
     {
+      "displayName": "City of Houston Swimming Pool",
+      "id": "cityofhouston-1d14a3",
+      "locationSet": { "include": ["us-tx.geojson"] },
+      "tags": {
+        "amenity": "community_centre",
+        "operator": "Houston Parks and Recreation Department",
+        "operator:short": "HPARD",
+        "operator:type": "government",
+        "operator:wikidata": "Q115018877",
+        "owner": "City of Houston",
+        "owner:wikidata": "Q16555"
+      }
+    },
+    {
       "displayName": "City of Kitchener",
       "id": "cityofkitchener-6e645e",
       "locationSet": {


Snowysauce:
My concern with the `leisure/sports_centre` preset as proposed is that it will cause iD to suggest `sport=swimming` for any `leisure=sports_centre` operated by the Houston Parks and Recreation Department, regardless of whether or not it's a facility where swimming is the primary activity.

On a lesser note, `"preserveTags": ["^name"]` is unnecessary since the lone preset in the proposed category doesn't contain `name` tag suggestions.

EvanCarroll:
@Snowysauce is there a way to address your concern within the current implementation of the technology such that it's only suggested for swimming sports centers? Also wouldn't this also be the same problem in [`brands/leisure/sports_centre`](https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/leisure/sports_centre.json) which currently exists?

```json
{
  "displayName": "Bear Paddle Swim School",
  "id": "bearpaddleswimschool-f2066e",
  "tags": {
    "brand": "Bear Paddle Swim School",
    "brand:wikidata": "Q121333406",
    "leisure": "sports_centre",
    "name": "Bear Paddle Swim School",
    "sport": "swimming"
  }
}
```

What about making the name "City of Houston Swimming Pool"

Snowysauce:
>  is there a way to address your concern within the current implementation of the technology such that it's only suggested for swimming sports centers?

There isn't, no.

> Also wouldn't this also be the same problem in [`brands/leisure/sports_centre`](https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/leisure/sports_centre.json) which currently exists?

To me, this is an apples-to-oranges comparison - a parks and rec department can operate a variety of facilities, whereas the brand of swim schools presumably only use swimming venues. I personally wouldn't have included the `"sport": "swimming"` tag suggestion with the brand, but it's arguably a correct suggestion that I don't take an exception to.

> What about making the name "City of Houston Swimming Pool"

The preset can be called anything, but it still won't impact how iD processes the preset and what editors do with the suggestions that iD presents. (I've learned in the last few months that a majority of editors accept anything iD presents to them, even if it's something as incorrect as a European fuel brand for a fuel station in India.)

EvanCarroll:
Going to rollback that file and commit it to end the controversy.

PR #11100, opened 2025-06-24
Chinese fast food chain now has a Canadian location

ShinyDog2636:
Also updated English name to so only first letter of each word is capitalized.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -14189,16 +14189,16 @@
     {
       "displayName": "正新鸡排",
       "id": "zhengxinchickensteak-65b4e8",
-      "locationSet": {"include": ["cn"]},
+      "locationSet": {"include": ["cn","ca"]},
       "tags": {
         "amenity": "fast_food",
         "brand": "正新鸡排",
-        "brand:en": "ZHENGXIN CHICKEN STEAK",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Well, the all capitalized name is years ago I copy from the front signboard of its branch shop. This truly a mistake because all capitalized is a design style, not means name spelling method. 

But I'm curious that whether this brand in Canada also use Chinese character on signboard?

---
If not, I'd suggest create a global version and add Canada to that

ShinyDog2636:
The English is slightly larger but both are present on the sign.

https://nomsmagazine.com/wp-content/uploads/2024/12/zhengxin-chicken-steak-burnaby-4.jpg

I just did a quick search and it looks like this chain has also opened locations in USA, Australia, and Singapore. Maybe you are right that a global version would be better. I will review later, Thanks.

Snowysauce:
Canada (and any other countries where the primary language isn't Chinese) should have a separate preset, with the Latin text being the values for `name` and `brand`.

ShinyDog2636:
It looks like the number of locations in the English-primary world is very small (1 in Canada, 2 in USA). So small that I think it may not deserve it's own listing in the NSI.

I will just manually update the location in Canada that triggered me to create this patch instead of updating the NSI.

Thanks.

PR #11102, opened 2025-06-25
Supercor duplicated

Crashillo:
Brand Supercor is duplicated, both as a [convenience](https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/shop/convenience.json#L5594-L5603) and as a supermarket. This last one is wrong, therefore is removed

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -8195,17 +8195,6 @@
       }
     },
     {
-      "displayName": "Supercor",
-      "id": "supercor-d9bffc",
-      "locationSet": {"include": ["es", "pt"]},
-      "tags": {
-        "brand": "Supercor",
-        "brand:wikidata": "Q6135841",
-        "name": "Supercor",
-        "shop": "supermarket"
-      }
-    },
-    {
       "displayName": "SuperDino",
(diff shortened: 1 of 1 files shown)


Snowysauce:
According to the workflow tests, the convenience preset needs to be tagged with `"matchTags": ["shop/supermarket"],`; otherwise, the supermarket preset will be re-added by the build script on its first run post-merge.

UKChris-osm:
According to the [Supercor Web Site](https://www.supercor.es/tiendas/) they do trade as both Supermarkets and Convenience stores!

Crashillo:
> According to the [Supercor Web Site](https://www.supercor.es/tiendas/) they do trade as both Supermarkets and Convenience stores!

Well, that's a language misunderstading: in spanish, everything is a "supermarket". _Supercor_ belongs to _El Corte Inglés_ chain, as well as _Hipercor_, however this last one would match better with the english term supermarket, and Supercor would match with convenience stores. Indeed, _Supercor_ has also _Supercor Exprés_, but both are convenience stores, business-wise speaking.

If you allow to tag Supercor as a supermarket also, everybody would tag it only as a supermarket.

Crashillo:
No comments?

PR #11104, opened 2025-06-25
Remove pakketautomaat text from DHL and PostNL entries

RobinLinde:
Split off from #11071.

Removes the "pakketautomaat" text from two entries, which can be seen as descriptive rather than being part of the brand. Draft for now, waiting for more community feedback

### DIFF
--- a/data/brands/amenity/parcel_locker.json
+++ b/data/brands/amenity/parcel_locker.json
@@ -271,6 +271,17 @@
       }
     },
     {
+      "displayName": "DHL",
+      "id": "dhl-476b76",
+      "locationSet": {"include": ["nl"]},
+      "matchNames": ["dhl pakketautomaat"],
+      "tags": {
+        "amenity": "parcel_locker",
+        "brand": "DHL",
+        "brand:wikidata": "Q132858576"
+      }
+    },
+    {
       "displayName": "DHL BOX 24/7",
       "id": "dhlbox247-ce4e28",
       "locationSet": {"include": ["pl"]},
@@ -311,17 +322,6 @@
         "brand": "Paketbox",
         "brand:wikidata": "Q2046604",
         "name": "DHL Paketbox"
-      }
-    },
-    {
-      "displayName": "DHL Pakketautomaat",
-      "id": "dhlpakketautomaat-476b76",
-      "locationSet": {"include": ["nl"]},
-      "tags": {
-        "amenity": "parcel_locker",
-        "brand": "DHL",
-        "brand:wikidata": "Q132858576",
-        "name": "DHL Pakketautomaat"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


bhousel:
It's been 4 months - was there ever any community feedback?

rogoyski:
I strongly support this. In fact, I would like to one day propose removal of the generic term for parcel lockers from brand names in my country, together with all the substitutions companies made up, because one company somehow manage to get the term _parcel locker_ registered as a trademark.

The names we provide are supposed to used in maps. And in maps, as much information as possible is supposed to be conveyed through symbols. It can be expected that a map showing parcel lockers will have an icon for that POI type and so repeating the same information in text is an overkill and just adds clutter to precious map space.

Dwarf-Nebula:
@RobinLinde it's currently marked as draft. It looks fine to me, so could you mark it as final?

Aimmig:
The germany entry for DHL Paketbox has very similar problems with purely descriptive name and brand which is not really a brand. Paketbox just means "box to put your parcel in or pick it up", they also exist from different post services.

Dwarf-Nebula:
> The germany entry for DHL Paketbox has very similar problem

I think it helps if you create an issue for it with exactly which nsi entries have this problem. Because this pull request is only for NL.

Dwarf-Nebula:
@bhousel @GA-Kevin-Codes @1ec5 @UKChris-osm could one of you change this PR to final state and merge it? The change looks fine to me and would help remove the fake names on these parcel lockers.

UKChris-osm:
If "_pakketautomaat_" is to be removed, would it make more sense to use the generic DHL wikidata entry (with usable logo), rather than the current [DHL Pakketautomaat](https://www.wikidata.org/wiki/Q132858576) wikidata that seems to imply the branding is unique, and as the wikidata itself is quite minimal in terms of information?

For example, the [Royal Mail](https://nsi.guide/index.html?t=*&k=amenity&v=parcel_locker&tt=roy#royalmail-ad77d0) parcel lockers in the UK use the general purpose [Royal Mail wikidata](https://www.wikidata.org/wiki/Q638098) which seems to work well.

Also, with "_PostNL_", I don't see the benefit of having the `brand` and `operator` tags if both are the same. I'd be inclined to remove the operator tags in this entry as they serve no additional purpose.

Dwarf-Nebula:
> If "_pakketautomaat_" is to be removed, would it make more sense to use the generic DHL wikidata entry (with usable logo), rather than the current [DHL Pakketautomaat](https://www.wikidata.org/wiki/Q132858576) wikidata that seems to imply the branding is unique, and as the wikidata itself is quite minimal in terms of information?
> 
> Also, with "_PostNL_", I don't see the benefit of having the `brand` and `operator` tags if both are the same. I'd be inclined to remove the operator tags in this entry as they serve no additional purpose.

I would say both are valid comments and they could both be implemented.

Dwarf-Nebula:
@UKChris-osm do you need anything to merge this pull request? I can't create a commit in this PR to solve the issues you mentioned. When needed I could create a new PR based on this one, and this one could then be closed.

PR #11118, opened 2025-07-01
Add Café Local

peternewman:
Alternative branding of Pumpkin.

I didn't know if we should rename, or add a second (as I've currently done) or...?

See:
https://en.wikipedia.org/wiki/Pumpkin_Caf%C3%A9_Shop
https://www.wikidata.org/wiki/Q27825961 - Already been updated by others.
https://tpt.amey.co.uk/news/all-change-at-havant-as-ssp-continue-cafe-local-conversions/

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -938,6 +938,18 @@
       }
     },
     {
+      "displayName": "Café Local",
+      "locationSet": {"include": ["gb"]},
+      "tags": {
+        "amenity": "cafe",
+        "brand": "Café Local",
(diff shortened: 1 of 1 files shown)


ilias52730:
It seems like a rebranding.
1. Have all the locations been rebranded? I can't find any recent data.
2. *If* all locations have been rebranded, you should probably just replace the Pumpkin entry with the Café Local entry, and add "Pumpkin", "Pumpkin Café" and "Pumpkin Café Shop" as Matchnames, so that old entries on OSM get a warning to be updated.

Issue #5693, opened 2021-11-16
Superchargers want to flip over to destination chargers


tas50:
I've noticed that a lot of previously tagged superchargers want to update to destination chargers. I think we're going to lose a lot of good data as folks unintentionally update these. Here's an example of one with Supercharger in the name that really wants to become a destination charger.

arnowrld:
That's a good point. Is there any possibility of making sure that these are separate in the near future?

bhousel:
Just opened https://github.com/openstreetmap/iD/pull/8828  with a fix for this..

arch0345:
Looks like the PR was merged - is this issue safe to close now?

tas50:
Seems good now

arch0345:
It looks like the same thing is happening with the `man_made=charge_point` preset for superchargers. I've opened #8237 to track this

bhousel:
At some point this bug regressed and it's back.  
I'm seeing the Supercharger preset immediately offer the "upgrade" to the Destination Charger preset.

This probably means that our matching code is getting confused by the tags on these presets.  I'll check it out.

bhousel:
I did a bit of digging today and here's some rough thoughts on what I found:

There is code in Rapid and iD which generates a "reverse index" of key/value pairs back to trees. (`vmap` in the code).

I guess this has come up several times in this project about the issues that can happen when a category like charge points or charging stations lives in 2 trees.  Some thoughts here:  https://github.com/osmlab/name-suggestion-index/pull/5623#issuecomment-956279182

But anyway, this reverse index means that the validator code in Rapid and iD is treating all of the charging stations and charge points as "operators" instead of "brands".  I think this is because when building the reverse index, the code first sees the "brands/amenity/charging_station" and decides "it's a brand", and then later sees "operators/amenity/charging_station" and then overwrites it with "it's an operator".

What this actually means in terms of matching is:  the `operator` tags are getting prioritized over the `brand` tags.  This causes problems for the Tesla entries: they exist as both supercharger or destination charger which both have the same `operator` but are distinguished by their `brand`.  But the matcher is looking mainly at `operator` and the `brand` is considered an alternate/backup match.

There is no great way to fix this, but I'm going to try by just - generate the dist files in reverse, so that the code in Rapid and iD will build their reverse index so that "brand" overwrites "operator".  I think this is a quick fix but might solve the problem well enough for now, and get the matcher code to work more like how people expect it to work.

[...]

PR #11120, opened 2025-07-02
[website] support viewing all trees together (brand+operator+...)

### LINKED ISSUE #8706: Search through all types of tree
It's quite confusing, UX-wise, that using the search input cannot return _flags, operators, or transit_ results, until you change that scope.

As a user, I expect to find any result just typing the name, and not clicking firstly which tree I want to use, as I could not know where the result lies.

Front-end wise, there are not enough elements in the secondary trees, to avoid merge them all in the homepage. Something like this:

imagen

By the way, notice the misplacement of the filters bar after typing something

k-yle:
Closes #8706 and unblocks #10601

This PR is deployed to GitHub Pages, so it can be tested from [https://kyle.kiwi/name-suggestion-index](https://kyle.kiwi/name-suggestion-index/?t=brands)

---

You can now browse https://nsi.guide by preset, without first choosing whether you want `brands` or `operators`. I found this distinction quite confusing, and it seems like others did too. So hopefully it's easier to understand now.

These links:

image

Have been replaced with a dropdown:

image

and one option is <kbd>*</kbd> which shows data from all 4 trees. 

For example: you can now view all `amenity=library` on a single page: https://kyle.kiwi/name-suggestion-index/index.html?k=amenity&v=library


bhousel:
I”ll have to think a lot about this.  The `tree` distinctions are pretty important for how the tags are expected to work.

bhousel:
> For example: you can now view all `amenity=library` on a single page: https://kyle.kiwi/name-suggestion-index/index.html?k=amenity&v=library

I'm also concerned that there are 4 library "networks" in the brand tree now?  
Who did that?  Are we sure that won't cause surprises?

update: I guess it was mentioned here,  https://github.com/osmlab/name-suggestion-index/issues/5625#issuecomment-956513420
It makes me super nervous.  We've had issues with the tag matching for [EV chargers](https://github.com/osmlab/name-suggestion-index/issues/5693#issuecomment-2819259226) most recently.

k-yle:
> The `tree` distinctions are pretty important for how the tags are expected to work.

I realise this concept must be straightforward for NSI maintainers, but as an ordinary mapper, I find it really confusing that some entries can have both `brand` and `operator` tags[^1] **but** if you click on <kbd>operators/</kbd> in the website, you won't find these entries because they're actually under <kbd>brands/</kbd> (or <kbd>networks/</kbd>).

Example: I see a strange warning in Singapore about `operator=Best Express`. Taginfo doesn't take me to the NSI documentation (#10601), so I go to [nsi.guide](https://nsi.guide/index.html?t=operators&k=amenity&v=post_office) and click on [amenity/post_office](https://nsi.guide/index.html?t=operators&k=amenity&v=post_office) but there's nothing related to `Best Express`. So to an ordinary user, it looks like NSI doesn't have data about this brand 🤔 It takes a lot of searching to realise that there's a default filter in the navbar which is hiding 47 out of 196 post offices, despite some of them having both `brand` and `operator` tags.

> I'm also concerned that there are 4 library "networks" in the brand tree now?
> Who did that? Are we sure that won't cause surprises?

This is a great example of why I think this would be helpful. Since these quirks do exist, it would be great if the website helps users to understand the situation, rather than adding to the confusion

[^1]: or `operator` _and_ `network`

UKChris-osm:
I think this is a nice addition to the web site, although I think the dark text with a dark background makes the new navigation a little difficult to read.

new-nsi-tree

Regarding the library examples though, I do wonder if the four branded libraries would fit into the operator tree without any issue?

k-yle:
thanks Chris, the colour issue on windows should be fixed now

bhousel:
I'm just going to merge this, even though I still have doubts.

Issue #10158, opened 2024-11-19
add Wrocławskie krasnale


PavLL:
{
"displayName": "Wrocławskie krasnale",
"locationSet": {"include": ["pl"]},
"tags": {
   "artwork_type": "dwarf",
   "brand": "Wrocławskie krasnale",
   "brand:wikidata": "Q910242",
   "brand:wikipedia": "pl:Wrocławskie krasnale",
   "tourism": "artwork"
   }
}

PavLL:
It's been proposed on the [forum thread](https://community.openstreetmap.org/t/oznaczanie-wroclawskich-krasnali/124013) to use `artwork_type=statue` instead as the figurine type can be differentiated by the `brand` tag.

Personally I'm in favor of this change but I think we should ask @maro-21 , the creator of the wiki page for [artwork_type=dwarf](https://wiki.openstreetmap.org/wiki/Pl:Tag:artwork_type%3Ddwarf) for his opinion.

maro-21:
I'm against adding it to NSI or changing the tagging.

Wrocławskie krasnale is not a brand, but network.

The [brand:wikidata=Q910242] tag is superfluous - [network=Wrocławskie krasnale] alone already uniquely identifies Wrocław dwarves.

PavLL:
Why is one's opinion more important than another's? Wroclaw's dwarfs should be added as to NSI because it would be very useful to be able to add the tags from a preset for new figurines.

In my opinion Wroclaw's dwarfs are a brand but sure, that's up to debate by the community once the issue is reopened.

I don't know why is the argument saying that the network tag already tells everything is accepted. The *:wikidata tag is obviously needed to add extra important such as the Wikipedia article, website, etc.

Please reopen this issue.

matkoniecz:
I assume that NSI maintainers prefer to wait for clarity on preferred tagging.

Especially if it introduces new or unusual tagging.

maro-21:
> Why is one's opinion more important than another's?

No one's opinion is more important than another's. An opinion is an opinion.

> Wroclaw's dwarfs should be added as to NSI because it would be very useful to be able to add the tags from a preset for new figurines

It can be added, but using the current, established tagging. Previously you proposed tagging that has never been used ["artwork_type": "figurine" + "brand": "Wrocławskie krasnale"]. That's why I was against. I am strongly against changing the tagging of the same objects that has been established for years **using NSI**. NSI is not an OSM database.

[...]

PR #11124, opened 2025-07-03
Add Wrocław's dwarfs

PavLL:
Added Wroclaw's dwarfs and its category. I don't how what the ID should be.


bhousel:
This is just the thing we decided we weren't going to do in #10158

PavLL:
> This is just the thing we decided we weren't going to do in #10158

Why? You left my questions unanswered.

Wrocław's dwarfs are an instance of a brand so there's nothing stopping them from being added to NSI. Please reopen this PR.

PR #11125, opened 2025-07-03
Added British Columbia car repair chain

ShinyDog2636:
I am somewhat hesitant regarding the use of "and" vs "&". The official website uses "and" but signage uses "&".

OSM default to "on the ground truth" but also "If the name can be spelled without an abbreviation, then don't abbreviate it".

I see many other car repair chains in the NSI have used "&".

I have used "and" but will accept if this project's maintainers prefer "&".

Thanks.

### DIFF
--- a/data/brands/shop/car_repair.json
+++ b/data/brands/shop/car_repair.json
@@ -1752,12 +1752,12 @@
     },
     {
       "displayName": "Budget Brake & Muffler Auto Centres",
-      "locationSet": {"ca-bc.geojson"},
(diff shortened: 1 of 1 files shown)


bhousel:
The ampersand in the name is fine - our matching scripts can match either '&' or 'and' equally.
I'd just use the ampersand it if that is what the signs say, and their social media accounts use it.

https://x.com/BBMACofficial
https://www.facebook.com/BudgetBrakeAuto

ShinyDog2636:
Ok thanks. I have updated the file.

PR #10806, opened 2025-03-27
Update paint.json(Update CAN brand name to match Wikidata)

meyerk1456:
Updated brand name for Dulux to match Wikidata and official website.

https://www.wikidata.org/wiki/Q50921
https://dulux.ca/

PR #11126, opened 2025-07-03
Update Dulux paint.json

Happy2482:
update brand Dulux Paint to Dulux for AUS as per Wikidata and official website, because Dulux Paint is the name of the product https://www.dulux.com.au/specifier/products/, https://www.dulux.com.au/about-us/ Dulux (Q50921)

### DIFF
--- a/data/brands/shop/paint.json
+++ b/data/brands/shop/paint.json
@@ -137,14 +137,14 @@
       }
     },
     {
-      "displayName": "Dulux Paints",
+      "displayName": "Dulux",
       "id": "duluxpaints-573ec1",
       "locationSet": {"include": ["001"]},
       "matchNames": ["dulux"],
       "tags": {
-        "brand": "Dulux Paints",
+        "brand": "Dulux",
         "brand:wikidata": "Q50921",
-        "name": "Dulux Paints",
+        "name": "Dulux",
         "shop": "paint"
       }
     },


Snowysauce:
This PR as currently coded would change the preset globally, not just in Australia, and #10806 established that the current preset values are correct for at least one country (Canada).

UKChris-osm:
Looking at the web site, do Dulux only have stores called "Dulux Trade Centre" in Australia?

Happy2482:
> Looking at the web site, do Dulux only have stores called "Dulux Trade Centre" in Australia?

Yes, agree... only Dulux Trade Centre, the rest are more like maybe stores (not owned by Dulux) selling Dulux.

UKChris-osm:
In that case I'd probably recommend splitting this entry so that Australia has its own listing using the name "_Dulux Trade Centre_".

Happy2482:
> In that case I'd probably recommend splitting this entry so that Australia has its own listing using the name "_Dulux Trade Centre_".

Thanks for checking, but since Dulux Trade Center, also has Dulux as the brand, so maybe only Dulux as the brand still fine, but not sure, if we really need to add another for Dulux Trade Center, then please allow me to ask, you mean we will have 2 brands right with same brand, but different name... if yes, not sure if this is fine... maybe do you have example for this case? interesting to me...

ilias52730:
@Happy2482 just look at Google Street View. If a country has Dulux stores but with a different name displayed, then you need a different preset. Please start another PR with the correct information.

Example in Scotland:
image
Example in Australia:
image

PR #11127, opened 2025-07-04
Add Quality RV Resorts

EvanCarroll:


UKChris-osm:
They all seem to have unique names, so it might be worth adding a `preserveTags` for that in this case.

I'm not sure about `brand:short`, is that something that's added to the index these days? I'll wait for other opinions before merging.

EvanCarroll:
@UKChris-osm yes, I know we use it for HPARD and such.

PR #11131, opened 2025-07-06
Add Fluevog Shoes to shoes.json

ShinyDog2636:
Canadian shoe store chain operating in CA and US

### DIFF
--- a/data/brands/shop/shoes.json
+++ b/data/brands/shop/shoes.json
@@ -2802,12 +2802,12 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Is this brand not called "_John Fluevog_"?

ShinyDog2636:
I have revised to "John Fluevog Shoes" but "John Fluevog" appears about as often on OSM. From the storefront images on their '[location' page](https://www.fluevog.com/stores/) it appears both are used. Maybe either is acceptable, depending on the specific location?

PR #11133, opened 2025-07-07
Switzerland: Add Blackout clothing store chain

habi:
In the course of fixing some POIs mentioned in https://community.openstreetmap.org/t/vogele-shoes-pois-loschen/132208 I found out that the https://www.blackout.ch/ shops are missing from NSI.

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -1204,6 +1204,16 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


habi:
"More than 90 branches" according to https://www.blackout.ch/storefinder

codeinabox (approved):
Well spotted, and thank you!

PR #11136, opened 2025-07-08
Adding NSI entry  for french food_bank "Les Restos du Coeur"

oldnab:
Major french food_bank with 2300 sites needs "official" tags in OSM.
This proposal is to add an entry in NSI for this target.

### DIFF
--- a/data/operators/amenity/social_facility.json
+++ b/data/operators/amenity/social_facility.json
@@ -26,7 +26,7 @@
         "amenity": "social_facility",
         "social_facility": "food_bank",
         "social_facility:for": "underprivileged",
-        "operator": "Les Restos du cœur",
+        "operator": "Les Restos du Cœur",
         "operator:type": "charitable",
         "operator:wikidata": "Q171205"
       }
--- a/data/operators/amenity/social_facility.json
+++ b/data/operators/amenity/social_facility.json
@@ -12,7 +12,7 @@
   "items": [
   {
       "displayName": "Les Restos du Cœur",
-      "locationSet": {"include": ["fr"]},
+      "locationSet": {"include": ["fx"]},
       "matchNames": [
         "Les Restos du Coeur",
         "Les Restaurants du Cœur",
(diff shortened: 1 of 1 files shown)


data/operators/amenity/social_facility.json
```diff
@@ -10,6 +10,27 @@
     }
   },
   "items": [
+  {
+      "displayName": "Les Restos du Cœur",
+      "locationSet": {"include": ["fr"]},
```
codeinabox: Are the food banks only in metropolitan France, or do they also appear in the overseas French territories? If just the former, you can use `fx` as the location

oldnab: After checking : Only in metropolitan France.
Thanks for alerting.

oldnab: In fact, I think I'll keep fr because :

1.  fx is not yet / no more  an iso3166-2 code (seems to me)
2. It is unlikely that a food bank in France overseas will be created with operator name restos du coeur except if the metropolitan association  decides to extend.

Snowysauce: The `locationSet`s of presets in this project use many codes beyond those in ISO 3166-2, and it's preferred to use whatever code best suits the actual operating area of the entity. In this case, that code would be "fx".

oldnab:
Ok. Moved fr to fx
Thanks both

ilias52730:
Why did you choose _Les Restos du cœur_ instead of _Les Restos du Cœur_? I see a capital C everywhere, starting with their [website](https://www.restosducoeur.org/). Apart from this the PR looks good to me.

oldnab:
You are right. This was an error. I moved to capital C
Thanks

ilias52730:
It seems it… [already exists](https://github.com/search?q=repo%3Aosmlab%2Fname-suggestion-index%20les%20restos&type=code), somehow? I guess it was added automatically in the meantime. Nevertheless, the C should still be capitalized. I can correct it and this PR can be closed.

PR #11138, opened 2025-07-08
More Best Western/BWH brands

1ec5:
Added the rest of the hotel chains by BWH Hotels, including SureStay and WorldHotels. (I didn’t distinguish the various classes of WorldHotels since it isn’t clear to me whether anything besides booking sites distinguish them.) Updated the Wikidata tags of various existing Best Western brands to stop overloading [Q830334](https://www.wikidata.org/wiki/Q830334), especially since some of them have very different branding.

### DIFF
--- a/data/brands/tourism/hotel.json
+++ b/data/brands/tourism/hotel.json
@@ -21,6 +21,18 @@
   },
   "items": [
     {
+      "displayName": "@HOME",
+      "id": "home-779ccb",
+      "locationSet": {"include": ["001"]},
+      "tags": {
+        "brand": "@HOME",
+        "brand:wikidata": "Q135249100",
+        "name": "@HOME",
+        "official_name": "@HOME by Best Western",
+        "tourism": "hotel"
+      }
+    },
+    {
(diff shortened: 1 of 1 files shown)


data/brands/tourism/hotel.json
```diff
@@ -20,6 +20,18 @@
...
   },
   "items": [
+    {
+      "displayName": "@HOME",
+      "id": "home-779ccb",
+      "locationSet": {"include": ["001"]},
+      "tags": {
+        "brand": "@HOME",
+        "brand:wikidata": "Q135249100",
+        "name": "@HOME",
```
1ec5, line 30: I’m unsure if this becomes “At Home” or just “Home” in normal writing, so I left it as the stylized name. By contrast, I left GLō and Vīb as the plain names instead of turning them into the stylized form.

1ec5:
2a01e63eb8aec66e2e085a8ccf8f6626c2139d93 adds Sure Hotel, the alter ego of SureStay.

PR #11139, opened 2025-07-11
Add flagpole entries for Japanese prefectural flags

momiji-takinogawa:


LaoshuBaby:
Looks good because those flag can be speculated that been used widly, like America's 50 states' flag was recorded.

(Additional, for second level administration region 市区町村 because the number is large, I think there isn't such notability for add them. This was recorded in [Judge Case](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case))

PR #11140, opened 2025-07-11
Update video_games.json

Mannivu:
Add Italian retailer brand


Cj-Malone:
See #11093, if GameStop is gone from IT, we can remove it.

Mannivu:
I'm unsure every GameStop shop has already become a Game life shop. I think they're in process of changing every shops.

Anyway, I'll close this PR since it's a duplicate.

PR #11149, opened 2025-07-14
Drop cuisine=vegetarian

Cj-Malone:
`cuisine=vegetarian` is strongly discouraged in favour of `diet:vegetarian=*`, and has basically [no usage](https://taginfo.openstreetmap.org/keys/cuisine#values) due to people cleaning it up. We shouldn't have it.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -14555,7 +14555,7 @@
         "brand:en": "FJ Veggie",
         "brand:wikidata": "Q130458837",
         "brand:zh": "芳珍蔬食",
-        "cuisine": "vegetarian",
+        "diet:vegetarian": "yes",
         "name": "芳珍蔬食",
         "name:en": "FJ Veggie",
         "name:zh": "芳珍蔬食",
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -2151,7 +2151,8 @@
         "amenity": "restaurant",
         "brand": "De Beren",
         "brand:wikidata": "Q57076079",
-        "cuisine": "burger;chicken;grill;vegetarian",
+        "cuisine": "burger;chicken;grill",
+        "diet:vegetarian": "yes",
         "name": "De Beren"
       }
     },


tiptoptom:
I don't know these brands but diet:vegetarian=only would also be an option.

ilias52730:
By reading [Key:diet:vegerarian](https://wiki.openstreetmap.org/wiki/Key:diet:vegetarian), it seems FJ Veggie should have `diet:vegetarian=only` because it's a vegetarian restaurant. De Beren is fine with just `diet:vegetarian=yes` because they offer other non-vegetarian options.

Cj-Malone:
Fixes #11313

data/brands/amenity/fast_food.json
```diff
@@ -14555,7 +14555,7 @@
         "brand:en": "FJ Veggie",
         "brand:wikidata": "Q130458837",
         "brand:zh": "芳珍蔬食",
-        "cuisine": "vegetarian",
```
UKChris-osm, line 14921: Is there any value that could be added to `cuisine` here, rather than leave it blank?

ilias52730: From what I read, `cuisine=chinese;dumpling;noodle`

Cj-Malone:
Rel: https://github.com/osmlab/name-suggestion-index/pull/11468

maxwellward:
Hi all, has there been any movement on this PR? Is there anything else needed for it to be merged other than potentially adding the proper cuisine=* tags mentioned above?

bhousel:
I think this is a mistake, but I'm going to merge anyway.

`diet:vegetarian` is for saying that the restaurant offers vegetarian options (most do).
`cuisine=vegetarian` is for saying that the restaurant specializes in vegetarian food (most don't).

maxwellward:
`cuisine=vegetarian` is an unused/deprecated tag. If a restaurant only offers vegetarian food, then `diet:vegetarian=only` should be used, as per the approved tag here: https://wiki.openstreetmap.org/wiki/Key:diet:vegetarian

PR #11150, opened 2025-07-14
Updating the name of BP Pulse to match capitalization on website and signs.

GA-Kevin-Codes:
Changing the tags and name of BP Pulse to bp pulse, to match the branding of the stations present online and on signs.

Please see the bp pulse website where all mentions of the brand are in lowercase: https://www.bppulse.com/en-us/public-ev-charging

Also included on signage including up to the gigahub:

image

### DIFF
--- a/data/brands/amenity/charging_station.json
+++ b/data/brands/amenity/charging_station.json
@@ -128,7 +128,7 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks @GA-Kevin-Codes

PR #11152, opened 2025-07-16
Improve Holiday Oil

dknelson9876:
I previously added Holiday Oil under `amenity=fuel`, but forgot to add it under `shop=convenience`. 

Additionally, to prevent further confusion Holiday Stationstores (Q5880490) are excluded from Utah to prevent bad suggestions, since they do not operate in Utah according to the map on their website

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -3173,7 +3173,7 @@
     {
       "displayName": "Holiday",
       "id": "holiday-c5cf43",
-      "locationSet": {"include": ["us"]},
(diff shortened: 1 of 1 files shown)


EvanCarroll:
Are there really so many convenience stores in Utah with the name Holiday to warrant a commit to stop suggesting an almost nationwide franchise that only doesn't exist in Utah?

It also seems they don't exist in Nebraska.

ilias52730:
It seems that even in Utah are called Holiday.
image

bhousel:
Based on the image above this seems like a mistake, so let's close until we know better.
They might be going through a rebrand and just haven't updated all their signs yet.

PR #11153, opened 2025-07-16
Update shoes.json: Add Ritmo Shoes

airon90:
+Ritmo Shoes


AntiCompositeNumber:
Per https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#notability, brands should generally have at least 30 locations to be added to NSI. This brand currently only has 15 locations. Please re-open or recreate this pull request when the brand has more locations.

PR #11157, opened 2025-07-20
Add Trojan Energy

peternewman:
I don't know if we want AON ( https://trojan.energy/aon ) and Hub ( https://trojan.energy/ev-charger-locations/ ) as separate sub brands?


bhousel:
No wikidata tags, so this won't do anything, but we can merge it anyway.

PR #11161, opened 2025-07-20
Add Kengur (RS)

u-spec-png:


data/brands/shop/stationery.json
```diff
@@ -440,6 +440,21 @@
...
     },
+    {
+      "displayName": "Кенгур",
+      "id": "kengur-f0e3r3",
+      "locationSet": {
+        "include": ["rs"]
+      },
+      "tags": {
+        "brand": "Кенгур",
+        "barnd:wikidata": "Q135428341",
```
Outlet2048: This seems like a typo
Should probably be "brand" instead of "barnd"

PR #11165, opened 2025-07-23
Update Stromnetz Hamburg replacement

DaN0mic:
Stromnetz Hamburg doesnt exist as an entity anymore, there is Hamburger Energiewerke and Hamburger Energienetze now.

### DIFF
--- a/data/operators/amenity/charging_station.json
+++ b/data/operators/amenity/charging_station.json
@@ -3414,6 +3414,16 @@
       }
     },
     {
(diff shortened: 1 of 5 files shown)


Outlet2048:
Does the ID need to change or can we continue to reuse the existing ID with updated name?

bhousel:
> Does the ID need to change or can we continue to reuse the existing ID with updated name?

The ID will get automatically set when we run the build.  It is a combination of the name + tagging + location, so it will get a new ID.

PR #11166, opened 2025-07-25
MA - Société Générale rebranded as Saham Bank

ilias52730:
In Morocco, Société Générale was bought by Saham and was rebranded Saham Bank ([source](https://maroc-diplomatique.net/la-societe-generale-maroc-devient-saham-bank/)). They display the name in French and Arabic (see image below). I updated the names accordingly.

image

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -12196,6 +12196,13 @@
     {
       "displayName": "Saham Bank سهام بنك",
       "locationSet": {"include": ["ma"]},
+      "matchNames": [
+        "société générale سوسيتيه جنرال",
+        "سوسيتيه جنرال",
+        "société générale",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
If there's been a name change, I'd keep the `matchNames` tag with the old name so that users will see a suggestion for updating older entries.

ilias52730:
@UKChris-osm I plan on personally updating all the entries through an automated edit, after having previously discussed it with the community. Is matchnames still necessary in this case?

UKChris-osm:
I'd say yes, because it gives other people a point of reference later on if they aren't familiar with the new branding.

ilias52730:
Thank you, I’ll update my PR shortly with all the matchnames.

ilias52730:
Matchnames have been added.

PR #11169, opened 2025-07-28
Add Univerexport (RS) as convenience store

u-spec-png:
Univerexport can be a convenience store (example: [1](https://www.openstreetmap.org/node/9329529177))

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -6074,6 +6074,21 @@
       }
     },
     {
+      "displayName": "Univerexport",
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Does zz want to me more than a match name? The example you gave has:
```
name 	Univerexport
name:sr 	Универекспорт
name:sr-Latn 	Univereksport
```

u-spec-png:
Serbian mapping community still hasn't agreed if such tagging is necessary. It's not wrong, and it's how I do it. But until we agree it will stay this way.

Issue #5292, opened 2021-08-31
```hot_pot``` or ```hotpot```?


LaoshuBaby:
I use ```cuisine=chinese;hotpot``` for those restaurant that serves "火锅" in ```restaurant.json```

But today I found ```hot_pot``` also been widely used, and another "hot-prefix" western food "hot dog" usually use ```hot_dog```

So I want to ask for community suggestion that which I should use in NSI

This question I will ask both in Github and Chinese local community.

![图片](https://user-images.githubusercontent.com/42690037/131494022-4bdfa424-eead-4e25-9bd2-106e78302998.png)

tntchn:
I'm not sure what kind of `hotpot` or `hot pot` those existing restaurants in OSM are, but I think hot pot is much better.

After googling for a while, I found that there was a English dish call "[Lancashire hotpot](https://www.bbcgoodfood.com/recipes/lancashire-hotpot)", not sure if this dish is widespread or not. But I still think using `hot_pot` is a easy way to avoid the confusion.

sorz:
[enwiki](https://en.wikipedia.org/wiki/Hot_pot) uses "Hot pot" as page title.
Although they also mentioned and used "hotpot" in the article and the talk page, "hot pot" is use more frequently (13 vs 82 times by a quick Ctrl-F).

1ec5:
At least in the U.S., “hotpot” or “hot pot”, unqualified, would refer to the Asian dish of Chinese origin; a Lancashire hotpot would need to be qualified as such to be recognizable.

I don’t think it’s necessary to avoid every potential naming conflict. After all, `cuisine=noodles` is reserved for Asian-style noodle dishes, as opposed to American-style chicken noodle soup, for example.

bhousel:
I think it's ok to just use `cuisine=hotpot` (or `cuisine=chinese;hotpot` is fine too) 
because it is the most frequently occurring value right now: 

<img width="832" alt="Screen Shot 2021-09-01 at 5 52 30 PM" src="https://user-images.githubusercontent.com/38784/131750341-cf794ccb-1fc1-4b15-b380-baa489973595.png">


A long time ago, this index contained a mix of `hotdog` and `hot_dog` tags, but I [changed them all to `hot_dog`](https://www.openstreetmap.org/changeset/74910608) just to match what the [OSM wiki](https://wiki.openstreetmap.org/wiki/Tag:cuisine%3Dhot_dog) said to do.

[...]

PR #11171, opened 2025-07-29
Rename cuisine tag `hotpot` to `hot_pot`

yaindrop:
The phrase "[Hot pot](https://en.wikipedia.org/wiki/Hot_pot)", like "Hot dog", should alway be used with an whitespace. The informal `hotpot` tag should be deprecated, and existing usages should be changed to `hot_pot`

### DIFF
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -3066,7 +3066,7 @@
         "brand:en": "Haidilao Hot Pot",
         "brand:wikidata": "Q5638920",
         "brand:zh": "海底捞火锅",
-        "cuisine": "chinese;hotpot",
+        "cuisine": "chinese;hot_pot",
         "name": "Haidilao Hot Pot",
         "name:en": "Haidilao Hot Pot",
         "name:zh": "海底捞火锅"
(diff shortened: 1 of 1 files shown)


peternewman:
Has this been agreed by the community @yaindrop or have you just decided to do it?
https://wiki.openstreetmap.org/w/index.php?title=Tag:cuisine%3Dhotpot&action=history
https://wiki.openstreetmap.org/w/index.php?title=Template:Map_Features:Cuisine_(style)&action=history

You should possibly add it to the ID presets, but it may not be popular enough:
https://github.com/openstreetmap/id-tagging-schema/pull/1146

Also for info:
https://github.com/openstreetmap/id-tagging-schema/pull/1146#issuecomment-1983511332

LaoshuBaby:
I noticed I also raised a same question in #5292 years ago because I think there should be more clarification on this special cuisine. (And I remember that I've changed both hotpot and hot_pot 2 pages on osmwiki but now I can't find the edit history)

------

@peternewman Indeed I've ask for advice in local community after #5292 (https://t.me/osmchina/1/26602) and got some opinion, some suggest that hot_pot is better and also have voice that both is acceptable.

LaoshuBaby:
> Has this been agreed by the community @yaindrop or have you just decided to do it?

This question also need to ask the user "something B" that whether this page movement is talked or just the user want to do

图片
图片
图片

data/brands/amenity/restaurant.json
```diff
@@ -9333,7 +9333,7 @@
         "brand:en": "Chien Yen Shabu Shabu",
         "brand:wikidata": "Q67936077",
         "brand:zh": "千葉火鍋",
-        "cuisine": "japanese;hotpot",
```
LaoshuBaby: Wait, looks like hot_pot was also used in other すき焼き like cuisine?

bhousel:
Going to just merge this, as it's been sitting for a while and the osm wiki says `hot_pot` is the tag now.

PR #11173, opened 2025-07-30
add Return and Earn bottle return for AU-NSW

andrewharvey:


GA-Kevin-Codes:
Hey @andrewharvey do all the locations have all the returns from this PR? Does the operator have a Wikidata? Thanks in advance!

andrewharvey:
Yes I'm pretty sure they all operate with the same criteria of what is accepted -> https://returnandearn.org.au/what-you-can-return

That said maybe we should drop those tags, I used `recycling:cans`, `recycling:glass_bottles`, `recycling:plastic_bottles` because it's the closest fit of what the accept but in reality it's only those bottles part of the scheme, with a specific label indicating the item is accepted as a redeemable. Ok I just did.

I just created and added the operator wikidata.

PR #11176, opened 2025-07-31
Update branding for Moss

TheUKHighStreet:
Brand changed from "Moss Bros" to "Moss" in 2023
https://www.moss.co.uk/about-us

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -7409,6 +7409,7 @@
         "clothes": "men",
         "name": "Moss",
         "alt_name": "Moss Bros",
+        "matchNames": ["Moss Bros"],
         "shop": "clothes"
       }
     },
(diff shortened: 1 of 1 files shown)


peternewman:
Do we need a matchNames for Moss Bros so it doesn't try and use Bros as a branch name or does the alt_name cover that?

TheUKHighStreet:
> Do we need a matchNames for Moss Bros so it doesn't try and use Bros as a branch name or does the alt_name cover that?

Yes - I think you're right - thanks. I'll update.

peternewman:
The only other thing I'd say is at least one shop I've seen recently still hasn't rebranded...

TheUKHighStreet:
Yes - tricky. The website lists all the stores as Moss but then has photos of Moss Bros. I think they're just being very slow with the rebranding.
https://www.moss.co.uk/store/269-moss-bros-derby

PR #11178, opened 2025-07-31
Add Charles Tyrwhitt

peternewman:
Probably has other countries from the store changer:
- [ ] DE
- [ ] FR
- [ ] AU
- [ ] CA
- [ ] CH
- [ ] NL

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -2211,6 +2211,21 @@
       }
     },
     {
+      "displayName": "Charles Tyrwhitt",
+      "locationSet": {
+        "include": [
+          "gb",
+          "us",
+        ]
+      },
+      "tags": {
+        "brand": "Charles Tyrwhitt",
+        "brand:wikidata": "Q924963",
+        "name": "Charles Tyrwhitt",
+        "shop": "clothes",
+      }
+    },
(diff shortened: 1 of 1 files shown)


ilias52730:
It seems they also have a store in France, in La Vallée Village shopping mall. Their website says it's "closed temporarily for refurbishment" but Google Maps says it's open and has reviews from a few days ago. IMO their website is not updated. I would add "fr" to the locationset.

GA-Kevin-Codes:
Instead of trying to capture where it is now we should consider where they may expand in the future, this seems to be anywhere worldwide, so locationSet should be [001](https://ideditor.codes/?code=Q2&map=1.77/-2.7102/0) in my opinion.

ilias52730:
> we should consider where they may expand in the future, this seems to be anywhere worldwide

I have no knowledge of Charles Tyrwhitt. Are they actively planning an expansion in the whole world? And even then, how much time would it take to open a store in 5, 10, 15, 20 countries? I don't think 001 should be added, but rather only the countries where they actually open a store. Each country can be added to the locationset accordingly.

Cj-Malone:
We don't need to sweat about a single POI, it location set can be edited later.

PR #11179, opened 2025-07-31
Update charging_station.json (Add Milence)

sn02:
Added Milence. (Operator of EV-Truck charging hubs in the EU. Regular EV-Cars are not allowed on any Milence location.)


GA-Kevin-Codes:
Hey @sn02! Thanks for the contribution! Is this strictly an operator or is it a signed /labelled brand as well? I tend to put most charging stations into Brand.

As far as the PR, you are missing a very minor but significant close bracket after the tags section, which is causing errors when we run our tests.

Finally, the amenity=charging_station wiki page states
> Charging Stations imply [motorcar](https://wiki.openstreetmap.org/wiki/Key:motorcar)=[yes](https://wiki.openstreetmap.org/wiki/Tag:motorcar%3Dyes), if not please specify with [motorcar](https://wiki.openstreetmap.org/wiki/Key:motorcar)=[no](https://wiki.openstreetmap.org/wiki/Tag:motorcar%3Dno) and reference [vehicle](https://wiki.openstreetmap.org/wiki/Key:vehicle) values for appropriate access.

So to me this should only be motorcar=no and possibly hgv=designated if only these vehicles can use them. I'll divert to your knowledge on the yes/designated point!

Looking forward to this PR! Thanks for the work!

sn02:
Hi @GA-Kevin-Codes. Thanks for the kind review. I think you're right in every point. I've edited the file, so hopefully everything is fine now.

GA-Kevin-Codes:
Looks good! Just make sure you're editing the brand file and not the operator file before we merge this :)

GA-Kevin-Codes:
Great work, thanks for the PR!

PR #11194, opened 2025-08-02
[JP] Remove telephone operator NTT Osaka Branch

momiji-takinogawa:
大阪支店 (Osaka branch office) of NTT West does not exist anymore, and its former operation area is now covered by 関西支店 (Kansan branch office)[^1].

I think it's better to just remove this entry for now to prevent mistagging, until someone wants to map telephone booths with their operating branch office information.

[^1]: https://www.ntt-west.co.jp/corporate/branch/

### DIFF
--- a/data/operators/amenity/telephone.json
+++ b/data/operators/amenity/telephone.json
@@ -355,17 +355,6 @@
       }
     },
     {
-      "displayName": "NTT大阪支店",
-      "id": "9b5070-944417",
-      "locationSet": {
-        "include": ["jp-27.geojson"]
-      },
-      "tags": {
-        "amenity": "telephone",
-        "operator": "NTT大阪支店"
-      }
-    },
-    {
       "displayName": "NTT東日本",
       "id": "nippontelegraphandtelephoneeast-a6a893",
       "locationSet": {


momiji-takinogawa:
Even if I remove this entry in NSI, will it revive next time the automated import runs?

Snowysauce:
> Even if I remove this entry in NSI, will it revive next time the automated import runs?

This depends on whether or not there are still objects in OSM with that operator value, and if so, how many there are. The import script has a minimum threshold of 10 instances for operators, and anything at or above that will get an auto-generated preset.

That said, we can use `matchNames` to reassign any automatically read value to other, more desirable presets.

momiji-takinogawa:
Thanks. Then I think I need some on-site survey to retag existing OSM objects.

LaoshuBaby:
> and anything at or above that will get an auto-generated preset.

but looks like we still can prevent it generate preset by add it to a generic name if urgent needed? although this isn't recommended.

(Besides, now how often do we run the nsi-collector?)

LaoshuBaby:
> 大阪支店 (Osaka branch office) of NTT West does not exist anymore, and its former operation area is now covered by 関西支店 (Kansan branch office)[1](#user-content-fn-1-ba94fb4037ca0d2a934b51e6dd27bff6).

Besides I also think there don't need to seperate Kansai branch office as a single entry since NTT West exist as a headquarter of many prefectural level branch ofice, only if we want to tag them all. 🤣

图片

Snowysauce:
> but looks like we still can prevent it generate preset by add it to a generic name if urgent needed?

Correct.

> (Besides, now how often do we run the nsi-collector?)

I've been running it more or less monthly since March. I make a fresh run when OSMF puts out a planet file with data from around the 10th of a given month, which usually happens a week later.

Up until June 2023, the collector was run approximately every other week. Although that interval is great for staying on top of the latest OSM data trends, I just don't have the time to commit to that kind of cycle. Monthly is at least better than what we had between June 2023 and March 2025, when there was only one run during that entire span.

PR #11196, opened 2025-08-03
Add Meilleurtaux to office=financial data

andreadecorte:
Adding French financial service brand with 350 offices as of today (see [official website](https://www.meilleurtaux.com/a-propos-de-meilleurtaux/toutes-nos-agences-en-france/nos-agences-de-courtage.html))
- fx code as only Metropolitan France
- category is aligned with existing competitors like Cafpi and Vousfinancer

### DIFF
--- a/data/brands/office/financial.json
+++ b/data/brands/office/financial.json
@@ -159,6 +159,18 @@
         "office": "financial",
         "operator": "Premista"
       }
+    },
+    {
+      "displayName": "Meilleurtaux",
+      "id": "meilleurtaux-1791d3",
+      "locationSet": {"include": ["fx"]},
+      "matchNames": ["Meilleurtaux.com"],
+      "tags": {
+        "brand": "Meilleurtaux",
+        "brand:wikidata": "Q3304699",
+        "name": "Meilleurtaux",
+        "office": "financial"
+      }
     }
   ]
 }


data/brands/office/financial.json
```diff
@@ -159,6 +159,18 @@
         "office": "financial",
         "operator": "Premista"
       }
+    },
+    {
+      "displayName": "Meilleurtaux",
+      "id": "meilleurtaux-1791d3",
+      "locationSet": {"include": ["fx"]},
+      "matchNames": ["Meilleurtaux.com"],
```
LaoshuBaby, line 167: Usually NSI don't record website and url, add this only if this url was used on board or common appear with its office.

andreadecorte: Yes, this is not the website but it used to be the previous name, then they dropped the .com. It's already present in OSM so for me it makes sense to add it here. One existing example: https://www.openstreetmap.org/node/5987094561

It seems name change is from 2021

https://www.argusdelassurance.com/les-distributeurs/courtiers/courtage-meilleurtaux-com-change-de-nom.192347

ilias52730: Old name is indeed "meilleurtaux.com", so it makes sens to add it to matchnames.

ilias52730:
Looks good to me. Merging PR.

Issue #6247, opened 2022-03-03
NSI should stop asking to replace actual names with `name=YMCA` (preserveTags)


matkoniecz:
properties object needs a preserveTags property

see https://osmus.slack.com/archives/C2VJAJCS0/p1646219577188909 at US Slack

pardthemonster:
I posted this on Slack conversation about topic, but will here too...Any benefit for YMCA now having short_name=YMCA as part of NSI?

PR #11210, opened 2025-08-05
au features where name is commonly "{brand} {branch}"

andrewharvey:
it's common for these features in AU to have a name set something like "{brand} {branch}" in this case iD suggesting to change the name to match the brand is problematic.

See Discord discussion at https://discord.com/channels/413070382636072960/471231032645910529/1401452321841352704

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -5346,7 +5346,6 @@
           "ye"
         ]
       },
-      "preserveTags": ["^name"],
       "tags": {
         "alt_name": "Kentucky Fried Chicken",
         "amenity": "fast_food",
@@ -6342,7 +6341,6 @@
           "ua"
         ]
       },
-      "preserveTags": ["^name"],
       "tags": {
         "amenity": "fast_food",
         "brand": "McDonald's",
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -5346,6 +5346,7 @@
           "ye"
         ]
       },
+      "preserveTags": ["^name"],
       "tags": {
         "alt_name": "Kentucky Fried Chicken",
         "amenity": "fast_food",
@@ -6341,6 +6342,7 @@
           "ua"
         ]
       },
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "fast_food",
         "brand": "McDonald's",
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -725,6 +725,7 @@
       "id": "amp-510934",
       "locationSet": {"include": ["au", "nz"]},
       "matchNames": ["amp bank"],
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "AMP",
@@ -778,6 +779,7 @@
       "id": "anz-510934",
       "locationSet": {"include": ["au", "nz"]},
       "matchNames": ["anz bank"],
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "ANZ",
@@ -2994,6 +2996,7 @@
       "displayName": "Bankwest (Australia)",
       "id": "bankwest-f304bd",
       "locationSet": {"include": ["au"]},
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "Bankwest",
@@ -3769,6 +3772,7 @@
       "displayName": "Bendigo Bank",
       "id": "bendigobank-f304bd",
       "locationSet": {"include": ["au"]},
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "Bendigo Bank",
@@ -4182,6 +4186,7 @@
       "displayName": "BOM",
       "id": "bom-f304bd",
       "locationSet": {"include": ["au"]},
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "BOM",
@@ -4194,6 +4199,7 @@
       "displayName": "BOQ",
       "id": "boq-f304bd",
       "locationSet": {"include": ["au"]},
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "BOQ",
@@ -7738,6 +7744,7 @@
       "displayName": "Heritage Bank (Australia)",
       "id": "heritagebank-f304bd",
       "locationSet": {"include": ["au"]},
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "Heritage Bank",
@@ -9498,6 +9505,7 @@
       "id": "nab-f304bd",
       "locationSet": {"include": ["au"]},
       "matchNames": ["national australia bank"],
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "NAB",
@@ -11009,6 +11017,7 @@
       "locationSet": {
         "include": ["au", "id", "nl", "nz", "us"]
       },
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "Rabobank",
@@ -12500,6 +12509,7 @@
       "id": "stgeorge-f304bd",
       "locationSet": {"include": ["au"]},
       "matchNames": ["st.george bank"],
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "St.George",
@@ -12691,6 +12701,7 @@
       "displayName": "Suncorp",
       "id": "suncorp-f304bd",
       "locationSet": {"include": ["au"]},
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "Suncorp",
@@ -14044,6 +14055,7 @@
       "displayName": "Westpac",
       "id": "westpac-510934",
       "locationSet": {"include": ["au", "nz"]},
+      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "Westpac",
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -725,7 +725,6 @@
       "id": "amp-510934",
       "locationSet": {"include": ["au", "nz"]},
       "matchNames": ["amp bank"],
-      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "AMP",
@@ -779,7 +778,6 @@
       "id": "anz-510934",
       "locationSet": {"include": ["au", "nz"]},
       "matchNames": ["anz bank"],
-      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "ANZ",
@@ -2996,7 +2994,6 @@
       "displayName": "Bankwest (Australia)",
       "id": "bankwest-f304bd",
       "locationSet": {"include": ["au"]},
-      "preserveTags": ["^name"],
       "tags": {
         "amenity": "bank",
         "brand": "Bankwest",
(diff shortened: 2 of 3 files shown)


data/brands/amenity/fast_food.json
```diff
@@ -5346,6 +5346,7 @@
           "ye"
         ]
       },
+      "preserveTags": ["^name"],
```
1ec5: Does KFC normally go by the individual branch locations in all but 24 countries? I don’t think KFC locations in the U.S. even list their individual location names on their receipts. If this is unique to Australia, there needs to be a separate entry for Australia.

andrewharvey: Yeah I wasn't sure how best to apply those to Australia only but thought it could apply globally.

I'm not sure about KFC, for example with McDonalds it would only appear on the receipt, but with bank branches it may be signposted next to the brand logo signage.

I'll change this PR to only cover AU banks and supermarkets. Then can think about fast food later.

1ec5: > Yeah I wasn't sure how best to apply those to Australia only but thought it could apply globally.

You’d duplicate the preset and change the new preset’s `locationSet` array to only include `au`. If the old preset lists `au` in its `locationSet`, remove it; otherwise, if it’s a `001` preset (meaning worldwide), add `au` to the `exclude` list inside `locationSet`.

> I'm not sure about KFC, for example with McDonalds it would only appear on the receipt, but with bank branches it may be signposted next to the brand logo signage.

If the branch name is signposted, that’s a pretty strong case for including it in `name=*`. If it only appears on the receipt and people don’t normally refer to it by the branch name in conversation, then consider leaving the branch name in `branch=*` and the fully qualified name in `official_name=*`. This has often been the outcome of past global discussions on the tagging mailing list, wiki talk pages, and the [forum](https://community.openstreetmap.org/t/restructure-wiki-page-key-name/110631/24).

I can’t see the cited Discord discussion for some reason. If the on-the-ground situation in Australia differs from the general case internationally, that would be worth documenting somewhere beyond NSI, or at least here in this discussion, so that mappers will know what to look for when filling out `name=*`.

andrewharvey: > You’d duplicate the preset and change the new preset’s locationSet array to only include au. If the old preset lists au in its locationSet, remove it; otherwise, if it’s a 001 preset (meaning worldwide), add au to the exclude list inside locationSet.

Yeah.

> If the branch name is signposted, that’s a pretty strong case for including it in name=*. If it only appears on the receipt and people don’t normally refer to it by the branch name in conversation, then consider leaving the branch name in branch=* and the fully qualified name in official_name=*. This has often been the outcome of past global discussions on the tagging mailing list, wiki talk pages, and the [forum](https://community.openstreetmap.org/t/restructure-wiki-page-key-name/110631/24).

Sorry I don't keep up with all the forum discussions, in my view name="{brand} {branch}" is a more accurate representation of the "on the ground" name in that even if it the branch only appears on receipts and not signposted.

I realise that with `brand=*` and `branch=*` tagged applications can discard the `name=*` tag and use whatever format they like, but as a default `name={brand} {branch}` is better. If I search for nearby "{brand}" in Nominatim, a list of names of form `{brand} {branch} is most useful. Google Maps will use that format too.

Regardless it seems this isn't the widely accepted view globally so since it's usually only banks where the branch appears in signage (and this was the example raised in Discord that spurred this PR) I'll update this PR again to only change the banks.

bhousel: I'm kind of against making the banks work differently in Australia just because someone doesn't like the `branch` tag that has become the convention almost everywhere else. 

Are you sure there is no way we can find a solution where the Australian community is happy with the current behavior?

data/brands/amenity/fast_food.json
```diff
@@ -6341,6 +6342,7 @@
           "ua"
         ]
       },
+      "preserveTags": ["^name"],
```
bhousel: Yeah we're obviously not going to do this.

data/brands/amenity/fast_food.json
andrewharvey: I do like the `branch` tag and always tag it where possible. `brand` and `branch` allow you to specify each part separately allowing a more accurate representation and providing more flexibility for downstream data users to choose what to show.

I guess I see the name tag as being for the full or complete or fully qualified name of the feature and in that case it's often something like "{brand} {branch}".

iD should make suggestions in line with what's expected/accepted by the community, we don't want iD suggesting to change X->Y but then users who do so have edits reverted back to X and those reverts are upheld by the community consensus.

In this case it seems most prefer name to be the "{brand}" or what's on the storefront signage. Going with storefront signage means you'll get variation's one brand might do it, another brand might not.

The example given on Discord was this where it's signposted with the brand/logo "Bendigo Bank" on the left, then "Community Bank" on the right with the branch name "Kalbar" below.

image

I suggested we go with `brand=Bendigo Bank` + `branch=Community Bank Kalbar` or `branch=Kalbar`, then maybe just `name=Bendigo Community Bank Kalbar` which had some support from other mappers but iD suggests to change the name back to "Bendigo Bank".

1ec5: Totally understand that the local community doesn’t want to lose information about the branch name or full name that someone has gone through the trouble of collecting. Ideally, actual real-world usage would be the difference between `name=*` and `official_name=*` for a fully composed name. The “Kalbar” on the sign does make it more likely that a layperson would know the branch name. We already have some `preserveTags` where the signs can look similar, like YMCA in the U.S. (#6247) and hotels and car dealerships in general. Allowing the `name=*` to be “Bendigo Community Bank Kalbar” means we don’t expect most people to be familiar with it as simply “the Bendigo Bank in Kalbar” (based on the monument sign).

bhousel:
I hate this and I think it's the wrong thing to do.

I did about 10 minutes of research and Google and Apple Maps don't show these bank brands this way.  
It's always just labelled "Westpac", "Bendigo Bank", "Suncorp".

But anyway, I'm merging it because it's what someone wants.  😞 

> iD should make suggestions in line with what's expected/accepted by the community, we don't want iD suggesting to change X->Y but then users who do so have edits reverted back to X and those reverts are upheld by the community consensus.

This is classic "weaponized language of community" that you'll see everywhere in OSM spaces now.  Someone sees a thing they don't like - every time it's a small thing that doesn't matter much in the big picture - but they frame it as "you're going against the community" just to get what they want.  (#10158 contains a lot of this too).   Congrats @andrewharvey , you win.

1ec5:
> (https://github.com/osmlab/name-suggestion-index/issues/10158 contains a lot of this too)

(By way of an update, [the discussion on the forum](https://community.openstreetmap.org/t/figurine-tagging/132824) finally went in a different direction. They’re now considering a new key that won’t need to involve NSI.)

bhousel:
#5500 is like this too, where we basically needed to add a bunch of special items _just for Norway_ because they want their names to look a certain way that is different from the guidance on the OSM wiki. 

I think it's the wrong thing to do, and it makes POI data less consistent, but the software does support it and we try to give people what they want.

1ec5:
In my own local community, we used to put the fully qualified name in `name=*` but eventually migrated to `official_name=*` when we started taking our POI coverage more seriously, as something ordinary people would use day to day. I think the main motivation for using `name=*` originally was just a fear of losing out on this information if the mapper isn’t prompted for it. `branch=*` and `official_name=*` aren’t visible by default in most POI presets coming from id-tagging-schema.

Did you mention somewhere that NSI can generate presets that add their own fields, or was I hallucinating? If so, maybe for certain entries we could generate a preset that shows the `branch=*` and `official_name=*` fields by default instead of relegating them to the `moreFields` menu. These keys don’t get locked by default, so there’s no need for a `preserveTags` exception.

Secondarily, people have brought up the inconvenience of seeing [a screen full of “McDonald’s” search results](https://community.openstreetmap.org/t/restructure-wiki-page-key-name/110631/28) without any further context. But that’s actually a UI bug, particularly osmandapp/OsmAnd#2993, which finally got fixed about a year ago. Maybe some of these regional consensuses need to be revisited in light of changing data consumption trends. When I’ve argued for `preserveNames` exceptions in the past, it’s been based on [how ordinary people use these POIs in the real world](https://community.openstreetmap.org/t/name-tag-on-chain-hotels/131973/15) – not just a matter of taste, an urge for completeness, or a pedantic rule about what’s on the sign.

bhousel:
> In my own local community, we used to put the fully qualified name in `name=*` but eventually migrated to `official_name=*` when we started taking our POI coverage more seriously, as something ordinary people would use day to day. I think the main motivation for using `name=*` originally was just a fear of losing out on this information if the mapper isn’t prompted for it. `branch=*` and `official_name=*` aren’t visible by default in most POI presets coming from id-tagging-schema.

I think we mostly agree on this.  The `name` field should be a reasonable thing to show in a search result or on a map pin or read out loud by a screen reader or as a waypoint in a navigation app, etc..  It doesn't need to be everything for everyone, and there are other fields better suited for more detailed information about the branch location.   Unsophisticated map apps already use `name` for all these things.  (There were a few things about this PR that I didn't like, but the sneaky attempt to modify the main McDonald's preset definitely put my guard up).

> Secondarily, people have brought up the inconvenience of seeing [a screen full of “McDonald’s” search results](https://community.openstreetmap.org/t/restructure-wiki-page-key-name/110631/28) without any further context.

I dunno, Apple Maps handles this fine.  It sounds like OSM needs to get with the times.

Screenshot 2025-08-08 at 10 00 16 AM

> Did you mention somewhere that NSI can generate presets that add their own fields, or was I hallucinating? If so, maybe for certain entries we could generate a preset that shows the branch=* and official_name=* fields by default instead of relegating them to the moreFields menu. These keys don’t get locked by default, so there’s no need for a preserveTags exception.

The iD presets file that we generate can contain "fields", "presets", "categories", and "defaults", so it is allowed to contain fields too. (As you know, I'm the person who previously, _and still should be,_ the maintainer of iD and its presets, so I'm very familiar with how this works). 

Rapid currently uses this feature to override the default iD crossing presets and fields with better ones, so I know it works there.   That said, I think the iD presets currently contain too much stuff, so adding more stuff to what is displayed there would be a step in the wrong direction.  I guess we could also use this feature to remove some of the fields too, but really this whole UI component could use a redesign. It fills too much space with details that aren't important.

Screenshot 2025-08-08 at 10 15 16 AM

Apple maps is _not an editor_ but they do a better job of getting users to capture details about a place.  The screen space is better used and the important details are easy to find. 

Screenshot 2025-08-08 at 10 25 57 AM

Cj-Malone:
IDK why I'm commenting, but osm.org doesn’t show a full screen of McDonald's, it may not be pretty, but it is better than it used to be.

[image](https://www.openstreetmap.org/search?query=McDonald%27s%2C+Baltimore&zoom=7&minlon=-87.02270507812501&minlat=31.737511125687828&maxlon=-69.77416992187501&maxlat=39.816975090490004#map=12/39.2958/-76.5781)

I also made [this](https://cj-malone.github.io/dinosaur-juice/list.html) a while ago to try and show how logos and `branch` could be used to make lists more interesting.

1ec5:
Yeah, it was really just OsmAnd that insisted on showing exactly the same thing a bunch of times in a list without a map, but that’s fixed now. A restaurant chain like McDonald’s should really be a no-brainer. It’s the poster child of corporate brand uniformity. I think this will all become clearer once renderers start doing obnoxious things like plastering brand logos everywhere: osm-americana/openstreetmap-americana#809. 😉

andrewharvey:
>  The name field should be a reasonable thing to show in a search result or on a map pin or read out loud by a screen reader or as a waypoint in a navigation app, etc.. It doesn't need to be everything for everyone, and there are other fields better suited for more detailed information about the branch location.

I guess to me I'd expect search results, labels on a map pin, tts output and waypoint names in a navigation app to include the branch in the name, to provide a more specific result. eg. so I can tell which "McDonalds" location is being referenced and not just a generic "McDonalds" or list of "McDonalds" without any details about the branch names.

>  A restaurant chain like McDonald’s should really be a no-brainer. It’s the poster child of corporate brand uniformity

The brand may be uniform, but the branches are usually signposted on the front entrance door, listed on the receipt, shown on the website store locator.

PR #11217, opened 2025-08-06
fix Komunálna poisťovňa name

Oivo35:
Changed spelling according to https://wiki.openstreetmap.org/wiki/Names#Proper_name_spelling.

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -854,13 +854,13 @@
       }
     },
     {
-      "displayName": "ID ŽK (Slovensko)",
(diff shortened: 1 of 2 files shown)


data/transit/route/train.json
```diff
@@ -854,13 +854,13 @@
       }
     },
     {
-      "displayName": "ID ŽK (Slovensko)",
-      "id": "integrovanadopravazilinskehokraja-b6707b",
```
LaoshuBaby, line 858: Thank you but you don't need to edit manually because it will be updated when publish a new version

PR #11218, opened 2025-08-06
add transit network IDS Plus

Oivo35:


Snowysauce:
Based on #11217, I'm not seeing why "Plus" is in all-caps.

PR #11227, opened 2025-08-08
Add Claudio

Crashillo:
Spanish supermarket chain

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -1542,6 +1542,16 @@
       }
     },
     {
+      "displayName": "Claudio",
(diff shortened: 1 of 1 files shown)


ilias52730:
Hi Hugoren! You added Portugal as a location, however I can only find that Claudio supermarkets are present in Galicia, Castilla y León, Madrid, and Asturias. Do you have more info?

Crashillo:
Nope, nice catch! I saw this map in https://www.claudio.es/localizador, and I thought in Portugal were present also :sweat_smile:
I will update it
imagen

ilias52730:
I also had the same thought, and then I zoomed in and found out these Claudios are really at the border with Portugal.

Merged PR!

PR #11228, opened 2025-08-09
Added Dogs Trust charity

TheUKHighStreet:
32 shops.

### DIFF
--- a/data/brands/shop/charity.json
+++ b/data/brands/shop/charity.json
@@ -16,6 +16,16 @@
       }
     },
     {
+      "displayName": "Dogs Trust",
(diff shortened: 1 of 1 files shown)


data/brands/shop/charity.json
```diff
@@ -15,6 +15,16 @@
         "shop": "charity"
       }
     },
+    {
+      "displayName": "Dogs Trust",
+      "locationSet": {"include": ["gb"]},
```
codeinabox: Thank you for this! Looking at the shop map, they only have locations in England and Wales

```suggestion
      "locationSet": {"include": ["gb-eng", "gb-wls"]},
```

PR #11235, opened 2025-08-11
Updating Bosch eBike Systems to only be in Europe

GA-Kevin-Codes:
According to Overpass, the only stations of these exist in Europe, code 150.

https://overpass-turbo.eu/s/29Io

### DIFF
--- a/data/brands/amenity/charging_station.json
+++ b/data/brands/amenity/charging_station.json
@@ -114,7 +114,7 @@
       "displayName": "Bosch eBike Systems",
       "id": "boschebikesystems-67e574",
       "locationSet": {
-        "include": ["001"],
+        "include": ["150"],
         "exclude": ["no"]
       },
       "matchNames": ["robert bosch gmbh"],


LaoshuBaby:
> According to Overpass

Is there any first-hand reference from Bosch that said the system only in Europe?

(Anyway, thank you for your research on EV energy)

GA-Kevin-Codes:
To be honest, I can't find a map or list of these stations anywhere so I turned to searching OSM for it where I found it to only be in Europe. Instead, should I just exclude "us"? Happy to do what's best for NSI :)

LaoshuBaby:
> To be honest, I can't find a map or list of these stations anywhere so I turned to searching OSM

That's very hard. But I guess for those kind of situation we shouldn't assume that it won't appear in US? (This is not like NIO in #11239 because bosch isn't from a "evil communist country", XD)

(I exclude China at some brand's presets years ago because I'm sure they won't appear in Chian, and Norway community also open a PR ask to exclude them from many presets, if you insist to so I think it is also acceptable)

GA-Kevin-Codes:
Considering this is not mapped at all in the US, as well as the inability to verify any locations digitally, I don't think this raises to the level of inclusion in the NSI for the US. I can certainly periodically check to see if the situation changes.

UKChris-osm:
I'd only worry about removing the US locationSet if there is another brand currently in the US that's clashing with this one.

GA-Kevin-Codes:
Fair enough, was just trying to do some cleanup of the EV elements in the US, but I can see the appeal in leaving it in. The charging station wiki page states it implies `motorcar=yes`, for items like this should the NSI specify `motorcar=no` + `bicycle=designated`? Thanks in advance @UKChris-osm!

PR #11236, opened 2025-08-12
Add Hapik - Update sports_centre.json

versgui:


GA-Kevin-Codes:
Hi @versgui! It seems from the [Hapik website](https://hapik.fr/centres-tarifs) there are 19 locations, [Wikidata P8368](https://www.wikidata.org/wiki/Q135729937) states there are 19 as well. Typically, our [minimum to be included in the NSI](https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#notability) is 30, so I will mark this as needs discussion and notability for the time being so we can keep an eye on things.

versgui:
Hello @GA-Kevin-Codes, I didn't know this rule, thanks for the information.

UKChris-osm:
They appear to also trade in the US and Germany, and in the UK but under a different name (Rock Up) so I think this should be added.

PR #11237, opened 2025-08-12
Add Volfoni - Update restaurant.json

versgui:


ilias52730:
Extra info: 35 restaurants in Metropolitan France and 1 in La Réunion according to their [website](https://www.volfoni.fr/nos-restaurants).

Merging PR.

PR #11239, opened 2025-08-12
Updating NIO EV Infrastructure locationSet

GA-Kevin-Codes:
NIO primarily sells in China, which their brand websites show, so I have removed "exclude" cn. In addition, they are not present in the US due to political decisions and economic tariffs so I have added "us" to "exclude".

https://www.onvo.cn/map

https://chargermap.eu.nio.com/pe/h5/static/chargermap?channel=official

### DIFF
--- a/data/brands/amenity/charging_station.json
+++ b/data/brands/amenity/charging_station.json
@@ -643,7 +643,7 @@
       "id": "niopowercharger-9dbcd7",
       "locationSet": {
         "include": ["001"],
-        "exclude": ["us"]
+        "exclude": ["cn", "us"]
       },
       "tags": {
         "amenity": "charging_station",
@@ -661,7 +661,7 @@
       "id": "niopowerdestination-9dbcd7",
       "locationSet": {
         "include": ["001"],
-        "exclude": ["us"]
+        "exclude": ["cn", "us"]
       },
       "tags": {
         "amenity": "charging_station",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Thank you for notice this but indeed for China brand we usually separate a China (some times Mainland) only version because they are using non-latin character and contain multilanguage tag. 

And those preset have Chinese version recorded in NSI ("蔚来"):

image
https://nsi.guide/index.html?t=brands&k=amenity&v=charging_station&tt=%E8%94%9A%E6%9D%A5

those preset are using the same wikidata as modified.

They have a data of those power station on their Mainland of China market's official website

image
https://www.nio.cn/official-map

To avoid the global preset been suggested in China I think we still need to keep "cn" in exclude list.

---

For NIO in oversea, I only found a map that contain their power station in Europe but I think they have business not only in above countries, so use 001 is still suitable unless we found a exact list. (Because they may landed to new country quickly)

image
https://chargermap.eu.nio.com/pe/h5/static/chargermap?channel=official

(So I guess this map is asked by EU regulation?)

For US, because of the reason everyone know there truly haven't any Chinese vehicle brand, so I think add "us" to exclude list is not a bad idea, but we may need to edit it in future.

LaoshuBaby (changes requested):
Removing "cn" will allow Chinese users to see two presets, one "NIO" and the other "NIO"

GA-Kevin-Codes:
Thanks @LaoshuBaby for clarifying that for me, I've added back "cn" to the exclude and kept the change as adding "us" to the exclude list as agreed :)

PR #11245, opened 2025-08-13
Update pet.json

Patitasco:
Adding Patitas&Co, an online pet shop that delivers dog and cat food and accessories across Europe.


GA-Kevin-Codes:
Hi there. The format indent seems off, the Wikidata ID isn't valid, and it seems there is only one location on the website, so this would not reach the level of inclusion in NSI.

PR #11252, opened 2025-08-17
Update bicycle_rental.json (Modified Stadtrad Hamburg)

glawle:
Modified Stadtrad Hamburg according to the wiki.

### DIFF
--- a/data/brands/amenity/bicycle_rental.json
+++ b/data/brands/amenity/bicycle_rental.json
@@ -3698,14 +3698,11 @@
         "amenity": "bicycle_rental",
         "brand": "Stadtrad Hamburg",
         "brand:wikidata": "Q2326366",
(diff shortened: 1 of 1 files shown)


GA-Kevin-Codes:
This seems like a lot of duplicate information. Why not just link the Wikipedia article to the wikidata item and include the brand wikidata here on NSI?

glawle:
Thanks for the tip and for your time. Fixed PR. I did some extra research, first time editing bike rental network on OSM. I will start a discussion about the general tagging standard of Stadtrad Hamburg in the OSM wiki. I just found issue [#6481](https://github.com/osmlab/name-suggestion-index/issues/6481) and wanted to know if there are any other restrictions on allowed tags? There might be a need for an NSI wiki article because the wikipedia tags are widely used for the direct connection besides it might be a duplicate information. In theory, a website link is also duplicate info when you have wikidata.

Issue #4768, opened 2020-12-04
Road Routes 🛣


bhousel:
Whenever I get some free time, I'll add the road routes to the index.  We are already collecting the `network` tag values from osm planet.
https://wiki.openstreetmap.org/wiki/Relation:route#Road_routes

The eventual goal is to have every road route in the world tagged with a `network` and `network:wikidata` so that software can link the routes to Wikidata, e.g. to display route shields.

1ec5:
To my knowledge, `network` on road route relations is seldom used in conjunction with Wikidata, unlike `network` on public transportation routes. I think this is because `network` on road route relations has been pretty well-organized and systematized by comparison, as opposed to accepting freeform keys. (Same goes with `cycle_network` for bicycle routes.) So this is the one circumstance where Wikidata and OSM Wiki data items already capably perform the function that this ticket proposes for NSI. For example, [this Wikidata Query Service query](https://w.wiki/p5H) returns a correspondence between `network` values and Wikidata items (and associated metadata like ref formats). It isn’t quite comprehensive: the OSM Wiki has some more values for networks that aren’t in Wikidata yet. Unfortunately, Sophox is down (Sophox/sophox#24), so I can’t point to a convenient query that includes even undocumented networks. But those networks aren’t quite ready for NSI anyways.

It is possible to [directly query the OSM Wiki](https://wiki.openstreetmap.org/wiki/Special:ApiSandbox#action=wbsearchentities&format=json&search=network%3D&language=en&strictlanguage=1&limit=max) for these data items, but you’d have to do another API call to get the Wikidata QIDs claimed by each one. If it’s desirable for NSI to mirror the OSM Wiki and Wikidata this way, then that’s perfectly fine because it makes the data more immediately usable by editors and other tools. But I don’t think we’d want to encourage people to hand-author any additions to the mirror, which would get out of sync with two different wikis that are already hand-authorable.

1ec5:
For better or worse, some mappers have also added road routes to [`network`](https://wiki.openstreetmap.org/wiki/Relation:network) relations, which means NSI would only need to query the Overpass API to generate a list of these networks at build time.

1ec5:
[...]

PR #11261, opened 2025-08-21
Add Junta de Andalucía as operator of a route=road

Robot8A:


bhousel:
Thanks for the PR, but road routes isn't really the sort of thing we track in NSI yet.

We had some discussion on #4768  about adding road routes, but this was more in the context of supporting `wikidata` and highway shields.

Issue #10411, opened 2025-01-29
Improving multilingualism


Cj-Malone:
NSI is already improving multilingualism in OSM through split entries for different countries or regions and having multiple `brand:XX` values available. However I think there is so much more to be done and NSI is in a key place to start to some bigger improvements.

Why should a Japanese person visiting the US or even just browsing an OSM map not see `マクドナルド` instead of `McDonald's`?

NSI for US (And most of the world) gives us this:

```
brand=McDonald's
brand:wikidata=Q38076
```
But imagine if it was:

```
brand=McDonald's
brand:en=McDonald's
brand:ko=맥도날드
brand:ja=マクドナルド
brand:zh-Hant=麥當勞
brand:zh-Hans=麦当劳
brand:ar=ماكدونالدز
brand:wikidata=Q38076
```

Snowysauce:
I feel like this and #10410 are discussions that should be had with the OSM community at-large. As far as I'm aware, what the NSI is doing in both situations reflects how the OSM community expects the presets in question to be/how the OSM community expects the objects to be tagged. Some of that practice may be because of the NSI, but I also think that if the NSI was really out of step with the community's preferred way of doing things, we'd have heard about it by now.

1ec5:
Data consumers should be using `brand:wikidata=*`. `brand=*` is just a convenience for fellow mappers who don’t want to hop over to Wikidata or install [a browser extension](https://community.openstreetmap.org/t/announcing-human-readable-wikidata-browser-plugins-for-openstreetmap-org/108180) when verifying that `brand:wikidata=*` is correct. It’s a compatibility shim. NSI is sort of localizing `brand=*` because it doesn’t really serve that purpose if local mappers don’t understand it.

Cj-Malone:
> Data consumers should be using `brand:wikidata=*`.

Totally agree, I use it for brand contact info, and OSM for POI specific contact info.

`brand=*` is just a convenience for fellow mappers who don’t want to hop over to Wikidata or install [a browser extension](https://community.openstreetmap.org/t/announcing-human-readable-wikidata-browser-plugins-for-openstreetmap-org/108180) when verifying that `brand:wikidata=*` is correct. It’s a compatibility shim.

Hard no. `brand` is not deprecated or a compatibility shim.

I love `brand:wikidata`, I just called out Vespucci for not using it. But it is not an alternative to `brand` or `brand:XX`.

[...]

PR #11272, opened 2025-08-24
Add Mongolian name & brand for Starbucks (China)

OverflowCat:


LaoshuBaby (changes requested):
I don't think we need to add Mongolian name for Mainland of China

1. This is a nationwide preset and if you add this tag, it will apply to all user in China, not only Inner Mongolian AR.
2. I remember (and I've lived in China for many years) that Starbucks only use its Latin spelling name and “星巴克” outside, because OSM based on ground fact, I think you need to provide some photo or report about adopting to bilingual of Chinese and Mongolian.

---

In Chinese:

牢猫你搁这儿整活呢？不是你要推广小语种也不是这么来的啊？而且上面还mn下面就mn-Mong拜托至少统一一下都用mn-Mong喽？（流汗黄豆）

OverflowCat:
1. Is there such an locationSet for that? Also, I think this is a
#10411 problem.
2. There is no usage of brand:mn-Mong, only brand:mn. I originally wanted
to add both *:mn & *:mn-Mong. As for whether to use mn / mn-*, I think it
is the same with zh and zh-*.
3. (Xilinhot) https://www.nmgsb.com.cn/system/mshang/2023/0621910R023.html
and (Hulunbuir) https://www.sohu.com/a/474236358_121123912

Am Mo., 25. Aug. 2025 um 06:32 Uhr schrieb 快乐的老鼠宝宝 ***@***.***

bhousel:
Closing for now - it sounds like it doesn't make sense to add this multilingual name tag here.
I agree #10411 is worth thinking about though.  OSM's tagging model doesn't make this easy. 
We are kind of stuck with either - sticking a lot of extra name tags on every item in the world - or we need to encourage renderers to look up multilingual names another way.

PR #11279, opened 2025-08-25
[MA] Updated Marjane Market and Marjane

ilias52730:


ilias52730:
@Cj-Malone I see you reacted with a 👎. I forgot to add a description, my mistake:

Marjane and Marjane Market are different brands in Morocco, just like Carrefour and Carrefour Market. Marjane is the hypermarket brand and Marjane Market is the supermarket brand. Hypermarkets are still tagged as `shop=supermarket` on OSM so I don't have a choice but to tag both as such.

Issue #11264, opened 2025-08-22
Various warnings generated by the build script


Snowysauce:
Warnings generated by a run of the build script on 2025-08-21:

```
⚠️   Warning - matchIndex errors:
----------------------------------------------------------------------
  `key/value/name` occurs multiple times in the match index.
  To resolve these, make sure the key/value/name does not appear in multiple trees
    (e.g. `amenity/post_office/ups` should not be both a "brand" and an "operator"
----------------------------------------------------------------------
Duplicate cache key "man_made/charge_point/q105883058" in trees "operators" and "brands", check items: shellrechargesolutions-7872f6
Duplicate cache key "man_made/charge_point/shellrecharge" in trees "operators" and "brands", check items: shellrecharge-9b2e7d,shellrechargesolutions-7872f6
Duplicate cache key "man_made/charge_point/volta" in trees "operators" and "brands", check items: volta-416623
Duplicate cache key "amenity/charging_station/q105883058" in trees "operators" and "brands", check items: shellrechargesolutions-c22d3a
Duplicate cache key "amenity/charging_station/shellrecharge" in trees "operators" and "brands", check items: shellrecharge-67e574,shellrechargesolutions-c22d3a
Duplicate cache key "amenity/charging_station/volta" in trees "operators" and "brands", check items: volta-994ed4
total 6
```

```
⚠️   Warning - Missing tag:
----------------------------------------------------------------------
  To resolve these, add the missing tag.
----------------------------------------------------------------------
  "Modern Nails (modernnails-0c474f)" -> missing tag? -> "beauty"
  "Aurora (aurora-2fabb0)" -> missing tag? -> "cuisine"
  "Normandin (normandin-71aa4c)" -> missing tag? -> "cuisine"
total 3
```

GA-Kevin-Codes:
For the Charging Station ones, Shell bought Ubitricity and Volta and has been folding them into their wider Shell Recharge brand. So while Shell doesn't directly operate all of these stations due to past agreements, they are largely a brand/operator. Much like many high-power charging station brands.

Snowysauce:
The NSI is currently designed to allow for only one use of a QID or name across all category files with a specific key-value combination. For `amenity=charging_station`, this means that entities have to be classified as a "brand" or "operator" - in the eyes of the NSI, they cannot be both. @bhousel would be the one best suited to explain why this setup exists.

UKChris-osm:
[...]

PR #11282, opened 2025-08-26
Removing Shell Recharge from Operators in favor of Brand

GA-Kevin-Codes:
Fixes part of #11264 related to EV charging.


Cj-Malone:
Shell Recharge Solutions was added back automatically, it probably needs to be added to the Shell Recharge match names.

PR #11287, opened 2025-08-26
Set correct name for VAG Nürnberg operator

besentv:
See: https://www.wikidata.org/wiki/Q1557682

### DIFF
--- a/data/transit/route/subway.json
+++ b/data/transit/route/subway.json
@@ -953,7 +953,7 @@
         "network": "Verkehrsverbund Großraum Nürnberg",
         "network:short": "VGN",
         "network:wikidata": "Q2516463",
-        "operator": "Verkehrs-Aktiengesellschaft Nürnberg",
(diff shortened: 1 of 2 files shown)


besentv:
Right, additionally the name contains the city's name to not confuse it with https://www.vag-freiburg.de

ilias52730:
VAG seems to be the abbreviation of Verkehrs-Aktiengesellschaft, and then Nürnberg is the city, just like VAG Freiburg. It seems redundant to have VAG followed by Verkehrs-Aktiengesellschaft Nürnberg, as VAG has exactly the same meaning as Verkehrs-Aktiengesellschaft.

It's like having SNCF Société Nationale des Chemins de Fer.

Is this how it's commonly referred to in Germany? VAG Verkehrs-Aktiengesellschaft?

besentv:
> It seems redundant to have VAG followed by Verkehrs-Aktiengesellschaft Nürnberg

This doesn't matter, they call themselves redundant like this.

PR #11288, opened 2025-08-26
[MA] Added matchname to Bank Al-Maghrib

ilias52730:


ilias52730:
@Cj-Malone I see you reacted with a 👎. My mistake for not providing a description (even though it seems pretty intuitive to me):

Here I added a matchname, because when people type "bank al maghrib", the bank doesn't appear. One had to type bank al-maghrib, and no one really adds a hyphen when searching for a brand.

Cj-Malone:
`bank al maghrib` can't be a machname because it is the same (post normalization) as `brand:fr`.

I am also very :-1: mapping for the renderer by putting both languages in `brand` and `name`, but that isn't related to this.

ilias52730:
> bank al maghrib can't be a machname because it is the same (post normalization) as brand:fr.

Here's a screenshot of me typing "bank al maghrib" in Morocco:
image

Here's a screenshot of me typing "bank al-maghrib" in Morocco:
image

I don't know what you mean by "it is the same (post normalization) as brand:fr", but typing "bank al maghrib" doesn't display it, so adding "bank al maghrib" as a matchname will make it appear.

> I am also very 👎 mapping for the renderer by putting both languages in brand and name, but that isn't related to this.

I'm simply following the guidelines of my country, which themselves simply follow the [Map what's on the ground](https://wiki.openstreetmap.org/wiki/Good_practice#Map_what's_on_the_ground) principle, so it's not really mapping for the renderer but mapping what's actually displayed on signs on the ground.
image

bhousel:
Oh yeah, I guess multiple languages in the `name` tag is actually a thing in Morocco.
@ilias52730 I think NSI can try to respect this convention, just please understand that it's an exception.  Most communities in OSM prefer only a single name in the `name` tag and that value should match one of the other language name tags.

Issue #2245 and #6788 have some more on this.
I think the "multiple names in name tag" convention is also used in Hong Kong, Macau, and Belgium.  It can generate warnings in Osmose, KeepRight, and JOSM.

Screenshot 2025-09-02 at 11 52 53 AM

ilias52730:
> Most communities in OSM prefer only a single name in the `name` tag

I completely understand @bhousel, and thank you for respecting this convention. I would also prefer to make things simpler, but unfortunately both French and Arabic have around the same level of importance here. It's normal that people speak 2 or 3 languages minimum. Arabic and Tamazight (which is used less, especially in cities) are the official languages. French is used literally everywhere.

It's mostly the big, old, local brands that use both languages (such as banks or supermarket chains). If we only put their French name in `name`, we neglect the official language people speak. If we only put their Arabic or Tamazight name in `name`, a person looking at a map won't understand anything because nearly everything is also written in French, and some locals speak more French than Arabic. It has nothing to do with tagging for the renderer. I wish I could make things simpler, but for now, I can't.

When a new place opens, they very probably only display a sign in French. I notice a slow and steady change towards only using French names for brands recently in Morocco, and I reflect this on the NSI everytime a brand does. Maybe this is a step towards the right direction.

Thanks for sharing an image of Casablanca, where I edit and added most of the place names you see.

PR #11300, opened 2025-08-28
Update car.json

jhutton1:
update countries and names for Toyota and Lexus to local languages.

### DIFF
--- a/data/brands/shop/car.json
+++ b/data/brands/shop/car.json
@@ -13,6 +13,62 @@
   },
   "items": [
     {
+      "displayName": "凌志",
+      "locationSet": {"include": ["hk"]},
+      "tags": {
+        "brand": "凌志",
+        "brand:en": "Lexus",
+        "brand:zh-Hant": "凌志",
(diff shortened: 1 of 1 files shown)


ilias52730:
Lexus store in Hong Kong:
image
Toyota store in Hong Kong:
image
Toyota store in Taiwan:
image
Toyota store in China:
image


I don't see a reason to split all these entries if the stores are the same all over the world. We [map what's on the ground](https://wiki.openstreetmap.org/wiki/Ground_truth).

bhousel:
> I don't see a reason to split all these entries if the stores are the same all over the world. We [map what's on the ground](https://wiki.openstreetmap.org/wiki/Ground_truth).

Some local communities want to have name tags in multiple languages and some don't want this.
There are a bunch of old issues in this project where we've tried to give people what they want regarding localized names.

Anyway I guess what I'm saying is - don't just go changing stuff because of some rule you read on the OSM wiki.  We probably have separate Toyota entries for the CJK languages because people asked us to do this for good reasons.

PR #11320, opened 2025-09-01
Add ten more Quebec school boards.

waptaff:
Lots more to do.

### DIFF
--- a/data/operators/amenity/school.json
+++ b/data/operators/amenity/school.json
@@ -10755,6 +10755,22 @@
       }
     },
     {
+      "displayName": "Centre de services scolaire de l'Énergie",
+      "id": "centredeservicesscolairedelenergie-d1e97c",
+      "locationSet": {
+        "include": ["ca-qc.geojson"]
+      },
+      "matchNames": [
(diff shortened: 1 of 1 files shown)


UKChris-osm:
How many schools does each board represent? Are there lots of different locations to be mapped for each school board?

waptaff:
Number of schools highly depends on the board.

From the commit, Laval has about 70, Énergie, Val-de-Cerfs, Haute-Rivière and Marie-Victorin have about 45 each, Saint-Hyacinthe about 40, Laurentides and Sommets about 35 each, Découvreurs about 25, Littoral about 10.

There are about 75 total school boards in Québec, managing around 3000 total schools, so the average count per school board is around 40.

One thing of note: in 2020, most of the school boards in Québec were recreated under new names, moving from “Commission scolaire X” to “Centre de services scolaire X”, so having the new ones in NSI help find and fix schools still tagged with the old operator names in OpenStreetMap.

EvanCarroll:
* Wikidata links check out
* ids are unique
* matchnames present in the OSM set, seem good to me.

PR #11324, opened 2025-09-02
Update fuel.json for new brand Gazprom Neft

katslodz:
Update fuel.json for the third largest oil producer in Russia and a subsidiary of Gazprom

Adding new brand for amenity=fuel called Gazprom Neft

https://www.wikidata.org/wiki/Q1461799
https://www.gazprom-neft.ru/
https://www.gazprom-neft.com/

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -8960,6 +8960,8 @@
    },
    "tags": {
        "brand": "Газпром Нефть",
+       "brand:ru": "Газпром Нефть",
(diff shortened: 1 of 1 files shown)


ilias52730:
It seems they only display their Russian name:
image

So IMO:
```
name=Газпром Нефть
name:ru=Газпром Нефть
name:en=Gazprom Neft
```

katslodz:
I've made changes in accordance with your feedback. Thank you for this suggestion!

PR #11325, opened 2025-09-02
Update tyres.json for new brand Michelin

katslodz:
Update tyres.json for the French multinational tyre manufacturing company

Adding new brand for shop=tyres called Michelin

https://www.wikidata.org/wiki/Q151107
https://www.michelin.com/

### DIFF
--- a/data/brands/shop/tyres.json
+++ b/data/brands/shop/tyres.json
@@ -677,6 +677,19 @@
         "name:ko": "타이어뱅크",
         "shop": "tyres"
       }
-    }
+    },{
+   "displayName": "Michelin",
+   "locationSet": {
+       "include": [
+           "001"
+       ]
+   },
+   "tags": {
+       "brand": "Michelin",
+       "brand:wikidata": "Q151107",
+       "name": "Michelin", 
+       "shop": "tyres"
+   }
+}
   ]
 }


Cj-Malone:
Have you got any examples of branded locations?

katslodz:
Thank you for your question regarding branded locations. 
I was following the example of 'Continental AG,' which is already included in the NSI shop=tyres. 
Similar to Michelin, Continental sells their tires through distributors rather than operating branded locations, with branded Logo is clearly visible on the storefronts.
Michelin's globally established reputation makes it a fitting addition to the index, just like Continental. Thank you for considering this perspective!

ilias52730:
Continental should be deleted and Michelin shouldn't be added, IMO. Neither Continental nor Michelin operate tire shops. As you mentioned, they sell them through distributors. This index is for branded locations.

[Michelin's store locator](https://www.michelinman.com/auto/dealer-locator) is called "dealer locator".
Similarly, [Continental's store locator](https://continentaltire.com/Store-finder) lists distributors such as Mavis Tires & Brakes, B & B Tire & Wheel, etc.

katslodz:
All right, I fully agree. 
I will avoid such situations in the future and will only check branded locations carefully. Thank you for the detailed explanation of this case. I am closing the Pull Request for Michelin.

Best regards.

katslodz:
FYI: 
In accordance with your feedback, I am attaching a PR to remove the Continental brand, as it does not have branded locations.
https://github.com/osmlab/name-suggestion-index/pull/11341
Kind Regards

PR #11339, opened 2025-09-03
Fix Harris County

EvanCarroll:
Add/fix support for community centers and parks


EvanCarroll:
There is nothing different here. It's just more of the same. We already supported Community Centers just like this for HPARD. This just the county. And the other operator:type should have always been government, since they're taxing authorities.

PR #11348, opened 2025-09-05
Added pharmacy brand Burchacińscy

### LINKED ISSUE #11109: Add Burchacińscy
amenity=pharmacy
healthcare=pharmacy
dispensing=yes

Wikidata: Q133856230
https://apteka-burchacinscy.pl/
PL

sarvenazranjbar:
Fix #11109

### DIFF
--- a/data/brands/amenity/pharmacy.json
+++ b/data/brands/amenity/pharmacy.json
@@ -2037,6 +2037,20 @@
       }
     },
     {
+      "displayName": "Burchacińscy",
+      "id": "",
+      "locationSet": {"include": ["pl"]},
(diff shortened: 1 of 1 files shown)


UKChris-osm:
At the top of this PR is the text `The head ref may contain hidden characters: "Burchaci\u0144scy"` but I've never seen this before, so I'm not sure if merging this would cause any problems?

sarvenazranjbar:
Hello, this warning means that the branch name includes some unicode in this case the accent which shouldn't be an issue but just to be safe I've made a new branch with the same solution and a new name so we don't get this warning here's the link to it https://github.com/osmlab/name-suggestion-index/pull/11350

PR #11349, opened 2025-09-05
Update cafe.json

jhutton1:
update display, name, brand to only be in local language. This to be consistent with how all other Starbucks brands in various countries are available.


jhutton1:
Understand HK displays in both languages.  Closing PR.

PR #11353, opened 2025-09-06
add ownership information to HPL

EvanCarroll:
Add ownership information to Houston Public Library

### DIFF
--- a/data/operators/amenity/library.json
+++ b/data/operators/amenity/library.json
@@ -1345,7 +1345,9 @@
         "operator": "Houston Public Library",
         "operator:short": "HPL",
         "operator:type": "government",
-        "operator:wikidata": "Q1647690"
+        "operator:wikidata": "Q1647690",
+        "owner": "City of Houston",
+        "owner:wikidata": "Q16555"
       }
     },
     {


EvanCarroll:
If this isn't problematic, I would like to get this commited. This follows the trend we did with HPARD and would allow us to simplify queries. It also does a bit to document `operator:type=government`

UKChris-osm:
I haven't kept up with the NSI much this year, so don't know if this should be added or not, but I do see the `owner` tag already exists within the NSI data.

I feel like it isn't needed though, as the information could be made available via the Wikidata of the current `operator`.

Also, the newly added Wikidata seems to point to the [City of Houston](https://www.wikidata.org/wiki/Q16555) itself - would it not be the [City Council](https://www.wikidata.org/wiki/Q5916428) / local authority who own the library?

EvanCarroll:
@UKChris-osm The City Council doesn't own anything. That's just a legislative body, like congress or whatever. The City of Houston proper owns the library buildings. We're a strong mayor city too. The reason why we need this is to do geospatial queries across all public municipal assets.

From the perspective as the user this is the most useful information. It's actually _more_ important than the operator information. For example, in the City of Houston, some meeting centers are operated by the Houston Parks and Recreation, some are operated by Houston Department of Health, some by the Houston Public Library, we have even more that are operated by certain corporations TIRZ's. But ultimately, if you have a problem with City of Houston resources you need to call `311` to file a ticket -- they all use the same ticketing system. You may think `owernship:type=government` works, but not really because we also have dual-governments in Texas. We still have a commissioner system, with HCP1, HCP2, HCP3, and HCP4 representing rival governments that all have their own distinct reservation and ticketing sytems.

You're right to say you _could_ use wikidata for this, but on that note you could use Wikidata for everything in OSM less the link to the way which wikidata has a link for (OSM Way). 

This is for the Houston Hackathon, we're trying to index all Houston-owned assets for Overpass queries. Specifically for the purposes of acquiring a meeting room.

Cj-Malone:
Given the high amount of cross tagged places (POI + building or POI + land use) I am also wary of making this addition. I don't know the specifics this organisation, they may currently own all there buildings, OSM may currently have these tagged separately to the buildings, but neither of these things are constants.

You can still add these tags to OSM where true, it also means you can do your overpass query a minute after editing instead of waiting years for it to happen via NSI and iD.

GA-Kevin-Codes:
I am in Georgia so not intimately familiar with Texas government organization but it seems fairly common for a city to have different departments and those departments to be the operator of the POI. A local park is operator of Parks Department, a library operator of the local library system. Owner = city seems to delve into more of a org chart which is not what OSM is for but is really great for Wikidata. There, you can follow it all the way up. This particular library is operated by the library system, that system by the city, that city by the county, that county by the state, that state by the US government. OSM would just be interested in the immediate operator.

bhousel:
I agree with the others here - if you want to push `owner` tags to all the libraries in Houston and you have the support of the community there, you could do this with Overpass and JOSM.  

Adding `owner` tags through NSI is probably a mistake as this is information that can be looked up from Wikidata.  [There are 6 items in NSI](https://github.com/search?q=repo%3Aosmlab%2Fname-suggestion-index+%22owner%3Awikidata%22&type=code) where we set the `owner` tag, and we should maybe stop doing it.

PR #11358, opened 2025-09-07
[Londis GB and IE] Remove name

TheUKHighStreet:
I'd like it to be an option to add the name of the shop for Londis as many of them have a name. Eg https://hisdeedsaredust.com/tmp/20250907T1223-londis.jpg

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -3362,7 +3362,6 @@
       "tags": {
         "brand": "Londis",
         "brand:wikidata": "Q21015800",
-        "name": "Londis",
         "shop": "convenience"
       }
     },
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Personally I'd leave  the `name` tag as it is, but add `"preserveTags": ["^name"]` to allow the name to be overridden in cases like this.

TheUKHighStreet:
Thanks - that was exactly what I wanted but didn't know it existed.

UKChris-osm:
Sorry, I wasn't clear about the placement of the `"preserveTags": ["^name"]` tag.

TheUKHighStreet:
Thanks - I'll know for next time - I think there may be some other convenience chains which could do with this adding too.

PR #11372, opened 2025-09-10
Merge subway network

hlfan:
The subway doesn't have its own network.
See also https://c.osm.org/t/135462 and https://ptna.openstreetmap.de/results/AT/AT-VOR-Analysis.html#notconsiderednetworks

### DIFF
--- a/data/transit/route/subway.json
+++ b/data/transit/route/subway.json
@@ -917,9 +917,10 @@
         "include": ["at-9.geojson"]
       },
       "tags": {
-        "network": "U-Bahn Wien",
-        "network:wikidata": "Q209400",
+        "network": "VOR",
+        "network:wikidata": "Q2516485",
         "operator": "Wiener Linien",
+        "operator:short": "WL",
         "operator:wikidata": "Q668849",
         "route": "subway"
       }


ilias52730:
Merged.

Cj-Malone:
U-Bahn Wien will just be auto added back after the next build

ilias52730:
> U-Bahn Wien will just be auto added back after the next build

Isn't there a way to stop that from happening? exclude or whatever?

Cj-Malone:
Yeah, match names or exclude

ilias52730:
You seem to know how to do that, could you please add the small line of code that's needed to avoid this from happening?

Cj-Malone:
I don't know which? Should OSM cases of "U-Bahn Wien" be encouraged to turn into "VOR" (match names) or are they different things (exclude).

hlfan:
There is no separate "U-Bahn Wien" network; it's all in the same network - VOR.
The cases should be changed so the lines aren't excluded from the [AT-VOR PTNA](https://ptna.openstreetmap.de/results/AT/AT-VOR-Analysis.html#notconsiderednetworks).

PR #11375, opened 2025-09-10
Added Taylor Corporation

### LINKED ISSUE #9097: Add Taylor Communications
https://www.wikidata.org/wiki/Q17108752

sarvenazranjbar:
Fixes #9097

### DIFF
--- a/data/brands/office/company.json
+++ b/data/brands/office/company.json
@@ -26,6 +26,17 @@
       }
     },
     {
+      "displayName": "Taylor Corporation",
+      "id": "",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
They seems to have a couple of [locations](https://www.taylor.com/locations) in Canada and Mexico as well, although perhaps these are branded differently?

sarvenazranjbar:
I've added Canada and Mexico to the locationSet. Checking the their website they don't seem to advertise those locations as a separate brand so I think it's safe to have all as one item.

PR #11398, opened 2025-09-19
Add OVHcloud (telecom/data_center)

davidhicks:
Locations listed at https://www.ovhcloud.com/en-au/about-us/global-infrastructure/regions/


davidhicks:
Replaced by #11415

PR #11402, opened 2025-09-19
Refactor loop usage to comply with Sonar rule S1751

jayamrutiya:
### What kind of change does this PR introduce?
- Code refactor for better readability and maintainability.
- Removed unnecessary for loop with unconditional break.

### Why was this change needed?
- The original loop only executed once due to an unconditional break.
- Sonar flagged this as a violation of rule javascript:S1751 (loops should not contain more than a single break/continue).
- Refactoring improves code clarity and eliminates the Sonar warning.

### Other information:
- Replaced loop with direct access to the first element of keylist.
- No functional impact — behavior remains the same.

### DIFF
--- a/lib/idgen.js
+++ b/lib/idgen.js
@@ -67,10 +67,12 @@ export function idgen(item, tkv, locationID) {
           if (ex.test(key)) keylist.push(key);
         });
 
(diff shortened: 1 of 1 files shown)


bhousel:
thanks!

PR #11405, opened 2025-09-22
All'Antico Vinaio

codeinabox:
Add the Italian sandwich shop brand, All'Antico Vinaio, which has locations in Italy, New York, Los Angeles, Las Vegas, and London

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -275,6 +275,20 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


ilias52730:
It seems they also have a store in the Dubai airport https://www.allanticovinaio.com/dubai/

codeinabox:
> It seems they also have a store in the Dubai airport https://www.allanticovinaio.com/dubai/

Well spotted, have added it to the location set.

PR #11407, opened 2025-09-22
Add Tuttogiardino (shop/garden_centre)

ilselvaggio:
Locations listed at https://www.tuttogiardino.it/

### DIFF
--- a/data/brands/shop/garden_centre.json
+++ b/data/brands/shop/garden_centre.json
@@ -280,6 +280,7 @@
         "name": "Tuttogiardino",
         "shop": "garden_centre"
       }
+    },
     {
       "displayName": "Villaverde",
       "id": "villaverde-afed88",


bhousel:
I don't think this locationSet will work
```
      "locationSet": {
        "include": ["it"],
        "exclude": ["Bolzano-Bozen"]
      }
```
"Bolzano-Bozen" isn't a thing our scripts would recognize.

ilias52730:
> I don't think this locationSet will work
> 
> ```
>       "locationSet": {
>         "include": ["it"],
>         "exclude": ["Bolzano-Bozen"]
>       }
> ```
> 
> "Bolzano-Bozen" isn't a thing our scripts would recognize.

It's a province that's part of a bigger region. [features/it](https://github.com/osmlab/name-suggestion-index/tree/main/features/it) only has regions. In any case, I don't see the point in excluding a province. I removed the line.

PR #11411, opened 2025-09-23
Adding Canadian linen store, Daniadown

ShinyDog2636:


ilias52730:
From their website it looks like a regular `shop=bed`, it's a _shop that specializes in selling beds, mattresses, pillows, duvets, covers and other bedding products._

Plus the tag `shop=household_linen` only has 1,862 uses.

ShinyDog2636:
Where I am from bed stores and bedding/linen stores are distinct amenities. But reading through the wiki and I see that I could use the shop=bed with mattress=no, furniture=no, linen=yes. I will resubmit with these changes.

PR #11420, opened 2025-09-29
Velvet Taco

codeinabox:


codeinabox:
Could someone please confirm if this should be tagged as a restaurant or fast food. I had a look in Overpass, and there was a mix, though it appeared the fast food tagging was in the majority.

GA-Kevin-Codes:
I believe they refer to themselves as "fast casual" which makes me think of something like Chipotle which is fast food in the NSI.

PR #11423, opened 2025-09-30
7 brew

### LINKED ISSUE #11422: 7 Brew — amenity=cafe (us)
```json
{
  "displayName": "7 Brew",
  "locationSet": {
    "include": [
      "us"
    ]
  },
  "tags": {
    "amenity": "cafe",
    "brand": "7 Brew",
    "brand:wikidata": "Q131838275"
  }
},
```

**Notes to editors**

Created with https://ga-kevin-codes.github.io/Suggest-A-Brand/
data file: https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/amenity/cafe.json

GA-Kevin-Codes:
Closes #11422

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -44,6 +44,11 @@
     {
       "displayName": "7 Brew",
       "locationSet": {"include": ["us"]},
+      "matchNames": [
+        "7Brew",
+        "7 Brew Coffee",
+        "Seven Brew Coffee"
+      ],
       "tags": {
         "amenity": "cafe",
         "brand": "7 Brew",


data/brands/amenity/cafe.json
```diff
@@ -41,6 +41,23 @@
     }
   },
   "items": [
+    {
+      "displayName": "7 Brew",
+      "locationSet": {"include": ["us"]},
+      "matchNames": [
+        "7Brew",
```
1ec5, line 48: By the way, you don’t need to worry about case or spaces in matchNames. Clients like iD and Rapid already do a bit of munging to catch similar names. Stuff like “coffee” at the end or spelling out “seven” is useful though.

PR #11433, opened 2025-10-05
Update estate_agent.json

Mannivu:
Add alt name for RE/MAX company, since some users could use "remax"as search key

### DIFF
--- a/data/brands/office/estate_agent.json
+++ b/data/brands/office/estate_agent.json
@@ -967,7 +967,7 @@
       "displayName": "RE/MAX",
       "id": "remax-08bb13",
       "locationSet": {"include": ["001"]},
-      "matchNames": ["re/max of america"],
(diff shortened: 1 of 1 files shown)


Cj-Malone:
This won't work

Mannivu:
Why not? Wrong syntax?

Cj-Malone:
NSI sees it as a duplicate of "RE/MAX" and will remove it.

Mannivu:
So there's no way to alert NSI that a user could write "remax" while looking for "RE/MAX" (since, at the moment, looking for "remax" gives no result)?
NSI.guide page. Looking for remax gives no result

bhousel:
Ah yeah, this won't really do anything.    
We could probably change the nsi.guide frontend code to do more fuzzy matching.

PR #11434, opened 2025-10-05
fast food Белоруснефть

Michael21011:


Cj-Malone:
What ate you trying to do? Do you want to remove Белоруснефть? Why change the name to Заўжды цікаVа but not the brand or display name.

tohaklim:
Uh, so from what I know:
- Белоруснефть is the garage operator (hence their brand on amenity=fuel and, historically, everything that encompasses, including retail and food outlets)
- At some point they decide to rebrand their fast food outlets at all (most?) garages to Заўжды цікаVа. Белоруснефть remains the operator/owner (through their local branches) of the site and the fast food outlet too
- Current iD behaviour seems to be suggesting Белоруснефть for these fast food entries, which is I guess what @Michael21011 is trying to correct?

Michael21011:
> What ate you trying to do? Do you want to remove Белоруснефть? Why change the name to Заўжды цікаVа but not the brand or display name.

In additional to @tohaklim answer I will show this photo.  It is fuel station with fast food. Owner is Белоруснефть but fastfood has own name "Заўжды цікаVа ".
But ID editor insist to add name=Белоруснефть. I want to remove this suggestion and set new names.

Screenshot_2025-10-05-20-43-39-081-edit_com mapillary app

https://mapillary.com/map/im/1283546929650741

bhousel:
Thanks, this makes sense and is probably fine - I agree there should probably be an entry in the fast_food file for "Заўжды цікаVа", so I created [Q136565015](https://www.wikidata.org/wiki/Q136565015) for it and updated this entry a bit in 4fe8ab70a - hope I got everything right.

PR #11440, opened 2025-10-07
Removed incorrect amenity=fuel

SteveLz:
uSmile 昆仑好客 is a convenience store at a PetroChina gas station, not the gas station itself.

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -6980,20 +6980,6 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


SteveLz:
Added the match name "uSmile 昆仑好客" for PetroChina, as many people mistake the name of a convenience store for a gas station, which is why this preset is here (19a8d3e3cb51cd53c93fbcadf27ed5644afcc92a).

PR #11442, opened 2025-10-08
Add `mul` wikidata support

### LINKED ISSUE #10112: Wikidata sync script may need modifying to accommodate Wikidata's new multilingual label feature
At some point this year, Wikidata implemented a [multilingual label feature](https://www.wikidata.org/wiki/Help:Default_values_for_labels_and_aliases) intended to cut down on the number of identical labels and aliases for items across different languages. I ran the Wikidata sync script earlier, and it came across an item where a user utilized the multilingual label feature and subsequently blanked the labels of most languages, including English. The Wikidata sync script ignored the multilingual label and added a default English label as it does for Wikidata items without multilingual or English labels. (Given how new the feature is, I admit that this was the expected behavior.)

The reason I used the phrase "may need modifying" in the issue title is that in this specific case, the chosen English label was different than the value set for the multilingual label. While we may want to avoid duplicating multilingual labels if the value the script would set for English is the same as the multilingual label, there may also be situations where the multilingual label is flawed and an override would be appropriate.

Dimitar5555:
resolves #10112 
ref https://github.com/osmlab/name-suggestion-index/discussions/11270#discussioncomment-14266671

I haven't tested it but following the logic of the code, it should work correctly.

### DIFF
--- a/scripts/build_wikidata.js
+++ b/scripts/build_wikidata.js
@@ -235,8 +235,9 @@ function processEntities(result) {
     const meta = _qidMetadata[qid];
     let target = _wikidata[qid];
     let entity = result.entities[qid];
-    let label = entity.labels && entity.labels.en && entity.labels.en.value;
-    let labelmul =  entity.labels && entity.labels.mul && entity.labels.mul.value;
+    const labelEn = entity.labels && entity.labels.en && entity.labels.en.value;
+    const labelMul =  entity.labels && entity.labels.mul && entity.labels.mul.value;
+    let label = labelEn ? labelEn : labelMul;
 
     if (entity.redirects) {
       const warning = { qid: qid, msg: `Wikidata QID redirects to ${entity.redirects.to}` };


scripts/build_wikidata.js
```diff
@@ -235,8 +235,9 @@ function processEntities(result) {
     const meta = _qidMetadata[qid];
     let target = _wikidata[qid];
     let entity = result.entities[qid];
-    let label = entity.labels && entity.labels.en && entity.labels.en.value;
-    let labelmul =  entity.labels && entity.labels.mul && entity.labels.mul.value;
+    const labelEn = entity.labels && entity.labels.en && entity.labels.en.value;
+    const labelMul =  entity.labels && entity.labels.mul && entity.labels.mul.value;
+    let label = labelEn ? labelEn : labelMul;
```
1ec5, line 240: I think we’ll just barely get lucky and avoid editing Wikidata to set the English label to what’s in the multilingual label, which is the issue you’re trying to fix here. The logic below is confusing, but essentially we’re trying to avoid going down the code path where we override the Wikidata label with what’s in OSM. Long-term, it would be safer to separate out the tasks of pulling from and pushing to Wikidata, especially since the pushes are so English-centric.

1ec5 (approved):
I haven’t tested this change either, since I don’t have NSI hooked up to a bot account on Wikidata. However, it does look like an improvement over the state of the code in 2761e9a6a9ff3e8e804ea504fafc01608c960714.

scripts/build_wikidata.js
Dimitar5555: See the logic below. If an item doesn't have `en` and `mul` name in Wikidata, it will get the `en` label from NSI and set it as `en` label in Wikidata. If that item has either `mul` or `en` in Wikidata, it won't change anything.

```js
// Get label...
if (label) {
  target.label = label;
} else {
  // try to pick an English label.
  label = enLabelForQID(qid);
  if (label && _wbEdit) {   // if we're allowed to make edits, just set the label
    target.label = label;
    const msg = `Adding English label for ${qid}: ${label}`;
    wbEditQueue.push({ id: qid, language: 'en', value: label, msg: msg });

  }
}
```

1ec5: I know, it is safe, it’s just a bit confusing.

bhousel:
Seems fine, I'll test it in a few days.. Really having our script update labels on wikidata at all is kinda optional..

PR #11462, opened 2025-10-18
Fixes for Calgary Transit

TagaSanPedroAko:
Replaces CT with Calgary Transit, moving the CT tagging to network:short. Change in response to problems with current network tagging for Calgary Transit at https://community.openstreetmap.org/t/gtfs-data-and-ptna-calgary-transit/136904

### DIFF
--- a/data/transit/route/light_rail.json
+++ b/data/transit/route/light_rail.json
@@ -53,15 +53,17 @@
       }
     },
     {
-      "displayName": "CTrain",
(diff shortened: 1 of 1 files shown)


TagaSanPedroAko:
This also changes the tagging for CTrain, moving the system branding under network:metro and tagging them under network=Calgary Transit. The bus and CTrain routes use the same fare system and are also operated by Calgary Transit.

PR #11478, opened 2025-10-24
Update tags for Toronto subway

TagaSanPedroAko:
This updates the tag preset for Toronto subway, moving "Toronto subway" to network:metro and its associated Wikidata to network:metro:wikidata, making TTC the network value (in line with other transit modes under the Toronto Transit Commission such as its buses and streetcars/trams which have been using network=TTC for long). This follows up a similar change completed for Calgary Transit's CTrain light rail transit (LRT) network (part of #11462), which moves the system branding under network:metro.

### DIFF
--- a/data/transit/route/subway.json
+++ b/data/transit/route/subway.json
@@ -908,7 +908,7 @@
         "network": "TTC",
         "network:metro": "Toronto subway",
         "network:metro:wikidata": "Q20379",
-        "network:wikidata": "Q6612865"
+        "network:wikidata": "Q6612865",
         "operator": "Toronto Transit Commission",
         "operator:short": "TTC",
         "operator:wikidata": "Q17978",


bhousel:
I haven't seen tags like `network:metro` and `network:metro:wikidata` before.
I admit I missed it when I merged #11462 - maybe we shouldn't add these?
Are they documented anywhere?

TagaSanPedroAko:
those keys aren't documented, but network:metro has been added to the preset for Warsaw Metro as for usage with route=subway.

bhousel:
> those keys aren't documented, but network:metro has been added to the preset for Warsaw Metro as for usage with route=subway.

Is that were you got the idea to add it?  or is there some other place where I can read about this tag.

TagaSanPedroAko:
I didn't get the idea directly how Warsaw Metro is tagged. Well, it had about 2,000 uses according to Taginfo, mostly in parts of Europe as well as East and Southeast Asia. It seems to have been created to handle cases of a rail transit system with distinctive branding that is tagged as part of a larger multimodal transit network with unified fare system and usually of regional scope (such as regional transport unions in parts of Europe and regional transit authorities in parts of North America).

bhousel:
Ok thanks for the explanation.
I think the tag is fine - and if it offers some value then NSI would be a good place to include it - but I'm not comfortable including right now until it gets some documentation.  As soon as iD starts suggesting adding the tag, people will get mad at us.  I'm going to merge this, but then go through and remove the `network:metro` tags from the index, for now.  We can always put them back later.

PR #11481, opened 2025-10-24
Add preserveTags for 'De Echte Bakker' brand

wvanderp:
De Echte Bakker is a suffix for the name of the baker.

So it would be Simon, De Echte Bakker (translated Simon, real baker) 

De echte bakker is a shortened form of the real bakkers guild


bhousel:
Makes sense, thanks!

Issue #11311, opened 2025-09-01
add Williams Fresh Cafe


jarek:
It's a cafe, seems a bit fancier than fast food.

Currently all 13 locations are in Ontario, Canada.

All but one locations are already in OSM and tagged with `brand:wikidata` (one is not certain - I set a fixme), but it probably can't hurt to have the preset in case they open or move?

Wikidata: https://www.wikidata.org/wiki/Q106195633
Website: https://williamsfreshcafe.com/

Name variations: "William's Fresh Cafe" (apostrophe has previous been used in OSM but it is incorrect); they used to be called "Williams Coffee Pub" and I've also seen "William's Coffee Pub" tagged

UKChris-osm:
If there's only 13 locations, and they are quite local to each other, and all but one are already mapped, I don't think it really needs to be a part of the NSI, unless there's a plan for expansion?

One of the general ideas of the NSI is to help multiple mappers from around the world map the same items in a consistent manor, by giving them hints and suggestions, but if, in this case it seems, _Williams Fresh Cafe_ is already mapped in a consistent way, there's no real advantage to adding it to the NSI at the moment.

jarek:
> If there's only 13 locations, and they are quite local to each other, and all but one are already mapped, I don't think it really needs to be a part of the NSI, unless there's a plan for expansion?

Sure, that's fine by me if you decide it's out of scope, I just wasn't really sure

UKChris-osm:
I'll leave this open however, in case others disagree with me and wish to add it.

bhousel:
Sure this is fine, I added it in 6fa151ee9 as an `amenity=fast_food`  (we use this tag for "fast casual" restaurants)

PR #11484, opened 2025-10-24
No apostrophe in Williams Fresh Cafe

jarek:
ref https://github.com/osmlab/name-suggestion-index/issues/11311

see website https://williamsfreshcafe.com/ , Wikidata https://www.wikidata.org/wiki/Q106195633

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -11436,18 +11436,18 @@
       }
     },
     {
-      "displayName": "William's Fresh Cafe",
+      "displayName": "Williams Fresh Cafe",
       "id": "williamsfreshcafe-72d462",
       "locationSet": {
         "include": ["ca-on.geojson"]
       },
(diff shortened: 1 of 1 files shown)


data/brands/amenity/fast_food.json
```diff
@@ -11436,18 +11436,18 @@
...
     },
     {
-      "displayName": "William's Fresh Cafe",
+      "displayName": "Williams Fresh Cafe",
       "id": "williamsfreshcafe-72d462",
       "locationSet": {
         "include": ["ca-on.geojson"]
       },
-      "matchNames": ["williams coffee pub"],
+      "matchNames": ["williams coffee pub", "william's fresh cafe", "william's coffee pub"],
```
1ec5, line 11444: I think punctuation like apostrophes gets ignored when matching tags to presets anyways, but it’s probably harmless.

1ec5 (approved):
Good catch, thanks!

PR #11489, opened 2025-10-27
use ETS as network=* for Edmonton LRT

TagaSanPedroAko:
this changes the network value for the Edmonton LRT preset to ETS, and also removes operator tags to account for one line and associated infrastructure not directly operated by ETS.

### DIFF
--- a/data/transit/route/light_rail.json
+++ b/data/transit/route/light_rail.json
@@ -132,7 +132,8 @@
       },
       "matchNames": [
         "edmonton transit system",
-        "ets",
+        "edmonton transit service",      
+        "edmonton lrt",
       ],
       "tags": {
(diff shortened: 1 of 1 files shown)


data/transit/route/light_rail.json
```diff
@@ -132,13 +132,11 @@
       },
       "matchNames": [
         "edmonton transit system",
-        "ets"
+        "ets",
```
bhousel: We should matchName some of the less-preferred names here so that they doesn't appear back in the index, and so mappers are offered to upgrade.  (We don't need to matchName "ets" now that the value is being used as the `network` tag).

```suggestion
        "edmonton transit service",      
        "edmonton lrt",
```

PR #11498, opened 2025-10-30
Update convenience.json

VodevilMark:
My first edit. Added a brand


bhousel:
Thanks, looks mostly good 👍 
We don't add email, phone, or website tags to the items here - that information can be looked up from the brand's wikidata page - so I removed these tags in 1e40574a3.

PR #11523, opened 2025-11-03
Update OMV brand

Oivo35:
Update OMV fuel station and car wash brand - locationSet, matchNames and preserveTags.

### DIFF
--- a/data/brands/amenity/car_wash.json
+++ b/data/brands/amenity/car_wash.json
@@ -20,6 +20,27 @@
   },
   "items": [
     {
+      "displayName": "OMV Top Wash",
+      "locationSet": {
+        "include": [
+          "at",
+          "bg",
+          "cz",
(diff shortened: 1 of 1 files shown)


kjonosm:
Is OMB a typo?

Oivo35:
This name "OMB" is in Bulgarian Cyrillic.

I tried to add also combination
name=OMV
name:bg=OMB
name:en=OMV
But there was an error during the check.

Dimitar5555:
> Is OMB a typo?

It seems to be auto added because there were 52 entries on OSM from many years ago. It is incorrect since OMV has always been written in Latin and never in Cyrillic. I've corrected the data in OSM, feel free to remove the Cyrillic version.

kjonosm:
To prevent auto adding, you should probably also add a matchName.

PR #11530, opened 2025-11-05
Update Transports Urbans de Sabadell

swaree:
Updating bus network in Sabadell with Wikidata ID and correct name.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -2876,13 +2876,19 @@
       }
     },
     {
-      "displayName": "Bus de Sabadell",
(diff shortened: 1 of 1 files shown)


kjonosm:
@swaree it seems that there are two different operators within the network. If that's the case please leave out `operator` entirely.

swaree:
Hi, do you mean the comma in `TUS, SCCL`? It is the legal name of a single company: SCCL is a denomination much like Inc. or GmbH

kjonosm:
Ok. Thanks for the clarification.

PR #11552, opened 2025-11-11
Adding Canadian gift store chain, Calendar Club

ShinyDog2636:


LaoshuBaby:
image

I think current store number may have problem in current standard

ShinyDog2636:
I noticed that, as well. Their website shows an incomplete list of locations until you search for a specific geography. I count more than thirty between the provinces of BC, AB, and ON after I search for Vancouver, Calgary, Edmonton, Toronto.

ilias52730:
It's because you need to remove the 100 km radius. When you change it to "No limit" you see all the stores, which are more than enough. Merging PR.
image

PR #11553, opened 2025-11-11
Update PH-LTFRB presets

TagaSanPedroAko:
Update tagging for Philippine LTFRB network tags, replacing the current PUB and PUJ values applying nationwide with geofenced regional values (e.g. LTFRB National Capital Region) covering most bus routes as well as jeepneys, and a Inter-Regional Inter-Provincial value to cover long-distance provincial bus routes (e.g. Manila to Baguio). This does not merge the presets for UV Express and Premium Point-to-Point (P2P) buses, which will remain as-is.

For regional networks (covering provincial bus routes not crossing regional boundaries, city buses and jeepneys), these ones will be supported based on current local and regional bus network mapping. Name matching is also provided to handle routes and stops still tagged with legacy values such as PH:[region]:[locale]:
- Metro Manila (National Capital Region)
- Cordillera Administrative Region
- Negros Island Region
- Central Luzon (Region III)
- CALABARZON (Region IV-A)
- Western Visayas (Region VI)
- Central Visayas (Region VII)
- Davao Region (Region XI)

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -10097,7 +10097,7 @@
     },
     {
       "displayName": "LTFRB Cordillera Administrative Region",
-      "locationSet": {"include": ["ph-15"]},
+      "locationSet": {"include": ["ph-15.geojson"]},
       "matchNames": [
         "baguio city jeepney",
         "baguio city puj",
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Generic entries come back for: LTFRB PUB, LTFRB PUJ, PUB and PUJ because they don't have entries/matchnames/excludes

PR #11565, opened 2025-11-16
Update Czech transit networks

Oivo35:
Updated names, wikidata and operators.


Cj-Malone:
MHD v Pardubicích will be readded automatically

PR #11574, opened 2025-11-17
feat(brands/supermarket): add Belarusian chains + multilingual support

VodevilMark:


ilias52730:
This is probably the biggest PR I've ever seen. Please understand we're not from Belarus and we have to check every single entry. Please split this PR into one single PR per brand, this is much easier for us and will make your PRs be added much faster.

PR #11576, opened 2025-11-17
[nsiguide] support overpass link when `tree=*`

### LINKED ISSUE #11536: Fix overpass links on nsi.guide when `tree = *`
> I've also noticed that the Overpass Turbo link for each entry disappears whenever we have * selected, which was added as part of #11120 @k-yle 

 _Originally posted by @UKChris-osm in [#11490](https://github.com/osmlab/name-suggestion-index/issues/11490#issuecomment-3480250796)_

k-yle:
Closes #11536

I didn't add this feature at the time, beacuse I wasn't sure how we should handle features with multiple wikidata tags (for example, [this one which has both `brand` & `operator`](https://nsi.guide/?t=*&k=amenity&v=bank#absa-9d35ea)).

Currently it just picks one of the `*:wikidata` tags for the query. Maybe it would be better to construct a more complex query for this case?

### DIFF
--- a/nsiguide/CategoryRow.jsx
+++ b/nsiguide/CategoryRow.jsx
@@ -141,6 +141,35 @@ relation[${k}=${v}][network:wikidata=${qid}]
     qid = Object.values(TREES)
       .map(tree => item.tags[tree.wikidataTag])
       .find(Boolean);
+
+    // for the green styling, every [*:]wikidata tag must be correct.
+    // merge these tags into an overpass tag filter string.
+    const filter = Object.keys(item.tags)
+      .filter(key => key.endsWith(":wikidata"))
+      .map(key => {
+        const plainKey = key?.replace(":wikidata", "");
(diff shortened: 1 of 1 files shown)


k-yle:
@UKChris-osm do you have an opinion about how the overpass query should be constructed for features with both `brand` and `operator`?

should [the green styling](https://overpass-turbo.eu/s/2fR5) require both `brand[:*]` and `operator[:*]` to be set correctly?

UKChris-osm:
It would be handy if the green styling could be retained when an exact match is found, with perhaps a different colour for one that matches just the brand, but not operator (and vice versa) but for entries with both it can be tricky, and so in that respect it might be safter to removed the styling and go for a more basic match which we used to do, which was a name / type match.

k-yle:
okay, updated. Here's [an example](https://overpass-turbo.eu/s/2g9k) of the query that's generated for presets with both `brand[:*]` and `operator[:*]`

PR #11577, opened 2025-11-17
Bakery chain in Lithuania and grocery chain in Georgia

ekatyshev:


ekatyshev:
Added Georgian payment terminals brand 1 Pay

PR #11579, opened 2025-11-18
Added two coffee shop chains in Lithuania, and added wikidate for one existing

ekatyshev:


ekatyshev:
Also added grocery chop chains in Turkiye and Lithuania

ilias52730:
Huracan Coffee only has 7 locations, so it cannot be added to the NSI (minimum 30 locations). See their website https://huracan.lt/en/kavines/

ekatyshev:
> Huracan Coffee only has 7 locations, so it cannot be added to the NSI

I remove it

ilias52730:
You added some brands that already exist under shop=supermarket to shop=convenience. Do these brands operate both types of shops?

ekatyshev:
> You added some brands that already exist under shop=supermarket to shop=convenience. Do these brands operate both types of shops?

I'm not sure if Sok and Macrocenter operate as supermarket (somebody else add them to there) but all, including these two, 100% operate as a convenience shops.

PR #11580, opened 2025-11-18
Update fuel.json

monk198230:


monk198230:
added EG on the move - large UK based petrol retailer

PR #11582, opened 2025-11-19
Update signmaker.json: Add Signwave (AU name for Fastsigns)

impiaaa:


data/brands/craft/signmaker.json
```diff
@@ -39,6 +39,17 @@
         "name": "Signarama"
       }
     },
+    {
+      "displayName": "Signwave",
+      "locationSet": {"include": ["au"]},
+      "matchTags": ["shop/sign"],
+      "tags": {
+        "brand": "Signwave",
+        "brand:wikidata": "Q5437127",
```
1ec5: Can you split out a new item about Signwave as a chain owned by Fastsigns, so that we don’t conflate the two brands from the beginning? Ideally the [Fastsigns <small>(Q5437127)</small>](https://www.wikidata.org/wiki/Q5437127) item would get split up too, but that can wait.

impiaaa: Hm, actually I'm not sure anymore that it is the same company. The AU website footer says "© 2025 SIGNWAVE International. Inc." and the LinkedIn URL slug is "white-room-group" (neither of which I can find on OpenCorporates, fwiw). The US, CL, MT, and UK sites do all say "© 2025 FASTSIGNS International. Inc." I also found out that at least that brand (or company?) is owned by one "[Propelled Brands](https://www.propelledbrands.com/)." So I guess, Propelled Brands owns FASTSIGNS which owns SIGNWAVE, which was once known as White Room Group? But if that's true, it's confusing that FASTSIGNS is introduced with the name SIGNWAVE in the press release.

1ec5: As far as I can tell, Fastsigns International is the company that started the original chain. They partnered with Propelled Brands, a franchise consultant, to [franchise the chain](https://www.fsfastsigns.com/). Signwave International appears to be the name of an Australian franchising partner or subsidiary – most likely subsidiary, as the same C-suite is listed on the sites of both Fastsigns International and Signwave International. In the Australian market, one can franchise [either Signwave or Fastsigns](https://www.fdsignwave.com.au/), which are equivalent brands. There seems to be some relationship between Fastsigns International and the Signwave brand, given that the company’s conference is called the “FASTSIGNS/Signwave International Convention”.

PR #11584, opened 2025-11-20
Add Fruitisimo

Oivo35:
Added fast food chain.


ilias52730:
Are you sure Fruitisimo is takeaway only? I found this picture on their website, so surely it's not takeaway only in *at least* one location.

Oivo35:
You are right. I changed value to takeaway=yes.

PR #11585, opened 2025-11-20
Update Czech optician brands

Oivo35:


ilias52730:
Extra info: Orange Optik has 67 branches according to [their website](https://www.orangeoptik.cz/pobocky). GrandOptical have branches in Czech Republic (https://www.grandoptical.cz/). Merging PR.

Issue #10939, opened 2025-05-06
Nonsense suggestion to change post-high school into kindergarten


matkoniecz:
Is NSI is supposed to be usable in such context?

https://www.openstreetmap.org/edit?editor=id&node=10568650202#map=19/51.231081/22.603083

![Image](https://github.com/user-attachments/assets/9a2c6a46-f025-40a2-85d2-0ccc4f5ee42e)

Maybe iD should ignore school entries etc? (see also https://github.com/osmlab/name-suggestion-index/issues/10926 )

bhousel:
I'm not sure..
I can see we have "Gmina Lublin"  Q37333  in :
* [`operators/amenity/school`](https://nsi.guide/index.html?t=operators&k=amenity&v=school&tt=q37333)
* [`operators/amenity/kindergarten`](https://nsi.guide/index.html?t=operators&k=amenity&v=kindergarten&tt=q37333)

So I guess we need to add it to `operators/amenity/college` also?
Otherwise NSI will just think this organization only operates kindergartens and primary schools.

matkoniecz:
If there is "Gmina Lublin" Q37333 in [operators/amenity/school](https://nsi.guide/index.html?t=operators&k=amenity&v=school&tt=q37333) and [operators/amenity/kindergarten](https://nsi.guide/index.html?t=operators&k=amenity&v=kindergarten&tt=q37333) then maybe it is iD bug to suggest such "upgrade"?

> So I guess we need to add it to operators/amenity/college also?
> Otherwise NSI will just think this organization only operates kindergartens and primary schools.

independently, it also would probably help this specific case?

bhousel:
> If there is "Gmina Lublin" Q37333 in [operators/amenity/school](https://nsi.guide/index.html?t=operators&k=amenity&v=school&tt=q37333) and [operators/amenity/kindergarten](https://nsi.guide/index.html?t=operators&k=amenity&v=kindergarten&tt=q37333) then maybe it is iD bug to suggest such "upgrade"?

I don't see it as a bug - this is just the code doing what it does.
(If the situation were switched and someone mistakenly tagged a kindergarten as a college, this would be a useful warning)

matkoniecz:
> I don't see it as a bug - this is just the code doing what it does.

well, it already has data that it can be school or kindergarten

so maybe it should have extrapolate that other values may be present but not covered by NSI yet? Rather than assuming that NSI has full coverage?

[...]

Issue #10926, opened 2025-05-01
Incorrect iD suggestion by Orlen brand


Oivo35:
iD incorrect suggests to change [Orlen](https://nsi.guide/index.html?t=brands&k=amenity&v=fuel#orlen-5730fb) brand to [Orlen express](https://nsi.guide/index.html?t=brands&k=amenity&v=fuel#orlenexpress-83551a) at advertising totem. Both presets are in NSI. This doesn't happen at `Orlen` fuel stations.
e.g. this object:
https://www.openstreetmap.org/node/12314042068

Snowysauce:
[Both advertising presets do indeed exist in Slovakia](https://nsi.guide/index.html?t=brands&k=advertising&v=totem&tt=orlen), so in theory this shouldn't be happening. I'm not familiar enough with the inner workings of template-generated presets to say whether this is a bug on our end or with iD.

Snowysauce:
I looked at a couple comparable presets on a whim, and it further muddies the waters as to what is happening with the Orlen presets.

Compare the presets for Lotos/Lotos Optima to those of Orlen/Orlen express:

```
    "advertising/totem/lotos-d7e9e3": {
      "name": "Lotos",
      "locationSet": {"include": ["pl"]},
      "geometry": ["point", "line"],
      "matchScore": 2,
      "terms": ["lotos"],
      "tags": {"advertising": "totem", "brand:wikidata": "Q1256909"},
      "addTags": {
        "advertising": "totem",
        "brand": "Lotos",
        "brand:wikidata": "Q1256909"
      }
    },
    "advertising/totem/lotosoptima-d7e9e3": {
      "name": "Lotos Optima",
      "locationSet": {"include": ["pl"]},
      "geometry": ["point", "line"],
      "matchScore": 2,
      "terms": ["lotos optima"],
      "tags": {"advertising": "totem", "brand:wikidata": "Q1256909"},
      "addTags": {
        "advertising": "totem",
        "brand": "Lotos Optima",
        "brand:wikidata": "Q1256909"
      }
    },
```

```
    "advertising/totem/orlen-5e42d4": {
      "name": "Orlen",
      "locationSet": {"include": ["de", "hu", "lt", "pl", "sk"]},
      "geometry": ["point", "line"],
      "matchScore": 2,
      "imageURL": "https://graph.facebook.com/GrupaORLEN/picture?type=large",
      "terms": ["orlen"],
[...]

PR #11597, opened 2025-11-24
remove entry causing bogus suggestions

### LINKED ISSUE #11479: wrong suggestion to change post-high school into kindergarten
this is NOT a duplicate of #10939 - that case is now fixed in live iD

Maybe iD should ignore school entries etc? (see also https://github.com/osmlab/name-suggestion-index/issues/10926 )

https://www.openstreetmap.org/edit?editor=id&node=11117465383#map=19/52.224661/21.033926

Image

```
amenity=college
email=sekretariat@zs23.edu.pl
isced:level=4
name=Szkoła Policealna nr 37 dla Dorosłych
note=Parent RSPO: 25358
official_name=Szkoła Policealna nr 37 dla Dorosłych
operator=Śródmieście
operator:type=government
phone=+48 226286803
ref:regon=015812870
ref:rspo=30617
website=https://www.zs23.edu.pl/
```

some other mapper claimed that it is a very widespread problem across Warsaw, affecting many objects with `isced:level=4` (not verified that claim, just asked for example POI so I can make an issue)

matkoniecz:
fixes #11479

there may be less ugly way if fixing it, or more radical like iD ignoring some parts of NSI

see also #10926 #10939

### DIFF
--- a/data/operators/amenity/kindergarten.json
+++ b/data/operators/amenity/kindergarten.json
@@ -1911,19 +1911,6 @@
       }
     },
     {
-      "displayName": "Śródmieście",
-      "id": "srodmiescie-d9c326",
-      "locationSet": {
-        "include": ["pl-14.geojson"]
-      },
-      "tags": {
-        "amenity": "kindergarten",
-        "operator": "Śródmieście",
-        "operator:type": "government",
-        "operator:wikidata": "Q7095"
-      }
-    },
-    {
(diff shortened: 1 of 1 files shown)


bhousel:
In 9ce796eeb I added `skipCollection` for the school-like categories. This means that going forward we won't collect names from the planet for these.

PR #11598, opened 2025-11-24
Kineya Mugimaru

codeinabox:


data/brands/amenity/fast_food.json
```diff
@@ -14549,6 +14567,22 @@
         "takeaway": "yes"
       }
     },
+    {
```
LaoshuBaby, line 14570: Screenshot 2025-11-26 at 15-40-29 店舗検索 グルメ杵屋レストラン
This brand is a little strange, if only focus on the "kineya mugimaru" brand, all over Japan is not required to at least 10 shops' reqiurement, but the operator "gourmet kineya" has [many brand](https://gourmet-kineya.co.jp/brands/) and each brand have a small group of shops, so if calculated for the corporation company it obviously have a strong notability.

But the problem is that those brand owned by "gourmet kineya" have different brand name and shop name, such as "kineya", "kanouya", "dondontei", "honoka" and many others. and they need to create a series of presets.

codeinabox: I'm only familiar with the Kineya Mugimaru brand, as they've started opening franchises in the United Kingdom. There was already an existing Wikidata entry for it, so I thought it'd be a quick win to add it to the NSI. I can help with the other sub-brands if you think that's worthwhile.

LaoshuBaby: > I can help with the other sub-brands if you think that's worthwhile.

Maybe don't need to do such more, I'm finding which brands are major brand in this company and choose which really should be added. I'll done this in a later commit

PR #11604, opened 2025-11-24
Add Laser Quest brand for UK

wmorland:
- There are enough locations on https://laserquest.co.uk/locations/ to meet notability requirement
- id generated with `bun run build`
- Confirmed changes built and linted with `bun run all`


codeinabox (approved):
Thank you!

PR #11607, opened 2025-11-27
Add new network tags for De Lijn (Belgium)

bxl-forever:
The organisation of public transport in Flanders (Belgium) has changed.
Tram and bus routes operated by De Lijn are now split into [15 smaller networks](https://community.openstreetmap.org/t/rethink-the-network-tag-for-de-lijn-stops-and-routes/135072), each with its own transit authority (vervoerregio).

This change adds 15 extra network values for buses and trams.

Tram: legacy values (DLAn, DLOV) can be removed because OSM data is already 100% updated to the new tagging scheme.
Bus: we temporarily keep legacy values (DLAn, DLLi, DLOV, DLVB, DLWV) along the new ones. This is to prevent automatic suggestions in iD. Once data migration is complete, we will issue a new pull request to remove the 5 legacy entries from bus.json.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -22647,6 +22647,186 @@
       }
     },
     {
+      "displayName": "Vervoerregio Antwerpen",
+      "id": "vvrant-c3736a",
+      "locationSet": {"include": ["be", "nl"]},
+      "tags": {
+        "network": "VVR Antwerpen",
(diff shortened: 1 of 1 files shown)


bxl-forever:
There was a typo in one of the files → Let’s ignore this pull request.  Redone as https://github.com/osmlab/name-suggestion-index/pull/11608

PR #11615, opened 2025-11-29
add locationSet to all IKEA presets

k-yle:
I just tried to map the well-known shop, but all I found was some random stuff like "Power Generator". So i think we should make the `locationSet` consistent for all IKEA entries.

and to solve my original problem, I added NZ to the `locationSet` ([reference](https://stuff.co.nz/nz-news/360901349))

image


data/operators/advertising/billboard.json
```diff
@@ -570,8 +570,58 @@
     },
     {
       "displayName": "IKEA",
-      "id": "ikea-d4816c",
-      "locationSet": {"include": ["001"]},
+      "id": "ikea-1dde13",
+      "locationSet": {
+        "include": [
```
LaoshuBaby, line 575: I guess this is unnecessary? because the wholesale market IKEA is truly have business in too many countries, so the advertising board can be seen in those so many countries, I guess it is also OK to keep the 001 and don't need to list such long list?

LaoshuBaby: Oh my fault, just noticed that currently all IKEA related things are described in a full list. 🥺

PR #11569, opened 2025-11-16
Add Otrębusy bakery chain in Poland

rogoyski:
Add another bakery chain from Warsaw area, Poland. Smaller than _Putka_, _Gorąco Polecam_ or _SPC_, but comparable to _Gromulski_, which are all part of NSI. On Instagram they invite to _over 30 locations_.

Instagram: https://www.instagram.com/ak_otrebusy/
Facebook: https://www.facebook.com/AK.Otrebusy/

PR #11617, opened 2025-11-30
update capitalisation of Geers brand

rogoyski:
Geers is not an acronym. If it is spelled all-caps, it is done so just for the purpose of increasing the visibility of the brand name in text. Neither [Wikidata](https://www.wikidata.org/wiki/Q1497707) nor [Wikipedia](https://de.wikipedia.org/wiki/Geers_H%C3%B6rakustik) found it warranted to apply an all-caps spelling.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,6 @@ node_modules/
 
 .DS_Store
 .vscode
-.idea
 npm-debug.log
 lerna-debug.log
 package-lock.json


ilias52730:
Yes it was, but I removed it since it got no answer.

Do I have to add .idea to gitignore or no need?

rogoyski:
Hi! No need, I will perhaps try adding it again one day with an other pull request, and giving an appropriate comment.

What I am afraid is that I can't seem to find any trace of a question being asked about that line :/

ilias52730:
Weird, it was a review of that line. Perhaps reviews don't get sent as an email? I'll simply comment next time then.

rogoyski:
FWIW, I also made the same edit to `.gitignore` in #11569. I also don't see any code comments there besides the one I added just now. I hope I am not missing something in the pull request process since I am not used to GitHub.

Thanks anyway!

PR #11622, opened 2025-12-01
Add TRAID charity shop

peternewman:
Whilst they probably technically don't meet the notability requirements with only twelve stores, we've got their recycling bins in the system already, so this is all it takes to add the shops too:
https://traid.org.uk/shop-at-traid/


peternewman:
Thanks @Cj-Malone

PR #11625, opened 2025-12-02
Vintage Cellars replaced by Liquorland Cellars in AU

andrewharvey:
in Australia, the Vintage Cellars brand was retired, all stores have now been rebranded into either "Liquorland Cellars" or "Liquorland".

The Liquorland brand already exists, so in this change I've removed the now retired "Vintage Cellars" and added the new "Liquorland Cellars" brand.

### DIFF
--- a/data/brands/shop/alcohol.json
+++ b/data/brands/shop/alcohol.json
@@ -527,6 +527,16 @@
       }
     },
     {
+      "displayName": "Liquorland Cellars",
+      "locationSet": {"include": ["au"]},
+      "tags": {
+        "brand": "Liquorland Cellars",
+        "brand:wikidata": "Q137190282",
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Vintage Cellars will be re added, it should be excluded?

andrewharvey:
> Vintage Cellars will be re added, it should be excluded?

Sorry, where/why will it be re added?

Cj-Malone:
`bun run build` adds entries from OSM

andrewharvey:
> bun run build adds entries from OSM

well in this case, the brand has been retired and (apparently) all the stores have now been re-branded, so I think we should not bring it back in. over time as mappers catch up they'll disappear from OSM.

PR #11627, opened 2025-12-04
"Bajs" - Zagreb public bicycle rental system

mnalis:
(some ~180 stations, more added every day)


mnalis:
is there anything else needed here?

PR #11630, opened 2025-12-06
Update clothes.json

Mannivu:
Fixed name of Italian clothes chain


Cj-Malone:
Primigi Store should be a match name, it'll be readded automatically

Mannivu:
Yes, you're right, I forgot to include that.

PR #11636, opened 2025-12-06
fix Aida cafe name and extra info

cuatim:
Hi,

I fixed the name of this coffee house chain (source: I have been going to Aida for 20+ years). Aïda, (with a diaeresis) is not actually used beyond the logo, and that is technically Ɑïdɑ according to [Wikipedia](https://de.wikipedia.org/wiki/Aida_(Unternehmen)).

Thanks!
C

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -282,8 +282,7 @@
         "amenity": "cafe",
         "brand": "Aida",
         "brand:wikidata": "Q794636",
-        "cuisine": "coffee_shop;cake;pastry",
-        "drink:espresso": "yes",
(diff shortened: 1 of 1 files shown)


cuatim:
I could add potentially add `alt_name=Aïda`?

cuatim:
@ilias52730 made the requested changes :)

LaoshuBaby:
> I could add potentially add `alt_name=Aïda`?

Of course if it also often been seen in real life

ilias52730:
> I could add potentially add `alt_name=Aïda`?

I don't think this is an actual alternative name, just a different way to write it. No need to add both Aida and Aïda, really. This PR looks great to me and can be merged.

PR #11637, opened 2025-12-06
Add Filmstaden, Svenska Bio cinema brands

sp1187:


ilias52730:
From my research, there seem to be so many different cinema brands here. Do you have any more info?

- Filmstaden
- Svenska Bio
- Cinema 3
- Cinemascenen
- Röda Kvarn

sp1187:
The rest seems to be too small to put on NSI, but I added Svenska Bio to the PR (distinct brand and company though it is partially owned by Filmstaden and uses the same website).

PR #11643, opened 2025-12-07
Add New York City Housing Authority

akadouri:


GA-Kevin-Codes:
From a QA perspective is it typically the land or building / floors / units owned by the authority? Also, are these clearly marked so to avoid improper tagging? (i.e. preventing "This is an apartment in NYC, so it must be managed by the NYCHA")

akadouri:
It is the authority that owns the land & apartments but I recently learned they do not operate all the buildings, let me close this and rethink the tags. 

I don't think there's a risk of privately owned complexes being tagged as owned by the public housing.

PR #11646, opened 2025-12-09
Commit built JSON on pull request

Cj-Malone:
See https://github.com/osmlab/name-suggestion-index/commit/dc99d363e72f3834be0da89dfa28cadae8a4cc17#commitcomment-172375894 cleaning the JSON on the pull request helps people see how there changes are actually going to land, avoids maintainers having to do it periodically (and the git conflicts that may come with it) and helps with the git log (if people merging are doing squash and merge)

### DIFF
--- a/data/brands/shop/motorcycle.json
+++ b/data/brands/shop/motorcycle.json
@@ -20,16 +20,6 @@
       }
     },
     {
-      "displayName": "BMW",
-      "id": "bmw-7610a8",
-      "locationSet": {"include": ["001"]},
-      "tags": {
-        "brand": "BMW",
-        "name": "BMW",
-        "shop": "motorcycle"
-      }
-    },
-    {
       "displayName": "BMW Motorrad",
       "id": "bmwmotorrad-7610a8",
       "locationSet": {"include": ["001"]},
--- a/data/operators/power/line.json
+++ b/data/operators/power/line.json
@@ -8096,12 +8096,6 @@
       }
     },
     {
-      "displayName": "SER",
-      "id": "ser-116894",
-      "locationSet": {"include": ["001"]},
-      "tags": {"operator": "SER", "power": "line"}
-    },
-    {
       "displayName": "SEV",
       "id": "sev-636b4f",
       "locationSet": {"include": ["fo"]},


Cj-Malone:
I will be surprised if the workflow works first time, I never manage that.

Cj-Malone:
Seems to work. I would argue that there should be a build committed directly, then this main merged into this pr, just for a clean commit. But I'll leave that up to whoever reviews this. (cc @bhousel)

Cj-Malone:
```yaml
on:
  push:
    branches: [ main ]
  pull_request:
    branches: [ main ]
```

This shouldn't add a commit when on main, just PRs.

Cj-Malone:
Maybe this should be ran on main too, half the point is to keep main built.

The issue would be if main gets lots of unbuilt commits directly to it (not PRs and pushed separately), causing lots of build commits. IDK how frequently that happens.

data/brands/shop/telecommunication.json
```diff
@@ -655,6 +645,25 @@
...
+          "es-or.geojson",
+          "es-po.geojson"
+        ]
+      },
+      "tags": {
+        "brand": "R",
+        "brand:wikidata": "Q3326234",
+        "name": "R",
+        "office": "telecommunication",
+        "shop": "telecommunication"
```
Cj-Malone: Presumably you didn't want both tags

Cj-Malone:
`main` has been built recently by other people, and I have just cleaned some mistakes, there are more, but it's not worth crying about.

This PR is now exclusively for PRs, I think `main` should be kept good, but that is up to the people who have access to commit directly to it.

PRs (after/if this is merged) will often get an extra commit, generating the id, sorting the json file. If the PR does something incorrectly (remove without a match name/exclude, have a different category to the file) the commit will show that. If a PR isn't good, fix it in the PR before merging, if a PR is fine, merging should clean.

bhousel:
ok, we can try it

bhousel:
It looks like this does generate a reasonable diff and commit, but I'm seeing a permissions issue when trying to push that commit to the PR branch.

Screenshot 2025-12-29 at 10 49 02 AM

The original code did this too, I've since tried swapping in https://github.com/stefanzweifel/git-auto-commit-action (which does the same thing), and tried manually specifying the job to have write permissions, but I still get the error.

PR #11657, opened 2025-12-12
E->É in Bouygues Énergies & Services

Marc-marc-marc:


ilias52730:
https://www.bouygues-es.fr

Even though it's a French company, they brand themselves in English as _Bouygues Energies & Services_.

PR #11662, opened 2025-12-14
Enhance Republic of Türkiye Ministry of National Education

w2r4:
It had a typo (i vs î) and it was missing the `operator:en` tag.
Those tags are fixed using the data from the [wikipedia page](https://en.wikipedia.org/wiki/Ministry_of_National_Education_(Turkey)).

### DIFF
--- a/data/operators/amenity/school.json
+++ b/data/operators/amenity/school.json
@@ -31498,7 +31498,7 @@
         "operator": "Millî Eğitim Bakanlığı",
         "operator:en": "Ministry of National Education",
         "operator:short": "MEB",
-        "operator:type": "public",
+        "operator:type": "government",
         "operator:tr": "Millî Eğitim Bakanlığı",
         "operator:wikidata": "Q1359675"
       }


ilias52730:
Agree with the changes. I just noticed: all other "Ministry of Education" entries in the file have operator:type=government, not operator:type=public. What do you think?

Cj-Malone:
If adding operator:en, you should really add operator:tr. https://wiki.openstreetmap.org/w/index.php?title=Multilingual_names&mobileaction=toggle_view_desktop#Repeating_name_with_language_specific_tag

w2r4:
> Agree with the changes. I just noticed: all other "Ministry of Education" entries in the file have operator:type=government, not operator:type=public. What do you think?

I tried to classify government and public for ~20 minutes and I'm confused. OSM wiki defines them as:
> `public`: A public operator, thus operated by a public organisation linked to the government but which has its own director and management structure
> `government`: A governmental organisation is the operator, thus operated directly by a local, regional or country government, for example a city tax office, the city municipality, ... 

Do we have any well-known example for these? Something like "X thing is operated by Y which is linked to government thing Z"

---


> If adding operator:en, you should really add operator:tr. https://wiki.openstreetmap.org/w/index.php?title=Multilingual_names&mobileaction=toggle_view_desktop#Repeating_name_with_language_specific_tag

Learned something new today! I fixed it.

w2r4:
Although the schools are *public*, if we are speaking about the *operator type* and if the operator is the ministry, it seems reasonable tag it as "operator:type=government"

ilias52730:
> Do we have any well-known example for these? Something like "X thing is operated by Y which is linked to government thing Z"

For example, I can think of energy companies (e.g. EDF in France). In many countries they are state-owned (public companies) but it's not directly the government that operates them, it's like its own separate company, it's just "owned" by the government. So yes in the case of public schools, it's operator=government.

w2r4:
Ok, its government then. Let me fix it

PR #11672, opened 2025-12-18
Update school.json

monk198230:
added new MAT in NW England


bhousel:
It looks like this is just 8 schools around Liverpool.  It would be easier to just edit these 8 things in OSM to have the missing `operator` tags.

PR #11673, opened 2025-12-20
Update fast_food.json : Add Tasty Crousty

remsoff:


ilias52730:
Fixed Wikidata, changed `diet:halal=yes` to `diet:halal=only` and added `takeaway=yes`.

Issue #11142, opened 2025-07-12
There seems to be some kind of incorrect overlap between Goodwill presets.


Udarthegreat:
I could not tell you exactly when this started, but over the last month or so Goodwill POI's have been creating errors were they ask you to convert POI's of `Goodwill Industries of South Florida` to `Goodwill Industries of Central Florida` and `Goodwill Industries of North Florida` as seen in the image below (from iD) although `not:operator:wikidata=Q126897703;Q126845603` is tagged on the way and `Goodwill Industries of South Florida` seems to still exist in real life as their website is [still up](https://goodwillsouthflorida.org/) still up, as is the [wikidata](https://www.wikidata.org/wiki/Q54624094) page. From my understanding this should not be happening and I would consider this a bug. As an example POI see node [6020061550](https://www.openstreetmap.org/node/6020061550).  If I am missing something please feel free to correct me.

<img width="367" height="402" alt="Image" src="https://github.com/user-attachments/assets/8e002d00-ad9b-4682-a214-374879e6391e" />

Snowysauce:
Related to #11089 / https://github.com/openstreetmap/iD/issues/11146. This is an issue with iD, not the NSI.

bhousel:
It's probably related to this:  https://github.com/osmlab/name-suggestion-index/issues/5693#issuecomment-2819259226

iD/Rapid has code that was built on the assumption that a category could only live in one tree, and when we started having categories that could be indexed under either "brand" or "operator", it caused this issue where the match might be surprising.  

Before, it would primary-match brands on their "operator" tag, and by resorting the NSI files in #5693, now we are seeing it primary-match operators on their "brand" tag - so if it sees `brand=Goodwill` it probably just grabs the first "Goodwill"-like thing out of the index that it finds and suggests those tags.

Snowysauce:
A possible stop-gap solution for this specific case would be to add extra definitions to the Goodwill `locationSet`s using coordinate-defined radius areas. In New York State at least, the operating areas of the Goodwill organizations don't really overlap. Finger Lakes is mostly Rochester; Goodwill of WNY is centered on Buffalo; etc. I don't know if that's the case nationwide, but perhaps if we can narrow the number of Goodwill presets for any given area to two at most, `not:operator:wikidata` would work to suppress the warning for the "wrong" Goodwill?

impiaaa:
[...]

PR #11675, opened 2025-12-21
Add recycling brands

Oivo35:
Add 2 recycling brands with same name, but different capitalization.


bhousel:
How is this going to work in CZ when a user enters one of these?  Both names will match and the validator may try switching them.

Oivo35:
Unfortunately both entries have the same name. Would moving them to the operator tree help since they have different operators?

bhousel:
> Unfortunately both entries have the same name. Would moving them to the operator tree help since they have different operators?

Maybe - I guess I'm trying to understand the situation in the real world, and whether users will be confused by what the software does, and whether we are trying to map some detail about these bins (the operator) that doesn't matter much?

Each item in NSI is scoped by its name, tags, and location - so if we multiple items that match the same name, tags, and location (all of CZ), the software won't be able to identify the two items.  If they occur in different parts of CZ, then it would work fine (like Aldi in Germany, there are 2 Aldis, but they are geographically distinct).  

(This is similar to the problem we had with Goodwill operators #11142 and others - we weren't able to make this work).

Oivo35:
There are 2 different clothing collecting projects throughout the Czech Republic
1. Textileco - [entry](https://nsi.guide/index.html?t=operators&k=amenity&v=recycling#textileco-778b5b) in NSI

turn</a>

3. TextilEco - not in NSI

turn</a>

Should this entry be removed from the NSI to prevent false positives?

Dimitar5555:
You could change one of the display names to "Textileco (Český červený kříž)" which should be enough to differentiate them IMO.

PR #11676, opened 2025-12-21
Updated few amenties in Armenia

MKo-xx:
1. Armbusinessbank was rebranded as AMIO.
2. Rostelecom was rebranded as OVIO.
3. HSBC Armenia was sold merged with Ardshinbank and no longer exist.
4. Added Team Energy car charging chain.


bhousel:
probably needs matchNames for all of these old names

MKo-xx:
> probably needs matchNames for all of these old names

Added match names for old names

PR #11680, opened 2025-12-22
Add ION Hotels entry to hotel.json

LuisAPI:
Adds the ION Hotels chain to OSM Name Suggestion Index, which includes multiple branches labeled with descriptions of either the location or the vibe (or marketing strategy) the branch is aiming for.


bhousel:
As far as I can tell, it's just 2 hotels.  This isn't the sort of thing we would add to this index.

PR #11684, opened 2025-12-23
Adding and updating school operators in Southernwestern British Columbia

ShinyDog2636:
Updated existing entries and added missing entries for schools operated in the Lower Mainland. Several school districts use a different operating name than their official name so I captured both names, as well as any acronyms.

I may have over specified the matchnames. I anticipate many different names to be tried by mappers so tried to consider several variations. I included in the matchnames their operating name (tagged as operator:short) since I am not sure if it would be suggested by default.

### DIFF
--- a/data/operators/amenity/school.json
+++ b/data/operators/amenity/school.json
@@ -56380,7 +56380,7 @@
         "amenity": "school",
         "operator": "School District Number 44 (North Vancouver)",
         "operator:short": "North Vancouver School District",
-        "operator:abbr": "NVSD"
+        "operator:abbr": "NVSD",
         "operator:type": "public",
         "operator:wikidata": "Q7432191"
       }


Cj-Malone:
School District 67 Okanagan Skaha and School District No. 43 will be readded when `bun run build` is ran because they weren’t added as match names or excludes.

Cj-Malone:
Also this is just confusing. It's strange that operator is now basically a ref and operator:short is a clean name, but not a short version of name.

bhousel:
Oh for what it's worth, I put `"skipCollection": true` on all the school-like categories yesterday in 9ce796eeb  (see #11597) so at least they won't be re-added anymore.  I'm on-the-fence about whether we should even be including school districts at this point.

bhousel:
Agree that the tagging is kind of weird - `operator:short` usually is the abbreviation, and `operator:abbr` is a new thing I've never seen.

ShinyDog2636:
Thanks for your comments. It was confusing to me, too, and why I included multiple names for some school districts.

All the school districts in BC have an official government name. Some only use this official name in certain legal contexts and use a different name for all other purposes (operating, signage, etc).

I originally prepared this with `operator:official_name`=government name and `operator`=signed name but didn't like the inconsistency across districts. Any school district that has not adopted a different operating name would not have an operator:official_name.

Would this be preferred? Thank you.

Edit: I am also okay if the direction of this project is to no longer include school districts and doesn't find this debate/discussion worth pursuing.

PR #11698, opened 2025-12-27
network=VRS and network=VRM

### LINKED ISSUE #11387: network=VRS and network=VRM
Until May 2025, all bus lines in the VRS area were mapped with network=VRS.
Since May 2025, the Id Editor changes this to “Verkehrsverbund Rhein-Sieg.”
This is supposed to be due to the NSI.
How can it be that an established mapping in the NSI was incorrectly stored and is now being continuously changed via the Id Editor?
Can this please be fixed? Because all previous Overpass queries no longer return the usual results.
Incidentally, the same problem exists in VRM. In VRR, however, this is not the case. There, the abbreviation is accepted as correct.
I see no reasonable reason why VRR, VRS, and VRM should not be treated equally. Everyone is talking about the transport associations with abbreviations  and the wiki says that the abbreviation is allowed in this case.

Translated with DeepL.com (free version)

ilias52730:
Closes #11387

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -22478,14 +22478,13 @@
       }
     },
     {
-      "displayName": "Verkehrsverbund Rhein-Mosel",
-      "id": "verkehrsverbundrheinmosel-b96e87",
+      "displayName": "VRM",
       "locationSet": {
         "include": ["de-rp.geojson"]
       },
+      "matchNames": ["verkehrsverbund rhein-mosel"],
       "tags": {
-        "network": "Verkehrsverbund Rhein-Mosel",
-        "network:short": "VRM",
+        "network": "VRM",
         "network:wikidata": "Q2516481",
         "route": "bus"
       }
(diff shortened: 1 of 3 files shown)


kjonosm:
Before deploying this change, please discuss in the German OSM forum.

bernwardhanssen:
> Before deploying this change, please discuss in the German OSM forum.

I raised this issue at the "OSM Stammtisch" in Bonn.
They agreed with me.

I think that was the right place to clarify something so local.

bhousel:
bump.. this has been sitting for a while, is it ok to merge?

kjonosm:
unfortunately there are several discussions in the forum on how to deal with the issue, all without a clear consensus...

https://community.openstreetmap.org/t/problematische-id-vorschlage-fur-network-tag/136198/

https://community.openstreetmap.org/t/what-is-the-definition-of-a-public-transport-network/117250/

https://community.openstreetmap.org/t/verkehrsverbunde-taggen/91491

PR #11706, opened 2025-12-29
Remove GitHub Action build-pr

Cj-Malone:
Rel: https://github.com/osmlab/name-suggestion-index/pull/11646#issuecomment-3696882067


bhousel:
Hey cool, it looks like this worked!
Maybe the errors I saw were just because they were older PR branches.

Cj-Malone:
I'm not that optimistic, I see 2 errors on the previous actions, checkout and push.

https://github.com/osmlab/name-suggestion-index/actions/runs/20575891067/job/59092694308
https://github.com/osmlab/name-suggestion-index/actions/runs/20577248563/job/59096946324

I wonder if the push is because was on forks, not a branch within the repo. Checkout, IDK.

Cj-Malone:
Push may or may not be a permissions thing https://stackoverflow.com/questions/73687176/permission-denied-to-github-actionsbot-the-requested-url-returned-error-403#answer-75175628

Cj-Malone:
So checkout of forks is fixed, but it's basically impossible to commit to them (without secrets everywhere). We could:

1. Switch to a comment on the PR :/
2. Delete it

I'm not in love with comments, especially because main will still end up in an unbuilt state, and the comments will be too noisy.

.github/workflows/build-pr.yml
```diff
@@ -1,7 +1,7 @@
 name: build pr
 
 on:
-  pull_request:
+  pull_request_target:
```
Cj-Malone: This is naughty if NSI has secrets in GitHub, they can be leaked by PRs

.github/workflows/build-pr.yml
```diff
@@ -32,7 +32,16 @@ jobs:
         echo "::remove-matcher owner=eslint-compact::"
         echo "::remove-matcher owner=eslint-stylish::"
     - run: bun run all
-    - run: git status
-    - uses: stefanzweifel/git-auto-commit-action@v7
+    - run: |
+        DIFF=$(git diff data/)
+        echo "DIFF<<EOF" >> $GITHUB_ENV
+        echo "$DIFF" >> $GITHUB_ENV
+        echo "EOF" >> $GITHUB_ENV
```
Cj-Malone: I don't really understand this, but it's what GH needs for multiline env variables: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#multiline-strings

.github/workflows/build-pr.yml
```diff
@@ -32,7 +32,16 @@ jobs:
...
+    - name: Comment PR
+      if: ${{ env.DIFF }}
+      uses: thollander/actions-comment-pull-request@v3
       with:
-        add_options: '-A'
+        message: |
+          ```diff
```
Cj-Malone: Some words would be good

Cj-Malone:
```suggestion
          The following changes will be applied before NSI is released, your changes may have unintended consequences.
          ```diff
```

Cj-Malone:
Spending the "morning" on this has demotivated me a bit, but I would still like to see main be constantly clean, or close to it.

Having another action build main and open a PR for review (or commit straight to main) every day/week/hour should be significantly easier, I've done similar before, it's possible. It shouldn’t get in the way of rapidly merging PRs occasionally, and if the existing action is changed to comment, the comments would stay relevant.

PR #11708, opened 2025-12-29
Add 4all brand (Greece)

geraki:
Convenience/Coffee shops with 147 locations.


Cj-Malone:
Parsing OSM data with mixed categories is awful, I don't think NSI should encourage it.

geraki:
> Parsing OSM data with mixed categories is awful, I don't think NSI should encourage it.

Any other approach would violate [one feature, one OSM element](https://wiki.openstreetmap.org/wiki/One_feature,_one_OSM_element). In this case there is an advantage that no semicolon operator is needed, only the appropriate tags.

bhousel:
I merged but kept it as just `shop=convenience`.. I agree that mixing toplevel tags can lead to problems.

PR #11716, opened 2026-01-02
[shop/truck][shop/truck_repair] change Wikidata for Mercedes-Benz Trucks and add repair

panasonic-mafia:
Previous wikidata is a list of Mercedes-Benz trucks ([Q1586306](https://www.wikidata.org/wiki/Q1586306)), it's probably better to link nsi entry to just Mercedes-Benz ([Q36008](https://www.wikidata.org/wiki/Q36008)).


bhousel:
Thanks, good catch!

PR #11718, opened 2026-01-02
Fix typo in brand name for Tzomet Sfarim

mattanelkaim:
The value for the `name:en` tag was missing an 'i', although it exists in the `brand:en` tag.
I fixed it to avoid false warnings in OSM iD editor that suggest "fixing" mapped elements with an incorrect name.


mattanelkaim:
Hi @bhousel, when is it anticipated to be updated in the DB? It keeps warning in iD editor, and still not updated through [the website](https://nsi.guide/index.html?t=brands&k=shop&v=books&tt=tzomet).

PR #11724, opened 2026-01-04
Add tag education=kindergarten

cuatim:
Hi,

As by recent [proposal](https://wiki.openstreetmap.org/wiki/Proposal:Education_tag), the education key should be added to all relevant amenities. This PR adds it to all kindergarten.

Thanks :)

### DIFF
--- a/data/brands/amenity/kindergarten.json
+++ b/data/brands/amenity/kindergarten.json
@@ -19,6 +19,7 @@
       "matchNames": ["asb"],
       "tags": {
         "amenity": "kindergarten",
+        "education": "kindergarten",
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks for the suggestion, but we don't need to add tags like that here, because the presets for this project inherit from the ones released by https://github.com/openstreetmap/id-tagging-schema .

You can go there to search for the status of the `education` tag.  As far as I know, the tag got approved via the proposal process, then they added it to the presets, then people complained, and now they are trying to un-add it.  It's a mess.

cuatim:
Ah, I was not aware. Thanks! :)

PR #11731, opened 2026-01-05
add service=commuter tag to some networks

aaei924:
added service=commuter tag to commuter train routes.

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -1241,6 +1241,7 @@
         "operator:short": "KORAIL",
         "operator:type": "public",
         "operator:wikidata": "Q18169",
+        "service": "high_speed",
         "route": "train"
       }
     },


bhousel:
I haven't seen the `service=` tag used this way before.  I'm not really comfortable accepting a pull request that adds a tag like this that's not consistent with what's written here:  https://wiki.openstreetmap.org/wiki/Key:service .  Is there some discussion where other people have said they are ok with this?

aaei924:
> I haven't seen the `service=` tag used this way before. I'm not really comfortable accepting a pull request that adds a tag like this that's not consistent with what's written here: https://wiki.openstreetmap.org/wiki/Key:service . Is there some discussion where other people have said they are ok with this?

That key on train routes seems to be properly used. You can find out at both [Train routes section of Key:service](https://wiki.openstreetmap.org/wiki/Key:service#Train_routes) and [route=train](https://wiki.openstreetmap.org/wiki/Tag:route%3Dtrain) pages.

LaoshuBaby:
@aaei924 I don't think this is a good idea because we can't assume the train route only focus on commuter, even the duck really looks like a commuter route. I think you can add those tags to routes and network relation manually with review. (And another reason is transport's mapping method still have many discussion)

And another way is that you can see other countries' major commuter system also recorded in NSI's preset, such as DB, or many private railway company in Japan (obviously they are commuter route, but I think NSI isn't a encyclopedia about everything, it should focus on detect "who is this" and tag with something core tag that let you match it with a network. For example we seldom add vegan/vegetarian and seldom add )

----
So I tend to not add this tag at this time

LaoshuBaby:
> I haven't seen the `service=` tag used this way before. I'm not really comfortable accepting a pull request that adds a tag like this that's not consistent with what's written here: https://wiki.openstreetmap.org/wiki/Key:service . Is there some discussion where other people have said they are ok with this?

such of tag service=yard/siding/spur is for describe the railway infrastructure on a way or a relation with type=route+route=railwway

And service=commuter is used on a relation with type=route+route=something tag, describe the stop, platform and other thing on transport operation level. (Especially for those train routes across different railway).

I thinks this is a shortage in OSM key design, like branch=* can both have meaning in chain shop and railway.

data/transit/route/train.json
```diff
@@ -1241,6 +1241,7 @@
         "operator:short": "KORAIL",
         "operator:type": "public",
         "operator:wikidata": "Q18169",
+        "service": "high_speed",
```
LaoshuBaby, line 1244: If so I still think this is unnecessary to record this, I think NSI preset shouldn't describe whether a train route is fast or slow

aaei924:
> @aaei924 I don't think this is a good idea because we can't assume the train route only focus on commuter, even the duck really looks like a commuter route. I think you can add those tags to routes and network relation manually with review. (And another reason is transport's mapping method still have many discussion)
> 
> And another way is that you can see other countries' major commuter system also recorded in NSI's preset, such as DB, or many private railway company in Japan (obviously they are commuter route, but I think NSI isn't a encyclopedia about everything, it should focus on detect "who is this" and tag with something core tag that let you match it with a network. For example we seldom add vegan/vegetarian and seldom add )
> 
> So I tend to not add this tag at this time

Okay. I understood what NSI intends. I'll close this PR and make further contributions considering roles and purposes of NSI.

PR #11732, opened 2026-01-06
[amenity/cafe] remove Ikea RU entry

panasonic-mafia:
Closed in RU: https://www.ikea.com/gb/en/customer-service/knowledge/articles/57geb729-2fe9-4fe9-994d-b6ce49g125cg.html

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -5196,23 +5196,6 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
Thank you for reporting this, after this you may need to re-check current OSM data about their name and brand, or set life-cycle tag on vital tag "amenity=cafe" or it may been collected to this registry again by automatically scripts.

PR #11735, opened 2026-01-06
Blank out short_name and offical_name in templates when name is also removed

Cj-Malone:
We currently end up with some weird entries eg alt_name on post boxes, and no name.


bhousel:
Yeah this makes sense.  When I made templates I wasn't really thinking of all the name tags that might not want to be copied, it was more about "make this thing a copy of this other thing".

There are other related issues because it's a problem that keeps coming up, #10414, #10378, #9260, etc.   😓

Cj-Malone:
Thank you

PR #11747, opened 2026-01-10
Further Japanese regions

HayaKoh-WeldyAlin:
Add regions larger than prefectures (or smaller in the case of Hokkaido) that are commonly used in Japan.

- Soya (宗谷)
- Rumoi (留萌)
- Kamikawa (上川)
- Okhotsk (オホーツク)
- Nemuro (根室)
- Kushiro (釧路)
- Tokachi (十勝)
- Sorachi (空知)
- Ishikari (石狩)
- Shiribeshi (後志)
- Iburi (胆振)
- Hidaka (日高)
- Oshima (渡島)<sup>※1</sup>
- Hiyama (檜山)<sup>※1</sup>

※1 Both include the former Kumaiishi (熊石町, now Yakumo (八雲町)).

- Tohoku region (東北地方)
- Kanto region (関東地方)
- Chubu region (中部地方)
- Hokuriku region (北陸地方)
- Tokai region (東海地方)
- Kinki/Kansai region (近畿/関西地方)
- Chugoku region (中国地方)
- Shikoku region (四国地方)
- Kyushu region (九州地方)

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -6124,6 +6124,18 @@
       }
     },
     {
+      "displayName": "Solo Oil",
+      "id": "solo-9ed9c0",
+      "locationSet": {"include": ["au"]},
+      "matchNames": ["solo oil"],
+      "tags": {
+        "amenity": "fuel",
+        "brand": "Solo Oil",
+        "brand:wikidata": "Q115517074",
+        "name": "Solo Oil"
+      }
+    },
+    {
       "displayName": "Sonangol",
       "id": "sonangol-a1d251",
       "locationSet": {"include": ["ao"]},


LaoshuBaby:
Adding a region means it need to be loaded everytime when a user open the iD or other editors, and don't related to where you are. So I think add detailed administrative region geojson of Hokkaido will contribute less to make NSI preset more helpful for osm mapper 

And the another reason is Hokkaido 's population is decreasing, I don't think there are such many brands need to be tag with this level, just add jp-01 is enough
(I guess even for Tokyo, create geojson files like jp-13-shinjuku, jp-13-toshima is still too detailed)

LaoshuBaby:
> So I think add detailed administrative region geojson of Hokkaido will contribute less to make NSI preset more helpful for osm mapper

But for those larger area such as shikoku or kyushu will be useful for brands not around the country, for example, 551 horai in kansai (or kinki?)

This was raise not only 1 time by me and other Japanese mapper but until now nothing was made because some area are hard to define or included by another one, example is above kansai or kinki area. Or, will 山阳 山阴 be calculated? I hope to find a list that define which area contain which prefecture exactly.

HayaKoh-WeldyAlin:
> So I think add detailed administrative region geojson of Hokkaido will contribute less to make NSI preset more helpful for osm mapper

> And the another reason is Hokkaido 's population is decreasing

I certainly agree with that view. However, Hokkaido is simply too vast compared to other prefectures, hence the existence of [subprefectures](https://en.wikipedia.org/wiki/Subprefectures_of_Japan).
This is a larger framework than cities (such as Sapporo) or towns (such as Niseko), and is effectively represented in a context akin to a prefecture to denote a location.
I myself have not seen any concrete examples, so I cannot say for certain, but I believe it would be effective in describing chain stores operating within Hokkaido that are limited to specific subprefectures.
(However, this may only be effective in the southern regions such as Ishikari and Tokachi.)

HayaKoh-WeldyAlin:
> I hope to find a list that define which area contain which prefecture exactly.

Unfortunately, it appears that no authoritative legal documents exist on this point.
However, as the eight regional divisions (Hokkaido, Tohoku, Kanto, Chubu, Kinki, Chugoku, Shikoku, Kyushu) are widely used in Japanese school education, many Japanese people share this sense.
For more granular regions such as Hokuriku and Tokai (which nonetheless encompass several prefectures), the divisions used are largely consistent with those employed by various organisations, including JR, sports bodies, and national agencies (such as the Ministry of Land, Infrastructure, Transport and Tourism).
The final decision was based on the page linked below (where multiple candidates existed, the broader classification was adopted).
https://ja.wikipedia.org/wiki/%E6%97%A5%E6%9C%AC%E3%81%AE%E5%9C%B0%E5%9F%9F

LaoshuBaby:
> This is a larger framework than cities (such as Sapporo) or towns (such as Niseko), and is effectively represented in a context akin to a prefecture to denote a location.

locationSet have a feature that can let you select a circle area with a central point, for example if want to describe commercial near Sapporo city you can write like this (and add a "note" in the entry)

https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#locationset

HayaKoh-WeldyAlin:
> Or, will 山阳 山阴 be calculated?

Should it prove necessary, I may create this region in future; however, these issues are more complex than those encountered with the region created this time.
Several categories were similarly omitted for similar reasons.

- Sanyo region (山陽地方)
- Sanin region (山陰地方)

It is difficult to determine whether Hyogo Prefecture and Kyoto Prefecture are included.
In cases where this is included (which is historically accurate), it would divide the prefecture into northern and southern parts; however, consideration must be given to which region would be most effective for the division.
There is also debate over whether Sanyo should include Osaka, and this too requires careful consideration as it necessitates disregarding prefectural boundaries (indeed, even municipal boundaries may be involved).

- Northern Hokkaido region (道北地方)
- Eastern Hokkaido region  (道東地方)
- Central Hokkaido region (道央地方)
- Southern Hokkaido region (道南地方)

Examples of its use can be seen in various places, but its definition is less established than in other regions.
Using this will reduce the number of regions (such as Soya and Hidaka) that are underutilised and go to waste, but to employ it, a clear definition must be sought.

- Other, Kanto-Koshin region (関東甲信地方) etc

For those cases that can be reproduced simply by adding a few prefectures to the newly added region, I believe there is no need to introduce them at all.
However, should a need arise for more granular regions such as Northern-Kanto (北関東), I may create them.

HayaKoh-WeldyAlin:
> for example if want to describe commercial near Sapporo city you can write like this (and add a "note" in the entry)

While this might be feasible within a single city, many subprefectures in Hokkaido (e.g., [Tokachi Subprefecture](https://en.wikipedia.org/wiki/Tokachi_Subprefecture)) are considerably larger than many [prefectures](https://en.wikipedia.org/wiki/Prefectures_of_Japan), meaning this may not always be the wisest choice.

LaoshuBaby:
> Should it prove necessary, I may create this region in future; however, these issues are more complex than those encountered with the region created this time.

Yes and this is also what I concerned, I even have seen the "甲信越" for several times after I move to Japan, and if people want they even can find some smaller area name, so to prevent add too much of them I guess find a case for those "customized" region is important? We may not add all of them, but, if find a brand can described more detailed and don't need to write a complex location set, if we add a region, then I guess the region is well for merge? 

> * Other, Kanro-Koshin region (関東甲信地方) etc

obviously kanto is more common used XD

> For more granular regions such as Hokuriku and Tokai (which nonetheless encompass several prefectures), the divisions used are largely consistent with those employed by various organisations, including JR, sports bodies, and national agencies (such as the Ministry of Land, Infrastructure, Transport and Tourism).

image

https://github.com/ffff-23/Japan-regions/compare/main...LaoshuBaby:Japan-regions:main

Indeed I also want to add those area for organizations, JR, NTT, NEXCO, and we truly can find brand related to them. (I want to draw some polygon file and then make a pr to ffff-23's repository, looks like I forgot to do this.)

NTT-West and NTT-East region was already included in our repository.

HayaKoh-WeldyAlin:
> but, if find a brand can described more detailed and don't need to write a complex location set, if we add a region, then I guess the region is well for merge?

If it consists of only one to three prefectures, it may not be worth adding in the first place.
However, for chains spanning regions that cannot be combined using any single term (e.g. Gifu, Aichi, Mie, Shiga, Kyoto, Osaka, Nara), detailed classification is necessary to prevent redundancy and define the scope more accurately. (Organised examples: Tokai, Kinki)

> Kanro-Koshin region (関東甲信地方)

It's just a simple spelling mistake. Don't mind it. (It's the fault of t and r being next to each other.)

> Indeed I also want to add those area for organizations, JR, NTT, NEXCO, and we truly can find brand related to them.

Such clearly defined categories are often more intuitive, so I believe it would be a good idea to actively add them.
Whilst a complete replacement cannot be expected due to the occasional presence of peculiar divisions...

LaoshuBaby:
> If it consists of only one to three prefectures, it may not be worth adding in the first place.

> However, for chains spanning regions that cannot be combined using any single term (e.g. Gifu, Aichi, Mie, Shiga, Kyoto, Osaka, Nara), detailed classification is necessary to prevent redundancy and define the scope more accurately. (Organised examples: Tokai, Kinki)

For example, the convenience seicomart have branch shop in Hokkaido, and Saitama, and Ibaraki, this can't been included by a single JP-Hokkaido or JP-Kanto, so I recommend use JP-01(Hokkaido) + JP-11(Saitama) + JP-08(Ibaraki) to describe its location, and it will not recommended to other Kanto area mapper, such as Chiba. 551 and seicomart are both my favourite to tell people how to combine proper value and those 2 are known by many people in Japan.

HayaKoh-WeldyAlin:
> https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#notability

What was the purpose of posting this link?
If you mean that Wikidata is required, then it exists.

- [Tohoku (東北)](https://www.wikidata.org/wiki/Q129465)
- [Kanto (関東)](https://www.wikidata.org/wiki/Q132480)
- [Chubu (中部)](https://www.wikidata.org/wiki/Q134638)
- [Hokuriku (北陸)](https://www.wikidata.org/wiki/Q381025)
- [Tokai (東海)](https://www.wikidata.org/wiki/Q398989)
- [Kinki/Kansai (近畿/関西)](https://www.wikidata.org/wiki/Q164256)
- [Chugoku (中国)](https://www.wikidata.org/wiki/Q127864)
- [Shikoku (四国)](https://www.wikidata.org/wiki/Q60213044)
- [Kyushu (九州)](https://www.wikidata.org/wiki/Q13393883)

Regardless of whether the intention is correct, I think using Wikidata as the criteria for judgement is a good idea.

HayaKoh-WeldyAlin:
> If it consists of only one to three prefectures, it may not be worth adding in the first place.

Ah, I see. I finally understand. Here, “not be worth adding” was intended to mean “locationset”.
I should add that this does not mean that NSI does not need to be added unless it is a national chain.
I apologise if any parts are unclear, as I have relied on machine translation.

bhousel:
Just checking in - is this ok to merge?  I haven't reviewed the files carefully.  It's ok to add custom geojsons if it makes sense to regionally restrict some of the brands, and as long as the files aren't too large/detailed.

HayaKoh-WeldyAlin:
This data is fully grounded in the rationale for the scope setting and the necessary voices from the community.

> and as long as the files aren't too large/detailed

However, there was discussion regarding sections that might be overly detailed.
I have submitted a pull request as I considered it acceptable, but there may still be points requiring discussion.
(Although I do not believe there is a problem, as there has been no progress in the discussions for two weeks)

GA-Kevin-Codes:
I'm of the mind that these 100-liners wouldnt significantly impact NSI performance. It seems reasonably well formatted and discussed, and the concerns were more policy than format so I will merge this PR and if we start to have issues we know where to look. Thank you @HayaKoh-WeldyAlin for your work on this and engagement on this thread.

PR #11751, opened 2026-01-12
Add matchNames for The Bottle-O

peternewman:


peternewman:
I do sort of wonder if this should be done automagically and centrally via NSI, rather than expecting the end user to, but then you'd need a list of suitable words to drop to generate them...

Cj-Malone:
"bottleo" and "the bottleo" are handled automatically by matchers (and will be removed by the NSI build process (`bun run build`))

peternewman:
> "bottleo" and "the bottleo" are handled automatically by matchers (and will be removed by the NSI build process (`bun run build`))

Do you mean this should just be:
`      "matchNames": ["bottle-o"],`

Also where is the code set, I couldn't obviously find it, but obviously searching for `the` is quite hard!

Specifically my issue was with StreetComplete which obviously uses different matching code to iD.

Cj-Malone:
I'm not sure where "the" is removed either, but I encourage you to run `bun run build` and see the output.

peternewman:
> I'm not sure where "the" is removed either, but I encourage you to run `bun run build` and see the output.

You are of course correct @Cj-Malone 
https://github.com/osmlab/name-suggestion-index/commit/68a8aa857c6a4f48a6193cbc27c8367df8ef8859#diff-0372c5c3e1bbb4bc865e629d1792f764eb50b6791af43942e7264b6e974ff16bL883

https://github.com/osmlab/name-suggestion-index/blob/8ff831be6b7fb38efe1ac1cd69f84eb6d55cd82d/data/brands/shop/alcohol.json#L883

Do you know where the "the" is removed @bhousel so we could get the same behaviour as NSI/iD in other tools consuming the data?

Cj-Malone:
Oh, the brand includes "The" so is probably normalised to "the bottleo". I can check when I'm next at a computer

PR #11752, opened 2026-01-12
Add Laser Clinics entry to beauty.json

peternewman:


data/brands/shop/beauty.json
```diff
@@ -319,6 +319,18 @@
...
       }
     },
+    {
+      "displayName": "Laser Clinics",
+      "id": "",
+      "locationSet": {"include": ["au", "ca", "gb", "nz"]},
+      "tags": {
+        "beauty": "hair_removal",
+        "brand": "Laser Clinics",
+        "brand:wikidata": "",
```
peternewman: Please could someone magic up some Wikidata for this.

kjonosm: Is this https://www.wikidata.org/wiki/Q126539797?

peternewman: Yes, sorry @kjonosm I think I got sidetracked and didn't even go looking for Wikidata.

I guess the follow on is this item is for "Laser Clinics Australia" and mentions them being Australian. All the other sites have the same corporate history as them originating in Australia first, so is it one NSI entry with that Wikidata, and maybe just broaden the scope of the Wikidata/add the other country URLs, or should it be one per country or something (but then Canada only has a handful of clinics)?

bhousel:
```suggestion
        "brand:wikidata": "Q126539797",
```

PR #11755, opened 2026-01-12
Add new ice-cream vending machine brands and details

peternewman:
Ben & Jerry's and Häagen-Dazs


peternewman:
Both vending machines seen in the UK at least, but their respective locationsets copied from their main shop brands...

PR #11756, opened 2026-01-12
Add Damir billboard

Mannivu:
Add Italian operator Damir ([website](https://www.damir.it))

### DIFF
--- a/data/operators/advertising/billboard.json
+++ b/data/operators/advertising/billboard.json
@@ -401,6 +401,14 @@
       }
     },
     {
+      "displayName": "Damir",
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Without wikidata, this wont make it into iD.

Mannivu:
I suspected this, but I've seen that most of the other billboard operators are without Wikidata, so I thought I could give it a try.

Cj-Malone:
They'll have been added a long time ago, due to being in OSM, or via PRs. They'll make it to nsi.guide, but not iD.

PR #11767, opened 2026-01-20
add AU medical imaging providers

andrewharvey:


bhousel:
Oh, it looks like `healthcare=medical_imaging` is a [proposed tag](https://wiki.openstreetmap.org/wiki/Proposal:Medical_Imaging), not approved yet.

I've noticed we already have a bunch of imaging and radiology facilities under `amenity=clinic` [`brands/amenity/clinic.json`](https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/amenity/clinic.json) - maybe these should go there for now?

andrewharvey:
> Oh, it looks like healthcare=medical_imaging is a [proposed tag](https://wiki.openstreetmap.org/wiki/Proposal:Medical_Imaging), not approved yet.

Do tags need to be approved to be used here? If that's the case we should update the [contribution guidelines](https://github.com/osmlab/name-suggestion-index/blob/main/CONTRIBUTING.md).

Wouldn't a draft PR imply I'm not finished with the PR, I think best to mark it as ready for review, but add the label to `waitfor` [the tag to become more accepted], because from my side the PR is good to go.

> I've noticed we already have a bunch of imaging and radiology facilities under amenity=clinic [brands/amenity/clinic.json](https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/amenity/clinic.json) - maybe these should go there for now?

I'd rather not have these items in the index than to to suggest mappers use the wrong tags, or worse suggest mappers change the tags to the incorrect tag. That would also then make it harder to get the tag more accepted and used, because the usage numbers would be pushed lower due to this.

Certainly some `healthcare=clinic` would provide in-house medical imaging, but these entries are for external medical imaging centres that customers visit directly.

bhousel:
> Do tags need to be approved to be used here? If that's the case we should update the [contribution guidelines](https://github.com/osmlab/name-suggestion-index/blob/main/CONTRIBUTING.md).

Yes, I think tags should be approved before NSI encourages their use, or we are likely to catch complaints from people.

> Wouldn't a draft PR imply I'm not finished with the PR, I think best to mark it as ready for review, but add the label to waitfor [the tag to become more accepted], because from my side the PR is good to go.

Ok

> I'd rather not have these items in the index than to to suggest mappers use the wrong tags, or worse suggest mappers change the tags to the incorrect tag. 

That works for me too.. We can put the PR on hold if you aren't comfortable using the `amenity=clinic` tag.

PR #11786, opened 2026-01-27
Enhance Clinic Operators

mattanelkaim:
- Added Wikidata code & name:en to Meuhedet
- Added new operator entry - Leumit (without an `id` field)


mattanelkaim:
@kjonosm Thanks for merging! How long until the changes take effect?

PR #11797, opened 2026-01-31
Stonegate brands

TheUKHighStreet:
Add Stonegate as operator to Stonegate brands.


TheUKHighStreet:
Not sure what happened here. I thought I'd already added these but something obviously went wrong and 2 pull request have become merged.

One is just adding Stonegate as an operator for the pubs and bars it runs.

The other is adding different tags for TGJones shops.

GA-Kevin-Codes:
Hey @TheUKHighStreet, the Stonegate entries look good to go. As far as the TGJones shops, please only include it in the book shop file: data/brands/shop/books.json. The rest are incorrectly placed.

data/brands/shop/gift.json
GA-Kevin-Codes, line 1: This does not belong in the gift shop file based on the tag `shop=books`.

data/brands/shop/stationery.json
GA-Kevin-Codes, line 1: This does not belong in the stationary shop file based on the tag `shop=books`.

GA-Kevin-Codes (changes requested):

TheUKHighStreet:
Thanks - I've removed it from gifts and stationery. I think a lot of WHSmith shops were tagged as stationery (but I'm not sure how to find out how many) so when they are changed to TGJones, the brand tag isn't being added.

TheUKHighStreet:
I've removed all the TGJones edits and will start again with a new pull request just for that.

PR #11804, opened 2026-02-03
Updated PubliBike Velospot

dafadllyn:
Updated PubliBike Velospot:

- new brand name: PubliBike → PubliBike Velospot
- official company name: PubliBike SA (in French) / PubliBike AG (in German)
- removed discontinued service in Lausanne
- new network name in Bern: Velo Bern → Velo Region Bern

### DIFF
--- a/data/brands/amenity/bicycle_rental.json
+++ b/data/brands/amenity/bicycle_rental.json
@@ -3271,7 +3271,7 @@
         "brand": "PubliBike Velospot",
         "brand:wikidata": "Q138003595",
         "network": "Agglo Fribourg-Freiburg",
-        "operator": "PubliBike SA",
+        "operator": "PubliBike AG",
         "operator:type": "private",
         "operator:wikidata": "Q3555363"
       }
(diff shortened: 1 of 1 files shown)


kjonosm:
If PubliBike SA is different from PubliBike AG they shouldn't have the same Wikidata ID...

dafadllyn:
> If PubliBike SA is different from PubliBike AG they shouldn't have the same Wikidata ID...

It's one single company, see the [company's entry](https://zefix.ch/en/search/entity/list/firm/967437) in the Swiss Central Business Name Index.

'SA' is the French abbreviation for 'société anonyme' and 'AG' the German abbreviation for 'Aktiengesellschaft', both meaning 'joint-stock company'. In Switzerland, where there are four official national languages, it is quite common for a company to have more than one name. The [Swiss Federal Railways](https://en.wikipedia.org/wiki/Swiss_Federal_Railways) even have five official names, one in each official national language and one in English (see https://zefix.ch/en/search/entity/list/firm/462042).

dafadllyn:
I've now changed PubliBike SA to PubliBike AG (which is the first listed name of the company in the Central Business Name Index) to avoid confusion.

Although not correct (or official), the 'AG' can be omitted as far as I am concerned, as is currently the case.

PR #11814, opened 2026-02-08
fix: run build

Johannes-Andersen:
Not sure what the correct process is here, just noticed when building my other PR #11813, it caused a bunch of noice as there seems to be a lot of uncommited lint fixes/missing builds


Cj-Malone:
Thank you, I personally think main should be kept cleaner.

PR #11817, opened 2026-02-09
Refine geography of Pizza Inn

peternewman:
The wikidata is for:
https://www.pizzainn.com/

The pages for even the most obvious overlap have very different history and branding, I think it's just a very popular name/lacking in strong trademarking
https://www.pizzainn.com/about-us/
https://www.pizzainn.com.au/about-us/

Presumably the original matcher would have just found lots of similar named stores across the world?

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -8265,7 +8265,7 @@
       }
     },
     {
-      "displayName": "Pizza Inn",
+      "displayName": "Pizza Inn (America)",
       "id": "pizzainn-658eea",
       "locationSet": {"include": ["us"]},
       "tags": {


data/brands/amenity/fast_food.json
```diff
@@ -8267,7 +8267,7 @@
     {
       "displayName": "Pizza Inn",
       "id": "pizzainn-658eea",
-      "locationSet": {"include": ["001"]},
+      "locationSet": {"include": ["us"]},
```
peternewman, line 8270: We could narrow to particular states, but I'm not sure if that's worth it?
https://www.pizzainn.com/location-finder/

data/brands/amenity/fast_food.json
```diff
@@ -8267,7 +8267,7 @@
     {
       "displayName": "Pizza Inn",
```
peternewman: Should I make the display name more unique?
```suggestion
      "displayName": "Pizza Inn (America)",
```

Cj-Malone: Definitely, if we were adding AU or another entry. But since there is only 1 Pizza Inn, it's less important. But I'd still do it, for JOSM users.

data/brands/amenity/fast_food.json
Cj-Malone: IMO it just adds more of a maintenance bourdon, unless there is a specific reason to shrink it below a country level.

data/brands/amenity/fast_food.json
peternewman: Ah do they get the presets with no geofencing then?

TBH I guess I could/should have added the AU one too, although I didn't have any Wikidata for them.

data/brands/amenity/fast_food.json
peternewman: I guess what I don't know and didn't bother checking, is how many other places called Pizza Inn exist in the rest of the US that are nothing to do with this specific brand...

PR #11822, opened 2026-02-11
Remove Consejería de Educación de la Junta de Castilla y León

Crashillo:
According to the local community, no need to use different operators. _Junta de Castilla y León_, plain and simple, is enough

https://github.com/osmlab/name-suggestion-index/blob/main/data/operators/amenity/school.json#L23794-L23816

### DIFF
--- a/data/operators/amenity/school.json
+++ b/data/operators/amenity/school.json
@@ -12627,29 +12627,6 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


Robot8A:
The edit removes consejeriadeeducaciondelajuntadecastillayleon-5f247a as a duplicate of juntadecastillayleon-5f247a, which is kept. Lgtm

PR #11836, opened 2026-02-16
Add Art Lunch cafeterias

aq7443:
The largest cafeteria chain in Armenia (26 venues)
https://artlunch.am/en/about-us

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -474,7 +474,7 @@
     {
       "displayName": "Art Lunch",
       "id": "artlunch-c2a495",
-      "locationSet": {"include": ["am"]},
+      "locationSet": {"include": ["am", "ru"]},
       "tags": {
         "amenity": "fast_food",
         "brand": "Art Lunch",


GA-Kevin-Codes:
Thanks for your PR! From the website linked, it seems they also state they have locations in Russia, Ukraine and Kazakhstan. If so, these should be added to the `locationSet`.

data/brands/amenity/fast_food.json
```diff
@@ -471,6 +471,24 @@
         "takeaway": "yes"
       }
     },
+    {
+      "displayName": "Art Lunch",
+      "id": "artlunch-c2a495",
+      "locationSet": {"include": ["am"]},
```
GA-Kevin-Codes: specifically, this line would be updated.

GA-Kevin-Codes (changes requested):

aq7443:
Oh, I didn't know they operate in other countries.

As far as I can see now, currently they have cafeterias:
- 8 in Moscow
- 3 in Almaty
- 1 in Pasadena
And there was 1 in Kiev, but it's closed since 2023...

`      "locationSet": {"include": ["am", "ru"]},`
Will this be enough, or should I also include Kazakhstan and the USA?

aq7443:
Added Russia

PR #11837, opened 2026-02-16
Add Grand Seiko

akadouri:
Add Grand Seiko watch brand, they have stores around the world https://www.grand-seiko.com/us-en/storesinfo

### DIFF
--- a/data/brands/shop/watches.json
+++ b/data/brands/shop/watches.json
@@ -86,6 +86,18 @@
       }
     },
     {
+      "displayName": "Grand Seiko",
+      "matchNames": ["グランドセイコー"],
+      "matchTags": ["shop/jewelry"],
+      "tags": {
+        "name": "Grand Sieko",
+        "name:ja": "グランドセイコー",
+        "brand": "Grand Sieko",
+        "brand:wikidata": "Q11300325",
+        "shop": "watches"
+      }
+    },
+    {
       "displayName": "IWC",
       "id": "iwc-101248",
       "locationSet": {"include": ["001"]},
@@ -233,6 +245,7 @@
         "brand": "Seiko",
         "brand:wikidata": "Q322601",
         "name": "Seiko",
+        "name:ja": "セイコー",
         "shop": "watches"
       }
     },


GA-Kevin-Codes:
Hey @akadouri, I appreciate the PR! A big confused on this one. Grand Seiko is a subset of Seiko watches, which [is in the NSI already](https://nsi.guide/index.html?t=*&k=shop&v=watches&tt=seiko). From the site linked, there are [9 "boutique" stores](https://www.grand-seiko.com/us-en/storesinfo/list?storetype=5cb7ddd02f264d01bab89350bb19ffb2) with the Grand Seiko name, but the rest seem to be other jewelers. Finally, you added a `name:jp` tag to the worldwide Seiko listing. There is certainly debate about what to do next on multilingualism on the NSI but I don't think adding `name:jp` to entities worldwide would be the purview of this project. Currently, the Japanese name is in the `matchName` field, so those searching for the Japanese name can already find the Seiko listing. Thanks again for your contribution and work here! I'll mark this as a draft until we are ready for final review.

akadouri:
Grand Seiko is the luxury brand within Seiko, similar to Lexus as a luxury sub-brand of Toyota. The stores are branded differently and some only sell Grand Seiko watches. While there are 9 stores in the US, there are a few in Europe and ~20 or so in Japan (though some may be within malls, they could still be micromapped).

The brand is Japanese and has an official Japanese name, I would argue that the `name:jp` is important for those 20 or so locations in Japan and there already seem to be many translations on other NSI listings.

GA-Kevin-Codes:
Hey @akadouri, I see now about the locations, thanks for pointing that out! My concern with the `name:jp` is that it is on a `001` entry, not a `jp` locationSet. So if I added one of the 9 locations in the US, the NSI would automatically be filling in `name:jp` which quickly turns into "Is OSM a translator or is this better housed elsewhere"... The translation is already in match names in the NSI and can and should be added to the Wikidata language property on [Q322601](https://www.wikidata.org/wiki/Q322601)!

akadouri:
That makes sense, thanks for the quick response.

GA-Kevin-Codes:
No problem, one more thing, be sure to include locationSet, sounds like "US", "EU", "JP" but thats just from this conversation. if you want an easy way to add these I made [Suggest-A-Brand](https://ga-kevin-codes.github.io/Suggest-A-Brand/) to help!

PR #11843, opened 2026-02-18
update german regional infrastructure network operator (Badenova)

nalxnet:
The german utility company re-branded itself and all subsidiary companies beginning of this year. This commit reflects these changes. Additionally, it replaces the parent company with its subsidiaries and adds them to new features where applicable. Also the location scope has been narrowed down.

### DIFF
--- a/data/operators/man_made/pipeline.json
+++ b/data/operators/man_made/pipeline.json
@@ -484,26 +484,39 @@
       }
     },
     {
-      "displayName": "badenova",
-      "id": "badenova-da7c89",
-      "locationSet": {"include": ["de"]},
-      "tags": {
-        "man_made": "pipeline",
-        "operator": "badenova",
-        "operator:wikidata": "Q798941"
-      }
-    },
-    {
-      "displayName": "badenovaWÄRMEPLUS",
-      "id": "badenovawarmeplus-da7c89",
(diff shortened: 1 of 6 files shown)


data/operators/man_made/pipeline.json
```diff
@@ -484,26 +484,41 @@
...
-      "id": "badenovawarmeplus-da7c89",
-      "locationSet": {"include": ["de"]},
+      "displayName": "Badenova Wärmeplus",
+      "id": "badenovawarmeplus-0b5f1c",
+      "locationSet": {
+        "include": ["de-bw.geojson"]
+      },
       "matchNames": [
-        "badenovawärmeplus gmbh & co. kg"
+        "badenova",
```
nalxnet: This matches the same string as the other subsidiary company above, is this OK?

One subsidiary (Badenova Wärmeplus) is just for hot water pipelines used in heating, the other one (Badenova Netze) for all other substances (gas, water, hydrogen).

kjonosm: this won't work, better delete "badenova" and "badenova ag & co. kg" from matchNames for this item.

nalxnet: I see, thanks. Removed it from "Badenova Wärmeplus" but kept it at "Badenova Netze".

PR #11844, opened 2026-02-18
Update Žabka

Oivo35:
The convenience chain Žabka in the Czech Republic was taken over and is operated by Tesco under a different logo.
Updated locationSet of original item and added new brand.

### DIFF
--- a/data/brands/shop/clothes.json
+++ b/data/brands/shop/clothes.json
@@ -2133,7 +2133,8 @@
       "id": "cartersoshkosh-9e1367",
       "locationSet": {"include": ["ca"]},
       "tags": {
-        "brand": "Carter's OshKosh",
+        "brand": "Carter's",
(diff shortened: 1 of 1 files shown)


data/brands/shop/convenience.json
```diff
@@ -56,6 +56,19 @@
...
   "items": [
+    {
+      "displayName": "Žabka",
+      "locationSet": {"include": ["cz"]},
+      "matchNames": ["žabka potraviny"],
+      "tags": {
+        "brand": "Žabka",
+        "brand:wikidata": "Q133445159",
+        "name": "Žabka",
+        "operator": "Tesco Franchise Stores ČR",
```
matkoniecz: if these are still franchise stores then such `operator` is wrong

Oivo35: Yes, the stores are operated as franchises. I removed operator.

PR #11845, opened 2026-02-18
Georgian POIs update

empty-child:
- Several brands fixed
- Wikidata added
- Several POIs added

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -8435,7 +8435,6 @@
         "name:en": "Magniti",
         "name:ka": "მაგნიტი",
         "name:ru": "Магнити",
-        "operator:wikidata": "Q118792780",
         "shop": "convenience"
       }
     },
@@ -8470,7 +8469,6 @@
         "name:en": "Nikora",
         "name:ka": "ნიკორა",
         "name:ru": "Никора",
-        "operator:wikidata": "Q115949612",
         "shop": "convenience"
       }
     },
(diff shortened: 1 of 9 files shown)


Cj-Malone:
operator:wikidata without operator?

empty-child:
Ouch, sorry. Removed

data/brands/shop/cosmetics.json
```diff
@@ -1120,7 +1120,7 @@
         "brand:en": "Voulez-Vous",
         "brand:ka": "ვულე ვუ",
         "brand:ru": "Вуле Ву",
-        "brand:wikidata": "Q131469965",
+        "brand:wikidata": "Q131469964",
```
bhousel, line 1123: This change definitely seems wrong

data/brands/shop/convenience.json
bhousel, line 8361: Carrefour City is being tracked as a different brand from Carrefour, so I think we should keep as the "Carrefour City" QID, if that's really what this item is supposed to represent.

data/brands/shop/cosmetics.json
```diff
@@ -1153,7 +1169,7 @@
         "brand": "მოზაიკა",
         "brand:en": "Mosaic",
         "brand:ka": "მოზაიკა",
-        "brand:wikidata": "Q131469966",
+        "brand:wikidata": "Q131469965",
```
bhousel, line 1172: another QID mixup

bhousel:
I started reviewing but found numerous issues with QIDs getting mixed up, so I'm just going to close this.  For future try to make smaller more deliberate changes, it will avoid making so many mistakes.

balsoft:
Just for context, this was a semi-automated sync from [our JOSM template](https://github.com/komachi/josm-ge-preset) using [a script I wrote](https://github.com/komachi/josm-ge-preset/blob/main/toNSI.py). I think a better solution is to fix our template accordingly and then continue with semi-automated syncs, rather than manually transferring information piece-wise.

PR #11847, opened 2026-02-19
Add Indomaret Fresh

crse:
Add spin-off brand of Indonesian Indomaret with bigger space and green produce.


crse:
Some reference for validation:
https://id.wikipedia.org/wiki/Indomaret
https://www.instagram.com/reels/DJlhxp9qBYh/

bhousel:
Thanks @crse 👍

PR #11852, opened 2026-02-19
Russian POIs synced with template

empty-child:
Some POIs were transferred from [Russian POI template](https://github.com/ruosm-presets/literan-moscow)


data/brands/amenity/fast_food.json
```diff
@@ -1319,6 +1301,22 @@
         "takeaway": "yes"
       }
     },
+    {
+      "displayName": "Burger King (Russia)",
```
bhousel, line 1305: This duplicates an existing Burger King in Russia entry (Бургер Кинг)

data/brands/amenity/fast_food.json
```diff
@@ -2261,6 +2259,21 @@
         "takeaway": "yes"
       }
     },
+    {
+      "displayName": "Cofix",
```
bhousel, line 2263: This duplicates an existing Cofix entry.

bhousel:
Sorry I'm going to close this.  The few that I spot checked seemed to be a mess.
They are either duplicating existing entries that we have, or lots of them are just in the wrong category (`amenity=fast_food` vs `amenity=cafe` vs `amenity=restaurant`).
It seems like the source data wasn't very carefully compiled.

PR #11861, opened 2026-02-22
TGJones - add other categories

TheUKHighStreet:


bhousel:
Seems ok - why did you mark it "draft"?

TheUKHighStreet:
> Seems ok - why did you mark it "draft"?

There was some discussion about whether to include the gift category or not but I don't think any decision was made. I'll mark it ready. Thanks.

PR #11862, opened 2026-02-22
Add TOMONY convenience store entry

j9d3it:


data/brands/shop/convenience.json
```diff
@@ -6014,6 +5981,17 @@
...
       }
     },
+    {
+      "displayName": "TOMONY",
+      "locationSet": {"include": ["jp"]},
+      "matchNames": ["トモニー"],
+      "tags": {
+        "brand": "TOMONY",
+        "brand:wikidata": "Q11249798",
+        "name": "TOMONY",
```
LaoshuBaby: I want to point out that Tomony is indeed operated by FamilyMart and only have this specially name in Seibu Railway's station area. So I guess we can tag it with FamilyMart as brand and with Tomony as name. 
Is there any suggestion from local community?

 (I live in Japan but have just a little connection to local community, so above is my personal suggestion)

---
Besides I'll add Katakana name as Japanese name

j9d3it:
Ah sorry, I didn't get notified about it. I will implement it.

PR #11924, opened 2026-03-10
Add & update Danish chains

mikini:
Bank:
* Sydbank -> AL Sydbank
* Skjern Bank

Optician:
* Thiele
* Profil Optik
* Synoptik: +dk

Bike:
* Bike & Co
* Fri BikeShop: fix wikidata

Convenience:
* Nærkøb

Clothes:
* Tøjeksperten

Restaurant:
* Pincho Nation: da & no

Garden centre:
* Plantorame

Bed:
* Drømmeland

Second hand:
* Kirppu

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -12344,6 +12344,16 @@
       }
     },
     {
+      "displayName": "Skjern Bank",
+      "locationSet": {"include": ["dk"]},
+      "tags": {
+        "amenity": "bank",
+        "brand": "Skjern Bank",
+        "brand:wikidata": "Q43523656",
+        "name": "Skjern Bank"
+      }
+    },
+    {
       "displayName": "Sicoob",
       "id": "sicoob-0ed44d",
       "locationSet": {"include": ["br"]},
@@ -13024,16 +13034,16 @@
       }
     },
     {
-      "displayName": "Sydbank",
+      "displayName": "AL Sydbank",
       "id": "sydbank-9eeba7",
       "locationSet": {
         "include": ["de", "dk", "se"]
       },
       "tags": {
         "amenity": "bank",
-        "brand": "Sydbank",
+        "brand": "AL Sydbank",
         "brand:wikidata": "Q825159",
-        "name": "Sydbank"
+        "name": "AL Sydbank"
       }
     },
     {
--- a/data/brands/amenity/restaurant.json
+++ b/data/brands/amenity/restaurant.json
@@ -6182,6 +6182,30 @@
       }
     },
     {
+      "displayName": "Pincho Nation",
+      "locationSet": {"include": ["dk"]},
+      "tags": {
+        "amenity": "restaurant",
+        "brand": "Pincho Nation",
+        "brand:wikidata": "Q42298558",
+        "cuisine": "tapas",
+        "name": "Pincho Nation"
+      }
+    },
+    {
+      "displayName": "Pincho Nation",
+      "locationSet": {"include": ["no"]},
+      "issues": [5500],
+      "preserveTags": ["^name"],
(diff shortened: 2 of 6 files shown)


kjonosm:
The suggested change to Al Sydbank is not quite clear to me. Has this been rolled out to all branches etc.?

mikini:
> The suggested change to Al Sydbank is not quite clear to me. Has this been rolled out to all branches etc.?

The merger was formally approved and done on 2025-12-08 with the [former "Sydbank A/S" changing name to "AL Sydbank A/S"](https://ml-eu.globenewswire.com/Resource/Download/a81c2693-83a3-4c75-af33-ca3f95803b92) (I have just updated [wikidata](https://www.wikidata.org/wiki/Q825159) with old/new name, date and new website). 

They are currently doing lay-offs  ([47 in February](https://www.dr.dk/nyheder/seneste/al-sydbank-vinker-farvel-til-47-ansatte-efter-fusion), [84 in March](https://www.dr.dk/nyheder/seneste/ny-fyringsrunde-rammer-ansatte-i-al-sydbank)) I guess because they are  determining which branches to close and which to continue.

There is no announced time frame for the practical merger, current wording is;

> "Selvom vi er fu­sio­ne­ret, tager det tid, før alt er på plads. Det betyder, at vi rent teknisk stadig er tre banker." ([source](https://www.al-sydbank.dk/))

Meaning; "In spite of us merging, it will be a while before everything has settled. This means that technically we are still three banks."

> "Hvornår er arbejdet med at samle til én bank færdigt (AL Sydbank-skilte på filialer, fælles netbank/mobilapp, osv.)? Vi kan endnu ikke sige, hvornår alt er lagt sammen. I en periode vil du stadig møde vores gamle brands Sydbank, Arbejdernes Landsbank og Vestjysk Bank.."([source](https://www.al-sydbank.dk/spoergsmaal-og-svar))

Meaning; "When is it expected that one bank emerges from this (AL Sydbank signage at all branches, common net banking/app etc.)? At this point we do not know when all this has been completed. For some periode you will still see our old branding as Sydbank, Arbejdernes Landsbank and Vestjysk Bank"


I could revert it, as I kind of agree that it won't respect ground truth before all branches have been converted. But then brand:wikidata of "Sydbank" will still point to an entity which is now correctly "AL Sydbank". Inevitable it will end up being like this.

But NSI is still just a suggestion which mappers are free to ignore if it doesn't fit reality.

UKChris-osm:
How come "_Pincho Nation_" is separated? This is currently causing a build error due to two `displayName` tags being the same! They both seem to have the same Wikidata, so could be listed as one entry using both locations.

mikini:
> How come "_Pincho Nation_" is separated? This is currently causing a build error due to two `displayName` tags being the same! They both seem to have the same Wikidata, so could be listed as one entry using both locations.

Sorry for that. Hope the PR checks can be enabled again to prevent this.

I followed the Norwegian's wish of having a preserveTags to keep names with branch identifiers in their area (see issue #5500 referenced in the entry). This is not something Danish entries use unless it is common for the branches of the particular chain to advertise using this on their facades ("on the ground" rule) which is the reason for the plain Danish entry without preserveTags (see [recent talk-dk discussion](https://lists.openstreetmap.org/pipermail/talk-dk/2026-February/006057.html)). That entry could have included Sweden also, unfortunately in Sweden the restaurants are called just "Pinchos" so "se" needed a separate entry also.

I would prefer to not have name preserved in "dk" and "se" but I see a lot of other entries also done like this so I guess it makes little sense to hunt this one.
A way to keep the "no' entries separate from others could be to give them unique displayNames. I guess the pattern of separation I copied must have done that (not able to relocate it ATM from phone, however).

PR #11925, opened 2026-03-10
Split AlzaBox

RickeyRichards:
Split AlzaBox parcel locker brand by country-specific operator.

AlzaBox lockers are operated by separate Alza group companies in each country:
CZ (Alza.cz) - operator will be feeded directly from Wikidata item (Q115254158)
SK (Alza.sk) - "operator" and "operator:wikidata" specified
HU (Alza.hu) - "operator" and "operator:wikidata" specified
AT (Alza.at) - "operator" and "operator:wikidata" specified

This change splits initial NSI entry, "operator" and "operator:wikidata" will now be based on location.

### DIFF
--- a/data/brands/amenity/parcel_locker.json
+++ b/data/brands/amenity/parcel_locker.json
@@ -27,16 +27,49 @@
       }
     },
     {
-      "displayName": "AlzaBox",
-      "id": "alzabox-bee63b",
-      "locationSet": {
-        "include": ["at", "cz", "hu", "sk"]
-      },
+      "displayName": "AlzaBox (Česko)",
+      "locationSet": {"include": ["cz"]},
       "tags": {
         "amenity": "parcel_locker",
         "brand": "AlzaBox",
         "brand:wikidata": "Q115254158",
         "name": "AlzaBox"
+      }
+    },
+    {
+      "displayName": "AlzaBox (Magyarország)",
+      "locationSet": {"include": ["hu"]},
+      "tags": {
+        "amenity": "parcel_locker",
+        "brand": "AlzaBox",
+        "brand:wikidata": "Q115254158",
+        "name": "AlzaBox",
+        "operator": "Alza.hu",
+        "operator:wikidata": "Q138637883"
+      }
+    },
+    {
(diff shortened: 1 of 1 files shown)


kjonosm:
Maybe just leave out `operator` altogether? parcel_locker.json is part f the /brand tree.

RickeyRichards:
> Maybe just leave out `operator` altogether? parcel_locker.json is part f the /brand tree.

I'm okay with leaving out the operator, but right now it's being pulled directly from the AlzaBox Wikidata item since an operator is defined there. I don't see a way around it other than deleting the operator statement in the Wikidata item for AlzaBox (Q115254158) or proposed change in this Pull request. Do you happen to know if there’s a better way to handle this?

Cj-Malone:
> but right now it's being pulled directly from the AlzaBox Wikidata item since an operator is defined there

What is pulling the operator from Wikidata?

RickeyRichards:
> > but right now it's being pulled directly from the AlzaBox Wikidata item since an operator is defined there
> 
> What is pulling the operator from Wikidata?

>> As I understand it (correct me if I'm wrong please), parameters `operator` and `operator:wikidata` are not defined in [parcel_locker.json](https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/amenity/parcel_locker.json) for AlzaBox, but are still present when creating new POI in iD editor. So my initial understanding was that operator is somehow inserted straight from Wikidata item [Q115254158](https://www.wikidata.org/wiki/Q115254158). Would you be able to explain how is it in reality, that `operator` tag is automatically inserted when adding AlzaBox to OSM?

Oivo35:
"I think it’s already been fixed in 9e0c6b004a2d26ccc97e35b5164e99ed9b636be0. It should be available in the next iD release."

RickeyRichards:
> "I think it’s already been fixed in [9e0c6b0](https://github.com/osmlab/name-suggestion-index/commit/9e0c6b004a2d26ccc97e35b5164e99ed9b636be0). It should be available in the next iD release."

@Oivo35 See it, so operator was removed. In this case, pull request can be closed. Thank you all.

PR #11931, opened 2026-03-11
Update funeral_directors.json

TheUKHighStreet:
Is removing the name field correct to enable other names? Lots of the Coop Funeral Directors have a different name.

### DIFF
--- a/data/brands/shop/funeral_directors.json
+++ b/data/brands/shop/funeral_directors.json
@@ -122,9 +122,11 @@
         "the co-operative group"
       ],
       "matchTags": ["office/company"],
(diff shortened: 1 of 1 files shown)


bhousel:
> Is removing the name field correct to enable other names? Lots of the Coop Funeral Directors have a different name.

Normally the correct solution is to leave the `name` tag there so that it defaults to something reasonable, but add a `"preserveTags": ["^name"]` property so that mappers can change the name to something else without getting nagged by the validator.

TheUKHighStreet:
Thanks - I thought there was something slightly more complicated but couldn't remember. I'll update.

PR #11932, opened 2026-03-12
changed spelling of myflexbox

peter-myflexbox:
changed spelling to match https://www.wikidata.org/wiki/Q117313525

### DIFF
--- a/data/brands/amenity/parcel_locker.json
+++ b/data/brands/amenity/parcel_locker.json
@@ -744,12 +744,12 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Hey, we don't necessary always follow Wikidata, but based on your username I assume you're related to the brand? It's always nice to have brands directly involved.

[All the Places](https://alltheplaces.xyz/spiders.html?search=myflexbox&link_format=map) will auto update (after NSI is released) OSM may take a bit longer.

PR #11938, opened 2026-03-14
Update bank.json

NFarrasP:
Added matchname for BSI to make it easier to search


Cj-Malone:
I don't think this will do anything, there ds already a short name

NFarrasP:
Okay and sorry. I initially made this request to fix search issue for a bank in Indonesia. However, I just checked once again on StreetComplete and iD to find that BSI is not the one I was looking for.

I will close this and open another one once I get the correct one.

PR #11942, opened 2026-03-15
Add brand tags to Russian Post

aq7443:


aq7443:
Added one more common name

PR #11952, opened 2026-03-19
Rename Total by Verizon to Total Wireless

JoshM2:
I saw a store today that said Total Wireless on the sign.

This article on the Verizon website says "Total Wireless, formerly Total by Verizon"
https://www.verizon.com/about/news/total-wireless-launches-bold-new-offers-new-look-outshine-rivals-prepaid-wireless

### DIFF
--- a/data/brands/shop/mobile_phone.json
+++ b/data/brands/shop/mobile_phone.json
@@ -1091,14 +1091,14 @@
       }
     },
     {
-      "displayName": "Total by Verizon",
+      "displayName": "Total Wireless",
       "id": "totalbyverizon-ce3e5c",
       "locationSet": {"include": ["us"]},
-      "matchNames": ["total", "total wireless"],
+      "matchNames": ["total", "total by verizon"],
       "tags": {
-        "brand": "Total by Verizon",
+        "brand": "Total Wireless",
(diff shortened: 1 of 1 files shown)


ilias52730:
Surely this is `shop=telecommunication` and not `shop=mobile_phone`

JoshM2:
Yeah, shop=telecommunication is probably better

JoshM2:
Though actually it appears that it is very common for similar stores to be in the shop=mobile_phone category. I also see T-Mobile, Verizon, and AT&T, which are the three biggest cellular data providers in the United States.

ilias52730:
Right. There's been [endless discussions](https://community.openstreetmap.org/t/rfc-feature-proposal-shop-telecommunication/135231) about this. Always the same people who think a mobile phone shop is the same as a telecom shop. Thankfully there are some communities with rational people (like Spain and France) that accepted the change when I proposed it, without hesitation. It's a slow process, but it's going in the right direction. In the meantime I can't single handedly make the change for all instances of T-Mobile, Verizon, and AT&T in the US, or I'll get in trouble. So I'll get to creating a post on the US community forum to propose the change there.

PR #11954, opened 2026-03-20
Update blood_donation.json

Mannivu:
Removed wrong operator: AVIS doesn't operate any blood donation structure, they are all within an hospital an operated by the hospital's staff

### DIFF
--- a/data/brands/healthcare/blood_donation.json
+++ b/data/brands/healthcare/blood_donation.json
@@ -33,19 +33,6 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


Cj-Malone:
It'll get auto re added because they are still in OSM eg https://www.openstreetmap.org/node/5814382734

Mannivu:
After further investigations, I've seen that there maybe some blood donation points directly operated by AVIS, so it might be useful to have this preset.

PR #11971, opened 2026-03-23
Removed forced name of AVIS from blood_donation.json

Mannivu:
Removing name since it could have a different name, e.g. it could be dedicated to someone

### DIFF
--- a/data/brands/healthcare/blood_donation.json
+++ b/data/brands/healthcare/blood_donation.json
@@ -41,8 +41,8 @@
         "brand": "AVIS",
         "brand:wikidata": "Q3627251",
         "donation:compensation": "no",
-        "preserveTags": ["^name"],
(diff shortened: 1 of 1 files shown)


bhousel:
You probably still want it to have _some_ `name` and "AVIS" is a reasonable default.

Normally the correct solution is to leave the name tag there so that it defaults to something reasonable, but add a `"preserveTags": ["^name"]` property so that mappers can change the name to something else without getting nagged by the validator.

Mannivu:
Thanks, I added it (I didn't know about that tag).

bhousel:
Sorry no, It's not a tag, it's a property to control the behavior of items that are in the name-suggestion-index.
see: d30a92ade - I added the property there.

Mannivu:
Yes, sorry. Property, I misspoke.

PR #11985, opened 2026-03-27
broaden area to cover actual extend

### LINKED ISSUE #11928: Q5155093 operates for example in Galway, not only in Dublin
this is pulling one case from https://github.com/osmlab/name-suggestion-index/issues/10469 - and it looks like at least in this case NSI has bug (which does not invalidate `downstream-bug` label)

`network=Commuter network:wikidata=Q5155093` at https://nsi.guide/index.html?t=*&k=route&v=train&tt=q73043 is marked as `{"include":["ie-d.geojson"]}` which restricts it to https://github.com/osmlab/name-suggestion-index/blob/main/features/ie/ie-d.geojson

but per https://en.wikipedia.org/wiki/Commuter_(Iarnr%C3%B3d_%C3%89ireann)

> Commuter ([Irish](https://en.wikipedia.org/wiki/Irish_language): Comaitéir) is a brand of suburban rail services operated by [Iarnród Éireann](https://en.wikipedia.org/wiki/Iarnr%C3%B3d_%C3%89ireann) in the [Republic of Ireland](https://en.wikipedia.org/wiki/Republic_of_Ireland), serving the cities of [Dublin](https://en.wikipedia.org/wiki/Dublin), [Cork](https://en.wikipedia.org/wiki/Cork_(city)), [Limerick](https://en.wikipedia.org/wiki/Limerick) and [Galway](https://en.wikipedia.org/wiki/Galway). This brand is distinct from the longer distance [InterCity](https://en.wikipedia.org/wiki/InterCity_(Iarnr%C3%B3d_%C3%89ireann)) brand, and Dublin's higher frequency [DART](https://en.wikipedia.org/wiki/Dublin_Area_Rapid_Transit) brand.

it seems that say https://www.openstreetmap.org/relation/10377553 is potentially valid example of this NSI entry, despite being outside declared catchment area

If I understand situation correctly its area restriction should be changed

matkoniecz:
fixes #11928

some routes were outside claimed coverage

note: maybe another custom area would be better than extending claimed coverage to entire country

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -425,7 +425,7 @@
       "displayName": "Commuter",
       "id": "commuter-91413c",
       "locationSet": {
-        "include": ["ie-d.geojson"]
+        "include": ["ie.geojson"]
       },
       "matchNames": ["dublin commuter"],
       "tags": {


UKChris-osm:
Thanks, but just as an FYI, as Ireland is part of [Country Coder](https://github.com/rapideditor/country-coder), it doesn't require the `.geojson` at the end, as that is only required for any custom `.geojson` files we add.

matkoniecz:
So it was likely outright broken? Sorry.

PR #12006, opened 2026-03-29
Tax, customs and border control offices

DevManiak17:
This change introduces three new preset categories to the Name Suggestion Index: tax offices, customs offices, and border control facilities. These are widely used types of government infrastructure that are currently underrepresented in NSI, despite being mapped frequently in OpenStreetMap across many countries.

Adding these categories improves consistency in tagging and naming, helps mappers select standardized names more easily, and reduces duplication and spelling variations. It is particularly valuable in regions where such institutions follow recognizable naming patterns but are currently entered manually.

Including these presets will enhance data quality, support better search and rendering, and make it easier for contributors to map administrative and border-related infrastructure accurately.


rogoyski:
You are proposing to add tagging under the `amenity=*` key that is either documented as deprecated or not documented at all.

I guess `barrier=border_control` might not be supported by NSI, but what about `office=government` with `government=customs` and `government=tax`?

DevManiak17:
I’m not suggesting specific tags; the ones used in the files are only examples of what I would like to include in NSI. Currently, in Poland, there is a lot of inconsistency in tagging this type of object (for example, many Polish Border Guard offices are marked as military areas, which, in my opinion, is not correct). That’s why I decided to propose this change.
Your suggestion sounds good.

Cj-Malone:
Since the specific tags aren't final I've turned this into a draft. I have strong objections to the current tags (3 uses of amenity=tax), NSI should be contain we'll documented standard tags, not used to push new ones.

DevManiak17:
I've deleted created files and edited government.json.

DevManiak17:
Is there any problem with merging this change?

data/operators/office/government.json
```diff
@@ -1165,6 +1165,32 @@
...
+      "displayName": "Krajowa Administracja Skarbowa",
+      "id": "krajowaadministracjaskarbowa-9c7901",
+      "locationSet": {"include": ["pl"]},
+      "matchNames": ["oddział celny"],
+      "tags": {
+        "office": "government",
+        "government": "customs",
+        "operator": "Krajowa Administracja Skarbowa",
+        "operator:type": "government",
+        "operator:wikidata": "Q30904637"
```
rogoyski, line 1178: This duplicates too much of the entry below, doesn't it? In particular, it shouldn't be that both have the same `displayName`. I think also there can also be more keywords here that @DevManiak17 even used in his post announcing this proposal in community channels that I am not able to locate now. What about _urząd celny_ and _izba celna_ in particular? I would need a refresher on the structure of that administration after the latest reforms 😅

rogoyski (changes requested):

data/operators/office/government.json
rogoyski: After a brief glimpse into the structure, I think the display name for `government=tax` could be _Urząd Skarbowy_. For this entry with `government=customs` it gets more complicated. The tagging should be applied to _Urząd Celno-Skarbowy_ and to _Oddział Celny_, so it's not clear which to use as a display name. Maybe these two things should be two different NSI entries 🤔 For sure, _urząd celno-skarbowy_ and _oddział celny_ should go to `matchNames`, the obsolete terms _urząd celny_ and _izba celna_ too.

DevManiak17: _Izba Celna_ no longer exists - it functioned until 2017. After the Służba Celna was merged with the tax administration, the following entities were established: 'Urząd Celno-Skarbowy', which replaced the former _Urząd Celny_, as well as _Oddział Celny_ and _Miejsce wyznaczone_ - locations where customs control takes place, such as railway stations in Hrebenne and Werchrata.

DevManiak17: There is also an _Izba Administracji Skarbowej_ in each voivodeship, which makes the system more complex.

DevManiak17: Krajowa Administracja Skarbowa:
- Izba Administracji Skarbowej (mixed Izba Skarbowa and Izba Celna)
- Urząd Celno-Skarbowy (formerly Urząd Celny)
- Urząd Skarbowy
- Oddział Celny
- Miejsce Wyznaczone
- [Krajowa Szkoła Skarbowości](https://pl.wikipedia.org/wiki/Krajowa_Szko%C5%82a_Skarbowo%C5%9Bci)

rogoyski: > Izba Celna no longer exists

Yes, I learned that before writing the second comment. This is why I called it _obsolete_ there.

> There is also an Izba Administracji Skarbowej in each voivodeship

Correct, but it is not clear which tagging should it use if it's responsibility is both `tax` and `customs`. Should probably be another NSI entry anyway, though since there are only 16 of them, I also thought we could consider them out of scope.

> Miejsce Wyznaczone

I would not call this a `government=office` anymore. We should first determine tagging for this if we want to get that deep.

---

I would suggest to first cover in this pull request the types that are not ambiguous and approach the rest later if deemed necessary:
* Urząd Skarbowy
* Urząd Celno-Skarbowy
* Oddział Celny
* Straż Graniczna

rogoyski: I have made a commit in my own fork sorting this out to a state that I would be happy with. Not sure how I can deliver it. @DevManiak17 , can you pull or cherry-pick it into your branch?

https://github.com/rogoyski/name-suggestion-index/commit/b82066bde29d9b35fa3936619fdb77943da4ad44

PR #12007, opened 2026-03-30
add UniCredit in Poland and unclutter brand name

rogoyski:
This pull request covers two things:

* adds the Unicredit brand to the location `pl` (Poland) where the bank started offering services under its own brand late last year [[1](https://unicredit.pl/media/UniCredit-w-Polsce)]
* unclutters the brand name by removing the word _bank_ which is not part of the name

To justify removal of the word _bank_ let me point you at [Wikipedia](https://en.wikipedia.org/wiki/UniCredit), [Wikidata](https://www.wikidata.org/wiki/Q45568) (which was linked here all along) and finally the [international website](https://www.unicreditgroup.eu/en.html) of the bank. Neither in the logotype nor in narratives is the word _bank_ used as part of the brand name or even the legal entity operating it.

image

image

image

image

We all know UniCredit is a bank, that's why we tag is as `amenity=bank`, but there is absolutely no reason to repeat this category in the `name` or `brand` tags.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ pnpm-lock.yaml
 yarn.lock
 
 .vscode/
+.idea/
 coverage/
 dist/
 node_modules/
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -13698,14 +13698,15 @@
       }
     },
     {
-      "displayName": "UniCredit Bank",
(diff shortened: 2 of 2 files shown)


bhousel:
This is ok, but please make sure to also add the old name "unicredit bank" to the matchNames array, so that existing data will be matched and users will be prompted to change the tags as you are proposing.

PR #12009, opened 2026-03-31
Rename Pathé Gaumont to Pathé Cinémas

theoricher:
Updated displayName, tags.name, and tags.brand to reflect the new name "Pathé Cinémas". The id remains unchanged (pathegaumont-22090e) to preserve entity continuity and all existing references. No changes to locationSet or brand:wikidata.

### DIFF
--- a/data/brands/amenity/cinema.json
+++ b/data/brands/amenity/cinema.json
@@ -902,22 +902,11 @@
     },
     {
       "displayName": "Pathé",
-      "id": "pathe-484389",
-      "locationSet": {"include": ["nl"]},
-      "tags": {
-        "amenity": "cinema",
-        "brand": "Pathé",
-        "brand:wikidata": "Q61959392",
-        "name": "Pathé"
-      }
-    },
-    {
-      "displayName": "Pathé",
       "id": "pathegaumont-22090e",
       "locationSet": {
-        "include": ["be", "ch", "ci", "fr", "ma", "sn", "tn"]
+        "include": ["be", "ch", "ci", "fr", "ma", "nl", "sn", "tn"]
       },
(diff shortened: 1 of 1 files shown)


UKChris-osm:
As there is a name change I'd suggest including `"matchNames": ["pathé gaumont"];` (underneath `locationSet`) so that the old name is highlighted to mappers as requiring an update.

ilias52730:
- I would even suggest changing the display name, name and brand to just _Pathé_. That's how the cinemas are branded on the ground. _Pathé Cinémas_ is just the legal corporate entity.
- The Dutch already understood this as there is a separate entry, just for Netherlands, named Pathé.
- There's also a duplicate entry with no id or locationset that needs to be removed.
- The locationset also needs to be updated as they now also operate in Morocco, Senegal and Ivory Coast.

image

theoricher:
Hello,

Thanks for your feedback. This is my first time contributing to NSI, so I'm still learning the best practices :)

@UKChris-osm I've added your feedback :)

@ilias52730 I think I made the changes correctly, but I couldn't find the duplicate entry. Could you tell me where it is?

We agree that I shouldn't merge with Pathé Netherlands since they have a different ID and a different Wikidata entry, right?

ilias52730:
> Thanks for your feedback. This is my first time contributing to NSI, so I'm still learning the best practices :)

Of course, we've all started somewhere, you're doing great!

> but I couldn't find the duplicate entry

My mistake, I must have hallucinated that.

> We agree that I shouldn't merge with Pathé Netherlands

Actually, it should be merged, because in Netherlands, they're also branded as just "Pathé". If it's the same branding on the facade, then it's the same entry. The PR seems flawless to me now and will be merged in a few hours.

image

PR #12022, opened 2026-04-02
Removing wrong network tags

giopera:
After a discussion on the relevant telegram groups of the Italian community the following networks resulted to be inserted wrongfully and are outside the scope of the network tag.
A discussion on the forum was also opened and advertised but no discussion happened there.
https://community.openstreetmap.org/t/discussione-riguardante-luso-di-network-rfi-su-stazioni-e-relazioni/142687

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -2081,16 +2081,6 @@
       }
     },
     {
-      "displayName": "RFI",
-      "id": "rfi-66b763",
-      "locationSet": {"include": ["it"]},
-      "tags": {
-        "network": "RFI",
-        "network:wikidata": "Q1060049",
-        "route": "train"
-      }
-    },
-    {
       "displayName": "RhB",
       "id": "rhb-0c66c7",
       "locationSet": {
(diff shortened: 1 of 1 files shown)


Cj-Malone:
RFI was readded https://github.com/osmlab/name-suggestion-index/pull/12033/changes#diff-e5c5c8a385f3e502dfd85f6d088c63ced9a222bf9c38e6ddbe59cb039bcb9c24R2097

giopera:
Is there any explanation on why? t shouldn't be there....

Cj-Malone:
Probably because it is (or recently was) in OSM.

giopera:
Is there a way to blacklist from reinsertion?

bhousel:
Our collection script just collects all the values that are used in OSM.  
The last time we pulled data from the planet file, `network=RFI` [was used 49 times.](https://github.com/ideditor/nsi-collector/blob/99f9deebbad15aa2da11409481f64a1c101e7b8b/dist/osm/networks_all.json#L5094)

We can blacklist it from reinsertion or match it to a different name (e.g. if "RFI" now has a new name or was merged with another network).  It's not clear from the linked forum discussion what you want - I don't think I've ever heard of "out of scope of the network tag" before. 🤷

giopera:
So, generally RFI is the operator of the railway infrastructure, it's not a network of trains and it only operates infrastructure, the operator on the route is the company Trenitalia, while a network may be something like STIBM, a system that unifies prices across various routes, so when a network=RFI is used it means that RFI decides the prices for the service, which it doesn't. This is the wiki page for it https://it.wikipedia.org/wiki/Sistema_tariffario_integrato.

bhousel:
Thanks for the context, if they are actually operators and not networks this makes sense.
I added these these names to the exclude regexes in 12aaa3704

PR #12026, opened 2026-04-03
Add Frankfurter Sparkasse to banks

thedrstrangelove:
Add Frankfurter Sparkasse in the State of Hesse, Germany.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -545,6 +545,19 @@
       }
     },
     {
+      "displayName": "AL Sydbank",
+      "id": "alsydbank-9eeba7",
+      "locationSet": {
+        "include": ["de", "dk", "se"]
+      },
+      "tags": {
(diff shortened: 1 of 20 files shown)


data/brands/amenity/bank.json
```diff
@@ -7484,6 +7484,24 @@
...
       }
     },
+    {
+      "displayName": "Frankfurter Sparkasse",
+      "id": "frankfurtersparkasse-9ffcae",
+      "locationSet": {
+        "include": ["de-he.geojson"]
+      },
+      "matchNames": [
+        "sparkasse",
```
kjonosm: please remove matchName "sparkasse" as it will trigger ID to suggest to switch ALL Sparkassen banks in Hessen to Frankfurter Sparkasse.

thedrstrangelove: Thank you for the feedback. I have removed matchName "sparkasse".

PR #12031, opened 2026-04-04
Update convenience.json

TheUKHighStreet:
Added Southern Coop as the operator for Welcome stores.


UKChris-osm:
My local "_Welcome_" shops are named uniquely, such as "_Welcome placename_", so I'd suggest adding `"preserveTags": ["^name"];` to this to allow names to stay unique where requied.

PR #12054, opened 2026-04-12
Update pastry.json

JhowieNitnek:
Corrected some things about Maison Dandoy + added contact info


UKChris-osm:
Thanks @JhowieNitnek - however, we tend not to include any additional information such as `contact` details, as all these can be found in the Wikidata item we link to with the `brand:wikidata` key.

bhousel:
FYI, I skipped the contact details and added only the identifying tags in 38e6bcb78

PR #12059, opened 2026-04-14
Remove false operator

Crashillo:


UKChris-osm:
Thanks @Crashillo 👍 

I've added "_sícoris club_" to the exclusion list at the top of the file to prevent it from being added back automatically, as it appears more than 10 times within OSM data.

Crashillo:
> more than 10 times

10 times? Wasn't the threshold 50 times?

bhousel:
> 10 times? Wasn't the threshold 50 times?

The threshold for collection is 50x for brands and 10x for operators.

(This doesn't mean that those numbers are official numbers for notability - it just means that those are the frequencies of occurrance that the script considers "worth collecting").

PR #12067, opened 2026-04-15
add typescript definitions for lib/*

k-yle:
It would be helpful for downstream consumers to have complete TypeScript definitions for the public API.

This PR adds all the missing annotations to `lib/*`. I could send a separate PR for the internal `scripts/*` files

### DIFF
--- a/lib/matcher.ts
+++ b/lib/matcher.ts
@@ -2,8 +2,9 @@ import whichPolygon from 'which-polygon';
 import { simplify } from './simplify.ts';
 
 import type LocationConflation from '@rapideditor/location-conflation';
-import type { Vec2 } from '@rapideditor/location-conflation';
-import type { MatchHit, MatchIndexBranch, NsiData } from './types.ts';
+import type { FeatureProperties, Vec2 } from '@rapideditor/location-conflation';
+import type { MatchHit, MatchIndexBranch, NsiData, NsiTree } from './types.ts';
+import type { Feature, Geometry } from 'geojson';
 
 // Imported JSON (will be inlined when generating a bundle)
 import matchGroupsJSON from '../config/matchGroups.json' with {type: 'json'};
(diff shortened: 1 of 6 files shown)


lib/matcher.ts
```diff
@@ -12,39 +15,12 @@ const matchGroups = matchGroupsJSON.matchGroups;
...
-  private matchIndex;
-  private genericWords = new Map();
-  private itemLocation;
-  private locationSets;
-  private locationIndex;
+  private matchIndex: Map<string, Branch>;
+  private genericWords = new Map<string, RegExp>();
+  public itemLocation: Map<string, string> | undefined;
+  public locationSets: Map<string, Feature<Geometry, MatchLocation>> | undefined;
+  public locationIndex: whichPolygon.Query<MatchLocation> | undefined;
```
k-yle: these were changed to `public` since they're used by downstream apps

lib/matcher.ts
```diff
@@ -462,10 +438,10 @@ export class Matcher {
 
     // If we were supplied a location, and a that.locationIndex has been set up,
     // get the locationSets that are valid there so we can filter results.
-    let matchLocations;
+    let matchLocations: MatchLocation[];
     if (Array.isArray(loc) && that.locationIndex) {
       // which-polygon query returns an array of GeoJSON properties, pass true to return all results
-      matchLocations = that.locationIndex([loc[0], loc[1], loc[0], loc[1]], true);
+      matchLocations = that.locationIndex([loc[0], loc[1]], true)!;
```
k-yle: as far as I can tell, [which-polygon expects `Vec2` here](https://github.com/mapbox/which-polygon/blob/ad814168b4876eeade591dcd69264c3914577541/index.js#L30-L36), the 3rd and 4th values are not used.

bhousel:
I'm already working on this today so it looks like unfortunately we crossed paths - does https://github.com/osmlab/name-suggestion-index/commit/67f1f75675d8f7af8568aefef4f5e9d4fc8845d1  head in the right direction?

lib/matcher.ts
bhousel: oh yea, I recently ended up needing to add ambient declarations for which-polygon and other dependencies over in Rapid.  I'll probably just copy whatever I did there for this project too.
https://github.com/bhousel/Rapid/tree/main/modules/types

k-yle:
lol what a coincidence 😄 

but it's great because it makes this PR much simpler! I've rebased and kept only the changes that affect downstream users.

mostly replacing implicit `any`s and `unknown`s with the defintions that you just added in 67f1f75

bhousel:
Thanks! I'm going to close this but I think I got everything from here into 
* 8ddafa50f 2026-04-16 Split tsconfig, improve types for PR #12067 alignment
* b7b5ac30f 2026-04-16 Tighten TypeScript strictness and remove locale-compare

If I missed anything, let me know - I feel like none of my projects are really set up ideally for downstream TypeScript use, but I'm learning and I'm trying to get better at that.

PR #12069, opened 2026-04-17
Replace Wikidata places with governments

1ec5:
I [queried Wikidata](https://qlever.dev/wikidata/F5k53v) for any item about a place in the United States that has an NSI identifier and also has a corresponding item about the government of that place. The latter item is more appropriate for `operator=*` tags. I’ve updated the relevant QIDs in this repository.

This process will need to be repeated in the future as more government items get split out from place items.

### DIFF
--- a/data/operators/leisure/park.json
+++ b/data/operators/leisure/park.json
@@ -959,7 +959,7 @@
         "leisure": "park",
         "operator": "City of Anaheim",
         "operator:type": "government",
-        "operator:wikidata": "Q49247"
+        "operator:wikidata": "Q139044657"
       }
     },
     {
@@ -1709,7 +1709,7 @@
         "leisure": "park",
         "operator": "City of Detroit",
         "operator:type": "government",
-        "operator:wikidata": "Q12439"
+        "operator:wikidata": "Q2824588"
       }
     },
     {
@@ -2281,7 +2281,7 @@
         "leisure": "park",
         "operator": "City of Irvine",
         "operator:type": "government",
-        "operator:wikidata": "Q49219"
+        "operator:wikidata": "Q139033177"
       }
     },
     {
@@ -2750,7 +2750,7 @@
         "leisure": "park",
         "operator": "City of Miami",
         "operator:type": "government",
-        "operator:wikidata": "Q8652"
+        "operator:wikidata": "Q5589249"
       }
     },
     {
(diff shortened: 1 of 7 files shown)


data/operators/leisure/park.json
```diff
@@ -4146,6 +4146,19 @@
...
     },
+    {
+      "displayName": "County of San Bernardino",
+      "id": "countyofsanbernardino-dc1787",
+      "locationSet": {
+        "include": ["us-ca.geojson"]
+      },
+      "tags": {
+        "leisure": "park",
+        "operator": "County of San Bernardino",
```
1ec5, line 4157: In California, county governments are consistently known by the form “County of …” to distinguish them from the counties as places. I flipped it around here, but the build script keeps revivifying a San Bernardino County entry because scores of elements were mistakenly tagged `operator=San Bernardino County`. [Changeset 181,445,178](https://www.openstreetmap.org/changeset/181445178) retags these elements, so this will only be a temporary annoyance. Similarly, [changeset 181,445,979](https://www.openstreetmap.org/changeset/181445979) corrects tags for the County of Santa Clara.

GA-Kevin-Codes:
I've been quietly keeping up with your Wikidata California work as more of an interest than anything. One question on it here though is if it's only the label that's wrong ("San Bernardino County" vs "County of San Bernardino"), shouldnt the label / alias just be flipped on Wikidata? Or is this saying one way is the geographic feature, the other is county government?

1ec5:
The labels on Wikidata are correct: “[County of San Bernardino](https://www.wikidata.org/wiki/Q138891274)” is the common name of the government of San Bernardino County, just as “[State of California](https://www.wikidata.org/wiki/Q5589189)”, “[City of Los Angeles](https://www.wikidata.org/wiki/Q5589234)”, and “[City of Redwood City](https://www.wikidata.org/wiki/Q121024392)” are the governments of California, Los Angeles, and Redwood City, respectively. When a layperson chooses not to use this name, they end up qualifying it, similar to how the English Wikipedia prefers “[government of San Francisco](https://en.wikipedia.org/wiki/Government_of_San_Francisco)” purely for global consistency. But those descriptive names aren’t as suitable for OSM tags.

Other states I’m familiar with typically make this distinction only for municipal corporations but not necessarily at the county or township level, or it’s the stuff of corporate seals and court cases but rarely used by the public. But in California, almost every county insists on the “County of” style. The County of San Bernardino is one of only a few that uses both forms. I think it’s partly because the county government shares jurisdiction with so many other independent authorities with similar names and coextensive boundaries. I haven’t changed the labels of individual agencies and departments to say “County of … Department of …”, because in practice people do use the shorter form.

PR #12072, opened 2026-04-17
New road route operator category

Agazzr:
Created new "Top-Level Tag" Category for road route operators. Added some road operators from Europe and Morocco.


bhousel:
This maybe relates to #4768 
@1ec5 , can you weigh in on whether this is ok?  I think road routes probably have some meaningful information to tag about them, but I'm not sure whether "operator" is it.

1ec5:
I think #4768 was more about `network=*` than `operator=*`. In general, `network=*` matters more to a highway route’s identity than `operator=*`. Until we have a distinct tree that’s all about `network=*`, Go Map!!/iD/Rapid would show `operator=*` as the name of the preset. I think I’ve mentioned before that naming the preset by an operator can get awkward if the organization isn’t as closely identified with the feature. (Everyone knows what “a McDonald’s” is but not “a City of Springfield”.) Unless we or downstream projects address that design issue somehow, we probably should reserve this file for organizations that have a more or less one-to-one relationship to a highway system and avoid adding more tenuous connections to this file, even if the tagging is correct.

At a glance, most of the entries in this PR are toll road authorities. I think this is fine for NSI, as long as the organization operates the entire length of the tollway and doesn’t operate a variety of other kinds of highways. After all, we already added toll authorities as operators of toll gantries in #6703 and elsewhere.

As an example of what not to add yet, some highway routes in the U.S. have `operator=*` set to a state transportation department, such as “Oklahoma Department of Transportation”. But a preset for ODOT would be confusing, because they operate a wide variety of highways from Interstates to state routes, while some Interstates and state routes in the same state are operated by a different authority. I see some [trunk road agents](https://en.wikipedia.org/wiki/Trunk_road_agent) in this PR, which seem to serve a similar purpose as the DOTs, but I don’t know if presets for them would be helpful or unhelpful to UK mappers.

Concession companies could also be problematic. For example, the Chicago Skyway is operated by the Skyway Concession Company, a relatively obscure company, probably not that useful as a preset name. Its parent company, [Atlas Arteria](https://www.atlasarteria.com/), is a huge multinational that’s even more obscure to the average mapper. At a glance, I don’t see any concessionaires in this PR, but we would need to be vigilant about it once this file is created.

data/operators/route/road.json
```diff
@@ -0,0 +1,173 @@
...
+      "locationSet": {"include": ["gb-wls"]},
+      "tags": {
+        "operator": "North and Mid Wales Trunk Road Agent",
+        "operator:short": "NMWTRA",
+        "operator:wikidata": "Q55625449",
+        "route": "road"
+      }
+    },
+    {
+      "displayName": "Red de Carreteras del Estado",
```
Robot8A, line 108: "Red de Carreteras del Estado" would be more akin to a network (though it's true that, for roads, the community is adding identifiers rather than network names to the `network=*` tag). `operator=*` would be wrong in this case for Spain.

data/transit/route/bus.json
```diff
@@ -21135,12 +21135,15 @@
       "locationSet": {
         "include": ["es-na.geojson"]
       },
-      "matchNames": ["tuc pamplona"],
+      "matchNames": ["tuc pamplona","ehg iruña"],
```
Robot8A, line 21138: This change is correct, but probably should be split from this PR

PR #12103, opened 2026-04-19
Bus route networks ref colour

Agazzr:
Added default network colours for some Spanish bus route systems. In these networks, all routes consistently use the same reference colour (ref_colour) as per official branding guidelines. This helps maintain consistency across all route relations in Spain.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -2302,6 +2302,8 @@
       "tags": {
         "network": "Bizkaibus",
         "network:wikidata": "Q879270",
+        "ref:colour": "#9ACD32",
+        "ref:colour_tx": "white",
         "route": "bus"
       }
     },
@@ -2895,6 +2897,8 @@
         "network": "Autobuses de España",
         "network:short": "bus.es",
         "network:wikidata": "Q108443913",
+        "ref:colour": "#0FDDC4",
+        "ref:colour_tx": "#003A47",
         "route": "bus"
       }
     },
@@ -4574,6 +4578,8 @@
       "tags": {
         "network": "Consorcio de Transporte Metropolitano Área de Almería",
         "network:wikidata": "Q5784309",
+        "ref:colour": "#009639",
+        "ref:colour_tx": "white",
         "route": "bus"
       }
     },
@@ -4586,6 +4592,8 @@
       "tags": {
         "network": "Consorcio de Transporte Metropolitano del Área de Granada",
         "network:wikidata": "Q5784314",
+        "ref:colour": "#009639",
+        "ref:colour_tx": "white",
         "route": "bus"
       }
     },
@@ -4599,6 +4607,8 @@
         "network": "Consorcio de Transporte Metropolitano del Área de Málaga",
         "network:short": "CTMAM",
         "network:wikidata": "Q6949617",
+        "ref:colour": "#009639",
+        "ref:colour_tx": "white",
         "route": "bus"
       }
     },
@@ -4612,6 +4622,8 @@
         "network": "Consorcio de Transporte Metropolitano del Campo de Gibraltar",
         "network:short": "CTMCG",
         "network:wikidata": "Q5784307",
+        "ref:colour": "#009639",
+        "ref:colour_tx": "white",
         "route": "bus"
       }
     },
@@ -4652,6 +4664,8 @@
         "network": "Consorcio Metropolitano de Transportes de la Bahía de Cádiz",
         "network:short": "CMTBC",
         "network:wikidata": "Q5784325",
+        "ref:colour": "#009639",
+        "ref:colour_tx": "white",
         "route": "bus"
       }
     },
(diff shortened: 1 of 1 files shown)


kjonosm:
Thank you @Agazzr for your suggestion. `ref:colour_tx` seems to be almost exclusively used in Spain. Could you document the key in the OSM wiki?

Agazzr:
Hello @kjonosm,
I have already created the documentation page for `Key:ref:colour_tx` in OSM Wiki. Here you have the link: https://wiki.openstreetmap.org/wiki/Key:ref:colour_tx.

I have detailed its importance for maintaining brand consistency in transit networks, especially when the official identity uses a specific pair of corporate colors. I also included a couple of examples.

Thank you for your suggestion!

mxdanger:
The white text on the following background colors [#9ACD32](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=9ACD32), [#0FDDC4](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=0FDDC4), [#62A70B](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=62A70B) fails sufficient contrast for WCAG AA large text but the other colors used do pass WCAG AA and some AAA. Is there a source for the colors? Such as route guide PDF, website or other example where the color is being used?

If the colors are exactly as the official source, then there's really no choice and I wouldn't have any issue with getting this merged. Data consumers can still add text outlines or shift the color to meet legibility.

Robot8A:
One issue I see with this PR is that the tags will also be copied onto the bus stops, and we probably only want to keep the colors for the bus routes.

data/transit/route/bus.json
```diff
@@ -2302,6 +2302,8 @@
       "tags": {
         "network": "Bizkaibus",
         "network:wikidata": "Q879270",
+        "ref:colour": "#9ACD32",
```
Dwarf-Nebula, line 2305: Why not use "colour=" here?

data/transit/route/bus.json
```diff
@@ -2302,6 +2302,8 @@
...
+        "ref:colour_tx": "white",
```
Dwarf-Nebula, line 2306: Why not use "colour:text=" here?

1ec5:
[`ref:colour=*`](https://wiki.openstreetmap.org/wiki/Key:ref:colour), `ref:colour_tx=*`, and `ref:colour_bg=*` were coined for `route=road`, because each highway route in Spain has a route marker that is color-coded but not color-coordinated. That is, the colors are important for distinguishing one route from another, but they aren’t necessarily consistent within a given network as they are in other countries. On the other hand, if all the routes in a bus network have the same colors, then this is not a property of the route that ought to be tagged individually on the route.

The [`ref:colour_tx=*`](https://wiki.openstreetmap.org/wiki/Key:ref:colour_tx) documentation you’ve written up seems inconsistent with the key’s predominant usage (on highway routes). Based on the explanation there, I’m afraid that iD’s relation label design has been misinterpreted. When possible, iD labels the route with:

* The logo of the `network=*`, in order to distinguish this network from other networks.
* The route number in a bubble colored according to `colour=*`, in order to distinguish this route from other routes within the same network – not from other routes in other networks.

To put a finer point on it, NSI’s McDonald’s entry doesn’t have `ref:colour_tx=#ffbc0d` `ref:colour_bg=#db0007` just so that the restaurant number appears in the chain’s brand colors in iD. That would be mapping for the editor. It would be pointless, as the logo is already present by virtue of an NSI entry.

bhousel:
Bump?  This has been sitting for a while and I don't know what to do with this. 
As far as I can tell these tags seem unaccepted?

1ec5:
I have a feeling this particular use of the keys wouldn’t down well with the global community. There’s already so much skepticism about the way they’re used on highway routes, which at least has a pretty solid justification.

bhousel:
> I have a feeling this particular use of the keys wouldn’t down well with the global community. There’s already so much skepticism about the way they’re used on highway routes, which at least has a pretty solid justification.

Yes but it's only used in Spain and it's apparently supported by OpenMapTiles?
It seems unfair to say that the Spanish community can't have their tag because the "global community" (whatever that means) might say no.

I mean, yeah I'd reject the PR if it touched the [main McDonald's entry](https://github.com/osmlab/name-suggestion-index/pull/12103#issuecomment-4756867524) - that's a strawman argument.

I guess I'm just trying to determine - how likely is it that someone will complain if I accept this?

1ec5:
The Spanish community is using these keys for semantic reasons on highway routes but for presentational reasons on public transportation routes. OpenMapTiles only supports the keys on highway routes.

The semantic use is more defensible than the presentational one. If I understand correctly, this PR would open the door to coloring a restaurant preset based on its logo, but I’d be happy to be corrected.

bhousel:
Ok, you've convinced me - we can close it.

PR #12108, opened 2026-04-20
Add Raiffeisen stavební spořitelna

Oivo35:
Added Czech bank with more than 50 locations.

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -23,6 +23,16 @@
   },
   "items": [
     {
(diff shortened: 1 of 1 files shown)


kjonosm:
@Oivo35 Thank you! How is this different to existing `Raiffeisenbank (Česko)`?

Oivo35:
@kjonosm 
`Raiffeisenbank (Česko)` is a commercial bank accepting deposits and providing loans. It provides credit cards, insurance, business loans, investment products, etc.
`Raiffeisen stavební spořitelna` is only a home (building) savings bank.

PR #12132, opened 2026-04-26
Colizeum game clubs chain added (Russia, CIS, etc.)

empty-child:


bhousel:
Interesting!  I think this is the first dedicated e-sports venue I've seen.  

I guess tagging it as an "internet cafe" is ok for now, but I do wonder whether it's worth a conversation on the forums to discuss tagging.  We can always change the osm tags later- getting the `brand`/`brand:wikidata` pair locked in is the important thing.

empty-child:
I checked other possible tags for this type of venue, but internet_cafe is the nearest in meaning. And, frankly, I don't see enough reasons to extract game clubs into separate POI: big intersection in services. That's why "service:gaming": "yes" as an additional flag

PR #12133, opened 2026-04-26
Zgva Ludi alcohol shop added (Georgia)

empty-child:


bhousel:
Thanks!  If you have the time, please try to add some more supporting to claims to Q136806338 so that the Wikidata folks don't delete it.

PR #12143, opened 2026-04-28
move battery shop into shop=battery per community discussion

andrewharvey:
https://community.openstreetmap.org/t/shop-exclusively-for-batteries/143333 showed 7 :+1: for using `shop=battery` for a battery shop, so moving this battery shop into a more appropriate tag.

Proposed mass edit suggested at https://discord.com/channels/413070382636072960/471231032645910529/1498505094055067710

### DIFF
--- a/data/brands/shop/battery.json
+++ b/data/brands/shop/battery.json
@@ -1,0 +1,19 @@
+{
+  "properties": {
+    "path": "brands/shop/battery",
+    "exclude": {"generic": ["^battery$"]}
+  },
+  "items": [
+    {
+      "displayName": "Battery World",
+      "id": "batteryworld-31899b",
+      "locationSet": {"include": ["au"]},
+      "tags": {
+        "brand": "Battery World",
+        "brand:wikidata": "Q124003678",
+        "name": "Battery World",
+        "shop": "battery"
+      }
+    }
+  ]
+}
--- a/data/brands/shop/electrical.json
+++ b/data/brands/shop/electrical.json
@@ -5,17 +5,6 @@
   },
   "items": [
     {
-      "displayName": "Battery World",
-      "id": "batteryworld-31899b",
-      "locationSet": {"include": ["au"]},
-      "tags": {
-        "brand": "Battery World",
-        "brand:wikidata": "Q124003678",
-        "name": "Battery World",
-        "shop": "electrical"
-      }
-    },
-    {
       "displayName": "CEF",
       "id": "cef-7c8ef7",
       "locationSet": {"include": ["gb"]},


Kovoschiz:
`=electrical` is supported by mainstream apps (with CoMaps assumed from Organic Maps). For me, I would be reluctant to re-tag it. https://taginfo.openstreetmap.org/tags/shop=electrical#projects
Besides, this still doesn't distinguish removable batteries or battery charger of consumer electronic devices vs non-consumer electronics, vehicle, and electrical or industrial equipment. The former may be more common, prominent, and well-known. 
Is any rule about /openstreetmap/id-tagging-schema on changing away from a preset, and adopting a non-preset ?

andrewharvey:
The wiki says shop=electrical is

> An electrical store is a shop focused on selling different electrical supplies (cable, cable channels, electric sockets etc) and electrical devices (electric meters, junction boxes, fuse boxes etc).

ie. equipment an electrician would need.

That’s quite different to a shop just selling batteries.

> Besides, this still doesn't distinguish removable batteries or battery charger of consumer electronic devices vs non-consumer electronics, vehicle, and electrical or industrial equipment. The former may be more common, prominent, and well-known.

`shop=battery` could cover any or all of those. Take a look at https://www.batteryworld.com.au/products which is what this particular entry sells.

> Is any rule about /openstreetmap/id-tagging-schema on changing away from a preset, and adopting a non-preset ?

I can open a PR for it to be a preset, if we are reluctant to add the NSI because there is no Preset, and reluctant to add a Preset because there is no usage, well there is no usage because the NSI entries don't support it, we'll never have new tags that way.

It's clear from the discussion thread there is support from the community, the local community seems happy to have the NSI entry changed, so let's do it and wider downstream support can follow.

Kovoschiz:
As you replied there, let's continue to discuss the concept in Discourse. To highlight the findings there, NSI has a `=car_parts` for car batteries. Seems strange to have this, then `=battery` for all batteries only. 
For presets, shouldn't there be some organic use first?

mxdanger:
@Kovoschiz All I can say is at this point you are hindering progress and not making a good argument. Trying to incorrectly use tagging of shop=car_parts + car_parts=battery for a generic shop that sells, for example (unrelated to Battery World), all kinds of lithium batteries from rc cars to ebikes is bad suggestion.

I mean just take a look at all the [Overture categories](https://wiki.openstreetmap.org/wiki/Overture_categories). Clearly there is some demand for a generic `battery_store` category. NSI doesn't move at a glacial pace anyhow, if Battery World needs to be changed to something else then that's an easy change to make here and the 2 dozen that exist in the OSM database.

Your average 1 edit user on iD trying to add a shop or their own store that sells batteries is going to search the preset in iD and if it's not there just give up or pick the next closest option even if it doesn't make sense, organic growth is really just left up to power users. Leave this up to debate on openstreetmap/id-tagging-schema for presets like "Battery shop", "Car battery store", "Solar equipment supplier", "Battery manufacturer"

> Is any rule about /openstreetmap/id-tagging-schema on changing away from a preset, and adopting a non-preset ?

It would be using the generic shop preset. However, given the community support I don't see any reason not to merge this pr. All the instances of Battery World use shop=battery anyways.

Kovoschiz:
@mxdanger You are making things up. When did I say to use `=car_parts` for non-car batteries? My question for that was the adoption of `=battery` would lead to `car_parts=battery` in need of change to avoid 2 competing formats in parallel. 
`car_parts=battery` was also argued by one user to be not `=car_parts` if it doesn't sell multiple car parts. However, I disagree with this from the use of `car_parts=battery` for specializing in batteries.

mxdanger:
`car_parts=*` is used almost exclusively (96% combination [according to taginfo](https://taginfo.openstreetmap.org/keys/car_parts#combinations)) with `shop=car_parts`. Adoption of `shop=battery` would not be competing in parallel, rather it would be one generalized encompassing POI category (**Battery shop**) and a more specific `shop=car_parts` + `car_parts=battery` (**Car battery store**).

Just think how easy. Editor searches for "battery" and sees those two preset names. They're going to just pick the car battery shop preset if the shop just sells car batteries. If the shop sells more than just automotive batteries, then they're picking the other option. It's not all or nothing, both can coexist.

Kovoschiz:
I'm not talking about the presets, but the tags. Using `=battery` for all batteries favors shifting away from `car_parts=battery` for cars. `shop=` already has `=tyres` , and some thinks the `car_parts=` list is worth promoting to individual top-level features
#12234 practical question: If it mainly sells car batteries, but also some other batteries, should it be `car_parts=battery` for the characterization, or `=battery` for the breadth? Is the former expected to be specialized, and the latter having similar proportions?

andrewharvey:
>  Using =battery for all batteries favors shifting away from car_parts=battery for cars. shop= already has =tyres , and some thinks the car_parts= list is worth promoting to individual top-level features

If the shop is mostly a car battery shop, then `shop=car_parts` + `car_parts=battery` is fine to continue using.

> If it mainly sells car batteries, but also some other batteries

Like with all shops, best to focus on what the shop is promoted as and predominantly focuses on. We aren't trying to tag the whole inventory. A butcher that also sells some spices to go with the meat doesn't become a `shop=butcher;spices`, it's still a `shop=butcher`.

bhousel:
Ugh, I actually agree with this, but from what I can tell `shop=battery` isn't accepted.  It has a wiki stub page and only 140 uses in OSM.

https://wiki.openstreetmap.org/wiki/Tag:shop%3Dbattery

Let's put it in draft until there is more consensus around the tag.

andrewharvey:
I've expanded the wiki page.

PR #12151, opened 2026-05-02
Update O!Save hard discount store brand

LuisAPI:
- Add Indonesia to chain that was previously only in the PH
- Add common misspellings as matching names, per [Taginfo search](https://taginfo.openstreetmap.org/search?q=name%3Do%21save)


UKChris-osm:
Thanks @LuisAPI 👍 

Just an FYI that our build script has removed the `matchNames`, as the built-in matcher already considers those options.

PR #12158, opened 2026-05-05
Added international ramen chain

ShinyDog2636:


data/brands/amenity/restaurant.json
```diff
@@ -11735,6 +11735,31 @@
...
+          "my",
+          "ph",
+          "th"
+        ]
+      },
+      "tags": {
+        "amenity": "restaurant",
+        "brand": "Hokkaido Ramen Santouka",
+        "brand:en": "Hokkaido Ramen Santouka",
+        "brand:ja",:"らーめん山頭火",
```
LaoshuBaby: Please re-check the json’s format

ShinyDog2636 (commented):
Sorry. Fixed syntax errors

LaoshuBaby:
Sorry I noticed another things that it include some major market and not using English, so I separate it with another entry. Thank you for let us know this brand chain because it only have 4 shops in Tokyo area so maybe that's the reason that it haven't been added by Japan living mappers until now.

PR #12167, opened 2026-05-06
Add places leisure to UK Sports Centre Brands

Diacius:
Adding Places Leisure who operate about 90 leisure centres in England and Scotland.

### DIFF
--- a/data/brands/leisure/sports_centre.json
+++ b/data/brands/leisure/sports_centre.json
@@ -307,11 +307,7 @@
       "tags": {
         "brand": "Places Leisure",
         "brand:wikidata": "Q130223830",
-        "fee": "yes",
-        "leisure": "sports_centre",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
Thank you for your first submission @Diacius - although I would suggest removing the `operator` information as it generally repeats what is in the `brand` info (such as using the same wikidata number) and the `fee` tag as we try to focus mainly on `name` / `brand` / `operator` tagging with the NSI.

Diacius:
> Thank you for your first submission @Diacius - although I would suggest removing the `operator` information as it generally repeats what is in the `brand` info (such as using the same wikidata number) and the `fee` tag as we try to focus mainly on `name` / `brand` / `operator` tagging with the NSI.

No worries, will do, I’d copied from another similar entry and wasn’t sure what needed to be included.

Diacius:
Thanks!

PR #12169, opened 2026-05-07
Update subway operator details to Macao Light Rapid Transit

qgmzmy:


data/transit/route/subway.json
```diff
@@ -1699,12 +1699,13 @@
         "network:zh": "澳門輕軌",
         "network:zh-Hans": "澳门轻轨",
         "network:zh-Hant": "澳門輕軌",
-        "operator": "香港鐵路有限公司 MTR Corporation",
-        "operator:en": "MTR Corporation",
+        "operator": "澳門輕軌股份有限公司 Macao Light Rapid Transit Corporation, Limited",
+        "operator:en": "Macao Light Rapid Transit Corporation, Limited",
+        "operator:pt": "Sociedade do Metro Ligeiro de Macau, S.A.",
         "operator:wikidata": "Q1479375",
```
Vectorial1024: The Wikidata tag should probably also be updated (if exists).

Vectorial1024:
The Wikidata tag should probably also be updated (if exists).

Vectorial1024 (approved):

bhousel:
We generally don't include corporate forms like "Limited" and "S.A." in the tag values.  There have even been concerted efforts to remove these from OSM in the past.  

If there is a strong community preference to include them, then sure, but most communities prefer not to.  
Just wanted to ask "Are you really sure?" before I merge.

Vectorial1024:
To clarify, I am only here because I scrolled by this and noticed the PR back then did not update the Wikidata field.

This Macao Light Rail started as a subsidiary of HK MTR, and I would agree/prefer to remove the "Limited" from MTR even before the change of operators.

qgmzmy:
> We generally don't include corporate forms like "Limited" and "S.A." in the tag values. There have even been concerted efforts to remove these from OSM in the past.
> 
> If there is a strong community preference to include them, then sure, but most communities prefer not to. Just wanted to ask "Are you really sure?" before I merge.

Not really sure either, I was just following what the wiki seemed to suggest.

PR #12178, opened 2026-05-08
Add Triodos Bank

Robot8A:


Cj-Malone:
~~Do they have physical locations?~~ Seems like they might, I've heard of them in terms of ethical banks, but in terms of physical locations.

Robot8A:
They do have, I mapped this one a couple of years ago (survey) https://www.openstreetmap.org/node/11944368959

PR #12179, opened 2026-05-08
Update bus.json

HayaKoh-WeldyAlin:
I edited and added information about bus companies in Japan.

- ジェイ・アール北海道バス / JR Hokkaido Bus
- ジェイアールバス東北 / JR Bus Tohoku
- ジェイアールバス関東 / JR Bus Kanto
- ジェイアール東海バス / JR Tokai Bus
- 西日本ジェイアールバス / West Japan JR Bus
- JRバス中国 / JR Bus Chugoku
- ジェイアール四国バス / JR Shikoku Bus
- JR九州バス / JR Kyushu Bus
- ヤサカバス / Yasaka Bus
- 京都バス / Kyoto Bus
- 京都市営バス（京都市交通局） / Kyoto City Bus (Kyoto Municipal Transportation Bureau)

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -8477,20 +8477,6 @@
       }
     },
     {
-      "displayName": "JRバス関東",
-      "id": "jrbuskanto-54d9ea",
-      "locationSet": {
-        "include": ["jp-12.geojson"]
-      },
-      "tags": {
-        "network": "JRバス関東",
-        "network:en": "JR Bus Kanto",
-        "network:ja": "JRバス関東",
-        "network:wikidata": "Q10851155",
-        "route": "bus"
-      }
-    },
-    {
       "displayName": "JTA (Jackson)",
       "id": "jacksontransitauthority-ae7ebe",
       "locationSet": {
@@ -25553,6 +25539,13 @@
       "locationSet": {
         "include": ["jp-01.geojson"]
       },
+      "matchNames": [
+        "JR北海道バス",
+        "ジェイアール北海道バス",
+        "JRバス北海道",
+        "ジェイアールバス北海道",
+        "北海道ジェイアールバス"
+      ],
       "tags": {
         "network": "ジェイ・アール北海道バス",
         "network:en": "JR Hokkaido Bus",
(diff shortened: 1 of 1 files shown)


HayaKoh-WeldyAlin:
By the way, the `bus.json` file also included data for larger bus companies such as Hankyu Bus and Hanshin Bus, categorised by region—is this perhaps excessive?

Example:
```
    {
      "displayName": "千里ニュータウン線",
      "id": "senrinewtownline-b748f9",
      "locationSet": {
        "include": ["jp-27.geojson"]
      },
      "note": "looks like it is part of 阪急バス? At least its operator is 阪急バス.",
      "tags": {
        "network": "千里ニュータウン線",
        "network:en": "Senri New Town Line",
        "network:ja": "千里ニュータウン線",
        "route": "bus"
      }
    },
```
Glossary:
阪急バス = Hankyu Bus, Q11656745
千里ニュータウン = Senri New Town, New towns in northern Osaka Prefecture

I was unable to find any Wikidata entries directly related to this example.
In Japan, there are many bus companies that are subsidiaries of private railway operators in particular, and these companies sometimes use multiple bus lines with the same number within the same organisation.
In that case, this might help you tell the difference.
However, if these are divided too extensively, mappers may be unable to determine which group a bus route belongs to (or may use tags of companies that have not been defined), which could prevent NSI from functioning properly.
What would be the ideal way to operate these NSIs?

bhousel:
> However, if these are divided too extensively, mappers may be unable to determine which group a bus route belongs to (or may use tags of companies that have not been defined), which could prevent NSI from functioning properly.
> What would be the ideal way to operate these NSIs?

A good way to think of NSI is that it's like _"if this then that"_ for OSM.

The code basically does:  
IF we see this string "千里ニュータウン線" on a `route=bus` in this region "jp-27.geojson", 
THEN these are the tags that we believe that feature should have.

For situations that don't fit this model ("blah" sometimes A and sometimes B within the same location") it won't be a good fit for NSI.

You can always specify narrower geofence regions, even custom ones, but they ideally shouldn't overlap.

PR #12182, opened 2026-05-10
Fixed brands which left Russia and added "import-substituted" surrogate brands

empty-child:


data/brands/amenity/fast_food.json
```diff
@@ -2842,6 +2842,23 @@
...
+      "displayName": "DOMИNO PIZZA",
+      "id": "dominopizza-7582d1",
+      "locationSet": {"include": ["ru"]},
+      "matchNames": ["domino's", "доминоc пицца"],
+      "tags": {
+        "amenity": "fast_food",
+        "brand": "DOMИNO PIZZA",
+        "brand:wikidata": "Q139597698",
+        "cuisine": "pizza",
+        "name": "DOMИNO PIZZA",
```
Cj-Malone, line 2855: What language is "DOMИNO PIZZA"? It should be repeated in a "name:XX" tag. https://wiki.openstreetmap.org/wiki/Multilingual_names#Repeating_name_with_language_specific_tag

Cj-Malone: `not:brand:wikidata` should be removed also

empty-child: It's a mix of Cyrillic and Latin letters to "play with fonts", to show how Russian И plays with N. Frankly, don't know to what language key to assign. Ready to use any other value, but this one is on the label. And I specifically left not:brand:wikidata, because it's a bootleg brand, not the original Domino's, but looks like it

Issue #12190, opened 2026-05-12
Sibuya — amenity=restaurant (ES, PT, AD)


Crashillo:
```json
{
  "displayName": "Sibuya",
  "locationSet": {
    "include": [
      "es",
      "pt",
      "ad"
    ]
  },
  "tags": {
    "amenity": "restaurant",
    "name": "Sibuya",
    "brand": "Sibuya",
    "brand:wikidata": "Q139770153",
    "cuisine": "japanese"
  }
},
```

**Notes to editors**

Created with https://ga-kevin-codes.codeberg.page/Suggest-A-Brand/
Wikidata item: https://www.wikidata.org/wiki/Q139770153
Data file: https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/amenity/restaurant.json

PR #12200, opened 2026-05-15
Add new restaurant entry for Sibuya

GA-Kevin-Codes:


GA-Kevin-Codes:
Resolves #12190 by @Crashillo

PR #12206, opened 2026-05-17
Added Chinese Name & Payment Methods for Sheng Siong Supermarket

YJJcoolcool:


UKChris-osm:
Hi @YJJcoolcool 

I don't think we generally add payment methods to entries within the NSI, except maybe for payment terminals such as ticket machines.

I'll leave this for now for others to comment.

YJJcoolcool:
Alright, makes sense, I've removed the payment types

PR #12209, opened 2026-05-17
moved Matöppet, Tempo and Handlar'n from convenience to supermarket

matthiasfeist:
After re-reading definitions and discussions in the Swedish OSM forum, I decided to move Matöppet, Tempo and Handlar'n to the supermarket category.
Also to distinguish them better from conventional convenience stores like "pressbyrån" or "seven-eleven"


matthiasfeist:
ping @sp1187 for your information, since you introduced these brands first in https://github.com/osmlab/name-suggestion-index/pull/10665

sp1187:
I guess it's a matter of definition. I set it as `convenience` based on the suggestion on the OSM wiki that "One suggestion for distinguishing between a supermarket and a convenience shop, is that a supermarket is large enough to use trolleys." (though ICA, Coop and even Hemköp should presumably be split into both based on that definition) and also the fact that Handlar'n and Matöppet explicitly define themselves as "närbutik" on their websites.

> Handlar’n är närbutiken på hörnet eller kanske ortens enda butik

> Matöppet är närbutikskedjan som alltid sätter kunden först

matthiasfeist:
yeah I've seen that too. I'm more thinking that there's not really any difference between an ICA nära (which is a supermarket) and any of those other stores. Hence my PR to move them to supermarkets

PR #12210, opened 2026-05-17
configured Norwegian seven-eleven to be "Nordic" (also usable in Sweden)

matthiasfeist:


UKChris-osm:
Hi @matthiasfeist 

I'm not sure I see the benefit of this entry, as its using the same Wikidata as the current [7-Eleven](https://nsi.guide/index.html?t=brands&k=shop&v=convenience&tt=q259340#7eleven-e1b4e3), and thus would result in the same tagging that's already available!

matthiasfeist:
Ah OK. I just added it since there are already different country version of SevenEleven. Like the Norwegian version just above my change.

So would it better to merge the Swedish and Norwegian version into one @UKChris-osm ?

UKChris-osm:
I'm not sure why the Norwegian one is a separate entry either to be honest, as again the end result seems to be the same tagging as the main 7-Eleven.

The only main difference I can see is the inclusion of `"preserveTags": ["^name"]` (and a couple of different `matchNames`) but I feel like that tag could simply be added to the original entry anyway, thus covering the entire 7-Eleven brand with a single entry (excluding the Asia languages entries).

I might leave this for others to consider, but regarding your PR, as you've added an entry for Sweden, you'd also need to exclude Sweden from the main 7-Eleven entry (by added `se` to the `exclude` listing within the `locationSet`).

Cj-Malone:
I'm not in favour of adding preserveTags=name to the global entry, it really lowers the quality of OSM over time.

matthiasfeist:
@UKChris-osm: I configured the 7-eleven entity that was tagged with Norway before to be a more generic "Nordics" entity
@Cj-Malone: I also remove the "preserve-tags" as per your comment.

Cj-Malone:
Some Norwegian mappers want weird names, that's probably why the split happened in the first place, I've restored it. And I've excluded SE from the generic.

I too am unsure what you are trying to achieve here? SE has now moved from the generic entry to the NO entry. The only difference between the them is NO allowing weird names. (match tags=shop/kiosk does nothing here and will be auto removed when the project is build)

bhousel:
Ah yeah, the Scandinavians protested on #5500 and we tried to give them their way.  😑 
This is why sometimes you'll see Norway/Sweden brands with their own entry and with `preserveTags: [/name/]`.

I don't remember the details, but I think someone was also running a script to sync store names up with an official business database, so they need the names to be a certain way for the script to work.

matthiasfeist:
Hmm OK. I wasn't aware of that.
Actually, I guess in that case I think I can close my PR since it doesn't achieve anything that the generic one doesn't already cover. I got mainly confused by the Norwegian 7-eleven and wanted to make sure the Swedish is tagged in the same way. 
Thanks for all your comments!

PR #12230, opened 2026-05-21
Correct wikidata for Bam

Robot8A:


mxdanger:
I took a look at the wikidata and it looks like it was modified by a now globally blocked user. See last valid edit before it was changed. https://www.wikidata.org/w/index.php?title=Q129652114&oldid=2236094161 What they should have done is merge Q129652114 with Q5718037 thus turning Q129652114 into a redirect. You're not allowed to repurpose QIDs for this exact reason.

Q129652114 is going to get merged and shouldn't be used for anything else anymore.

mxdanger:
For this PR you should also update the wikidata for Bam ATMs as seen here: https://nsi.guide/index.html?t=*&k=amenity&v=atm&tt=q129652114

Robot8A:
The ATM preset gets automatically populated from the bank preset

PR #12232, opened 2026-05-22
Supeco is the same brand in these 4 countries

ilias52730:


Cj-Malone:
This reverted https://github.com/osmlab/name-suggestion-index/commit/9e30077d09a4b6989fc01d9668725f53982b0c6a by @kjonosm. It seems intentionally split, https://www.supeco.ro/ looks different to https://www.supeco.es/ and has different logos, different social medias, different websites and different wikidata.

PR #12237, opened 2026-05-25
Add La Pizza de Nico restaurant chain

versgui:


mxdanger:
You should also add some social media Identifiers if it has them (Facebook, Instagram, etc) to the wikidata item. That way the icon can be fetched.

PR #12241, opened 2026-05-27
amenity(cafe): Expand Tealive's operating countries

LuisAPI:
Adds (alphabetically based on ISO 3166 code) the following countries in which Tealive also operates, aside from their home country of Malaysia: the UAE, Brunei, Canada, India, Cambodia, Myanmar, Mauritius, the Philippines, Singapore and Vietnam.
 
Source for that list is the Malay-language Wikipedia: https://ms.wikipedia.org/w/index.php?title=Tealive&oldid=6670539

(Considering they're still looking for more countries to expand to, maybe the proper way should instead be put the World -- 001 -- as their location set? Hmm)

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -4423,7 +4423,7 @@
     {
       "displayName": "Tealive",
       "id": "tealive-2aa1ef",
-      "locationSet": {"include": ["my"]},
+      "locationSet": {"include": ["ae", "bn", "ca", "in", "kh", "mm", "mu", "my", "ph", "sg", "vn"]},
       "tags": {
         "amenity": "cafe",
         "brand": "Tealive",


mxdanger:
You should also add that data to the wikidata item with operating area P2541. Looks like someone accidently added it to P17 country (which is used for country of origin). Would also be good to add P12454.

LuisAPI:
> Looks like someone accidently added it to P17 country (which is used for country of origin).

Understood. I read [the English Wikipedia infobox for companies](https://en.wikipedia.org/w/index.php?title=Template:Infobox_company&action=edit), and it does seem like that Wikidata property should be specifically renamed `country of origin` or `country of headquarters` to avoid such further conflation.

> You should also add that data to the wikidata item with operating area P2541.

I [just tagged](https://www.wikidata.org/w/index.php?title=Q95722097&diff=2499145163&oldid=2499143218) all those other `country` value claims as being "on the wrong property", and [moved them all](https://www.wikidata.org/w/index.php?title=Q95722097&diff=2499146200&oldid=2499145556) to the `operating area` property.

PR #12243, opened 2026-05-27
Add wikidata Vivari

Robot8A:


data/brands/amenity/cafe.json
```diff
@@ -5170,7 +5170,7 @@
       "tags": {
         "amenity": "cafe",
         "brand": "Vivari",
-        "cuisine": "coffee_shop",
```
UKChris-osm, line 5173: Any reason for removing this @Robot8A?

Robot8A: I feel like it's unnecessary with the "amenity"="cafe", but maybe we can put it back. I never fully got what that tag is supposed to mean.

UKChris-osm: All I can recall is that a warning appears if we missed out the `cuisine` tag for cafes and restaurants.

Issue #4053, opened 2020-07-30
nsi.guide canonical urls / SEO


bhousel:
I noticed that Googlebot is not crawling the nsi.guide site very well.  This is because we built it as a single page app (SPA) with react router parsing params to make the key/value subpages:

<img width="1137" alt="Screenshot 2020-07-30 13 41 31" src="https://user-images.githubusercontent.com/38784/88955902-936d4680-d26a-11ea-897a-6511018a948e.png">

Put another way, Google won't index these pages because it thinks they are all just non-canonical variations on the same page.  This is a [well known problem](https://blog.theodo.com/2019/11/your-react-app-deserves-proper-seo/) for single page apps.

I'm ok with **not** solving this, but it would be nice to figure out a way to get the site indexed better.  Some solutions are:
* Just generate a separate static copy of the entire site for crawlers - maybe we just do it on the GitHub page, maybe we leverage something like prerender.io.
* generate a sitemap to help Googlebot what the pages all are
* Rails Router trickery?  Maybe we force full page refreshes if we detect a crawler and set appropriate `<link rel="canonical">` in the `<head>`, maybe using [React Helmet](https://github.com/nfl/react-helmet)
* put it on a real server that generates the pages and pushes them to the client.. The fancy word for this is [isomorphic](https://stackoverflow.com/questions/43064707/what-does-isomorphic-react-mean) (I'd rather not).

Just throwing out ideas / learning in the open.  I don't know what I'm doing.

UKChris-osm:
Does the [nsi.guide](https://nsi.guide) web site need to be completely indexed? In terms of SEO what would be the ultimate goal? That every page comes up highly in Google (such as a Costa search brings up the cafe page), or that the home page comes up high (such as a Costa search brings up the home page)?

At the moment, I expect there isn't much indexing going on because there isn't a great deal of content available, and most pages have the same content to begin with at the top (with the "Some things you can do here:" intro), and have the same page title. To get better indexed each page would need to be a bit more unique with the content on offer, such as the "amenity / cafe" page having a cafe brand introduction and examples.

I don't know if React can accommodate unique page titles easily? I don't think the URL's are much of an issue, as "?k=acb&v=xyz" are unique URL's and can be indexed as such, unless you tell Google to ignore the URL Parameters, and I think you can still tell it to explicitly follow URL Parameters via the Search Console.

[...]

PR #12247, opened 2026-05-28
chore: regenerate sitemap

Johannes-Andersen:
Sitemap was quite far behind main, so it was missing a few things, so why not regenerate it :)

Since it seems to have been disabled in a previous commit to avoid having to push the changed file on every dist generation, I moved it out to its own script, so Github actions can create a PR whenever a new tag is released.

Also added some code to get the last modification date from commit history to better utilise crawl budgets on changes URLs

### DIFF
--- a/.github/workflows/sitemap.yml
+++ b/.github/workflows/sitemap.yml
@@ -1,0 +1,35 @@
+name: regenerate sitemap
+
+on:
+  push:
+    tags:
+      - '**'
+  workflow_dispatch:
+
+env:
+  FORCE_COLOR: 2
+
+jobs:
+  sitemap:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v6
+      with:
+        ref: main
+        fetch-depth: 0    # full history so `git log` can compute per-file lastmod
+    - name: Setup Bun
+      uses: oven-sh/setup-bun@v2
+      with:
+        bun-version: latest
+    - run: bun install
+    - run: bun run dist:sitemap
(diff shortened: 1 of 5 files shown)


UKChris-osm:
Updatable `<sitemaps>` seems like a good idea 👍 I don't know enough about the core usage of `bun`, `typescript` and `packages` to know if this is the best way to do it, so I'll leave that for review by others, but thank you @Johannes-Andersen 👍

bhousel:
Ah yeah I did try a few things to get nsi.guide crawled better - these are in #4053

From what I remember, Google didn't really respond to any of the things that I tried.

I had hoped that the sitemaps could be generated dynamically by React, mostly because I really don't want to see a bunch of commits and git diffs that just update the paths and timestamps.  Site indexing worked best back when we were statically generating the whole nsi.guide, but that was really a lot of git noise, and the currrent React SPA approach allows search to work the way it does.

Johannes-Andersen:
Ah craph yeah, that sucks. Forgot that Google isn't really crawling and indexing SPAs that rely on URL params that well
I'll close this for now, and can maybe be looked into as part of getting canonical URLs

UKChris-osm:
> From what I remember, Google didn't really respond to any of the things that I tried.

Searching for "site:nsi.guide" shows a lot more indexed pages that the previous 4 that I saw back in 2020, so that's good 👍 

But, each page shows the same page title as "Name Suggestion Index", rather than perhaps a category name - could the page title be changed within SPA for each category, giving Google something else to focus on?

PR #12248, opened 2026-05-28
feat: modernize Wikidata warnings page

Johannes-Andersen:
As someone who uses https://nsi.guide/wikidata-warnings.htm a lot, there has been a few things I have been missing:
- Darkmode
- Ability to sort
- Ability to filter

This PR aims to solve that by adding the above mentioned features and modernizing the styling to look more like the main site.

Technical changes:
- `scripts/wikidata.ts` now also outputs a category
- `scripts/build_js.ts` creates its own bundle for the warning page (to avoid it bundling unrelated code to main site)
- Moved the page over to using React 

Demo:

https://github.com/user-attachments/assets/22d9e55e-1ccc-4151-8fef-a5d5dc6fcb31


bhousel:
Can we just have everything bundled into a single react app?  
I'm not enthusiastic about creating a separate app just for the warnings page.

Johannes-Andersen:
@bhousel I can try later to see if I can figure out a smart way to do it ☺️
Not a 100% sure how it can be easily done with Github pages since routing isn't really possible from my understanding

Johannes-Andersen:
There @bhousel, had a go at moving it into its own `view` param to select different pages ☺️

UKChris-osm:
The page is great @Johannes-Andersen 👍 

I remember creating the original page a couple years ago as a quick way to have a look at the warnings, but never got around to adding any type of filters or sorting, although @Snowysauce did some updates to it as well, so thanks for this.

bhousel:
Thanks!

PR #12252, opened 2026-05-29
feat: link to wikiwarnings and fix mobile styling

### LINKED ISSUE #12250: Better way to navigate to Wikidata Warnings
On the root page you have an instructions paragraph with several external links and then the Wikidata Warnings is just snuck in there with a hidden link named "improving". All this time I've never known it existed until I saw a PR for it. Now that #12248 is improving that page I think it may be worth moving the link to a more prominent location like the footer or header or just restructuring the instructions so the link is given a descriptive name. 

CC @Johannes-Andersen what's your thoughts.

Johannes-Andersen:
This is an attempt to fix https://github.com/osmlab/name-suggestion-index/issues/12250 by adding a more prominent link that encourages people to use https://nsi.guide/wikidata-warnings.htm to fix Wikidata issues.

It also resolves issues with mobile header responsiveness.

<details><summary>Before</summary>
before-index-mobile
before-index
before-page-mobile

</details> 

<details><summary>After</summary>
after-index-mobile
after-index
after-page-mobile

</details>

### DIFF
--- a/docs/style.css
+++ b/docs/style.css
@@ -74,11 +74,14 @@ input[type=text] {
   text-overflow: ellipsis;
 }
 .dark textarea,
-.dark input[type=text],
-.dark input[type=checkbox] {
+.dark input[type=text] {
   background-color: #666;
   color: #ddd;
   border: 1px solid #777;
+}
+.dark input[type=checkbox] {
+  accent-color: #666;
+  color-scheme: dark;
 }
 
 button {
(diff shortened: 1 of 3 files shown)


bhousel:
Thanks!

UKChris-osm:
Does the website now only update via a fully published release? Rather than the previous `npm run appbuild`?

Johannes-Andersen:
Looks like the latest tag published on jsdelivr is https://cdn.jsdelivr.net/npm/name-suggestion-index@7.2.20260515/dist/
From my quick and limited understanding, it does indeed need a new tag release to get the latest js files needed for the site. 

However, the css is loaded directly using relative paths. So its kinda breaking when the css isn’t versioned but the js is 😅

bhousel:
> However, the css is loaded directly using relative paths. So its kinda breaking when the css isn’t versioned but the js is

Good catch - I just fixed this in 3d67064c

Update:  I also needed to publish a new release because we hadn't published `docs/` in a long time, so `latest` was picking up a very old file.  But it looks like things are working now and I do see the latest `style.css` and the warnings page on `nsi.guide` now.

PR #12278, opened 2026-06-03
Move Husqvarna and Stihl

Oivo35:
Husqvarna and Stihl are categorised under `shop=hardware`, which is a shop that mainly sells metal fasteners (screws, bolts, nails, hooks). However, Stihl and Husqvarna sell lawn mowers, chainsaws or string trimmers, and therefore the `shop=groundskeeping` category is more appropriate. They often also have a local shop name, so they should have preserveTags.

### DIFF
--- a/data/brands/shop/groundskeeping.json
+++ b/data/brands/shop/groundskeeping.json
@@ -16,6 +16,17 @@
       }
     },
     {
+      "displayName": "Husqvarna",
+      "locationSet": {"include": ["001"]},
+      "preserveTags": ["^name"],
+      "tags": {
+        "brand": "Husqvarna",
+        "brand:wikidata": "Q899356",
+        "name": "Husqvarna",
+        "shop": "groundskeeping"
+      }
+    },
+    {
       "displayName": "Mountfield",
       "id": "mountfield-9aa91c",
       "locationSet": {"include": ["cz", "sk"]},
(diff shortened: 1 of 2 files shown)


bhousel:
I'm kind of surprised to see these brands in the index.  They don't operate the stores, they are just sold through resellers which are independently owned.  If you see a Stihl or Husqvarna sign outside a groundskeeping store, that doesn't make it a Stihl store or a Husqvarna store - they are probably also a dealer for other things and often also a repair shop.
(This was discussed a while ago on #5233).

Oivo35:
In that case, most car brands operating stores on the same principle as Husqvarna a Stihl should be removed from the NSI. Almost all car brands sell through dealerships which are independently owned. Similarly, authorised service centres listed in the NSI, such as VW, Opel, and Peugeot, also repair cars of other brands.

Husqvarna and Stihl use the same sales concept as car brands. They have authorised dealerships where you can buy only one brand of garden machinery (sometimes consumables of other brands, such as when VW sells tyres or roof boxes by other brands), or they sell their products alongside other brands.

</a>
</a>

bhousel:
Ok I guess they can stay, if people are really making concept stores like this.

The ones I'm familiar look more like this:
Screenshot 2026-06-04 at 11 53 21 AM

Oivo35:
Of course, there are also independent dealers who offer multiple brands, but then there are also retailers who only offer the Stihl brand. Similar shops can be found across the USA and Europe.

In USA

https://www.facebook.com/STIHLdealerburlington

</a>

In Europe

https://www.facebook.com/stihlemspol

</a>

https://www.vondryska.cz/

</a>

https://www.facebook.com/centrum.elektrotechniczne

</a>

PR #12291, opened 2026-06-07
Add some Catalan bakery and café chains

swaree:
Valero is a bakery chain in the Vallès region of Catalonia, with [29 locations](https://www.pavalero.com/ca/botigues) as of June 2026. `locationSet` is `es` rather than `es-b.geojson` to be more future-proof.

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -766,8 +766,7 @@
         "amenity": "cafe",
         "brand": "Boheme",
         "brand:wikidata": "Q140089658",
-        "name": "Boheme",
-        "shop": "bakery"
+        "name": "Boheme"
       }
     },
     {
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -759,6 +759,18 @@
       }
     },
     {
+      "displayName": "Boheme",
+      "id": "boheme",
+      "locationSet": {"include": ["es"]},
+      "tags": {
+        "amenity": "cafe",
+        "brand": "Boheme",
+        "brand:wikidata": "Q140089658",
+        "name": "Boheme",
+        "shop": "bakery"
+      }
+    },
+    {
       "displayName": "Bombay Cutting Chai",
       "id": "bombaycuttingchai-12649c",
       "locationSet": {"include": ["in", "qa"]},
--- a/data/brands/shop/pastry.json
+++ b/data/brands/shop/pastry.json
@@ -174,6 +174,22 @@
       }
     },
     {
+      "displayName": "Demasié",
+      "id": "demasie",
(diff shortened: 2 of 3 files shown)


UKChris-osm:
Thanks @swaree 👍 

Do the `name`, `brand` and `display name` need to be different? Is setting the `name` and `brand` simply as **Valero** not okay?

swaree:
Hi @UKChris-osm , the signage outside of the shops are not the same across locations, but I think Valero for all keys would work. I'll edit the commit, thanks!

data/brands/amenity/cafe.json
```diff
@@ -758,6 +758,18 @@
...
+    {
+      "displayName": "Boheme",
+      "id": "boheme",
+      "locationSet": {"include": ["es"]},
+      "tags": {
+        "amenity": "cafe",
+        "brand": "Boheme",
+        "brand:wikidata": "Q140089658",
+        "name": "Boheme",
+        "shop": "bakery"
```
UKChris-osm: Did you mean to leave the shop tag here @swaree? I think it'll need removing if `Boheme` is within the café category.

swaree: I based it on the `Santagloria` entry, which is a similar type of establishment, though that one is in `bakery.json`. What do you think is best?

UKChris-osm: I didn't realise `Santagloria` had both tags, I'm not 100% sure, but as it seems very few entries include both tags, I'd suggest playing it safe and removing the `shop` tag from the `Boheme` entry. We can always add it back if I'm wrong.

swaree:
That should be it for now, thanks for the comments! @UKChris-osm

data/brands/shop/pastry.json
```diff
@@ -173,6 +173,22 @@
...
+      "matchNames": [
+        "demasie",
+        "cookies demasie"
+      ],
+      "tags": {
+        "brand": "Demasié",
+        "brand:wikidata": "Q140089894",
+        "name": "Demasié",
+        "shop": "pastry",
+        "alt_name": "Demasie"
```
UKChris-osm, line 189: I don't think an `alt_name` would be required in this case, as the matcher (NSI scanner) is good at matching up names with acute characters.

UKChris-osm:
As you're adding to this PR @swaree I've converted it into a `draft` for now. Feel free to remove the `draft` tag when you've finished adding to your entry.

data/brands/shop/pastry.json
swaree: The rationale behind the `alt_name` is that the brand seems to use both names interchangeably, but definitely use the _Demasié_ /de.maˈsi̯e/ pronunciation, so I'd say it's a stylistic choice while the original is the accented one.

PR #12312, opened 2026-06-11
Tax, customs and border control offices (Poland)

rogoyski:
This PR replaces https://github.com/osmlab/name-suggestion-index/pull/12006 as the author didn't react there to the proposal of of including my changes in his PR.

### DIFF
--- a/data/operators/office/government.json
+++ b/data/operators/office/government.json
@@ -1180,24 +1180,34 @@
     },
     {
       "displayName": "Krajowa Administracja Skarbowa (urzędy i oddziały celne)",
+      "id": "krajowaadministracjaskarbowa-995fc9",
       "locationSet": {"include": ["pl"]},
-      "matchNames": ["urząd celno-skarbowy, izba celna", "urząd celny", "oddział celny"],
+      "matchNames": [
+        "oddział celny",
+        "urząd celno-skarbowy, izba celna",
+        "urząd celny"
+      ],
       "tags": {
-        "office": "government",
         "government": "customs",
+        "office": "government",
         "operator": "Krajowa Administracja Skarbowa",
+        "operator:short": "KAS",
         "operator:type": "government",
         "operator:wikidata": "Q30904637"
       }
     },
     {
       "displayName": "Krajowa Administracja Skarbowa (urzędy skarbowe)",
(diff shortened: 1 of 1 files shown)


UKChris-osm:
In this case you've got "_Krajowa Administracja Skarbowa_" appearing as the `operator` multiple times in the same file, which is causing a build fail.

To use the same operator multiple times when referring to a different department, in this case "_Urząd Celno-Skarbowy_", "_Oddział Celny_" and "_Urząd Skarbowy_", you'd need to put them in the `brand` tree, or have each one as the `operator` separately instead of "_Krajowa Administracja Skarbowa_".

rogoyski:
I see. If I combined them in one entry, I would not be however able to specify different values for `government=*`, which was the main point of this initiative, because people get it wrong all the time. Would be nice if the value for `government=*` was included in that ID hashing.

If I moved this to the _brand_ tree, I would have to specify a `brand` for each entry, right?

I found a way to cheat the system too, but you may not like it ;)

UKChris-osm:
> I found a way to cheat the system too, but you may not like it ;)

I'm not actually sure - it does create unique ID's which is important! I'll have to wait for others opinion!

bhousel:
Ah yeah, using one locationSet that includes 'pl' and another locationSet that draws a circle around Poland to "cheat the system" will result in problems.   We really don't want our index to contain duplicate entries in the same location, because this index powers Rapid and iD validator and that code will have a hard time determining what is really meant when it sees certain name values.

rogoyski:
I'm fine with the verdict on _Administracja Celno-Skarbowa_, but then you could at least let me merge the _Straż Graniczna_ entry. Instead, the whole PR is closed, while for example the original one by @DevManiak17 still isn't :confused: I will resubmit it.

I also must note that by calling these _duplicate entries in the same location_ you seem not to acknowledge that they are only duplicate from the perspective of NSI's technical design. From the perspective of OpenStreetMap and mapping workflows in general, which I guess you are trying to facilitate, both are distinctive and valuable real-world types of features (POIs). So ultimately it is arguably a flaw in NSI's design that standardising both entries is not possible. I understand you will not turn the design up side down because of this case, but if I were in your place, I would acknowledge this when declining a PR.

PR #12315, opened 2026-06-13
Update Audika operating countries

saugatkarke:
Updates Audika’s locationSet to include the countries listed in the issue and source page.

Note: I used `dk` for Denmark instead of `da` from the original issue text, because NSI uses country/location codes and `dk` is the correct code for Denmark.

Source: https://www.audikagroup.com/find-local-clinic


GA-Kevin-Codes:
@saugatkarke Please see https://ideditor.codes for regions, `DK` is the Kingdom of Denmark (the whole kingdom), Denmark proper would be Q35 as an example. Play around on there and make sure the locations you set are wide enough to capture where you expect a place to be, but not elsewhere. Thanks for your PR!

saugatkarke:
Thanks for the catch! I didn't realise "dk" included the autonomous territories. 
I've updated the location set to use "Q35" for Denmark proper so it excludes Greenland and the Faroe Islands.

GA-Kevin-Codes:
Thank you for your contribution to the NSI!

PR #12321, opened 2026-06-15
Update brand name Spar Express convenience.json

BruniSchulz:
update brand name to 'Spar Express':
SPAR Express is an own brand within the SPAR group, only the display name is 'Spar Express' but not the brand name 'Spar'. references: https://spar-international.com/aboutus/store-formats/ Wikidata Q-Code for Spar Express (supermarket brand of SPAR) is also available Q12321004 and should be changed. The same change should  also be applied to "Spar Express (Norge)"

### DIFF
--- a/data/brands/shop/convenience.json
+++ b/data/brands/shop/convenience.json
@@ -5651,8 +5651,8 @@
       },
       "matchTags": ["shop/supermarket"],
       "tags": {
-        "brand": "Spar",
-        "brand:wikidata": "Q610492",
+        "brand": "Spar Express",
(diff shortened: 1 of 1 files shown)


Cj-Malone:
rel https://github.com/osmlab/name-suggestion-index/discussions/12306

GA-Kevin-Codes:
Seems to me that we (and Wikidata) should have the 4 different types as that's what's on the door from what I can tell. Wikidata would also benefit from an identically named parent group linking all the sub-brands. Thanks for your work on this!

PR #12323, opened 2026-06-15
Update atm.json Adding new brand OTP Banka in SI

vinzdb:
Added new ATM network in Slovenia
https://www.otpbanka.si/
OTP Banka Slovensko (Q13537695)

### DIFF
--- a/data/brands/amenity/atm.json
+++ b/data/brands/amenity/atm.json
@@ -708,6 +708,15 @@
         "brand:wikidata": "Q709259",
         "brand:zh": "中華郵政"
       }
+    },
+    {
+      "displayName": "OTP Banka",
+      "locationSet": {"include": ["si"]},
+      "tags": {
+        "amenity": "atm",
+        "brand": "OTP Banka",
+        "brand:wikidata": "Q13537695"
+      }
     }
   ]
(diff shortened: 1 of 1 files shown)


Dimitar5555:
OTP Banka is a bank and it looks like it operates as such in Slovenia. They also have what looks like bank offices.

https://www.otpbanka.si/branches-atm

vinzdb:
Thanks. OTP Banka operates both bank branches and branded ATMs in Slovenia. This entry was added for standalone amenity=atm features, which are commonly mapped separately in OSM.

I noticed there is currently no corresponding amenity=bank entry in NSI. I can add the bank entry as well so that both feature types are represented consistently.

Dimitar5555:
Having it as bank only is enough. All banks are automatically duplicated from `banks.json` to `atm.json`.

https://github.com/vinzdb/name-suggestion-index/blob/patch-9/data/brands/amenity/atm.json#L9-L26

vinzdb:
Thanks for the clarification. I wasn't aware that bank entries are automatically duplicated from banks.json to atm.json.

I'll close this PR and add OTP Banka under amenity=bank.

PR #12324, opened 2026-06-16
Add a new bank entry OTP Banka in Slovenia

vinzdb:
Added new bank entry for Slovenia
https://www.otpbanka.si/
OTP Banka Slovensko (Q13537695)


Dimitar5555:
Thanks!

PR #12341, opened 2026-06-18
Remove fake names from Dutch parcel lockers

Dwarf-Nebula:
These objects don't really have a name according to the OSM standards. So, the name tag shouldn't be added.


Dwarf-Nebula:
Should have scrolled trough the pull requests. There already is a pull request. #11104

PR #12342, opened 2026-06-18
Move Utrecht Binnen from tram to light rail

Dwarf-Nebula:
The tram in Utrecht is tagged as light_rail, since it's more light_rail than a tram. This should make sure the NSI entry is for the right route type.

### DIFF
--- a/data/transit/route/light_rail.json
+++ b/data/transit/route/light_rail.json
@@ -82,6 +82,21 @@
       }
     },
     {
+      "displayName": "Utrecht Binnen",
+      "id": "utrechtbinnen-50b0a0",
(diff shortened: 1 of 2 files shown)


bhousel:
Did you mean to remove the `operator` and `brand` tags?

Dwarf-Nebula:
> Did you mean to remove the `operator` and `brand` tags?

I didn't remove them, right? I've added them. Or should these be in another file?

bhousel:
> I didn't remove them, right? I've added them. Or should these be in another file?

oh sorry, I misread the diff 🙈   yeah it looks correct.

PR #12348, opened 2026-06-19
Add CDEM & NZRT

k-yle:
i'm still pretty confused about NSI's usage of `brand` vs `operator`, but this seems to work

### DIFF
--- a/data/brands/emergency/disaster_response.json
+++ b/data/brands/emergency/disaster_response.json
@@ -1,0 +1,41 @@
+{
+  "properties": {
+    "path": "brands/emergency/disaster_response",
+    "exclude": {
+      "generic": ["^disaster response$"]
+    }
+  },
+  "items": [
+    {
+      "displayName": "Civil Defence",
+      "id": "civildefence-75c8ff",
+      "locationSet": {"include": ["nz"]},
+      "matchNames": [
+        "cdem",
+        "emergency management",
+        "national emergency management",
+        "nema"
+      ],
+      "tags": {
+        "brand": "Civil Defence",
+        "brand:wikidata": "Q6866662",
+        "emergency": "disaster_response",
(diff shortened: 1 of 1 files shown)


k-yle:
well, now i'm really confused. why are supermarkets allowed duplicate `brand`s/`operator`s, but not this file?

bhousel:
> well, now i'm really confused. why are supermarkets allowed duplicate brands/operators, but not this file?

They aren't allowed to be duplicate.  
When I pull the changes in your branch and `bun run build` I get an error.

```
build:json | Error - Duplicate id 'nationalemergencymanagementagency-5beaa6' in:
build:json |   New Zealand Response Team
build:json |   ./data/operators/emergency/disaster_response.json
build:json | Exited with code 1
error: script "build" exited with code 1
```

UKChris-osm:
Even though these two entries have unique `brands`, you've placed them in the `operator` tree in the same `disasterter_response.json` file with matching operator information, i.e. "_National Emergency Management Agency_", which is causing the duplicate error, as the `operator` data takes priority within the `operator` tree, and in this case, "_National Emergency Management Agency_" would appear twice.

In the "_Civil Defence_" entry, you've also used the same wikidata as "_National Emergency Management Agency_".

In this case, you'd need to either place both entries under the `brand` tree, which would allow the `brand` data to take priority (and thus remove the clash) while also including the `operator` data., or, remove the branding and focus on a single `operator` entry, or have the the new `brands` that you've added actually be the `operator`.

k-yle:
thanks both, that clears up my misunderstanding! another thing I didn't realise was that `brand:wikidata` may have duplicate values, but `brand` cannot.

anyway, that error should be fixed now. i didn't see it at first, because `bun run all` did nothing on my computer[^1] 😅 

image


[^1]: not sure why, but reinstalling bun fixed it

PR #12366, opened 2026-06-23
Add operator (:type) Gemeente Utrecht

Dwarf-Nebula:


data/operators/amenity/bicycle_parking.json
```diff
@@ -190,6 +190,17 @@
         "operator:wikidata": "Q2355889",
         "operator:zh": "運輸署"
       }
+    },
+    {
+      "displayName": "Gemeente Utrecht",
+      "id": ,
```
Dwarf-Nebula: `build | build:json | Error - JSON5: invalid character ',' at 196:13 reading:`

Should the id field be left out completely? Or an empty string instead of nothing?

Cj-Malone: Either, the id will be generated but for now the json needs to be valid.

```suggestion
```
```suggestion
      "id": "",
```

PR #12376, opened 2026-06-26
Clean up DHL parcel_locker DE

### LINKED ISSUE #12349: Clean up amenity parcel_locker - DHL [DE]
I'm currently discussing with the community to make some changes for `amenity=parcel_locker` for the entries regarding DE and DHL/Deutsche Post:

- Entry "brand": "Deutsche Post" would be deleted, because Deutsche Post doesn't exist as brand for parcels/lockers and they also do not operate any of these boxes - it's all DHL.
- fake name "name": "DHL Paketbox" should be removed, it doesn't have any meaning as wikidata is already present.
- operator and operator:wikidata pointing to DHL should be added to the entries "DHL Paketbox", "DHL Packstation", "DHL Poststation" (this is documented here [Wiki-Packstation](https://wiki.openstreetmap.org/wiki/Packstation) and established tagging)
- potentially add matchnames to point to the correct entries if mistakenly (the confusions happens quite often) using "Deutsche Post". E.g. "Deutsche Post Packstation" should match with "DHL Packstation" etc.

I will open a PR after discussion of these points is agreed on.

Aimmig:
Closes #12349

### DIFF
--- a/data/brands/amenity/parcel_locker.json
+++ b/data/brands/amenity/parcel_locker.json
@@ -285,17 +285,21 @@
       "tags": {
         "amenity": "parcel_locker",
         "brand": "DHL Packstation",
-        "brand:wikidata": "Q1766703"
+        "brand:wikidata": "Q1766703",
+        "operator": "DHL",
+        "operator:wikidata": "Q489815"
       }
     },
     {
       "displayName": "DHL Paketbox",
-      "id": "dhlpaketbox-83cc4d",
+      "id": "paketbox-83cc4d",
       "locationSet": {"include": ["de"]},
       "tags": {
         "amenity": "parcel_locker",
         "brand": "Paketbox",
-        "brand:wikidata": "Q2046604"
+        "brand:wikidata": "Q2046604",
+        "operator": "DHL",
+        "operator:wikidata": "Q489815"
       }
     },
     {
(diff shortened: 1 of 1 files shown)


data/brands/amenity/parcel_locker.json
```diff
@@ -292,22 +281,29 @@
...
     },
     {
       "displayName": "DHL Paketbox",
-      "id": "dhlpaketbox-83cc4d",
+      "id": "paketbox-83cc4d",
       "locationSet": {"include": ["de"]},
+      "matchNames": ["deutsche post paketbox"],
       "tags": {
         "amenity": "parcel_locker",
         "brand": "Paketbox",
```
kjonosm, line 303: Why is this not `brand=DHL Paketbox`? See https://www.dhl.de/de/privatkunden/pakete-versenden/pakete-abgeben/paketbox.html

Aimmig: As this was already present I can only explain what was discussed or assume:
- These special parcel lockers (which are only for dropping of parcels) are rather rare and date from ~2013. From looking at pictures these had a different branding which didn't include DHL in the branding only as operator.
- To avoid confusion with the way more common "DHL Packstation". In fact some parcel lockers are actually tagged as brand="DHL Paketbox" but wrongly suggest you could also pick up parcels which is not ideal. If someone types DHL and then doesn't know what the differences of these are most likely it just a regular "Packstation"
- the wikidata entry only points to the general type of boxes and I also did not find any other company which has these type of boxes.

PR #12384, opened 2026-06-29
Update Cepsa branding to Moeve

Robot8A:


UKChris-osm:
What about the `displayName`?

Robot8A:
> What about the `displayName`?

fixed

PR #12393, opened 2026-06-30
Create mobile_money_agent add M-Pesa

### LINKED ISSUE #12088: M-Pesa — amenity=mobile_money_agent (001)
```json
{
  "displayName": "M-Pesa",
  "id": "M-Pesa",
  "locationSet": {
    "include": [
      "001"
    ]
  },
  "tags": {
    "amenity": "mobile_money_agent",
    "name": "M-Pesa",
    "brand": "M-Pesa",
    "brand:wikidata": "Q252382"
  }
},
```

**Notes to editors**

Created with https://ga-kevin-codes.github.io/Suggest-A-Brand/
data file: https://github.com/osmlab/name-suggestion-index/blob/main/data/brands/amenity/mobile_money_agent.json

Dwarf-Nebula:
This pull request adds a new categorie (11.000 uses in OSM database) and adds M-Pesa to is. Which closes #12088.

### DIFF
--- a/data/brands/amenity/mobile_money_agent.json
+++ b/data/brands/amenity/mobile_money_agent.json
@@ -1,0 +1,19 @@
+{
+  "properties": {
+    "path": "brands/amenity/mobile_money_agent",
+    "skipCollection": false,
+    "exclude": {"generic": [""]}
+  },
+  "items": [
+    {
+      "displayName": "M-Pesa",
+      "locationSet": {"include": ["ke", "et"]},
+      "tags": {
+      "amenity": "mobile_money_agent",
+      "name": "M-Pesa",
+      "brand": "M-Pesa",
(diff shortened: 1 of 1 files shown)


Kovoschiz:
Question https://github.com/osmlab/name-suggestion-index/issues/12088#issuecomment-4881209279

UKChris-osm:
I don't think adding this file would work at the moment as `mobile_money_agent` isn't part of the tagging scheme (which the NSI uses to match up entries), and looking at comments such as [4273542330](https://github.com/openstreetmap/id-tagging-schema/issues/1641#issuecomment-4273542330), I'm not sure it'll even be added as an option as the 11k entries seem to have been an import dump, rather than organically added.

bhousel:
Ah yeah, if the iD folks are opposing it, let's pick a different tag.

PR #12399, opened 2026-07-02
OTP banka - Slovakia fix, add Slovenia

mnalis:
- fixes wrong entry for Slovakia (`.sk`) - Q13537695 is _"OTP Banka Slovensko"_ which is Slovakian, not Slovenian (yeah they get mixed up often)
- also adds missing Slovenian (`.si`) "OTP Banka"

### DIFF
--- a/data/brands/amenity/bank.json
+++ b/data/brands/amenity/bank.json
@@ -10725,9 +10725,9 @@
       }
     },
     {
-      "displayName": "OTP Banka",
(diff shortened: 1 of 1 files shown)


Oivo35:
OTP banka hasn't operated in Slovakia for several years. Just update the correct Wikidata on the original item.

mnalis:
Ok; I've removed OTP banka in Slovakia, and left the one in Slovenia.

PR #12413, opened 2026-07-09
add Service NSW for au-nsw

andrewharvey:
We have a few different `government=*` values in use, `services`, `register_office`, `social_services`, `public_service`, `administrative`.

* They do provide `register_office` services as one function, but given that's only a small subset of what they provide likely another tag is best.
* I would say they are less likely to be `social_services` as most of that goes through the federal Medicare/Centerlink service centres.
* `administrative` seems more like backend government administration offices than service centers
* `public_service` on the wiki is described as "An office with public services for government affairs" seems best, given `services` isn't even on the list at https://wiki.openstreetmap.org/wiki/Key:government

Personally I also tag these with `admin_level=4` but I haven't included that in this PR.

Discord ref https://discord.com/channels/413070382636072960/471231032645910529/1524680365086740563

### DIFF
--- a/data/operators/office/government.json
+++ b/data/operators/office/government.json
@@ -1954,6 +1954,19 @@
       }
     },
     {
+      "displayName": "Service NSW",
+      "locationSet": {"include": ["au-nsw.geojson"]},
+      "tags": {
+        "office": "government",
+        "government": "public_service",
+        "operator": "Service NSW",
+        "operator:wikidata": "Q29324815",
+        "operator:type": "government",
+        "brand": "Service NSW",
+        "brand:wikidata": "Q29324815"
+      }
+    },
+    {
       "displayName": "Servicio Nacional de Sanidad Agraria del Perú",
       "id": "servicionacionaldesanidadagrariadelperu-b5f2e9",
       "locationSet": {"include": ["pe"]},


UKChris-osm:
I'd suggest removing the `brand` tagging in this PR as the information is repeating that of the `operator `tagging.

andrewharvey:
> I'd suggest removing the brand tagging in this PR as the information is repeating that of the operator tagging.

I'm not sure, but I'm thinking they should be tagged with both, as the branding is "Service NSW" and they are also operated by "Service NSW".

For example with Australia Post post offices/post shops, you can have some which have the "AusPost" branding but are operated by a private company under a license, and others with both with the "AusPost" branding and operated by AusPost. So it's helpful to record both tags as they have different meanings.

Though I do understand that for many branded things, where the operator and brand always go hand in hand it's common to only map one tag. For example a public telephone like https://nsi.guide/index.html?t=*&k=amenity&v=telephone&tt=telstra has `operator=Telstra` but not `brand=Telstra` though I guess adding both would be correct?

For infrastructure like power utilities, water utilities we are less likely to use brand since the concept of branding doesn't really apply in the same was as a fast food chain.

Though in this case, Service NSW brands itself and presents itself like a shop, with strong branding, rather than a generic office location where people work.

This also raises a point about how do we distinguish the customer service centre's for Service NSW, from the  office where Service NSW staff work from. Though they probably they work from within the Department of Customer Service offices, rather than their own dedicated Service NSW offices.

bhousel:
Thanks!  It's ok to have both `brand` and `operator` tags filled in with the same thing if it really does reflect the real world situation.  (Worst case, someone will complain later).

PR #12435, opened 2026-07-16
Add KR fast food brand

chiho-kr:
Add Korea-specific Pizza Hut entry (fast_food) and exclude kr from worldwide restaurant entry.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -137,6 +137,7 @@
     {
       "displayName": "피자헛",
       "locationSet": {"include": ["kr"]},
+      "matchNames": ["Pizza Hut Delivery", "Pizza Hut Express"],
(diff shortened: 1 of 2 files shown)


data/brands/amenity/fast_food.json
```diff
@@ -134,6 +134,22 @@
         "takeaway": "yes"
       }
     },
+    {
```
LaoshuBaby, line 137: I'd suggest that you add those 2 name (Pizza Hut Delivery and Pizza Hut Express) to matchNames, to let existing OSM element named with this can be recognised to new one.

https://github.com/osmlab/name-suggestion-index/wiki/Item-Property-Reference#matchNames-matchTags

chiho-kr: Done, Thanks!

PR #12446, opened 2026-07-21
Fixed brand name, it's First Hawaiian Bank not Hawaiian First Bank

TheAdventurer64:


Cj-Malone:
Probably best to have a match name so any in OSM get prompted for an update

TheAdventurer64:
> Probably best to have a match name so any in OSM get prompted for an update

How do you add a match name? Does it require updates to the iD Validator?

Cj-Malone:
It's just needed here, in the entry. https://github.com/osmlab/name-suggestion-index/pull/12446/commits/65cdbf7ae7cc5e499d41b51dd09cdb28626cf884

PR #12449, opened 2026-07-22
update to id-tagging-schema v7

k-yle:
[id-tagging-schema](https://github.com/openstreetmap/id-tagging-schema) recently released v7, which is the first major release in 3 years. 

~~NSI does not need to do anything~~ (see [the migration guide](https://github.com/openstreetmap/id-tagging-schema/blob/main/MIGRATION_GUIDE.md) for more context).


tyrasd:
> NSI does not need to do anything

actually, I think `buildFields` does need to be adjusted ([here](https://github.com/k-yle/name-suggestion-index/blob/c90f03d632db067aa134929700bdb3b8131e8d94/lib/presets_id.ts#L185)). Currently, iD does not load the fields of NSI presets correctly:

UKChris-osm:
Thanks for this - though as my scripting skills are low I'll wait for more experienced coders / maintainers to review & action this PR.

lib/presets_id.ts
```diff
@@ -178,10 +179,13 @@ function collectTerms(
...
  */
-function buildFields(t: string, presetID: string, preserveTags: string[]): string[] | undefined {
+function buildFields(t: string, preset: Preset & { originalFields?: string[] }, preserveTags: string[]): string[] | undefined {
   if (!preserveTags.some(s => s === '^name')) return undefined;
-  if (t === 'brands')    return ['name', 'brand', `{${presetID}}`];
-  if (t === 'operators') return ['name', 'operator', `{${presetID}}`];
+  // `originalFields` is preferred over `fields` to be backwards compatible with old versions
+  // of iD (released between May 2026 and July 2026). This can eventually be removed once there
+  // is no one using these old versions of iD.
+  if (t === 'brands')    return ['name', 'brand', ...(preset.originalFields || preset.fields || [])];
```
k-yle: this part is annoying. iD currently misuses `buildIDPresets()`, see [openstreetmap/iD#12610](https://github.com/openstreetmap/iD/pull/12610) for details. 

Three options:
1. temporarily add this `originalFields` hack, to make the migration smoother. 
2. do not add this `originalFields` hack. Instead, wait for [openstreetmap/iD#12610](https://github.com/openstreetmap/iD/pull/12610) to be merged and released, and accept that old versions of iD will be broken
3. treat this PR as a breaking change, and release it as v8, so that old versions of iD won't use it.


This PR currently implements option 1

k-yle (commented):
good point, PR updated

lib/presets_id.ts
```diff
@@ -178,10 +179,13 @@ function collectTerms(
...
+  if (t === 'operators') return ['name', 'operator', ...(preset.originalFields || preset.fields || [])];
```
tyrasd: If I remember correctly, it should also handle the case where `t` is not brands/operators (which previously used to rely on the implicit field inheritance from the respective parent preset)
```suggestion
  if (t === 'brands')    return ['name', 'brand', ...(preset.originalFields || preset.fields || [])];
  if (t === 'operators') return ['name', 'operator', ...(preset.originalFields || preset.fields || [])];
  return preset.originalFields || preset.fields || [];
```

lib/presets_id.ts
bhousel, line 248: Is this still true?   The removal of implicit inheritance is my main concern about the v7 upgrade.

bhousel:
Thanks @k-yle - I'll just merge as-is and see what happens.

lib/presets_id.ts
k-yle: this comment is out of date now
- this was discussed in https://github.com/openstreetmap/schema-builder/issues/204
- presets were mass-updated in https://github.com/openstreetmap/id-tagging-schema/pull/1855
- iD code was deleted in https://github.com/openstreetmap/iD/pull/12546

k-yle:
https://pr-12644--ideditor.netlify.app - if anyone wants to test NSI v8 in iD

PR #12454, opened 2026-07-24
chore: update deps and use common Bun version

Johannes-Andersen:
PR consists of:
- update `@openstreetmap/id-tagging-schema`, `typescript-eslint`, and transitive dependencies
- use the pinned Bun version consistently across GitHub Actions
	- The workflow checks the packageManager field and selects the correct version, ref: https://github.com/oven-sh/setup-bun#usage
- improve workflow step names

### DIFF
--- a/.github/workflows/build-commit.yml
+++ b/.github/workflows/build-commit.yml
@@ -12,24 +12,18 @@ jobs:
   build-nsi:
     runs-on: ubuntu-latest
 
-    strategy:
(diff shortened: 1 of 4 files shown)


bhousel:
Thanks for this, but I'm going to close it because #12449 already includes the update for `@openstreetmap/id-tagging-schema` and that dependency requires a bunch more review than any of the other things.

PR #12480, opened 2026-08-01
Tax, customs and border control offices (Poland)

rogoyski:
This PR replaces #12312, which got prematurely closed, which in turn replaced #12006, whose author abandoned it.

For visibility, let me repeat my comment about closing #12312:

I'm fine with the verdict on _Administracja Celno-Skarbowa_, but then you could at least let me merge the _Straż Graniczna_ entry. Instead, the whole PR is closed, while for example the original one by @DevManiak17 still isn't 😕 I will resubmit it.

I also must note that by calling these _duplicate entries in the same location_ you seem not to acknowledge that they are only duplicate from the perspective of NSI's technical design. From the perspective of OpenStreetMap and mapping workflows in general, which I guess you are trying to facilitate, both are distinctive and valuable real-world types of features (POIs). So ultimately it is arguably a flaw in NSI's design that standardising both entries is not possible. I understand you will not turn the design up side down because of this case, but if I were in your place, I would acknowledge this when declining a PR.

### DIFF
--- a/data/operators/office/government.json
+++ b/data/operators/office/government.json
@@ -1180,24 +1180,34 @@
     },
     {
       "displayName": "Krajowa Administracja Skarbowa (urzędy i oddziały celne)",
+      "id": "krajowaadministracjaskarbowa-995fc9",
       "locationSet": {"include": ["pl"]},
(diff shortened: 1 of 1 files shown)


bhousel:
Thanks, this seems ok 👍

PR #12490, opened 2026-08-03
Added hk, mo to location_exclude

ckpiper:
Adding to excludes as they have there own entry.


LaoshuBaby:
The same reason as #12491 and also suggest to close this 

https://github.com/osmlab/name-suggestion-index/blob/d208dd2f3079995010abd79ffd775d9b99c384c3/data/brands/amenity/restaurant.json#L10651-L10689

PR #12491, opened 2026-08-03
Added hk & mo to global starbucks exclude entry

ckpiper:
Added hk & mo to the global starbucks location.exludes list as hk & mo have their own entry in NSI.

### DIFF
--- a/data/brands/amenity/cafe.json
+++ b/data/brands/amenity/cafe.json
@@ -4402,7 +4402,7 @@
     },
     {
       "displayName": "Starbucks",
-      "id": "starbucks-795f60",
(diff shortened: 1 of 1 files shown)


LaoshuBaby:
I guess there are no need to make this changes because HK and MO are already excluded in Mainland of China entry, and Mainland of China entry was already excluded in global one as you seen, so this is a chain relation, the global one will also not have effect in those areas, while the localized one effect.

https://github.com/osmlab/name-suggestion-index/blob/19ee2da5585d34752dc50eab9cbbfee60a85c156/data/brands/amenity/cafe.json#L7593-L7628

So I suggest close this PR. Also for #12490

PR #12508, opened 2026-08-09
Add Dreammedia

thomraider12:
Portuguese billboard company

### DIFF
--- a/data/operators/advertising/billboard.json
+++ b/data/operators/advertising/billboard.json
@@ -468,8 +468,7 @@
         "advertising": "billboard",
         "operator": "Dreammedia",
         "operator:wikidata": "Q140956057",
-        "message": "commercial",
-        "contact:phone": "+351 808200600"
+        "message": "commercial"
       }
     },
     {
--- a/data/operators/advertising/billboard.json
+++ b/data/operators/advertising/billboard.json
@@ -461,6 +461,18 @@
       }
     },
     {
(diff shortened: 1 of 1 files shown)


data/operators/advertising/billboard.json
```diff
@@ -460,6 +460,18 @@
...
+    {
+      "displayName": "Dreammedia",
+      "id": "dreammedia-28ca1f",
+      "locationSet": {"include": ["pt"]},
+      "tags": {
+        "advertising": "billboard",
+        "operator": "Dreammedia",
+        "operator:wikidata": "Q140956057",
+        "message": "commercial",
+        "contact:phone": "+351 808200600"
```
LaoshuBaby: Sorry we don't record any detailed contact in NSI, if this is the headquater's contact, you can add it to related wikidata. According to https://github.com/osmlab/name-suggestion-index/wiki/Judge-Case#adding-specific-keys

LaoshuBaby (changes requested):

data/operators/advertising/billboard.json
thomraider12: Done.

bhousel:
Thanks, I merged and also removed the `"message": "commercial"` tag in 851ee2f21
The [Key:message](https://wiki.openstreetmap.org/wiki/Key:message) page on the OSM wiki says that this is the default value, so it seems unnecessary.

PR #12509, opened 2026-08-09
Update union.json

hamdigumus:
added a union data to nsi category


hmdgms (approved):
Added turkish education union

PR #12518, opened 2026-08-10
Add Shell Mobility

Yivan000:
"Shell Mobility" is a name for some new branches of Shell in the Philippines that have additional amenities.

Adding this also to reduce edit wars at OSM over removing the "Mobility" name even though on-the-ground signages have that.

### DIFF
--- a/data/brands/amenity/fuel.json
+++ b/data/brands/amenity/fuel.json
@@ -6103,6 +6103,20 @@
       }
     },
     {
+      "displayName": "Shell Mobility",
+      "id": "shell-ph1898",
+      "locationSet": {
+        "include": ["ph"]
+      },
+      "note": "Philippine variant of Shell with additional amenities",
+      "tags": {
+        "amenity": "fuel",
+        "brand": "Shell",
+        "brand:wikidata": "Q110716465",
+        "name": "Shell Mobility"
+      }
+    },
+    {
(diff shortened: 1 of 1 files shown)


Cj-Malone:
Is this the fuel station or the shop attached? I had a little search and only it's franchise thing showed up.

Yivan000:
Fuel station. It's just Shell but these branches are named "Shell Mobility".

Example branch:

[From Street View](https://maps.app.goo.gl/4JwPBjxnExih9iis8):
image
[Place in OSM wherein a user removed the "Mobility"](https://www.openstreetmap.org/way/1420187510)

Cj-Malone:
They removed it from brand (which I currently agree with) not name. https://osm.mapki.com/history/way/1420187510

This PR adds it as a name, but leaves the standard brand. (I think this is better than the OSM edit). But I don't think this will actually do anything, or maybe make iD worse. Best case scenario, iD acts the same, it currently allows you to add any name to fuel stations. Worst case, iD isn't the greatest at matching NSI.. it often ends in a endless argument about which entry to match. "Upgrade to Shell" "Upgrade to Shell Mobility" "Upgrade to Shell"... 

https://find.shell.com/ph/fuel/12983551-sh-tplex-rosario-la-union/en_PH

PR #12520, opened 2026-08-10
Update Innogy

Oivo35:
Innogy operates only in the Czech Republic, and the other European branches have been rebranded.


bhousel:
I'm not sure it's ok to delete the German Innology without adding some kind of matching tag where it's been rebranded.

Oivo35:
Innogy was acquired by E.ON, but I've restored the deleted entry as not all original charging stations in Germany can carry the E.ON brand yet. I just added the country to the name to differentiate it.

PR #12526, opened 2026-08-12
nsiguide: Force display: none on .hide CSS class

### LINKED ISSUE #12466: New Search feature on the NSI Web Site - bug or by design?
I'm not a fan of the new search feature on the web site, where it displays every category on screen, even if the search finds 0 results. I found the previous method of hiding each category when it returned 0 results to be a lot easier to use.

Is this a new feature by design, or an unintended bug?

Image

willemarcel:
The empty categories are not getting hidden when a filter is applied ([example](https://nsi.guide/?t=*&tt=geely)). This PR forces `display: none` on the `.hide` CSS class.

image


bhousel:
fixes #12466

PR #10842, opened 2025-04-01
[amenity/fast_food] add McDonald's Ukraine

riwnodennyk:

PR #12529, opened 2026-08-13
Modify Philippine fastfood chains

Yivan000:
Modified Philippine fastfood chains:
* Chowking is `"cuisine": "chinese;filipino"`
* Jollibee is `"cuisine": "burger;chicken"`
* McDonald's Philippines has a separate Wikidata, is `"cuisine": "burger;chicken"`, and has `"short_name": "McDo"`.

### DIFF
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -5341,7 +5341,7 @@
         "amenity": "fast_food",
         "brand": "Jollibee",
         "brand:wikidata": "Q37614",
-        "cuisine": "burger",
+        "cuisine": "burger;chicken",
         "name": "Jollibee",
         "takeaway": "yes"
       }
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -2165,7 +2165,7 @@
         "amenity": "fast_food",
         "brand": "Chowking",
         "brand:wikidata": "Q1076816",
-        "cuisine": "asian",
+        "cuisine": "chinese;filipino",
         "name": "Chowking",
         "takeaway": "yes"
       }
--- a/data/brands/amenity/fast_food.json
+++ b/data/brands/amenity/fast_food.json
@@ -6776,7 +6776,7 @@
       "tags": {
         "amenity": "fast_food",
         "brand": "McDonald's",
-        "brand:wikidata": "Q60775813",
+        "brand:wikidata": "Q38076",
         "cuisine": "burger;chicken",
         "name": "McDonald's",
         "short_name": "McDo",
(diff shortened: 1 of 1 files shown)


data/brands/amenity/fast_food.json
```diff
@@ -6768,6 +6769,20 @@
...
       }
     },
+    {
+      "displayName": "McDonald's (Philippines)",
+      "id": "mcdonalds-ph1898",
+      "locationSet": {"include": ["ph"]},
+      "tags": {
+        "amenity": "fast_food",
+        "brand": "McDonald's",
+        "brand:wikidata": "Q60775813",
```
1ec5: McDonald’s has divisions in all the countries they operate in, and then some. These restaurant locations are operated by the McDonald’s Philippines division, or are franchised by it, but do they have a brand distinct from the main McDonald’s brand?

data/brands/amenity/fast_food.json
```diff
@@ -6704,6 +6704,7 @@
           "ma",
           "no",
           "om",
+          "ph",
```
1ec5, line 6707: By both adding the Philippines to this list and adding the entry below, you’re giving the user a choice between “McDonald’s” and “McDonald’s (Philippines)” whenever they map a new restaurant. And if a restaurant is ambiguously tagged with the name but no brand tag, iD will avoid suggesting a tag upgrade because of the ambiguity.

Yivan000: Oh, perhaps I misread the documentation on this one. I thought by adding "ph" here, I exclude “McDonald’s” from the list and only “McDonald’s (Philippines)” will show.

data/brands/amenity/fast_food.json
Yivan000: No, it generally follows the main McDonald's brand. 

I was just following the entry right before this ("McDonald's (Україна)") wherein it added its separate McDonald's Ukraine Wikidata item. From my limited research, McDonald's Ukraine also does not have a distinct brand, yet its entry lists its separate Wikidata item.

data/brands/amenity/fast_food.json
1ec5: Oh, you’re right, it is the exclude list. I misread the diff.

data/brands/amenity/fast_food.json
1ec5: I see, that was introduced in #10842, which turned out to be quite a controversial change. The name got undone, but it looks like the Wikidata QID got overlooked amid all the debate. This separate item is a good idea if “McDo” is common shorthand for the brand name, since that isn’t the case globally. But it should share the same QID as the main entry.

Yivan000: I have now changed the Wikidata item to be in line with the main entry. 

In addition to "McDo" being a common shorthand, the cuisine is also `"burger;chicken"`, as opposed to the main entry `"burger"`.

PR #12533, opened 2026-08-16
add Flip Out for AU as leisure=trampoline_park

andrewharvey:
I opted to use a new wikidata ID for the AU business as a separate business entity, sharing the same branding.

cc @UKChris-osm (who added the UK one)

I got a check error

> build | build:json | Error - duplicate displayName 'Flip Out' in:

therefore I've specified the countries within the `displayName`, but ideally we could just use the name "Flip Out" for both given the `locationSet` is disjoint.

### DIFF
--- a/data/brands/leisure/trampoline_park.json
+++ b/data/brands/leisure/trampoline_park.json
@@ -29,7 +29,7 @@
       }
     },
     {
-      "displayName": "Flip Out",
+      "displayName": "Flip Out (UK)",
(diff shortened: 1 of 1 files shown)


bhousel:
> therefore I've specified the countries within the `displayName`, but ideally we could just use the name "Flip Out" for both given the `locationSet` is disjoint.

Yes, you got it  👍  
The Display Name is mostly so that they look different in a list of presets.

PR #12534, opened 2026-08-16
Add Archaeological Survey of India

Robot8A:


bhousel:
This adds new categories for  `"historic": "archaeological_site"` and `"tourism": "museum"`.
I think that's ok, just wanted to mention that these are new things we hadn't tracked before.

PR #12584, opened 2026-08-21
Update trolley_bay.json : Delete Chariots

remi-sim:
"Chariots" is the french term for trolleys. This is not a brand at all.

### DIFF
--- a/data/brands/amenity/trolley_bay.json
+++ b/data/brands/amenity/trolley_bay.json
@@ -4,7 +4,8 @@
     "exclude": {
       "generic": [
         "^cart return$",
-        "^trolley bay$"
+        "^trolley bay$",
+        "^chariots?$"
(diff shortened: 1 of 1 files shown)


data/brands/amenity/trolley_bay.json
bhousel: Additionally we should add `"^chariots?$"` in here, so that the matcher is aware that it's a generic word.   The validator in Rapid and iD can warn the user if they use generic words in the `name` tag.

bhousel (changes requested):

data/brands/amenity/trolley_bay.json
remi-sim: Done

bhousel: thanks!

PR #12585, opened 2026-08-21
Update public_bookcase.json : Generic name

remi-sim:


data/brands/amenity/public_bookcase.json
```diff
@@ -6,7 +6,8 @@
         "^(offener|öffentlicher)?\\s?bücher(box|schrank|zelle)$",
         "^bo[îi]te [aà] liv?res?$",
         "^knihobudka$",
-        "^public bookcase$"
+        "^public bookcase$",
+        "^cabane [aà] livres?$"
```
bhousel, line 10: oh nice, looks like for this one you added the generic regex - that was the thing I commented on #12584 about.

remi-sim: I understood how it works with this one 🙂. I will update my trolley bay commit.

PR #12592, opened 2026-08-23
Update supermarket.json

cheesecak3h:


data/brands/shop/supermarket.json
```diff
@@ -7717,7 +7717,7 @@
         "brand": "Sheng Siong",
         "brand:wikidata": "Q3481878",
         "name": "Sheng Siong",
-        "name:zh": "昇松超市",
```
bhousel, line 7720: Should probably move this old name into `matchNames: [...]` so that it is still matched if found.

bhousel (changes requested):

PR #12597, opened 2026-08-23
"Split" Aldi Süd into Aldi Süd and Aldi Suisse

habi:
Remove location `ch` from Aldi Süd, add https://www.wikidata.org/wiki/Q111030009 as Aldi Suisse


bhousel:
Thanks, in 068c4a0b4 I also applied this change to the other categories with Aldi stuff:

- amenity: charging_station
- man_made: surveillance
- power: generator

PR #12619, opened 2026-08-27
Add tinyfish, a Swiss sushi fast food chain

habi:
15 branches according to their website, all cards-only


Robot8A:
Hi, why does the name start with lowercase, shouldn't it start with uppercase instead?

habi:
> Hi, why does the name start with lowercase, shouldn't it start with uppercase instead?
> 
I added it as such because it’s signed in all lowercase on the branch I visited.

PR #12630, opened 2026-08-29
Point San Antonio and San Diego libraries at their library systems

watmildon:
Two entries named a city rather than the library system that runs the libraries, and carried the city's Wikidata QID:

  City of San Antonio -> Q975, the city itself (P31: big city)
  City of San Diego   -> Q138816781, municipal government of California

Replace them with the actual systems, and add San Diego County Library, which is a separate system from San Diego Public Library:

  San Antonio Public Library  Q5642025
  San Diego Public Library    Q5486355
  San Diego County Library    Q7413630

I  have already moved San Diego branches off the city QIDs.

Also change Austin and Houston Public Library from operator:type=government to operator:type=public, which makes operator:type uniform across US entries.

### DIFF
--- a/data/operators/amenity/library.json
+++ b/data/operators/amenity/library.json
@@ -8,6 +8,10 @@
         "^gemeinde$",
         "^mairie$",
         "^public$"
+      ],
+      "named": [
+        "^city of san antonio$",
+        "^city of san diego$"
       ]
     }
   },
@@ -94,7 +98,7 @@
       "tags": {
         "amenity": "library",
         "operator": "Austin Public Library",
-        "operator:type": "government",
+        "operator:type": "public",
         "operator:wikidata": "Q69491633"
       }
     },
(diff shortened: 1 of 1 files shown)


michaelblyons:
What does the `"named"` change do? Did you mean to put those in matches for the respective systems?

Otherwise LGTM.

watmildon:
My impression was that is says "this is not an acceptable operator name" to keep folks from adding it back. But it's my first time in the repo so maybe I don't understand that bit. :-D

michaelblyons:
It has been a hot minute since I made NSI updates, but I think you're looking for something like [the `matchNames` on Auckland's library](https://github.com/osmlab/name-suggestion-index/blob/main/data/operators/amenity/library.json#L75).

I think the generic names thing is for libraries *named* "Library" or sports centers *named* "Sports Center." (If it works differently for operators, I apologize.)

watmildon:
Oh all good. I will poke around some.

It's also potentially just overkill?

watmildon:
Okay, I have bailed on the wacky thing I misunderstood. I think it's overkill anyway as all of those tags were cleared from the DB by me a few weeks ago.

michaelblyons (approved):

PR #12636, opened 2026-08-30
Update locationSet for brands that are not global.

mutipg:
According to data on Wikidata and on brand websites, they are limited to an area smaller than the entire world.

### DIFF
--- a/data/brands/shop/car_repair.json
+++ b/data/brands/shop/car_repair.json
@@ -645,6 +645,18 @@
       }
     },
     {
+      "displayName": "GlassAuto",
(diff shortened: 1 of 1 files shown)


Robot8A:
JCDecaux is also present in Spain, I have seen advertising signs operated by it

mutipg:
> JCDecaux is also present in Spain, I have seen advertising signs operated by it

Thanks, 
I've found it also in Warsaw Airport https://www.jcdecauxairport.com.pl/ - I'm dismiss this PR. I need to check that once more.

PR #12648, opened 2026-08-31
Update bus.json & tram.json: unify Kraków's bus and tram entries to be uniform

TranslatorPS:
This PR aims to unify the two networks' entries. Functionally, the bus and tram networks are part of one entity (same authority, same fares, same rules, for the most part the same operator), but their entries have been different in practice. This PR introduces a new Wikidata item and unifies the tagging across the two.

There is no operator added to the bus.json entry as there are two bus operators contracted in Kraków.


Robot8A:
Looks good, thanks!

PR #12664, opened 2026-09-04
Libraries/add us library systems

watmildon:
This adds library operators for systems that have 10+ branches already in OSM. Many of these still have missing branches in OSM so it would be nice for folks to get offered the correct tagging when they get added.


michaelblyons:
lovelovelove!

PR #12665, opened 2026-09-04
Add Evolve E-Bike Share

maxwellward:
Adds the Evolve E-Bike Share program rental stands. BCAA owns Evo Car Share which operates the Evolve E-Bike Share program.

"Evolve E-Bike Share" is the program name, but the shortened "Evolve" is the brand name.

This service only operates in British Columbia, Canada.


Robot8A:
Looks good to me!

PR #12671, opened 2026-09-04
Remove Tesla Destination Charger and update Tesla DCFC

GA-Kevin-Codes:
Removed Tesla Destination Charger entry and updated Tesla DCFC tags.

### DIFF
--- a/data/brands/amenity/charging_station.json
+++ b/data/brands/amenity/charging_station.json
@@ -1493,18 +1493,6 @@
       }
     },
     {
-      "displayName": "Tesla Destination Charger",
-      "id": "tesladestinationcharger-4bd82e",
-      "locationSet": {"include": ["001"]},
-      "matchNames": ["tesla wall connector"],
-      "tags": {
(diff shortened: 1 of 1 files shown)


data/brands/amenity/charging_station.json
```diff
@@ -1492,18 +1492,6 @@
...
-      "matchNames": ["tesla wall connector"],
-      "tags": {
-        "amenity": "charging_station",
-        "brand": "Tesla",
-        "brand:wikidata": "Q478214",
-        "name": "Tesla Destination Charger"
-      }
-    },
     {
       "displayName": "Tesla Supercharger",
```
bhousel: This should just say "Tesla" now, since going forward we won't differentiate between different kinds of charging stations.

data/brands/amenity/charging_station.json
bhousel: also should add the old names here:  "tesla supercharger", "tesla destination charger", "tesla wall connector"

PR #12681, opened 2026-09-07
Fix MN Metro Transit operator

cfinn68916:
The operator tag for metro transit is currently BNSF railroad. This is probably a holdover from the [Northstar Commuter Line](https://en.wikipedia.org/wiki/Northstar_Line), which has been closed since Jan 2026. The only currently operating rail routes in the Metro Transit network are light rail, which is operated by Metro Transit itself.

### DIFF
--- a/data/transit/route/train.json
+++ b/data/transit/route/train.json
@@ -1539,8 +1539,8 @@
       "tags": {
         "network": "Metro Transit",
         "network:wikidata": "Q3307442",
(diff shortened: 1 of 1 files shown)


Robot8A:
If there are no more train routes, then the entry should be removed (or changed to dissolved). The entry for light rail should then be kept.

bhousel:
matchName the old name, dont' just delete it

PR #12697, opened 2026-09-09
Update sports.json : Add Skiset

remi-sim:


Robot8A:
The Wikidata identifier seems like it points to the wrong element

remi-sim:
> The Wikidata identifier seems like it points to the wrong element

My mistake, thanks for the warning.

Robot8A:
Looks good now, merging!

PR #12705, opened 2026-09-10
Update Moneta Bank

Oivo35:
Updated the brand and operator names for the bank and ATM.

### DIFF
--- a/data/brands/amenity/atm.json
+++ b/data/brands/amenity/atm.json
@@ -10,6 +10,7 @@
       "templateExclude": [
         "abnamro",
         "bankinghub",
+        "monetabank",
         "monobank",
         "postfinance"
       ],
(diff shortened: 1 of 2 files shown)


Cj-Malone:
Why does this need to be special cased in data/brands/amenity/atm.json?

Oivo35:
The brand and operator names are different. The bank has rebranded, but the operator remains the same. Consequently, iD suggests the incorrect brand name for the ATMs.

bhousel:
> The brand and operator names are different. The bank has rebranded, but the operator remains the same. Consequently, iD suggests the incorrect brand name for the ATMs.

Oh yeah, it's kind of weird for the same QID to have different `brand` and `operator` English-language tags.  I understand what you're saying "the bank has rebranded but the operator remains the same" and, sure we can do this, but it's just weird.  I feel like it would be fine to just call it "Moneta Bank" everywhere.

Oivo35:
> I feel like it would be fine to just call it "Moneta Bank" everywhere.

The consensus within the Czech community is that the `operator` tag should reflect the official name in the Czech companies register, which is still `Moneta Money Bank`, whilst the `brand` name is `Moneta Bank`.

PR #12708, opened 2026-09-10
Update Czech bus networks

Oivo35:
Add MHD Valašské Meziříčí network and remove outdated entries.

### DIFF
--- a/data/transit/route/bus.json
+++ b/data/transit/route/bus.json
@@ -29,6 +29,19 @@
   },
   "items": [
     {
+      "displayName": "MHD Valašské Meziříčí",
+      "locationSet": {
+        "include": ["cz-72.geojson"]
+      },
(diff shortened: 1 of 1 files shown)


Cj-Malone:
The outdated entries were added back because they weren't added as a match name or excluded. It will continue to happen until they are added as match names or excluded, or they are removed from osm (and the osm cache is updated).

Oivo35:
These outdated networks were removed from OSM at least six months ago. Should they still be added as matchNames?

bhousel:
> These outdated networks were removed from OSM at least six months ago. Should they still be added as matchNames?

Yes let's leave them for now.. We'd need to re-run the collection script again to get fresh names from the planet file.
This is annoying to do, so I'm not surprised that it hasn't happened in 6 months.

PR #12710, opened 2026-09-10
Remove Supersol (ES) from supermarket.json: brand dissolved into Carr…

mjoe999:
Carrefour bought all 172 Supersol stores in 2020-21 and converted them to Carrefour Express, Carrefour Market and Supeco by end of 2021 (38 were sold on to Cash Lepe, Froiz, Maskom and Ecomor). The Supersol brand no longer trades anywhere.

Wikidata https://www.wikidata.org/wiki/Q62073427 has been amended to reflect the dissolution.

### DIFF
--- a/data/brands/shop/supermarket.json
+++ b/data/brands/shop/supermarket.json
@@ -8625,19 +8625,6 @@
       }
     },
     {
-      "displayName": "Supersol",
-      "id": "supersol-0a1807",
-      "locationSet": {
-        "include": ["ar", "es", "ma", "uy"]
-      },
-      "tags": {
-        "brand": "Supersol",
-        "brand:wikidata": "Q62073427",
-        "name": "Supersol",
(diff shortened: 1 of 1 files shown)


Crashillo:
But you don't need to remove the brand since the wikidata item contains the date dissolution, right?

Cj-Malone:
I think in a perfect world, we'd keep dissolved brands in here. It's good to have for QA purposes. In that perfect world, I'd change the details (why's it in a separate dist file? why's on a qcode level, not item level?), we need a better source (Wikidata isn't great here, it mixes brands and companies).

But this isn't perfect, iD and Rapid couldn't (probably still can't in some brand/country combos) add brands in GB for years because they were dissolved globally in Wikidata. Other editors don't use the extra dissolved file, so still see the brands. Removing them is an option that we do sometimes. It's not perfect, it is what it is.

PR #12742, opened 2026-09-16
Update of the citiz brand and its local operators

MeGilles:
The Citiz brand and its local operators have changed over the past few years.

This PR clarifies the relationship between the national Citiz cooperative—which owns the brand, website, and reservation platform—and the local operators, which are public companies or cooperatives responsible for managing the cars and car-sharing parking spaces (amenity=car_sharing).


Robot8A:
Is the PR ready for review?

Cj-Malone:
I don't think this will really work in iD. It'll probably end up being 1 brand entry.

bhousel:
Yeah I haven't tried building it, but I'd expect we would get duplicate key issues because of multiple things named "cititz" in "fr-pdl" and maybe other issues.

Maybe I'm wrong though?  I'm surprised the CI build didn't raise this.

PR #12743, opened 2026-09-16
Add EVN (България) charging station entry

vzaimov500:
Add EVN (България) to operators/amenity/charging_station
Adds the Bulgarian EVN charging network as an operator entry.

Wikidata: Q20498243 (EVN, Bulgarian energy company)
Website: https://www.evn.bg

This QID is already used in NSI at operators/power/generator
and operators/office/energy_supplier.

Follows the naming pattern of the existing EVN (Österreich)
and EVN (Северна Македонија) entries.

Note: the EVN2GO app is a roaming platform that also lists
other operators' stations (e.g. EVPoint), so it is deliberately
not included as a matchName.


vzaimov500:
WhatsApp Image 2026-09-17 at 13 03 39

Added brand tags after surveying a unit (photo attached).

The charger carries the EVN logo as its primary marking, so
`brand=EVN` applies alongside the operator tags. Both point to
the same Wikidata item since EVN is both brand and operator
here — same pattern as the ChargePoint and Circle K entries.

No EVN2GO branding appears on the hardware, which supports
leaving it out as discussed above.

Scoped to Bulgaria only — I haven't surveyed the Austrian or
North Macedonian units, so I'm not proposing the same change
for those entries.

Dimitar5555:
Thanks for the PR! I'm not sure if having both `operator` and `brand` tags with the same values is really useful.

vzaimov500:
Thanks, that's fair. I checked the data: the operators file is operator-only across all 732 entries, so brand tags there would be the only exception. The two closest entries, EVN (Österreich) and EVN (Северна Македонија), are operator-only as well, so
following that pattern makes sense.

The brands file is different. 
Of 136 entries, 
58 carry both brand and operator with the same value, 
55 are brand-only, 
23 carry both with different values.

So if the brand tags are worth keeping, this entry belongs in the brands file rather than the operators one. Happy to move it
there, or to just drop the brand tags and leave it as an operator entry. Which would you prefer?

Dimitar5555:
Having brand tags only and moving it to the brand tree is the best option. It also aligns with the other brands in Bulgaria (see https://nsi.guide/index.html?t=*&k=amenity&v=charging_station&cc=bg).

vzaimov500:
Makes sense, thanks. Closing this and opening a brand entry instead.

PR #12747, opened 2026-09-17
Add operator tags to eMotion (amenity=charging_station)

vzaimov500:
The eMotion brand entry currently fills in brand tags only, so a
mapper selecting it gets no operator. A separate "OMV eMotion"
entry in operators/amenity/charging_station fills in operator=OMV
but no brand — meaning eMotion surfaces as two presets and neither
gives the complete tagging.

This adds the operator tags to the brand entry, matching how
ChargePoint, Circle K and Tesla Supercharger carry both pairs.

OMV describes eMotion as its own charging brand
(https://www.omv.com/en/expertise/fuels/electric-mobility), so
brand=eMotion with operator=OMV reflects the relationship.

Leaving the operators entry untouched — it is scoped worldwide
and may deliberately cover OMV-operated chargers under the
Petrom electric and E-charge brands. Happy to follow up if
maintainers would rather consolidate.


Dimitar5555:
Thanks!

Oivo35:
Some communities prefer using the official national company name for the operator=* tag, such as operator=OMV Slovensko in Slovakia.  To avoid creating multiple country-specific entries, only the entry with `brand=eMotion` is used, omitting the `operator=*` tag.

[OMV eMotion](https://nsi.guide/index.html?t=operators&k=amenity&v=charging_station#omv-c22d3a) entry with `operator=OMV` was added automatically as these tags still occur in OSM, and it should therefore be merged with the newer entry.

PR #12749, opened 2026-09-17
Add eCars (amenity=charging_station, bg)

vzaimov500:
Adds eCars, a Bulgarian EV charging brand.

Wikidata: Q141489203
Website: https://ecars.bg
Legal entity: Екарс БГ ЕООД, EIK 204632421, registered 2017

Brand tags only, no operator. eCars manufactures charging
stations and operates the VsichkoTok station locator
(https://vsichkotok.bg). Their terms distinguish stations managed
by that platform from stations managed by other operators via
e-roaming, so the locator is not evidence of who operates a given
station. eCars does operate some stations directly, but not all
eCars-branded hardware is eCars-operated.

Note: ESB ecars (Q134882112, Ireland) is a different company.

Same pattern as the EVPoint and Tesla Destination Charger
entries.


Dimitar5555:
I'm not sure about this one. Their stations seem to be branded by whoever bought them. See the photos on their [facebook profile](https://www.facebook.com/eCars.bg/photos).

It might be a better fit for `manafacturer=eCars` but I don't think that we store such information here.

vzaimov500:
You're right, and thanks for checking the photos

PR #12759, opened 2026-09-18
Add EVN (България) to brands/amenity/charging_station

vzaimov500:
Moving this from #12743 as suggested.
Wikidata: Q20498243 (EVN, Bulgarian energy company)
Brand tags only, matching the other Bulgarian entries.


Dimitar5555:
Thanks!

Issue #12744, opened 2026-09-16
Public transport around Nantes (France)


Mbodin:
Hi,
The public transport network previously named TAN in Nantes has been renamed to NaoLib.
Currently, iD isn't aware of this change and suggest to wrongly rename any `network=NaoLib` to `network=TAN`.
I think that it corresponds to Wikidata's Q17519150.
Have a nice day,
Martin.

Robot8A:
Created the PR from your commit, and merged. Thanks!

PR #12760, opened 2026-09-18
Update name of TAN to Naolib (solves #12744)

Robot8A:


bhousel:
You should `matchName` the old name, so that the validators suggest an upgrade.

bhousel:
fixed this in 5087e6bfc

Robot8A:
Thanks @bhousel, missed that

PR #12761, opened 2026-09-18
Add STUV

### LINKED ISSUE #12625: Add STUV - Serviço de Transportes Urbanos de Viseu
The Public Transportes company (buses) of Viseu. We can use in bus routes, route_master and public_transports=platforms

network=STUV - Serviço de Transportes Urbanos de Viseu
network:wikidata=Q10370644

Robot8A:
Would close #12625


Robot8A:
Another question would be, should we change https://nsi.guide/?t=transit&k=route&v=bus#muv-fdc094 to match (`network`, `network:short` and `operator`)?
@stgarmm

Robot8A:
If I understand the Wikipedia articles correctly, MUV replaced STUV for everything except a couple of lines. When STUV lines stop existing, we can always add a "dissolved" status, but the STUV lines still exist.

stgarmm:
Hello @Robot8A , yes, it would be possible to make that change to `network` and `network:short` in MUV.

However, since the decision has been made to hold public tenders for the concession of various transportation networks in Portugal - and MUV is no exception - the `operator` field should not be used, as it is not always the same. This would allow for local modifications when necessary, without having to make changes to this database. 

Everything indicates that the STUV lines will continue to exist for at least a few more years.

Robot8A:
Fixed

PR #12777, opened 2026-09-22
use brand instead of operator for Australia Post amenity=post_office

### LINKED ISSUE #12731: Tagging issue with Australia Post australiapost-536bec
The [entry](https://nsi.guide/index.html?t=*&k=amenity&v=post_office&tt=australia%20#australiapost-536bec) for Australia Post suggests tagging an `amenity=post_office` with `operator=Australia Post` and `operator:wikidata=Q1142936`. However about [80%](https://www.poaal.com.au/looking-to-buy-a-post-office) of the post offices in Australia are not directly operated by Australia Post, rather they are run by private operators under contract. 

It might be more accurate if these default tags were `brand` and `brand:operator`.

joelkoen:
Fixes #12731


matkoniecz:
do you need help with fixing data? I have run bot edit for https://github.com/osmlab/name-suggestion-index/pull/10076 - if Australian community would approve, I can run similar one here

(unless you already have plans for making one? then feel free to ignore me)

FrakGart:
It's a bit tricky as ~20% of them are operated directly by AusPost, so we can't do a bulk edit without some research before hand. I was more interested in not automatically adding more.

matkoniecz:
I would probably assume that all can be broken and move them to brand, but obviously local community may have a different plan for repair

Cj-Malone:
If the ~20% are identified on the Australia Post website, it's possible ATP could collect operator info.

